PR #598, opened 2016-09-27
DRILL-4906 CASE Expression with constant generates class exception

Serhii-Harnyk:


zfong:
@amansinha100 - can you review this pull request.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillConstExecutor.java
```diff
@@ -143,116 +156,176 @@ public void reduce(RexBuilder rexBuilder, List<RexNode> constExps, List<RexNode>
...
-          case DATE:
-            reducedValues.add(rexBuilder.makeLiteral(
-                new DateTime(((DateHolder) output).value, DateTimeZone.UTC).toCalendar(null),
-                TypeInferenceUtils.createCalciteTypeWithNullability(typeFactory, SqlTypeName.DATE, newCall.getType().isNullable()),
-                false));
-            break;
-          case DECIMAL9:
-            reducedValues.add(rexBuilder.makeLiteral(
-                new BigDecimal(BigInteger.valueOf(((Decimal9Holder) output).value), ((Decimal9Holder)output).scale),
+      Function<ValueHolder, RexNode> literator = new Function<ValueHolder, RexNode>() {
```
amansinha100: Interesting that you are using a Function interface here within the original function.  Is this helping much ? It is not eliminating the need for the large switch statement.  I would think a valid use case for Function interface would be when you are passing a function pointer (borrowing a term from c++) to another function and that function calls apply().   We can keep this as-is but I am curious about the rationale.

exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestCaseNullableTypes.java
```diff
@@ -0,0 +1,107 @@
...
+    testBuilder()
+        .sqlQuery("select (case when (false) then null else 1 end) res1 from (values(1))")
+        .unOrdered()
+        .baselineColumns("res1")
+        .baselineValues(1)
+        .go();
+  }
+
+  @Test
+  public void testCaseNullableTypesVarchar() throws Exception {
```
amansinha100: Its good that you have added tests for different data types.  One suggestion: could you add a test where the constant reduction rule has to do 2 or more iterations .. i.e for 2 or more different data types ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillConstExecutor.java
Serhii-Harnyk: I tried to minimize code under the switch cases. As you can see, before my pull request "reducedValues.add()" calls under any of cases. 
At first I tried to move switch to private method, somethig like "private RexNode getRexNode()" but there was a lot of arguments to pass - typeFactory, builder, etc, etc. Unfortunately we can't use Java8 lambdas, so the function interface(de facto it's nested function) is the optimal solution I think.

exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestCaseNullableTypes.java
Serhii-Harnyk: Done

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillConstExecutor.java
amansinha100: Sounds good.

PR #600, opened 2016-09-30
DRILL-4373: Drill and Hive have incompatible timestamp representations in parquet

vdiravka:
- added sys/sess option "store.parquet.int96_as_timestamp";
- added int96 to timestamp converter for both readers;
- added unit tests;


bitblender:
I can't see NullableFixedByteAlignedReaders.java. Shows up as a binary file.

vdiravka:
@bitblender Sorry about this. That was hidden `\u0000` symbols.
Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -132,6 +132,8 @@
   OptionValidator PARQUET_VECTOR_FILL_CHECK_THRESHOLD_VALIDATOR = new PositiveLongValidator(PARQUET_VECTOR_FILL_CHECK_THRESHOLD, 100l, 10l);
   String PARQUET_NEW_RECORD_READER = "store.parquet.use_new_reader";
   OptionValidator PARQUET_RECORD_READER_IMPLEMENTATION_VALIDATOR = new BooleanValidator(PARQUET_NEW_RECORD_READER, false);
+  String PARQUET_READER_INT96_AS_TIMESTAMP = "store.parquet.int96_as_timestamp";
```
parthchandra: Should rename this to store.parquet.reader.abc to make it clear this is a reader only property

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -754,15 +764,45 @@ public void testImpalaParquetVarBinary_DictChange() throws Exception {
     compareParquetReadersColumnar("field_impala_ts", "cp.`parquet/int96_dict_change.parquet`");
   }
 
+  @Test
+  public void testImpalaParquetBinaryTimeStamp_DictChange() throws Exception {
+    try {
+      test("alter session set %s = true", ExecConstants.PARQUET_READER_INT96_AS_TIMESTAMP);
+      compareParquetReadersColumnar("field_impala_ts", "cp.`parquet/int96_dict_change.parquet`");
```
parthchandra: This is not a good enough test. Both the baseline and test case queries will use the getDateTimeValueFromBinary method and if there is a bug in that method, the test will still pass as both will produce the same incorrect value. Better to compare with the actual baseline value in the file.

vdiravka: 1. Is it better to compare result with baseline columns and values from the file or it is ok to compare with `sqlBaselineQuery` and disabled new `PARQUET_READER_INT96_AS_TIMESTAMP` option?
2. In the process of investigating this test I found that the primitive data type of the column in the file `int96_dict_change.parquet`  is BINARY, not INT96.  
I am a little bit confused with this. Do we need convert this BINARY to TIMESTAMP as well?
CONVERT_FROM function with IMPALA_TIMESTAMP argument works properly for this field.
I will investigate a little more about does impala and hive can store timestamps into parquet BINARY.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -45,4 +53,34 @@ public static int getIntFromLEBytes(byte[] input, int start) {
...
+   *          are stored in parquet Binary as INT96
+   *
+   * @return  the number of milliseconds since January 1, 1970, 00:00:00 GMT
+   *          represented by @param binaryTimeStampValue .
+   */
+    public static long getDateTimeValueFromBinary(Binary binaryTimeStampValue) {
+      NanoTime nt = NanoTime.fromBinary(binaryTimeStampValue);
+      int julianDay = nt.getJulianDay();
+      long nanosOfDay = nt.getTimeOfDayNanos();
+      return DateTimeUtils.fromJulianDay(julianDay-0.5d) + nanosOfDay/NANOS_PER_MILLISECOND;
```
bitblender: Sorry for the late reply. For some reason, I did not see these comments till now. 
About 1) Yes, you are correct. I just want the comments in ConvertFromImpalaTimestamp to be removed.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -739,30 +739,54 @@ public void runTestAndValidate(String selection, String validationSelection, Str
...
+    Impala encodes timestamp values as int96 fields. Test the reading of an int96 field with two converters:
+    the first one converts parquet INT96 into drill VARBINARY and the second one (works while
+    store.parquet.reader.int96_as_timestamp option is enabled) converts parquet INT96 into drill TIMESTAMP.
    */
   @Test
   public void testImpalaParquetInt96() throws Exception {
     compareParquetReadersColumnar("field_impala_ts", "cp.`parquet/int96_impala_1.parquet`");
+    try {
+      test("alter session set %s = true", ExecConstants.PARQUET_READER_INT96_AS_TIMESTAMP);
+      compareParquetReadersColumnar("field_impala_ts", "cp.`parquet/int96_impala_1.parquet`");
```
parthchandra: Github seems to have swallowed the previous comments so including @vdiravka's questions here:

>  1) Is it better to compare result with baseline columns and values from the file or it is ok to compare with sqlBaselineQuery and disabled new PARQUET_READER_INT96_AS_TIMESTAMP option?
> In the process of investigating this test I found that the primitive data type of the column in the file int96_dict_change.parquet is BINARY, not INT96.
> 2) I am a little bit confused with this. Do we need convert this BINARY to TIMESTAMP as well? CONVERT_FROM function with IMPALA_TIMESTAMP argument works properly for this field. I will investigate a little more about does impala and hive can store timestamps into parquet BINARY.

For 1) I think it is better to compare values from the file as opposed to running with the the PARQUET_READER_INT96_AS_TIMESTAMP disabled.
For 2) Can you correct the int96 data in the file? AFAIK, the data should be int96 for the test.

vdiravka:
@bitblender @parthchandra 
Changes according to the comments were made, the branch version was rebased to the master version. 
Could you please review?

parthchandra:
+1. LGTM

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -739,30 +741,76 @@ public void runTestAndValidate(String selection, String validationSelection, Str
...
+  public void testTimestampImpalaConvertFrom() throws Exception {
     compareParquetReadersColumnar("convert_from(field_impala_ts, 'TIMESTAMP_IMPALA')", "cp.`parquet/int96_impala_1.parquet`");
   }
 
   /*
+     Test reading parquet Int96 as TimeStamp and comparing obtained values with the
+     old results (reading the same values as VarBinary and convert_fromTIMESTAMP_IMPALA function using)
+   */
+  @Test
+  public void testImpalaParquetTimestampInt96AsTimeStamp() throws Exception {
```
parthchandra: The test testImpalaParquetTimestampInt96AsTimeStamp fails when run in  a different timezone. Can you mark this as @Ignore unless you can fix the test to run across different timezones?

parthchandra:
Changing this to -1 until unit test failure is addressed.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
vdiravka: This test compares the results between new converter (Int96 to TimeStamp) and the old one (Int96 to VarBinary) with `convert_fromTIMESTAMP_IMPALA` function. 
The issue was in the `ConvertFromImpalaTimestamp` [link to the code
](https://github.com/apache/drill/pull/600/commits/a45490af2dd663168220cc3bda62a2d79170db62#diff-5d8360c5e3cf7d2f6ac7bfe58b6d319aL57) Because the timezone changing shouldn't affect on the result timestamp values.
I deleted timezone consideration there, so now all tests passed successfuly even across different timezones.

parthchandra:
@vdiravka Looks like the test TestHiveStorage.readAllSupportedHiveDataTypesNativeParquet:214 is also failing. (The timestamp_field value is not matching the baseline). Can you take a look?

vdiravka:
@parthchandra The known issue with hive that it stores timestamp values into parquet files with local zone retain. That's why when we want to retrieve the data from such table we should consider the local timezone.
On the other hand parquet files don't involve the particular time zone and when we just read the file we shouldn't consdier a local timezone. And this is also standard drill behaviour with normal int64 timestamps.
So I decided that we need two `IMPALA_TIMESTAMP` functions: for hive and for regular parquet files.
I left  `IMPALA_TIMESTAMP` function without local timezone retain and I added `IMPALA_TIMESTAMP_LOCALTIMEZONE` function (implicit using with hive timestamps and enabled drill native parquet reader). 

Please let me know if this approach is good.
Changes in a new commit for easy review.

parthchandra:
+1. Yes this is fine, and the tests all pass as well.

PR #601, opened 2016-10-03
Improve JDBC Metadata support

laurentgo:
Improve Drill JDBC metadata support by upgrading Avatica version, adding types filter to getTable Metadata API, and letting the JDBC driver using new Metadata API instead of SQL queries.


laurentgo:
Closing: same as #613

PR #603, opened 2016-10-05
DRILL-4927: Add support for Null Equality Joins

KulykRoman:
- Added changes to support Null Equality Joins;
- Created tests for it.


KulykRoman:
@amansinha100 The commit message is updated.

PR #604, opened 2016-10-06
DRILL-4862: binary_string should use another buffer as out buffer to …

chunhui-shi:
…be used in more generic usage.


common/src/main/java/org/apache/drill/common/util/DrillStringUtils.java
```diff
@@ -160,10 +160,9 @@ private static void appendByte(StringBuilder result, byte b) {
    *
```
gparai: Please update the comment - it no longer does in-place parsing.

common/src/main/java/org/apache/drill/common/util/DrillStringUtils.java
```diff
@@ -160,10 +160,9 @@ private static void appendByte(StringBuilder result, byte b) {
    *
    * @return Index in the byte buffer just after the last written byte.
    */
-  public static int parseBinaryString(ByteBuf str, int strStart, int strEnd) {
-    int length = (strEnd - strStart);
-    int dstEnd = strStart;
-    for (int i = strStart; i < strStart+length ; i++) {
+  public static int parseBinaryString(ByteBuf str, int strStart, int strEnd, ByteBuf out) {
+    int dstEnd = 0;
+    for (int i = strStart; i < strEnd ; i++) {
```
gparai: Please change to `strEnd;`

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java
```diff
@@ -1540,15 +1540,16 @@ public void eval() {
...
 
     @Override
     public void eval() {
-      out.buffer = in.buffer;
-      out.start = in.start;
-      out.end = org.apache.drill.common.util.DrillStringUtils.parseBinaryString(in.buffer, in.start, in.end);
+      out.buffer = buffer.reallocIfNeeded(in.end - in.start);
+      out.start = out.end = 0;
+      out.end = org.apache.drill.common.util.DrillStringUtils.parseBinaryString(in.buffer, in.start, in.end, out.buffer);
       out.buffer.setIndex(out.start, out.end);
```
gparai: Do we still need to set the index?

common/src/main/java/org/apache/drill/common/util/DrillStringUtils.java
chunhui-shi: Done

common/src/main/java/org/apache/drill/common/util/DrillStringUtils.java
chunhui-shi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java
chunhui-shi: Yes, we need to set readerIndex and writerIndex. Otherwise if there is another function consume the output of this function, it will hit error.

gparai:
+1

sudheeshkatkam:
+1

PR #605, opened 2016-10-07
DRILL-4870 drill-config.sh sets JAVA_HOME incorrectly for the Mac

paul-rogers:
Recent script revisions uncommented a line to export the Drill-computed JAVA_HOME. Turns out, we should not do this; we should allow Java to figure it out as the location differs depending on platform. This change comments out that export line to restore the pre-1.8 functionality.

### DIFF
--- a/distribution/src/resources/drill-config.sh
+++ b/distribution/src/resources/drill-config.sh
@@ -392,5 +392,7 @@ export DRILL_HOME
 export DRILL_CONF_DIR
 export DRILL_LOG_DIR
 export CP
-export JAVA_HOME
+# DRILL-4870: Don't export JAVA_HOME. Java can find it just fine from the java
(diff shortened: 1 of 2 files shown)


distribution/src/resources/drill-config.sh
```diff
@@ -392,5 +392,7 @@ export DRILL_HOME
 export DRILL_CONF_DIR
 export DRILL_LOG_DIR
 export CP
-export JAVA_HOME
+# DRILL-4870: Don't export JAVA_HOME. Java can find it just fine from the java
+# command. If we attempt to work out out, we do so incorrectly for the Mac.
```
paul-rogers: Yikes! Fixed the typo.

Changed the reference to JAVA_HOME to a reference relative to java, which we must already know.

parthchandra:
+1

PR #607, opened 2016-10-11
Drill 3423:Add New HTTPD format plugin

cgivre:
This PR adds the capability to Drill to query HTTPD web server logs natively.  In addition to the HTTPD parser, it also includes two UDFs: parse_url() and parse_query() which return maps of the URL and the query string respectively.


chunhui-shi:
Some generic questions: 
Do we expect to develop one format plugin for one kind of log?
And is this format plugin designed to supporting multiple(e.g. hundreds) formats for one kind of log?
Is there any correlation functionality that allow users to define a way to correlate multiple lines and extract information from them to build a 'record'?

cgivre:
@chunhui-shi 
I've actually been thinking about writing a generic log parser for Drill in which the user would provide a regex with groups and a list of fields.  For instance consider the following sshd log file:

```
070823 21:00:32       1 Connect     root@localhost on test1
070823 21:00:48       1 Query       show tables
070823 21:00:56       1 Query       select * from category
070917 16:29:01      21 Query       select * from location
070917 16:29:12      21 Query       select * from location where id = 1 LIMIT 1
```
You can't really split this by space or tab, and dissecting it with various string slicing functions would lead to some very complex and ugly queries.   But with the following regex:
```
^(\d{6}\s\d{2}:\d{2}:\d{2})\s+(\d+)\s(\w+)\s+(.+)$
```
You can extract all the fields and query them.  
With respect to the HTTPD log parser, the log parser accepts a format string in the configuration (https://issues.apache.org/jira/browse/DRILL-3423) and with that you can parse any kind of HTTPD log.

parthchandra:
+1. Let's get this into the 1.9 release.

PR #608, opened 2016-10-11
fix the wrong content

lvxin1986:
modify the content "Tells HiveServer2 to execute Hive operations as the user submitting the query. Must be set to false for the Storage  based model" to "Tells HiveServer2 to execute Hive operations as the user submitting the query. Must be set to false for the SQL standard based model" in the section SQL Standard Based Authorization


ilooner:
Change is still valid, file has just moved to https://github.com/apache/drill/blob/gh-pages/_docs/configure-drill/securing-drill/060-configuring-user-impersonation-with-hive-authorization.md

ilooner:
This change will be added by Bridget. So I will close this PR.

PR #609, opened 2016-10-11
DRILL-4882 - Support for MongoDB Direct Connection to a node in a Replica Set.

tgrall:
See issue DRILL-4882

In some case, for analytics the applicatio, Drill, has to connect to node directly without access the replica set or shard globally.

The new configuration option `direct-connection` true|false allows drillbit to be directly connected to a node.

References:
* [Connection String](http://mongodb.github.io/mongo-java-driver/3.0/driver/reference/connecting/connection-settings/)
* [Direct Connection term](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#terms)

### DIFF
--- a/contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoStoragePlugin.java
+++ b/contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoStoragePlugin.java
@@ -59,6 +59,7 @@ public class MongoStoragePlugin extends AbstractStoragePlugin {
   private final MongoSchemaFactory schemaFactory;
   private final Cache<MongoCnxnKey, MongoClient> addressClientMap;
   private final MongoClientURI clientURI;
+  private final boolean directConnection;
 
   public MongoStoragePlugin(MongoStoragePluginConfig mongoConfig,
       DrillbitContext context, String name) throws IOException,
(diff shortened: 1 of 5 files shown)


contrib/storage-mongo/src/main/resources/bootstrap-storage-plugins.json
```diff
@@ -3,7 +3,8 @@
     mongo : {
       type:"mongo",
       enabled: false,
-      connection:"mongodb://localhost:27017/"
+      connection:"mongodb://localhost:27017/",
+      "direct-connection" : false
     }
```
chunhui-shi: Could you make "direct-connection" optional but default to false?

contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoStoragePlugin.java
```diff
@@ -136,17 +138,40 @@ public synchronized MongoClient getClient(List<ServerAddress> addresses) {
     if (client == null) {
       if (credential != null) {
         List<MongoCredential> credentialList = Arrays.asList(credential);
-        client = new MongoClient(addresses, credentialList, clientURI.getOptions());
+        if (!isDirectConnection()) {
```
chunhui-shi: Could you add Unit test for this fix?

PR #611, opened 2016-10-12
Drill-4800: Improve parquet reader performance

parthchandra:
Added a Buffering input stream
Updated parquet reader to optionally use the buffering input stream
Added optional asynchronous reading of page data
Added optional parallel decompression and decoding of columns
    Decompression of data using Gzip/Snappy bypasses the Parquet APIs and calls the decompressors directly (there were concurrency issues with using the Parquet APIs)
Added new operator metrics for asynchronous page reading.


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -0,0 +1,409 @@
...
+      threadPool = parentColumnReader.parentReader.getOperatorContext().getScanExecutor();
+    }
+    asyncPageRead = threadPool.submit(new AsyncPageReaderTask());
+  }
+
+  @Override protected void loadDictionaryIfExists(final ColumnReader<?> parentStatus,
+      final ColumnChunkMetaData columnChunkMetaData, final DirectBufInputStream f) throws UserException {
+    if (columnChunkMetaData.getDictionaryPageOffset() > 0) {
+      try {
+        dataReader.skip(columnChunkMetaData.getDictionaryPageOffset() - dataReader.getPos());
```
ppadma: I see that input argument f is not used. should we be using f instead of dataReader here ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ColumnReader.java
```diff
@@ -234,4 +286,48 @@ public static int readIntLittleEndian(DrillBuf in, int offset) {
...
+  private class ColumnReaderProcessPagesTask implements Callable<Long> {
+
+    private final ColumnReader parent = ColumnReader.this;
+    private final long recordsToReadInThisPass;
+
+    public ColumnReaderProcessPagesTask(long recordsToReadInThisPass){
+      this.recordsToReadInThisPass = recordsToReadInThisPass;
+    }
+
+    @Override public Long call() throws IOException{
```
ppadma: Space between IOException and {

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/BufferedDirectBufInputStream.java
```diff
@@ -0,0 +1,467 @@
...
+    super.init();
+    this.internalBuffer = this.allocator.buffer(this.bufSize);
+    this.tempBuffer = this.allocator.buffer(defaultTempBufferSize);
+    int bytesRead = getNextBlock();
+    if (bytesRead <= 0) {
+      throw new IOException("End of stream reached while initializing buffered reader.");
+    }
+  }
+
+  private DrillBuf reallocBuffer(int newSize ){
```
ppadma: realloc seem to imply we copy the old data which is not the case here. Should we name this function better ?

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/BufferedDirectBufInputStream.java
```diff
@@ -0,0 +1,467 @@
...
+  /**
+   * Creates a <code>BufferedDirectBufInputStream</code>
+   * with the specified buffer size.
+   */
+  public BufferedDirectBufInputStream(InputStream in, BufferAllocator allocator, String id,
+      long startOffset, long totalByteSize, int bufSize, boolean enableHints) {
+    super(in, allocator, id, startOffset, totalByteSize, enableHints);
+    Preconditions.checkArgument(bufSize >= 0);
+    // We make the buffer size the smaller of the buffer Size parameter or the total Byte Size
+    // rounded to next highest pwoer of two
```
ppadma: typo: power

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/BufferedDirectBufInputStream.java
```diff
@@ -0,0 +1,467 @@
...
+    DrillBuf buffer = getBuf();
+    buffer.clear();
+    this.count = this.curPosInBuffer = 0;
+
+    // We *cannot* rely on the totalByteSize being correct because
+    // metadata for Parquet files is incorrect (sometimes). So we read
+    // beyond the totalByteSize parameter. However, to prevent ourselves from reading too
+    // much data, we reduce the size of the buffer, down to 64KiB.
+    if (buffer.capacity() >= (totalByteSize + startOffset - curPosInStream)) {
+      if (buffer.capacity() > smallBufferSize) {
```
ppadma: combine these 2 ifs ?

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/DirectBufInputStream.java
```diff
@@ -0,0 +1,183 @@
...
+    buf.writerIndex(len);
+    return len;
+  }
+
+  public synchronized DrillBuf getNext(int bytes) throws IOException {
+    DrillBuf b = allocator.buffer(bytes);
+    int bytesRead = -1;
+    try {
+    bytesRead = read(b, 0, bytes);
+    } catch (IOException e){
```
ppadma: space )  {

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/DirectBufInputStream.java
```diff
@@ -0,0 +1,183 @@
...
+      b.release();
+      throw e;
+    }
+    if (bytesRead <= -1) {
```
ppadma: bytesRead = 0 ?

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/DirectBufInputStream.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  protected void checkInputStreamState() throws IOException {
+    if (in == null) {
+      throw new IOException("Input stream is closed.");
+    }
+  }
+
+  public synchronized void close() throws IOException {
+    InputStream inp;
```
ppadma: inStream might be a better choice than inp.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -0,0 +1,409 @@
...
+        long e = parent.dataReader.getPos();
+        if(logger.isTraceEnabled()) {
+          logger.trace("[{}]: Read Page Header : ReadPos = {} : Bytes Read = {} ", name, s, e - s);
+        }
+        int compressedSize = pageHeader.getCompressed_page_size();
+        s = parent.dataReader.getPos();
+        pageData = parent.dataReader.getNext(compressedSize);
+        e = parent.dataReader.getPos();
+        bytesRead = compressedSize;
+
```
parthchandra: Nope. e-s is  the size of the header

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -0,0 +1,409 @@
...
+    // I think we are clobbering parts of the dictionary if there can be multiple pages of dictionary
+
+    do {
+      if (pageHeader.getType() == PageType.DICTIONARY_PAGE) {
+        readDictionaryPageData(readStatus, parentColumnReader);
+        // Ugly. Use the Async task to make a synchronous read call.
+        readStatus = new AsyncPageReaderTask().call();
+        pageHeader = readStatus.getPageHeader();
+      }
+    } while (pageHeader.getType() == PageType.DICTIONARY_PAGE);
```
parthchandra: Because this method has to block until it has the dictionary page(s) read. This part of the code is simply retaining the old code that seems to imply that more than one dictionary page may exist, though it is very unlikely to be so.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -0,0 +1,409 @@
...
+  private DrillBuf getDecompressedPageData(ReadStatus readStatus) {
+    DrillBuf data;
+    boolean isDictionary = false;
+    synchronized (this) {
+      data = readStatus.getPageData();
+      readStatus.setPageData(null);
+      isDictionary = readStatus.isDictionaryPage;
+    }
+    if (parentColumnReader.columnChunkMetaData.getCodec() != CompressionCodecName.UNCOMPRESSED) {
+      DrillBuf uncompressedData = data;
```
parthchandra: yes :)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
parthchandra: No we should be using dataReader here. I override the method of parent class which implements the original (synchronous) PageReader, so am forced to match the signature.

parthchandra: The assumption is true but added the assert anyway.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/OperatorContext.java
```diff
@@ -43,10 +45,18 @@
 
   public abstract OperatorStats getStats();
 
+  public abstract ExecutorService getExecutor();
+
+  public abstract ExecutorService getScanExecutor();
```
parthchandra: I kept it like this to minimize touching parts of the code that have nothing to do with the Parquet reader.

jinfengni:
+1

Look good to me.

Very good documentation of design and configuration parameters.

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/DirectBufInputStream.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+    try {
+      Class<? extends FSDataInputStream> inputStreamClass = inputStream.getClass();
+      readAhead =
+          inputStreamClass.getMethod("adviseFile", new Class[] {adviceType, long.class, long.class});
+    } catch (NoSuchMethodException e) {
+      logger.info("Unable to call fadvise due to: {}", e.toString());
+      readAhead = null;
+      return;
+    }
```
paul-rogers: The above is relatively expensive. Should the work be done only once, then the method handle cached? If we are making many fadvise calls, then caching the method handle might provide a nice performance bump.

exec/java-exec/src/main/java/org/apache/drill/exec/util/filereader/DirectBufInputStream.java
```diff
@@ -0,0 +1,166 @@
...
+    return;
+  }
+
+  public int read() throws IOException {
+    return getInputStream().read();
+  }
+
+  public synchronized int read(DrillBuf buf, int off, int len) throws IOException {
+    buf.clear();
+    ByteBuffer directBuffer = buf.nioBuffer(0, len);
```
paul-rogers: Seems to create garbage on every read (to create the ByteBuffer with the proper offset and length.) Any way to do the read without garbage?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBinaryReader.java
```diff
@@ -43,43 +51,153 @@ public VarLenBinaryReader(ParquetRecordReader parentReader, List<VarLengthColumn
...
+    ArrayList<Future<Boolean>> futures = Lists.newArrayList();
+    for (VarLengthColumn<?> columnReader : columns) {
+      Future<Boolean> f = columnReader.readPageAsync();
+      futures.add(f);
+    }
+    for (Future<Boolean> f : futures) {
+      try {
+        isDone = isDone || f.get().booleanValue();
+      } catch (Exception e) {
+        f.cancel(true);
```
paul-rogers: Same question about future failure & handling as above.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBinaryReader.java
```diff
@@ -43,43 +51,153 @@ public VarLenBinaryReader(ParquetRecordReader parentReader, List<VarLengthColumn
...
+    ArrayList<Future<Integer>> futures = Lists.newArrayList();
+    for (VarLengthColumn<?> columnReader : columns) {
+      Future<Integer> f = columnReader.readRecordsAsync(columnReader.pageReader.valuesReadyToRead);
+      futures.add(f);
+    }
+    for (Future f : futures) {
+      try {
+        f.get();
+      } catch (Exception e) {
+        f.cancel(true);
```
paul-rogers: Ditto about future failure handling.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -859,5 +865,40 @@ public void testLastPageOneNull() throws Exception {
         "cp.`parquet/last_page_one_null.parquet`");
   }
 
+  @Test
```
paul-rogers: Should there be tests for other code paths introduced in this commit?

* Failures in the parallel tasks (see comments above about handling such failures.)
* Verifying that the various "determine size parallel" code paths are exercised?
* Verify that pages were, indeed, read in parallel?
* What happens if more parallel readers are needed than threads available in the pool?
* What happens with parallelization if the file has 1000 columns (which, presumably, could cause 1K parallel reads?) Some upper limit on parallelization?
* Verify the five decompression paths?

Might the above code need a bit of instrumentation so the test can validate that the parallel reader was, in fact, used and exercised the interesting code paths?

PR #613, opened 2016-10-12
DRILL-4730: Update JDBC DatabaseMetaData implementation to use new Metadata APIs

laurentgo:
Update JDBC driver to use Metadata APIs instead of executing SQL queries

### DIFF
--- a/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillCursor.java
+++ b/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillCursor.java
@@ -43,6 +43,7 @@ import org.apache.drill.exec.exception.SchemaChangeException;
 import org.apache.drill.exec.proto.UserBitShared.QueryId;
 import org.apache.drill.exec.proto.UserBitShared.QueryResult.QueryState;
 import org.apache.drill.exec.proto.UserBitShared.QueryType;
+import org.apache.drill.exec.proto.UserProtos.PreparedStatement;
 import org.apache.drill.exec.proto.helper.QueryIdHelper;
 import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.RecordBatchLoader;
@@ -527,10 +528,18 @@ class DrillCursor implements Cursor {
         : "currentBatchHolder.getRecordCount() not 0 (is "
           + currentBatchHolder.getRecordCount() + " in loadInitialSchema()";
 
+    final PreparedStatement preparedStatement;
     if (statement instanceof DrillPreparedStatementImpl) {
       DrillPreparedStatementImpl drillPreparedStatement = (DrillPreparedStatementImpl) statement;
-      connection.getClient().executePreparedStatement(drillPreparedStatement.getPreparedStatementHandle().getServerHandle(), resultsListener);
+      preparedStatement = drillPreparedStatement.getPreparedStatementHandle();
     } else {
+      preparedStatement = null;
+    }
+
+    if (preparedStatement != null) {
+      connection.getClient().executePreparedStatement(preparedStatement.getServerHandle(), resultsListener);
+    }
+    else {
       connection.getClient().runQuery(QueryType.SQL, signature.sql, resultsListener);
     }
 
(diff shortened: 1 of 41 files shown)


sudheeshkatkam:
I started looking into this. Can you please split the commit into two (refactoring changes, and the core changes that use new API) for quicker review?

laurentgo:
the only "refactoring" I can think of is between DrillCursor and DrillResultSetImpl, anything else is related to the feature. I'll try to split it, but hopefully in the current state, it should still be pretty readable.

sudheeshkatkam:
@laurentgo thanks for splitting the commit!

Looks like queries are replaced by API calls in the core changes, so can the new driver talk to old servers?

sudheeshkatkam:
I just realized you mentioned about this on the mailing list. In that case, let's defer committing this pull request to the next release.

> That said, the metadata change is not merged yet (and I will add the extra
> logic), which is not the case regarding prepared statement support.

laurentgo:
I added the necessary fallbacks for prepared statement and metadata, so a new client should now be able to talk with an old server without adverse effects.

I would prefer that this change to be in this release, to match the state of the C++ client, but I realized this is getting late.

laurentgo:
@sudheeshkatkam can you please give it another review?

exec/jdbc/src/test/java/org/apache/drill/jdbc/DatabaseMetaDataGetColumnsTest.java
```diff
@@ -2712,22 +2712,18 @@ public void test_SOURCE_DATA_TYPE_hasSameNameAndLabel() throws SQLException {
 
   @Test
   public void test_SOURCE_DATA_TYPE_hasRightTypeString() throws SQLException {
-    // TODO(DRILL-2135):  Resolve workaround:
-    //assertThat( rsMetadata.getColumnTypeName( 22 ), equalTo( "SMALLINT" ) );
-    assertThat( rowsMetadata.getColumnTypeName( 22 ), equalTo( "INTEGER" ) );
+    assertThat( rowsMetadata.getColumnTypeName( 22 ), equalTo( "SMALLINT" ) );
```
laurentgo: no, it isn't unfortunately

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillCursor.java
```diff
@@ -24,33 +24,260 @@
...
+        throttle = null;
+      }
+      return stopped;
+    }
+
+    public void awaitFirstMessage() throws InterruptedException {
+      firstMessageReceived.await();
+    }
+
+    private void releaseIfFirst() {
```
laurentgo: unfortunately, github doesn't show the complete change, but if the question is about DrillCursor refactoring, the answer is yes! Since metadata queries resultset are plain avatica resultsets (and not using drillcursor at all), having some of the drillcursor business logic in DrillResultSetImpl caused some issues during refactoring.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillMetaImpl.java
```diff
@@ -78,19 +111,237 @@ private MetaResultSet s(String s) {
...
+
+
+  protected static ColumnMetaData drillColumnMetaData(String name, int index,
+      Class<?> type, boolean notNullable) {
+    TypeInfo pair = TypeInfo.get(type);
+    ColumnMetaData.Rep rep =
+        ColumnMetaData.Rep.VALUE_MAP.get(type);
+    ColumnMetaData.AvaticaType scalarType =
+        ColumnMetaData.scalar(pair.sqlType, pair.sqlTypeName, rep);
+    return new ColumnMetaData(
```
laurentgo: yes, it is fine. This method is only used for ResultSetMetaData, it just creates column metadata based on a pojo field.and none of the types used here have precision nor scale. 

Metadata for queries are handled in a different way and are already derived from the BatchSchema.

sudheeshkatkam:
+1

laurentgo:
yes, pull requests have been kept in sync.

PR #616, opened 2016-10-14
fix the default drill directory in zookeeper

lvxin1986:
Fix the bug:correct the drill directory in zookeeper to dirll from Drill,because the value is case Sensitive.


paul-rogers:
Oh, can you file a JIRA for this, then change the title to include the JIRA number:

JIRA-1234: fix the default drill directory in zookeeper

We use JIRA entries to track workflow. Thanks.

PR #617, opened 2016-10-14
DRILL-4934 ServiceEngine does not use property useIP for DrillbitStartup

joeswingle:
Pull Request for DRILL-4934


ilooner:
@joeswingle It looks like useIp was removed from ServiceEngine in this PR https://github.com/apache/drill/pull/578 . @parthchandra @sohami since you guys looked at the PR, do you know why it was removed and if it would make sense to add back?

jnturton:
@joeswingle the removed code that obtained an IP address fr the Drillbit when this flag was set was
```
useIP ? InetAddress.getLocalHost().getHostAddress() : InetAddress.getLocalHost().getCanonicalHostName();
```
That, from what I can tell, is generally going to return the IP address of the local loopback interface which would not have any effect on embedded Drill but would be useless for distributed Drill which is the only place it would be of interest. Now that we have `DRILL_HOST_NAME` I think we can get by without a feature like this but let me know if you'd like me to reopen here.

PR #619, opened 2016-10-17
DRILL-4946: redirect System.err so users under embedded mode won't se…

chunhui-shi:
…e error stacks printed due to errors occurred in scalar replacement.

### DIFF
--- a/distribution/src/resources/sqlline
+++ b/distribution/src/resources/sqlline
@@ -59,7 +59,7 @@ done
 # This is not currently needed as the new SQLLine we are using doesn't isolate.
 # DRILL_SHELL_JAVA_OPTS="-Dsqlline.isolation=TRANSACTION_NONE $DRILL_SHELL_JAVA_OPTS"
 
-DRILL_SHELL_LOG_OPTS="-Dlog.path=$DRILL_LOG_DIR/sqlline.log -Dlog.query.path=$DRILL_LOG_DIR/sqlline_queries.json"
+DRILL_SHELL_LOG_OPTS="-Dlog.path=$DRILL_LOG_DIR/sqlline.log -Dlog.query.path=$DRILL_LOG_DIR/sqlline_queries.json -Dorg.apache.drill.exec.server.Drillbit.system_options=\"org.apache.drill.exec.compile.ClassTransformer.scalar_replacement=off\""
 
 # Use either the SQLline options (for remote Drill) or full Drill options
 # (embedded Drill)


exec/java-exec/src/main/java/org/apache/drill/exec/util/StdErrCapturer.java
```diff
@@ -0,0 +1,81 @@
...
+
+
+import java.io.ByteArrayOutputStream;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * The safe way of this class is to redirect System.err to nullOutputStream
+ * If
```
chunhui-shi: The comment is now finished.

exec/java-exec/src/main/java/org/apache/drill/exec/util/StdErrCapturer.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.PrintStream;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * The safe way to use this class is to redirect System.err to nullOutputStream.
+ * If callers want to get the redirected output -ito print or log elsewhere - as used in unit tests and they
+ * need the output to be neat, then there should be advanced ways to preserve the output to avoid concurrent 
+ * inputs messing up. But this class is not designed for that.
+ */
+public class StdErrCapturer {
```
paul-rogers: If an exception occurs, this class can leave the System.stderr redirected, causing mysterious errors. Suggestion: make this AutoClosable: 

try ( StdErrCapturer cap = new StdErrCapturer( ) ) {
  // Do your stuff here
}

The above will ensure that stderr is restored whether the enclosed code succeeds or fails.

exec/java-exec/src/main/java/org/apache/drill/exec/util/StdErrCapturer.java
```diff
@@ -0,0 +1,83 @@
...
+  static private PrintStream savedStdErr;
+  static private OutputStream buffer;
+  static private PrintStream capturingStream;
+  static Boolean redirected = false;
+
+  public static void redirect() {
```
paul-rogers: Do in constructor (so it can be reversed in close.)

ilooner:
@chunhui-shi @paul-rogers Since Drill has moved to Java 8, which is supposed to do scalar replacement for us, would it be appropriate to commit this change to turn off scalar replacement? Or is this change no longer applicable?

chunhui-shi:
I suggest to set scalar replacement flag to off by default. I think we have compared the performance between on/off before.
------------------------------------------------------------------
Sender:Timothy Farkas <notifications@github.com>
Sent at:2018 Jun 21 (Thu) 10:45
To:apache/drill <drill@noreply.github.com>
Cc:Chunhui Shi <shi.chunhui@aliyun.com>; Mention <mention@noreply.github.com>
Subject:Re: [apache/drill] DRILL-4946: redirect System.err so users under embedded mode won't se… (#619)

@chunhui-shi @paul-rogers Since Drill has moved to Java 8, which is supposed to do scalar replacement for us, would it be appropriate to commit this change to turn off scalar replacement? Or is this change no longer applicable?

paul-rogers:
@ilooner, yes, the change here seems fine. We're just passing an option to Sqlline (which passes it to Drill). This is a cleaner way to avoid the original issue. So,
+1

vvysotskyi:
I don't think it is the correct decision to disable scalar replacement for embedded mode only just because of printing stack trace to stderr. 

Also, I think making more differences in work between embedded and distributed mode is not a good idea. 

If we want to disable scalar replacement, we should do that independently of running mode or used client.

ilooner:
Fair point @vvysotskyi . @chunhui-shi Would you have time to disable scalar replacement for all modes?

vvysotskyi:
Closing this PR, since most of the problems which caused errors during scalar replacement were resolved.

PR #620, opened 2016-10-18
add the RDBMS to the schema types

lvxin1986:
Add the RDBMS to the schema types which driver supports.


lvxin1986:
I think RDBMS should be the schema type which driver supports

PR #621, opened 2016-10-18
DRILL-4950: Remove incorrect false condition; consume all empty batches

sudheeshkatkam:
@parthchandra please review


parthchandra:
+1. Thanks for fixing this!

PR #627, opened 2016-10-21
DRILL-4812: Fix for detecting wildcards in paths when running on windows

mlavende:
By the time the path makes it to the `handleWildCard` method the separators have all been normalized to the forward slash.  This method, like `removeLeadingSlash` are safe, and must only check directly for the `/` separator.


sudheeshkatkam:
+1

PR #628, opened 2016-10-26
DRILL-4964: Drill fails to connect to hive metastore after hive metas…

sohami:
…tore is restarted unless drillbits are restarted also

            Changes: For certain methods in DrillHiveMetaStoreClient like getAllDatabases and getAllTables the HiveMetaStoreClient
                     wraps the exception of type MetaException and TException both into MetaException. So in case of connection
                     failure which is thrown as TException it is difficult to categorize at DrillClient level. Hence the fix
                     is to close older connection and reconnect in case of these 2 api's. In all other cases proper set of exceptions
                     are thrown where we can handle each one individually.


sohami:
@jinfengni - Can you please help to review ?

jinfengni:
+1 

LGTM.

sohami:
@jinfengni - Thanks for review.

PR #631, opened 2016-10-26
DRILL-4968: Add column size to ColumnMetadata

laurentgo:
Add a column size to ColumnMetadata so that JDBC and ODBC clients share
the implementation and don't have to recompute it client side.


adeneche:
+1, LGTM

PR #632, opened 2016-10-27
DRILL-4969: Basic implementation of display size for column metadata

laurentgo:
Add a simple implementation of display size, based on the ODBC table
available at: https://msdn.microsoft.com/en-us/library/ms713974(v=vs.85).aspx

### DIFF
--- a/common/src/main/java/org/apache/drill/common/types/Types.java
+++ b/common/src/main/java/org/apache/drill/common/types/Types.java
@@ -275,6 +275,84 @@ public class Types {
     return isSigned;
   }
 
+  public static int getJdbcDisplaySize(MajorType type) {
+    if (type.getMode() == DataMode.REPEATED || type.getMinorType() == MinorType.LIST) {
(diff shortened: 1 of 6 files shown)


common/src/main/java/org/apache/drill/common/types/Types.java
```diff
@@ -275,6 +275,84 @@ public static boolean isJdbcSignedType( final MajorType type ) {
     return isSigned;
   }
 
+  public static int getJdbcDisplaySize(MajorType type) {
+    if (type.getMode() == DataMode.REPEATED || type.getMinorType() == MinorType.LIST) {
+      return 0;
```
sudheeshkatkam: Why does 0 imply?

laurentgo: 0 implies that we don't know the display size

sudheeshkatkam:
+1

PR #633, opened 2016-10-27
DRILL-4972: Remove setDaemon(true) call in WorkManager.StatusThread

sudheeshkatkam:


exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -295,7 +295,7 @@ public FragmentExecutor getFragmentRunner(final FragmentHandle handle) {
    */
   private class StatusThread extends Thread {
     public StatusThread() {
-      setDaemon(true);
+      setDaemon(false);
```
sohami: Since the StatusThread is started by "main" thread by default the StatusThread will be non-Daemon, so we can just remove the setDaemon call.

sohami: Not sure why removing it will not ensure the same. I think it's redundant and like overriding a default already set in constructor with default value using setter method.

parthchandra:
+1. LGTM

PR #634, opened 2016-10-27
DRILL-4974: NPE in FindPartitionConditions.analyzeCall() for 'holistic' expressions

bitblender:
Changes: Added a missing null check in FindPartitionConditions.analyzeCall(), to ensure that opStack.peek() value is dereferenced only after a null-check. Without this check, if the expression is holistic, opStack can be null, so using the value of opStack.peek() without a check can cause an NPE.


bitblender:
@amansinha100 Can you please review this change.

PR #635, opened 2016-10-28
DRILL-4927 (part 2): Add support for Null Equality Joins (mixed compa…

KulykRoman:
…rators)

This changes are a subset of the original pull request from DRILL-4539 (PR-462).
- Added changes to support mixed comparators;
- Added tests for it.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/TestJoinNullable.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/TestJoinNullable.java
@@ -20,7 +20,6 @@ package org.apache.drill;
 import static org.junit.Assert.assertEquals;
 
 import org.apache.drill.common.util.TestTools;
-import org.junit.Ignore;
 import org.junit.Test;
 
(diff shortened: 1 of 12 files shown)


exec/java-exec/src/test/java/org/apache/drill/TestJoinNullable.java
```diff
@@ -493,4 +496,81 @@ public void withNullEqualAdditionFilter() throws Exception {
         .go();
   }
 
+  @Test
+  public void withMixedEqualAndIsNotDistinctHashJoin() throws Exception {
```
amansinha100: Could you add a test with the predicate of type: 
    t1.key = t2.key AND ((t1.data=t2.data) OR (t1.data IS NULL AND t2.data IS NULL))
where the right side of the AND gets converted to IS NOT DISTINCT FROM.

KulykRoman:
Added tests with t1.key = t2.key AND ((t1.data=t2.data) OR (t1.data IS NULL AND t2.data IS NULL)) pattern.

PR #636, opened 2016-10-30
DRILL-4951: Do Guava.patch earlier so we can run single Hbase's unit …

chunhui-shi:
…test through command line or IDE


amansinha100:
+1

PR #637, opened 2016-10-31
Drill 1950 : Parquet row group filter pushdown. 

jinfengni:


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -1000,6 +1053,81 @@ public long getColumnValueCount(SchemaPath column) {
...
+      final ImplicitColumnExplorer columnExplorer = new ImplicitColumnExplorer(optionManager, this.columns);
+      Map<String, String> implicitColValues = columnExplorer.populateImplicitColumns(file.getPath(), selectionRoot);
+
+      for (RowGroupMetadata rowGroup : file.getRowGroups()) {
+        ParquetMetaStatCollector statCollector = new ParquetMetaStatCollector(
+            parquetTableMetadata,
+            rowGroup.getColumns(),
+            implicitColValues);
+
+        Map<SchemaPath, ColumnStatistics> columnStatisticsMap = statCollector.collectColStat(schemaPathsInExpr);
```
parthchandra: Shouldn't we be able to build the filter predicate once outside the for loop? Or is it needed because the implicit columns are needed here?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/RangeExprEvaluator.java
```diff
@@ -0,0 +1,282 @@
...
+        return getStatistics( ((Float8Holder)minFuncHolder).value, ((Float8Holder)maxFuncHolder).value);
+      default:
+        return null;
+      }
+    } catch (Exception e) {
+      throw new DrillRuntimeException("Error in evaluating function of " + holderExpr.getName() );
+    }
+  }
+
+  static Map<TypeProtos.MinorType, Set<TypeProtos.MinorType>> CAST_FUNC = new HashMap<>();
```
amansinha100: What's the need for a new mapping for the casts  instead of leveraging TypeCastRules ?  I see that you are restricting the casts to only 4 data types .. is that what Parquet filter pushdown supports ?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/ParquetPredicates.java
```diff
@@ -0,0 +1,334 @@
...
+  /**
+   * EQ (=) predicate
+   */
+  public static class EqualPredicate extends ParquetCompPredicate {
+    public EqualPredicate(LogicalExpression left, LogicalExpression right) {
+      super(left, right);
+    }
+
+    @Override
+    public boolean canDrop(RangeExprEvaluator evaluator) {
```
amansinha100: Since the canDrop() and associated predicate analysis has overlaps with what the Parquet library itself does, it would be helpful to add comments somewhere that indicates why we need to customize it for Drill (e.g handle CAST expressions etc.).

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -77,7 +78,7 @@
 public class Metadata {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Metadata.class);
 
-  public static final String[] OLD_METADATA_FILENAMES = {".drill.parquet_metadata.v2"};
+  public static final String[] OLD_METADATA_FILENAMES = {".drill.parquet_metadata.v3"};
```
jinfengni: I reversed this change.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -329,79 +330,118 @@ private OriginalType getOriginalType(Type type, String[] path, int depth) {
...
+    }
+    Type t = ((GroupType) type).getType(path[depth]);
+    return getColTypeInfo(schema, t, path, depth + 1);
+  }
+
+  private class ColTypeInfo {
+    public OriginalType originalType;
+    public int precision;
+    public int scale;
+    public int repetitionLevel;
```
jinfengni: Included definition level for future use.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
jinfengni: Right. Filter predicate should be build only once. It's inside the loop just we need the column type information during filter expression materialization, for both regular columns and implicit columns. 

I put a check if (filterPredicate == null) inside the loop, so that filter predicate is built only once.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -1000,6 +1053,81 @@ public long getColumnValueCount(SchemaPath column) {
 
   @Override
   public List<SchemaPath> getPartitionColumns() {
-    return new ArrayList<>(columnTypeMap.keySet());
+    return new ArrayList<>(partitionColTypeMap.keySet());
   }
+
+  public GroupScan applyFilter(LogicalExpression filterExpr, UdfUtilities udfUtilities,
+      FunctionImplementationRegistry functionImplementationRegistry, OptionManager optionManager) {
+    if (fileSet.size() == 1 || ! (parquetTableMetadata instanceof Metadata.ParquetTableMetadata_v3)) {
```
jinfengni: I added a isRowGroupPrunable in Metadata class, and removed "instanceof" check.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRGFilterEvaluator.java
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    Set<LogicalExpression> constantBoundaries = ConstantExpressionIdentifier.getConstantExpressionSet(materializedFilter);
+    ParquetFilterPredicate parquetPredicate = (ParquetFilterPredicate) ParquetFilterBuilder.buildParquetFilterPredicate(
+        materializedFilter, constantBoundaries, udfUtilities);
+
+    return canDrop(parquetPredicate, columnStatisticsMap, rowCount);
+  }
+
+//  public static boolean evalFilter2(LogicalExpression expr, ParquetMetadata footer, int rowGroupIndex, OptionManager options, FragmentContext fragmentContext, Map<String, String> implicitColValues) {
```
jinfengni: My negligence. I should have clean up the commented code.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/stat/ParquetFooterStatCollector.java
```diff
@@ -0,0 +1,176 @@
...
+import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.joda.time.DateTimeConstants;
+import org.joda.time.DateTimeUtils;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class ParquetFooterStatCollector implements ColumnStatCollector {
```
jinfengni: Added stopwatch to stat collector.

Also, the pushdown rule has a stopwatch to log the elapsed time for the entire pruning process. That will give us some sense how much overhead filter pruning process introduces.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/stat/ParquetMetaStatCollector.java
```diff
@@ -0,0 +1,146 @@
...
+        final Long numNull = columnMetadata.getNulls();
+
+        primitiveType = this.parquetTableMetadata.getPrimitiveType(columnMetadata.getName());
+        originalType = this.parquetTableMetadata.getOriginalType(columnMetadata.getName());
+        final Integer repetitionLevel = this.parquetTableMetadata.getRepetitionLevel(columnMetadata.getName());
+
+        statMap.put(schemaPath, getStat(min, max, numNull, primitiveType, originalType, repetitionLevel));
+      } else {
+        final String columnName = schemaPath.getRootSegment().getPath();
+        if (implicitColValues.containsKey(columnName)) {
```
jinfengni: Without knowledge of implicit columns, expression materialization will treat dir0 in dir0 = 1995 as NULLEXPRESSION, and could not differentiate from a regular non-exist column.  A condition on a regular non-exist column will always lead to canDrop = true. 

That's the main reason we have to pass in the list of implicit columns when do expression materialization.  

In addition to implicit column name, we also pass in implicit column values, and wrap them in Statistics instance with min and max having same value.  That is done to expand the possibility of pruning. 

For example, if we have regCol = 5 or dir0 = 1995.  If regCol is not a partition column, we would not do any partition pruning in the current partition pruning logical. Pass the implicit column values may allow us to prune some row groups using condition regCol = 5 or dir0 = 1995.

 I added explanations in comment.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
```diff
@@ -346,6 +355,20 @@ static RuleSet getPruneScanRules(OptimizerRulesContext optimizerRulesContext) {
   }
 
   /**
+   *   Get an immutable list of partition pruning rules that will be used in logical planning.
```
jinfengni: Modified the comment.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
```diff
@@ -346,6 +355,20 @@ static RuleSet getPruneScanRules(OptimizerRulesContext optimizerRulesContext) {
...
+   */
+  static RuleSet getPhysicalPruneScanRules(OptimizerRulesContext optimizerRulesContext) {
+    final ImmutableSet<RelOptRule> pruneRules = ImmutableSet.<RelOptRule>builder()
+        .add(
+            ParquetPushDownFilter.getFilterOnProject(optimizerRulesContext),
```
jinfengni: Opened DRILL-4998, and put comments to explain why we put this rule in physical planning phase.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/stat/ParquetMetaStatCollector.java
amansinha100: Ok, I was thinking that you may not even pass in the implicit columns to the expression materializer and let the PruneScanRule rule handle the dirN column filters by pushing down into the scan.. but since your new rule is applied later (during physical planning phase) it implies that those  were not removed.  So now if we want to leverage them for row-group pruning, it makes sense to create the materialized expression with them.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/RangeExprEvaluator.java
jinfengni: Yes, I'm working with @bbevens  to document the types of cast supported for such pruning.

This static HashMap only needs populate once when this class is loaded, since it just keeps track of the type mapping, not the cast function instance. At time of evaluation, each thread's operation is read operation, no write operation. This is similar to what we have in TypeCastRules. Do you think it's fine in terms of thread-safety?

jinfengni:
@amansinha100 , @parthchandra , I pushed three commits, to address the review comments, and add one option.  Please kindly take another look at the commits. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/RangeExprEvaluator.java
amansinha100: you're right..it is only being read by the queries, so should be ok.

amansinha100:
LGTM.  +1

PR #639, opened 2016-10-31
New fragment placement algorithm based on locality of data.

ppadma:


ppadma:
Updated the JIRA with details on how current algorithm works, why remote reads were happening and the new algorithm details.
https://issues.apache.org/jira/browse/DRILL-4706

sudheeshkatkam:
@vkorukanti if you don't mind, can you review this?

exec/java-exec/src/main/java/org/apache/drill/exec/store/schedule/AssignmentCreator.java
```diff
@@ -146,6 +153,32 @@ private AssignmentCreator(List<DrillbitEndpoint> incomingEndpoints, List<T> unit
...
+      }
+
+      FragIteratorWrapper iteratorWrapper = endpointIterators.get(endpoint);
+      if (iteratorWrapper == null) {
+        currentUnassignedList.add(workPair);
+        continue;
+      }
+
+      Integer assignment = iteratorWrapper.iter.next();
+      iteratorWrapper.count++;
```
sohami: Here shouldn't we check if the "iteratorWrapper.count" is exceeding the "iteratorWrapper.maxCount" ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/LocalAffinityFragmentParallelizer.java
```diff
@@ -0,0 +1,149 @@
...
+        // 4. Make sure width is at least the min width enforced by operators
+        width = Math.max(parallelizationInfo.getMinWidth(), width);
+
+        // 5. Make sure width is at most the max width enforced by operators
+        width = Math.min(parallelizationInfo.getMaxWidth(), width);
+
+        // 6: Finally make sure the width is at least one
+        width = Math.max(1, width);
+
+        List<DrillbitEndpoint> endpointPool = Lists.newArrayList();
```
sudheeshkatkam: final (and wherever possible, generously)

exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/LocalAffinityFragmentParallelizer.java
```diff
@@ -0,0 +1,149 @@
...
+            endpointAssignments.put(ep, 1);
+        }
+
+        // Keep allocating from endpoints in a round robin fashion upto
+        // max(targetAllocation, maxwidthPerNode) for each endpoint and
+        // upto width for all together.
+        while(totalAssigned < width) {
+            int assignedThisRound = 0;
+            for (DrillbitEndpoint ep : endpointPool) {
+                if (remainingEndpointAssignments.get(ep) > 0 &&
```
sudheeshkatkam: get value once into local var (and reuse)

exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/LocalAffinityFragmentParallelizer.java
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }
+            if (assignedThisRound == 0) {
+                break;
+            }
+        }
+
+        // This is for the case where drillbits are not running on endPoints which have data.
+        // Allocate them from the active endpoint pool.
+        int totalUnAssigned =
```
sudheeshkatkam: So this parallelizer is not strictly local? Why not fail?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -822,10 +838,103 @@ private void getFiles(String path, List<FileStatus> fileStatuses) throws IOExcep
...
+  /*
+   * Figure out the best node to scan each of the rowGroups and update the preferredEndpoint.
+   * Based on this, update the total work units assigned to the endpoint in the endpointAffinity.
+   */
+  private void computeRowGroupAssignment() {
+    Map<DrillbitEndpoint, Integer> numEndpointAssignments = Maps.newHashMap();
+    Map<DrillbitEndpoint, Long> numAssignedBytes = Maps.newHashMap();
+
+    // Do this for 2 iterations to adjust node assignments after first iteration.
+    int numIterartions = 2;
```
sudheeshkatkam: Iterartions -> iterations

exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/LocalAffinityFragmentParallelizer.java
```diff
@@ -0,0 +1,149 @@
...
+import java.util.Collection;
+import java.util.HashMap;
+
+
+/**
+ * Implementation of {@link FragmentParallelizer} where fragment has zero or more endpoints.
+ * This is for Parquet Scan Fragments only. Fragment placement is done preferring strict
+ * data locality.
+ */
+public class LocalAffinityFragmentParallelizer implements FragmentParallelizer {
```
sudheeshkatkam: When to use this vs HardAffinityFragmentParallelizer?

ppadma:
Updated with all review comments taken care of.  Added TestLocalAffinityFragmentParallelizer.java which has bunch of test cases with examples.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/LocalAffinityFragmentParallelizer.java
ppadma: I got all unit and regression tests pass with localAffinity=true.  If this algorithm fails, that is not possible. Also, we are doing this only for the case when drillbits are not running on the nodes which have data.

ppadma:
Some initial comments.

The issue is regarding assigning fragments based on strict locality. So why is the parallelization logic affected, and not exclusively locality?

Parallelization logic is affected because it decides how many fragments to run on each node and that is dependent on locality.

sudheeshkatkam:
Hmm the answer seems like a rephrasing of the question. Sorry, I misspoke. Better asked:

The issue is regarding assigning **_work to_** fragments based on strict locality (**_decide which fragment does what_**). So why is the parallelization (**_decide how many fragments_**) logic affected?

ppadma:
Parallelization logic is affected for following reasons:
Depending upon how many rowGroups to scan on a node (based on locality information) i.e. how much work the node has to do, we want to adjust the number of fragments on the node (constrained to usual global and per node limits). 
We do not want to schedule fragment(s) on a node which do not have data. 
Because we want pure locality, we may have fewer fragments doing more work.

ppadma:
Merged with latest code. All review comments taken care of. All tests pass with the option `store.parquet.use_local_affinity` = true and false, both.

kkhatua:
@ppadma  was this merged? I don't see a `plus one` and the PR isn't closed.

ppadma:
Even though it is old, this PR is still very much relevant and useful feature to have in Drill for certain use cases/scenarios. 
I request a committer to work with me so we can get it in. Any volunteers ?

PR #644, opened 2016-11-03
DRILL-4980: Upgrading of the approach of parquet date correctness status detection

vdiravka:
- Parquet writer version is added;
- Updated the detection method of parquet date correctness.


vdiravka:
@parthchandra @paul-rogers @jaltekruse coud you please review?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -927,15 +927,11 @@ public void setMax(Object max) {
...
+    @JsonProperty int writerVersion;
 
     public ParquetTableMetadata_v2() {
-      super();
-    }
-
-    public ParquetTableMetadata_v2(boolean isDateCorrect) {
       this.drillVersion = DrillVersionInfo.getVersion();
-      this.isDateCorrect = isDateCorrect;
+      this.writerVersion = ParquetWriter.WRITER_VERSION;
```
paul-rogers: We set the writer version to the current version when we create the metadata. Is this same metadata used for both read and write? If so, we have the potential for a nasty bug. A (new) reader fails to set the writerVersion value from actual file metadata. The value will default to the latest, even if the file itself happens to be older.

I wonder if it makes sense to pass the version into the constructor. The Writer passes in the current writer version. The reader must pass in the value found in the file.

Or, is this metadata used only for writing, but not reading? If that is true perhaps we can document that in the code somewhere. (I looked but did not anything.)

Or, is this metadata cached from scanning actual files? If so, isn't defaulting the writer version simply asking for trouble if someone forgets to set this field based on actual file version?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -927,15 +927,11 @@ public void setMax(Object max) {
     @JsonProperty List<ParquetFileMetadata_v2> files;
     @JsonProperty List<String> directories;
     @JsonProperty String drillVersion;
-    @JsonProperty boolean isDateCorrect;
+    @JsonProperty int writerVersion;
```
paul-rogers: This property is Jackson-serialized. How does Jackson handle older files written without this version? According to this post (http://stackoverflow.com/questions/8320993/jackson-what-happens-if-a-property-is-missing), "Setter methods are only invoked for properties with explicit values." This means that older files without the writerVersion set won't call the function to deserialize the writer version, and the version will default to the newest version. I suspect that either A) I'm misunderstanding what this code does, or B) we have a backward-compatibility issue. Please explain which.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -944,14 +940,16 @@ public ParquetTableMetadata_v2(ParquetTableMetadataBase parquetTable,
...
+      this.writerVersion = ParquetWriter.WRITER_VERSION;
     }
 
     public ParquetTableMetadata_v2(List<ParquetFileMetadata_v2> files, List<String> directories,
         ConcurrentHashMap<ColumnTypeMetadata_v2.Key, ColumnTypeMetadata_v2> columnTypeInfo) {
       this.files = files;
       this.directories = directories;
       this.columnTypeInfo = columnTypeInfo;
+      this.drillVersion = DrillVersionInfo.getVersion();
+      this.writerVersion = ParquetWriter.WRITER_VERSION;
```
paul-rogers: Given a set of files and directories, we're assuming the writer version? This is supposed to be the version of the Drill Parquet writer that wrote the actual Parquet files, right? Not the version of the metadata files that hold information about the Parquet files? Or, are we using the same writer version for both purposes? If we need two versions (one for Parquet, another for metadata), then we should introduce a separate Parquet metadata version. (Or, I'm misunderstanding the code...)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -214,14 +218,24 @@ public static DateCorruptionStatus detectCorruptDates(ParquetMetadata footer,
...
-          return checkForCorruptDateValuesInStatistics(footer, columns, autoCorrectCorruptDates);
         }
       }
+    } catch (VersionParser.VersionParseException e) {
+      // If we couldn't parse "created by" or "drill version", check column metadata of date columns
+      return checkForCorruptDateValuesInStatistics(footer, columns, autoCorrectCorruptDates);
     }
   }
 
+  public static boolean isDrillVersionHasCorrectDates(String drillVersion) throws VersionParser.VersionParseException {
```
paul-rogers: Might flow better as "drillVersionHasCorrectDates"

But, see comment above about whether we need this check.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -189,19 +189,23 @@ public static DateCorruptionStatus detectCorruptDates(ParquetMetadata footer,
...
-        // loop through parquet column metadata to find date columns, check for corrupt values
-        return checkForCorruptDateValuesInStatistics(footer, columns, autoCorrectCorruptDates);
+    String writerVersion = footer.getFileMetaData().getKeyValueMetaData().get(ParquetRecordWriter.WRITER_VERSION_PROPERTY);
+    // This flag can be present in parquet files which were generated with 1.9.0-SNAPSHOT drill version
+    final String isDateCorrectFlag = "is.date.correct";
+    String isDateCorrect = footer.getFileMetaData().getKeyValueMetaData().get(isDateCorrectFlag);
+    try {
+      if (drillVersion != null) {
+        return (writerVersion != null && Integer.parseInt(writerVersion) >= 2) || Boolean.valueOf(isDateCorrect)
+            || isDrillVersionHasCorrectDates(drillVersion)
```
paul-rogers: Isn't the Drill version check redundant? We know that all Drill versions from 1.9.0 onwards will have a Drill Parquet writer version in the file. So, we only need check the writer version, if we have it. If we don't have it, then the only info we have is the Drill version.

We might want to explain this logic in a comment.

The purpose of adding the writer version is that all future format decisions can be made on the writer version independent of Drill version. For example, suppose we change something in Drill 1.10. Drill 1.10.0-SNAPSHOT will start with writer version 2. Later we'll make a change and 1.10.0-SNAPSHOT will writer files with writer version 3. The Drill version is ambiguous, the writer version is spot on.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -214,14 +218,24 @@ public static DateCorruptionStatus detectCorruptDates(ParquetMetadata footer,
...
 
+  public static boolean isDrillVersionHasCorrectDates(String drillVersion) throws VersionParser.VersionParseException {
+    VersionParser.ParsedVersion parsedDrillVersion = parseDrillVersion(drillVersion);
+    SemanticVersion semVer = parsedDrillVersion.getSemanticVersion();
+    // true for 1.9.0, 2.0.0-SNAPSHOT, 2.0.0 etc.                                             (false for 1.9.0-SNAPSHOT, 1.8.0, 1.8.0-SANPSHOT etc.)
+    return semVer != null && semVer.compareTo(new SemanticVersion(1, 9, 0)) >= 0;
+  }
+
+  public static VersionParser.ParsedVersion parseDrillVersion(String drillVersion) throws VersionParser.VersionParseException {
+    return VersionParser.parse("drill version " + drillVersion + " (build 1234)");
```
paul-rogers: The " (build 1234)" deserves some explanation. Does the parse function demand it so that we have to make up something here just to get the version to parse?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -78,7 +78,7 @@
   private static final int MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
 
   public static final String DRILL_VERSION_PROPERTY = "drill.version";
-  public static final String IS_DATE_CORRECT_PROPERTY = "is.date.correct";
+  public static final String WRITER_VERSION_PROPERTY = "parquet-writer.version";
```
paul-rogers: We know we are Drill, but others might be confused. Perhaps "drill-writer.version" or "drill.writer-version" (to be consistent with "drill.version".)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetWriter.java
```diff
@@ -40,6 +40,8 @@
 public class ParquetWriter extends AbstractWriter {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ParquetWriter.class);
 
+  public static final int WRITER_VERSION = 2;
```
paul-rogers: This deserves an explanation. Something like:

Version of Drill's Parquet writer. Increment this version (by 1) any time we make any format change to the file. Format changes include 1) supporting new data types, 2) changes to the format of data fields, 3) adding new metadata to the file footer, etc.

Newer readers must be able to read old files. The Writer version tells the Parquet reader how to interpret fields or metadata when that data changes format from one writer version to another.

vdiravka: I copied this comment into the project.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
paul-rogers: Not sure I completely follow... Here's what I understood:

1. The class is used when ready a Parquet file footer. (In such a case, the writer version would presumably be the actual version from the footer, or null if the file is old, was not written by Drill, etc.)
2. The class is used when reading metadata, but then some fields are not used.
3. In the case where the metadata is for one file (only) is the writer version stored in the metadata? So, object created when reading metadata will have the same information as the original file?
4. If the metadata is for multiple files, then the writer version is never set? (Or, should the rule be that the writer version is set if all files have the same version?)

Since this is a bit hard to understand, it would be very helpful to describe all this in a Javadoc comment for the class.

vdiravka:
@paul-rogers Could you please review? I answered on the last comment and rebased the branch with resolving conflicts (ParquetTableMetadata_v3 was added)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -185,7 +185,8 @@ private Metadata(FileSystem fs, ParquetFormatConfig formatConfig) {
         childFiles.add(file);
       }
     }
-    ParquetTableMetadata_v3 parquetTableMetadata = new ParquetTableMetadata_v3(true);
+    ParquetTableMetadata_v3 parquetTableMetadata = new ParquetTableMetadata_v3(DrillVersionInfo.getVersion(),
+        ParquetWriter.WRITER_VERSION);
```
paul-rogers: I'm a bit confused. The writer version applies to the Parquet files which Drill writes. (Or, at least, that was the intention.)

Here, we're talking about metadata. There may well be a metadata writer, but that should be a different writer, with a different version.

Not sure we want to initialize the metadata object with the current writer version: there seems to be no correlation between the metadata object and the writer version.

On the other hand, the metadata can certainly hold the writer version, but it must be the value read from the Parquet file itself; not a value set by the code. Else, we have the difficult problem of making sure that the code-set version number agrees with the actual version number in the file.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -59,19 +59,24 @@
...
+   * This balances two possible cases of bad auto-correction. External tools writing dates in the future will not
+   * be shifted unless they are past this threshold (and we cannot identify them as external files based on the metadata).
+   * On the other hand, historical dates written with Drill wouldn't risk being incorrectly shifted unless they were
+   * something like 10,000 years in the past.
+   */
   public static final int DATE_CORRUPTION_THRESHOLD =
       (int) (UTC.getDateTimeMillis(5000, 1, 1, 0) / DateTimeConstants.MILLIS_PER_DAY);
-
+  /**
+   * The version of drill parquet writer with date values corruption fix
```
paul-rogers: Maybe explain this a bit better. Something like:

Version 2 (and later) of the Drill Parquet writer uses the date format described (in the Parquet spec? URL?). Prior versions had dates formatted (copy description from above.)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -59,19 +59,24 @@
...
+   */
+  public static final int DRILL_WRITER_VERSION_WITHOUT_CORRUPTION = 2;
```
paul-rogers: Maybe call this DRILL_WRITER_VERSION_STD_DATE_FORMAT

The old format was not "corrupted", it just used a date format that was non-standard.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -189,10 +194,14 @@ public static DateCorruptionStatus detectCorruptDates(ParquetMetadata footer,
 
     String createdBy = footer.getFileMetaData().getCreatedBy();
     String drillVersion = footer.getFileMetaData().getKeyValueMetaData().get(ParquetRecordWriter.DRILL_VERSION_PROPERTY);
-    String isDateCorrect = footer.getFileMetaData().getKeyValueMetaData().get(ParquetRecordWriter.IS_DATE_CORRECT_PROPERTY);
+    String writerVersion = footer.getFileMetaData().getKeyValueMetaData().get(ParquetRecordWriter.WRITER_VERSION_PROPERTY);
+    // This flag can be present in parquet files which were generated with 1.9.0-SNAPSHOT drill version
+    final String isDateCorrectFlag = "is.date.correct";
```
paul-rogers: Maybe here you want to special case the "is.date.correct" flag.

1) If the writer version is present, use it.
2) If "is.date.correct" is present, set the writer version to "2".
3) If neither are present, set the writer version to 1.

That way we don't have to have (much) extra logic for the "is.date.correct" handling.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetWriter.java
```diff
@@ -40,6 +40,16 @@
 public class ParquetWriter extends AbstractWriter {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ParquetWriter.class);
 
+/** Version of Drill's Parquet writer. Increment this version (by 1) any time we make any format change to the file.
+ * Format changes include
+ * 1) supporting new data types,
+ * 2) changes to the format of data fields,
+ * 3) adding new metadata to the file footer, etc.
+ * Newer readers must be able to read old files. The Writer version tells the Parquet reader how to interpret fields
```
paul-rogers: This is Javadoc, so needs HTML formatting. For the list:

    <ul>
    <li>Supporting new data types,</li>
    ...
    </ul>

vdiravka: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
vdiravka: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
vdiravka: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
vdiravka: `is.date.correct` or `parquet-writer.version` were needed in metadata cache file for quick detection of date values correctness. Otherwise need to check `files.rowGroups.columns.mxValue` values from this cache file. 
But thought a little, I've understood that due to new added `ParquetTableMetadata_v3` we can check:
If version of parquet metadata cache file is 3, the date values are definitely correct. Otherwise (when parquet metadata cache file was generated earlier) need to check date values from this file. 
So `writerVersion` is redundant in the `ParquetTableMetadataBase` now. I deleted it. Please approve does it make sense?

vdiravka:
@parthchandra Could you please review this PR?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -189,10 +195,16 @@ public static DateCorruptionStatus detectCorruptDates(ParquetMetadata footer,
...
-    String isDateCorrect = footer.getFileMetaData().getKeyValueMetaData().get(ParquetRecordWriter.IS_DATE_CORRECT_PROPERTY);
+    String stringWriterVersion = footer.getFileMetaData().getKeyValueMetaData().get(ParquetRecordWriter.WRITER_VERSION_PROPERTY);
+    // This flag can be present in parquet files which were generated with 1.9.0-SNAPSHOT drill version.
+    // If this flag is present it means that the version of the drill parquet writer is 2
+    final String isDateCorrectFlag = "is.date.correct";
+    String isDateCorrect = footer.getFileMetaData().getKeyValueMetaData().get(isDateCorrectFlag);
     if (drillVersion != null) {
-      return Boolean.valueOf(isDateCorrect) ? DateCorruptionStatus.META_SHOWS_NO_CORRUPTION
-          : DateCorruptionStatus.META_SHOWS_CORRUPTION;
+      int writerVersion = (stringWriterVersion != null) ? Integer.parseInt(stringWriterVersion)
```
parthchandra: Use if-then-else to make this a little easier to read?

vdiravka: Agree. It makes sense. 
I even found one redundant check in the return statement.
Changes in a new commit.

parthchandra:
+1

PR #647, opened 2016-11-06
DRILL-4935 Allow Drill to advertise a specific hostname to Zookeeper

harrisonmebane:
This addition adds a parameter to the Drill configuration that allows Drill to advertise a specific hostname to Zookeeper (other than the one given by `INetAddress.getLocalHost()`.  The parameter is named `drill.exec.rpc.bit.advertised.host`.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -49,7 +49,7 @@ public interface ExecConstants {
   String USER_RPC_TIMEOUT = "drill.exec.rpc.user.timeout";
   String METRICS_CONTEXT_NAME = "drill.exec.metrics.context";
   String USE_IP_ADDRESS = "drill.exec.rpc.use.ip";
-  String BIT_ADVERTISED_HOST = "drill.exec.rpc.bit.advertised.host";
+  String BIT_ADVERTISED_HOST = "drill.exec.rpc.bit.advertised-host";
   String CLIENT_RPC_THREADS = "drill.exec.rpc.user.client.threads";
   String BIT_SERVER_RPC_THREADS = "drill.exec.rpc.bit.server.threads";
   String USER_SERVER_RPC_THREADS = "drill.exec.rpc.user.server.threads";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -49,7 +49,6 @@ public interface ExecConstants {
   String USER_RPC_TIMEOUT = "drill.exec.rpc.user.timeout";
   String METRICS_CONTEXT_NAME = "drill.exec.metrics.context";
   String USE_IP_ADDRESS = "drill.exec.rpc.use.ip";
-  String BIT_ADVERTISED_HOST = "drill.exec.rpc.bit.advertised-host";
   String CLIENT_RPC_THREADS = "drill.exec.rpc.user.client.threads";
   String BIT_SERVER_RPC_THREADS = "drill.exec.rpc.bit.server.threads";
   String USER_SERVER_RPC_THREADS = "drill.exec.rpc.user.server.threads";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -49,6 +49,7 @@ public interface ExecConstants {
   String USER_RPC_TIMEOUT = "drill.exec.rpc.user.timeout";
   String METRICS_CONTEXT_NAME = "drill.exec.metrics.context";
   String USE_IP_ADDRESS = "drill.exec.rpc.use.ip";
+  String BIT_ADVERTISED_HOST = "drill.exec.rpc.bit.advertised.host";
   String CLIENT_RPC_THREADS = "drill.exec.rpc.user.client.threads";
   String BIT_SERVER_RPC_THREADS = "drill.exec.rpc.bit.server.threads";
   String USER_SERVER_RPC_THREADS = "drill.exec.rpc.user.server.threads";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
@@ -144,9 +144,9 @@ public class ServiceEngine implements AutoCloseable {
     int userPort = userServer.bind(config.getInt(ExecConstants.INITIAL_USER_PORT), allowPortHunting);
     String address = null;
     if (config.hasPath(ExecConstants.BIT_ADVERTISED_HOST)) {
-        address = useIP ?  InetAddress.getLocalHost().getHostAddress() : InetAddress.getLocalHost().getCanonicalHostName();
-    } else {
         address = config.getString(ExecConstants.BIT_ADVERTISED_HOST);
+    } else {
+        address = useIP ?  InetAddress.getLocalHost().getHostAddress() : InetAddress.getLocalHost().getCanonicalHostName();
     }
     checkLoopbackAddress(address);
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
@@ -142,12 +142,11 @@ public class ServiceEngine implements AutoCloseable {
 
   public DrillbitEndpoint start() throws DrillbitStartupException, UnknownHostException{
     int userPort = userServer.bind(config.getInt(ExecConstants.INITIAL_USER_PORT), allowPortHunting);
-    String configIP = config.getString(ExecConstants.BIT_ADVERTISED_HOST);
     String address = null;
-    if (configIP == "") {
+    if (config.hasPath(ExecConstants.BIT_ADVERTISED_HOST)) {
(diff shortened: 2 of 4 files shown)


harrisonmebane:
I'm aware that the tests will not pass yet; putting what I have up here to make sure I'm on the right track.

exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
```diff
@@ -142,7 +142,13 @@ private static BufferAllocator newAllocator(
 
   public DrillbitEndpoint start() throws DrillbitStartupException, UnknownHostException{
     int userPort = userServer.bind(config.getInt(ExecConstants.INITIAL_USER_PORT), allowPortHunting);
-    String address = useIP ?  InetAddress.getLocalHost().getHostAddress() : InetAddress.getLocalHost().getCanonicalHostName();
+    String configIP = config.getString(ExecConstants.BIT_ADVERTISED_HOST);
+    String address = null;
+    if (configIP == "") {
```
xhochy: You could use `config.hasPath` instead of relying on matching on an empty string.

harrisonmebane: Good suggestion!  Much cleaner.

harrisonmebane:
I have implemented @xhochy 's suggestion.  I will need some guidance on the best way to unit test this new configuration.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -49,6 +49,7 @@
   String USER_RPC_TIMEOUT = "drill.exec.rpc.user.timeout";
   String METRICS_CONTEXT_NAME = "drill.exec.metrics.context";
   String USE_IP_ADDRESS = "drill.exec.rpc.use.ip";
+  String BIT_ADVERTISED_HOST = "drill.exec.rpc.bit.advertised.host";
```
paul-rogers: advertised.host -> advertised-host (or advertised_host). Here, "advertised" does not seem to form a group, so no need for dot.

exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
```diff
@@ -142,7 +142,12 @@ private static BufferAllocator newAllocator(
 
   public DrillbitEndpoint start() throws DrillbitStartupException, UnknownHostException{
     int userPort = userServer.bind(config.getInt(ExecConstants.INITIAL_USER_PORT), allowPortHunting);
-    String address = useIP ?  InetAddress.getLocalHost().getHostAddress() : InetAddress.getLocalHost().getCanonicalHostName();
+    String address = null;
```
paul-rogers: Factor out into a getHostName( ) method.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
paul-rogers: Is this really what we want to do? After this change, each host will have its on Drill config file. That means that, on every config change, the admin must:

1. Modify a master file.
2. Use a script to regenerate the per-host files.
3. Push the per-host files to the remote systems.
4. Restart the cluster.

If Drill is deployed on a system such as YARN, it is not possible to have per-host config files.

harrisonmebane: @paul-rogers Thank you for your helpful suggestions.  I understand what you are saying, and indeed, I was using exactly the flow you outlined above for my own use case.

Let me make sure I understand what you are proposing.  We can either:

1. Access the host name as an environment variable, in which case we could access it via `System.getEnv(...)`
2. Pass the host name in as a system variable (``-Ddrill.exec.rpc.bit.advertised-host=myhost`), which could just override the parameter I've already specified in `ExecConstants.java`

In either case, I guess I don't understand why I'd need to add anything to `drill-env.sh`, unless you mean just to add a commented out line for documentation purposes.  In the second case, is there any change I would need to make at all?

harrisonmebane: That all seems reasonable to me.  I can't find any examples of environment variables being accessed from in the Java code.  I assume we don't want the name of the environment variable hard-coded in the hostname resolution code, but I'm not sure of the best place to define it, i.e. ```String DRILL_HOST_NAME = "DRILL_HOST_NAME"```
Thoughts?  I could put it in `ExecConstants` but it would be the only env variable in there.  I could also just make it a static variable in the `ServiceEngine` class.

Do we still want to provide the option to override through a system property, or just rely on the environment variable?

paul-rogers: For an example, see `LogsResources.java` in package `org.apache.drill.exec.server.rest`:

    private File getLogFolder() {
      return new File(Preconditions.checkNotNull(System.getenv("DRILL_LOG_DIR"), "DRILL_LOG_DIR variable is not set"));
    }

The above example hard-codes the env var name. For now, you can probably do the same. When you reference the env var, maybe say something like:

    // Use a custom host name. See drill-env.sh for details.

    ... System.getenv( "DRILL_HOST_NAME" ) ...

Then, in drill-env.sh, add a commented-out entry:

    # Provide a customized host name for when the default mechanism
    # is not accurate.

    # export DRILL_HOST_NAME=`hostname`

I suspect that should be plenty good to document the purpose.

With this, I think you can omit the config variable as Drill's config system doesn't seem to have the right semantics for a per-host setting.

harrisonmebane:
I have made the changes suggested by @paul-rogers .  Are there any unit tests we should add?

paul-rogers:
Two ways to test.

First, try the fix manually. Start Drill once without the env var set to ensure Drill uses the default. Then, set the env var and ensure that Drill uses the correct value.

Second, would be a unit test. I'm not familiar with any existing ZK unit tests. Do any of the other committers know if we have such tests? Something that validates Drillbit registration, etc? If so, then that test can be modified for this use case.

But, there is one gotcha: in Java, env vars are read-only, there is no way to set env vars in a test (that I've ever found.)  Anyone have a work-around?

exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
```diff
@@ -140,9 +140,19 @@ private static BufferAllocator newAllocator(
...
   }
 
+  private String getHostName() throws UnknownHostException{
+    // DRILL_HOST_NAME sets custom host name.  See drill-env.sh for details.
+    String customHost = System.getenv("DRILL_HOST_NAME");
+    if (customHost == null) {
+      return useIP ? InetAddress.getLocalHost().getHostAddress() : InetAddress.getLocalHost().getCanonicalHostName();
+    } else {
+      return customHost;
+    }
```
paul-rogers: Minor code flow suggestion:

    if (customHost == null) {
      return customHost;
    }
    return useIP ? InetAddress.getLocalHost().getHostAddress() : InetAddress.getLocalHost().getCanonicalHostName();

harrisonmebane:
I have tested this fix manually, in the following way:

* Deploy Drill in Docker containers on an existing cluster on AWS, with the line 
```export DRILL_HOST_NAME=`curl http://169.254.169.254/latest/meta-data/local-ipv4` ``` in `drill-env.sh`, to ensure that the variable is populated with the host machine's IP.
* Start a Drill session in the docker container and run `select * from sys.drillbits;`.  The result was: 
```
+----------------+------------+---------------+------------+----------+
|    hostname    | user_port  | control_port  | data_port  | current  |
+----------------+------------+---------------+------------+----------+
| 172.31.21.207  | 31010      | 31011         | 31012      | false    |
| 172.31.29.130  | 31010      | 31011         | 31012      | true     |
| 172.31.22.200  | 31010      | 31011         | 31012      | false    |
+----------------+------------+---------------+------------+----------+
```
* On one of the drillbits, comment out the `DRILL_HOST_NAME` line in `drill-env.sh`, unset the variable, and restart the drillbit.
* Log into Drill shell again and run `select * from sys.drillbits;`  Result is now:
```
+----------------+------------+---------------+------------+----------+
|    hostname    | user_port  | control_port  | data_port  | current  |
+----------------+------------+---------------+------------+----------+
| 172.31.22.200  | 31010      | 31011         | 31012      | false    |
| a53b37888f62   | 31010      | 31011         | 31012      | true     |
| 172.31.29.130  | 31010      | 31011         | 31012      | false    |
+----------------+------------+---------------+------------+----------+
```
* Running queries that require the other nodes results in a `Error: SYSTEM ERROR: UnresolvedAddressException`, indicating that the new address registered on Zookeeper no longer works

paul-rogers:
Great! LGTM.

harrisonmebane:
Do I need to do anything more to push this forward?

paul-rogers:
The committers are in the middle of doing a release. Once that is done, you need one of them to do a quick review and merge your pull request.

zfong:
@harrisonmebane - for pull requests submitted by non-committers, each week, one of the committers will do a batch commit of changes that have gone through review.  This is normally done towards the end of the week.  There may be an exception this week because as @paul-rogers has noted, we're in the middle of trying to push out a release candidate build for the 1.9 release.

paul-rogers:
Please update the JIRA ticket to explain the solution. What does an admin need to know to use the feature? How can the admin verify that it works? This will allow the documentation team to add the needed information for folks to use this feature.

Also, assign the JIRA ticket to yourself, since you're working on it.

harrison-svds:
@paul-rogers I added a comment detailing the basic solution.  I don't know that I have the permissions to assign the story to myself.

parthchandra:
+1

PR #649, opened 2016-11-09
DRILL-4979: Make port of the DataConnection configurable

xhochy:


xhochy:
No, this patch is actually functioning as the data port is already included in the probuf: https://github.com/apache/drill/blob/master/protocol/src/main/protobuf/Coordination.proto#L11 The `+1 logic` is only used in the initialization code for the port.

Also I did a test run with 5 drillbits and random dataports assigned to them. They each were able to talk to them showing that the correct ports are communicated.

paul-rogers:
Thanks for the explanation.
LGTM

parthchandra:
+1

PR #650, opened 2016-11-09
DRILL-5007: Dynamic UDF lazy-init does not work correctly in multi-no…

arina-ielchiieva:
…de cluster


parthchandra:
+1. Good catch guys!

PR #651, opened 2016-11-09
DRILL-5009: Skip reading of empty row groups while reading Parquet me…

parthchandra:
…tadata.

We will no longer attempt to scan such row groups.


sudheeshkatkam:
+1

PR #652, opened 2016-11-15
DRILL-4990:Use new HDFS API access instead of listStatus to check if …

ppadma:
…users have permissions to access workspace.

Manually tested the fix with impersonation enabled. All unit and regression tests pass.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
@@ -70,6 +70,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.permission.FsAction;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.security.AccessControlException;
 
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -151,15 +152,11 @@ public WorkspaceSchemaFactory(
...
-      // We have to rely on the listStatus as a FileSystem can have complicated controls such as regular unix style
-      // permissions, Access Control Lists (ACLs) or Access Control Expressions (ACE). Hadoop 2.7 version of FileSystem
-      // has a limited private API (FileSystem.access) to check the permissions directly
-      // (see https://issues.apache.org/jira/browse/HDFS-6570). Drill currently relies on Hadoop 2.5.0 version of
-      // FileClient. TODO: Update this when DRILL-3749 is fixed.
-      fs.listStatus(wsPath);
+      fs.access(wsPath, FsAction.READ);
     } catch (final UnsupportedOperationException e) {
-      logger.trace("The filesystem for this workspace does not support this operation.", e);
+      logger.debug("The filesystem for this workspace does not support this operation.", e);
```
sohami: We should return false in this case as well. Better to have a local boolean variable (like boolean isAccessAllowed;) and set it accordingly in try and catch. Then have just 1 return statement in the end.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -151,15 +152,11 @@ public WorkspaceSchemaFactory(
   public boolean accessible(final String userName) throws IOException {
     final FileSystem fs = ImpersonationUtil.createFileSystem(userName, fsConf);
     try {
-      // We have to rely on the listStatus as a FileSystem can have complicated controls such as regular unix style
-      // permissions, Access Control Lists (ACLs) or Access Control Expressions (ACE). Hadoop 2.7 version of FileSystem
```
sohami: Also can we please replace the comments with why we are using "fs.access" api instead. Will be good for future use.

ppadma:
I did not add new unit tests because existing tests already provide enough coverage and they run on local file system.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
ppadma: I am not returning false to be consistent with the existing behavior.  If the user does not have permission, we fail when we try to read the file during execution.

sohami:
LGTM

ppadma:
This pull request was never merged because of a problem with windows test setup we have.  As a workaround, I added code to fall back to using old API if new API fails for some reason. All tests are passing fine with this change. 
This is nice to include in 1.12 as it provides performance improvement for all DFS based queries especially when there are large number of files. 
Can we review the new diffs please ?

ppadma:
@sohami Sorabh, since you reviewed the original pull request, can you please review the updated diffs ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -151,17 +152,32 @@ public WorkspaceSchemaFactory(
...
+    }
+
+    // if fs.access fails for some reason, fall back to listStatus.
+    if (tryListStatus) {
+      try {
+        fs.listStatus(wsPath);
+      } catch (final UnsupportedOperationException e) {
+        logger.debug("The filesystem for this workspace does not support listStatus operation.", e);
+      } catch (final FileNotFoundException | AccessControlException e) {
+        logger.debug("file {} not found or cannot be accessed", wsPath.toString(), e);
```
sohami: let's add `using listStatus` in the log statement to differentiate with previous case

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -151,17 +152,32 @@ public WorkspaceSchemaFactory(
...
+      // complicated access control logic underneath.
+      fs.access(wsPath, FsAction.READ);
     } catch (final UnsupportedOperationException e) {
-      logger.trace("The filesystem for this workspace does not support this operation.", e);
+      logger.debug("The filesystem for this workspace does not support access operation.", e);
+      tryListStatus = true;
     } catch (final FileNotFoundException | AccessControlException e) {
-      return false;
+      logger.debug("file {} not found or cannot be accessed", wsPath.toString(), e);
+      tryListStatus = true;
```
sohami: Is access function never trustworthy for negative cases ? Or is it windows platform specific issue ?

ppadma: It can be trusted fine for regular file systems HDFS, MapR FS, Mac OS Extended File System etc.  I am not sure if it works for windows or not. We were never able to figure out whether it is a problem with the test setup or it is actually not supported. 

Updated the diffs with your review comments taken care of. Please take a look.

sohami:
Based on the last in-person discussion it was decided to further look into why the test was failing on Windows platform. Is it the test environment setup issue or an actual issue w.r.t platform implementation of access method ?

kkhatua:
@ppadma can you rebase this with the current master, and include the workaround? It does not make sense to hold up this commit for so long if a workaround for the Windows platform is sufficient.

kkhatua:
You can close this PR, @ppadma .
This seems to have been resolved by @chunhui-shi 's commit (#1032) for DRILL-5089. 
https://github.com/ppadma/drill/commit/18a71a38f6bd1fd33d21d1c68fc23c5901b0080a#diff-b89fd230da4ac69b7a8b6caf7677f77a

ppadma:
this issue is addressed through another PR #1032. closing it.

PR #653, opened 2016-11-16
DRILL-4831: Running refresh table metadata concurrently randomly fail…

ppadma:
…s with JsonParseException

To prevent metadata cache files from getting corrupted, write to a temporary file and do rename at the end. All unit and regression tests pass.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Iterator;
+import java.util.UUID;
 
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
@@ -40,6 +41,9 @@ import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.FileContext;
+import org.apache.hadoop.fs.Options;
+
 import org.apache.parquet.column.statistics.Statistics;
 import org.apache.parquet.hadoop.ParquetFileReader;
 import org.apache.parquet.hadoop.metadata.BlockMetaData;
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -495,31 +499,65 @@ private ParquetFileMetadata_v3 getParquetFileMetadata_v3(ParquetTableMetadata_v3
...
+
+    // Use fileContext API as FileSystem rename is deprecated.
+    FileContext fileContext = FileContext.getFileContext(tmpPath.toUri());
+    Path finalPath = new Path(path, METADATA_FILENAME);
+
+    try {
+      fileContext.rename(tmpPath, finalPath, Options.Rename.OVERWRITE);
+    } catch (Exception e) {
+      logger.info("Rename from {} to {} failed", tmpPath.toString(), finalPath.toString(), e);
+    }
```
gparai: Should we add a `finally` section which removes the temp file if it exists i.e. case rename failed?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -495,31 +499,65 @@ private ParquetFileMetadata_v3 getParquetFileMetadata_v3(ParquetTableMetadata_v3
...
+
+    // Use fileContext API as FileSystem rename is deprecated.
+    FileContext fileContext = FileContext.getFileContext(tmpPath.toUri());
+    Path finalPath = new Path(path,  METADATA_DIRECTORIES_FILENAME);
+
+    try {
+      fileContext.rename(tmpPath, finalPath, Options.Rename.OVERWRITE);
+    } catch (Exception e) {
+      logger.info("Rename from {} to {} failed", tmpPath.toString(), finalPath.toString(), e);
+    }
```
gparai: Same comment as earlier regarding `finally`

gparai: It looks like the function throws an IOException. We should figure out the code doing so and handle it if related to tmp file creation?

ppadma:
Thanks Gautam for the review.  Updated with all review comments addressed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -495,31 +499,75 @@ private ParquetFileMetadata_v3 getParquetFileMetadata_v3(ParquetTableMetadata_v3
...
     ObjectMapper mapper = new ObjectMapper(jsonFactory);
     SimpleModule module = new SimpleModule();
     module.addSerializer(ColumnMetadata_v3.class, new ColumnMetadata_v3.Serializer());
     mapper.registerModule(module);
-    FSDataOutputStream os = fs.create(p);
+
+    // If multiple clients are updating metadata cache file concurrently, the cache file
+    // can get corrupted. To prevent this, write to a unique temporary file and then do
+    // atomic rename.
+    UUID randomUUID =  UUID.randomUUID();
```
amansinha100: Is there a way to get the global queryId created by the Foreman ?  The name of the temp file would best be associated with the queryId.  If that is not possible (I am not sure if the query context is accessible easily from here), at the very least we should re-use the UUID for both the METADATA_FILENAME and METADATA_DIRECTORIES_FILENAME .  It is not a correctness issue but will avoid confusion.

ppadma: Yes, I wanted to use queryId as well. But, it is not easily accessible as you mentioned. I made the change to use same UUID for METADATA_FILENAME and METADATA_DIRECTORIES_FILENAME.

ppadma:
Updated diffs with all review comments taken care of.

amansinha100:
LGTM.  +1

parthchandra:
+1

PR #654, opened 2016-11-17
DRILL-5032: Drill query on hive parquet table failed with OutOfMemoryError

Serhii-Harnyk:
List of columns in every hive partition increases size of serialized physical plan that causes OOM. 
As Hive allows every partition to have its own set of columns now system checks, and if all column sets for all partitions are equal, list of columns stores only on hive table level.

### DIFF
--- a/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveTableWrapper.java
+++ b/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveTableWrapper.java
@@ -26,7 +26,6 @@ import org.apache.hadoop.hive.metastore.api.Order;
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.SerDeInfo;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
-import org.apache.hadoop.hive.metastore.api.Table;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
@@ -35,10 +34,10 @@ import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.google.common.collect.Lists;
 
 @JsonTypeName("table")
-public class HiveTable {
+public class HiveTableWrapper {
 
   @JsonIgnore
-  private Table table;
+  private HiveTableWithColumnCache table;
 
   @JsonProperty
   public String tableName;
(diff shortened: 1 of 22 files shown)


contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -398,12 +398,18 @@ public static void addConfToJob(final JobConf job, final Properties properties)
...
-   * @param partition
-   * @param table
-   * @return
+   * @param partition {@link Partition} instance
+   * @param table {@link Table} instance
+   * @return properties
    */
   public static Properties getPartitionMetadata(final Partition partition, final Table table) {
-    final Properties properties = MetaStoreUtils.getPartitionMetadata(partition, table);
+    final Properties properties;
```
Serhii-Harnyk: @chunhui-shi let me clarify. Do you propose something like creating "partition groups" and storing columns in this "groups"?

chunhui-shi: I was thinking about adding columns to a list if it is not already in the list. This list should belong to the table, and each partition will refer to an index of this list for columns. Do you think this may be a way to store columns?

Serhii-Harnyk:
@chunhui-shi, I have made some changes. Could you please review it?

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -398,17 +399,13 @@ public static void addConfToJob(final JobConf job, final Properties properties)
...
    */
-  public static Properties getPartitionMetadata(final Partition partition, final Table table) {
+  public static Properties getPartitionMetadata(final HivePartition partition, final HiveTable table) {
     final Properties properties;
-    // exactly the same column lists for partitions and table
-    // stored only in table to reduce physical plan serialization
-    if (partition.getSd().getCols() == null) {
-      partition.getSd().setCols(table.getSd().getCols());
-    }
+    restoreColumns(table, partition);
```
chunhui-shi: could this restoreColumns fail the purpose of this fix? Since it is setting columns back to each partition, if getPartitionMetadata was called before the final physical plan generated, then these columns could be printed out to the plan again, could you check this possibility?

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/store/hive/schema/TestColumnListCache.java
```diff
@@ -0,0 +1,78 @@
+/*
```
chunhui-shi: Could you add a unit test to verify when there are two kind (or more) of columns in e.g. 10 partitions, the physical plan text has only two copies of columns?

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
Serhii-Harnyk: No, it couldn't. Before this fix table and partition columns serialized with help of StorageDescriptorWrapper class, but now field "columns" in this class ignored for serialization.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/store/hive/schema/TestColumnListCache.java
Serhii-Harnyk: Done

chunhui-shi:
+1

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveTableWrapper.java
```diff
@@ -0,0 +1,483 @@
...
+      }
+    }
+
+    @JsonIgnore
+    public List<FieldSchemaWrapper> getColumns() {
+      return columns;
+    }
+  }
+
+  public static class SerDeInfoWrapper {
```
Serhii-Harnyk: Fixed

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveTableWrapper.java
```diff
@@ -0,0 +1,483 @@
...
+    @JsonIgnore
+    public void setColumns(List<FieldSchema> columns) {
+      sd.setCols(columns);
+      if (columns != null) {
+        this.columns = Lists.newArrayList();
+        for (FieldSchema fieldSchema : columns) {
+          this.columns.add(new FieldSchemaWrapper(fieldSchema));
+        }
+      } else {
+        this.columns = null;
```
Serhii-Harnyk: This method has stayed from the first realization, and now it does not needed. So I removed it.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveTableWrapper.java
```diff
@@ -0,0 +1,483 @@
...
+    public StorageDescriptorWrapper(StorageDescriptor storageDescriptor) {
+      sd = storageDescriptor;
+      location = storageDescriptor.getLocation();
+      inputFormat = storageDescriptor.getInputFormat();
+      outputFormat = storageDescriptor.getOutputFormat();
+      compressed = storageDescriptor.isCompressed();
+      numBuckets = storageDescriptor.getNumBuckets();
+      serDeInfo = new SerDeInfoWrapper(storageDescriptor.getSerdeInfo());
+      sortCols = Lists.newArrayList();
+      for (Order order : storageDescriptor.getSortCols()) {
```
Serhii-Harnyk: Fixed

Serhii-Harnyk:
@jinfengni, I have made some changes. Could you please review it?

jinfengni:
+1 

LGTM.

PR #655, opened 2016-11-17
DRILL-5047: When session option is string, query profile is displayed…

arina-ielchiieva:
… incorrectly on Web UI

### DIFF
--- a/exec/java-exec/src/main/resources/rest/profile/profile.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/profile.ftl
@@ -107,7 +107,8 @@
   <p>FOREMAN: ${model.getProfile().getForeman().getAddress()}</p>
   <p>TOTAL FRAGMENTS: ${model.getProfile().getTotalFragments()}</p>
 
-  <#if (model.getOptionList()?size > 0)>
+  <#assign options = model.getOptions()>
+  <#if (options?keys?size > 0)>
     <div class="page-header"></div>
     <h3>Session Options</h3>
     <div class="panel-group" id="session-options-accordion">
@@ -129,10 +130,10 @@
                 </tr>
               </thead>
               <tbody>
-                <#list model.getOptionList() as option>
+                <#list options?keys as name>
                   <tr>
-                    <td>${option.getName()}</td>
-                    <td>${option.getValue()?c}</td>
+                    <td>${name}</td>
+                    <td>${options[name]}</td>
                   </tr>
                 </#list>
               </tbody>
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.Maps;
 import org.apache.commons.lang3.tuple.ImmutablePair;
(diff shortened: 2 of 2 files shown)


Ben-Zvi:
LGTM

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -132,7 +132,7 @@
                 <#list model.getOptionList() as option>
                   <tr>
                     <td>${option.getName()}</td>
-                    <td>${option.getValue()?c}</td>
+                    <td>${option.getValue()?string}</td>
```
gparai: I am not familiar with FreeMarker so I do not understand the change. In Drill-4792 you mentioned 
> Since org.apache.drill.exec.server.options.OptionValue.getValue() returns Object, Freemarker built-in c is used to convert Object to string.
Could you please explain why that was not sufficient and how using `string` changes that?

paul-rogers: According to the [documentation](), ?string works only for numbers and formats them for "human display". ?c also works for numbers (and formats them as "C" language constants.)

The challenge is that this template line works for all data types. So, the suggestion of using toString( ) is good. But, since toString( ) may be used for debugging, perhaps add a toDisplayString( ) that will format the value as we want it to appear in the Web UI.

Another issue: are we sure that the value of option is always non-null at this point?

kkhatua:
@arina-ielchiieva Your fix will not conflict, but is in a branch rebased off 4b1902c . 
@sudheeshkatkam  had reverted the commit for DRILL-4373 2 days later. He is using the following branch to prepare for the Apache commit (including the invite to vote).
[Ref: https://github.com/sudheeshkatkam/drill/commits/drill-1.9.0 ]
So, you'll need to rebase it before making a pull request.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
kkhatua: Tried this out. All the options take non-null values.  Any attempt to set a property (String, for e.g.) to null failed. So, I guess as long as a property takes only non-null values, things should be fine.

arina-ielchiieva:
Answering to all comments in one:
1. `${option.getValue()}` is not sufficient since method returns Object and freemarker does not know how to cast it.
2. `?c` only works with boolean and numbers, since options can be string, it not sufficient
3. `?string` works with most of the types but it's true for boolean seems to be deprecated
4. to set null value to option is not possible (handled in SetOptionHandler.class) but option.getValue() may return null according to the code.

Paul's idea to prepare options to display on Java side seems to be the most reasonable.
So I have updated the pull request, added new method in ProfileWrapper.class with transforms `OptionValueList` into `Map<String, String>`. If option value is null, it will be displayed as word 'null'.

sudheeshkatkam:
+1

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -129,10 +130,10 @@
                 </tr>
               </thead>
               <tbody>
-                <#list model.getOptionList() as option>
+                <#list options?keys as name>
```
paul-rogers: Neither this nor the original design displays the keys as sorted. Sorted display is a fairly traditional to save the user from a random search for each value. Since we are changing the behavior, should we fix the sorting issue as well?

arina-ielchiieva:
Replaced `HashMap` with `TreeMap` for sorted display.

gparai:
Thanks for the explanation and the corresponding changes. LGTM.
+1

PR #656, opened 2016-11-17
DRILL-5034: Select timestamp from hive generated parquet always return in UTC

vdiravka:
- TIMESTAMP_IMPALA function is reverted to retaine local timezone
- TIMESTAMP_IMPALA_LOCALTIMEZONE is deleted
- Retain local timezone for the INT96 timestamp values in the parquet files while
  PARQUET_READER_INT96_AS_TIMESTAMP option is on

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
@@ -323,18 +323,28 @@ public class ParquetReaderUtility {
    * @param binaryTimeStampValue
    *          hive, impala timestamp values with nanoseconds precision
    *          are stored in parquet Binary as INT96 (12 constant bytes)
-   *
+   * @param retainLocalTimezone
+   *          parquet files don't keep local timeZone according to the
+   *          <a href="https://github.com/Parquet/parquet-format/blob/master/LogicalTypes.md#timestamp">Parquet spec</a>,
+   *          but some tools (hive, for example) retain local timezone for parquet files by default
+   *          Note: Impala doesn't retain local timezone by default
    * @return  Unix Timestamp - the number of milliseconds since January 1, 1970, 00:00:00 GMT
    *          represented by @param binaryTimeStampValue .
    */
-    public static long getDateTimeValueFromBinary(Binary binaryTimeStampValue) {
+    public static long getDateTimeValueFromBinary(Binary binaryTimeStampValue, boolean retainLocalTimezone) {
       // This method represents binaryTimeStampValue as ByteBuffer, where timestamp is stored as sum of
       // julian day number (32-bit) and nanos of day (64-bit)
       NanoTime nt = NanoTime.fromBinary(binaryTimeStampValue);
       int julianDay = nt.getJulianDay();
       long nanosOfDay = nt.getTimeOfDayNanos();
-      return (julianDay - JULIAN_DAY_NUMBER_FOR_UNIX_EPOCH) * DateTimeConstants.MILLIS_PER_DAY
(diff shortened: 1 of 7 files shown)


vdiravka:
I've rebased the branch to the latest master version.
@bitblender Could you please review?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -323,18 +323,28 @@ public static DateCorruptionStatus checkForCorruptDateValuesInStatistics(Parquet
...
       // julian day number (32-bit) and nanos of day (64-bit)
       NanoTime nt = NanoTime.fromBinary(binaryTimeStampValue);
       int julianDay = nt.getJulianDay();
       long nanosOfDay = nt.getTimeOfDayNanos();
-      return (julianDay - JULIAN_DAY_NUMBER_FOR_UNIX_EPOCH) * DateTimeConstants.MILLIS_PER_DAY
+      long dateTime = (julianDay - JULIAN_DAY_NUMBER_FOR_UNIX_EPOCH) * DateTimeConstants.MILLIS_PER_DAY
           + nanosOfDay / NANOS_PER_MILLISECOND;
+      if (retainLocalTimezone) {
+        return new org.joda.time.DateTime(dateTime, org.joda.time.chrono.JulianChronology.getInstance())
+            .withZoneRetainFields(org.joda.time.DateTimeZone.UTC).getMillis();
```
bitblender: Trying to understand this: Why are you calling .withZoneRetainFields(org.joda.time.DateTimeZone.UTC) if retainLocalTimezone is true ?

vdiravka: `withZoneRetainFields` method calculates the difference between local timezone and UTC (parameter of that method) and returns original dateTime with a shift of that difference. This approach is used frequently in drill code.
But thinking a little more on this I decided that it is possible to use more simpler statement, without creating DateTime object. 
`DateTimeZone.getDefault().convertUTCToLocal(dateTime)`. I think it's more clear.

vdiravka:
@bitblender 
The branch is rebased to the master version.

vdiravka:
@bitblender 
I've disabled one test, which relies on particular timezone (separate commit) and rebased the branch to the master version. Could you please review?

vdiravka:
@bitblender Parquet files with INT96 TIMESTAMP values usually are generated with hive. Those TIMESTAMP values represent the local timezone of the host where the data was written. To read that values hive considers the local timezone (in fact shift between local and UTC timezones is adding to the timestamp values).
The aim of this patch to make the same behaviour like in HIVE while reading parquet INT96 TIMESTAMP vals.
As the result in different timezones we have different data after query or in other words this test depends from the local timezone and `baselineValues` for different timezones will be different. 

However I applied the logic of converting timestamps to the local timezone for the test's `baselineValues`.
So now this test works on the every timezone properly.

The last commit is updated.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -969,12 +969,15 @@ public void testInt96TimeStampValueWidth() throws Exception {
     try {
       testBuilder()
           .ordered()
-          .sqlQuery("select c, d from cp.`parquet/data.snappy.parquet` where d = '2015-07-18 13:52:51'")
+          .sqlQuery("select c, d from cp.`parquet/data.snappy.parquet` " +
+              "where `a` is not null and `c` is not null and `d` is not null")
```
arina-ielchiieva: If you have several baseline values and indicate that output is ordered `order by` clause must be present in query to ensure that result will always be returned in expected order, otherwise use `unOrdered()`.

vdiravka: Done. Thanks.

PR #660, opened 2016-11-20
DRILL-5052: Option to debug generated Java code using an IDE

paul-rogers:
Drill makes extensive use of Java code generation to implement its operators. Drill uses sophisticated techniques to blend generated code with pre-compiled template code. An unfortunate side-effect of this behavior is that it is very difficult to visualize and debug the generated code.

As it turns out, Drill's code-merge facility is, in essence, a do-it-yourself version of subclassing. The Drill "template" is the parent class, the generated code is the subclass. But, rather than using plain-old subclassing, Drill combines the code from the two classes into a single "artificial" packet of byte codes for which no source exists.

This PR modifies the code generation path to optionally allow "plain-old Java" compilation: the generated code becomes a subclass of the template. The generated code can then be compiled as a plain-old Java class with no byte-code fix-up. A further option allows writing the code to a known location that the IDE can search when looking for source files.

With this change, developers can turn on the above feature, set a breakpoint in a template, then step directly into the generated Java code called from the template.

This provides an option to do the above, enabled by developers when needed. The existing byte-code technique is used for production code generation.

Included are a few examples from the External Sort Batch that shows the kinds of minor fix-ups that might be needed for a template in order for the "plain-old Java" technique to work.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/compile/AbstractClassCompiler.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/compile/AbstractClassCompiler.java
@@ -20,12 +20,14 @@ package org.apache.drill.exec.compile;
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.util.Map;
 
 import org.apache.drill.common.util.DrillStringUtils;
 import org.apache.drill.exec.compile.ClassTransformer.ClassNames;
 import org.apache.drill.exec.exception.ClassTransformationException;
 import org.codehaus.commons.compiler.CompileException;
 
+@SuppressWarnings("unused")
 public abstract class AbstractClassCompiler {
   protected boolean debug = false;
 
@@ -74,9 +76,10 @@ public abstract class AbstractClassCompiler {
     return out.toString();
   }
 
-  protected abstract byte[][] getByteCode(ClassNames className, String sourcecode)
+  protected abstract byte[][] getByteCode(final ClassNames className, final String sourcecode)
       throws CompileException, IOException, ClassNotFoundException, ClassTransformationException;
-
+  public abstract Map<String,byte[]> compile(final ClassNames className, final String sourceCode)
+      throws CompileException, IOException, ClassNotFoundException;
   protected abstract org.slf4j.Logger getLogger();
 
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
@@ -1,0 +1,189 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.compile;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.Map;
+
+import org.apache.drill.common.config.DrillConfig;
+import org.apache.drill.exec.compile.ClassTransformer.ClassNames;
+import org.apache.drill.exec.exception.ClassTransformationException;
+import org.apache.drill.exec.expr.CodeGenerator;
+import org.apache.drill.exec.server.options.OptionManager;
+import org.codehaus.commons.compiler.CompileException;
+
+/**
+ * Implements the "plain-old Java" method of code generation and
+ * compilation. Given a {@link CodeGenerator}, obtains the generated
+ * source code, compiles it with the selected compiler, loads the
+ * byte-codes into a class loader and provides the resulting
+ * class. Compared with the {@link ClassTransformer} mechanism,
+ * this one requires the code generator to have generated a complete
+ * Java class that is capable of direct compilation and loading.
+ * This means the generated class must be a subclass of the template
+ * so that the JVM can use normal Java inheritance to associate the
+ * template and generated methods.
+ * <p>
+ * Here is how to use the plain-old Java technique to debug
+ * generated code:
+ * <ul>
+ * <li>Set the config option <var>drill.exec.compile.save_source</var>
+ * to <var>true</var>.</li>
+ * <li>Set the config option <var>drill.exec.compile.code_dir</var>
+ * to the location where you want to save the generated source
+ * code.</li>
+ * <li>Where you generate code (using a {@link CodeGenerator}),
+ * set the "plain-old Java" options:<pre>
+ * CodeGenerator&lt;Foo> cg = ...
+ * cg.plainOldJavaCapable(true); // Class supports plain-old Java
+ * cg.preferPlainOldJava(true); // Actually generate plain-old Java
+ * ...</pre></li>
+ * <li>In your favorite IDE, add to the code lookup path the
+ * code directory saved earlier. In Eclipse, for example, you do
+ * this in the debug configuration you will use to debug Drill.</li>
+ * <li>Set a breakpoint in template used for the generated code.</li>
+ * <li>Run Drill. The IDE will stop at your breakpoint.</li>
+ * <li>Step into the generated code. Examine class field and
+ * local variables. Have fun!</li>
+ * </ul>
+ * <p>
+ * Note: not all generated code is ready to be compiled as plain-old
+ * Java. Some classes omit from the template the proper <code>throws</code>
+ * declarations. Other minor problems may also crop up. All are easy
+ * to fix. Once you've done so, add the following to mark that you've
+ * done the clean-up:<pre>
+ * cg.plainOldJavaCapable(true); // Class supports plain-old Java</pre>
+ * <p>
+ * The setting to prefer plain-old Java is ignored for generated
+ * classes not marked as plain-old Java capable.
+ */
+
+public class ClassBuilder {
+
+  public static final String SAVE_CODE_OPTION = CodeCompiler.COMPILE_BASE + ".save_source";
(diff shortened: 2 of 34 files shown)


paul-rogers:
Revised title to include correct JIRA title.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/JaninoClassCompiler.java
```diff
@@ -58,5 +56,27 @@ public JaninoClassCompiler(ClassLoader parentClassLoader, boolean debug) {
...
+    ClassFile[] classFiles = doCompile( className, sourceCode );
+    Map<String,byte[]> results = new HashMap<>( );
+    for(int i = 0;  i < classFiles.length;  i++) {
+      ClassFile classFile = classFiles[i];
+      results.put( classFile.getThisClassName(), classFile.toByteArray() );
+    }
+    return results;
+  }
+
+  public ClassFile[] doCompile(final ClassNames className, final String sourceCode)
```
arina-ielchiieva: 1. className parameter is never used.
2. May be make method private as in JDKClassCompiler?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Performs the actual work of compiling the code and loading the class.
+   *
+   * @param cg
+   * @return
+   * @throws IOException
+   * @throws CompileException
+   * @throws ClassNotFoundException
+   * @throws ClassTransformationException
+   */
```
arina-ielchiieva: Description for params, return, exceptions?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/CodeCompiler.java
```diff
@@ -33,36 +32,71 @@
...
   }
 
+  /**
+   * Create a single instance of the generated class.
+   *
+   * @param cg
+   * @return
+   * @throws ClassTransformationException
+   * @throws IOException
+   */
```
arina-ielchiieva: Description for param, return and exceptions?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/CodeCompiler.java
```diff
@@ -33,36 +32,71 @@
...
   }
 
+  /**
+   * Create multiple instances of the generated class.
+   *
+   * @param cg
+   * @param count
+   * @return
+   * @throws ClassTransformationException
+   * @throws IOException
```
arina-ielchiieva: Description for param, return and exceptions?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -0,0 +1,190 @@
...
+ * <li>Run Drill. The IDE will stop at your breakpoint.</li>
+ * <li>Step into the generated code. Examine class field and
+ * local variables. Have fun!</li>
+ * </ul>
+ * <p>
+ * Note: not all generated code is ready to be compiled as plain-old
+ * Java. Some classes omit from the template the proper <code>throws</code>
+ * declarations. Other minor problems may also crop up. All are easy
+ * to fix. Once you've done so, add the following to mark that you've
+ * done the clean-up:<pre>
```
arina-ielchiieva: May be to create Jira to add this support?

arina-ielchiieva:
Paul, it's really good feature! I have left some comments in PR.
Other than that:
1. Please format the code to remove unnecessary spaces.
Examples:
addClass( key, results.get( key ) );
public CachedClassLoader( ) {
codeDir = new File( config.getString( CODE_DIR_OPTION ) );

2. May we can add plain-old Java support for Projector in this PR, it's the easiest template to start with.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
```diff
@@ -246,6 +246,12 @@ public void rotateBlock() {
     rotateBlock(BlkCreateMode.TRUE);
   }
 
+  /**
+   * Create a new code block, closing the current block.
+   *
+   * @param mode
+   */
```
arina-ielchiieva: IntelliJ, it's more strict :)

arina-ielchiieva:
Looks good.

paul-rogers:
Rebased on master as master had drifted along in the three weeks this PR has been active.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+    ClassTransformer.ClassNames name = new ClassTransformer.ClassNames(className);
+
+    // A key advantage of this method is that the code can be
+    // saved and debugged, if needed.
+
+    saveCode(code, name);
+
+    // Compile the code and load it into a class loader.
+
+    CachedClassLoader classLoader = new CachedClassLoader();
```
jinfengni: Why do we created a new CachedClassLoader per compileClass() call? Does it mean that we always have cache miss, and end up with re-compile the code?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+    ClassCompilerSelector compilerSelector = new ClassCompilerSelector(classLoader, config, options);
+    Map<String,byte[]> results = compilerSelector.compile(name, code);
+    classLoader.addClasses(results);
+
+    // Get the class from the class loader.
+
+    try {
+      return classLoader.findClass(className);
```
jinfengni: Per the comment above, this findClass() call will always end up with  cache-miss?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/AbstractClassCompiler.java
```diff
@@ -74,9 +76,10 @@ protected String prefixLineNumbers(String code) {
     return out.toString();
   }
 
-  protected abstract byte[][] getByteCode(ClassNames className, String sourcecode)
+  protected abstract byte[][] getByteCode(final ClassNames className, final String sourcecode)
       throws CompileException, IOException, ClassNotFoundException, ClassTransformationException;
-
+  public abstract Map<String,byte[]> compile(final ClassNames className, final String sourceCode)
```
jinfengni: can you please add comment to explain the change (the return type is changed)?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/JaninoClassCompiler.java
```diff
@@ -58,5 +56,27 @@ public JaninoClassCompiler(ClassLoader parentClassLoader, boolean debug) {
...
 
   @Override
+  public Map<String,byte[]> compile(final ClassNames className, final String sourceCode)
+      throws CompileException, IOException, ClassNotFoundException {
+
+    ClassFile[] classFiles = doCompile(sourceCode);
+    Map<String,byte[]> results = new HashMap<>();
+    for(int i = 0;  i < classFiles.length;  i++) {
+      ClassFile classFile = classFiles[i];
+      results.put(classFile.getThisClassName(), classFile.toByteArray());
```
jinfengni: Will it defeat the purpose of caching the compiled class, if the key in the cache is the class name? The class name, if I understand correctly, is like org.apache.drill.exec.test.generated.ProjectGenNNN, where NNN is a sequence number. It's possible that two queries may require same generated code, but with different sequence number. Using class name this way seems to indicate multiple copies of code would be complied.

For now, it is fine if the purpose is for adding debugging option using plain-old class. But it may cause some performance overhead, in the future.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/DrillJavaFileObject.java
```diff
@@ -67,6 +72,22 @@ public boolean isCompiled() {
...
+   * classes.
+   *
+   * @return map of fully-qualified class names to byte codes
+   * for the class
+   */
+
+  public Map<String,byte[]> getClassByteCodes() {
+    Map<String,byte[]> results = new HashMap<>();
+    for(DrillJavaFileObject outputFile : outputFiles.values()) {
+      results.put(outputFile.className, outputFile.outputStream.toByteArray());
```
jinfengni: Same as the comment for JaninoClassComplier regarding the key = ClassName.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/AbstractClassCompiler.java
paul-rogers: We spoke in person. There was some confusion about the two caches involved. One is the original class cache, which has not changed. This is a way for the compiler to give us a (name, code) map so we can match up byte codes to compiled code. This map is used only after the global class cache has decided we have a cache miss.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: Please see note above. Here is what we have:

```
Global class cache --> class --> class loader
```
The above pattern is the same in the byte-code fixup path and in the "plain old Java" path. The Class Loader is Java's way to get at the class.

Why we have one per class? That is the original design and it was replicated here. An advantage is that we don't actually need to add a numeric class suffix (though we do anyway) as each class loader creates its own name space.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: Again, this is the class loader, not the global class cache. For the program to run, this class loader must always find the class, else we have a serious runtime error.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/DrillJavaFileObject.java
paul-rogers: See note above.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/JaninoClassCompiler.java
paul-rogers: As we've discussed, this is the class loader, not the global class cache. The non-ASM path can't get the class name from the byte codes, so the compiler has to tell us, via this map, which byte code corresponds to which compiled class name. The "plain old Java" class loader then fetches classes given their compiled names using this map.

jinfengni:
+1

LGTM. Thanks for the PR!

PR #663, opened 2016-11-21
DRILL-5054: Provide source files of jquery-ui and jquery-dataTables locally

zbdzzg:
Hi,

Currently Drill uses CDN for serving source files of jquery-ui and jquery-dataTables. This is OK for most cases, but not working in an isolated environment.

This is a patch adding these files so that Drill will work fine in intranet.


exec/java-exec/src/main/resources/rest/query/result.ftl
```diff
@@ -11,14 +11,13 @@
 
 <#include "*/generic.ftl">
 <#macro page_head>
-  <link rel="stylesheet" type="text/css" href="//code.jquery.com/ui/1.10.3/themes/smoothness/jquery-ui.css">
-  <link rel="stylesheet" type="text/css" href="//cdn.datatables.net/plug-ins/be7019ee387/integration/jqueryui/dataTables.jqueryui.css">
+  <link rel="stylesheet" type="text/css" href="/static/css/jquery-ui.css">
```
parthchandra: It's a good idea to have both CDN and local loading. Try CDN first and if that fails, try the local set. Typically, the browser would have already got jquery and other common js and css files cached and pages will therefore load faster. 
See here: http://stackoverflow.com/questions/3832446/benefits-vs-pitfalls-of-hosting-jquery-locally
and here: http://stackoverflow.com/questions/1014203/best-way-to-use-googles-hosted-jquery-but-fall-back-to-my-hosted-library-on-go

Also, not too sure about including large js, png, and css files in the source tree. Should we just put in code to try to load from a local location, and provide instructions to users on how to download and set it up?

If we do include the third party scripts, then I would also suggest coding the version number into the name we keep locally: jquery-1.10.2.min.js instead of jquery.min.js, dataTables.colVis-1.1.0.min.js instead of dataTables.colVis.min.js, etc.

zbdzzg: Hi Parthchandra,

I have reconsidered this.
 
Simply putting these things to source code is indeed not a graceful solution, even this method is just temporarily working in my case.

I agree with your second advice: to make these stuffs downloadable and configurable. But as the typical scene for now is that users can already have them in cache, so may be just customizing and recompiling the code for special requirements like this is more reasonable for users like me (the scripts are not fetchable from my area, the network hangs while accessing cdn, util click "stop" actively. To have both the cdn and local files is therefore not work for me), I am so going to shelve this issue util someone report the same problem. Thank you.

PR #664, opened 2016-11-21
DRILL-5058: External sort does not handle its own UserExceptions

paul-rogers:
Solution is to catch the exception, clean up, and fail the operator.

Test case is the existing {{TestSortSpillWithException}} test case.


paul-rogers:
Hold off on this for a bit. It turns out that the proper handling of exceptions is a bit murky. Should a UserException be handled locally in each operator (as are all other exceptions?) That's what this fix does. Or, should all other exceptions be translated to a UserException and thrown up the stack?

PR #665, opened 2016-11-21
DRILL-5056: UserException does not write full message to log

paul-rogers:
A case occurred in which the External Sort failed during spilling. All
that was written to the log was:

2016-11-18 ... INFO  o.a.d.e.p.i.xsort.ExternalSortBatch - User Error
Occurred

Modified the logging code to provide more information to aid tracking
down problems that occur in the field.

Also cleaned up some warnings found while tracking down this issue.

The test case for this is {{TestSortSpillWithException}}, then look at the log output.

### DIFF
--- a/common/src/main/java/org/apache/drill/common/exceptions/UserException.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/UserException.java
@@ -38,6 +38,7 @@ import org.slf4j.Logger;
  * @see org.apache.drill.exec.proto.UserBitShared.DrillPBError.ErrorType
  */
 public class UserException extends DrillRuntimeException {
+  private static final long serialVersionUID = -6720929331624621840L;
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(UserException.class);
 
   public static final String MEMORY_ERROR_MSG = "One or more nodes ran out of memory while executing the query.";
@@ -549,7 +550,15 @@ public class UserException extends DrillRuntimeException {
       if (isSystemError) {
         logger.error(newException.getMessage(), newException);
       } else {
-        logger.info("User Error Occurred", newException);
+        StringBuilder buf = new StringBuilder();
+        buf.append("User Error Occurred");
+        if (message != null) {
+          buf.append(": ").append(message);
+        }
+        if (cause != null) {
+          buf.append(" (").append(cause.getMessage()).append(")");
+        }
+        logger.info(buf.toString(), newException);
       }
 
       return newException;


common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -37,6 +37,7 @@
  *
  * @see org.apache.drill.exec.proto.UserBitShared.DrillPBError.ErrorType
  */
+@SuppressWarnings("serial")
```
Ben-Zvi: Should there instead be in this class :
     private static final long serialVersionUID = -3796081521525479249L;
as does its super-class DrillRuntimeException, and as does its super-class RuntimeException, and as does its super-class Exception, and as does its super-class Throwable (which implements Serializable ........) ??
(Not sure how to pick the specific UID number ...)

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -549,7 +550,12 @@ public UserException build(final Logger logger) {
       if (isSystemError) {
         logger.error(newException.getMessage(), newException);
       } else {
-        logger.info("User Error Occurred", newException);
+        String msg = "User Error Occurred";
+        if (message != null) {
+          msg += ": " + message; }
+        if (cause != null) {
```
Ben-Zvi: Wonder if the root cause (see line 540 above) may also be useful (in case it is below/different from 'cause') ?

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
paul-rogers: There are three ways to handle serializable types. 1) assign a number (start with 1 and increase, say). 2. Let an IDE generate one, such as the one you mentioned. or 3) state that the class is never serialized, so don't worry about the ID, which is the approach I took. Drill never uses Java serialization so the serial ID is unused and unnecessary.

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -549,7 +550,12 @@ public UserException build(final Logger logger) {
       if (isSystemError) {
         logger.error(newException.getMessage(), newException);
       } else {
-        logger.info("User Error Occurred", newException);
+        String msg = "User Error Occurred";
```
sudheeshkatkam: Use StringBuilder?

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -549,7 +550,12 @@ public UserException build(final Logger logger) {
...
+        if (message != null) {
+          msg += ": " + message; }
```
sudheeshkatkam: `}` on new line, here and below.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -224,6 +224,7 @@ public void run() {
           ImpersonationUtil.getProcessUserUGI();
 
       queryUserUgi.doAs(new PrivilegedExceptionAction<Void>() {
+        @Override
```
sudheeshkatkam: Please undo these overrides. Although useful, these are unrelated to the issue. Maybe another ticket?

paul-rogers:
Used a generated serial id for UserException (even though it is never serialized.) Remove the code cleanup. Changed code to use a string builder.

sudheeshkatkam:
+1

PR #666, opened 2016-11-23
DRILL-4956: Temporary tables support

arina-ielchiieva:


exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JsonRecordWriter.java
```diff
@@ -82,6 +84,9 @@ public void init(Map<String, String> writerOptions) throws IOException {
     Path fileName = new Path(location, prefix + "_" + index + "." + extension);
     try {
       stream = fs.create(fileName);
+      // set storage strategy for folder and file
+      storageStrategy.apply(fs, fileName.getParent());
+      storageStrategy.apply(fs, fileName);
```
paul-rogers: Hmmm... The strategy has permission of "775" (permanent) and "700" (temporary). Those include executable. May be fine for the directory. Not so good for the data file. Maybe need an applyToFile and applyToDir? Or, let the storage strategy split the name?

Also, it is not clear if we are creating the directory here, or just changing an existing one. Seems we should honor existing perms on existing dirs; but set then only on new dirs. So, maybe have a mkdir() method that works like mkdirs(): create dir if needed (and imply permissions) or do nothing if the directory exists.

Does writer here create a directory? Will we know to remove it at end of session? We set delete on exit, but exit may be weeks away (for a long-lived Drillbit). Do we know to cascade deletes down into directories if the writer creates a partitioned file?

exec/java-exec/src/main/java/org/apache/drill/exec/store/StorageStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+
+/** Contains list of parameters that will be used to store files on file system. */
+public class StorageStrategy {
+
+  /**
+   * Primary is used for persistent tables,
+   * where owner and group have full access and others can not write,
+   * tables files are not deleted on file system close.
+   */
+  public static final StorageStrategy PERSISTENT = new StorageStrategy("775", false);
```
paul-rogers: How do we know the user wants to be this open? Is there a session (or global) option to lock this down more?

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/TextFormatPlugin.java
```diff
@@ -125,7 +125,7 @@ public RecordWriter getRecordWriter(final FragmentContext context, final EasyWri
 
     options.put("extension", ((TextFormatConfig)getConfig()).getExtensions().get(0));
 
-    RecordWriter recordWriter = new DrillTextRecordWriter(context.getAllocator());
+    RecordWriter recordWriter = new DrillTextRecordWriter(context.getAllocator(), writer.getStorageStrategy());
```
paul-rogers: Stopped here, need to review from here down.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -374,6 +381,13 @@ public void endRecord() throws IOException {
 
   @Override
   public void abort() throws IOException {
+    cleanup();
```
paul-rogers: Does the new code call abort()? Searched master source and found no references. On the other hand, cleanup() is called from WriterRecordBatch. Should the file system cleanup stuff move there?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -382,4 +396,20 @@ public void cleanup() throws IOException {
...
+   * Creates directory if not present, applies storage strategy.
+   *
+   * @return path to files location
+   * @throws IOException during directory creation or permission setting problems
+   */
+  private Path prepareLocationPath() throws IOException {
+    if (locationPath == null) {
+      locationPath = new Path(location);
+      fs.mkdirs(locationPath);
+      storageStrategy.apply(fs, locationPath);
```
paul-rogers: This will apply the permissions only to the deepest of the directories created by mkdirs?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
```diff
@@ -207,25 +239,165 @@ public SchemaPlus getDefaultSchema(SchemaPlus rootSchema) {
...
+
+    /**
+     * Removes temporary table from temporary tables cache
+     * by temporary table original name and its schema.
+     *
+     * @param fullSchemaName table schema name which is used as key in temporary tables cache
+     * @param tableName original table name
+     * @param action action applied to temporary table and its schema before removing from cache
+     */
+    public void remove(String fullSchemaName, String tableName, BiConsumer<AbstractSchema, String> action) {
```
paul-rogers: Oh, clever -- now I see you added a Drill-specific version.

exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/TemporaryTablesAutomaticDropTest.java
```diff
@@ -0,0 +1,74 @@
+/**
```
paul-rogers: Nice set of unit tests! Seems to cover the cases I could think of.

arina-ielchiieva:
@paul-rogers
I have re-designed CTTAS, here is the main changes:
1. Temp table name is generated using  UUID.randowmUUID().
2. All session temp tables are stored in session temp location (corresponds to session id). On session close this folder is deleted recursively.
3. One session can have several temp locations (since default temporary workspace config can change at runtime).
4. Temp table can be created only in one workspace (declared as default in configuration).
5. Permissions are applied on folder and file levels: persistent tables (folder -775, files - 644), temporary tables (folders - 700. files - 600).

Please note, drill-override-example.conf contains link to CTTAS documentation on gist, currently it's empty but I'll update it once CTTAS design is final.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DropTableHandler.java
```diff
@@ -55,35 +58,79 @@ public DropTableHandler(SqlHandlerConfig config) {
...
 
-    if (dropTableNode.checkTableExistence()) {
-      final Table tableToDrop = SqlHandlerUtil.getTableFromSchema(drillSchema, tableName);
-      if (tableToDrop == null || tableToDrop.getJdbcTableType() != Schema.TableType.TABLE) {
-        return DirectPlan.createDirectPlan(context, true,
-            String.format("Table [%s] not found", tableName));
-      }
+    String temporaryTable = context.getSession().findTemporaryTable(tableName);
+    if (temporaryTable == null) {
+      return false;
```
paul-rogers: Here we return false if the name is a temp table, but the temp table does not exist. This sends us down the non-temp table path in the above method. Since the table won't exist there, the user will get an error.

I wonder if this can be written as
```
if ( temp table ) {
  go down temp route
} else {
  go down non-temp route
}
```

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DropTableHandler.java
```diff
@@ -55,35 +58,79 @@ public DropTableHandler(SqlHandlerConfig config) {
...
     }
 
-    drillSchema.dropTable(tableName);
+    AbstractSchema drillSchema = SchemaUtilites.resolveToMutableDrillSchema(defaultSchema,
+        Lists.newArrayList(temporaryWorkspace));
 
-    return DirectPlan.createDirectPlan(context, true,
-        String.format("Table [%s] %s", tableName, "dropped"));
+    Table tableToDrop = SqlHandlerUtil.getTableFromSchema(drillSchema, temporaryTable);
+    if (tableToDrop != null && tableToDrop.getJdbcTableType() == Schema.TableType.TABLE) {
```
paul-rogers: This logic looks very similar to that earlier for non-temp tables. Can it be factored out?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ViewHandler.java
```diff
@@ -49,6 +50,24 @@ public ViewHandler(SqlHandlerConfig config) {
...
   }
 
+  /**
+   * If view to be dropped is in default temporary workspace, checks if it's a temporary table or not.
+   *
+   * @param schema view schema
+   * @param viewName view name to be created
+   * @return true is object to be created is temporary table, false otherwise
+   */
+  protected boolean isTemporaryTable(AbstractSchema schema, String viewName) {
```
paul-rogers: Not sure why we are checking if a view is a temporary table? Or, should viewName be tableName?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ViewHandler.java
```diff
@@ -49,6 +50,24 @@ public ViewHandler(SqlHandlerConfig config) {
...
+    if (schema.getFullSchemaName().equals(context.getConfig().getString(ExecConstants.DEFAULT_TEMPORARY_WORKSPACE))) {
+      String temporaryTableName = context.getSession().findTemporaryTable(viewName);
+      if (temporaryTableName != null) {
+        Table temporaryTable = SqlHandlerUtil.getTableFromSchema(schema, temporaryTableName);
+        return temporaryTable != null && temporaryTable.getJdbcTableType() == Schema.TableType.TABLE;
```
paul-rogers: After reading this same chunk of code in multiple places, I wonder if we can just do this:

```
enum TempTableStatus { NOT_TEMP, NEW_TEMP, EXISTING_TEMP }
class TempTableResult {
    TempTableStatus status;
    String resolvedName;
}

TempTableStatus resolveTempTable(List<String> name, ... ) {
...
```

That is, in one place, do the work of checking if table is temp, if it exists in the name space, and if so, what its (internal?) name is.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ViewHandler.java
```diff
@@ -74,36 +74,51 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ValidationException, RelConv
...
+     * Validates if view can be created in indicated schema:
+     * checks if object (persistent / temporary table) with the same exists
+     * in indicated schema, or if view exists but replace flag is not set.
+     *
+     * @param drillSchema schema where views will be created
+     * @param viewName view name
+     * @param replaceView replace view if exists
+     * @throws UserException if views can be created in indicated schema
+     */
+    private void validateViewCreationPossibility(AbstractSchema drillSchema, String viewName, boolean replaceView) {
```
paul-rogers: So what happens?

```
CREATE TEMPORARY TABLE foo AS ...
CREATE VIEW bar AS SELECT * FROM foo;
```
Logout. Log back in:
```
SELECT * FROM bar;
```
The above fails, right?
```
CREATE TEMPORARY TABLE foo AS ...
SELECT * FROM bar;
```
Does this now succeed? If so, this raises another issue:
```
USE `my-schema`; // contains table foo
CREATE VIEW bar AS SELECT * FROM foo;
SELECT * FROM bar;
```
This selects data from `my-schema.foo`. Fine.
```
CREATE TEMPORARY TABLE foo AS ...
SELECT * FROM bar;
```
Does the existing view bar now resolve to the temp table `foo`? Or, does it remember that it used to resolve to the prior `my-schema.foo`?

This seems like a usability hole. It is one thing to change the meaning of
```
SELECT * FROM foo;
```
After I create a temp table of the same name: clearly I saw table names and knew what I was doing. But, table names in views are hidden.

So, is a solution to forbid the use of temporary tables in creating views?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ViewHandler.java
```diff
@@ -74,36 +93,49 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ValidationException, RelConv
...
+      final String schemaPath = drillSchema.getFullSchemaName();
+      final Table existingTable = SqlHandlerUtil.getTableFromSchema(drillSchema, viewName);
+
+      if ((existingTable != null && existingTable.getJdbcTableType() != Schema.TableType.VIEW) ||
```
paul-rogers: Yet another appearance of the temp table check...

Given how often we do these checks, should we factor them out into functions?

```
void assertTable(List<String> name, Boolean exists, Boolean isTemp)
```
Where this one function checks if the thing is or is not a table, does or does not exist (or doesn't care if exists is null), is or  is not a temp table (again with a doesn't care option).

Having simpler functions makes code review easier and will certainly make changing the code easier as the same logic does not have to be modified in many places.

The form of the function is not important; feel free to improve it. It is the repetition that we want to factor out.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/CompoundIdentifierConverter.java
```diff
@@ -161,7 +161,7 @@ public SqlNode visitChild(
   //SqlNode offset,
   //SqlNode fetch,
     rules.put(SqlSelect.class, R(D, E, D, E, E, E, E, E, D, D));
-    rules.put(SqlCreateTable.class, R(D, D, D, E));
+    rules.put(SqlCreateTable.class, R(D, D, D, E, D));
```
paul-rogers: Stopped here for now.

distribution/src/resources/drill-override-example.conf
```diff
@@ -184,7 +184,12 @@ drill.exec: {
       # Set this property if custom absolute root should be used for remote directories
       root: "/app/drill"
     }
-  }
+  },
+  # Setting for Temporary Tables.
+  # See https://gist.github.com/arina-ielchiieva/50158175867a18eee964b5ba36455fbf#file-temporarytablessupport-md.
+  # Temporary table can be created ONLY in default temporary workspace.
+  # Full workspace name should be indicated (including schema and workspace separated by dot). Workspace MUST be writable.
+  default_temporary_workspace: "dfs.tmp"
```
paul-rogers: Any restrictions here on which workspace can be used? Must it be a file-based workspace? (Probably can't be a JDBC storage plugin, right?) Do we check anywhere that the plugin referenced here is of the correct type?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
```diff
@@ -207,18 +245,85 @@ public SchemaPlus getDefaultSchema(SchemaPlus rootSchema) {
...
+   * @throws IOException if error during session temporary location creation
+   */
+  public String registerTemporaryTable(AbstractSchema schema, String tableName) throws IOException {
+    if (schema instanceof WorkspaceSchemaFactory.WorkspaceSchema) {
+      addTemporaryLocation((WorkspaceSchemaFactory.WorkspaceSchema) schema);
+      String temporaryTableName = Paths.get(sessionId, UUID.randomUUID().toString()).toString();
+      String oldTemporaryTableName = temporaryTables.putIfAbsent(tableName, temporaryTableName);
+      return oldTemporaryTableName == null ? temporaryTableName : oldTemporaryTableName;
+    } else {
+      throw new DrillRuntimeException("Temporary workspace must be instance of WorkspaceSchemaFactory.WorkspaceSchema");
```
paul-rogers: Should this check be done earlier? Say, on start up? Otherwise, the admin won't know about the problem until some random time after each Drillbit starts. Also, since this is a config problem, shouldn't we log this at the error level so that the admin can track down the problem? Telling the query user is fine, but the user is not in a position to fix the problem.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
```diff
@@ -207,18 +245,85 @@ public SchemaPlus getDefaultSchema(SchemaPlus rootSchema) {
...
   }
 
-  public boolean setSessionOption(String name, String value) {
-    return true;
+  /**
+   * Removes temporary table name from the list of session temporary tables.
+   *
+   * @param tableName original table name
+   */
+  public void removeTemporaryTable(String tableName) {
```
paul-rogers: This method is public and sure looks like what should be called to remove a table. But, removing a temp table should delete it from disk since, once it is removed from the table, we know longer know about it, right? Should the work done in the DROP TABLE command shift here so we have a single place that does the deregister-table-and-delete-file as a single operation in a single place?

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JsonRecordWriter.java
```diff
@@ -82,6 +84,9 @@ public void init(Map<String, String> writerOptions) throws IOException {
     Path fileName = new Path(location, prefix + "_" + index + "." + extension);
     try {
       stream = fs.create(fileName);
+      // set storage strategy for folder and file
+      storageStrategy.applyToFolder(fs, fileName.getParent());
```
paul-rogers: Does this handle the case where I ask to create "foo/bar/mumble/myFile" where "foo" exists but "bar" and "mumble" are both new? That is, does it set permissions on the full set of newly created folders? Should the `applyToFolder` also take on the `mkdirs` task so that it can create each folder in the path and set permissions on that folder?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -382,4 +391,20 @@ public void cleanup() throws IOException {
...
+   * Prepares location where files will be written to.
+   * Creates directory if not present, applies storage strategy.
+   *
+   * @return path to files location
+   * @throws IOException during directory creation or permission setting problems
+   */
+  private Path prepareLocationPath() throws IOException {
+    if (locationPath == null) {
+      locationPath = new Path(location);
+      fs.mkdirs(locationPath);
```
paul-rogers: See note above about question about setting permissions on each newly created folder in the path.

exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/TemporaryTablesAutomaticDropTest.java
```diff
@@ -0,0 +1,99 @@
...
```
paul-rogers: Maybe add a test for the mkdirs conditions discussed above? Make sure that permissions are set on the entire set of newly created folders?

exec/rpc/src/main/java/org/apache/drill/exec/rpc/RpcBus.java
```diff
@@ -164,6 +164,8 @@ public void operationComplete(ChannelFuture future) throws Exception {
       }
 
       final ChannelClosedException ex = future.cause() != null ? new ChannelClosedException(msg, future.cause()) : new ChannelClosedException(msg);
+      // close client connection to drop all temporary tables
+      clientConnection.close();
```
paul-rogers: Here we assume that the `close()` won't throw an exception. Clearly, it throws no checked exceptions. But, Drill likes to throw unchecked exceptions. Do we need to worry about that here? Even if such an exception were to occur; we'd still want to close the channel below.

arina-ielchiieva:
@paul-rogers 
I have made changes after second round of CR. Main changes include:
1. disallow view over temporary tables;
2. case-insensitive temporary tables names;
3. temporary workspace validity check at drillbit start up (writable, file-based);
4. permissions are applied to newly created directories only;
5. general refactoring.

Please review new changes when possible.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
```diff
@@ -177,4 +186,34 @@ public static AbstractSchema resolveToMutableDrillSchema(final SchemaPlus defaul
...
+  }
+
+  /**
+   * Checks that passed schema path is the same as temporary workspace path.
+   *
+   * @param schemaPath schema path
+   * @param config drill config
+   * @return true is schema path corresponds to temporary workspace, false otherwise
+   */
+  public static boolean isTemporaryWorkspace(String schemaPath, DrillConfig config) {
```
arina-ielchiieva: According to Drill documentation (and Drill implementation): storage plugin and workspace names are case-sensitive.
For example:
`use dfs.tmp` -> succeeds
`use DFS.TMP` or `use dfs.TMP` -> fails
That's why I have left check for temporary workspace case-sensitive as well.
I'll add comment that check is case-sensitive.

[1] https://drill.apache.org/docs/lexical-structure/ (Storage Plugin and Workspace References section)

arina-ielchiieva:
@paul-rogers 
I have squashed commits and updated documentation to reflect recent changes (also added design document content into Gist). If there are any other changes required, can we put ready-for-commit label?

zfong:
Checked with @paul-rogers, he's ok with the changes.  I'll add the ready-to-commit tag to the Jira.

distribution/src/resources/drill-override-example.conf
```diff
@@ -184,7 +184,13 @@ drill.exec: {
       # Set this property if custom absolute root should be used for remote directories
       root: "/app/drill"
     }
-  }
+  },
+  # Settings for Temporary Tables.
+  # See https://gist.github.com/arina-ielchiieva/50158175867a18eee964b5ba36455fbf#file-temporarytablessupport-md.
```
jinfengni: In stead of pointing to a gist file in Drill source code, can we point to DRILL-4956 which points to such document? After Drill doc updates with the temporary table, we can update the link in DRILL-4956.  That way, we do not have to update the code to change the link later on.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ViewHandler.java
```diff
@@ -62,9 +62,10 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ValidationException, RelConv
 
       final String newViewName = createView.getName();
 
+      // Disallow temporary tables usage in view definition
+      config.getConverter().disallowTemporaryTables();
```
jinfengni: Looks like we only disable temporary tables in create view flow? What about view expansion logic? 

Let's say under "dfs.tmp", I have a temporary table "t1" which is masked into what_ever_random_name, and I also have folder "t1" for a regular table "t1".  When we create a view on top of "t1", since temporary table is disabled, we will resolve to this regular table "t1". However, when we run a query against a view, do we resolve to the temporary table in stead? Is this what we expect to see?

jinfengni:
+1

LGTM

PR #667, opened 2016-11-23

spanchamiamapr:
We enabled _id only projection for JSON tables in MapR v5.2.0
Hence, removing the inefficient workaround which was getting entire document when
1. Only '_id' is projected, or
2. When the user runs a count(*) query


Agirish:
Hey @spanchamiamapr, may be we should create an Apache JIRA for this? Let me know if you want me to.

Agirish:
DRILL-5065 has been filed to track this issue. Kindly update the issue title and description accordingly. Thanks!

PR #668, opened 2016-11-24
DRILL-5068: Add a new system table for completed profiles

zhztheplayer:
Add table "sys.profiles" for completed queries.

Following fields added:

1. queryID (String)
2. time (Timestamp)
3. latency (long)
4. user (String)
5. query (String)
6. state (String)


kkhatua:
Addressed in #1077

PR #669, opened 2016-11-28
DRILL-5044: After the dynamic registration of multiple jars simultane…

arina-ielchiieva:
…ously not all UDFs were registered

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/CreateFunctionHandler.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/CreateFunctionHandler.java
@@ -90,7 +90,7 @@ public class CreateFunctionHandler extends DefaultSqlHandler {
 
       jarManager.initRemoteBackup();
       List<String> functions = validateAgainstLocalRegistry(jarManager, context.getFunctionRegistry());
-      initRemoteRegistration(functions, jarManager, remoteRegistry, remoteRegistry.getRetryAttempts());
+      initRemoteRegistration(functions, jarManager, remoteRegistry);
       jarManager.deleteQuietlyFromStagingArea();
 
       return DirectPlan.createDirectPlan(context, true,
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/CreateFunctionHandler.java
```diff
@@ -175,22 +175,20 @@ private void initRemoteRegistration(List<String> functions,
...
     jarManager.copyToRegistryArea();
-    boolean cleanUp = true;
     List<Jar> jars = Lists.newArrayList(remoteJars);
     jars.add(Jar.newBuilder().setName(jarManager.getBinaryName()).addAllFunctionSignature(functions).build());
     Registry updatedRegistry = Registry.newBuilder().addAllJar(jars).build();
     try {
       remoteRegistry.updateRegistry(updatedRegistry, version);
-      cleanUp = false;
     } catch (VersionMismatchException ex) {
+      jarManager.deleteQuietlyFromRegistryArea();
```
paul-rogers: I wonder about the retry approach. If we get a version mis-match, it means that something has changed: some other agent is modifying the registry at the same time. How should this be handed? Fixed number of retry? Wait and retry? Presumably, after some time, the system will get less busy, or we'll finally get our crack at the registry.

Also, is it wise to use recursion to retry? Doing so causes a stack depth proportional to the retry count, which may not be ideal. Loop instead?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/CreateFunctionHandler.java
```diff
@@ -175,22 +175,20 @@ private void initRemoteRegistration(List<String> functions,
...
       if (retryAttempts-- == 0) {
         throw new DrillRuntimeException("Failed to update remote function registry. Exceeded retry attempts limit.");
       }
       initRemoteRegistration(functions, jarManager, remoteRegistry, retryAttempts);
-    } finally {
-      if (cleanUp) {
-        jarManager.deleteQuietlyFromRegistryArea();
-      }
+    } catch (Exception e) {
```
paul-rogers: Since recursion is used on retry, won't we end up deleting the jar multiple times? No harm, but seems odd...

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DropFunctionHandler.java
```diff
@@ -143,7 +143,7 @@ private Jar unregister(String jarName, RemoteFunctionRegistry remoteFunctionRegi
         if (retryAttempts-- == 0) {
           throw new DrillRuntimeException("Failed to update remote function registry. Exceeded retry attempts limit.");
         }
-        unregister(jarName, remoteFunctionRegistry, retryAttempts);
+        return unregister(jarName, remoteFunctionRegistry, retryAttempts);
```
paul-rogers: Comments from registration process apply here as well.

exec/java-exec/src/test/java/org/apache/drill/TestDynamicUDFSupport.java
```diff
@@ -271,6 +271,75 @@ public void testDuplicatedFunctionsInLocalRegistry() throws Exception {
...
+  }
+
+  @Test
+  public void testSuccessfulUnregistrationAfterSeveralRetryAttempts() throws Exception {
+    RemoteFunctionRegistry remoteFunctionRegistry = spyRemoteFunctionRegistry();
+    copyDefaultJarsToStagingArea();
+    test("create function using jar '%s'", default_binary_name);
+
+    reset(remoteFunctionRegistry);
+    doThrow(new VersionMismatchException("Version mismatch detected", 1))
```
paul-rogers: Perhaps a few words of explanation about what doThrow does?

exec/java-exec/src/test/java/org/apache/drill/TestDynamicUDFSupport.java
```diff
@@ -271,6 +271,75 @@ public void testDuplicatedFunctionsInLocalRegistry() throws Exception {
   }
 
   @Test
+  public void testSuccessfulRegistrationAfterSeveralRetryAttempts() throws Exception {
```
paul-rogers: The case in question would seem to occur when multiple registrations happen concurrently. Do we have a test for that case? That is, do we try to run two thread (two Drillbits?) and have them each attempt to register jars?

arina-ielchiieva:
@paul-rogers, I have made changes after CR, please review when possible.

sudheeshkatkam:
+1

PR #670, opened 2016-11-28

spanchamiamapr:


adityakishore:
Could you please add couple of unit test to verify the plan?

PR #672, opened 2016-11-30
DRILL-5085: Add / update description for dynamic UDFs directories in …

arina-ielchiieva:
…drill-env.sh and drill-module.conf

### DIFF
--- a/distribution/src/resources/drill-env.sh
+++ b/distribution/src/resources/drill-env.sh
@@ -142,3 +142,8 @@
 # Drill is embedded in Sqlline or not.
 
 #export DRILL_SHELL_JAVA_OPTS="..."
+
+# Location Drill should use for temporary files, such as downloaded dynamic UDFs jars.
+# Set to "/tmp" by default.
+#
+# export DRILL_TMP_DIR="..."
--- a/distribution/src/resources/drill-override-example.conf
+++ b/distribution/src/resources/drill-override-example.conf
@@ -170,7 +170,21 @@ drill.exec: {
     threadpool_size: 8,
     decode_threadpool_size: 1
   },
-  debug.error_on_leak: true
+  debug.error_on_leak: true,
+  # Settings for Dynamic UDFs.
+  # See https://gist.github.com/arina-ielchiieva/a1c4cfa3890145c5ecb1b70a39cbff55#file-dynamicudfssupport-md.
+  udf: {
+    # number of retry attempts to update remote function registry
+    # if registry version was changed during update
+    retry-attempts: 10,
+    directory: {
+      # Override this property if custom file system should be used to create remote directories
+      # instead of default taken from Hadoop configuration
+      fs: "hdfs:///",
+      # Set this property if custom absolute root should be used for remote directories
+      root: "/app/drill"
+    }
+  }
 }
 
 # Below SSL parameters need to be set for custom transport layer settings.
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
@@ -377,7 +377,6 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
    * Creates local udf directory, if it doesn't exist.
    * Checks if local udf directory is a directory and if current application has write rights on it.
    * Attempts to clean up local udf directory in case jars were left after previous drillbit run.
-   * Local udf directory path is concatenated from drill temporary directory and ${drill.exec.udf.directory.local}.
    *
    * @param config drill config
    * @return path to local udf directory
@@ -395,22 +394,24 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
     } catch (IOException e) {
       throw new DrillRuntimeException("Error during local udf directory clean up", e);
     }
+    logger.info("Created and validated local udf directory [{}]", udfPath);
     return new Path(udfDir.toURI());
   }
 
   /**
-   * First tries to get drill temporary directory value from environmental variable $DRILL_TMP_DIR,
-   * then from config ${drill.tmp-dir}.
+   * First tries to get drill temporary directory value from from config ${drill.tmp-dir},
+   * then checks environmental variable $DRILL_TMP_DIR.
    * If value is still missing, generates directory using {@link Files#createTempDir()}.
    * If temporary directory was generated, sets {@link #deleteTmpDir} to true
    * to delete directory on drillbit exit.
    * @return drill temporary directory path
    */
   private File getTmpDir(DrillConfig config) {
-    String drillTempDir = System.getenv("DRILL_TMP_DIR");
-
-    if (drillTempDir == null && config.hasPath(ExecConstants.DRILL_TMP_DIR)) {
+    String drillTempDir;
+    if (config.hasPath(ExecConstants.DRILL_TMP_DIR)) {
       drillTempDir = config.getString(ExecConstants.DRILL_TMP_DIR);
+    } else {
+      drillTempDir = System.getenv("DRILL_TMP_DIR");
     }
 
     if (drillTempDir == null) {
@@ -477,7 +478,7 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
    * Will unregister all functions associated with the jar name
    * and delete binary and source associated with the jar from local udf directory
    */
-  public class UnregistrationListener implements TransientStoreListener {
+  private class UnregistrationListener implements TransientStoreListener {
 
     @Override
     public void onChange(TransientStoreEvent event) {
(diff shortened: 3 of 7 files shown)


distribution/src/resources/drill-env.sh
```diff
@@ -141,4 +141,9 @@
...
 # Drill is embedded in Sqlline or not.
 
-#export DRILL_SHELL_JAVA_OPTS="..."
\ No newline at end of file
+#export DRILL_SHELL_JAVA_OPTS="..."
+
+# Location Drill should use for temporary files, such as downloaded dynamic UDFs jars.
+# Set to "/tmp" by default.
+#
+# export DRILL_TMP_DIR="..."
```
paul-rogers: Nit: add a newline after this line.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -207,11 +207,12 @@ drill.exec: {
...
 
-      // Relative base directory for all udf directories (local and remote).
+      // Base directory for all udf directories (local and remote).
       base: ${drill.exec.zk.root}"/udf",
 
-      // Relative path to local udf directory.
-      // Generated temporary directory is used as root unless ${DRILL_TMP_DIR} or ${drill.tmp-dir} is set.
+      // Relative local udf directory. Drill temporary directory is used as root for this directory.
+      // Drill temporary directory is set to generated temporary directory
+      // unless ${DRILL_TMP_DIR} or ${drill.tmp-dir} is set.
```
paul-rogers: Still not sure this entirely makes sense. This is a local file system directory. We describe ow the default is based on DRILL_TMP_DIR, etc. That all make sense.

But the actual default value is based on the base directory in the file system configured by fs, which might be HDFS.

And, since we specify a value here, the `drill.tmp.dir` will never be used.

Do we really mean:
```
local: "${drill-tmp-dir}/udfs"
```

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -45,7 +45,7 @@ drill.client: {
   supports-complex-types: true
 }
 
-// Directory is used as base for temporary storage of Dynamic UDF jars.
+// Directory is used as root for temporary storage of Dynamic UDF jars.
```
paul-rogers: Where does the user configure the suffix added to this directory to get the temp dir for a given cluster? For example, should we have something like:

drill.cluster-temp-dir: "${drill.tmp-dir}/${drill.exec.zk.root}"

Note that tmp itself is a shared resource. If it is in /tmp, then multiple clusters will use it. If it is in $DRILL_HOME/tmp, then still multiple clusters will use it. Only if it is configured as $DRILL_SITE/tmp can we be guaranteed only one cluster will use the directory (because only one cluster can use each $DRILL_SITE dir -- that's how $DRILL_SITE is defined.)

paul-rogers:
Oh, and can we copy some of these changes into drill-override-example.conf in distribution/src/resources? Users see the defaults & explanations there, but not those in drill-module.conf.

Maybe just fs, base and local. The others should be fine at their default value for almost all users.

distribution/src/resources/drill-env.sh
arina-ielchiieva: Done.

exec/java-exec/src/main/resources/drill-module.conf
arina-ielchiieva: Agree. Added new config parameter ${drill.cluster-temp-dir}

exec/java-exec/src/main/resources/drill-module.conf
arina-ielchiieva: Updated to local: "${drill.cluster-tmp-dir}/udf/local".

arina-ielchiieva:
1. Made changes after CR.
2. Added udf config example into drill-override-example.conf.
3. Minor refactoring.

@paul-rogers please review when possible.

distribution/src/resources/drill-override-example.conf
```diff
@@ -170,7 +170,17 @@ drill.exec: {
     threadpool_size: 8,
     decode_threadpool_size: 1
   },
-  debug.error_on_leak: true
+  debug.error_on_leak: true,
+  udf: {
```
paul-rogers: Comment?

# Settings for Dynamic UDFs. See (link to docs)

distribution/src/resources/drill-override-example.conf
```diff
@@ -170,7 +170,17 @@ drill.exec: {
...
+    retry-attempts: 10,
```
paul-rogers: Comment to explain this? Just lift wording from the docs.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
```diff
@@ -377,14 +374,12 @@ private ScanResult scan(ClassLoader classLoader, Path path, URL[] urls) throws I
...
    * Attempts to clean up local udf directory in case jars were left after previous drillbit run.
-   * Local udf directory path is concatenated from drill temporary directory and ${drill.exec.udf.directory.local}.
    *
    * @param config drill config
    * @return path to local udf directory
    */
   private Path getLocalUdfDir(DrillConfig config) {
-    tmpDir = getTmpDir(config);
-    File udfDir = new File(tmpDir, config.getString(ExecConstants.UDF_DIRECTORY_LOCAL));
+    File udfDir = new File(config.getString(ExecConstants.UDF_DIRECTORY_LOCAL));
```
paul-rogers: Although we talked about using the new system you've implemented here, we have to consider backward compatibility.

Since the original behavior is already visible to users in Drill 1.8, I think we need to leave your original design.

At some point, we'll need to rationalize how Drill handles temp files and storage in DFS. But, until then, your 1.8 design is fine.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/RemoteFunctionRegistry.java
```diff
@@ -189,6 +188,7 @@ private void prepareStores(PersistentStoreProvider storeProvider, ClusterCoordin
    * if not set, uses user home directory instead.
    */
   private void prepareAreas(DrillConfig config) {
+    logger.info("Preparing three remote udf areas: staging, registry and tmp.");
```
paul-rogers: Although we suggested keeping the file path stuff as-is, the new logging messages are great; let's keep them.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -45,11 +45,13 @@ drill.client: {
   supports-complex-types: true
 }
 
-// Directory is used as base for temporary storage of Dynamic UDF jars.
-// Set this property if you want to have custom temporary directory, instead of generated at runtime.
-// By default ${DRILL_TMP_DIR} is used if set.
-// drill.tmp-dir: "/tmp"
-// drill.tmp-dir: ${?DRILL_TMP_DIR}
+// Location Drill uses for temporary files, such as downloaded dynamic UDFs jars.
```
paul-rogers: Can we leave it that, if blank, it will use the system-generated directory as you have in 1.9? This means we'll still need your getTmpDir method. To do the check for blank.

Also, as it turns out, we already have a temp dir setting used elsewhere. Look in drill-override-example.conf under drill.sys.store.provider.local.path. It uses "/tmp/drill". Then it adds subdirs for storage plugins and what-not.

Can we use "/tmp/drill/udf" as the base temp storage location for udfs to be consistent? Then we'd add the cluster-id, etc. as in your "cluster-temp-dir" below.

And, yes, we have lots of places where we specify the temp file system (UDFs, the one cited above, the (unused) cache.hazel, the temp dirs for spilling, ...). We do need to create a project (later) to rationalize all of this...

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
arina-ielchiieva: Ok, reverted usage of generated temporary directory logic.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/RemoteFunctionRegistry.java
arina-ielchiieva: Sure.

exec/java-exec/src/main/resources/drill-module.conf
arina-ielchiieva: 1. Reverted generated directory logic.
2. "/tmp/drill/udf" + cluster-id is not good idea, since local udf directory has similar location as remote. And remote directory is generated as cluster-id + /udf. So I suggest to keep it as is. Local registry is returned to use its previous composition logic.
3. Removed "cluster-temp-dir", we don't need it for now.

arina-ielchiieva:
@paul-rogers made changes after 2CR. Partially reverted temporary directory implementation logic to provide backward compatibility. Please review.

sudheeshkatkam:
+1

PR #675, opened 2016-12-02
DRILL-5094: Comparator should guarantee transitive attribute.

chunhui-shi:


zfong:
Looks good to me.  +1 (non-binding)

exec/java-exec/src/main/java/org/apache/drill/exec/store/schedule/AssignmentCreator.java
```diff
@@ -50,7 +50,8 @@
   private static Comparator<Entry<DrillbitEndpoint,Long>> comparator = new Comparator<Entry<DrillbitEndpoint,Long>>() {
     @Override
     public int compare(Entry<DrillbitEndpoint, Long> o1, Entry<DrillbitEndpoint,Long> o2) {
-      return (int) (o1.getValue() - o2.getValue());
+      long ret = o1.getValue() - o2.getValue();
```
laurentgo: Alternatively: 
`
return o1.getValue().compareTo(o2.getValue());
`

exec/java-exec/src/main/java/org/apache/drill/exec/store/schedule/AssignmentCreator.java
```diff
@@ -50,7 +50,8 @@
...
+      return ret > 0? 1 : ((ret < 0)? -1: 0);
```
laurentgo: this doesn't work for `o1.getValue() == Long.MAX_VALUE` and `o2.getValue() == Long.MIN_VALUE`

amansinha100:
Seems like some of the review spilled over to the JIRA.   @chunhui-shi  and @laurentgo , did you come to an agreement ? :)   I am fine with either way..this is a small change.

amansinha100:
+1

amansinha100:
To clarify, I decided to accept the fix because in the context of the comparator call, the o2.getValue() should never be negative (since it is the #bytes of data used for affinity).  Strictly speaking, @laurentgo is right about potential for overflow for the subtraction, so I would suggest @chunhui-shi keep that in mind for future, particularly since it doesn't make sense to try and optimize this particular code path given that it is a one-time planner decision, not run-time code :)  Let's move on to more complex problems :)

PR #677, opened 2016-12-05

spanchamiamapr:

Change the format plugin config parameter name.


contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBFormatPluginConfig.java
```diff
@@ -65,6 +68,15 @@ public void setAllTextMode(boolean mode) {
     allTextMode = mode;
   }
 
+  @JsonProperty("disableCountOptimization")
+  public void setReadDocumentForCount(boolean mode) {
```
adityakishore: setter function does not match the parameter.

PR #678, opened 2016-12-05

spanchamiamapr:

Change the format plugin config parameter name.

Fix setter of config parameter `disableCountOptimization` for drill-maprdb plugin


adityakishore:
+1

LGTM.

PR #679, opened 2016-12-06
DRILL-5098: Improving fault tolerance for connection between client a…

sohami:
…nd foreman node.

         Note: Adding tries config option in connection string.
               Improving fault tolerance in Drill client when trying to make first connection with foreman.
               The client will try to connect to min(tries, num_drillbits) unique drillbits unless a successfull
               connection is established.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
@@ -314,6 +314,7 @@ public class DrillClient implements Closeable, ConnectionThrottle {
     }
 
     final List<DrillbitEndpoint> endpoints = new ArrayList<>();
+
     if (isDirectConnection) {
       // Populate the endpoints list with all the drillbit information provided in the connection string
       endpoints.addAll(parseAndVerifyEndpoints(props.getProperty("drillbit"),
@@ -334,7 +335,6 @@ public class DrillClient implements Closeable, ConnectionThrottle {
 
     // shuffle the collection then get the first endpoint
     Collections.shuffle(endpoints);
-    final DrillbitEndpoint endpoint = endpoints.get(0);
 
     if (props != null) {
       final UserProperties.Builder upBuilder = UserProperties.newBuilder();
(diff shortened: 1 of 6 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/client/ConnectTriesPropertyTestClusterBits.java
```diff
@@ -0,0 +1,244 @@
...
+      bitInfo.append(currentBitIp);
+      bitInfo.append(":");
+      bitInfo.append(currentBitPort);
+    }
+  }
+
+  @AfterClass
+  public static void testCleanUp(){
+    for(int i=0; i < drillBitCount; ++i){
+      drillbits.get(i).close();
```
sudheeshkatkam: + spacing in the signature
+ Use `AutoCloseables.close(drillbits);`

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/JdbcConnectTriesTestEmbeddedBits.java
```diff
@@ -0,0 +1,162 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class JdbcConnectTriesTestEmbeddedBits extends JdbcTestBase {
+
+  @Test
+  public void testDirectConnectionConnectTriesEqualsDrillbitCount() throws SQLException {
+    Connection connection = null;
+    try {
+      connection = new Driver().connect("jdbc:drill:drillbit=127.0.0.1:5000,127.0.0.1:5001;" + "tries=2",
```
sudheeshkatkam: Create one driver and use across tests.

exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
```diff
@@ -357,10 +357,54 @@ protected void afterExecute(final Runnable r, final Throwable t) {
...
+        logger.error("Connection to {}:{} failed with error {}. Not retrying anymore", endpoint.getAddress(),
+                     endpoint.getUserPort(), ex.getMessage());
+        throw ex;
+      } catch (RpcException ex) {
+        ++triedEndpointIndex;
+        logger.error("Attempt {}: Failed to connect to server {}:{}", triedEndpointIndex, endpoint.getAddress(),
+                     endpoint.getUserPort());
+
+        // Close the connection
+        if (client.isActive()) {
```
sohami: Here client is userClient and "isActive" check for all the below conditions:

return connection != null
        && connection.getChannel() != null
        && connection.getChannel().isActive();

and close --> closes the corresponding channel.

Since we are retrying so other resources will be reused for next connection attempt. After all retries are exhausted those will be closed by the DrillClient.close method.

exec/java-exec/src/test/java/org/apache/drill/exec/client/ConnectTriesPropertyTestClusterBits.java
sohami: Fixed

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/JdbcConnectTriesTestEmbeddedBits.java
sohami: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
sudheeshkatkam: The loop recreates a new client anyway. So this should be closed, regardless of being active.

sohami: I looked more and to me it looks like we are not doing the close in BasicClient properly.

All the ServerConnection / ControlConnection / DataClientConnection are derived from RemoteConnection.

RemoteConnection itself has a close method which check for channel isActive and closes the channel. This is what is mimicked inside BasicClient::Close(). Ideally I think we should call the connection.close() (after checking for null) inside the BasicClient::close() method. And in RemoteConnection::close() we should check for null of channel and isActive and then call close on the channel. Let me know if it sounds good to you.

exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
```diff
@@ -357,10 +357,53 @@ protected void afterExecute(final Runnable r, final Throwable t) {
...
+        logger.error("Connection to {}:{} failed with error {}. Not retrying anymore", endpoint.getAddress(),
+                     endpoint.getUserPort(), ex.getMessage());
+        throw ex;
+      } catch (RpcException ex) {
+        ++triedEndpointIndex;
+        logger.error("Attempt {}: Failed to connect to server {}:{}", triedEndpointIndex, endpoint.getAddress(),
+                     endpoint.getUserPort());
+
+        // Close the connection
+        client.close();
```
sudheeshkatkam: Is `UserClient#close` idempotent?

Looks like `if (triedEndpointIndex == connectTriesVal)`, then `UserClient#close` is called twice, here and in `DrillClient#close`. The is unlike the previous catch clause i.e. `catch (InvalidConnectionInfoException ex)`.

sohami: I didn't found anywhere in documentation specifying the cause of calling close multiple times. It just says that internally close is being called for each handler in the pipeline. https://netty.io/4.0/api/io/netty/channel/Channel.html#close()

Though I do have unit tests which will result in covering that code path where close will be called twice. Didn't found any issue running the test.

sudheeshkatkam:
+1

PR #680, opened 2016-12-07
DRILL-5108: Reduce output from Maven git-commit-id-plugin

paul-rogers:
The git-commit-id-plugin grabs information from Git to display during a
build. It prints many e-mail addresses and other generic project
information. As part of the effort to trim down unit test output, we
propose to turn off the verbose output from this plugin by default.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -383,7 +383,7 @@
 
         <configuration>
           <dateFormat>dd.MM.yyyy '@' HH:mm:ss z</dateFormat>
-          <verbose>true</verbose>
+          <verbose>false</verbose>
           <skipPoms>false</skipPoms>
           <generateGitPropertiesFile>true</generateGitPropertiesFile>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -383,7 +383,7 @@
 
         <configuration>
           <dateFormat>dd.MM.yyyy '@' HH:mm:ss z</dateFormat>
-          <verbose>true</verbose>
+          <verbose>false</verbose>
```
Ben-Zvi: Can this be undone by giving the mvn command line option  -Dverbose  ?

sudheeshkatkam:
What does the non verbose message look like?

paul-rogers:
The non-verbose output is, indeed, blank. The value of the plugin seems to be that it creates Maven variables that can be used elsewhere in the POM for various purposes. Dumping to console is an "extra."

sudheeshkatkam:
+1

PR #681, opened 2016-12-07
DRILL-5112: Unit tests derived from PopUnitTestBase fail in IDE due to config errors

paul-rogers:
Tests rely on command-line settings in the pom.xml file. Those settings
are not available when tests are run in Eclipse. Replicated required
settings into the base test class (as in BaseTestQuery).

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/pop/PopUnitTestBase.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/pop/PopUnitTestBase.java
@@ -18,9 +18,12 @@
 package org.apache.drill.exec.pop;
 
 import java.io.IOException;
+import java.util.Properties;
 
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/pop/PopUnitTestBase.java
```diff
@@ -42,7 +46,17 @@
 
   @BeforeClass
   public static void setup() {
-    CONFIG = DrillConfig.create();
+    Properties props = new Properties();
```
chunhui-shi: Will the config here override the command line's setup if the unit test is initiated from Intellij? What about other unit tests - do they have the same problem?

sudheeshkatkam:
+1

PR #682, opened 2016-12-07
DRILL-5113: Upgrade Maven RAT plugin to avoid annoying XML errors

paul-rogers:
Upgrade to eliminate XML compiler warnings that appear for each sub-project in the Drill build.

Also fixes a Maven warning about a duplicated version error for the dependency plugin (version is inherited from root pom.xml). And removes some trailing spaces.


chunhui-shi:
+1

sudheeshkatkam:
+1

PR #683, opened 2016-12-07
DRILL-5114: Rationalize use of Logback logging in unit tests

paul-rogers:
Cleans up the Drill test Logback config files to reduce unnecessary
logging in tests and avoid Logback complaints about missing or
duplicate log config files.

Does not affect production logging.


exec/memory/base/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
chunhui-shi: Why could not the test here use existing logback-test.xml?

logical/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!-- Licensed to the Apache Software Foundation (ASF) under one or more contributor
+  license agreements. See the NOTICE file distributed with this work for additional
```
chunhui-shi: The same question for this file. They look identical.

paul-rogers:
Will fix later.

PR #684, opened 2016-12-08
DRILL-5070: Code gen: create methods in fixed order to allow test verification	

paul-rogers:
A handy technique in testing is to compare generated code against a "golden" copy that defines the expected results. However, at present, Drill generates code using the method order returned by Class.getDeclaredMethods, but this method makes no guarantee about the order of the methods. The order varies from one run to the next. There is some evidence this link that order can vary even within a single run, though a quick test was unable to reproduce this case.

The fix is simple, in the SignatureHolder constructor, sort methods by name after retrieving them from the class. The sort ensures that method order is deterministic. Fortunately, the number of methods is small, so the sort step adds little cost.

### DIFF
--- a/common/src/test/java/org/apache/drill/test/FileMatcher.java
+++ b/common/src/test/java/org/apache/drill/test/FileMatcher.java
@@ -1,0 +1,352 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


jinfengni:
It's fine to have fixed order for the generated methods. I'm not sure if it's appropriate to enforce  "golden" copy in the new tests. The golden copy is just one implementation in the current code; people may change the implementation in the future, and it's likely these new testcases will fail. The person to make the change in the future will have to deal with the overhead.

The question is what benefit we get from enforcing such "golden" copy in the testcases.

paul-rogers:
Golden copies are needed only for tests that, today, do their "testing" by dumping generated code to stdout. Such tests do absolutely nothing (other than detect crashes) when run under Maven. The golden copy assures that such low-level tests are verified.

Other tests verify results by executing code. No need to save a golden copy of code for those.

Yes, when code gen changes, the golden copies have to be recaptured. However, how else can we verify that the new changes do what we expect if we don't actually examine them and save the desired final state? How else do we detect unexpected changes to generated code? (For example, if we generated a bunch of unneeded boilerplate, we'd not detect that in an execution test.)

This test is also going to be a way to verify DRILL-5052: capture code using the current code gen technique, then verify against the code path described in DRILL-5052. The two paths should be identical except for an extra "extends" in the code from DRILL-5052.

Is it worth testing such cases or is it OK to simply trust that the generated code is probably OK?

paul-rogers:
Closing pull request since we don't want to have golden copies of generated code.

PR #685, opened 2016-12-08
Drill 5043: Function that returns a unique id per session/connection similar to MySQL's CONNECTION_ID()

nagarajanchinnasamy:
Please check [DRILL-5043](https://issues.apache.org/jira/browse/DRILL-5043) for the details on changes made and other details.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java
@@ -27,19 +27,19 @@ import org.apache.drill.exec.ops.ContextInformation;
 
 import javax.inject.Inject;
 
-@SuppressWarnings("unused")
 public class ContextFunctions {
 
   /**
    * Implement "user", "session_user" or "system_user" function. Returns the username of the user connected to Drillbit.
    */
-  @FunctionTemplate(names = {"user", "session_user", "system_user"}, scope = FunctionTemplate.FunctionScope.SIMPLE)
+  @FunctionTemplate(names = {"user", "session_user", "system_user"}, scope = FunctionTemplate.FunctionScope.SIMPLE, isNiladic = true)
   public static class User implements DrillSimpleFunc {
     @Output VarCharHolder out;
     @Inject ContextInformation contextInfo;
     @Inject DrillBuf buffer;
     @Workspace int queryUserBytesLength;
 
+    @Override
     public void setup() {
       final byte[] queryUserNameBytes = contextInfo.getQueryUser().getBytes();
       buffer = buffer.reallocIfNeeded(queryUserNameBytes.length);
@@ -47,6 +47,7 @@ public class ContextFunctions {
       buffer.setBytes(0, queryUserNameBytes);
     }
 
+    @Override
     public void eval() {
       out.start = 0;
       out.end = queryUserBytesLength;
@@ -57,13 +58,14 @@ public class ContextFunctions {
   /**
    * Implement "current_schema" function. Returns the default schema in current session.
    */
-  @FunctionTemplate(name = "current_schema", scope = FunctionTemplate.FunctionScope.SIMPLE)
+  @FunctionTemplate(name = "current_schema", scope = FunctionTemplate.FunctionScope.SIMPLE, isNiladic = true)
   public static class CurrentSchema implements DrillSimpleFunc {
     @Output VarCharHolder out;
     @Inject ContextInformation contextInfo;
     @Inject DrillBuf buffer;
     @Workspace int currentSchemaBytesLength;
 
+    @Override
     public void setup() {
       final byte[] currentSchemaBytes = contextInfo.getCurrentDefaultSchema().getBytes();
       buffer = buffer.reallocIfNeeded(currentSchemaBytes.length);
@@ -71,10 +73,37 @@ public class ContextFunctions {
       buffer.setBytes(0, currentSchemaBytes);
     }
 
+    @Override
     public void eval() {
       out.start = 0;
       out.end = currentSchemaBytesLength;
       out.buffer = buffer;
+    }
+  }
+
+  /**
+   * Implement "session_id" function. Returns the unique id of the current session.
+   */
+  @FunctionTemplate(name = "session_id", scope = FunctionTemplate.FunctionScope.SIMPLE, isNiladic = true)
+  public static class SessionId implements DrillSimpleFunc {
+    @Output VarCharHolder out;
+    @Inject ContextInformation contextInfo;
+    @Inject DrillBuf buffer;
+    @Workspace int sessionIdBytesLength;
+
+    @Override
+    public void setup() {
+        final byte[] sessionIdBytes = contextInfo.getSessionId().getBytes();
+        buffer = buffer.reallocIfNeeded(sessionIdBytes.length);
+        sessionIdBytesLength = sessionIdBytes.length;
+        buffer.setBytes(0, sessionIdBytes);
+    }
+
+    @Override
+    public void eval() {
+        out.start = 0;
+        out.end = sessionIdBytesLength;
+        out.buffer = buffer;
     }
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/ContextInformation.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/ContextInformation.java
@@ -28,12 +28,14 @@ public class ContextInformation {
   private final String currentDefaultSchema;
   private final long queryStartTime;
   private final int rootFragmentTimeZone;
+  private final String sessionId;
 
   public ContextInformation(final UserCredentials userCredentials, final QueryContextInformation queryContextInfo) {
     this.queryUser = userCredentials.getUserName();
     this.currentDefaultSchema = queryContextInfo.getDefaultSchemaName();
     this.queryStartTime = queryContextInfo.getQueryStartTime();
     this.rootFragmentTimeZone = queryContextInfo.getTimeZone();
+    this.sessionId = queryContextInfo.getSessionId();
   }
 
   /**
@@ -62,5 +64,12 @@ public class ContextInformation {
    */
   public int getRootFragmentTimeZone() {
     return rootFragmentTimeZone;
+  }
+
+  /**
+   * @return Unique id of the user session
+   */
+  public String getSessionId() {
+    return sessionId;
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlOperator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlOperator.java
@@ -28,13 +28,17 @@ import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlOperatorBinding;
+import org.apache.calcite.sql.SqlSyntax;
 import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.calcite.sql.type.OperandTypes;
 import org.apache.calcite.sql.type.SqlReturnTypeInference;
+import org.apache.calcite.sql.validate.SqlMonotonicity;
 import org.apache.drill.exec.expr.fn.DrillFuncHolder;
 
 public class DrillSqlOperator extends SqlFunction {
   // static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DrillSqlOperator.class);
   private final boolean isDeterministic;
+  private final boolean isNiladic;
   private final List<DrillFuncHolder> functions;
 
   /**
@@ -44,11 +48,12 @@ public class DrillSqlOperator extends SqlFunction {
    * In principle, if Drill needs a DrillSqlOperator, it is supposed to go to DrillOperatorTable for pickup.
    */
   @Deprecated
-  public DrillSqlOperator(final String name, final int argCount, final boolean isDeterministic) {
+  public DrillSqlOperator(final String name, final int argCount, final boolean isDeterministic, final boolean isNiladic) {
     this(name,
         argCount,
         isDeterministic,
-        DynamicReturnType.INSTANCE);
+        DynamicReturnType.INSTANCE,
+        isNiladic);
   }
 
   /**
@@ -59,13 +64,14 @@ public class DrillSqlOperator extends SqlFunction {
    */
   @Deprecated
   public DrillSqlOperator(final String name, final int argCount, final boolean isDeterministic,
-      final SqlReturnTypeInference sqlReturnTypeInference) {
+      final SqlReturnTypeInference sqlReturnTypeInference, final boolean isNiladic) {
     this(name,
         new ArrayList<DrillFuncHolder>(),
         argCount,
         argCount,
         isDeterministic,
-        sqlReturnTypeInference);
+        sqlReturnTypeInference,
+        isNiladic);
   }
 
   /**
@@ -75,7 +81,7 @@ public class DrillSqlOperator extends SqlFunction {
    * In principle, if Drill needs a DrillSqlOperator, it is supposed to go to DrillOperatorTable for pickup.
    */
   @Deprecated
-  public DrillSqlOperator(final String name, final int argCount, final boolean isDeterministic, final RelDataType type) {
+  public DrillSqlOperator(final String name, final int argCount, final boolean isDeterministic, final RelDataType type, final boolean isNiladic) {
     this(name,
         new ArrayList<DrillFuncHolder>(),
         argCount,
@@ -85,11 +91,11 @@ public class DrillSqlOperator extends SqlFunction {
           public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
             return type;
           }
-        });
+        }, isNiladic);
   }
 
   protected DrillSqlOperator(String name, List<DrillFuncHolder> functions, int argCountMin, int argCountMax, boolean isDeterministic,
-      SqlReturnTypeInference sqlReturnTypeInference) {
+      SqlReturnTypeInference sqlReturnTypeInference, boolean isNiladic) {
     super(new SqlIdentifier(name, SqlParserPos.ZERO),
         sqlReturnTypeInference,
         null,
@@ -98,6 +104,7 @@ public class DrillSqlOperator extends SqlFunction {
         SqlFunctionCategory.USER_DEFINED_FUNCTION);
     this.functions = functions;
     this.isDeterministic = isDeterministic;
+    this.isNiladic = isNiladic;
   }
 
   @Override
@@ -105,8 +112,19 @@ public class DrillSqlOperator extends SqlFunction {
     return isDeterministic;
   }
 
+  public boolean isNiladic() {
+    return isNiladic;
+  }
+
   public List<DrillFuncHolder> getFunctions() {
     return functions;
+  }
+
+  public SqlSyntax getSyntax() {
+    if(isNiladic) {
+      return SqlSyntax.FUNCTION_ID;
+    }
+    return super.getSyntax();
   }
 
   public static class DrillSqlOperatorBuilder {
@@ -115,6 +133,7 @@ public class DrillSqlOperator extends SqlFunction {
     private int argCountMin = Integer.MAX_VALUE;
     private int argCountMax = Integer.MIN_VALUE;
     private boolean isDeterministic = true;
+    private boolean isNiladic = false;
 
     public DrillSqlOperatorBuilder setName(final String name) {
       this.name = name;
@@ -147,6 +166,25 @@ public class DrillSqlOperator extends SqlFunction {
       return this;
     }
 
+    public DrillSqlOperatorBuilder setNiladic(boolean isNiladic) {
+      /*
+       * Set Operand type-checking strategy for an operator which takes no operands and need to be invoked
+       * without parentheses. E.g.: session_id
+       *
+       * Niladic functions override columns that have names same as any niladic function. Such columns cannot be
+       * queried without the table qualification. Value of the niladic function is returned when table
+       * qualification is not used.
+       *
+       * For e.g. in the case of session_id:
+       *
+       * select session_id from <table> -> returns the value of niladic function session_id
+       * select t1.session_id from <table> t1 -> returns session_id column value from <table>
+       *
+       */
+      this.isNiladic = isNiladic;
+      return this;
+    }
+
     public DrillSqlOperator build() {
       if(name == null || functions.isEmpty()) {
         throw new AssertionError("The fields, name and functions, need to be set before build DrillSqlAggOperator");
@@ -160,7 +198,8 @@ public class DrillSqlOperator extends SqlFunction {
           isDeterministic,
           TypeInferenceUtils.getDrillSqlReturnTypeInference(
               name,
-              functions));
+              functions),
+          isNiladic);
     }
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/util/Utilities.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/util/Utilities.java
@@ -49,13 +49,14 @@ public class Utilities {
    * @param defaultSchemaName
    * @return
    */
-  public static QueryContextInformation createQueryContextInfo(final String defaultSchemaName) {
+  public static QueryContextInformation createQueryContextInfo(final String defaultSchemaName, final String sessionId) {
     final long queryStartTime = System.currentTimeMillis();
     final int timeZone = DateUtility.getIndex(System.getProperty("user.timezone"));
     return QueryContextInformation.newBuilder()
         .setDefaultSchemaName(defaultSchemaName)
         .setQueryStartTime(queryStartTime)
         .setTimeZone(timeZone)
+        .setSessionId(sessionId)
         .build();
   }
 
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/client/DrillClientSystemTest.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/client/DrillClientSystemTest.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.client;
 
+import static org.junit.Assert.assertFalse;
 import java.util.List;
 
 import org.apache.drill.exec.DrillSystemTestBase;
@@ -72,5 +73,27 @@ public class DrillClientSystemTest extends DrillSystemTestBase {
       result.release();
     }
     client.close();
+  }
+
+  @Test
+  public void testSessionIdUDFWithTwoConnections() throws Exception {
+    final String sessionIdQuery = "select session_id as sessionId from (values(1));";
+    startCluster(1);
+
+    DrillClient client1 = new DrillClient();
+    client1.connect();
+    List<QueryDataBatch> results1 = client1.runQuery(QueryType.SQL, sessionIdQuery);
+    String sessionId1 = results1.get(0).getData().toString();
+    results1.get(0).release();
+    client1.close();
+
+    DrillClient client2 = new DrillClient();
+    client2.connect();
+    List<QueryDataBatch> results2 = client1.runQuery(QueryType.SQL, sessionIdQuery);
+    String sessionId2 = results2.get(0).getData().toString();
+    results2.get(0).release();
+    client2.close();
+
+    assertFalse(sessionId1.equals(sessionId2));
   }
 }
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestLocalExchange.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestLocalExchange.java
@@ -407,7 +407,7 @@ public class TestLocalExchange extends PlanTestBase {
 
     findFragmentsWithPartitionSender(rootFragment, planningSet, deMuxFragments, htrFragments);
 
-    final QueryContextInformation queryContextInfo = Utilities.createQueryContextInfo("dummySchemaName");
+    final QueryContextInformation queryContextInfo = Utilities.createQueryContextInfo("dummySchemaName", "938ea2d9-7cb9-4baf-9414-a5a0b7777e8e");
     QueryWorkUnit qwu = PARALLELIZER.getFragments(new OptionList(), drillbitContext.getEndpoint(),
         QueryId.getDefaultInstance(),
         drillbitContext.getBits(), planReader, rootFragment, USER_SESSION, queryContextInfo);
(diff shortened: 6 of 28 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
```diff
@@ -115,6 +118,7 @@ public UserSession build() {
 
   private UserSession() {
     queryCount = new AtomicInteger(0);
+    sessionId = nextSessionId.getAndIncrement();
   }
```
arina-ielchiieva: As far as I know, UserSession instance is created for each session, won't this mean that after your changes each session will have the same id? Actually CTTAS are going to add session id in its scope (it's represented as String and generated using UUID). You make take a look at those changes - https://github.com/apache/drill/pull/666

arina-ielchiieva: Sorry, I see it's static. But using this approach session id on different drillbits may overlap.

paul-rogers:
As it turns out, Arina is building something very similar for the CTTAS feature. See the PR: https://github.com/apache/drill/pull/666

Can the two of your coordinate so that the same ID can be used for both purposes? Can the function described here expose the ID that CTTAS has created? Is the GUID format for CTTAS adequate for the need here?

nagarajanchinnasamy:
@paul-rogers as per my requirement the SessionId needs to be unique. Thats all. It does not have to be a monotonically increasing sequence. So, UUID should be fine. Will look into #666

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
nagarajanchinnasamy: Thanks Arina. Just realized you are the author of #666 :) Could you please tell me how I can merge your changes into my branch. Am new to this process. Thanks.

nagarajanchinnasamy:
@paul-rogers @arina-ielchiieva I have incorporated only the session id related changes from #666. Checked in embedded mode and it does work. Yet to check in distributed mode. Is there a specific testcase that i can run to test this?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java
```diff
@@ -77,4 +77,28 @@ public void eval() {
...
+  public static class SessionId implements DrillSimpleFunc {
+    @Output VarCharHolder out;
+    @Inject ContextInformation contextInfo;
+    @Inject DrillBuf buffer;
+    @Workspace int sessionIdBytesLength;
+
+    public void setup() {
+        final byte[] sessionIdBytes = contextInfo.getSessionId().getBytes();
+        buffer = buffer.reallocIfNeeded(sessionIdBytes.length);
+        sessionIdBytesLength= sessionIdBytes.length;
```
arina-ielchiieva: space before =

exec/java-exec/src/main/java/org/apache/drill/exec/util/Utilities.java
```diff
@@ -43,19 +43,21 @@ public static String getFileNameForQueryFragment(FragmentContext context, String
...
 
   /**
-   * Create QueryContextInformation with given <i>defaultSchemaName</i>. Rest of the members of the
+   * Create QueryContextInformation with given <i>defaultSchemaName</i> and <i>sessionId</i>. Rest of the members of the
    * QueryContextInformation is derived from the current state of the process.
    *
    * @param defaultSchemaName
+   * @param sessionId
    * @return
    */
```
arina-ielchiieva: Please add description for two @param and return, since you are changing method description.

arina-ielchiieva:
@nagarajanchinnasamy 
1. Please address minor changes in review comments.
2. Squash / fixup everything in one commit.
3. Update commit message: Jira-number: Jira name (ex: DRILL-5043: Function that returns a unique id per session/connection similar to MySQL's CONNECTION_ID()).
4. Add unit test in TestContextFunctions, let's say compare that when you two times call session_id() in different queries and in the same query you'll get the same result.

nagarajanchinnasamy:
@arina-ielchiieva 

  - minor changes made
  - commit message amended
  - unit tests added

But am not sure If I have done the squashing correctly!!! tried `rebase` etc.. Need help regarding this.

nagarajanchinnasamy:
@arina-ielchiieva 

  - squashing done :)

arina-ielchiieva:
Looks good.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java
```diff
@@ -77,4 +76,28 @@ public void eval() {
...
+   * Implement "session_id" function. Returns the unique id of the current session.
+   */
+  @FunctionTemplate(name = "session_id", scope = FunctionTemplate.FunctionScope.SIMPLE)
+  public static class SessionId implements DrillSimpleFunc {
+    @Output VarCharHolder out;
+    @Inject ContextInformation contextInfo;
+    @Inject DrillBuf buffer;
+    @Workspace int sessionIdBytesLength;
+
+    public void setup() {
```
sudheeshkatkam: Annotate with `@Override`, here and below.

sudheeshkatkam:
+1, pending minor change

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java
nagarajanchinnasamy: Changes pushed.

nagarajanchinnasamy:
@sudheeshkatkam changes suggested are incorporated and committed.

sudheeshkatkam:
@nagarajanchinnasamy Thank you for the patch!

This will be committed today.

julianhyde:
To be consistent with other the other context functions in standard SQL ({{CURRENT_USER}}, {{SESSION_USER}}, {{CURRENT_PATH}}, {{CURRENT_ROLE}}, {{LOCALTIMESTAMP}}) this should be invoked without parameters: {{session_id}} rather than {{session_id()}}.

nagarajanchinnasamy:
@arina-ielchiieva @sudheeshkatkam I have made further re-factoring. Please review the changes.

Main changes are:
1. Introduced *`UserSessionContext`* class to encapsulate all session-context related information
2. Changed UserSession.Builder to accept UserSessionContext instance (withSessionContext())
3. UserServer creates the UserSessionContext and passes it to UserSession.Builder

Thanks.

nagarajanchinnasamy:
@julianhyde Thanks for your feedback. But I am not sure how to make a function not to have `()`. Need help.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/ContextInformation.java
```diff
@@ -19,6 +19,8 @@
 
 import org.apache.drill.exec.proto.BitControl.QueryContextInformation;
 import org.apache.drill.exec.proto.UserBitShared.UserCredentials;
+import org.apache.drill.exec.proto.UserBitShared.UserSessionContextInformation;
```
nagarajanchinnasamy: Unused imports need to be removed.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestLocalExchange.java
```diff
@@ -55,6 +55,7 @@
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.List;
+import java.util.UUID;
```
nagarajanchinnasamy: Unused import needs to be removed

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/partitionsender/TestPartitionSender.java
```diff
@@ -27,6 +27,7 @@
 import java.io.PrintWriter;
 import java.util.List;
 import java.util.Random;
+import java.util.UUID;
```
nagarajanchinnasamy: Unused import needs to be removed

exec/java-exec/src/test/java/org/apache/drill/exec/pop/TestFragmentChecker.java
```diff
@@ -18,6 +18,7 @@
 package org.apache.drill.exec.pop;
 
 import java.util.List;
+import java.util.UUID;
```
nagarajanchinnasamy: Unused import needs to be removed

arina-ielchiieva:
-1 to changes after refactoring.
using withSessionContext() is optional, so somebody can create a session without it's unique id (for example, for CTTAS it's not acceptable). I suggest we revert to previous design where session id is created in constructor, it's safe and transparent. And I am not sure that we need the whole new class `UserSessionContext` to hold session id. 

@nagarajanchinnasamy, next time you refactor your changes after PR was reviewed, I guess it's better to add these changes in separate commit, so they can be reviewed easier. You'll be asked to squash your commits when review is done.

nagarajanchinnasamy:
@arina-ielchiieva Sorry for the refactoring before consulting some of the thoughts I had. Thanks for the suggestion to add such changes in separate commit. Learning the process :)

1. Idea of having UserSessionContext as a separate class was to have the ability to hold additional session related parameters as a possible future extension. Basically have the ability to:
  - populate system captured session parameters
  - have custom authenticator populated session parameters

2. I agree to the point that session-id cannot be optional.

I will see how I can combine above points and submit another commit. Hope that's fine.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java
```diff
@@ -64,17 +65,45 @@ public void eval() {
...
+    }
+
+    @Override
+    public void eval() {
+        out.start = 0;
+        out.end = sessionIdBytesLength;
+        out.buffer = buffer;
+    }
+  }
 }
```
nagarajanchinnasamy: Invocation of  `session_id()` without parenthesis is pending

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -343,10 +346,11 @@ public BitToUserHandshake getHandshakeResponse(UserToBitHandshake inbound) throw
           if (inbound.getRpcVersion() != UserRpcConfig.RPC_VERSION) {
             final String errMsg = String.format("Invalid rpc version. Expected %d, actual %d.",
                 UserRpcConfig.RPC_VERSION, inbound.getRpcVersion());
-
```
nagarajanchinnasamy: Remove debug message

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -343,10 +346,11 @@ public BitToUserHandshake getHandshakeResponse(UserToBitHandshake inbound) throw
...
+            logger.debug(errMsg);
```
nagarajanchinnasamy: Remove debug message

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -359,12 +363,14 @@ public BitToUserHandshake getHandshakeResponse(UserToBitHandshake inbound) throw
                 }
               }
               authenticator.authenticate(inbound.getCredentials().getUserName(), password);
+              // TODO: make authenticator return a map of session parameters and store it in sessionParams.
             } catch (UserAuthenticationException ex) {
+              logger.debug(ex.getMessage());
```
nagarajanchinnasamy: remove debug message

nagarajanchinnasamy:
@arina-ielchiieva I have made the changes to:

1. Generate sessionId within the constructor of UserSession
2. Add a builder method in UserSession to construct UserSessionContextInformation
3. Session Parameters related changes are either commented or put in TODO (for future extension. Will address this in a separate PR)
4. @julianhyde 's comment of being able to invoke session_id() UDF without parenthesis is still pending. Have sent a mail to dev list. Need to know how to achieve that in drill. 

This time again I have squashed the commits.... as I already have done that. Will take care of your suggestion about squashing in future PRs.

Please review the changes.

protocol/src/main/protobuf/UserBitShared.proto
```diff
@@ -320,3 +320,17 @@ message Jar {
   repeated string function_signature = 2;
 }
 
+
+/* Session parameter
+message SessionParam {
+  optional string key = 1;
+  optional string value = 2;
+}
```
arina-ielchiieva: We have similar Property in User.proto, we may use it instead of creating new.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/ContextInformation.java
```diff
@@ -28,12 +29,14 @@
...
   private final int rootFragmentTimeZone;
+  private final String sessionId;
 
-  public ContextInformation(final UserCredentials userCredentials, final QueryContextInformation queryContextInfo) {
+  public ContextInformation(final UserCredentials userCredentials, final UserSessionContextInformation sessionContextInfo, final QueryContextInformation queryContextInfo) {
     this.queryUser = userCredentials.getUserName();
     this.currentDefaultSchema = queryContextInfo.getDefaultSchemaName();
     this.queryStartTime = queryContextInfo.getQueryStartTime();
     this.rootFragmentTimeZone = queryContextInfo.getTimeZone();
+    this.sessionId = sessionContextInfo.getSessionId();
```
arina-ielchiieva: If user didn't set sessionContextInfo in builder, we'll get NPE?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
```diff
@@ -160,6 +173,14 @@ public int getQueryCount() {
     return queryCount.get();
   }
 
+  public String getSessionId() {
+    return sessionId;
+  }
+
+  public UserSessionContextInformation getSessionContextInfo() {
+    return sessionContextInfo;
+  }
```
arina-ielchiieva: As above, will we return null if user didn't set session context info in builder?

protocol/src/main/protobuf/UserBitShared.proto
arina-ielchiieva: We have similar Property in User.proto, not sure we need to create new message with the same content.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -255,11 +257,12 @@ void disableReadTimeout() {
       getChannel().pipeline().remove(BasicServer.TIMEOUT_HANDLER);
     }
 
-    void setUser(final UserToBitHandshake inbound) throws IOException {
+    void setUser(final UserToBitHandshake inbound, Map<String, String> sessionParams) throws IOException {
```
arina-ielchiieva: sessionParams -> sessionParameters

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
```diff
@@ -97,6 +101,14 @@ public Builder withUserProperties(UserProperties properties) {
       return this;
     }
 
+    public Builder setSessionContextInfo(Map<String, String> sessionParams) {
+        UserSessionContextInformation.Builder sessionContextBuilder = UserSessionContextInformation.newBuilder()
+            .setSessionId(userSession.sessionId);
+        // TODO: Populate sessionParams into UserSesessionContextInfo
+        userSession.sessionContextInfo =  sessionContextBuilder.build();
+        return this;
+    }
```
arina-ielchiieva: I guess UserSessionContextInformation should exist disregarding calling setSessionContextInfo. setSessionContextInfo needs to enrich it.

protocol/src/main/protobuf/UserBitShared.proto
```diff
@@ -320,3 +320,17 @@ message Jar {
...
+*/
+
+/* User session context that has session id and session parameters
+*/
+message UserSessionContextInformation {
+  optional string session_id = 1;
+  // repeated SessionParam session_params = 2
+}
```
arina-ielchiieva: Why we are using protobuf for UserSessionContextInformation and not regular class?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -255,11 +257,12 @@ void disableReadTimeout() {
...
       session = UserSession.Builder.newBuilder()
           .withCredentials(inbound.getCredentials())
           .withOptionManager(worker.getSystemOptions())
           .withUserProperties(inbound.getProperties())
+          .setSessionContextInfo(sessionParams)
```
arina-ielchiieva: withSessionParameters(sessionParameters)?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
arina-ielchiieva: Can you give example of sessionParameters? How they are different from UserProperties? Can we combine them?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -343,10 +346,10 @@ public BitToUserHandshake getHandshakeResponse(UserToBitHandshake inbound) throw
           if (inbound.getRpcVersion() != UserRpcConfig.RPC_VERSION) {
             final String errMsg = String.format("Invalid rpc version. Expected %d, actual %d.",
                 UserRpcConfig.RPC_VERSION, inbound.getRpcVersion());
-
             return handleFailure(respBuilder, HandshakeStatus.RPC_VERSION_MISMATCH, errMsg, null);
           }
 
+          Map<String, String> sessionParams = new LinkedHashMap<String, String>();
```
arina-ielchiieva: May be it's better to keep setUser method as is, I don't really like the idea of creating empty map each time we getHandshakeResponce

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
nagarajanchinnasamy: Lets say there is an external custom authenticator.... It validates the credentials with that external system and loads contextual information from that external system. Examples:
  - Tenant_Id of the currently logged in user
  - Roles of the user logged in
  - Any user preference details that are loaded in the context of that external system
  - Etc..

User Properties, as I understood seem to be the configuration variables (E.g., drill-override.conf). These are known expected properties. Whereas Session Parameters may not be known properties.

protocol/src/main/protobuf/UserBitShared.proto
nagarajanchinnasamy: Though they have same key-value pairs, did not want to overload it for different purposes.

nagarajanchinnasamy: Though they have same key-value pairs, did not want to overload it for different purposes.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
sudheeshkatkam: Can you please open a ticket with a details about use case? How will these session parameters be used?

nagarajanchinnasamy: @sudheeshkatkam I had created this few days back. But updated details with more clarity now. The ticket is: [DRILL-5132](https://issues.apache.org/jira/browse/DRILL-5132). Pls let me know your views.

nagarajanchinnasamy:
@arina-ielchiieva @sudheeshkatkam as [DRILL-5132](https://issues.apache.org/jira/browse/DRILL-5132) is being explored to be addressed through UDFs... I am reverting back to original design as suggested by @arina-ielchiieva .

Will submit a new commit with original design shortly. Sorry about the detour.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -343,7 +343,6 @@ public BitToUserHandshake getHandshakeResponse(UserToBitHandshake inbound) throw
           if (inbound.getRpcVersion() != UserRpcConfig.RPC_VERSION) {
             final String errMsg = String.format("Invalid rpc version. Expected %d, actual %d.",
                 UserRpcConfig.RPC_VERSION, inbound.getRpcVersion());
-
```
nagarajanchinnasamy: Remove newline introduced

protocol/src/main/protobuf/UserBitShared.proto
```diff
@@ -319,4 +319,3 @@ message Jar {
   optional string name = 1;
   repeated string function_signature = 2;
 }
-
```
nagarajanchinnasamy: Remove newline introduced

exec/java-exec/src/main/java/org/apache/drill/exec/util/Utilities.java
```diff
@@ -47,7 +47,7 @@ public static String getFileNameForQueryFragment(FragmentContext context, String
    * QueryContextInformation is derived from the current state of the process.
    *
    * @param defaultSchemaName
-   * @return
+   * @return QueryContextInformation
```
nagarajanchinnasamy: Revert the change as this has got nothing to do with this PR.

exec/java-exec/src/test/java/org/apache/drill/exec/client/DrillClientSystemTest.java
```diff
@@ -17,6 +17,8 @@
  */
 package org.apache.drill.exec.client;
 
+import static org.junit.Assert.assertFalse;
+
```
nagarajanchinnasamy: Remove unnecessary newline introduced

protocol/src/main/java/org/apache/drill/exec/proto/UserProtos.java
```diff
@@ -25001,7 +25001,7 @@ public Builder setUpdatability(org.apache.drill.exec.proto.UserProtos.ColumnUpda
        *
        * <pre>
        *
-       * Defaults to READ_ONLU
+       * Defaults to READ_ONLY
```
nagarajanchinnasamy: Revert changes as this has got nothing to do with this PR

protocol/src/main/protobuf/User.proto
```diff
@@ -361,7 +361,7 @@ message ResultColumnMetadata {
   optional ColumnSearchability searchability = 13;
 
   /*
-   * Defaults to READ_ONLY
+   * Defaults to READ_ONLU
```
nagarajanchinnasamy: Revert the change

nagarajanchinnasamy:
@arina-ielchiieva @sudheeshkatkam I have reverted the changes back to original design and tested. However, invoking `session_id` without parenthesis is pending to be fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -141,7 +141,7 @@ public FragmentContext(final DrillbitContext dbContext, final PlanFragment fragm
     this.accountingUserConnection = new AccountingUserConnection(connection, sendingAccountor, statusHandler);
     this.fragment = fragment;
     this.funcRegistry = funcRegistry;
-    contextInformation = new ContextInformation(fragment.getCredentials(), fragment.getContext());
+    contextInformation = new ContextInformation(fragment.getCredentials(), fragment.getContext(), connection.getSession().getSessionId());
```
nagarajanchinnasamy: accessing `connection` results into NPE for non-root fragment contexts.

arina-ielchiieva:
@nagarajanchinnasamy, as discussed on handouts yesterday, could you please examine how current_date, current_time are implemented? May be session_id can be implemented the same way?

nagarajanchinnasamy:
@arina-ielchiieva , I have started examining the difference in implementations of functions like CURRENT_DATE, SYSTEM_USER etc. in Calcite and the implementation of DrillSqlOperator in Drill. Will get back once I find the solution.

nagarajanchinnasamy:
Sent a mail to Dev list with the findings... issue yet to be resolved:

[Link to the mail](http://mail-archives.apache.org/mod_mbox/drill-dev/201701.mbox/%3CCACd9g6hwRtfCKD_2JHmrexT_TB0ZWcRXTb7J5digKmY9ATKdaA@mail.gmail.com%3E)

nagarajanchinnasamy:
Success :) I could invoke SESSION_ID without parenthesis. Following did the trick:

1. In DrillSqlOperator, overriding getSyntax method to return SqlSyntax.FUNCTION_ID 
2. In DrillOperatorTable, adding operator when SqlSyntax is (FUNCTION || FUNCTION_ID)

Will submit a commit shortly.

nagarajanchinnasamy:
@arina-ielchiieva @sudheeshkatkam I have made the following changes to support invocation of UDFs without parenthesis.

1. Introduced `isNiladic` option in FunctionTemplate and added a property in DrillFuncHolder, FunctionAttributes, DrillSqlOperator classes

2. Overridden DrillSqlOperator.getSyntax() to return SqlSyntax.FUNCTION_ID when isNiladic is true

3. Modified all the places to pass false for isNiladic parameter to DrillSqlOperator constructor -- so that existing behavior is not impacted

4. In DrillOperatorTable, adding an operator to operatorList when the syntax of an operator is either FUNCTION or FUNCTION_ID

5. Modified tests to invoke session_id without parenthesis. 

Please review the changes.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlOperator.java
```diff
@@ -147,6 +166,14 @@ public DrillSqlOperatorBuilder setDeterministic(boolean isDeterministic) {
       return this;
     }
 
+    public DrillSqlOperatorBuilder setNiladic(boolean isNiladic) {
+      /*
+       * Set Operand type-checking strategy for an operator which takes no operands
+       */
```
arina-ielchiieva: Also let's add note here that making function niladic won't allow to select column with function name without table alias. Please also add example:
When querying table t with column session_id:
select session_id from table -> will return generated session_id 
select t1.session_id from table t1 -> will return session_id column value

arina-ielchiieva:
@nagarajanchinnasamy Good job, thank you!
Please address minor review comments, squash commits into one and rebase on current master.

nagarajanchinnasamy:
@arina-ielchiieva @sudheeshkatkam @julianhyde @paul-rogers Thank You All for your comments and guidance throughout. It was a nice learning experience for me :)

I have squashed the commits. I understand this function needs to be added to the documentation. Please let me know what is the process to help with the documentation.

sudheeshkatkam:
+1

Thank you for the patch!

arina-ielchiieva:
@nagarajanchinnasamy after this weeks commits, there is one conflicting file. Could you please rebase on master and resolve conflict?

nagarajanchinnasamy:
@arina-ielchiieva conflicts resolved and rebased on master. Please check.

arina-ielchiieva:
@nagarajanchinnasamy, thank you!

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java
```diff
@@ -64,17 +65,45 @@ public void eval() {
...
       out.start = 0;
       out.end = currentSchemaBytesLength;
       out.buffer = buffer;
     }
   }
+
+  /**
+   * Implement "session_id" function. Returns the unique id of the current session.
+   */
+  @FunctionTemplate(name = "session_id", scope = FunctionTemplate.FunctionScope.SIMPLE, isNiladic = true)
```
nagarajanchinnasamy: - Functions like current_time/current_date already function like niladic as Calcite recognizes them so. 

- IsNiladic flag can be set by any UDF. Not restricted only to session_I'd

- If I understand right, currently there is no categorization of UDFs in Drill. If we want to restrict isNiladic flag only to a particular category of UDFs, then categorization of UDFs need to be designed n implemented.

arina-ielchiieva: Each drill function is converted to DrillFunctionHolder and though Calcite recognizes current_time/current_date & other context functions (ex: user, current_schema) as niladic, after introduction of isNilladic flag, DrillFunctionHolder would describe such functions as non-niladic, though they are. So we are intentionally hold incorrect info.

Let's say in future I would need to take list of all function holders in Drill (form local function registry) and filter out all niladic functions, since we do have such isNiladic flag I would assume that I can use this flag as filter but since current_time/current_date & other context functions (ex: user, current_schema) have isniladic=false, they would never be filtered out.

Also please add `isNiladic=true` to other niladic context functions [1].

[1] https://github.com/apache/drill/blob/5a4ad2a88331dfe7561ee76fc87e882afe170681/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java

nagarajanchinnasamy: @jinfengni @arina-ielchiieva am in the process of adding `isNiladic=true` to the standard functions. I am referring to [SqlStdOperatorTable.java](https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/sql/fun/SqlStdOperatorTable.java) in Calcite and tried to find the equivalent functions in Drill.

I have marked following functions from [DateTypeFunctions.java](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/DateTypeFunctions.java) and [ContextFunctions.java](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/ContextFunctions.java) as isNiladic=true.

1. "current_date"
2. "timeofday" (**is this a niladic function? Not found in Calcite**)
3. {"localtimestamp", "current_timestamp", "now", "statement_timestamp", "transaction_timestamp"}
4. {"current_time", "localtime"}
5. "unix_timestamp" (**is this a niladic function? Not found in Calcite**)
6. {"user", "session_user", "system_user"}
7. "current_schema"
8. "session_id"

Please check the above list... and let me know if I should add more. Also, I should not be marking something niladic if it is not as that will break the existing queries.

arina-ielchiieva: Please exclude `timeofday, now, statement_timestamp, transaction_timestamp, unix_timestamp`. Just leave those that already behave like niladic functions (just execute function in Drill and see if it requires parenthesis or not, if not, then it's niladic :)). We don't won't to change existing behavior not to break queries.

nagarajanchinnasamy: :) I checked the niladic functions... Will make the commit shortly.

arina-ielchiieva:
@nagarajanchinnasamy new changes look good, please squash commits into one.

nagarajanchinnasamy:
@arina-ielchiieva  commits squashed. Thanks.

arina-ielchiieva:
@nagarajanchinnasamy thank you, looks good.

sudheeshkatkam:
Another +1 :)

arina-ielchiieva:
Changes have been merged into master. @nagarajanchinnasamy, please close pull request.

nagarajanchinnasamy:
Closing pull request

PR #686, opened 2016-12-09
DRILL-5117: Compile error when query a json file with 1000+columns

Serhii-Harnyk:


exec/java-exec/src/main/java/org/apache/drill/exec/expr/SizedJBlock.java
```diff
@@ -32,7 +32,11 @@
 
   public SizedJBlock(JBlock block) {
     this.block = block;
-    this.count = 0;
+    // count should be 1 because in some cases it is hard to increase it when
+    // Logical Expressions are added to JBlock.
+    // To avoid taking into account of some extra count from empty JBlock,
```
jinfengni: I spent some time to understand why modifying this count makes the query compilation works.  

The failing case encounter compilation failure in Copier (for SVR operator). Turns out that the idea of SizedJBlock (DRILL-4715) only works when we call ClassGenerator.addExpr(). This is fine with Project, Filter, Aggregator, etc, but not for Copier.  The Copier is doing kind of short-cut handling, by accessing the eval() and setup() directly [1].

Ideally, we probably should try to see if we can convert Copier using same mechanism in Project/Filter. After some thoughts, I realized doing that might add additional overhead, as the current way is doing the copying directly. 

Given that, I'm fine with this proposed change. Please add comments to explain why we set count = 1. 

[1] https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/vector/CopyUtil.java#L45-L50

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestLargeFileCompilation.java
```diff
@@ -154,4 +158,20 @@ public void testProject() throws Exception {
     testNoResult(ITERATION_COUNT, LARGE_QUERY_SELECT_LIST);
   }
 
+  @Test
+  public void testSelectAllFromFileWithManyColumns() throws Exception {
+    File path = new File(BaseTestQuery.getTempDir("json/input"));
```
jinfengni: You do not have to generate a new json file for writing a unit test. Not sure how the new created file is cleaned up if the testcase failed.

You may consider using the existing tpch sample file to do unit test, or re-enable the originally ignored test (testTOP_N_SORT).

```java
  @Test
  public void testLargeListColWithLimit() throws Exception {
    final int nCol = 1000;
    final StringBuilder sb = new StringBuilder();

    sb.append(" select n_nationkey ");
    for (int i = 0; i< nCol; i++) {
      sb.append(", " + "Col" + i );
    }
    sb.append(" from cp.`tpch/nation.parquet`");
    sb.append(" limit 1");

    test(sb.toString());
  }
```

Serhii-Harnyk:
@jinfengni, could you please review new changes? I squashed all changes into single commit and rebased into master.

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestLargeFileCompilation.java
Serhii-Harnyk: You are right, with this changes both tests testEXTERNAL_SORT and testTOP_N_SORT work. So I enabled them.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/SizedJBlock.java
Serhii-Harnyk: I changed previous comment and  added your explanations.

jinfengni:
+1

LGTM. Thanks for the PR.

sudheeshkatkam:
@Serhii-Harnyk I am curious how this resolves DRILL-1808. Can you provide some details?

Serhii-Harnyk:
@sudheeshkatkam, when running for example test testEXTERNAL_SORT(), generates class CopierGen4, in which methods doSetup() and doEval() does not splits correctly. But with this fix them both splits into the smaller methods.

jinfengni:
I think the cause of DRILL-1808 is same as DRILL-5117. We should mark them as related or duplicated in the JIRA.

PR #690, opened 2016-12-10
DRILL-5121 A memory leak is observed when exact case is not specified for a column in a filter condition

bitblender:
Fix for https://issues.apache.org/jira/browse/DRILL-5121.

Changes fieldVectorMap in ScanBatch to a CaseInsensitiveMap to workaround the memory leak described in the bug.


jinfengni:
The proposed change looks fine for case insensitive store. On the other hand, have you given some thoughts of @Ben-Zvi 's comment about case sensitive store? Will it work for case sensitive store?

You may add one simple testcase to verify that the fix works as expected.

PR #692, opened 2016-12-12
DRILL-5123: Write query profile after sending final response to client

paul-rogers:
Write query profile after sending final response to client to improve latency

In testing a particular query, I used a test setup that does not write
to the "persistent store", causing query profiles to not be saved. I
then changed the config to save them (to local disk). This produced
about a 200ms difference in query run time as perceived by the client.

I then moved writing the query profile after sending the client the
final message. This resulted in an approximately 100ms savings, as
perceived by the client, in query run time on short (~3 sec.) queries.


paul-rogers:
Fixed comment to explain reasoning without reference to prior behavior.

sudheeshkatkam:
+1

ppadma:
+1. LGTM

PR #693, opened 2016-12-13
DRILL-5122: DrillBuf performs expensive logging if -ea set

paul-rogers:
In Drill, once assertions are enabled, the dominant cost of a query
becomes DrillBuf logging. One example showed that DrillBuf logging
increases query cost by 10x.

Logging was enabled by the -ea option, which should be available for
many other uses. Changed to require a specific option to be turned on.

Enabled the option for memory tests (which require the option) but off
by default and for all other tests.

To do this, converted Surefire system options from command line args to
system variables so that they can be customized in sub-project poms.


paul-rogers:
Will fix later.

PR #694, opened 2016-12-13
DRILL-5104: Foreman should not alter physical plan memory

paul-rogers:
Foreman should not set external sort memory for a physical plan

Physical plans include a plan for memory allocations. However, the code
path in Foreman replans external sort memory, even for a physical plan.
This makes it impossible to use a physical plan to test memory
configuration.

This change avoids changing memory settings in a physical plan; while
preserving the adjustments for logical plans or SQL queries.


paul-rogers:
Will complete later.

PR #696, opened 2016-12-14
DRILL-4558: BSonReader should prepare buffer size as actual need

chunhui-shi:


exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -103,6 +103,17 @@ public void testStringType() throws IOException {
   }
 
   @Test
+  public void testSpecialCharStringType() throws IOException {
+    BsonDocument bsonDoc = new BsonDocument();
+    bsonDoc.append("stringKey", new BsonString("§§§§§§§§§1"));
```
paul-rogers: Might want to put your value in a string, convert to byte but, and assert that the byte buf length differs in length from the string length. This will verify that this test is, in fact, testing the problem case.

parthchandra:
+1

PR #697, opened 2016-12-15
DRILL-5097: Using store.parquet.reader.int96_as_timestamp gives IOOB whereas convert_from works

vdiravka:
When the int96 value is converted into timestamp (long in java) we cut the nanos precision to millis. But need to change dataTypeLengthInBits of columnReader (while converting parquet fixed binary type INT96 into drill TimeStamp) from 12 byte(s) to 8 byte(s) as well.


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/NullableFixedByteAlignedReaders.java
```diff
@@ -110,9 +110,14 @@ protected void readField(long recordsToReadInThisPass) {
...
   /**
    * Class for reading parquet fixed binary type INT96, which is used for storing hive,
-   * impala timestamp values with nanoseconds precision. So it reads such values as a drill timestamp.
+   * impala timestamp values with nanoseconds precision (12 bytes). So it reads such values as a drill timestamp (8 bytes).
    */
   static class NullableFixedBinaryAsTimeStampReader extends NullableFixedByteAlignedReader<NullableTimeStampVector> {
+    /**
+     * The width of each element of the TimeStampVector is 8 byte(s).
+     */
+    private static final int timestampLengthInBits = 64;
```
bitblender: I would recommend the use of uppercase in naming constants.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/NullableFixedByteAlignedReaders.java
```diff
@@ -132,6 +137,9 @@ protected void readField(long recordsToReadInThisPass) {
           valueVec.getMutator().setSafe(valuesReadInCurrentPass + i, getDateTimeValueFromBinary(binaryTimeStampValue));
         }
       }
+      // The nanos precision is cut to millis. Therefore the length of single timestamp value is 8 bytes(s)
+      // instead of 12 byte(s).
+      dataTypeLengthInBits = timestampLengthInBits;
```
bitblender: Just trying to understand this better: Is it OK to set the length to 64 bits even if PARQUET_READER_INT96_AS_TIMESTAMP is set to false?

jinfengni:
+1

PR #698, opened 2016-12-19
DRILL-5136: Server unable to prepare non SELECT queries

superbstreak:
Currently, the server makes every incoming queries a limit 0 query during server prepare. This is causing some non select queries to fail.

We need to add an additional check to determine if the query is a SELECT query before wrapping it with LIMIT 0.


superbstreak:
Closing PR due to widen scope.

PR #699, opened 2016-12-21
DRILL-5137 - Optimize count(*) queries for MapR-DB Binary Tables

spanchamiamapr:
This diff uses the same optimization as that for the rowKeyOnly queries.
We use the FirstKeyOnlyFilter for count(*) queries.
This fix will optimize these queries for HBase tables, as well as MapR-DB Binary tables.


spanchamiamapr:
Closing this pull request since it is showing unnecessary changes too.

PR #700, opened 2016-12-21
DRILL-5137 - Optimize count(*) queries on MapR-DB Binary Tables

spanchamiamapr:
This diff uses the same optimization as that for the rowKeyOnly queries.
We use the FirstKeyOnlyFilter for count(*) queries.
This fix will optimize these queries for HBase tables, as well as MapR-DB Binary tables.

### DIFF
--- a/contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
+++ b/contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
@@ -124,6 +124,11 @@ public class HBaseRecordReader extends AbstractRecordReader implements DrillHBas
     } else {
       rowKeyOnly = false;
       transformed.add(ROW_KEY_PATH);
+      /* DRILL-5137 - optimize count(*) queries on MapR-DB Binary tables */
+      if (isSkipQuery()) {
+        hbaseScan.setFilter(
+            HBaseUtils.andFilterAtIndex(hbaseScan.getFilter(), HBaseUtils.LAST_FILTER, new FirstKeyOnlyFilter()));
+      }
     }
 
 


contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
```diff
@@ -124,6 +124,11 @@ public HBaseRecordReader(Connection connection, HBaseSubScan.HBaseSubScanSpec su
     } else {
       rowKeyOnly = false;
       transformed.add(ROW_KEY_PATH);
+      /* DRILL-5137 - optimize count(*) queries on MapR-DB Binary tables */
```
adityakishore: This branches into the else part of
`if (!isStarQuery()) {`

Can you verify if a query can be both Star and Skip query at the same time when count(*) has been requested?

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
```diff
@@ -124,6 +124,11 @@ public HBaseRecordReader(Connection connection, HBaseSubScan.HBaseSubScanSpec su
...
+      if (isSkipQuery()) {
```
adityakishore: Further optimization can be name by returning only a `count` vector in the `next()` call, similar to [this](https://github.com/apache/drill/blob/master/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java#L203-L204).

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
spanchamiamapr: Yes, I verified this in my testing.

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
spanchamiamapr: I thought of this, however I got stuck because there is no vectorContainerWriter in HBaseRecordReader class. If you can send a diff of how to do this, I can include that too. Thanks!

PR #702, opened 2016-12-21
DRILL-5088: set default codec for toJson

chunhui-shi:


contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoGroupScan.java
```diff
@@ -503,7 +505,9 @@ public ScanStats getScanStats() {
       long numDocs = collection.count();
       float approxDiskCost = 0;
       if (numDocs != 0) {
-        String json = collection.find().first().toJson();
+        final DocumentCodec codec =
```
gparai: Please add a comment to describe the need for the codec along with the example from the bug.

chunhui-shi:
MongoGroupScan tries to read first record to guess record width. So this toJson is called for only first record.

Notice that MongoClient.getDefaultCodecRegistry() contains DBRefCodecProvider:

Shown in MongoClient.java(In MongoClient.java):

    private static final CodecRegistry DEFAULT_CODEC_REGISTRY =
            fromProviders(asList(new ValueCodecProvider(),
                    new DBRefCodecProvider(),
                    new DocumentCodecProvider(new DocumentToDBRefTransformer()),
                    new DBObjectCodecProvider(),
                    new BsonValueCodecProvider(),
                    new GeoJsonCodecProvider()));

but a default DocumentCodec constructor does not. And it was this constructor get called if we don't set a codec for toJson function to use. The list of CodecProviders is shown in DocumentCodec.java in MongoDb's code:

public class DocumentCodec implements CollectibleCodec<Document> {

    private static final String ID_FIELD_NAME = "_id";
    private static final CodecRegistry DEFAULT_REGISTRY = fromProviders(asList(new ValueCodecProvider(),
            new BsonValueCodecProvider(),
            new DocumentCodecProvider()));

gparai:
+1 LGTM

gparai:
Actually, can we also add a testcase?

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
```diff
@@ -234,15 +240,17 @@ private static void createDbAndCollections(String dbName,
...
-        SingleMode.cleanup();
+      if (initCount.decrementAndGet() == 0) {
+        if (mongoClient != null) {
+          mongoClient.dropDatabase(EMPLOYEE_DB);
+          mongoClient.close();
+        }
+        if (distMode) {
+          DistributedMode.cleanup();
+        } else {
+          SingleMode.cleanup();
```
paul-rogers: Should the above include error handling? It seems that if the `dropDatabase` fails, then `close()` won't be called, nor will the `cleanup` method.

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/TestTableGenerator.java
```diff
@@ -58,7 +59,16 @@ public static void generateTable(String dbName, String collection,
...
         .getDefaultInstance().prepare(mongoImportConfig);
-    importExecutable.start();
+    MongoImportProcess importProcess = importExecutable.start();
+
+    try {
+      while (importProcess.isProcessRunning()) {
+        Thread.sleep(1000);
+      }
+    }catch (Exception ex) {
+      logger.error("Import mongoDb failed", ex);
```
paul-rogers: Here we log the error but go ahead and return. Should we propagate an exception upward in the call stack so that the caller knows that "Something Is Wrong"? Otherwise, how will the caller know whether the import process is ready or not?

chunhui-shi: Paul, since there are two JIRAs here, I filed a new pull request  https://github.com/apache/drill/pull/741 which addressed your comments. Could you look at that pull request? Thanks.

gparai:
Thanks for adding the testcase. 
+1 LGTM.

sudheeshkatkam:
+1

PR #703, opened 2016-12-22
DRILL-5104: Foreman should not set sort memory for a physical plan

paul-rogers:
Physical plans include a plan for memory allocations. However, the code
path in Foreman replans external sort memory, even for a physical plan.
This makes it impossible to use a physical plan to test memory
configuration.

This change avoids changing memory settings in a physical plan; while
preserving the adjustments for logical plans or SQL queries.

Revised to put a property in the plan itself. Old plans, and those
generated from SQL, will have memory allocations applied. Plans
marked as already "resource management" planned will be used as-is.

Includes a unit test that demonstrates the new behavior.

### DIFF
--- a/logical/src/main/java/org/apache/drill/common/logical/PlanProperties.java
+++ b/logical/src/main/java/org/apache/drill/common/logical/PlanProperties.java
@@ -35,6 +35,12 @@ public class PlanProperties {
   public JSONOptions options;
   public int queue;
 
+  /**
+   * Indicates if the plan has been planned for resource management
+   * (memory, etc.) or if this plan must still be computed.
+   */
+  public boolean hasResourcePlan;
+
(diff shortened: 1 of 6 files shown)


logical/src/main/java/org/apache/drill/common/logical/PlanProperties.java
```diff
@@ -112,8 +121,13 @@ public PlanPropertiesBuilder generator(Generator generator) {
       return this;
     }
 
+    public PlanPropertiesBuilder generator(boolean hasResourcePlan) {
```
Ben-Zvi: The method's name should not be **generator** but something about having a resource plan

paul-rogers: Fixed.

PR #704, opened 2016-12-22
DRILL-5125: Provide option to use generic code for sv remover

paul-rogers:
Performance tests showed that, for queries with a large number of
columns, it is faster to use a “generic” implementation of the
selection vector remover “copier” than to use a generated version.

This PR provides "generic" versions of the SV2 and SV4 copiers
used by the selection vector remover. The generic forms are
enabled using a new boot-time config parameter that is disabled
by default (preserving the traditional generated code.)

The generic form relies on a "virtual function" (really, just a
plain Java function) defined in the base ValueVector class and
implemented by each concrete vector: both the pre-defined and
generated forms. This form "does the right thing" for the copy
operation so that we don't need to generate code just to handle
the method dispatch operation (which Java does quite well on its
own.)

A unit tests validates that the generic form works by runing
the existing SV remover tests with the generic option turned on.

See the DRILL-5125 for details.

Add test

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/GenericSV2Copier.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/GenericSV2Copier.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.svremover;
+
+import org.apache.drill.exec.ops.FragmentContext;
+import org.apache.drill.exec.record.RecordBatch;
+import org.apache.drill.exec.record.VectorWrapper;
+import org.apache.drill.exec.vector.ValueVector;
+
+/**
+ * Generic selection vector 2 copier implementation that can
+ * be used in place of the generated version. Relies on a
+ * virtual function in each value vector to choose the proper
+ * implementation. Tests suggest that this version performs
+ * better than the generated version for queries with many columns.
+ */
+
+public class GenericSV2Copier extends CopierTemplate2 {
+
+  private ValueVector[] vvOut;
+  private ValueVector[] vvIn;
+
+  @SuppressWarnings("unused")
+  @Override
+  public void doSetup(FragmentContext context, RecordBatch incoming, RecordBatch outgoing) {
+
+    // Seems to be no way to get the vector count without iterating...
+
+    int count = 0;
+    for(VectorWrapper<?> vv : incoming) {
+      count++;
+    }
+    vvIn = new ValueVector[count];
+    vvOut = new ValueVector[count];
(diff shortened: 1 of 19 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/svremover/TestSVRemover.java
```diff
@@ -34,4 +38,33 @@ public void testSVRWithNoFilter() throws Exception {
...
+    config.put(ExecConstants.HTTP_ENABLE, "false");
+    config.put(ExecConstants.REMOVER_ENABLE_GENERIC_COPIER, "true");
+    updateTestCluster(1, DrillConfig.create(config));
+
+    int numOutputRecords = testPhysical(getFile("remover/test1.json"));
+    assertEquals(50, numOutputRecords);
+    numOutputRecords = testPhysical(getFile("remover/sv_with_no_filter.json"));
+    assertEquals(100, numOutputRecords);
+  }
 }
```
ppadma: Do these tests cover all the value vector types ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/GenericSV4Copier.java
```diff
@@ -0,0 +1,67 @@
...
+
+public class GenericSV4Copier extends CopierTemplate4 {
+
+  private ValueVector[] vvOut;
+  private ValueVector[][] vvIn;
+
+  @SuppressWarnings("unused")
+  @Override
+  public void doSetup(FragmentContext context, RecordBatch incoming,
+      RecordBatch outgoing) {
```
paul-rogers: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/GenericSV2Copier.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class GenericSV2Copier extends CopierTemplate2 {
+
+  private ValueVector[] vvOut;
+  private ValueVector[] vvIn;
+
+  @SuppressWarnings("unused")
+  @Override
+  public void doSetup(FragmentContext context, RecordBatch incoming,
+      RecordBatch outgoing) {
```
paul-rogers: Fixed.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/svremover/TestSVRemover.java
paul-rogers: No, which is why the feature is off by default. More work is needed to provide a reliable way to generate data for all (supported) value vector types.

ppadma: It would be good to have the tests that cover all the vector types.  But,  since it is off by default and you are exercising the code through other unit tests,  this is fine. 
+1

exec/vector/src/main/codegen/templates/FixedValueVectors.java
```diff
@@ -281,6 +282,11 @@ public void copyFromSafe(int fromIndex, int thisIndex, ${minor.class}Vector from
     copyFrom(fromIndex, thisIndex, from);
   }
 
+  @Override
+  public void copyEntry(int toIndex, ValueVector from, int fromIndex) {
+    ((${minor.class}Vector) from).data.getBytes(fromIndex * ${type.width}, data, toIndex * ${type.width}, ${type.width});
```
jinfengni: Throughout all the value vector classes, this new "copyEntry" method has to explicitly downcast from interface ValueVector to a subclass. Such cast happens for each value copied. I' not sure how performance overhead it could possibly incur : JVM has to check the input type, and raise ClassCastException in case of error. 

Compared with the existing method,  I think cast only happens in doSetup(), which is called for each new schema. 

Also, "copyEntry" uses data.getBytes(), while copyFromSafe() uses setInt(...,   from.data.getInt()).  I'm not sure the performance difference comes from the difference of generated code vs generic class, or the difference of the underneath implementation.

paul-rogers:
Let's do more research on this one. Closing for now since that research is not a priority at the moment.

PR #705, opened 2016-12-23
DRILL-5159: Make sure Drill's ProjectMergeRule operate on RelNodes wi…

jinfengni:
…th same convention trait.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
@@ -18,6 +18,8 @@
 package org.apache.drill.exec.planner.logical;
 
 
+import org.apache.calcite.plan.Convention;
+import org.apache.calcite.plan.ConventionTraitDef;
 import org.apache.calcite.plan.RelOptRuleCall;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.rules.ProjectMergeRule;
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
```diff
@@ -48,6 +50,12 @@ public boolean matches(RelOptRuleCall call) {
     Project topProject = call.rel(0);
     Project bottomProject = call.rel(1);
 
+    // Make sure both projects be LogicalProject.
+    if (topProject.getTraitSet().getTrait(ConventionTraitDef.INSTANCE) != Convention.NONE ||
```
amansinha100: At first glance, it is non-intuitive that a Drill specific MergeProjectRule returns False if a DRILL_LOGICAL convention is encountered.  However, currently, this rule only checks for the complex expression (e.g convert_to_json), so your change is fine in the context of the current rule.  
In future, would we consider allowing merging 2 DrillProjectRels ? (I think yes).

jinfengni: You are right that the current Drill specific MergeProjectRule extends from Calcite's version and the only difference is to check complex output expression. Therefore, it make senses the currently Drill MergeProjectRule operates on Calcite logical convention only. 

Yes, in the future, we would consider allowing merging 2 DrillProjectRels. That should be another instance of rule, which operates on DRILL_LOGICAL convention only.

amansinha100:
+1

PR #706, opened 2016-12-27
DRILL-5039: NPE - CTAS PARTITION BY (<char-type-column>)

arina-ielchiieva:
1. Moved varchar `newPartitionValue` functions to `NewValueFunctions` template:
a. took advantage of code generation for varchar `newPartitionValue` functions;
b. all `newPartitionValue` functions will be in one file.
2. Fixed logic for reassigning nullable varchar holders during `newPartitionValue` function execution (previously resulted in NPE if varchar partition column contained nulls).
3. Added `alltypes_optional.parquet` (each type contains at least one null value) in test resources and updated `alltypes_required.parquet` to have the same types as in `alltypes_optional.parquet`.
4. Added new unit test `testPartitionByForAllTypes` to cover all types and cases (with and without nulls) during CTAS with partitioning.
5. Updated `minMaxEmptyNonNullableInput` test to reflect changes in `alltypes_required.parquet`.


exec/java-exec/src/main/codegen/templates/NewValueFunctions.java
```diff
@@ -17,6 +17,12 @@
  */
 <@pp.dropOutputFile />
 
+<#macro reassignHolder>
+        previous.buffer = buf.reallocIfNeeded(length);
+        previous.buffer.setBytes(0, in.buffer, in.start, in.end - in.start);
```
arina-ielchiieva: Done.

exec/java-exec/src/main/codegen/templates/NewValueFunctions.java
```diff
@@ -17,8 +17,14 @@
...
 
+<#macro reassignHolder>
+        previous.buffer = buf.reallocIfNeeded(length);
+        previous.buffer.setBytes(0, in.buffer, in.start, length);
+        previous.end = length;
+</#macro>
 
-<@pp.changeOutputFile name="/org/apache/drill/exec/expr/fn/impl/GNewValueFunctions.java" />
+
+<@pp.changeOutputFile name="/org/apache/drill/exeBasec/expr/fn/impl/GNewValueFunctions.java" />
```
amansinha100: Is this the correct output destination ?  I get an error in my IDE:
The declared package "org.apache.drill.exec.expr.fn.impl" does not match the expected package "org.apache.drill.exeBasec.expr.fn.impl"	GNewValueFunctions.java	/drill-java-exec/target/generated-sources/org/apache/drill/exeBasec/expr/fn/impl

arina-ielchiieva: Fixed. Thanks for noticing. Not sure how it has leaked out.

Ben-Zvi:
Is this the same bug as DRILL-4872 ?

amansinha100:
+1

PR #707, opened 2016-12-28
DRILL-5157: Multiple Snappy versions on class path

paul-rogers:
Multiple Snappy versions on class path; causes unit test failures.

Drill's pom.xml files bring in multiple Snappy versions. Drill itself brings in a very old version that has a known problem loading the snappy native library. Other libraries bring in a newer version. The one that ends up first on the class path is non-deterministic, leading to random test failures.

This fix updates the Snappy library to the latest and adds dependency management to exclude older versions brought in by Avro and Parquet.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -609,6 +609,11 @@
       <version>0.9.44</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.xerial.snappy</groupId>
+      <artifactId>snappy-java</artifactId>
+      <version>1.1.2.6</version>
+    </dependency>
 
   </dependencies>
 
@@ -1331,6 +1336,17 @@
             </exclusions>
           </dependency>
           <dependency>
+          	<groupId>org.apache.parquet</groupId>
+          	<artifactId>parquet-hadoop</artifactId>
+          	<version>${parquet.version}</version>
+          	<exclusions>
+          		<exclusion>
+          			<groupId>org.xerial.snappy</groupId>
+          			<artifactId>snappy-java</artifactId>
+          		</exclusion>
+          	</exclusions>
+          </dependency>
+          <dependency>
             <groupId>org.apache.hbase</groupId>
             <artifactId>hbase-server</artifactId>
             <version>${hbase.version}</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1331,6 +1336,28 @@
...
+          		<exclusion>
+          			<groupId>org.xerial.snappy</groupId>
+          			<artifactId>snappy-java</artifactId>
+          		</exclusion>
+          	</exclusions>
+          </dependency>
+          <dependency>
+          	<groupId>org.apache.parquet</groupId>
+          	<artifactId>parquet-hadoop</artifactId>
+          	<version>1.8.1-drill-r0</version>
```
jinfengni: Can we use ${parquet.version} here, which is defined in the beginning of root/pom.xml?  It will save effort, in case Drill changes version number for parquet library.

pom.xml
```diff
@@ -1772,6 +1799,28 @@
...
```
jinfengni: Use ${parquet.version}

pom.xml
```diff
@@ -1331,6 +1336,28 @@
             </exclusions>
           </dependency>
           <dependency>
+          	<groupId>org.apache.avro</groupId>
+          	<artifactId>avro</artifactId>
+          	<version>1.7.4</version>
```
jinfengni: Where did you get "1.7.4"?  Seems java-exec/pom.xml says that Drill uses 1.7.7.

paul-rogers: hive-exec-shade includes Avro 1.7.5. java-exec uses 1.7.7. Hive-exec uses ${avro.version} -- which is not set anywhere that I can find.

The key here is to exclude snappy, so I moved the exclusion into java-exec.

paul-rogers:
@bitblender - The general problem with jar dependencies is that each Maven project depends on a different version of library x. However, the JVM can load only one version (if everything goes into the global class loader, as in Drill.) So, we have to exclude all the random versions to ensure that a single version is included. In general, we must assume that newer versions are compatible with old versions. Else, we're in a world of hurt. (Guava is a good example: Guava removes symbols that used to exist, so that a newer version is not usable by code that needs an older one.)

All we can do is test to ensure that 1) we have a single version of the library, and 2) everything works with that version. The nature of this PR is that we violated item 1: we have multiple snappy versions. The one that gets used depends on how the class path is formed: one goes first in production, a different one in the IDE.

pom.xml
paul-rogers: Fixed.

paul-rogers:
Result:

```
[INFO] |  +- org.apache.avro:avro:jar:1.7.4:provided
[INFO] |  |  +- (org.xerial.snappy:snappy-java:jar:1.0.4.1:provided - omitted for conflict with 1.1.2.6)
...
\- org.xerial.snappy:snappy-java:jar:1.1.2.6:compile
```

paul-rogers:
@jinfengni - please review the revisions.

jinfengni:
+1

LGTM

paul-rogers:
Merged into master by @jinfengni , but he omitted the magic message to close the PR. Closing manually.

PR #429, opened 2016-03-14
DRILL-4504: Create an event loop for each of [user, control, data] RPC components

sudheeshkatkam:

PR #709, opened 2016-12-28
DRILL-5156: BootStrapContext should close threads

paul-rogers:
The Bit-Client thread (that's the thread name) finds a closed allocator in TestDrillbitResilience unit test. This fix (along with DRILL-5157) eliminates two run-time problems seen in this unit tests.

BootStrapContext creates two thread pools, but does not close them. This allows the code running in the threads to attempt to access their allocators after the allocator is closed. This fix ensures that the
thread pools are closed to avoid the issue.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/BootStrapContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/BootStrapContext.java
@@ -124,6 +124,16 @@ public class BootStrapContext implements AutoCloseable {
   @Override
   public void close() {
     try {
+      loop2.shutdownGracefully(0, 0, TimeUnit.SECONDS);
+    } catch ( Exception e ) {
+      logger.warn("Failure During Bit-Client shutdown.", e);
+    }
+    try {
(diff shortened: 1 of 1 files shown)


sudheeshkatkam:
This change may not be that simple, but I could be wrong. I tried to do something similar as part of [PR 429](https://github.com/apache/drill/pull/429/commits/0394f4caffff5aed142bb2ba0b192f3588cfda7b).

The close happens elsewhere. The "loop" is actually closed as part of 
[BasicServer#close](https://github.com/apache/drill/blob/master/exec/rpc/src/main/java/org/apache/drill/exec/rpc/BasicServer.java#L218). But it will be closed multiple times (because there may be multiple instances of sub-classes of BasicServer, and all use the same loop), and looks like "loop2" is not closed anywhere. The changes in PR 429 close the loops exactly once, but I do not recollect why the PR is not merged.

ilooner:
@paul-rogers is this fix still valid? Or can we close this?

paul-rogers:
We can close this. The origin was seeing a resource leak. There was debate. There have been other related fixes.

If the leak still exists, folks will find it and we can devise a fix based on current code.

PR #710, opened 2016-12-28
DRILL-5126: Provide simplified, unified "cluster fixture" for test

paul-rogers:
Drill provides a robust selection of test frameworks that have evolved to satisfy the needs of a variety of test cases. However, some do some of what a given test needs, while others to other parts. Also, the various frameworks make assumptions (in the form of boot-time configuration) that differs from what some test may need, forcing the test to start, then stop, then restart a Drillbit - an expensive operation.

Also, many ways exist to run queries, but they all do part of the job. Several ways exist to change runtime options.

This checkin shamelessly grabs the best parts from existing frameworks, adds a fluent builder facade and provides a complete, versitie test framework for new tests. Old tests are unaffected by this new code.

An adjustment was made to allow use of the existing TestBuilder mechanism. TestBuilder used to depend on static members of BaseTestQuery. A "shim" allows the same code to work in the old way for old tests, but with the new ClusterFixture for new tests.

Details are in the org.apache.drill.test.package-info.java file.

This commit modifies a single test case, TestSimpleExternalSort, to use the new framework. More cases will follow once this framework itself is committed.

Also, the framework will eventually allow use of the extended mock data source from SQL. However, that change must await checkin of the mock data source changes.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSimpleExternalSort.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSimpleExternalSort.java
@@ -22,67 +22,94 @@ import static org.junit.Assert.assertTrue;
 
 import java.util.List;
 
-import org.apache.drill.BaseTestQuery;
-import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.common.expression.ExpressionPosition;
 import org.apache.drill.common.expression.SchemaPath;
-import org.apache.drill.common.util.FileUtils;
 import org.apache.drill.common.util.TestTools;
-import org.apache.drill.exec.client.DrillClient;
+import org.apache.drill.exec.ExecConstants;
+import org.apache.drill.exec.exception.SchemaChangeException;
+import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.record.RecordBatchLoader;
 import org.apache.drill.exec.rpc.user.QueryDataBatch;
-import org.apache.drill.exec.server.Drillbit;
-import org.apache.drill.exec.server.RemoteServiceSet;
 import org.apache.drill.exec.vector.BigIntVector;
+import org.apache.drill.test.ClientFixture;
+import org.apache.drill.test.ClusterFixture;
+import org.apache.drill.test.DrillTest;
+import org.apache.drill.test.FixtureBuilder;
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestRule;
 
-import com.google.common.base.Charsets;
-import com.google.common.io.Files;
-
-public class TestSimpleExternalSort extends BaseTestQuery {
+public class TestSimpleExternalSort extends DrillTest {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(TestSimpleExternalSort.class);
-  DrillConfig c = DrillConfig.create();
-
 
   @Rule public final TestRule TIMEOUT = TestTools.getTimeoutRule(80000);
 
-  @Ignore
   @Test
-  public void mergeSortWithSv2() throws Exception {
-    List<QueryDataBatch> results = testPhysicalFromFileWithResults("xsort/one_key_sort_descending_sv2.json");
-    int count = 0;
-    for(QueryDataBatch b : results) {
-      if (b.getHeader().getRowCount() != 0) {
-        count += b.getHeader().getRowCount();
-      }
-    }
-    assertEquals(500000, count);
+  public void mergeSortWithSv2Legacy() throws Exception {
+    mergeSortWithSv2(true);
+  }
 
+  /**
+   * Tests the external sort using an in-memory sort. Relies on default memory
+   * settings to be large enough to do the in-memory sort (there is,
+   * unfortunately, no way to double-check that no spilling was done.)
+   * This must be checked manually by setting a breakpoint in the in-memory
+   * sort routine.
+   *
+   * @param testLegacy
+   * @throws Exception
+   */
+
+  private void mergeSortWithSv2(boolean testLegacy) throws Exception {
+    try (ClusterFixture cluster = ClusterFixture.standardCluster( );
+         ClientFixture client = cluster.clientFixture()) {
+      chooseImpl(client, testLegacy);
+      List<QueryDataBatch> results = client.queryBuilder().physicalResource("xsort/one_key_sort_descending_sv2.json").results();
+      assertEquals(500000, client.countResults( results ));
+      validateResults(client.allocator(), results);
+    }
+  }
+
+  private void chooseImpl(ClientFixture client, boolean testLegacy) throws Exception {
+  }
+
+  @Test
+  @Ignore
+  public void sortOneKeyDescendingMergeSortLegacy() throws Throwable {
+    sortOneKeyDescendingMergeSort(true);
+  }
+
+  private void sortOneKeyDescendingMergeSort(boolean testLegacy) throws Throwable {
+    try (ClusterFixture cluster = ClusterFixture.standardCluster( );
+         ClientFixture client = cluster.clientFixture()) {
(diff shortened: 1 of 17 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSimpleExternalSort.java
```diff
@@ -22,220 +22,167 @@
...
+         ClientFixture client = cluster.clientFixture()) {
+      chooseImpl(client, testLegacy);
+      List<QueryDataBatch> results = client.queryBuilder().physicalResource("xsort/one_key_sort_descending_sv2.json").results();
+      assertEquals(500000, client.countResults( results ));
+      validateResults(client.allocator(), results);
     }
+  }
 
-    System.out.println(String.format("Sorted %,d records in %d batches.", recordCount, batchCount));
+  private void chooseImpl(ClientFixture client, boolean testLegacy) throws Exception {
```
sohami: Will the implementation for this be added once the new sort mechanism is checked in ?

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -0,0 +1,217 @@
...
+      String halves[] = part.split( " " );
+      fields.add( new FieldDef( halves[1], halves[0] ) );
+    }
+    return fields;
+  }
+
+  public Map<Integer,String> getOperators( ) {
+    Map<Integer,String> ops = new HashMap<>();
+    int n = getPlans( ).size();
+    Pattern p = Pattern.compile( "\\d+-(\\d+)\\s+(\\w+)" );
```
sohami: Verified all the regex with a sample plan from a query profile. Looks good!

exec/java-exec/src/test/java/org/apache/drill/test/FixtureBuilder.java
```diff
@@ -0,0 +1,251 @@
...
+
+    configProperty(ExecConstants.ZK_REFRESH, DEFAULT_ZK_REFRESH);
+    return this;
+  }
+
+  /**
+   * Create the embedded Drillbit and client, applying the options set
+   * in the builder. Best to use this in a try-with-resources block:
+   * <pre><code>
+   * FixtureBuilder builder = ClientFixture.newBuilder()
```
paul-rogers: Yes. A case of the comment not staying current with the code... Also revised the example to show that you have to create both the cluster and client in the current version of the code.

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -0,0 +1,217 @@
...
+    plans = new ArrayList<>( );
+    while ( m.find() ) {
+      plans.add(m.group(1));
+    }
+    return plans;
+  }
+
+  public String getScan( ) {
+    int n = getPlans( ).size();
+    Pattern p = Pattern.compile( "\\d+-\\d+\\s+(\\w+)\\(" );
```
paul-rogers: Yes. What we *should* do is provide the JSON version of the query profile in the UI so that customers/users can send us that form instead of the text form. With the JSON form, parsing is much simpler.

Yes, this can be simplified. Originally grabbed the other information such as the numbers, but not now. Also, was being paranoid incase the word "Scan" showed up as a table or field name, but "Scan(" is probably safe.

And, yes, in the general case there are multiple scans. Changed that also.

exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  private QuerySummary produceSummary(BufferingQueryEventListener listener) {
+    long start = System.currentTimeMillis();
+    int recordCount = 0;
+    int batchCount = 0;
+    QueryId queryId = null;
+    loop:
+    for ( ; ; ) {
+      QueryEvent event = listener.get();
```
paul-rogers: Really? The listener handles errors as follows:
```
  @Override
  public void submissionFailed(UserException ex) {
    silentPut( new QueryEvent( ex ) );
  }
```

That is, when an error arrives from the server, we create an event and put it into the event queue, later to be retrieved by the above get. Perhaps show me the error case you discovered as I'm likely missing something here.

I've seen other tests where any failure is delivered using the `submissionFailed` call...

paul-rogers:
Rebased onto master. Pulled in additional changes from development branch that have accumulated in the three weeks that this PR has worked its way though the process. See the description in the third commit for details.

exec/java-exec/src/test/java/org/apache/drill/test/LogFixture.java
```diff
@@ -0,0 +1,255 @@
...
+    ple.setContext(lc);
+    ple.start();
+
+    appender = new ConsoleAppender<>( );
+    appender.setContext(lc);
+    appender.setName("Console");
+    appender.setEncoder( ple );
+    appender.start();
+
+    Logger root = (Logger)LoggerFactory.getLogger(Logger.ROOT_LOGGER_NAME);
```
sohami: We are checking only for console appender in DrillLogger. Can't root logger have the console appender and drillLogger may not ? In which case we will overwrite the console appender of root logger and while restoring we will remove the console appender from both.

exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
```diff
@@ -49,6 +57,107 @@
...
+
+    @Override
+    public void queryIdArrived(QueryId queryId) {
+      this.queryId = queryId;
+    }
+
+    @Override
+    public void submissionFailed(UserException ex) {
+      future.completed(new QuerySummary(queryId, recordCount, batchCount,
+          System.currentTimeMillis() - startTime, ex));
```
sohami: formatting

exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
```diff
@@ -49,6 +57,107 @@
...
+   * just the summary of the query.
+   */
+
+  public class QuerySummaryFuture implements Future<QuerySummary> {
+
+    /**
+     * Synchronizes the listener thread and the test thread that
+     * launched the query.
+     */
+
```
sohami: please remove extra space here and other places too. Like line 120

exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
```diff
@@ -57,20 +166,36 @@
...
+      this.queryId = queryId;
+      records = recordCount;
+      batches = batchCount;
+      ms = elapsed;
+      finalState = null;
+      error = ex;
+    }
 
+    public boolean failed() { return error != null; }
+    public boolean succeeded() { return error == null; }
```
sohami: formatting here ?

exec/java-exec/src/test/java/org/apache/drill/test/LogFixture.java
paul-rogers: As I understand the semantics, each logger can have multiple appenders. We add ours to any that exist. When we remove ours, we only remove that one, not any others that might be registered. Of course, I still might be missing something...

exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
paul-rogers: Fixed.

exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
paul-rogers: Fixed the other way: consistent white space before & after Javadoc comments.

sohami:
+1 LGTM

sudheeshkatkam:
+1, thanks!

paul-rogers:
Rebased on master, resolved conflict and squashed commits.

PR #711, opened 2016-12-28
DRILL-5164: Equi-join query results in CompileException

Serhii-Harnyk:
1. Added method call rotateBlock() to avoid adding expressions into single Jblock.
2. Added unit tests to cover three types of joins: hash join, merge join, nested loop join.


jinfengni:
+1

LGTM.

PR #713, opened 2016-12-29
DRILL-3562: Query fails when using flatten on JSON data where some do…

Serhii-Harnyk:
…cuments have an empty array
1. Added set for ListWriters tracking to keep empty arrays for further initializing in ensureAtLeastOneField method. 
2. Added check to avoid schema generating with field type "Late" and mode "Optional", replaced it to "Int" type in FlattenRecordBatch class.
3. Added unit tests to cover cases querying Json with empty arrays with flatten.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
@@ -27,6 +27,7 @@ import org.apache.drill.common.expression.FieldReference;
 import org.apache.drill.common.expression.LogicalExpression;
 import org.apache.drill.common.expression.SchemaPath;
 import org.apache.drill.common.logical.data.NamedExpression;
+import org.apache.drill.common.types.Types;
 import org.apache.drill.exec.exception.ClassTransformationException;
 import org.apache.drill.exec.exception.OutOfMemoryException;
 import org.apache.drill.exec.exception.SchemaChangeException;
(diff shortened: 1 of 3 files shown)


arina-ielchiieva:
Looks good.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
```diff
@@ -305,12 +306,23 @@ protected boolean setupNewSchema() throws SchemaChangeException {
...
+    final FieldReference fieldReference = flattenExpr.getRef();
+    final TransferPair transferPair = getFlattenFieldTransferPair(fieldReference);
+
+    if (transferPair != null) {
+      final ValueVector flattenVector = transferPair.getTo();
+
+      // checks that list has only default ValueVector and replaces resulting ValueVector to INT typed ValueVector
+      if (exprs.size() == 0 && flattenVector.getField().getType().equals(Types.LATE_BIND_TYPE)) {
+        final MaterializedField outputField = MaterializedField.create(fieldReference.getAsNamePart().getName(), Types.OPTIONAL_INT);
+        final ValueVector vector = TypeHelper.getNewVector(outputField, oContext.getAllocator());
```
paul-rogers: The fix appears to be to transform an empty list into an empty list of integers. That is, Drill does not have the concept of "empty list", only "empty list of type X" and we are guessing the type to be integer.

We've had issues elsewhere in the product where such guesses turn out to be wrong. Perhaps the next row/batch has a non-empty list, but of strings. Or worse, of objects (maps.) Downstream operators cannot handle this.

The result is that a query fails for no better reason than we caused it to fail by guessing the wrong type.

Clearly, fixing the broader problem is beyond the scope of this fix. I am pointing out, however, that a consequence of the assumptirnmade here is that some queries, somewhere later, will fail due to an artificial schema change.

The correct solution is to introduce an "Unknown" type and mark this a vector of type "Unknown". All we know is that it is a list; the member types are unknown. Then, in downstream operators, when we encounter a schema change, we know that an empty list of "Unknown" type is compatible with a list of any other type (say maps.)

exec/java-exec/src/main/java/org/apache/drill/exec/vector/complex/fn/JsonReader.java
```diff
@@ -59,6 +59,12 @@
   private final boolean readNumbersAsDouble;
 
   /**
+   * Collection for tracking empty array writers during reading
+   * and storing them for initializing empty arrays
+   */
+  private final Set<ListWriter> emptyArrayWritersSet = Sets.newHashSet();
```
amansinha100: Any reason why this needs to be HashSet rather than just a List ?  The HashSet may change the insertion order, so if you had 2 or more empty arrays in the same Json doc within one list,  the output of Flatten could end up changing the order.

Serhii-Harnyk: @amansinha100 Yes, you are right, in this place should be used List. Fixed it.

Serhii-Harnyk: But it should be mentioned that this list may contain many duplicates, as every ListWriter from every empty array field will be added into it. 
To avoid secondary call field.integer() on the same ListWriter object, which was added many times into the list, in the method JsonReader.ensureAtLeastOneField() we check valueCapacity of every ListWriter. 
Also LinkedHashSet may be used there, but in the case of many empty array fields it may cause additional extra cost.

Serhii-Harnyk:
@amansinha100, could you please review new changes?

amansinha100:
LGTM.  +1

PR #714, opened 2016-12-29
DRILL-4919: Fix select count(1) / count(*) on csv with header

arina-ielchiieva:
When columns from file are not needed in result (`isSkipQuery = true`), empty list of columns is replaced for text files with `columns[0]`. But currently text files with headers do not support `columns` syntax, so for these type of files we need to replace empty list of columns with default named column.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/CompliantTextRecordReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/CompliantTextRecordReader.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.store.easy.text.compliant;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
 import com.univocity.parsers.common.TextParsingException;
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/CompliantTextRecordReader.java
```diff
@@ -91,6 +96,9 @@ public boolean apply(@Nullable SchemaPath path) {
 
   @Override
   protected List<SchemaPath> getDefaultColumnsToRead() {
+    if (settings.isHeaderExtractionEnabled()) {
```
gparai: Please add a comment describing the change - something similar to your comment 
> ... currently text files with headers do not support columns syntax, so for these type of files we need to replace empty list of columns with default named column when header extraction is enabled

arina-ielchiieva: Added.

gparai:
+1 LGTM

parthchandra:
+1

PR #715, opened 2017-01-01
DRILL-5105: remove buffer size checking in getBuffers since it is onl…

chunhui-shi:
…y for debug and the redundancy increases exponentially.

### DIFF
--- a/exec/vector/src/main/java/org/apache/drill/exec/vector/complex/MapVector.java
+++ b/exec/vector/src/main/java/org/apache/drill/exec/vector/complex/MapVector.java
@@ -134,10 +134,9 @@ public class MapVector extends AbstractMapVector {
 
   @Override
   public DrillBuf[] getBuffers(boolean clear) {
-    int expectedSize = getBufferSize();
-    int actualSize   = super.getBufferSize();
-
-    Preconditions.checkArgument(expectedSize == actualSize);
+    //int expectedSize = getBufferSize();
+    //int actualSize   = super.getBufferSize();
+    //Preconditions.checkArgument(expectedSize == actualSize);
     return super.getBuffers(clear);
   }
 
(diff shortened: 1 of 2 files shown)


exec/vector/src/main/java/org/apache/drill/exec/vector/complex/MapVector.java
```diff
@@ -134,10 +134,6 @@ public int getBufferSizeFor(final int valueCount) {
 
   @Override
   public DrillBuf[] getBuffers(boolean clear) {
```
paul-rogers: perhaps replace with:
```
assert getBufferSize() == super.getBufferSize();
```

We still get the benefit of the check in tests; but the assert is ignored in production runs. This is, in fact, the advantage of the Java `assert` statement over the Guava `Preconditions` methods.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/RepeatedMapVector.java
```diff
@@ -413,14 +413,9 @@ public RepeatedMapAccessor getAccessor() {
 
   @Override
   public DrillBuf[] getBuffers(boolean clear) {
```
paul-rogers: Same comment as above.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/MapVector.java
chunhui-shi: Thanks for suggesting to use assert. It is a nice way to avoid doing this check in production. My thoughts is we don't need any check here, the reasons are:

1, mapVector does not have its own 'real' data vectors but just go to underlying vectors and get a sum of buffer sizes.
2, for non-mapVectors, there may be multiple vectors(offsets, bits, etc, such as ListVector) or the getBufferSize() is just simply get writeIndex, for which super.getBufferSize() is identical to this.getBufferSize(). And if there is any issues that non-mapVector did not calculate bufferSize correctly, we should have already seen in using that specific vector, thus we don't need to do it in mapVector code.

Actually my first thought was rewrite the logic about how to recursively doing bufferSize check. But after I read all getBufferSize on all vectors, I decided the check is not needed at all. Since any error in getBufferSize should immediately cause serious problem(serialize and deserialize) and can not pass the functional tests of that specific vector.

parthchandra:
+1

PR #716, opened 2017-01-01
DRILL-5116: Enable generated code debugging in each Drill operator

paul-rogers:
DRILL-5052 added the ability to debug generated code. The reviewer suggested permitting the technique to be used for all Drill operators. This PR provides the required fixes. Most were small changes, others dealt with the rather clever way that the existing byte-code merge converted static nested classes to non-static inner classes, with the way that constructors were inserted at the byte-code level and so on. See the JIRA for the details.

This code passed the unit tests twice: once with the traditional byte-code manipulations, a second time using "plain-old Java" code compilation. Plain-old Java is turned off by default, but can be turned on for all operators with a single config change: see the JIRA for info. Consider the plain-old Java option to be experimental: very handy for debugging, perhaps not quite tested enough for production use.

One or two operators were not converted as they appear to not be used at present, so no tests exercise them. See the JIRA for details.

This fix required visiting all operators. All generated methods declare exceptions. For plain-old Java, the based class methods must also declare exceptions. That afforded the opportunity to fix compiler warnings about resources (our over-use of {{AutoCloseable}}), missing type parameters, unused variables, and so on.


exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -64,14 +67,12 @@
...
- * to fix. Once you've done so, add the following to mark that you've
- * done the clean-up:<pre>
- * cg.plainOldJavaCapable(true); // Class supports plain-old Java</pre>
+ * Most generated classes have been upgraded to support Plain-old Java
+ * compilation. Once this work is complete, the calls to
+ * <tt>plainOldJavaCapable<tt> can be removed as all generated classes
+ * will be capable.
  * <p>
- * The setting to prefer plain-old Java is ignored for generated
+ * The setting to prefer plain-old Java is ignored for any remaining generated
```
Ben-Zvi: Just a future comment: When eventually the "plain old java" method takes over and the byte code manipulation code is eliminated, then the meaning of "plain old java" won't be clear. We should start thinking about a better term then.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -167,24 +167,61 @@ drill.exec: {
...
+    // Disable code cache. Only for testing.
+    disable_cache: false,
+    // Use plain Java compilation where available
+    prefer_plain_java: false
   },
   sort: {
     purge.threshold : 1000,
     external: {
-      batch.size : 4000,
+      // Drill uses the managed External Sort Batch by default.
```
Ben-Zvi: The following two changes are not related to DRILL-5116.  Should they be moved to another PR ?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: True. "Plain old Java" here is a reference to the commonly-used "POJO" in Java: "Plain Old Java Object." Could have called it "Plain Old Java Class"...

Yes, if/when we switch to this technique by default then, by definition, all operators must be able to generate such code and we won't need the method that declares that support (nor the code that chooses the compilation path.)

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -82,18 +83,18 @@
...
-    // The option to save code is a boot-time option because
-    // it is used selectively during debugging, but can cause
-    // excessive I/O in a running server if used to save all code.
+    // Code can be saved per-class to enable debugging.
+    // Just mark the code generator as to be persistented,
+    // point your debugger to the directory set below, and you
+    // can step into the code for debugging. Code is not saved
+    // be default because doing so is expensive and unnecessary.
 
-    saveCode = config.getBoolean(SAVE_CODE_OPTION);
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/CodeCompiler.java
```diff
@@ -117,18 +200,24 @@ public CodeCompiler(final DrillConfig config, final OptionManager optionManager)
...
-        // Generate class as plain old Java
+      return makeClass(cg);
+    }
+  }
 
-        c = classBuilder.getImplementationClass(cg);
-      } else {
-        // Generate class parts and assemble byte-codes.
+  /**
+   * Called when the requested class does not exist in the class and should
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/CodeCompiler.java
```diff
@@ -153,4 +242,17 @@ public GeneratedClassEntry(final Class<?> clazz) {
   public void flushCache() {
     cache.invalidateAll();
   }
+
+  /**
+   * Upon close, report the effectiveness of the code cache to the log.
+   */
+
+  public void close() {
+    logger.info("Class gen count: " + classGenCount);
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
```diff
@@ -374,6 +388,124 @@ public HoldingContainer declare(MajorType t, boolean includeNewInstance) {
...
+   * }</pre></code>
+   *
+   * Note: in Java 8 we'd use the <tt>Parameter</tt> class defined in Java's
+   * introspection package. But, Drill prefers Java 7 which only provides
+   * parameter types.
+   */
+
+  private void addCtor(Class<?>[] parameters) {
+    JMethod ctor = clazz.constructor(JMod.PUBLIC);
+    JInvocation superCall = null;
```
paul-rogers: Fixed.

exec/java-exec/src/main/resources/drill-module.conf
paul-rogers: Fixed. Leaked in from the original development branch.

arina-ielchiieva:
Looks good. Thank you, Paul!

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -167,10 +167,12 @@ drill.exec: {
     debug: true,
     janino_maxsize: 262144,
     cache_max_size: 1000,
-    // Enable to write generated source to disk. See ClassBuilder
-    save_source: false,
     // Where to save the generated source. See ClassBuilder
     code_dir: "/tmp/drill/codegen"
```
jinfengni: Writing to an external directory may require to add such source directory to the IDE project each time when people try to debug the generated code. 

One way I tried is to put the generated source under $drill-java-exec/target/generated-sources, which is currently for freemarker's generated source.  That way, Intellij is able to load these new generated class automatically when you debug. 

The above approach requires us to exclude those runtime generated class from the distribution jars, by adding the following in java-exec pom.xml. 

```
                  <excludes>
                    <exclude>**/test/generated/**</exclude>
                  </excludes>
``` 

The above comment is optional.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -64,36 +67,33 @@
  * local variables. Have fun!</li>
  * </ul>
  * <p>
- * Note: not all generated code is ready to be compiled as plain-old
- * Java. Some classes omit from the template the proper <code>throws</code>
- * declarations. Other minor problems may also crop up. All are easy
- * to fix. Once you've done so, add the following to mark that you've
- * done the clean-up:<pre>
- * cg.plainOldJavaCapable(true); // Class supports plain-old Java</pre>
+ * Most generated classes have been upgraded to support Plain-old Java
```
jinfengni: Do all drill operators support POJ class with this patch (the comment said "most")? If not, please list the one not supported in the comment.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -64,36 +67,33 @@
...
+ * compilation. Once this work is complete, the calls to
+ * <tt>plainOldJavaCapable<tt> can be removed as all generated classes
+ * will be capable.
  * <p>
- * The setting to prefer plain-old Java is ignored for generated
+ * The setting to prefer plain-old Java is ignored for any remaining generated
```
jinfengni: It would be good if we know what "any remaining" are exactly.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -64,36 +67,33 @@
...
   public ClassBuilder(DrillConfig config, OptionManager optionManager) {
     this.config = config;
     options = optionManager;
 
-    // The option to save code is a boot-time option because
-    // it is used selectively during debugging, but can cause
-    // excessive I/O in a running server if used to save all code.
+    // Code can be saved per-class to enable debugging.
+    // Just mark the code generator as to be persisted,
+    // point your debugger to the directory set below, and you
```
jinfengni: Since the code generator uses a sequence number as part of the class name (ProjectGen5.java etc), is it possible that in different run we may end up with different class name? Will it cause issue when debug? If that's the case, probably document such possiblity so that people is aware of.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
```diff
@@ -141,7 +142,9 @@ public ClassBuilder(DrillConfig config, OptionManager optionManager) {
     // A key advantage of this method is that the code can be
     // saved and debugged, if needed.
 
-    saveCode(code, name);
+    if (cg.persistCode()) {
```
jinfengni: I found "persistCode()" a bit confusing. From the method name, I thought it's the method to persist the code to directory, but looks like it's the method to check if we want to persistCode; the actual one is saveCode().

exec/java-exec/src/main/resources/drill-module.conf
paul-rogers: Using Eclipse, the /tmp/drill/codegen directory is very simple: Eclipse prompts for the source location on the first run, remembers it thereafter, and MacOS cleans up any left-over generated files on the next restart.

Looks like things are a bit more difficult in IntelliJ. I'll play around a bit to see if I can find a solution that works well for both tools. I'll create a separate JIRA and PR for that work.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: Hard to put that list anywhere but in the operators that have the issues (else we have to keep the comments in sync.) There are two. The old `SortBatch`. Also `OrderedPartitionRecordBatch`, which seems to not be used at present. These are called out in the JIRA.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: See above & the JIRA entry.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: It has not been an issue with Eclipse. I wonder if it will be for IntelliJ? Eclipse cares about the generated sources only when we step into them. Otherwise, Eclipse completely ignores the zillions of old generated files that might be lying around. And, since we generate a new copy of the file each time we generate it, the new copy will replace the old one. Even if Eclipse had the old file open from a previous session, it will replace the old one with the new one once Eclipse switches to that window.

Again, I need to look into IntelliJ to see if a more elaborate solution is needed for that tool.

I like your suggestion to document this. Once all this is committed, I'll submit a JIRA to add documentation about how to use this, along with the IntelliJ issues, to the Developer Information page on the Drill web site.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: It is the method to ask that the generated code be saved... Perhaps the confusing bit is that there are two methods of the same name: `persistCode(boolean)` which is called to ask to persist the code, and `persistCode()` which is used to check if the code should be persisted. (Somehow, `isPersistCode` or `isCodePersisted` just didn't seem as clear...)

On the next PR, I'll see if I can come up with a clearer name. At that time, I'll find a way to test the two "dead" operators so that all support plain Java. Once that is done, I can remove the calls to `supportsPlainJava` and comments about using `persistCode` that now appear in each operator to mark those that are plain-Java ready.

jinfengni: Sounds to me that `persistCode(boolean)` is to set the persist boolean flag in code generator if it's allowed, while `persistCode()` is to check if the boolean flag is turned on or not. In such case, can we call the former as `setPersistMode(boolean)`, and call the later as `isPersistable`?

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestClassTransformation.java
```diff
@@ -44,6 +44,9 @@
 
   @BeforeClass
   public static void beforeTestClassTransformation() throws Exception {
+    // Tests here require the byte-code merge technique and are meaningless
+    // if the plain-old Java technique is selected.
+    System.setProperty("drill.compile.prefer_plain_java", "false");
```
jinfengni: Any comment regarding this one? I thought the name of the options has a typo?

jinfengni:
In the current code generator, we log both the source code and the size of source / byte code, the time spent on compiling, and the complier (Janino vs JDK) at debug level. It's probably debatable to log the source code, but the other information is quite useful for analyzing customer issue. In the plain-old style generator, seems we do not put those information in debug log. For now, the patch is not enabled. But in the future, if we are going to use this new code generator in production, we had better log more information( either in the log, or in operator profile), as that will help improve service workload.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/CodeCompiler.java
```diff
@@ -41,12 +39,87 @@
...
+    /**
+     * Compile the code already generated by the code generator.
+     *
+     * @param cg the code generator for the class
+     * @return the compiled class
+     * @throws Exception if anything goes wrong
+     */
+
+    public Class<?> compile(final CodeGenerator<?> cg) throws Exception {
+       if (cg.isPlainOldJava()) {
```
jinfengni: Since we now have to modes for runtime code generation across almost all Drill operators (plain-old vs bytecode transformer), is there a way to check which mode is used for each operator in a query? It would be nice to include such information in log, to help analyze a potential issue (whether there is a bug in the bytecode transformer, or plain-old style class).

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
jinfengni: Good that you put those special operators in the JIRA. thx.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
jinfengni: What we talked about probably is slightly different. What you mean is to step into the generated code, without setting a breaking point in it.  What I mean is first run the query and have the code generated, then I setup some breakpoint in the code generated, re-run the query and start debugs. In such cases, we need make sure the two runs of query get the same class names, otherwise the breakpoint will not be effective. 
  
Either way, it would be good to doc the typical ways of using the generated code in debug mode.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
paul-rogers: Since I had to rename other method; went ahead and changed references to "plain-old Java" to just "plain Java."

exec/java-exec/src/main/java/org/apache/drill/exec/compile/ClassBuilder.java
jinfengni: It's probably fine to leave as it is now with proper document usage. If more people uses this feature and run into the class name difference issue, we can do that at that time.

jinfengni:
+1

The patch looks good to me. Thanks for the PR, @paul-rogers !

I'll squash and re-run the regression, before merge it to master branch.

PR #717, opened 2017-01-02
DRILL-5080: Memory-managed version of external sort

paul-rogers:
Please see the DRILL-5080 JIRA, and subtasks, for reasons for revision, design spec and list of changes. Basically the idea is to beef up the logic in the external sort to ensure that it operates within a defined memory budget. In order to do the work, the code was first refactored into smaller, easier to digest functions.

As a precaution, the original external sort code is unchanged. The "managed" version was created as a new operator that must be enabled. Reviewers can compare the old and new versions. Users can continue to favor the old version until the new version is fully ready for prime-time.

This PR covers the changes to the external sort itself. Tests for this operator require the test framework in DRILL-5126 and the mock data source in DRILL-5152. Tests for this operator will be issued as a separate PR once those two dependencies are committed.

Until then, the new operator is disabled by default. It can be enabled using
```
drill.sort.external.disable_managed: false
```

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
@@ -1,0 +1,293 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.spill;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.drill.exec.expr.TypeHelper;
+import org.apache.drill.exec.memory.BaseAllocator;
+import org.apache.drill.exec.record.BatchSchema;
+import org.apache.drill.exec.record.MaterializedField;
+import org.apache.drill.exec.record.VectorAccessible;
+import org.apache.drill.exec.record.VectorWrapper;
+import org.apache.drill.exec.record.selection.SelectionVector2;
+import org.apache.drill.exec.vector.BaseDataValueVector;
+import org.apache.drill.exec.vector.FixedWidthVector;
+import org.apache.drill.exec.vector.NullableVarCharVector;
+import org.apache.drill.exec.vector.NullableVector;
+import org.apache.drill.exec.vector.ValueVector;
+import org.apache.drill.exec.vector.VarCharVector;
+
+import io.netty.buffer.DrillBuf;
+
+/**
+ * Given a record batch or vector container, determines the actual memory
+ * consumed by each column, the average row, and the entire record batch.
+ */
+
+public class RecordBatchSizer {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(RecordBatchSizer.class);
+
+  /**
+   * Column size information.
+   */
+  public static class ColumnSize {
+    public final MaterializedField metadata;
+
+    /**
+     * Assumed size from Drill metadata.
+     */
+    public int stdSize;
+    /**
+     * Actual memory consumed by all the vectors associated with this column.
+     */
+    public int totalSize;
+    /**
+     * Actual average column width as determined from actual memory use. This
+     * size is larger than the actual data size since this size includes per-
+     * column overhead such as any unused vector space, etc.
+     */
+    public int estSize;
+
+    /**
+     * The size of the data vector backing the column. Useful for detecting
+     * cases of possible direct memory fragmentation.
+     */
+    public int dataVectorSize;
+    public int capacity;
+    public int density;
+    public int dataSize;
+
+    @SuppressWarnings("resource")
+    public ColumnSize(VectorWrapper<?> vw) {
+      metadata = vw.getField();
+      stdSize = TypeHelper.getSize(metadata.getType());
+
+      // Can't get size estimates if this is an empty batch.
+
+      ValueVector v = vw.getValueVector();
+      int rowCount = v.getAccessor().getValueCount();
+      if (rowCount == 0) {
+        return;
+      }
+      DrillBuf[] bufs = v.getBuffers(false);
+      for (DrillBuf buf : bufs) {
+        totalSize += buf.capacity();
+      }
+
+      // Capacity is the number of values that the vector could
+      // contain. This is useful only for fixed-length vectors.
+
+      capacity = v.getValueCapacity();
+
+      // Crude way to get the size of the buffer underlying simple (scalar) values.
+      // Ignores maps, lists and other esoterica. Uses a crude way to subtract out
+      // the null "bit" (really byte) buffer size for nullable vectors.
+
+      if (v instanceof BaseDataValueVector) {
+        dataVectorSize = totalSize;
+        if (v instanceof NullableVector) {
+          dataVectorSize -= bufs[0].getActualMemoryConsumed();
+        }
+      }
+
+      // Determine "density" the number of rows compared to potential
+      // capacity. Low-density batches occur at block boundaries, ends
+      // of files and so on. Low-density batches throw off our estimates
+      // for Varchar columns because we don't know the actual number of
+      // bytes consumed (that information is hidden behind the Varchar
+      // implementation where we can't get at it.)
+      //
+      // A better solution is to have each vector do this calc rather
+      // than trying to do it generically, but that increases the code
+      // change footprint and slows the commit process.
+
+      if (v instanceof FixedWidthVector) {
+        dataSize = stdSize * rowCount;
+      } else if ( v instanceof VarCharVector ) {
+        VarCharVector vv = (VarCharVector) v;
+        dataSize = vv.getOffsetVector().getAccessor().get(rowCount);
+      } else if ( v instanceof NullableVarCharVector ) {
+        NullableVarCharVector vv = (NullableVarCharVector) v;
+        dataSize = vv.getValuesVector().getOffsetVector().getAccessor().get(rowCount);
+      } else {
+        dataSize = 0;
+      }
+      if (dataSize > 0) {
+        density = roundUp(dataSize * 100, dataVectorSize);
+        estSize = roundUp(dataSize, rowCount);
+      }
+    }
+
+    @Override
+    public String toString() {
+      StringBuilder buf = new StringBuilder();
+      buf.append(metadata.getName());
+      buf.append("(std col. size: ");
+      buf.append(stdSize);
+      buf.append(", actual col. size: ");
+      buf.append(estSize);
+      buf.append(", total size: ");
+      buf.append(totalSize);
+      buf.append(", vector size: ");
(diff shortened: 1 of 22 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1321 @@
...
+    // Go no further if nothing changed.
+
+    if (estimatedRecordSize == origEstimate) {
+      return; }
+
+    // Maintain an estimate of the incoming batch size: the largest
+    // batch yet seen. Used to reserve memory for the next incoming
+    // batch.
+
+    estimatedInputBatchSize = Math.max(estimatedInputBatchSize, actualBatchSize);
```
Ben-Zvi: Need to move this line above the "return" on line 892 above -- Just in case a huge batch comes, but with a huge record count such that the average record does not exceed the previous estimate -- then we exit on line 892 without capturing that huge batch size.
  Actually the same applies to some more code below that makes use of the est. input batch size.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1321 @@
...
+
+    // Estimate the input record count. Add one due to rounding.
+
+    long estimatedInputRecordCount = estimatedInputBatchSize / estimatedRecordSize + 1;
+
+    // Estimate the total size of each incoming batch plus sv2. Note that, due
+    // to power-of-two rounding, the allocated size might be twice the data size.
```
Ben-Zvi: So if the record count __is always__ a power of two, then we are overestimating. E.g., 64K --> then we're adding 128KByte to the estimate.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1321 @@
...
+
+    if (freeMemory < 0) {
+      logger.error("ERROR: Free memory is negative: {}. Spill point = {}",
+                   freeMemory, spillPoint);
+    }
+    if (freeMemory <= spillPoint) {
+      return true; }
+
+    // Number of incoming batches (BatchGroups) exceed the limit and number of incoming
+    // batches accumulated since the last spill exceed the defined limit
```
Ben-Zvi: The comment is a little confusing - basically that's to force a spill (usually for testing purpose) after the number of batches in memory exceeded a special configuration parameter.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1321 @@
...
+    } catch (SchemaChangeException e) {
+      convertedBatch.clear();
+      throw UserException.unsupportedError(e)
+            .message("Unexpected schema change.")
+            .build(logger);
+    }
+    try {
+      sorter.sort(sv2);
+    } catch (SchemaChangeException e) {
+      throw UserException.unsupportedError(e)
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
paul-rogers: Fixed. Had assumed that the batch sizes were consistent, only the row widths varied. Now check for both.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
paul-rogers: The record count is not a power of two. The backing vectors are. We may have 2049 records. When we allocate the vector, we go to the next power of two, or 4096. So, we have to allow for that possible internal fragmentation. (Would be nice if we didn't have to...) So, yes, we assume, worst case, that we need twice the memory, as in the example above.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
paul-rogers: Stale comment about my attempt to understand the original implementation. Replaced with a better comment about the actual current implementation.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1321 @@
...
+    }
+
+    // We will merge. This will create yet another spilled
+    // run. Account for that.
+
+    mergeCount += 1;
+
+    // Must merge at least 2 batches to make progress.
+
+    mergeCount = Math.max(2, mergeCount);
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1321 @@
...
+   * minimizing unnecessary disk writes. The exact count must lie between
+   * the minimum and maximum spill counts.
+    */
+
+  private void spillFromMemory() {
+    int spillCount;
+    if (spillFileSize == 0) {
+      // If no spill limit given, guess half the batches.
+
+      spillCount = bufferedBatches.size() / 2;
```
paul-rogers: Actually, this line is here only as a fall back. Notice the surrounding if. We are, in fact, driven by the spill file size.

Changed this to set a minimum spill file size in setup, then eliminate this line here.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1321 @@
...
+   * @return a new, populated selection vector 2
+   */
+
+  private SelectionVector2 newSV2() {
+    SelectionVector2 sv2 = new SelectionVector2(allocator);
+    if (!sv2.allocateNewSafe(incoming.getRecordCount())) {
+      throw UserException.resourceError(new OutOfMemoryException("Unable to allocate sv2 buffer"))
+            .build(logger);
+    }
+    for (int i = 0; i < incoming.getRecordCount(); i++) {
```
paul-rogers: You're welcome.

Strangely, the `SingleBatchsorterTemplate` in the (unmanaged) external sort does not initialize the sv2.

Actually, this make sense. Notice that the code will reuse a selection vector from the upstream operator. In that case, we have to work with the sv values already applied and not try to recreate them as doing so will, say, unfilter the batch.

paul-rogers:
Build failure related to the slf4j library in the build process itself.

paul-rogers:
Rebased on latest master.

paul-rogers:
Some comment got lost in the force-push. One was related to the output batch size, suggesting we cap it at 16 MB. The reason is that value vectors about 16 MB cause memory fragmentation. A later fix will limit an output batch to either 64K rows (the size of an sv2) or so that the longest vector is smaller than 16 MB. The most recent commit added per-column size information so that we can enforce this limit. For example, we can have 64K rows with columns of size 256 bytes within a 16 MB vector. There is no reason not to allow 64K rows even for rows with four of the 256 columns. Total batch size would be 64 MB, but no single vector would be above 16 MB.

That fix will be offered, along with tests and enabling the managed sort by default, in a subsequent PR.

paul-rogers:
Rebased and squashed commits to prepare for pulling into master.

Revised the code to estimate batch size. @Ben-Zvi, can you take a quick look?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -0,0 +1,293 @@
...
+  public boolean hasSv2() { return hasSv2; }
+  public int getAvgDensity() { return avgDensity; }
+
+  public static final int MAX_VECTOR_SIZE = 16 * 1024 * 1024; // 16 MiB
+
+  /**
+   * Look for columns backed by vectors larger than the 16 MiB size
+   * employed by the Netty allocator. Such large blocks can lead to
+   * memory fragmentation and unexpected OOM errors.
+   * @return if any column is oversize
```
Ben-Zvi: @return "true" ... oversized

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -0,0 +1,293 @@
...
+   */
+  public boolean checkOversizeCols() {
+    boolean hasOversize = false;
+    for (ColumnSize colSize : columnSizes) {
+      if ( colSize.dataVectorSize > MAX_VECTOR_SIZE) {
+        logger.warn( "Column is wider than 256 characters: OOM due to memory fragmentation is possible - " + colSize.metadata.getPath() );
+        hasOversize = true;
```
amansinha100: Add break here once encountered oversized col ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1456 @@
...
+
+    if (incoming.getRecordCount() == 0) {
+      return;
+    }
+
+    // Determine actual sizes of the incoming batch before taking
+    // ownership. Allows us to figure out if we need to spill first,
+    // to avoid overflowing memory simply due to ownership transfer.
+
+    RecordBatchSizer sizer = analyzeIncomingBatch();
```
paul-rogers: In an ideal world, each batch would know its size and the code here would simply query that number. Yes, we do check on each batch to detect the case where, say, the first batch happened to have short columns and later batches have longer columns. We need this because the "copier" works only with row counts. To ensure we don't overflow memory, we need an accurate row width estimate, hence the checks.

Further, it turns out that Parquet (only some files) create strangely-sized batches, so simply taking the batch size / row count produces non-sensical numbers (rows of width 28 K when they are really 500 bytes, say.)

Of course, the best solution would be if the "copier" worked on actual memory usage: copy data until the outgoing batch reaches such-and-so size. Then the size of the average row would be moot and this whole chunk of code could be removed. That is left as an exercise for later.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -0,0 +1,1456 @@
...
+
+    int maxMergeWidth = (int) (mergeMemoryPool / targetMergeBatchSize);
+    maxMergeWidth = Math.min(mergeLimit, maxMergeWidth);
+
+    // If we can't fit all batches in memory, must spill any in-memory
+    // batches to make room for multiple spill-merge-spill cycles.
+
+    if (inMemCount > 0) {
+      if (spilledRunsCount > maxMergeWidth) {
+        spillFromMemory();
```
paul-rogers: The function is obligated to spill. If it can't (disk full? file system errors? OOM?), it will throw an exception. So, there is no status to return. Even if we did return a status, each caller would then be obligated to handle the error case, and so on up the call chain. Since exceptions do that for us, we use them rather than returning an error status.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
paul-rogers: Fixed.

amansinha100:
LGTM.  +1.  Thanks for the detailed writeup, comments and the responses to code review comments.

PR #718, opened 2017-01-04
DRILL-5127: Revert the fix for DRILL-4831

ppadma:
Please see DRILL-5127 for details.


amansinha100:
I suppose we need more time to investigate why the combination of DRILL-4831 plus DRILL-5082 fixes causes couple of other issues to show up at high concurrency.   I am ok with reverting this for now. 
+1

amansinha100:
Please make sure to run DRILL-5082 test to confirm if the regression is fixed.

PR #719, opened 2017-01-04
DRILL-5172: Display elapsed time for queries in the UI

kkhatua:
Displays the elapsed time for running queries and the total duration of
completed/failed/cancelled queries in the list of query profiles
displayed, and within a query's profile page as well.


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -72,16 +72,30 @@
...
     private String foreman;
     private String query;
     private String state;
     private String user;
 
-    public ProfileInfo(String queryId, long time, String foreman, String query, String state, String user) {
+    public ProfileInfo(String queryId, long startTime, String foreman, String query, String state, String user) {
       this.queryId = queryId;
-      this.time = new Date(time);
+      this.startTime = startTime;
```
sudheeshkatkam: Invoke the other ctor from here.

sudheeshkatkam:
+1, pending minor comment

exec/java-exec/src/main/resources/rest/profile/list.ftl
```diff
@@ -64,6 +65,8 @@
             <td>
               <div style="height:100%;width:100%">${query.getState()}</div>          
             <td>
+              <div style="height:100%;width:100%">${(.now?long - query.getStartTime()) / 1000} sec</div>
```
paul-rogers: Rounding? Should 999 ms be shown as 0 seconds or 1 second. Actually, given the short duration of some queries, should the time be shown down to the ms? That is either 1234 ms or 1.234 sec. Or, we could be fancy and show only three digits of precision: 1.23 seconds, 123,000 seconds. Further, do we want to shift units as the numbers increase: 234 ms, 23.4 seconds, 15 min 34 sec. 1 hr 23 min.?

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -106,7 +106,13 @@
   <p>STATE: ${model.getProfile().getState().name()}</p>
   <p>FOREMAN: ${model.getProfile().getForeman().getAddress()}</p>
   <p>TOTAL FRAGMENTS: ${model.getProfile().getTotalFragments()}</p>
-
+  <p>
+  <#if (model.getProfile().getState().name() == "RUNNING")> 
+    ELAPSED: ${(.now?long - model.getProfile().getStart()) / 1000}
```
paul-rogers: Rather than encoding this in the template three times, perhaps compute it ahead of time and store the computed result in the "model". Makes it easier to handle the multi-unit formatting, digits of precision, etc.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -106,7 +106,13 @@
   <p>STATE: ${model.getProfile().getState().name()}</p>
   <p>FOREMAN: ${model.getProfile().getForeman().getAddress()}</p>
   <p>TOTAL FRAGMENTS: ${model.getProfile().getTotalFragments()}</p>
-
+  <p>
+  <#if (model.getProfile().getState().name() == "RUNNING")> 
+    ELAPSED: ${(.now?long - model.getProfile().getStart()) / 1000} 
+  <#else> 
+    DURATION: ${(model.getProfile().getEnd() - model.getProfile().getStart()) / 1000}
```
paul-rogers: Units?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: Agreed. Collapsed into a single constructor. We'll use the current time directly to mark the current 'end' time.

exec/java-exec/src/main/resources/rest/profile/list.ftl
kkhatua: Good point. For long running queries, reporting in a single unit actually makes readability much harder. Modified in the next commit. Had to modify the UserBitShared class, however, to add this functionality for the query profile. Unless there is a change to the protobuf, we should be fine.

protocol/src/main/java/org/apache/drill/exec/proto/UserBitShared.java
```diff
@@ -13595,6 +13597,17 @@ public long getEnd() {
       return end_;
     }
 
+    public String getDuration() {
```
sudheeshkatkam: This is an auto-generated file. Do not define logic in this class.

kkhatua: Removed the logic. We'll do this within the Webpage using Javascript.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -105,6 +109,25 @@ public String getTime() {
...
 
+    public long getStartTime() {
+      return startTime;
+    }
+
+    public long getEndTime() {
+      return endTime;
+    }
+
+    public String getDuration() {
```
arina-ielchiieva: It seems we have similar logic here and in freemarker template for duration. May it can be removed in one of them?

protocol/src/main/java/org/apache/drill/exec/proto/UserBitShared.java
paul-rogers: Actually, the right place to do this work is in the Java code that creates the "model" that we pass to Freemarker. (Yeah, it is kind of complicated.)

Freemarker uses the classic MVC model. The template is the view. Java builds the model. (Let's ignore the controller for now.)

For an example, look in `ProfileResources.java`, which is in this PR. Look at the `ProfileInfo` class. Here we gather up the bits of information we want to display on the web page. Any calcs needed to prepare the data are done here. Then, Freemarker simply pulls out the pre-created data using the getter methods.

So, the `ProfileInfo` class could be extended to have, say, the raw duration in ms along with methods to return the computed values: `getDisplayDuration()` and `getDurationUnits()`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: The logic in this Java class provides the duration for the listing generated in the profile listing Freemarker file (list.ftl). For the actual query profile page, this gets trickier, since the freemarker template refers to methods from the auto-generated UserBitShared class in the protocol package.

arina-ielchiieva: How about we create wrapper class around `QueryProfile` (ex: `QueryProfileWrapper`) which will contain `getDuration` method along with needed `QueryProfile` methods. Logic for converting two long values (start and end) into human-readable format will be factored out into static method in `ProfileResources` (ex: `convertDurationIntoString(long start, long end)`). `QueryProfileWrapper` and `ProfileResources` will call this utility method from their `getDuration` methods.

protocol/src/main/java/org/apache/drill/exec/proto/UserBitShared.java
kkhatua: The ProfileInfo class is actually where the change was made to retrieve the value from the model. We don't create any duration field, as it is computed based on the other 2 fields (start,end) and simply provide a getter for that. However, the ProfileInfo class is used only by the profiles listing freemarker template (list.ftl). 
For the individual query profile's template (profile.ftl), the model from which we retrieve the duration value is auto-generated by the protobuf, making this a lot more complicated to implement. Hence, we took the JavaScript approach. 
I'll probably refine it to strictly follow the MVC model when I'm attempting a fix for DRILL-5190 https://issues.apache.org/jira/browse/DRILL-5190  [Display planning time for a query in its profile page ] , since that would require me to modify the protobuf.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: That could be one way to work around it. It would be good to have the human-readable formatting done through only one static method. I'm still trying to familiarize myself with the way these classes and the freemarker templates interact, so let me see if I can do with a wrapper. :)

kkhatua: Done. This is a much simpler and cleaner solution. Thank you, @arina-ielchiieva !

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -68,20 +68,35 @@
   @Inject DrillUserPrincipal principal;
   @Inject SecurityContext sc;
 
+  public static String getPrettyDuration(long durationInMillis) {
+    long hours = TimeUnit.MILLISECONDS.toHours(durationInMillis);
```
arina-ielchiieva: Please add description with output example.

arina-ielchiieva: May be this utility method can accept two parameters: `long start, long end`, so you won't have to repeat the following logic `profile.getEnd()-profile.getStart()` or `endTime - startTime`?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -105,6 +120,17 @@ public String getTime() {
...
+      return startTime;
+    }
+
+    public long getEndTime() {
+      return endTime;
+    }
+
+    public String getDuration() {
+      return getPrettyDuration(endTime - startTime);
+    }
```
arina-ielchiieva: May be we can factor out calculation? Please see comment above.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -114,6 +114,10 @@ public QueryProfile getProfile() {
     return profile;
   }
 
+  public String getProfileDuration() {
+    return ProfileResources.getPrettyDuration(profile.getEnd()-profile.getStart());
+  }
```
arina-ielchiieva: May be we can factor out calculation? Please see comment above.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -106,6 +106,7 @@
   <p>STATE: ${model.getProfile().getState().name()}</p>
   <p>FOREMAN: ${model.getProfile().getForeman().getAddress()}</p>
   <p>TOTAL FRAGMENTS: ${model.getProfile().getTotalFragments()}</p>
+  <p>DURATION: ${model.getProfileDuration()}</p>
```
arina-ielchiieva: For running queries, I see incorrect duration calculation. On `/profiles` elapsed time is shown correctly but when you step into running profile, duration is shown incorrectly.
image
image

kkhatua: The end time for a query is not set until the very end of the query, so the default value in 0L. For the `/profiles ` template, the elapsed time is computed using the current time in millis for running queries. Hence it shows correctly here. As per your suggestion, fixed it now by changing the signature to take the start and end times as arguments.

kkhatua:
Closing the pull request. Attempted to rebase and squash, but got interleaved commits with master. Will send out a new pull request.

PR #720, opened 2017-01-06
Indexr plugin

flowbehappy:
Add IndexR plugin.

IndexR project: https://github.com/shunfei/indexr
Introduction: https://github.com/shunfei/sfmind/blob/master/indexr.about.en.md
Document: https://github.com/shunfei/indexr/wiki
Mail list discussion: http://mail-archives.apache.org/mod_mbox/drill-user/201701.mbox/%3C64829063-6ACE-4D95-A238-E265A0BCAA4D%40sunteng.com%3E


contrib/storage-indexr/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+      <groupId>org.apache.drill</groupId>
+      <artifactId>drill-common</artifactId>
+      <classifier>tests</classifier>
+      <version>${project.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+  </dependencies>
+
+  <repositories>
```
jinfengni: Why is this repository declared for indexR storage plugin?

contrib/storage-indexr/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.drill.exec</groupId>
+      <artifactId>drill-java-exec</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.indexr</groupId>
```
jinfengni: Which repositories have the dependency of index-common, indexr-segment?  I tried to build drill after applying the patch, and hit mvn dependency error.

```
[ERROR] Failed to execute goal on project drill-indexr-storage: Could not resolve dependencies for project org.apache.drill.contrib:drill-indexr-storage:jar:1.10.0-SNAPSHOT: The following artifacts could not be resolved: io.indexr:indexr-common:jar:0.1.0, io.indexr:indexr-segment:jar:0.1.0, io.indexr:indexr-server:jar:0.1.0: Could not find artifact io.indexr:indexr-common:jar:0.1.0 in cdh.repo (https://repository.cloudera.com/artifactory/cloudera-repos) -> 
```

jinfengni:
Can you please open a Apache Drill JIRA at https://issues.apache.org/jira/browse/DRILL/? Put all the related links for IndexR storage plugin in the Drill JIRA. Also, update the commit message with something like "DRILL-XXXX: IndexR storage plugin ... ".

flowbehappy:
@jinfengni 

Sorry for the delay reply. 

My mistake to forget posting IndexR jars to public maven repository. And while doing this, we found that we cannot use the "io.indexr" groupid. Because we don't own this domain, which is registered by others already. So we plan to change to "com.sunteng.indexr", which is our company's official domain. And doing this will need to update the pom.xml and lots of package name, also some compatibility work to do. As it have been deployed into a large cluster, which make it a little complicated, like the "INPUTFORMAT" in Hive.

So we are going to close this pull request and reopen a new one when we are ready. 

Really sorry for that :)

paul-rogers:
There is another PR, for indexr-0.3.0. Is this an earlier version of that one? Is so, perhaps copy the links in this PR into the JIRA ticket needed for this PR, then close this PR in favor of the newer one. Or, if both are needed, perhaps explain how the two differ?

PR #723, opened 2017-01-20
DRILL-5207: Improve Parquet Scan pipelining.

parthchandra:
Add a configurable AsyncPageReader Queue.
  Enforce total size of parquet row group.
  Do not initialize BufferedDirectBufInputStream buffer in init. Wait for
first read.
  Change default size of BufferedDirectBufInputStream.
  Do not invoke getOptions too many times in Parquet reader.
  Add metrics for processing time, and decoding time for varlen and fixedlen columns.


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -74,12 +82,17 @@ protected void loadDictionaryIfExists(final ColumnReader<?> parentStatus,
         handleAndThrowException(e, "Error Reading dictionary page.");
       }
       // parent constructor may call this method before the thread pool is set.
```
sudheeshkatkam: Why not ensure parent calls the c'tor with pool != null?

That way all variables can be final and non-null, the `offer(...)` is only done once in the c'tor, and the internal classes can directly access the `pageQueue`.

sudheeshkatkam:
+1

PR #724, opened 2017-01-20
DRILL-5241: JDBC proxy driver: Do not put null value in map

davidhaller:
Hello everyone,

proxyReturnClass is always null, so interfacesToProxyClassesMap will contain null values only. Adding newProxyReturnClass should be correct.

This bug does not affect functionality, but probably decreases performance because you get "cache misses" all the time.

Best regards,
David.


paul-rogers:
Hi David,

Thanks much for the fix! Is there a JIRA entry for this fix? If not, please go ahead and create one. Then, rename the PR to:
```
DRILL-xxxx: DBC proxy driver: Do not put null value in map
```

That name format is needed by our committers to trace each PR back to the JIRA entry that describes it.

Thanks!

davidhaller:
JIRA issue created at https://issues.apache.org/jira/browse/DRILL-5241

PR #726, opened 2017-01-24
DRILL-5218: Support disabling hearbeats from C++ client

sudheeshkatkam:
+ remove invalid code (server should not request "handshake" type), in fact, client should fail in that case

### DIFF
--- a/contrib/native/client/src/clientlib/drillClientImpl.cpp
+++ b/contrib/native/client/src/clientlib/drillClientImpl.cpp
@@ -150,15 +150,18 @@ connectionStatus_t DrillClientImpl::connect(const char* host, const char* port){
 }
 
 void DrillClientImpl::startHeartbeatTimer(){
-    DRILL_MT_LOG(DRILL_LOG(LOG_TRACE) << "Started new heartbeat timer with "
-        << DrillClientConfig::getHeartbeatFrequency() << " seconds." << std::endl;)
-    m_heartbeatTimer.expires_from_now(boost::posix_time::seconds(DrillClientConfig::getHeartbeatFrequency()));
-    m_heartbeatTimer.async_wait(boost::bind(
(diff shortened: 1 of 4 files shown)


contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -1400,22 +1404,6 @@ void DrillClientImpl::handleRead(ByteBuf_t _buf,
             break;
 
         case exec::user::HANDSHAKE:
```
laurentgo: maybe you can also remove the case statement

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -179,9 +179,11 @@ connectionStatus_t DrillClientImpl::sendHeartbeat(){
 }
 
 void DrillClientImpl::resetHeartbeatTimer(){
-    m_heartbeatTimer.cancel();
-    DRILL_MT_LOG(DRILL_LOG(LOG_TRACE) << "Reset Heartbeat timer." << std::endl;)
-    startHeartbeatTimer();
+    if (DrillClientConfig::getHeartbeatFrequency() > 0) {
+        m_heartbeatTimer.cancel();
```
laurentgo: I was wondering if one needs to cancel the timer, as startHearbeatTimer sets it again to expire. Maybe all this logic can be done in one place (like startHeartbeatTimer?)

I also noticed another place where m_heartbeatTime.cancel() is called, in broadcastError. I guess this is fine (probably not an error of calling cancel() if not set, but haven't checked asio doc on it), but maybe this should be cleaned up/guarded too...

sudheeshkatkam: You are right, cancelling is not required because `timer.expires_from_now(...)` cancels the timer as well. So I will remove this method and move the check to `startHeartbeatTimer`.

PR #727, opened 2017-01-25
DRILL-5219: Relax user properties validation in C++ client

laurentgo:
Unlike Java client, C++ client only allows user properties present in a
whitelist. Relax this restriction so that user can add extra properties.


sudheeshkatkam:
+1

PR #728, opened 2017-01-25
DRILL-5220: Provide API to set application/client names in C++ connector

laurentgo:
Add method to DrillClientConfig to set the client and the application names
in the C++ connector.

Allow the ODBC driver (or any user of the C++ connector) to provide more
specific informations like the application using the client.


sudheeshkatkam:
+1

PR #731, opened 2017-01-27
DRILL-5224: CTTAS: fix errors connected with system path delimiters (…

arina-ielchiieva:
…Windows)

Replaced `java.nio.file.Paths.get()` to `org.apache.hadoop.fs.Path` as the latest works the same disregarding the system. Errors related to `java.nio.file.Paths.get()` usage are described in Jira DRILL-5224.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
@@ -19,7 +19,6 @@ package org.apache.drill.exec.rpc.user;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserSession.java
```diff
@@ -278,7 +277,7 @@ public boolean setSessionOption(String name, String value) {
    */
   public String registerTemporaryTable(AbstractSchema schema, String tableName) throws IOException {
       addTemporaryLocation((WorkspaceSchemaFactory.WorkspaceSchema) schema);
-      String temporaryTableName = Paths.get(sessionId, UUID.randomUUID().toString()).toString();
+      String temporaryTableName = new Path(sessionId, UUID.randomUUID().toString()).toUri().getPath();
```
paul-rogers: I wonder if we should use one of the many platform-independent path representations? We have the original Java `File`, the new Java `Path`, the Hadoop HDFS `Path`... If we use strings for paths, we run into the very problems that those packages are designed to address.

For example, do the above handle the "D:" directory suffix on Windows? Resolving "." and ".." in paths? The "~" indicator of use home folder? These all vary across platforms.

arina-ielchiieva: In this case we use path  just to concatenate two paths: session id + temp table generated name (to which refer as temp table full name). org.apache.hadoop.fs.Path uses common path delimiter which it can resolve for any system. The above just returns xxx/xxx without file system specifics.

paul-rogers: Sounds good.

parthchandra:
+1

PR #732, opened 2017-01-30
DRILL-5230: Translation of millisecond duration into hours is incorrect

kkhatua:
Fixed invalid representation of readable elapsed time using `TimeUnit` class in JDK.
e.g. 4545 sec is now correctly translated as `1h15m` instead of `17h15m`


kkhatua:
Pending testcases

PR #735, opened 2017-02-01
DRILL-5237: FlattenRecordBatch loses nested fields from the schema when returns empty batches for the first time

Serhii-Harnyk:
Fixed loss of nested fields from the schema when the first FlattenRecordBatch is empty


amansinha100:
lgtm.  +1

PR #736, opened 2017-02-02
DRILL-5238: CTTAS: unable to resolve temporary table if workspace is …

arina-ielchiieva:
…indicated without schema

1. Added additional check for passed table is temporary if default workspace is partially indicated (issue is described in Jira DRILL-5238).
2. Removed `UserSession.getDefaultSchemaName()` method as it contained exactly the same logic as `UserSession.getDefaultSchemaPath()`.
3. Since `UserSession.properties` are set using builder, they are not required but when not set, they may cause NPE when `UserSession.getProp()` method is called. Moved `UserSession.properties` into constructor.
4. Unit test for partial schema usage with temporary table.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
@@ -479,25 +479,27 @@ public class SqlConverter {
     private boolean allowTemporaryTables;
 
     DrillCalciteCatalogReader(CalciteSchema rootSchema,
-                                     boolean caseSensitive,
-                                     List<String> defaultSchema,
-                                     JavaTypeFactory typeFactory,
-                                     DrillConfig drillConfig,
-                                     UserSession session) {
+                              boolean caseSensitive,
+                              List<String> defaultSchema,
+                              JavaTypeFactory typeFactory,
+                              DrillConfig drillConfig,
+                              UserSession session) {
       super(rootSchema, caseSensitive, defaultSchema, typeFactory);
       this.drillConfig = drillConfig;
       this.session = session;
       this.allowTemporaryTables = true;
     }
 
-    /** Disallow temporary tables presence in sql statement (ex: in view definitions) */
+    /**
+     * Disallow temporary tables presence in sql statement (ex: in view definitions)
+     */
     public void disallowTemporaryTables() {
       this.allowTemporaryTables = false;
     }
 
     /**
      * If schema is not indicated (only one element in the list) or schema is default temporary workspace,
-     * we need to check among session temporary tables first in default temporary workspace.
+     * we need to check among session temporary tables in default temporary workspace first.
      * If temporary table is found and temporary tables usage is allowed, its table instance will be returned,
      * otherwise search will be conducted in original workspace.
      *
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -527,5 +529,31 @@ public RelOptTableImpl getTable(final List<String> names) {
...
+     * <li>select * from t<li/>
+     * <li>select * from dfs.tmp.t<li/>
+     * <li>use dfs; select * from tmp.t<li/>
+     *
+     * @param names             list of schema and table names, table name is always the last element
+     * @param defaultSchemaPath current schema path set using USE command
+     * @param drillConfig       drill config
+     * @return true if check for temporary table should be done, false otherwise
+     */
+    private boolean checkForTemporaryTable(List<String> names, String defaultSchemaPath, DrillConfig drillConfig) {
```
sudheeshkatkam: static

arina-ielchiieva: `checkForTemporaryTable` is just private method used only in `DrillCalciteCatalogReader` class. I don't think it should be `private static.`

sudheeshkatkam: To emphasize that this method does not rely on any internal state; it's fine either way.

arina-ielchiieva: Makes sense. I'll make a change.

arina-ielchiieva: Oh, actually I can't since `DrillCalciteCatalogReader` is inner class and inner classes can't have static declarations :)

paul-rogers: `checkForTemporaryTable` --> `isTemporaryTable` ?

Actually, given the first condition, is this `mightBeTemporaryTable`?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -527,5 +529,31 @@ public RelOptTableImpl getTable(final List<String> names) {
...
+      if (names.size() == 1) {
+        return true;
+      }
```
paul-rogers: A single name might be a temporary table, or it might not be. Hence the name suggestion above. Is the intent to simply apply a filter so that we separate out those that are not temporary tables from those that may or may not be?

If we want a positive check that this is, in fact, a temp table, wouldn't we have to resolve the name against known temp tables rather than just infer based on path length?

And, if we are just filtering out known non-temp tables, is that good enough for the use case? That is, can the case we are fixing handle the ambiguous "may or may not be temporary" case?

arina-ielchiieva: Renamed method to `mightBeTemporaryTable`. You are right the intent here is just to filter out cases when we don't need to check is passed table might be temporary. When table is passed to `getTableName` we don't know if it's temporary or not. If it is temporary we need to substitute the name to generated one and there are several indicators when we need to check this: table name is without schema or indicated schema is default temporary workspace,in all other cases we just check regular tables directly.

PR #738, opened 2017-02-02
DRILL-5190: Display planning and queued time for a query in its profile page

kkhatua:
Modified UserSharedBit protobuf for marking planning end time. This will allow for accurately reporting the planning time of a query.
In the absence of this property, for older profiles, the root fragment's (i.e. SCREEN operator) start time is taken as the estimated end of planning time.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
@@ -31,6 +31,7 @@ import org.apache.drill.exec.proto.UserBitShared.MajorFragmentProfile;
 import org.apache.drill.exec.proto.UserBitShared.MinorFragmentProfile;
 import org.apache.drill.exec.proto.UserBitShared.OperatorProfile;
 import org.apache.drill.exec.proto.UserBitShared.QueryProfile;
+import org.apache.drill.exec.proto.UserBitShared.QueryResult.QueryState;
 import org.apache.drill.exec.proto.helper.QueryIdHelper;
 import org.apache.drill.exec.server.options.OptionList;
 import org.apache.drill.exec.server.options.OptionValue;
@@ -41,6 +42,8 @@ import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
  * Wrapper class for a {@link #profile query profile}, so it to be presented through web UI.
  */
 public class ProfileWrapper {
+  private static final String ESTIMATED_LABEL = " (Estimated)";
+  private static final String NOT_AVAILABLE_LABEL = "Not Available";
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProfileWrapper.class);
   private static final ObjectMapper mapper = new ObjectMapper().enable(INDENT_OUTPUT);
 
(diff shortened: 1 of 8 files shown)


sudheeshkatkam:
+1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -77,12 +77,14 @@
    */
   public static String getPrettyDuration(long startTimeMillis, long endTimeMillis) {
     long durationInMillis = (startTimeMillis > endTimeMillis ? System.currentTimeMillis() : endTimeMillis) - startTimeMillis;
-    long hours = TimeUnit.MILLISECONDS.toHours(durationInMillis);
+    long days = TimeUnit.MILLISECONDS.toDays(durationInMillis);
```
paul-rogers: This logic more-or-less duplicates that in DRILL-5230. Please pull this logic out into a common utility, and adopt a common display format.

kkhatua: @paul-rogers , The format in DRILL-5230 is compacted to an approximation (i.e. we retain the 2 largest units of time) so as to fit within the cells of a table. 
This format is more verbose as it gives the complete picture, down to the smallest unit of time. Do we want to still merge the two into a single utility method?

paul-rogers: Good point. Maybe have a single utility class with two methods: one for the compact form, one for the verbose form. Callers can choose the format best for a particular presentation.

Note that, in both cases, we need to tear apart the duration into various units. This suggests not a static class, but an actual "real" class:
```
DurationFormat df = new DurationFormat( durationMs );
System.out.println( df.compact() );
System.out.println( df.verbose() );
```

kkhatua: Initializing an object for rendering compact or verbose duration formats might be an overkill. 
I'll change the signature for the existing method _getPrettyDuration_ in the _utility_ class - `ProfileResources.java` to carry an additional parameter to indicating verbose or compact representation and apply that in DRILL-5230. I can rebase this commit on top of that.

paul-rogers: The cost of initializing an object is trivial. Unless we are in critical performance code, code clarity and simplicity is more important than a few object creations. Indeed, in the code below, each time we do a
```
stringA + stringB
```
We create a temporary String object. So, we are already creating a large collection of objects.

The question is really 1) code repetition (not duplicating the code to split out fields) and 2) ease-of-use (making it easy to use the duration format utility.)

Thanks for combining this code into a utility. Once committed, I will shamelessly reuse the code to replace the hacks I've used for displaying durations.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -122,6 +125,31 @@ public String getQueryId() {
     return id;
   }
 
+  public String getPlanningDuration() {
+    //Check if Plan End is valid
+    if (profile.getStart() > profile.getPlanEnd()) {
```
paul-rogers: Code clarity is always a goal. One handy technique is to pick off conditions quickly. Here, we could do:
```
if (profile.getStart() <= profile.getPlanEnd()) {
  return ProfileResources.getPrettyDuration(profile.getStart(), profile.getPlanEnd());
}
// Don't estimate...

File: exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java [hunk 3]
```diff
@@ -417,6 +417,9 @@ private void parseAndRunPhysicalPlan(final String json) throws ExecutionSetupExc
 
   private void runPhysicalPlan(final PhysicalPlan plan) throws ExecutionSetupException {
     validatePlan(plan);
+    //Marking endTime of Planning
+    queryManager.markPlanningEndTime();
```
paul-rogers: Nit. The next line is also part of planning, though should execute quite quickly.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -106,6 +106,7 @@
   <p>STATE: ${model.getProfile().getState().name()}</p>
   <p>FOREMAN: ${model.getProfile().getForeman().getAddress()}</p>
   <p>TOTAL FRAGMENTS: ${model.getProfile().getTotalFragments()}</p>
+  <p>PLANNING: ${model.getPlanningDuration()}</p>
```
paul-rogers: Is planning part of duration? Or, is total time planning time + (execution) duration?

If planning time is part of the (overall) duration, may be show that visually somehow. Indented? And, if we show planning time, should we also not show queue time and execution time?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
kkhatua: Right. The method actually is multiple steps..
1. Validate Plan
2. SortMemAllocations
3. Wait In Queue
4. Setup Fragments
5. Execute... 

I wanted to explicitly provide the delta time until step 1. 
Step 2 is not related to planning (though we can consider including it). Logically, it is planning. 
Step 3 is probably useful only in the RM scenario, but I cannot estimate it for older profiles. Let's file a **separate** JIRA for this bit? It most certainly involves another addition to the protobuf. 
Step 4. Expected to be a low cost operation like Step 2 and not certainly related to planning.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
kkhatua: I agree. Working on restructuring it.

paul-rogers:
Please modify the title to include the JIRA ticket number:

DRILL-XXXX: Display planning and queued...

sudheeshkatkam:
+1

PR #739, opened 2017-02-03
DRILL-5230: Translation of millisecond duration into hours is incorrect

kkhatua:
Fixed invalid representation of readable elapsed time using `TimeUnit` class in JDK.
e.g. 4545 sec is now correctly translated as `1h15m` instead of `17h15m`
TestCase has been added

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
@@ -68,25 +68,6 @@ public class ProfileResources {
   @Inject DrillUserPrincipal principal;
   @Inject SecurityContext sc;
 
-  /**
-   * Returns elapsed time a human-readable format. If end time is less than the start time, current epoch time is assumed as the end time.
-   * e.g. getPrettyDuration(1468368841695,1468394096016) = '7 hr 00 min 54.321 sec'
-   * @param startTimeMillis Start Time in milliseconds
-   * @param endTimeMillis   End Time in milliseconds
-   * @return                Human-Readable Elapsed Time
-   */
-  public static String getPrettyDuration(long startTimeMillis, long endTimeMillis) {
-    long durationInMillis = (startTimeMillis > endTimeMillis ? System.currentTimeMillis() : endTimeMillis) - startTimeMillis;
-    long hours = TimeUnit.MILLISECONDS.toHours(durationInMillis);
-    long minutes = TimeUnit.MILLISECONDS.toMinutes(durationInMillis) - TimeUnit.HOURS.toMinutes(TimeUnit.MILLISECONDS.toHours(durationInMillis));
-    long seconds = TimeUnit.MILLISECONDS.toSeconds(durationInMillis) - TimeUnit.MINUTES.toSeconds(TimeUnit.MILLISECONDS.toMinutes(durationInMillis));
-    long milliSeconds = durationInMillis - TimeUnit.SECONDS.toMillis(TimeUnit.MILLISECONDS.toSeconds(durationInMillis));
-    String formattedDuration = (hours > 0 ? hours + " hr " : "") +
-      ((minutes + hours) > 0 ? String.format("%02d min ", minutes) : "") +
-      seconds + "." + String.format("%03d sec", milliSeconds) ;
-    return formattedDuration;
-  }
-
   public static class ProfileInfo implements Comparable<ProfileInfo> {
     public static final SimpleDateFormat format = new SimpleDateFormat("MM/dd/yyyy HH:mm:ss");
 
@@ -137,7 +118,7 @@ public class ProfileResources {
     }
 
     public String getDuration() {
-      return getPrettyDuration(startTime, endTime);
+      return (new SimpleDurationFormat(startTime, endTime)).verbose();
     }
 
     public String getState() {
@@ -237,8 +218,8 @@ public class ProfileResources {
       return new QProfiles(runningQueries, finishedQueries, errors);
     } catch (Exception e) {
       throw UserException.resourceError(e)
-          .message("Failed to get profiles from persistent or ephemeral store.")
-          .build(logger);
+      .message("Failed to get profiles from persistent or ephemeral store.")
+      .build(logger);
     }
   }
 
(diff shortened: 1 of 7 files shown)


sudheeshkatkam:
+1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -56,6 +47,30 @@ public TableBuilder(final String[] columns) {
...
+   */
+  public String shortDurationFormat(long millis) {
+    long days = TimeUnit.MILLISECONDS.toDays(millis);
+    long hours = TimeUnit.MILLISECONDS.toHours(millis) - TimeUnit.DAYS.toHours(TimeUnit.MILLISECONDS.toDays(millis));
+    long minutes = TimeUnit.MILLISECONDS.toMinutes(millis) - TimeUnit.HOURS.toMinutes(TimeUnit.MILLISECONDS.toHours(millis));
+    long seconds = TimeUnit.MILLISECONDS.toSeconds(millis) - TimeUnit.MINUTES.toSeconds(TimeUnit.MILLISECONDS.toMinutes(millis));
+    long milliSeconds = millis - TimeUnit.SECONDS.toMillis(TimeUnit.MILLISECONDS.toSeconds(millis));
+    String formattedDuration = "";
+    if (days >= 1) {
+      formattedDuration = days + "d" + hours + "h" + minutes + "m";
```
paul-rogers: Is this the display format? Maybe add some space:

```
5d 14h 2m
```

Instead of:
```
5d14h2m
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -56,6 +47,30 @@ public TableBuilder(final String[] columns) {
...
+    } else if (hours >= 1) {
+      formattedDuration = hours + "h" + minutes + "m";
+    } else if (minutes >= 1) {
+      formattedDuration = minutes + "m" + seconds + "s";
+    } else {
+      formattedDuration = seconds + "." + milliSeconds + "s";
```
paul-rogers: The above won't work in general. If the time is 2 ms, you'll get an output of 0.2s, which is not quite right...

Here, consider using the very handy `String.format` method:
```
String.format("%.3f", seconds + milliSeconds / 1000.0 );
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -56,6 +47,30 @@ public TableBuilder(final String[] columns) {
...
+    }
+    return formattedDuration;
```
paul-rogers: A nit, but the code is clearer if you do:

```
if (cond1) {
    return expr1;
} else if (cond2) {
    return expr2;
} else {
    return expr3;
}
```

The above eliminates the temp variable and makes it very clear what is happening.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -56,6 +47,30 @@ public TableBuilder(final String[] columns) {
     sb.append("</tr>\n");
   }
 
+  /**
+   * Representation of a millisecond duration in a short readable text
+   * @param millis  Duration in milliseconds
+   * @return        Human-Readable Duration Time
+   */
+  public String shortDurationFormat(long millis) {
```
paul-rogers: Used elsewhere? If so, consider making static (so we don't need an instance.) If only used here, consider making private.

In fact, maybe move this out to a utility class. This logic can be used anywhere we want to display a duration such as in logs, in tests, etc.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -78,29 +93,8 @@ public void appendTime(final long d, final String link) {
...
-    }else if (mins >= 1){
-      timeFormat = this.smins;
-    }else if (secs >= 10){
-      timeFormat = this.secs;
-    }else {
-      timeFormat = this.ssecs;
-    }
-    appendCell(timeFormat.format(new Date(p)), null);
+    String shortReadableDuration = shortDurationFormat(p);
+    appendCell(shortReadableDuration, link);
```
paul-rogers: In Drill, we like brevity:
```
appendCell(shortDurationFormat(p), link);
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
kkhatua: This display format is kept compact due to limited space within the cells of the table. Hence, we aim to keep only the first two major units. Hence, if something runs in hours, we report down to the minute, but not any more granular.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
kkhatua: Good catch. I'll update the test case as well and apply the String.format()

kkhatua:
@paul-rogers , @sudheeshkatkam 
Committed changes based on your recommendations. I also noticed that some of the calls were passing fragment IDs as links, which were never applied. Based on the file history, my hunch is that it was meant for debugging and forgotten ever since. Corrected it by replacing these with the appropriate `null` values.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -352,16 +369,26 @@ public String cancelQuery(@PathParam("queryid") String queryId) {
...
-          .message("Not authorized to cancel the query '%s'", queryId)
-          .build(logger);
+      .message("Not authorized to cancel the query '%s'", queryId)
+      .build(logger);
     }
   }
+
+  /**
+   * Duration format definition
+   * @author kkhatua
```
paul-rogers: Drill practice seems to be to omit the author. Many IDE's helpfully add this, but you can turn off that option.

kkhatua: My bad. Will fix that.

kkhatua:
@paul-rogers Created a new SimpleDurationFormat class. We can expand to have more formats, or reimplement on lines of SimpleDateFormat by passing format strings in the future. Hope this helps.

kkhatua:
This closes PR #739

PR #740, opened 2017-02-03
DRILL-5240: Parquet - fix unnecessary object creation while checking …

parthchandra:
…for null values in nullable var length columns


jinfengni:
+1

LGTM

PR #741, opened 2017-02-04
DRILL-5196: init MongoDB cluster when run a single test case directly…

chunhui-shi:
… through command line or IDE.

Other fixes include:
Sync mongo-java-driver versions to newer 3.2.0
update flapdoodle package to latest accordingly


contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
```diff
@@ -204,16 +209,20 @@ private static void cleanup() {
   @BeforeClass
   public static void initMongo() throws Exception {
     synchronized (MongoTestSuit.class) {
-      if (distMode) {
-        logger.info("Executing tests in distributed mode");
-        DistributedMode.setup();
-      } else {
-        logger.info("Executing tests in single mode");
-        SingleMode.setup();
+      if (initCount.get() == 0) {
```
paul-rogers: non-synchronized and if (initCount.getAndIncrement() == 0) ?

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
```diff
@@ -234,15 +243,25 @@ private static void createDbAndCollections(String dbName,
...
-    if (mongoClient != null) {
-      mongoClient.dropDatabase(EMPLOYEE_DB);
-      mongoClient.close();
-    }
     synchronized (MongoTestSuit.class) {
-      if (distMode) {
-        DistributedMode.cleanup();
-      } else {
-        SingleMode.cleanup();
+      if (initCount.decrementAndGet() == 0) {
```
paul-rogers: Because using the atomic int, also need the synchronized?

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
```diff
@@ -234,15 +243,25 @@ private static void createDbAndCollections(String dbName,
...
+        try {
+          if (mongoClient != null) {
```
paul-rogers: Move to top of function?
```
if (mongoClient==null) { return; }
if (initCount.decrementAndGet() == 0) { ...
```

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
chunhui-shi: Without Synchronized, there is still chance two threads go into this place. while thread 0 go to here and is in the middle of tearing down cluster, then thread 1 increases initCount to 1 and somehow report error then come to here to decrease initCount and continue... Then we don't know what will happen.

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
chunhui-shi: Then we will end up not to cleanup the cluster if mongoClient for some reason was set to null? I think it is better leave it as is.

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
chunhui-shi: Then it will increase the count when the init is not done. If init fail in the middle, then another test has no chance to try.

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/TestTableGenerator.java
```diff
@@ -58,7 +59,12 @@ public static void generateTable(String dbName, String collection,
         .jsonArray(jsonArray).importFile(jsonFile).build();
     MongoImportExecutable importExecutable = MongoImportStarter
         .getDefaultInstance().prepare(mongoImportConfig);
-    importExecutable.start();
+    MongoImportProcess importProcess = importExecutable.start();
+
+    while (importProcess.isProcessRunning()) {
```
gparai: Please add a comment explaining the change - Import process is running in a different thread/process and we do not to start running any test prior to import complete?

contrib/storage-mongo/src/test/resources/datatype.json
```diff
@@ -0,0 +1 @@
+[{"_id": {"$oid" : "582081d96b69060001fd8938"}, "account" : { "$ref": "contact", "$id": {"$oid": "999cbf116b69060001fd8611"} }}, {"_id" : {"$oid" : "582081d96b69060001fd8939" } } ]
```
gparai: Should we rename it to something like datatype_dbref.json? The name seems to suggest we have all datatypes in this document

gparai:
+1 LGTM

sudheeshkatkam:
+1

PR #742, opened 2017-02-04
DRILL-5242: The UI breaks when rendering profiles having unknown metrics

kkhatua:
Skip any metrics whose metric ID is unknown, This prevents any ArrayIndexOutOfBoundsException from being thrown and breaking the UI rendering.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
@@ -19,6 +19,8 @@ package org.apache.drill.exec.server.rest.profile;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 
 import com.google.common.base.Preconditions;
 import org.apache.commons.lang3.tuple.ImmutablePair;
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -163,11 +165,18 @@ public String getMetricsTable() {
...
-          values[metric.getMetricId()] = metric.getDoubleValue();
+        if (metric.getMetricId() < metricNames.length) {
+          if (metric.hasLongValue()) {
+            values[metric.getMetricId()] = metric.getLongValue();
+          } else if (metric.hasDoubleValue()) {
+            values[metric.getMetricId()] = metric.getDoubleValue();
+          }
+        } else {
+          //Tracking unknown metric IDs
+          unknownMetrics.add(metric.getMetricId());
```
paul-rogers: Will this work? We leave the metric unset, then iterate over them in the following loop. Also, we build the unknownMetrics set, but never use it.

Suggestion: if the metric is not known, just set its value to 0, and log a message to the log file.

This situation occurs only when 1) opening an old profile with metrics that no longer exist, or 2) when adding a metric but not registering it properly.

Note that this code does not handle the case where the metric is not registered (the metric names is null.)

For the external sort, the problem occurred because we have two sets of metric enums (two implementations) but only one registry of names. The fix was to keep the old and new ones in sync.

kkhatua: The unknownMetrics set is built incase we want to use this list of metrics to do something. I tried logging by publishing the number of unknown metrics found, but the problem is that it floods the log. A fix could be to do something to publish it just once per rendering of the page. But either way, it has no real value. 
For me, the unknown metric IDs existed because the profile was created from a prototype/forked Drill version. It would be nice if the registry of these IDs was part of the profile, but that is a separate problem. 
The simplest suggestion from @parthchandra and @sudheeshkatkam was to ignore these metrics for now. The expectation is that elements of the profile which are unknown to the rendering version of Drill should never break the UI.

kkhatua: @paul-rogers Can we bless this? :)

kkhatua:
@sudheeshkatkam Please review the PR and label the JIRA, so that @jinfengni can get it into the commit list for master.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
paul-rogers: Sure. Still not convinced we need to build a map to throw away, but I guess doing so, while silly, is harmless.

sudheeshkatkam:
+1

kkhatua:
This closes PR #742

PR #745, opened 2017-02-11
DRILL-5252: Fix a condition that always returns true

lifove:
Fix for https://issues.apache.org/jira/browse/DRILL-5252


jinfengni:
+1

PR #746, opened 2017-02-12
DRILL-5254: Enhance default reduction factors in optimizer

paul-rogers:
Provides enhanced default selectivity rules as defined in DRILL-5254. Please see the JIRA entry for the details.

The rules are selectable with a boot-time option and are disabled by default to allow full testing of the rules before they become enabled by default.

Also provides a complete unit test. The test required enhancements to the new test framework; those enhancements are also included.

### DIFF
--- a/exec/java-exec/src/main/resources/drill-module.conf
+++ b/exec/java-exec/src/main/resources/drill-module.conf
@@ -91,7 +91,22 @@ drill.exec: {
     use.ip : false
   },
   optimizer: {
-    implementation: "org.apache.drill.exec.opt.IdentityOptimizer"
+    implementation: "org.apache.drill.exec.opt.IdentityOptimizer",
+    enhanced_defaults: {
+      // If true, uses the enhanced Drill reduction factor rules.
+      // If false, uses the default Calcite rules.
+      enable: false,
+      // p(a = value)
+      // Probability that a column equals some value.
+      // Default is 0.15 in Calcite, 0.10 in most textbooks (and System-R)
+      prob_eq: 0.15,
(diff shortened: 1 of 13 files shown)


paul-rogers:
Community asks this not be in Drill but rather offered to Calcite.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -91,7 +91,22 @@ drill.exec: {
     use.ip : false
   },
   optimizer: {
-    implementation: "org.apache.drill.exec.opt.IdentityOptimizer"
+    implementation: "org.apache.drill.exec.opt.IdentityOptimizer",
+    enhanced_defaults: {
```
jinfengni: Any particular reason why you make this option as boot-time option, in stead of system option? Majority planner options, if not all, are able to reset either at system or session level. That will give the user option for better control.

paul-rogers: Could not work out, in the limited time available, how to pass a system option (which can change at runtime) into the default rules (which appears to not have visibility to the query context.)

Any suggestions for how to pass along the query context so we can get at the runtime options?

jinfengni: You may pass in a difference instance of DrillDefaultRelMetaDataProvider, based on context.getPlannerSetting(). 

https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java#L382

PR #748, opened 2017-02-14
DRILL-5263: Prevent left NLJoin with non scalar subqueries

Serhii-Harnyk:


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/NestedLoopJoinPrule.java
```diff
@@ -49,7 +49,7 @@ protected boolean checkPreconditions(DrillJoinRel join, RelNode left, RelNode ri
       PlannerSettings settings) {
     JoinRelType type = join.getJoinType();
 
-    if (! (type == JoinRelType.INNER || type == JoinRelType.LEFT)) {
+    if (!(type == JoinRelType.INNER || (type == JoinRelType.LEFT && JoinUtils.hasScalarSubqueryInput(left, right)))) {
```
zfong: Don't we want to disallow left outer joins in all cases, even if one of the join inputs is a scalar?

Serhii-Harnyk: In the case when table joining with the result of the query with aggregate function, Drill creates physical plan with left nested loop join.
It is due to this code line:
https://github.com/mapr/incubator-calcite/blob/DrillCalcite1.4.0/core/src/main/java/org/apache/calcite/sql2rel/SqlToRelConverter.java#L1078
So to allow execution of such queries without changes in calcite - allow left nested loop join for scalar subqueries.

zfong: I see.  So, Calcite is converting a query with a subquery into an equivalent left join with a scalar subquery.  Do you know if the query returns the correct result in that case if a nested loop join is used?  If not, then it seems like we should still be returning an error.

Also, how do you distinguish the case where the query has been converted to this form by Calcite vs a user explicitly doing a left outer join to a scalar subquery.  It seems like in this case, if Drill uses a nested loop join, it will also return wrong result.

Serhii-Harnyk: Yes, you a right, we should return error in this case too.

amansinha100: @Serhii-Harnyk were you able to confirm what @zfong asked about w.r.t correctness ? The code is still allowing left nested join with scalar subquery, so it seems her comment is still valid.

zfong: @amansinha100 - @Serhii-Harnyk  and I discussed this.  If we return an error in the case I noted, then we will actually fail a bunch of existing unit tests.  That's because there is Calcite code that flattens a subquery into an equivalent left outer join that uses scalar subqueries.  Those queries end up failing.  Given that, we will end up breaking queries that previously were able to run.  Your thoughts on this?

amansinha100: If we already have tests that are checking correctness for the left nested outer join where one side is scalar, then we are ok with this change.

amansinha100:
+1

Serhii-Harnyk:
Merged into apace master without message _close apache/drill#748_. Closing PR manually.

PR #750, opened 2017-02-17
DRILL-5273: CompliantTextReader excessive memory use

paul-rogers:
DRILL-5273
CompliantTextReader exhausts 4 GB memory when reading 5000 small files

Please see JIRA for details of problem and fix.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/CompliantTextRecordReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/CompliantTextRecordReader.java
@@ -118,12 +118,21 @@ public class CompliantTextRecordReader extends AbstractRecordReader {
    * @param outputMutator  Used to create the schema in the output record batch
    * @throws ExecutionSetupException
    */
+  @SuppressWarnings("resource")
   @Override
   public void setup(OperatorContext context, OutputMutator outputMutator) throws ExecutionSetupException {
 
     oContext = context;
-    readBuffer = context.getManagedBuffer(READ_BUFFER);
-    whitespaceBuffer = context.getManagedBuffer(WHITE_SPACE_BUFFER);
+    // Note: DO NOT use managed buffers here. They remain in existence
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/CompliantTextRecordReader.java
```diff
@@ -118,12 +118,21 @@ public boolean apply(@Nullable SchemaPath path) {
...
    */
+  @SuppressWarnings("resource")
   @Override
   public void setup(OperatorContext context, OutputMutator outputMutator) throws ExecutionSetupException {
 
     oContext = context;
-    readBuffer = context.getManagedBuffer(READ_BUFFER);
-    whitespaceBuffer = context.getManagedBuffer(WHITE_SPACE_BUFFER);
+    // Note: DO NOT use managed buffers here. They remain in existence
+    // until the fragment is shut down. The buffers here are large.
```
chunhui-shi: I think the reason you chose to use context.getAllocator() was you don't want to fragmentize managed buffer?
Otherwise you might just call readBuffer.close()? Was there any problem with managed buffer's release? Just curious about the "DO NOT use managed buffer here" part. Besides that, +1.

paul-rogers: The reason is a bit different. The original call allocates a managed buffer: it is freed only when the fragment context shuts down at the end of query execution. But, if we read many files (5000 in one test case), then we leave 5000 buffers in existence for the whole query.

Instead, we want to take control over buffer lifetime. We allocate a regular (not managed) buffer ourselves, and then release it when this reader closes.

That way, instead of accumulating 5000 buffers of 1 MB each, we have only one 1 MB buffer in existence at any one time.

Of course, a further refinement would be to allocate the buffer on the ScanBatch and have all 5000 readers sequentially share that same buffer. But, I was not sure that any performance benefit was worth the cost in extra code complexity...

sudheeshkatkam:
+1

PR #751, opened 2017-02-17

kkhatua:
Allow changing default number of finished queries in web UI, when starting up Drillbits
-Ddrill.exec.http.max_profiles=100
Alternatively, the page can be loaded dynamically for the same.
e.g. appending the parameter **max=\<listSize\>**
`https://<hostname>:8047/profiles?max=100`

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
@@ -31,12 +31,15 @@ import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.SecurityContext;
+import javax.ws.rs.core.UriInfo;
 import javax.xml.bind.annotation.XmlRootElement;
 
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.exceptions.UserException;
+import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.coord.ClusterCoordinator;
 import org.apache.drill.exec.coord.store.TransientStore;
 import org.apache.drill.exec.proto.GeneralRPCProtos.Ack;
@@ -60,8 +63,6 @@ import com.google.common.collect.Lists;
 @RolesAllowed(DrillUserPrincipal.AUTHENTICATED_ROLE)
 public class ProfileResources {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProfileResources.class);
-
-  public final static int MAX_PROFILES = 100;
 
   @Inject UserAuthEnabled authEnabled;
   @Inject WorkManager work;
@@ -171,10 +172,13 @@ public class ProfileResources {
     public List<String> getErrors() { return errors; }
   }
 
+  //max Param to cap listing of profiles
+  private static final String MAX_QPROFILES_PARAM = "max";
+
   @GET
   @Path("/profiles.json")
   @Produces(MediaType.APPLICATION_JSON)
-  public QProfiles getProfilesJSON() {
+  public QProfiles getProfilesJSON(@Context UriInfo uriInfo) {
     try {
       final PersistentStore<QueryProfile> completed = getProvider().getOrCreateStore(QueryManager.QUERY_PROFILE);
       final TransientStore<QueryInfo> running = getCoordinator().getOrCreateTransientStore(QueryManager.RUNNING_QUERY_INFO);
@@ -201,7 +205,15 @@ public class ProfileResources {
 
       final List<ProfileInfo> finishedQueries = Lists.newArrayList();
 
-      final Iterator<Map.Entry<String, QueryProfile>> range = completed.getRange(0, MAX_PROFILES);
+      //Defining #Profiles to load
+      int maxProfilesToLoad = work.getContext().getConfig().getInt(ExecConstants.HTTP_MAX_PROFILES);
+      String maxProfilesParams = uriInfo.getQueryParameters().getFirst(MAX_QPROFILES_PARAM);
+      if (maxProfilesParams != null && !maxProfilesParams.isEmpty()) {
+        maxProfilesToLoad = Integer.valueOf(maxProfilesParams);
+      }
+
+      final Iterator<Map.Entry<String, QueryProfile>> range = completed.getRange(0, maxProfilesToLoad);
+
       while (range.hasNext()) {
         try {
           final Map.Entry<String, QueryProfile> profileEntry = range.next();
@@ -226,8 +238,8 @@ public class ProfileResources {
   @GET
   @Path("/profiles")
   @Produces(MediaType.TEXT_HTML)
-  public Viewable getProfiles() {
-    QProfiles profiles = getProfilesJSON();
+  public Viewable getProfiles(@Context UriInfo uriInfo) {
+    QProfiles profiles = getProfilesJSON(uriInfo);
     return ViewableWithPermissions.create(authEnabled.get(), "/rest/profile/list.ftl", sc, profiles);
   }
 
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -190,10 +193,13 @@ public QProfiles(List<ProfileInfo> runningQueries, List<ProfileInfo> finishedQue
     public List<String> getErrors() { return errors; }
   }
 
+  //max Param to cap listing of profiles 
+  private static final String MAX_QPARAM = "max";
+
```
ppadma: please choose a better name than MAX_QPARAM something like MAX_QPROFILES_PARAM  ?

kkhatua: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -190,10 +193,13 @@ public QProfiles(List<ProfileInfo> runningQueries, List<ProfileInfo> finishedQue
     public List<String> getErrors() { return errors; }
   }
 
+  //max Param to cap listing of profiles 
+  private static final String MAX_QPROFILES_PARAM = "max";
```
sudheeshkatkam: "max" seems too generic for Context, say "drill.profiles.load.max"?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -61,7 +64,7 @@
 public class ProfileResources {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProfileResources.class);
 
-  public final static int MAX_PROFILES = 100;
+  public final static int MAX_PROFILES = Integer.valueOf(System.getProperty(ExecConstants.HTTP_MAX_PROFILES, "100"));
```
sudheeshkatkam: Make this part of `drill-override.conf`?

(Using an Inject, see [DrillRestServer](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRestServer.java#L88))

kkhatua: I was hoping to. Wasn't sure how to get access to the DrillConfig object.  Let me look at this.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: This is to define the param in the URL. This would be too verbose to type into the URL IMO.

sudheeshkatkam: sounds good

kkhatua: Lucky. There is a WorkManager instance available. We're able to load the default from `drilloverride.conf` and also have the user pass it via the URL.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -60,8 +63,7 @@
 @RolesAllowed(DrillUserPrincipal.AUTHENTICATED_ROLE)
 public class ProfileResources {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProfileResources.class);
-
-  public final static int MAX_PROFILES = 100;
+  private static Integer MaxProfiles = null;
```
paul-rogers: This can be an int. The config system requires that you set some value for the config property, unless you go through some hokey-pokey to check if the property exists.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -190,10 +192,13 @@ public QProfiles(List<ProfileInfo> runningQueries, List<ProfileInfo> finishedQue
     public List<String> getErrors() { return errors; }
   }
 
+  //max Param to cap listing of profiles
+  private static final String MAX_QPROFILES_PARAM = "max";
```
paul-rogers: Maybe limit? As in SQL?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -220,7 +225,21 @@ public QProfiles getProfilesJSON() {
 
       final List<ProfileInfo> finishedQueries = Lists.newArrayList();
 
-      final Iterator<Map.Entry<String, QueryProfile>> range = completed.getRange(0, MAX_PROFILES);
+      //Defining #Profiles to load
+      int maxProfilesToLoad;
+      if (MaxProfiles == null) {
+        MaxProfiles = work.getContext().getConfig().getInt(ExecConstants.HTTP_MAX_PROFILES);
```
paul-rogers: Just

{code}
int maxProfilesToLoad = work.getContext()...;
{code}

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -220,7 +225,21 @@ public QProfiles getProfilesJSON() {
...
+      }
+
+      String maxProfilesParams = uriInfo.getQueryParameters().getFirst(MAX_QPROFILES_PARAM);
```
paul-rogers: You can use annotations for the property. That's the best way in this particular context.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -220,7 +225,21 @@ public QProfiles getProfilesJSON() {
...
+      if (maxProfilesParams != null && !maxProfilesParams.isEmpty()) {
+        maxProfilesToLoad = Integer.valueOf(maxProfilesParams);
+      } else {
+        maxProfilesToLoad = MaxProfiles;
+      }
```
paul-rogers: No need for an else.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: I wasn't sure if I'd want to repeatedly load the value from the config object, as I was assuming it might have synchronized access. But considering this is not going to be called at a very high frequency, there shouldn't be any overhead.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: Good point and I was originally doing that. But considering that this is probably going to support a whole lot more parameters in the future, I thought I'll leave a generic Context annotation, and let the implementation within the method handle that.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: Agreed. Got rid of the MaxProfiles altogether. We'll read the config value and load the appropriate value accordingly.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: It's just a convenient variable name for now, since it preserves the constant originally there `MAX_PROFILES`
The patch came as a result of this [exchange](https://www.mail-archive.com/user@drill.apache.org/msg05992.html) on drill-user list, so I didn't changing it for the patch. If we use pagination, perhaps then we can. 
For now, someone in the community has proposed this, which takes this further ahead: [DRILL-5068](https://issues.apache.org/jira/browse/DRILL-5068)

sudheeshkatkam:
+1

PR #753, opened 2017-02-18
DRILL-5260: Extend "Cluster Fixture" test framework

paul-rogers:
- Config option to suppress printing of CSV and other output. (Allows
printing for single tests, not printing when running from Maven.)

- Parsing of query profiles to extract plan and run time information.

- Fix bug in log fixture when enabling logging for a package.

- Improved ZK support.

- Set up the new CTTAS default temporary workspace for tests.

- Clean up persistent storage files on disk to avoid CTTAS startup
failures.

- Provides a set of examples for how to use the cluster fixture.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
@@ -129,11 +129,18 @@ public class ClusterFixture implements AutoCloseable {
   private boolean ownsZK;
   private ZookeeperHelper zkHelper;
   private RemoteServiceSet serviceSet;
-  private String dfsTestTmpSchemaLocation;
+  private File dfsTestTempDir;
   protected List<ClientFixture> clients = new ArrayList<>();
   private boolean usesZk;
   private boolean preserveLocalFiles;
   private boolean isLocal;
+
+  /**
+   * Temporary directories created for this test cluster.
+   * Each is removed when closing the cluster.
+   */
+
+  private List<File> tempDirs = new ArrayList<>();
 
   ClusterFixture(FixtureBuilder builder) {
 
@@ -237,7 +244,7 @@ public class ClusterFixture implements AutoCloseable {
 //    new File(logDir).mkdirs();
 //    System.setProperty("drill.log-dir", logDir);
 
-     dfsTestTmpSchemaLocation = TestUtilities.createTempDir();
+    dfsTestTempDir = makeTempDir("dfs-test");
 
     // Clean up any files that may have been left from the
     // last run.
@@ -287,7 +294,7 @@ public class ClusterFixture implements AutoCloseable {
 
     @SuppressWarnings("resource")
     final StoragePluginRegistry pluginRegistry = bit.getContext().getStorage();
-    TestUtilities.updateDfsTestTmpSchemaLocation(pluginRegistry, dfsTestTmpSchemaLocation);
+    TestUtilities.updateDfsTestTmpSchemaLocation(pluginRegistry, dfsTestTempDir.getAbsolutePath());
     TestUtilities.makeDfsTmpSchemaImmutable(pluginRegistry);
 
     // Create the mock data plugin
@@ -333,6 +340,7 @@ public class ClusterFixture implements AutoCloseable {
   public RemoteServiceSet serviceSet() { return serviceSet; }
   public BufferAllocator allocator() { return allocator; }
   public DrillConfig config() { return config; }
+  public File getDfsTestTmpDir() { return dfsTestTempDir; }
 
   public ClientFixture.ClientBuilder clientBuilder() {
     return new ClientFixture.ClientBuilder(this);
@@ -404,6 +412,13 @@ public class ClusterFixture implements AutoCloseable {
         }
     }
 
+    // Remove temporary directories created for this cluster session.
+
+    try {
+      removeTempDirs();
+    } catch (Exception e) {
+      ex = ex == null ? e : ex;
+    }
   }
 
   /**
@@ -440,6 +455,27 @@ public class ClusterFixture implements AutoCloseable {
     }
   }
 
+  private void removeTempDirs() throws IOException {
+    IOException ex = null;
+    for (File dir : tempDirs) {
+      try {
+        removeDir(dir);
+      } catch (IOException e) {
+        ex = ex == null ? e : ex;
+      }
+    }
+    if (ex != null) {
+      throw ex;
+    }
+  }
+
+  public void removeDir(File dir) throws IOException {
+    if (! dir.exists()) {
+      return;
+    }
+    FileUtils.deleteDirectory(dir);
+  }
+
   /**
    * Close a resource, suppressing the exception, and keeping
    * only the first exception that may occur. We assume that only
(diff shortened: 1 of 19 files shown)


exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -99,33 +111,83 @@
...
+      zkHelper.startZookeeper(builder.localZkCount);
       ownsZK = true;
     }
+    if (zkHelper != null) {
+      zkConnect = zkHelper.getConnectionString();
+      // Forced to disable this, because currently we leak memory which is a known issue for query cancellations.
+      // Setting this causes unittests to fail.
+      builder.configProperty(ExecConstants.RETURN_ERROR_FOR_FAILURE_IN_CANCELLED_FRAGMENTS, true);
+    }
+    return zkConnect;
```
paul-rogers: Good catch. Fixed.

Lots of this code was adapted from elsewhere and I didn't always follow through the exact flow. Thanks for doing so.

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -295,8 +396,96 @@ public void close() throws Exception {
     if (ex != null) {
       throw ex;
```
paul-rogers: Fixed.

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -295,8 +396,96 @@ public void close() throws Exception {
...
+    // Don't delete if this is not a local Drillbit.
+
+    if (! isLocal) {
+      return;
+    }
+
+    // Don't delete if we did not write.
+
+    if (! config.getBoolean(ExecConstants.SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE)) {
+      return;
```
paul-rogers: Good point. The proper rule probably is that if we are using the local store, we need to clean it up prior to test start. Else, the new dynamic UDF stuff throw an error. So, for start-up, removes files is we are using the local store, write or not.

A separate option will leave files on exit when desired. (The options to leave files us usually just turned on during debugging...)

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -295,8 +396,96 @@ public void close() throws Exception {
...
+    }
+
+    // Remove the local files if they exist.
+
+    String localStoreLocation = config.getString(ExecConstants.SYS_STORE_PROVIDER_LOCAL_PATH);
+    File storeDir = new File(localStoreLocation);
+    if (! storeDir.exists()) {
+      return;
+    }
+    if (storeDir.exists()) {
```
paul-rogers: Fixed. It is a nuisance that `deleteDirectory` throws an exception if the directory is already deleted...

exec/java-exec/src/test/java/org/apache/drill/test/ExampleTest.java
```diff
@@ -0,0 +1,243 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.test;
+
+import static org.junit.Assert.*;
+
```
paul-rogers: The wild card is traditional: it allows use of any of the JUnit assertions without having to import them one-by-one.

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -42,44 +48,207 @@
...
+      String parts[] = plan.split("\n");
+      for (String part : parts) {
+        plans.add(part);
+        OpDefInfo opDef = new OpDefInfo( part );
+        operations.add(opDef);
+      }
+      sortList();
+    }
+
+    public void sortList() {
```
paul-rogers: Fixed.

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -42,44 +48,207 @@
...
+   */
+
+  private void parsePlans() {
+    PlanParser parser = new PlanParser();
+    String plan = getPlan( );
+    parser.parsePlans(plan);
+    plans = parser.plans;
+    topoOrder = parser.operations;
+    operations = parser.sorted;
+  }
```
paul-rogers: It is confusing... The query profile lists operators in topological order, so the original order is the topo order. We then sort the operators by operator ID so we can retrieve them by ID. That's what the operators list is.

Added Javadoc to explain this a bit.

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -42,44 +48,207 @@
...
+
+  private void aggregateOpers() {
+    for (FragInfo major : fragments.values()) {
+      for (OpDefInfo opDef : major.ops) {
+        for ( OperatorProfile op : opDef.opExecs) {
+          Preconditions.checkState( major.id == op.majorFragId );
+          Preconditions.checkState( opDef.stepId == op.opId );
+          opDef.actualRows += op.records;
+          opDef.actualBatches += op.batches;
+          opDef.actualMemory += op.peakMem * 1024 * 1024;
```
paul-rogers: OK. Multiplications by constant powers of 2 is pretty fast so didn't worry about it originally.

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -138,9 +414,208 @@ public long getMetric(int id) {
...
+      if (!m.matches()) {
+        throw new IllegalStateException( "Could not parse plan: " + plan );
+      }
+      majorId = Integer.parseInt(m.group(1));
+      stepId = Integer.parseInt(m.group(2));
+      name = m.group(4);
+      args = m.group(5);
+      String tail = m.group(6);
+      String indent = m.group(3);
+      globalLevel = (indent.length() - 4) / 2;
```
paul-rogers: Here we are reverse engineering a format created elsewhere in Drill. Since that format uses a minimum of four spaces, we have to subtract that off. Then, that code inserts two more spaces for each indent level.

Clearly, if we change the text format at all, the parser here breaks.

A much better solution would be to include all this info in a JSON format...

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -138,9 +414,208 @@ public long getMetric(int id) {
...
+
+      p = Pattern.compile("rowType = RecordType\\((.*)\\): (rowcount .*)");
+      m = p.matcher(tail);
```
paul-rogers: Why is that? Spaces are literals. Colons are literal in this location, they have meaning only in a pattern like (?:X). What am I missing?

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
```diff
@@ -138,9 +414,208 @@ public long getMetric(int id) {
...
+
+    @Override
+    protected void visitSubtree(OpDefInfo node, int i, int indent) {
+      System.out.print( indentString(indent) + "      " );
+      System.out.println(subtreeLabel(node, i));
+    }
+  }
+
+  public Map<Integer,OperatorProfile> getOpInfo( ) {
     Map<Integer,String> ops = getOperators( );
```
paul-rogers: Added some Javadoc. This whole class is under constant construction, so I've been lazy about adding comments for stuff that will likely change later...

sohami: Thanks!

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
sohami: Discussed offline.

sohami:
Apart from fixing regex in ProfileParser.java, changes looks good to me.
+1

exec/java-exec/src/test/java/org/apache/drill/test/ProfileParser.java
paul-rogers: As it turns out, different scan operators use a different syntax for this info. This syntax works for the scan operators used in tests thus far. We'll need to extend this for others.

Actually, we'll want to combine this functionality with the full-blown parser used earlier. This was a quick & dirty on that predated the fancier one. And, as noted, the real solution is to include the info in the profile in JSON to avoid the need to fiddle with parsing the text format.

sudheeshkatkam:
+1

PR #754, opened 2017-02-20
DRILL-5275: Sort spill is slow due to repeated allocations

paul-rogers:
DRILL-5275 -  Sort spill serialization is slow due to repeated buffer
allocations

Rather than create a heap buffer per vector when writing and reading,
the revised code creates a single, shared buffer used for all I/O
within a particular container. This improves performance by reducing GC
and CPU costs during I/Os.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorAccessibleSerializable.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorAccessibleSerializable.java
@@ -57,6 +57,12 @@ public class VectorAccessibleSerializable extends AbstractStreamSerializable {
   private BatchSchema.SelectionVectorMode svMode = BatchSchema.SelectionVectorMode.NONE;
   private SelectionVector2 sv2;
 
+  /**
+   * Disk I/O buffer used for all reads and writes of DrillBufs.
+   */
+
+  private byte buffer[] = new byte[32*1024];
+
   private boolean retain = false;
 
   public VectorAccessibleSerializable(BufferAllocator allocator) {
@@ -90,6 +96,7 @@ public class VectorAccessibleSerializable extends AbstractStreamSerializable {
    * @param input the InputStream to read from
    * @throws IOException
    */
+  @SuppressWarnings("resource")
   @Override
   public void readFromStream(InputStream input) throws IOException {
     final VectorContainer container = new VectorContainer();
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorAccessibleSerializable.java
```diff
@@ -57,6 +57,12 @@
   private BatchSchema.SelectionVectorMode svMode = BatchSchema.SelectionVectorMode.NONE;
   private SelectionVector2 sv2;
 
+  /**
+   * Disk I/O buffer used for all reads and writes of DrillBufs.
+   */
+
+  private byte buffer[] = new byte[32*1024];
```
Ben-Zvi: (1) Is 32K too small ? (because it is heap?) Does a higher size perform better ?
(2) Would it work better if the operator would allocate this serialization buffer from the direct memory (and only when spilling), and then free it at the end ?

paul-rogers: Read a 18 GB file on disk using just a test program that uses various size buffers.

{code}
32K buffer: Rate: 799 MB/s
64K buffer: Rate: 766 MB/s
{code}

So, seems no advantage of a larger buffer. (Tests with a smaller buffer do slow things down, hence the 32K size.)

On direct memory: can't use direct memory as the fundamental problem is that data is in a direct memory DrillBuf, and must be copied to heap memory for writing. The original code does the copy by allocating a heap buffer the same size as the vector (16 MB, 32 MB or larger.) This code does the copying by reusing the same buffer over and over.

No need to hold the buffer on the operator. This class is used for an entire spill/read session.

What may be an issue, however, is the merge phase of a sort when many files are open and so many buffers are created. The reads are synchronous, so they could share a buffer.

paul-rogers: The key issues with the location of the buffer are:

* Want to reuse the buffer as much as possible. (Hence, putting it on the operator is a good idea.)
* Want to keep interfaces simple (passing the buffer from the operator to everything that needs it is awkward.)
* Can only be shared by a single thread, obviously.

After playing around, it turns out we can move the read & write methods onto the {{BufferAllocator}} class. This makes them available to anything that uses Drillbufs. And, it allows the allocator to hold on to the (shared) I/O buffer.

Reflecting on the sort, it becomes clear that such a change is necessary. In the merge phase, the sort will have many spill runs open; each will have its own {{VectorAccessibleSerializable}}, each with its own buffer. Moving the buffer to the allocator reduces the needs to a single, shared, buffer.

sudheeshkatkam:
+1

PR #755, opened 2017-02-22

kkhatua:
Using Hadoop API to filter and reduce profile list load time
Using an in-memory treeSet-based cache, maintain the list of most recent
profiles.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/DrillSysFilePathFilter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/DrillSysFilePathFilter.java
@@ -1,0 +1,56 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.store.sys.store;
+
+import static org.apache.drill.exec.ExecConstants.DRILL_SYS_FILE_SUFFIX;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+
+/**
+ * Filter for Drill System Files
+ */
+public class DrillSysFilePathFilter implements PathFilter {
+
+  //NOTE: The filename is a combination of query ID (which is monotonically
+  //decreasing value derived off epoch timestamp) and a random value. This
+  //filter helps eliminate that list
+  String cutoffFileName = null;
+  public DrillSysFilePathFilter() {}
+
+  public DrillSysFilePathFilter(String cutoffSysFileName) {
+    if (cutoffSysFileName != null) {
+      this.cutoffFileName = cutoffSysFileName + DRILL_SYS_FILE_SUFFIX;
+    }
+  }
+
+  /* (non-Javadoc)
+   * @see org.apache.hadoop.fs.PathFilter#accept(org.apache.hadoop.fs.Path)
+   */
+  @Override
+  public boolean accept(Path file){
+    if (file.getName().endsWith(DRILL_SYS_FILE_SUFFIX)) {
+      if (cutoffFileName != null) {
+        return (file.getName().compareTo(cutoffFileName) <= 0);
+      } else {
+        return true;
+      }
+    }
+    return false;
+  }
+}
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
@@ -68,7 +68,7 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
   private final AutoCloseableLock readCachedProfilesLock = new AutoCloseableLock(new ReentrantLock());
 
   //Provides a threshold above which we report the time to load
-  private static final long LIST_TIME_THRESHOLD_MSEC = 2000L;
+  private static final long RESPONSE_TIME_THRESHOLD_MSEC = 2000L;
 
   private static final int drillSysFileExtSize = DRILL_SYS_FILE_SUFFIX.length();
   private final Path basePath;
@@ -76,12 +76,17 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
   private final DrillFileSystem fs;
   private int version = -1;
   private TreeSet<String> profilesSet;
+  private int profilesSetSize;
 //  private String mostRecentProfile;
   private long basePathLastModified;
   private long lastKnownFileCount;
   private int maxSetCapacity;
   private Stopwatch listAndBuildWatch;
+  private Stopwatch transformWatch;
   private PathFilter sysFileSuffixFilter;
+  private Function<String, Entry<String, V>> transformer;
+
+  private Iterable<Entry<String, V>> iterableProfileSet;
 
   public LocalPersistentStore(DrillFileSystem fs, Path base, PersistentStoreConfig<V> config) {
     super();
@@ -89,11 +94,20 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
     this.config = config;
     this.fs = fs;
     this.profilesSet = new TreeSet<String>();
+    this.profilesSetSize= 0;
     this.basePathLastModified = 0L;
     this.lastKnownFileCount = 0L;
     this.sysFileSuffixFilter = new DrillSysFilePathFilter();
     //this.mostRecentProfile = null;
     this.listAndBuildWatch = Stopwatch.createUnstarted();
+    this.transformWatch = Stopwatch.createUnstarted();
+    this.transformer = new Function<String, Entry<String, V>>() {
+      @Nullable
+      @Override
+      public Entry<String, V> apply(String key) {
+        return new ImmutableEntry<>(key, get(key));
+      }
+    };
 
     try {
       if (!fs.mkdirs(basePath)) {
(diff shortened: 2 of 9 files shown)


kkhatua:
A summary of the performance is available in this [comment](https://issues.apache.org/jira/browse/DRILL-5270?focusedCommentId=15877119&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15877119) on the JIRA (DRILL-5270)

kkhatua:
For 8266 profiles, when measured from Chrome browser's Network tool:
```
Load First Time: 2.43s 
Load Second Time (no new profiles): 829ms
```

kkhatua:
@sudheeshkatkam Can you please review the PR?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/DrillSysFilePathFilter.java
```diff
@@ -0,0 +1,53 @@
+/**
```
arina-ielchiieva: Please use comment for the header, not javadoc.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -61,16 +63,29 @@
   private final AutoCloseableLock readLock = new AutoCloseableLock(readWriteLock.readLock());
   private final AutoCloseableLock writeLock = new AutoCloseableLock(readWriteLock.writeLock());
 
+  //Provides a threshold above which we report the time to load
+  private static final long LISTTIME_THRESHOLD_MSEC = 2000L;
+
+  private static final int DrillSysFileExtSize = DRILL_SYS_FILE_SUFFIX.length();
```
arina-ielchiieva: `DrillSysFileExtSize` -> `drillSysFileExtSize`

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -61,16 +63,29 @@
   private final AutoCloseableLock readLock = new AutoCloseableLock(readWriteLock.readLock());
   private final AutoCloseableLock writeLock = new AutoCloseableLock(readWriteLock.writeLock());
 
+  //Provides a threshold above which we report the time to load
+  private static final long LISTTIME_THRESHOLD_MSEC = 2000L;
```
arina-ielchiieva: `LISTTIME_THRESHOLD_MSEC` -> `LIST_TIME_THRESHOLD_MSEC`

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -112,23 +127,65 @@ public static DrillFileSystem getFileSystem(DrillConfig config, Path root) throw
 
   @Override
   public Iterator<Map.Entry<String, V>> getRange(int skip, int take) {
+    //Marking currently seen modification time
+    long currBasePathModified = 0L;
+    try {
+      currBasePathModified = fs.getFileStatus(basePath).getModificationTime();
+    } catch (IOException ioexcp) {
+      ioexcp.printStackTrace();
```
arina-ielchiieva: Please do not use `printStackTrace()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -112,23 +127,65 @@ public static DrillFileSystem getFileSystem(DrillConfig config, Path root) throw
...
+    }
+
+    //Acquiring lock to avoid reloading for request coming in before completion of profile read
```
arina-ielchiieva: 1. Before reading lock acquirement was enough, with your changes you modify class fields. Since many threads can access this method, you'll end up with raise conditions, also class fields can be cached by threads as well... I think design here should be reconsidered.
2. Guava library has several cache implementations. Can we leverage any of them instead of using tree set?

Pinging @vlad since he is working on DRILL-6053 which intends to make changes in the same class to avoid excessive locking to be aware of intended changes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -141,11 +198,33 @@ public static DrillFileSystem getFileSystem(DrillConfig config, Path root) throw
     }
   }
 
+  /**
+   * Add profile name to a TreeSet
+   * @param profileName
```
arina-ielchiieva: Please do not leave `@param`, `@return` without description. IDE usually highlights them, asking to add description.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/DrillSysFilePathFilter.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static org.apache.drill.exec.ExecConstants.DRILL_SYS_FILE_SUFFIX;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.fs.PathFilter;
+
+/**
+ * Filter for Drill System Files
+ */
+public class DrillSysFilePathFilter implements PathFilter {
```
arina-ielchiieva: Please consider using `FileSystemUtil` which help to create filters. Passing custom filter is also possible.

kkhatua: Ok.  I was thinking of using
```
List<FileStatus> fileStatuses = DrillFileSystemUtil.listFiles(fs, basePath, false, sysFileSuffixFilter);
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
kkhatua: OK. Will fix this. Eclipse didn't pop it up for me.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
kkhatua: OK. Will fix this.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/DrillSysFilePathFilter.java
kkhatua: OK. Will fix this.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
kkhatua: I wanted to treat this like a constant, but this makes it confusing as a Class name

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
kkhatua: Will publish a log message and return an empty iterator for now. Not sure how to bubble up an error to the UI. I'll take a look at how we do so for profile deserialization as a guide

kkhatua:
@arina-ielchiieva I need to rebase this on top of the latest master considering it was originally based on nearly a year old code. When ready, i'll create a new PR or push to this one. Let me know which one works.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
kkhatua: I'll provide the explanation below.

kkhatua:
The choice for a `TreeSet` is to basically use a binary structure that keeps the (maximum permitted) profiles sorted and in memory. 

When Drill detect changes, 
(Refer https://github.com/kkhatua/drill/blob/f7ad29b9a322bb215d16b3c3b9a2bfc40abfc1ed/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java#L146) 
it will fetch all the available profiles in the PStore and reconstruct the tree (since the order of the profiles returned by the `FileSystem` is not guaranteed). 

I tried using the `PathFilter` to fetch only new profiles, but the cost of the `FileSystem` fetching only new profiles, versus the entire list is the same! Also, there is the possibility that some profiles might have been deleted as new ones were added, so a full reconstruction would take care of that scenario as well. 

To evict, as I construct the TreeSet, I simply pop the oldest (by filename) entry. The Guava cache options don't seem to provide a way to define the basis on which to evict entries.

I believe, @vrozov's work on DRILL-6053 is to address locking during writes specifically. The lock I used (and need) is for reads to ensure that multiple requests don't trigger an expensive FileSystem call for the same state of the PStore. 
e.g. consider T# as timestamps
* `currBasePathModified` = T0 
* _ThreadA_ requests at t=T1 and issues a read-lock
* _ThreadB_ requests at t=T2 but is waiting for read-lock

If the tree exists and no change is detected, _ThreadA_ will use the `TreeSet` contents and resume by releasing the lock. 

If the `TreeSet` exists and a change is detected, _ThreadA_ will reconstruct the `TreeSet` before using its contents and it will update `lastBasePathModified`, before releasing the lock.

When _ThreadB_ gets the read-lock, it discovers that during the wait, the `TreeSet` was already updated. So, in terms of t=T2, this is the most recent snapshot, so it proceeds to use the treeSet's contents rather than reconstruct. That will be deferred to the next request.

We're using the `lastBasePathModified` as a way to provide a pseudo-versioned access to the list. That means if there are more profiles added *after* _ThreadB_ was waiting for the read-lock, it will not trigger the `FileSystem` call right away.

vrozov:
@kkhatua
1. The read locks are not exclusive (single writer/multiple readers). To achieve the required functionality you need to introduce a different lock and use write (or exclusive) lock.
2. The choice for TreeSet is not obvious. What are the most common operations performed on the collection? Do you optimize for get, put or collection construction?

@arina-ielchiieva my github id is `vrozov`.

kkhatua:
Thanks, @vrozov. I'll make use of a separate lock for read-only purpose in case of `#1`.
For `#2`, I need to construct a size-limited ordered set from a list of unordered elements.
In this case, the elements (i.e. profiles) need to be ordered by file-name, which is a 1:1 mapping function of the start time epoch for the query.
So, I need to be able to add to such a datastructure in `O(log(n))` time, remove in `O(1)` and iterate through it in sequence. So, my puts are the most expensive operation.

kkhatua:
Holding off to do a rebase once @vrozov 's PR #1163 (DRILL-6053) goes into Apache.

kkhatua:
Closing this PR in favor of #1250

PR #756, opened 2017-02-22
DRILL-5195: Publish Operator and MajorFragment Stats in Profile page

kkhatua:
Improved UI
1. Introduction of Tooltips
2. Share of each operator as a percentages of the major fragment and of the query
  - This would help identify the most CPU intensive operators within a fragment and across the query
3. Rows emitted by each operator
4. For a running query, changes to 'last update' and 'last progress' now shows the elapsed time since.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
@@ -17,9 +17,12 @@
  */
 package org.apache.drill.exec.server.rest.profile;
 
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
+import java.util.Locale;
 
 import org.apache.drill.exec.proto.UserBitShared.MajorFragmentProfile;
 import org.apache.drill.exec.proto.UserBitShared.MinorFragmentProfile;
@@ -49,75 +52,174 @@ public class FragmentWrapper {
     return String.format("fragment-%s", major.getMajorFragmentId());
   }
 
-  public static final String[] FRAGMENT_OVERVIEW_COLUMNS = {"Major Fragment", "Minor Fragments Reporting",
-    "First Start", "Last Start", "First End", "Last End", "Min Runtime", "Avg Runtime", "Max Runtime", "Last Update",
-    "Last Progress", "Max Peak Memory"};
+  public static final String[] ACTIVE_FRAGMENT_OVERVIEW_COLUMNS = {
+      OverviewTblTxt.MAJOR_FRAGMENT, OverviewTblTxt.MINOR_FRAGMENTS_REPORTING,
+      OverviewTblTxt.FIRST_START, OverviewTblTxt.LAST_START, OverviewTblTxt.FIRST_END, OverviewTblTxt.LAST_END,
+      OverviewTblTxt.MIN_RUNTIME, OverviewTblTxt.AVG_RUNTIME, OverviewTblTxt.MAX_RUNTIME,
+      OverviewTblTxt.PERCENT_BUSY,
+      OverviewTblTxt.LAST_UPDATE, OverviewTblTxt.LAST_PROGRESS,
+      OverviewTblTxt.MAX_PEAK_MEMORY
+  };
+
+  public static final String[] ACTIVE_FRAGMENT_OVERVIEW_COLUMNS_TOOLTIP = {
+      OverviewTblTooltip.MAJOR_FRAGMENT, OverviewTblTooltip.MINOR_FRAGMENTS_REPORTING,
+      OverviewTblTooltip.FIRST_START, OverviewTblTooltip.LAST_START, OverviewTblTooltip.FIRST_END, OverviewTblTooltip.LAST_END,
+      OverviewTblTooltip.MIN_RUNTIME, OverviewTblTooltip.AVG_RUNTIME, OverviewTblTooltip.MAX_RUNTIME,
+      OverviewTblTooltip.PERCENT_BUSY,
+      OverviewTblTooltip.LAST_UPDATE, OverviewTblTooltip.LAST_PROGRESS,
+      OverviewTblTooltip.MAX_PEAK_MEMORY
+  };
 
   // Not including Major Fragment ID and Minor Fragments Reporting
-  public static final int NUM_NULLABLE_OVERVIEW_COLUMNS = FRAGMENT_OVERVIEW_COLUMNS.length - 2;
+  public static final int NUM_NULLABLE_ACTIVE_OVERVIEW_COLUMNS = ACTIVE_FRAGMENT_OVERVIEW_COLUMNS.length - 2;
 
   public void addSummary(TableBuilder tb) {
     // Use only minor fragments that have complete profiles
     // Complete iff the fragment profile has at least one operator profile, and start and end times.
     final List<MinorFragmentProfile> complete = new ArrayList<>(
-      Collections2.filter(major.getMinorFragmentProfileList(), Filters.hasOperatorsAndTimes));
+        Collections2.filter(major.getMinorFragmentProfileList(), Filters.hasOperatorsAndTimes));
 
-    tb.appendCell(new OperatorPathBuilder().setMajor(major).build(), null);
-    tb.appendCell(complete.size() + " / " + major.getMinorFragmentProfileCount(), null);
+    tb.appendCell(new OperatorPathBuilder().setMajor(major).build());
+    tb.appendCell(complete.size() + " / " + major.getMinorFragmentProfileCount());
 
     // If there are no stats to aggregate, create an empty row
     if (complete.size() < 1) {
-      tb.appendRepeated("", null, NUM_NULLABLE_OVERVIEW_COLUMNS);
+      tb.appendRepeated("", null, NUM_NULLABLE_ACTIVE_OVERVIEW_COLUMNS);
       return;
     }
 
     final MinorFragmentProfile firstStart = Collections.min(complete, Comparators.startTime);
     final MinorFragmentProfile lastStart = Collections.max(complete, Comparators.startTime);
-    tb.appendMillis(firstStart.getStartTime() - start, null);
-    tb.appendMillis(lastStart.getStartTime() - start, null);
+    tb.appendMillis(firstStart.getStartTime() - start);
+    tb.appendMillis(lastStart.getStartTime() - start);
 
     final MinorFragmentProfile firstEnd = Collections.min(complete, Comparators.endTime);
     final MinorFragmentProfile lastEnd = Collections.max(complete, Comparators.endTime);
-    tb.appendMillis(firstEnd.getEndTime() - start, null);
-    tb.appendMillis(lastEnd.getEndTime() - start, null);
+    tb.appendMillis(firstEnd.getEndTime() - start);
+    tb.appendMillis(lastEnd.getEndTime() - start);
 
-    long total = 0;
+    long cumulativeFragmentDurationInMillis = 0L;
+    long cumulativeProcessInNanos = 0L;
+    long cumulativeWaitInNanos = 0L;
     for (final MinorFragmentProfile p : complete) {
-      total += p.getEndTime() - p.getStartTime();
+      cumulativeFragmentDurationInMillis += p.getEndTime() - p.getStartTime();
+      //Capture Busy & Wait Time
+      List<OperatorProfile> opProfileList = p.getOperatorProfileList();
+      for (OperatorProfile operatorProfile : opProfileList) {
+        cumulativeProcessInNanos += operatorProfile.getProcessNanos();
+        cumulativeWaitInNanos += operatorProfile.getWaitNanos();
+      }
+    }
+    double totalProcessInMillis = Math.round(cumulativeProcessInNanos/1E6);
+    double totalWaitInMillis = Math.round(cumulativeWaitInNanos/1E6);
+
+    final MinorFragmentProfile shortRun = Collections.min(complete, Comparators.runTime);
+    final MinorFragmentProfile longRun = Collections.max(complete, Comparators.runTime);
+    tb.appendMillis(shortRun.getEndTime() - shortRun.getStartTime());
+    tb.appendMillis(cumulativeFragmentDurationInMillis / complete.size());
+    tb.appendMillis(longRun.getEndTime() - longRun.getStartTime());
+
+    tb.appendPercent(totalProcessInMillis / (totalProcessInMillis + totalWaitInMillis), null,
+        //#8721 is the summation sign: sum(Busy): ## + sum(Wait): ##
+        String.format("&#8721;Busy: %,.2fs + &#8721;Wait: %,.2fs", totalProcessInMillis/1E3, totalWaitInMillis/1E3));
+
+    final MinorFragmentProfile lastUpdate = Collections.max(complete, Comparators.lastUpdate);
+    tb.appendMillis(System.currentTimeMillis()-lastUpdate.getLastUpdate());
+
+    final MinorFragmentProfile lastProgress = Collections.max(complete, Comparators.lastProgress);
+    tb.appendMillis(System.currentTimeMillis()-lastProgress.getLastProgress());
+
+    // TODO(DRILL-3494): Names (maxMem, getMaxMemoryUsed) are misleading; the value is peak memory allocated to fragment
+    final MinorFragmentProfile maxMem = Collections.max(complete, Comparators.fragmentPeakMemory);
+    tb.appendBytes(maxMem.getMaxMemoryUsed());
+  }
+
+  public static final String[] COMPLETED_FRAGMENT_OVERVIEW_COLUMNS = {
+      OverviewTblTxt.MAJOR_FRAGMENT, OverviewTblTxt.MINOR_FRAGMENTS_REPORTING,
+      OverviewTblTxt.FIRST_START, OverviewTblTxt.LAST_START, OverviewTblTxt.FIRST_END, OverviewTblTxt.LAST_END,
+      OverviewTblTxt.MIN_RUNTIME, OverviewTblTxt.AVG_RUNTIME, OverviewTblTxt.MAX_RUNTIME,
+      OverviewTblTxt.PERCENT_BUSY, OverviewTblTxt.MAX_PEAK_MEMORY
+  };
+
+  public static final String[] COMPLETED_FRAGMENT_OVERVIEW_COLUMNS_TOOLTIP = {
+      OverviewTblTooltip.MAJOR_FRAGMENT, OverviewTblTooltip.MINOR_FRAGMENTS_REPORTING,
+      OverviewTblTooltip.FIRST_START, OverviewTblTooltip.LAST_START, OverviewTblTooltip.FIRST_END, OverviewTblTooltip.LAST_END,
+      OverviewTblTooltip.MIN_RUNTIME, OverviewTblTooltip.AVG_RUNTIME, OverviewTblTooltip.MAX_RUNTIME,
+      OverviewTblTooltip.PERCENT_BUSY, OverviewTblTooltip.MAX_PEAK_MEMORY
+  };
+
+  //Not including Major Fragment ID and Minor Fragments Reporting
+  public static final int NUM_NULLABLE_COMPLETED_OVERVIEW_COLUMNS = COMPLETED_FRAGMENT_OVERVIEW_COLUMNS.length - 2;
+
+  public void addFinalSummary(TableBuilder tb) {
+
+    // Use only minor fragments that have complete profiles
+    // Complete iff the fragment profile has at least one operator profile, and start and end times.
+    final List<MinorFragmentProfile> complete = new ArrayList<>(
+        Collections2.filter(major.getMinorFragmentProfileList(), Filters.hasOperatorsAndTimes));
+
+    tb.appendCell(new OperatorPathBuilder().setMajor(major).build());
+    tb.appendCell(complete.size() + " / " + major.getMinorFragmentProfileCount());
+
+    // If there are no stats to aggregate, create an empty row
+    if (complete.size() < 1) {
+      tb.appendRepeated("", null, NUM_NULLABLE_COMPLETED_OVERVIEW_COLUMNS);
+      return;
+    }
+
+    final MinorFragmentProfile firstStart = Collections.min(complete, Comparators.startTime);
+    final MinorFragmentProfile lastStart = Collections.max(complete, Comparators.startTime);
+    tb.appendMillis(firstStart.getStartTime() - start);
+    tb.appendMillis(lastStart.getStartTime() - start);
+
+    final MinorFragmentProfile firstEnd = Collections.min(complete, Comparators.endTime);
+    final MinorFragmentProfile lastEnd = Collections.max(complete, Comparators.endTime);
+    tb.appendMillis(firstEnd.getEndTime() - start);
+    tb.appendMillis(lastEnd.getEndTime() - start);
+
+    long totalDuration = 0L;
+    double totalProcessInMillis = 0.0d;
+    double totalWaitInMillis = 0.0d;
+    for (final MinorFragmentProfile p : complete) {
+      totalDuration += p.getEndTime() - p.getStartTime();
+      //Capture Busy & Wait Time
+      List<OperatorProfile> opProfileList = p.getOperatorProfileList();
+      for (OperatorProfile operatorProfile : opProfileList) {
+        totalProcessInMillis += operatorProfile.getProcessNanos()/1E6;
+        totalWaitInMillis += operatorProfile.getWaitNanos()/1E6;
+      }
     }
 
     final MinorFragmentProfile shortRun = Collections.min(complete, Comparators.runTime);
     final MinorFragmentProfile longRun = Collections.max(complete, Comparators.runTime);
-    tb.appendMillis(shortRun.getEndTime() - shortRun.getStartTime(), null);
-    tb.appendMillis(total / complete.size(), null);
-    tb.appendMillis(longRun.getEndTime() - longRun.getStartTime(), null);
+    tb.appendMillis(shortRun.getEndTime() - shortRun.getStartTime());
+    tb.appendMillis(totalDuration / complete.size());
+    tb.appendMillis(longRun.getEndTime() - longRun.getStartTime());
 
-    final MinorFragmentProfile lastUpdate = Collections.max(complete, Comparators.lastUpdate);
-    tb.appendTime(lastUpdate.getLastUpdate(), null);
-
-    final MinorFragmentProfile lastProgress = Collections.max(complete, Comparators.lastProgress);
-    tb.appendTime(lastProgress.getLastProgress(), null);
+    tb.appendPercent(totalProcessInMillis / (totalProcessInMillis + totalWaitInMillis), null,
+        //#8721 is the summation sign: sum(Busy): ## + sum(Wait): ##
+        String.format("&#8721;Busy: %,.2fs + &#8721;Wait: %,.2fs", totalProcessInMillis/1E3, totalWaitInMillis/1E3));
 
     // TODO(DRILL-3494): Names (maxMem, getMaxMemoryUsed) are misleading; the value is peak memory allocated to fragment
     final MinorFragmentProfile maxMem = Collections.max(complete, Comparators.fragmentPeakMemory);
-    tb.appendBytes(maxMem.getMaxMemoryUsed(), null);
+    tb.appendBytes(maxMem.getMaxMemoryUsed());
   }
 
   public static final String[] FRAGMENT_COLUMNS = {"Minor Fragment ID", "Host Name", "Start", "End",
-    "Runtime", "Max Records", "Max Batches", "Last Update", "Last Progress", "Peak Memory", "State"};
+      "Runtime", "Max Records", "Max Batches", "Last Update", "Last Progress", "Peak Memory", "State"};
 
   // Not including minor fragment ID
   private static final int NUM_NULLABLE_FRAGMENTS_COLUMNS = FRAGMENT_COLUMNS.length - 1;
 
   public String getContent() {
-    final TableBuilder builder = new TableBuilder(FRAGMENT_COLUMNS);
+    final TableBuilder builder = new TableBuilder(FRAGMENT_COLUMNS, null);
 
     // Use only minor fragments that have complete profiles
     // Complete iff the fragment profile has at least one operator profile, and start and end times.
     final List<MinorFragmentProfile> complete = new ArrayList<>(
-      Collections2.filter(major.getMinorFragmentProfileList(), Filters.hasOperatorsAndTimes));
+        Collections2.filter(major.getMinorFragmentProfileList(), Filters.hasOperatorsAndTimes));
     final List<MinorFragmentProfile> incomplete = new ArrayList<>(
-      Collections2.filter(major.getMinorFragmentProfileList(), Filters.missingOperatorsOrTimes));
+        Collections2.filter(major.getMinorFragmentProfileList(), Filters.missingOperatorsOrTimes));
 
     Collections.sort(complete, Comparators.minorId);
     for (final MinorFragmentProfile minor : complete) {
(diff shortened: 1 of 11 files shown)


kkhatua:
@sudheeshkatkam This commit is rebased on top of another pending PR (#738): DRILL-5190. You'll need to apply that before applying this.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -33,62 +33,91 @@
...
+      sb.append(String.format("<td>%s%s</td>", s, link != null ? link : ""));
+    }
     if (++w >= width) {
       sb.append("</tr>\n");
       w = 0;
+    }  }
+
+  public void appendCellWithBgColor(final String s, final String link, final String titleText, final String backgroundColor) {
+    if (w == 0) {
+      sb.append("<tr bgcolor=\""+backgroundColor+"\">");
```
paul-rogers: This is very old-school. Modern web practice is to define a CSS style, then reference that here:

```
highlight: { background: #aabbcc; }
...
<tr class="highlight"> ...
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -33,62 +33,91 @@
...
     }
-  }
+    if (titleText != null && titleText.length() > 0) {
+      sb.append(String.format("<td title=\""+titleText+"\">%s%s</td>", s, link != null ? link : ""));
+    } else {
+      sb.append(String.format("<td>%s%s</td>", s, link != null ? link : ""));
```
paul-rogers: Seems odd. If this is truly a link, you need the following;

```
<a href="link url">title text</a>
```

Maybe the "link" is something else?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -33,62 +33,91 @@
...
+  public void appendCell(final String s, final String link, final String titleText) {
     if (w == 0) {
       sb.append("<tr>");
     }
-    sb.append(String.format("<td>%s%s</td>", s, link != null ? link : ""));
+    if (titleText != null && titleText.length() > 0) {
+      sb.append(String.format("<td title=\""+titleText+"\">%s%s</td>", s, link != null ? link : ""));
+    } else {
+      sb.append(String.format("<td>%s%s</td>", s, link != null ? link : ""));
+    }
```
paul-rogers: The above code appears twice. Please avoid copy & paste. Move the logic to a method called both places.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -136,25 +215,25 @@ public String getContent() {
...
 
-      builder.appendTime(minor.getLastUpdate(), null);
-      builder.appendTime(minor.getLastProgress(), null);
+      builder.appendTime(minor.getLastUpdate(), null, null);
+      builder.appendTime(minor.getLastProgress(), null, null);
 
-      builder.appendBytes(minor.getMaxMemoryUsed(), null);
-      builder.appendCell(minor.getState().name(), null);
+      builder.appendBytes(minor.getMaxMemoryUsed(), null, null);
+      builder.appendCell(minor.getState().name(), null, null);
```
paul-rogers: All these extra ", null" parameters (for tool tip) are a clutter. The clean way to do this is:

```
void appendTime(long ts) {
    appendTime(ts, null, null);
}
void appendTime(long ts, String link) {
    appendTime(ts, link, null)
}
void appendTime(long ts, String link, String toolTip) {
   ... // Do stuff here
}
```

And so on.

Then, the above lines are just:

```
builder.appendBytes(minor.getMaxMemoryUsed())
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
```diff
@@ -33,62 +33,91 @@
...
     sb.append("<table class=\"table table-bordered text-right\">\n<tr>");
-    for (final String cn : columns) {
-      sb.append("<th>" + cn + "</th>");
+    for (int i = 0; i < columns.length; i++) {
+      String cn = columns[i];
+      String ctt = "";
+      if (columnTooltip != null) {
+        String tooltip = columnTooltip[i];
+        if (tooltip != null) {
+          ctt = " title=\""+tooltip+"\" style=\"color:#00008B;\"";
```
paul-rogers: Color should appear in a style sheet. See below.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -107,6 +107,9 @@
   <p>FOREMAN: ${model.getProfile().getForeman().getAddress()}</p>
   <p>TOTAL FRAGMENTS: ${model.getProfile().getTotalFragments()}</p>
   <p>DURATION: ${model.getProfileDuration()}</p>
+  <p>&nbsp;&nbsp;&nbsp;&nbsp;PLANNING: ${model.getPlanningDuration()}</p>
+  <p>&nbsp;&nbsp;&nbsp;&nbsp;QUEUED: ${model.getQueuedDuration()}</p>
+  <p>&nbsp;&nbsp;&nbsp;&nbsp;EXECUTION: ${model.getExecutionDuration()}</p>
```
paul-rogers: Thanks for the indent. But, using nbsp is old-school. Please create a style that sets the left margin to, say 4em.

Also, as long as we are here, might as well clean up the labels. Shift from UPPER CASE to Title Case.

And, for extra credit, change the whole mess to a table so that the values are all aligned in a nice column.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
     return String.format("fragment-%s", major.getMajorFragmentId());
   }
 
-  public static final String[] FRAGMENT_OVERVIEW_COLUMNS = {"Major Fragment", "Minor Fragments Reporting",
-    "First Start", "Last Start", "First End", "Last End", "Min Runtime", "Avg Runtime", "Max Runtime", "Last Update",
-    "Last Progress", "Max Peak Memory"};
+  public static final String[] ACTIVE_FRAGMENT_OVERVIEW_COLUMNS = {"Major Fragment", "Minor Fragments Reporting",
+    "First Start", "Last Start", "First End", "Last End", "Min Runtime", "Avg Runtime", "Max Runtime", "% Busy",
```
kkhatua: Not specifically for this, as it affects the Fragment Overview table. Which parts do you think need a reword?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+    "Last Update", "Last Progress", "Max Peak Memory"};
+
+  public static final String[] ACTIVE_FRAGMENT_OVERVIEW_COLUMNS_TOOLTIP = {null, "# Minor Fragments Spawned",
```
kkhatua: Agreed. Was trying to save real estate :)

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+    double totalProcessInMillis = 0.0d;
+    double totalWaitInMillis = 0.0d;
     for (final MinorFragmentProfile p : complete) {
-      total += p.getEndTime() - p.getStartTime();
+      totalDuration += p.getEndTime() - p.getStartTime();
+      //Capture Busy & Wait Time
+      List<OperatorProfile> opProfileList = p.getOperatorProfileList();
+      for (OperatorProfile operatorProfile : opProfileList) {
+        totalProcessInMillis += operatorProfile.getProcessNanos()/1E6;
+        totalWaitInMillis += operatorProfile.getWaitNanos()/1E6;
```
kkhatua: Wouldn't Rounding actually reduce accuracy as I accumulate values?
I was worried of a possible overflow (in the event of a large number of long running fragments) if I simply accumulated all the values in nanosec instead of millisec. Hence the conversion. 
But I see your point. Rounding wouldn't hurt because as we sum, the values would be relatively small. I'll apply this suggestion.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
-  public static final String[] FRAGMENT_OVERVIEW_COLUMNS = {"Major Fragment", "Minor Fragments Reporting",
-    "First Start", "Last Start", "First End", "Last End", "Min Runtime", "Avg Runtime", "Max Runtime", "Last Update",
-    "Last Progress", "Max Peak Memory"};
+  public static final String[] ACTIVE_FRAGMENT_OVERVIEW_COLUMNS = {"Major Fragment", "Minor Fragments Reporting",
+    "First Start", "Last Start", "First End", "Last End", "Min Runtime", "Avg Runtime", "Max Runtime", "% Busy",
+    "Last Update", "Last Progress", "Max Peak Memory"};
+
+  public static final String[] ACTIVE_FRAGMENT_OVERVIEW_COLUMNS_TOOLTIP = {null, "# Minor Fragments Spawned",
+    null, null, null, null, "Shortest duration of a fragment", "Avg duration of a fragment", "Longest duration of a fragment", "%time Fragments were Busy",
+    "Last time a running fragment's status was updated", "Last time we heard from a running fragment", null};
```
kkhatua: Should I retain the 'darkblue' text color change for columns, if they all eventually have tooltips??

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+      }
     }
 
     final MinorFragmentProfile shortRun = Collections.min(complete, Comparators.runTime);
     final MinorFragmentProfile longRun = Collections.max(complete, Comparators.runTime);
-    tb.appendMillis(shortRun.getEndTime() - shortRun.getStartTime(), null);
-    tb.appendMillis(total / complete.size(), null);
-    tb.appendMillis(longRun.getEndTime() - longRun.getStartTime(), null);
+    tb.appendMillis(shortRun.getEndTime() - shortRun.getStartTime(), null, null);
+    tb.appendMillis(totalDuration / complete.size(), null, null);
```
kkhatua: Nope. We return early in the method if we encounter `complete.size() == 0`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+    tb.appendMillis(longRun.getEndTime() - longRun.getStartTime(), null, null);
+
+    tb.appendPercent(totalProcessInMillis / (totalProcessInMillis + totalWaitInMillis), null,
```
kkhatua: Nope. We return early in the method if we encounter `complete.size() == 0`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+      String.format("&#8721;Busy: %,.2fs + &#8721;Wait: %,.2fs", totalProcessInMillis/1E3, totalWaitInMillis/1E3));
```
kkhatua: This is a tooltip to indicate that the %busy is based on the total time consisting of sum(busyNanos)+sum(waitNanos). Just a reference on how we are arriving at that percentage. I'll add the comment in the code.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+    tb.appendMillis(firstEnd.getEndTime() - start, null, null);
+    tb.appendMillis(lastEnd.getEndTime() - start, null, null);
 
-    long total = 0;
+    long totalDuration = 0L;
+    double totalProcessInMillis = 0.0d;
+    double totalWaitInMillis = 0.0d;
     for (final MinorFragmentProfile p : complete) {
-      total += p.getEndTime() - p.getStartTime();
+      totalDuration += p.getEndTime() - p.getStartTime();
```
kkhatua: This is actually measuring the 'life span' of the MinorFragmentProfile. It would be inclusive of the process and wait time, but (possibly) also more. We report the average fragment runtime on this.
I'll change it to cumulativeFragmentDuration as a more accurate representation.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+    tb.appendCell(new OperatorPathBuilder().setMajor(major).build(), null, null);
+    tb.appendCell(complete.size() + " / " + major.getMinorFragmentProfileCount(), null, null);
+
+    // If there are no stats to aggregate, create an empty row
+    if (complete.size() < 1) {
+      tb.appendRepeated("", null, NUM_NULLABLE_COMPLETED_OVERVIEW_COLUMNS, null);
+      return;
+    }
+
+    final MinorFragmentProfile firstStart = Collections.min(complete, Comparators.startTime);
```
kkhatua: Yes. This looks at only completed fragments (i.e. we know their start and end times).

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+      List<OperatorProfile> opProfileList = p.getOperatorProfileList();
+      for (OperatorProfile operatorProfile : opProfileList) {
+        totalProcessInMillis += operatorProfile.getProcessNanos()/1E6;
+        totalWaitInMillis += operatorProfile.getWaitNanos()/1E6;
+      }
+    }
+
+    final MinorFragmentProfile shortRun = Collections.min(complete, Comparators.runTime);
+    final MinorFragmentProfile longRun = Collections.max(complete, Comparators.runTime);
+    tb.appendMillis(shortRun.getEndTime() - shortRun.getStartTime(), null, null);
```
kkhatua: The code creates two sets and we construct our tables primarily on the collection of 'completed' fragments.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
kkhatua: +1 for Succinct way to convey what the function call is. I'll incorporate additional methods.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
kkhatua: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
kkhatua: Since I'll be adding tooltips for all the columns in the 2 overview tables, I'll do away with the style

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
kkhatua: I think it was being misused as a way to show the fragment ID. I'm not sure what purpose does it serve to have two strings concatenated with `Strong.format("%s%s", text1, text2)`
Didn't want to break anything, so we only inject the `title` (our tooltip) when it is available.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java
kkhatua: Just keeping it simple. Added this so that we can create alternating bands where needed. 
The `profile.ftl` file also uses in-place style formatting in a lot of places.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -83,56 +84,77 @@ public String getContent() {
...
   }
 
-  public static final String[] OPERATORS_OVERVIEW_COLUMNS = {"Operator ID", "Type", "Min Setup Time", "Avg Setup Time",
-    "Max Setup Time", "Min Process Time", "Avg Process Time", "Max Process Time", "Min Wait Time", "Avg Wait Time",
-    "Max Wait Time", "Avg Peak Memory", "Max Peak Memory"};
+  public static final String[] OPERATORS_OVERVIEW_COLUMNS = {"Operator ID", "Type",
+    "Avg Setup Time", "Max Setup Time",
+    "Avg Process Time", "Max Process Time", "Min Wait Time", "Avg Wait Time", "Max Wait Time",
+    "% Fragment Time", "% Query Time", "Rows",
+    "Avg Peak Memory", "Max Peak Memory"};
```
kkhatua: Not sure if we need to make a finalized set of string, since the array itself is finalized. Only in case of the Fragment Overview, would it make sense since the same text is used in two places.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -83,56 +84,77 @@ public String getContent() {
...
 
-  public void addSummary(TableBuilder tb) {
+  public static final String[] OPERATORS_OVERVIEW_COLUMNS_TOOLTIP = {null, null,
+    null, null, null, null, null, null, null,
+    "% of time that the major fragment spent in operator", "% of Query time spent in operator", "Records emitted by the operator",
```
kkhatua: Rows emitted would be correct. For all the operators, this is a tally of the outgoing rows and not the rows processed. 
Changed text for the rest.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -83,56 +84,77 @@ public String getContent() {
...
     final ImmutablePair<OperatorProfile, Integer> longWait = Collections.max(ops, Comparators.waitTime);
-    tb.appendNanos(shortWait.getLeft().getWaitNanos(), null);
-    tb.appendNanos(Math.round(waitSum / size), null);
-    tb.appendNanos(longWait.getLeft().getWaitNanos(), null);
+    tb.appendNanos(shortWait.getLeft().getWaitNanos(), null, null);
+    tb.appendNanos(Math.round(waitSum / size), null, null);
+    tb.appendNanos(longWait.getLeft().getWaitNanos(), null, null);
+
+    tb.appendPercent(processSum / majorBusyNanos, null, String.format(
+        "&#8721;OperatorBusy: %,.2fs / &#8721;MajorBusy: %,.2fs", processSum/1E9, majorBusyNanos/1E9));
```
kkhatua: This was just to quantify the numbers, but it probably isn't needed any more. I think we can remove this.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+    // TODO(DRILL-3494): Names (maxMem, getMaxMemoryUsed) are misleading; the value is peak memory allocated to fragment
+    final MinorFragmentProfile maxMem = Collections.max(complete, Comparators.fragmentPeakMemory);
+    tb.appendBytes(maxMem.getMaxMemoryUsed(), null, null);
+  }
+
+  public static final String[] COMPLETED_FRAGMENT_OVERVIEW_COLUMNS = {"Major Fragment", "Minor Fragments Reporting",
+    "First Start", "Last Start", "First End", "Last End", "Min Runtime", "Avg Runtime", "Max Runtime", "% Busy", "Max Peak Memory"};
+
+  public static final String[] COMPLETED_FRAGMENT_OVERVIEW_COLUMNS_TOOLTIP = {null, "# Minor Fragments Spawned",
+    null, null, null, null, "Shortest duration of a fragment", "Avg duration of a fragment", "Longest duration of a fragment", "%time Fragments were Busy", null};
```
kkhatua: Ok. I see what you meant by the very first comment in your review. Will change these to using String constants.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -49,58 +51,135 @@ public String getId() {
...
+    tb.appendMillis(shortRun.getEndTime() - shortRun.getStartTime(), null, null);
+    tb.appendMillis(totalDuration / complete.size(), null, null);
+    tb.appendMillis(longRun.getEndTime() - longRun.getStartTime(), null, null);
+
+    tb.appendPercent(totalProcessInMillis / (totalProcessInMillis + totalWaitInMillis), null,
+      String.format("&#8721;Busy: %,.2fs + &#8721;Wait: %,.2fs", totalProcessInMillis/1E3, totalWaitInMillis/1E3));
 
     final MinorFragmentProfile lastUpdate = Collections.max(complete, Comparators.lastUpdate);
-    tb.appendTime(lastUpdate.getLastUpdate(), null);
+    tb.appendMillis(System.currentTimeMillis()-lastUpdate.getLastUpdate(), null, new SimpleDateFormat().format(new Date(lastUpdate.getLastUpdate())));
```
kkhatua: This is actually a tooltip, to indicate the time on the clock when the last update/progress was made. Since we already have a delta, this might not be needed. The only time I can imagine is when I'd want to refer to that to go through the logs to narrow down on the cause of a slow-down/hang in a query. 
Do you think that would help, or should I just delete the tooltip, @paul-rogers ?

kkhatua:
@paul-rogers , checked in the final set of changes, accommodating most of your comments. 
TPCH Samples: 
http://10.10.100.127:8047/profiles/275020fe-f51c-b8f3-a7b4-9f54461c4606
http://10.10.100.127:8047/profiles/275021b9-c87e-efd9-135e-4172ef2d38b6

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -179,12 +207,47 @@ public String getMetricsTable() {
...
-          builder.appendCell("", null);
+          builder.appendCell("");
         }
       }
     }
     return builder.build();
   }
+
+  private class OverviewTblTxt {
+    static final String OperatorID = "Operator ID";
```
arina-ielchiieva: Is there any reason why variables start with upper case letter instead of lower case (ex: operatorID)?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -35,6 +38,10 @@
 public class FragmentWrapper {
   private final MajorFragmentProfile major;
   private final long start;
+  private final Locale currentLocale = Locale.getDefault();
+  private final String pattern = "dd-MMM-yyyy HH:mm:ss";
+  private final SimpleDateFormat simpleDateFormat = new SimpleDateFormat(
```
paul-rogers: The following is the same:
```
new SimpleDateFormat(pattern);
```

Note that the locale info *does not* adjust the time zone. Local time is quite clunky with the original Java classes; we need to use the Joda classes to better handle local date/times.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -136,26 +240,60 @@ public String getContent() {
...
     for (final MinorFragmentProfile m : incomplete) {
-      builder.appendCell(major.getMajorFragmentId() + "-" + m.getMinorFragmentId(), null);
+      builder.appendCell(major.getMajorFragmentId() + "-" + m.getMinorFragmentId());
       builder.appendRepeated(m.getState().toString(), null, NUM_NULLABLE_FRAGMENTS_COLUMNS);
     }
     return builder.build();
   }
+
+  private class OverviewTblTxt {
+    static final String MajorFragment = "Major Fragment";
```
paul-rogers: Java conventions are to use UPPER_CASE for constants.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -136,26 +240,60 @@ public String getContent() {
...
 
     for (final MinorFragmentProfile m : incomplete) {
-      builder.appendCell(major.getMajorFragmentId() + "-" + m.getMinorFragmentId(), null);
+      builder.appendCell(major.getMajorFragmentId() + "-" + m.getMinorFragmentId());
       builder.appendRepeated(m.getState().toString(), null, NUM_NULLABLE_FRAGMENTS_COLUMNS);
     }
     return builder.build();
   }
+
+  private class OverviewTblTxt {
```
paul-rogers: Thanks for adding the explanations!

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -136,26 +240,60 @@ public String getContent() {
...
+    static final String MaxRuntime = "Max Runtime";
+    static final String PercentBusy = "% Busy";
+    static final String LastUpdate = "Last Update";
+    static final String LastProgress = "Last Progress";
+    static final String MaxPeakMemory = "Max Peak Memory";
+  }
+
+  private class OverviewTblTooltip {
+    static final String MajorFragment = "Major Fragment ID";
+    static final String MinorFragmentsReporting = "Number of Minor Fragments Spawned";
```
paul-rogers: This is descriptive text, so use sentence case: "Number of minor fragments spawned". Maybe "spawened" --> "started".

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -136,26 +240,60 @@ public String getContent() {
...
+    static final String AvgRuntime = "Avg Runtime";
+    static final String MaxRuntime = "Max Runtime";
+    static final String PercentBusy = "% Busy";
+    static final String LastUpdate = "Last Update";
+    static final String LastProgress = "Last Progress";
+    static final String MaxPeakMemory = "Max Peak Memory";
+  }
+
+  private class OverviewTblTooltip {
+    static final String MajorFragment = "Major Fragment ID";
```
paul-rogers: Maybe: "Major fragment ID as shown in the visualization"

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -136,26 +240,60 @@ public String getContent() {
...
+    static final String MinorFragmentsReporting = "Number of Minor Fragments Spawned";
+    static final String FirstStart = "Earliest start of a fragment since query submission";
+    static final String LastStart = "Latest start of a fragment since query submission";
+    static final String FirstEnd = "Earliest completion time a fragment";
```
paul-rogers: "time a" --> "time of a".

Or, better, "Time at which the first fragment completed."

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -136,26 +240,60 @@ public String getContent() {
...
+    static final String LastEnd = "Latest completion time of a fragment";
+    static final String MinRuntime = "Shortest fragment runtime";
+    static final String AvgRuntime = "Average fragment runtime";
+    static final String MaxRuntime = "Longest fragment runtime";
+    static final String PercentBusy = "Percent time Fragments were busy doing work";
```
paul-rogers: "Percentage of run time that fragments". Note "Fragments" --> "fragments".

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
paul-rogers: These are constants. So, Java convention is upper case: OPERATOR_ID.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -179,12 +207,47 @@ public String getMetricsTable() {
...
+    static final String PercentQueryTime = "% Query Time";
+    static final String Rows = "Rows";
+    static final String AvgPeakMemory = "Avg Peak Memory";
+    static final String MaxPeakMemory = "Max Peak Memory";
+  }
+
+  private class OverviewTblTooltip {
+    static final String OperatorID = "Operator ID";
+    static final String Type = "Operator Type";
+    static final String AvgSetupTime = "Average Time in setting up fragments";
```
paul-rogers: Sentence case: "Time" --> "time"

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -179,12 +207,47 @@ public String getMetricsTable() {
...
+    static final String MaxSetupTime = "Longest Time a fragment took in setup";
```
paul-rogers: "Time" --> "time"

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -179,12 +207,47 @@ public String getMetricsTable() {
...
+    static final String AvgProcessTime = "Average processing time for a fragment";
+    static final String MaxProcessTime = "Longest Time a fragment took to process";
```
paul-rogers: "Time" --> "time"

Better, "Longest process time of any fragment"

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -179,12 +207,47 @@ public String getMetricsTable() {
...
+    static final String MinWaitTime = "Shortest time a fragment spent in waiting for data";
```
paul-rogers: Waits can be for any reason, so "Shortest time a fragment spent waiting"

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -179,12 +207,47 @@ public String getMetricsTable() {
...
+    static final String AvgWaitTime = "Average wait time for a fragment";
+    static final String MaxWaitTime = "Longest Time a fragment spent in waiting data";
```
paul-rogers: "Time" --> "time"

"waiting data" --> "waiting for data" or just "waiting"

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -179,12 +207,47 @@ public String getMetricsTable() {
...
+    static final String PercentFragmentTime = "Percentage of the total fragment time that was spent on the operator";
+    static final String PercentQueryTime = "Percentage of the total query time that was spent on the operator";
+    static final String Rows = "Rows emitted by the operator";
```
paul-rogers: "Rows emitted by scans, or consumed by other operators."

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
kkhatua: Oh. I thought the locale did that. But, as long as these timestamps are aligned with the timestamps on the server (where the logs reside), this should be fine, right? Otherwise, this is useful for only _running_ queries and we can skip it altogether. What do you suggest, @paul-rogers ?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
kkhatua: I'll change it to the Java convention. It was getting to be jarring to see everything in Upper case when defining the columns. Should follow the convention. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
kkhatua: +1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
kkhatua: +1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
kkhatua: +1 Applied changes to all.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
kkhatua: +1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
kkhatua: +1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
kkhatua: Rows emitted by scans is incomplete. Operators like Filter, HashAgg, etc will also change the number of outgoing rows wrt the number of incoming rows. 
I think we should leave this as is 'Rows emitter by the operator'

kkhatua:
@paul-rogers 
Incorporated most changes... except:
https://github.com/apache/drill/pull/756#discussion_r103000211
https://github.com/apache/drill/pull/756#discussion_r103005037

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
kkhatua: Went through some profiles in more detail. You were right about the row counts. HashAgg and HashJoins mention the rows processed. The following PROJECT operators indicate what the actual rowcounts were. 
I'll just revert to what you suggested.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
kkhatua: Dropping this as it's just a tooltip

kkhatua:
@paul-rogers Incorporated all changes.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
paul-rogers: Actually, accumulate in nanos, but round the total nanos to get ms. I probably missed the aggregation detail.

PR #757, opened 2017-02-22
DRILL-5290: Provide an option to build operator table once for built-…

ppadma:
…in static functions and reuse it across queries.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -413,4 +413,8 @@ public interface ExecConstants {
 
   String DYNAMIC_UDF_SUPPORT_ENABLED = "exec.udf.enable_dynamic_support";
   BooleanValidator DYNAMIC_UDF_SUPPORT_ENABLED_VALIDATOR = new BooleanValidator(DYNAMIC_UDF_SUPPORT_ENABLED, true, true);
+
+  String USE_DYNAMIC_UDFS_KEY = "exec.udf.use_dynamic";
+  BooleanValidator USE_DYNAMIC_UDFS = new BooleanValidator(USE_DYNAMIC_UDFS_KEY, true);
+
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
@@ -91,7 +91,14 @@ public class QueryContext implements AutoCloseable, OptimizerRulesContext, Schem
     executionControls = new ExecutionControls(queryOptions, drillbitContext.getEndpoint());
     plannerSettings = new PlannerSettings(queryOptions, getFunctionRegistry());
     plannerSettings.setNumEndPoints(drillbitContext.getBits().size());
-    table = new DrillOperatorTable(getFunctionRegistry(), drillbitContext.getOptionManager());
+
+    // If we do not need to support dynamic UDFs for this query, just use static operator table
+    // built at the startup. Else, build new operator table from latest version of function registry.
+    if (queryOptions.getOption(ExecConstants.USE_DYNAMIC_UDFS)) {
+      this.table = new DrillOperatorTable(drillbitContext.getFunctionImplementationRegistry(), drillbitContext.getOptionManager());
+    } else {
+      this.table = drillbitContext.getOperatorTable();
+    }
 
     queryContextInfo = Utilities.createQueryContextInfo(session.getDefaultSchemaPath(), session.getSessionId());
     contextInformation = new ContextInformation(session.getCredentials(), queryContextInfo);
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitContext.java
@@ -33,6 +33,7 @@ import org.apache.drill.exec.expr.fn.registry.RemoteFunctionRegistry;
 import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.physical.impl.OperatorCreatorRegistry;
 import org.apache.drill.exec.planner.PhysicalPlanReader;
+import org.apache.drill.exec.planner.sql.DrillOperatorTable;
 import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
 import org.apache.drill.exec.rpc.control.Controller;
 import org.apache.drill.exec.rpc.control.WorkEventBus;
@@ -62,6 +63,8 @@ public class DrillbitContext implements AutoCloseable {
   private final CodeCompiler compiler;
   private final ScanResult classpathScan;
   private final LogicalPlanPersistence lpPersistence;
+  // operator table for standard SQL operators and functions, Drill built-in UDFs
+  private final DrillOperatorTable table;
 
 
   public DrillbitContext(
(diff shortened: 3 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -91,7 +91,12 @@ public QueryContext(final UserSession session, final DrillbitContext drillbitCon
     executionControls = new ExecutionControls(queryOptions, drillbitContext.getEndpoint());
     plannerSettings = new PlannerSettings(queryOptions, getFunctionRegistry());
     plannerSettings.setNumEndPoints(drillbitContext.getBits().size());
-    table = new DrillOperatorTable(getFunctionRegistry(), drillbitContext.getOptionManager());
+
+    if (getOption(ExecConstants.USE_DYNAMIC_UDFS).bool_val) {
+      table = new DrillOperatorTable(getFunctionRegistry(), drillbitContext.getOptionManager());
```
sudheeshkatkam: Avoid calling methods in this class in ctor. Use

`this.table = new DrillOperatorTable(drillbitContext.getFunctionImplementationRegistry(), drillbitContext.getOptionManager());`

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -413,4 +413,8 @@
 
   String DYNAMIC_UDF_SUPPORT_ENABLED = "exec.udf.enable_dynamic_support";
   BooleanValidator DYNAMIC_UDF_SUPPORT_ENABLED_VALIDATOR = new BooleanValidator(DYNAMIC_UDF_SUPPORT_ENABLED, true, true);
+
+  String USE_DYNAMIC_UDFS = "exec.udf.use_dynamic";
```
sudheeshkatkam: + Looks like the above option "enables" dynamic UDFs, and this enables "using" dynamic UDFs. So why is this option required?
+ If this is intended to be a session option, and the above one is not, then make a check to ensure either of the options are enabled in QueryContext.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -91,7 +91,12 @@ public QueryContext(final UserSession session, final DrillbitContext drillbitCon
     executionControls = new ExecutionControls(queryOptions, drillbitContext.getEndpoint());
     plannerSettings = new PlannerSettings(queryOptions, getFunctionRegistry());
     plannerSettings.setNumEndPoints(drillbitContext.getBits().size());
-    table = new DrillOperatorTable(getFunctionRegistry(), drillbitContext.getOptionManager());
+
+    if (getOption(ExecConstants.USE_DYNAMIC_UDFS).bool_val) {
```
sudheeshkatkam: + `getOption` uses queryOptions. Is that intended?
+ If so, avoid calling methods in this class in ctor. Use `queryOptions.getOption(...)`
+ Change declaration to `BooleanValidator USE_DYNAMIC_UDFS_VALIDATOR ...` to avoid `.bool_val`

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
ppadma: I did not use the existing option "exec.udf.enable_dynamic_support" because if that option is enabled and then disabled later, expectation is that dynamic UDFs added in that window continue to be available and working even after disabling.  

All other comments addressed. Please review new diffs.

sudheeshkatkam: Hmm, how does recreating the table achieve that? In both contextual objects, the input params to the constructor are same.

ppadma: Currently, we have one FunctionRegistryHolder (LocalFunctionRegistry) for both static and dynamic functions, from which we register for each query.  It gets updated when we download new jars/functions from zookeeper and will not be same as what it is during startup if dynamic UDF support is enabled and disabled.  That means if I use table I built during startup, it will miss the dynamic UDFs that got added in between and to include them, you have to rebuild the table.  For that reason, I choose to  add a new option.

exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitContext.java
```diff
@@ -189,6 +192,10 @@ public ScanResult getClasspathScan() {
 
   public RemoteFunctionRegistry getRemoteFunctionRegistry() { return functionRegistry.getRemoteFunctionRegistry(); }
 
+  public DrillOperatorTable getOperatorTable() {
```
arina-ielchiieva: Please add java doc explaining reasons why we are caching operator table and pointing out to the USE_DYNAMIC_UDFS_KEY which triggers cached operator table usage.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
sudheeshkatkam: Why not update the table in DrillbitContext when the registry is updated, rather than rebuild the table?

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -413,4 +413,8 @@
 
   String DYNAMIC_UDF_SUPPORT_ENABLED = "exec.udf.enable_dynamic_support";
   BooleanValidator DYNAMIC_UDF_SUPPORT_ENABLED_VALIDATOR = new BooleanValidator(DYNAMIC_UDF_SUPPORT_ENABLED, true, true);
+
+  String USE_DYNAMIC_UDFS_KEY = "exec.udf.use_dynamic";
```
paul-rogers: ENABLE_DYNAMIC_UDF and exec.udf.enable_dynamic?

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -91,7 +91,12 @@ public QueryContext(final UserSession session, final DrillbitContext drillbitCon
     executionControls = new ExecutionControls(queryOptions, drillbitContext.getEndpoint());
     plannerSettings = new PlannerSettings(queryOptions, getFunctionRegistry());
     plannerSettings.setNumEndPoints(drillbitContext.getBits().size());
-    table = new DrillOperatorTable(getFunctionRegistry(), drillbitContext.getOptionManager());
+
+    if (queryOptions.getOption(ExecConstants.USE_DYNAMIC_UDFS)) {
```
paul-rogers: Maybe a comment to explain what we're doing here?

exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitContext.java
```diff
@@ -62,6 +63,7 @@
   private final CodeCompiler compiler;
   private final ScanResult classpathScan;
   private final LogicalPlanPersistence lpPersistence;
+  private final DrillOperatorTable table;
```
paul-rogers: Comment to explain that this is for static, built-in operators, not dynamic UDFs?

ppadma:
Thanks Sudheesh, Paul and Arina for the review.  Original thought was it is not worth the effort to keep operator table in sync with changes in dynamic UDFs. This option is meant to be used for short running operational queries, which most likely will not need dynamic UDF support (well, at least for now). For long running analytic queries, this is not considered an issue. 
However, I am looking into implementing Sudheesh's suggestion. I will provide an update soon.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
ppadma: ok,  we need to use readWriteLocks if we update the table each time a function gets added/removed. That is unnecessary overhead and will cause contention with concurrency. One option is to split the table into two, one for built-in functions (which can be accessed without locks) and other for dynamic functions.  That will be a bigger change and like I mentioned before, is not considered worth the effort.

sudheeshkatkam:
+1

Please squash commits, and open a ticket for the enhancement you mentioned.

PR #758, opened 2017-02-23
DRILL-5287: Provide option to skip updates of ephemeral state changes…

ppadma:
… in Zookeeper

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -413,4 +413,8 @@ public interface ExecConstants {
 
   String DYNAMIC_UDF_SUPPORT_ENABLED = "exec.udf.enable_dynamic_support";
   BooleanValidator DYNAMIC_UDF_SUPPORT_ENABLED_VALIDATOR = new BooleanValidator(DYNAMIC_UDF_SUPPORT_ENABLED, true, true);
+
+  String QUERY_TRANSIENT_STATE_UPDATE_KEY = "exec.query.progress.update";
+  BooleanValidator QUERY_TRANSIENT_STATE_UPDATE = new BooleanValidator(QUERY_TRANSIENT_STATE_UPDATE_KEY, true);
+
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
@@ -29,6 +29,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.exceptions.UserRemoteException;
+import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.coord.ClusterCoordinator;
 import org.apache.drill.exec.coord.store.TransientStore;
 import org.apache.drill.exec.coord.store.TransientStoreConfig;
@@ -106,6 +107,9 @@ public class QueryManager implements AutoCloseable {
 
   // How many fragments have finished their execution.
   private final AtomicInteger finishedFragments = new AtomicInteger(0);
+
+  // Is the query saved in transient store
+  private boolean inTransientStore;
 
   public QueryManager(final QueryId queryId, final RunQuery runQuery, final PersistentStoreProvider storeProvider,
       final ClusterCoordinator coordinator, final Foreman foreman) {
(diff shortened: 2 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -413,4 +413,8 @@
 
   String DYNAMIC_UDF_SUPPORT_ENABLED = "exec.udf.enable_dynamic_support";
   BooleanValidator DYNAMIC_UDF_SUPPORT_ENABLED_VALIDATOR = new BooleanValidator(DYNAMIC_UDF_SUPPORT_ENABLED, true, true);
+
+  String ZK_QUERY_STATE_UPDATE_KEY = "drill.exec.zk.query.state.update";
```
paul-rogers: Maybe ENABLE_QUERY_STATE_UPDATE_KEY? For the key name, maybe don't include ZK. ZK is our solution today, but this option would be equally valid if we used some other mechanism.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -1010,7 +1010,9 @@ public void addToEventQueue(final QueryState newState, final Exception exception
 
   private void recordNewState(final QueryState newState) {
     state = newState;
-    queryManager.updateEphemeralState(newState);
+    if (queryContext.getOptions().getOption(ExecConstants.ZK_QUERY_STATE_UPDATE)) {
+      queryManager.updateEphemeralState(newState);
+    }
```
paul-rogers: How does this affect query operation for long-running queries? How does it impact the query profile? If updates are enabled, do we still do an update at query completion to finalize the profile? If not, should writing of the profile be automatically disabled if status updates are disabled?

Do we do any timeout on updates? Will we notice that the query has not been updated and, say, kill the query due to timeouts?

ppadma: For long running queries, it may not make much difference. It adds latency of around ~50-60 msec for single query. However, with high concurrency, impact of contention because of zookeeper updates is significant. Like I mentioned in the JIRA, for concurrency=100, the average query response time for simple queries is 8 sec vs 0.2 sec with these updates disabled.  It does not impact the query profile. Query profile gets updated and written at the end of the query as usual.  This option affects only running queries. In Web UI, you will not see running queries and their state.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
ppadma: I changed it to QUERY_TRANSIENT_STATE_UPDATE_KEY  and exec.query.progress.update. Please review the new diffs.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
```diff
@@ -280,8 +281,15 @@ public void interrupted(final InterruptedException ex) {
     }
   }
 
-  QueryState updateEphemeralState(final QueryState queryState) {
-    switch (queryState) {
+  void updateEphemeralState(final QueryState queryState) {
+      // If query is already in zk transient store, ignore the transient state update option.
+      // Else, they will not be removed from transient store upon completion.
+      if (transientProfiles.get(stringQueryId) == null &&
```
sudheeshkatkam: Why not just check the option?

`transientProfiles.get(stringQueryId)` is quite expensive itself ([contacts ZooKeeper and deserializes data](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZkEphemeralStore.java#L61)).

ppadma: I want to bypass the option for the queries which are already in transient store when option is enabled. Otherwise, their state will never get updated and/or will never be removed from transient store. web UI will show these queries as running forever :-)

Thanks for raising a good point regarding using transientProfiles.get. I made the change to update and use in memory state instead. 

Please review the new diffs.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
```diff
@@ -280,31 +284,39 @@ public void interrupted(final InterruptedException ex) {
...
         }
-
         break;
 
       default:
         throw new IllegalStateException("unrecognized queryState " + queryState);
     }
 
-    return queryState;
+    return;
```
sudheeshkatkam: remove unnecessary return

sudheeshkatkam:
+1 (minor comment)

PR #759, opened 2017-02-24
DRILL-5255: Remove default temporary workspace check at drillbit star…

arina-ielchiieva:
…t up

Check if default temporary workspace is valid (exists, file-based and writable) will be performed only if we attempt to create / drop temporary table, this will allow users who don't use temporary tables successfully use drill functionality even if default temporary workspace is not valid.

1. Removed default temporary workspace validity check at drillbit start up.
2. Ensured that default temporary workspace validity checks are only performed when user tries to create / drop temporary tables.
3. Replaced PARSE errors to VALIDATION ones if workspace is not valid (for CTAS tables).
4. Minor refactoring.


sudheeshkatkam:
+1

PR #761, opened 2017-02-24
DRILL-5284: Roll-up of final fixes for managed sort

paul-rogers:
See subtasks for details.

* Provide detailed, accurate estimate of size consumed by a record batch
* Managed external sort spills too often with Parquet data
* Managed External Sort fails with OOM
* External sort refers to the deprecated HDFS fs.default.name param
* Config param drill.exec.sort.external.batch.size is not used
* NPE in managed external sort while spilling to disk
* External Sort BatchGroup leaks memory if an OOM occurs during read
* Ensure at least two batches are merged in low-memory conditions

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
@@ -200,6 +200,7 @@ public class ExternalSortBatch extends AbstractRecordBatch<ExternalSort> {
   public static final String INTERRUPTION_AFTER_SORT = "after-sort";
   public static final String INTERRUPTION_AFTER_SETUP = "after-setup";
   public static final String INTERRUPTION_WHILE_SPILLING = "spilling";
+  public static final String INTERRUPTION_WHILE_MERGING = "merging";
   public static final long DEFAULT_SPILL_BATCH_SIZE = 8L * 1024 * 1024;
   public static final long MIN_SPILL_BATCH_SIZE = 256 * 1024;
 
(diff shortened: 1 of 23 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -948,50 +1027,50 @@ private void updateMemoryEstimates(long memoryDelta, RecordBatchSizer sizer) {
...
-      // complete row.
+      // Figure out the minimum output batch size based on memory,
+      // must hold at least one complete row.
 
-      targetMergeBatchSize = Math.max(estimatedRowWidth, targetMergeBatchSize);
-      spillPoint = estimatedInputBatchSize + 2 * spillBatchSize;
-      minMergeMemory = minMergeBatches * targetMergeBatchSize;
+      long mergeAllowance = Math.round((memoryLimit - 2 * targetSpillBatchSize) * 0.95);
+      targetMergeBatchSize = Math.max(estimatedRowWidth, mergeAllowance / 2);
+      mergeBatchRowCount = (int) (targetMergeBatchSize / estimatedRowWidth / 2);
```
Ben-Zvi: If estimatedRowWidth is huge, then mergeBatchRowCount may be zero !

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -1231,52 +1308,44 @@ private boolean consolidateBatches() {
...
     spillCount = Math.min(batchGroups.size(), spillCount);
     assert spillCount > 0 : "Spill count to mergeAndSpill must not be zero";
-    long spillSize = 0;
     for (int i = 0; i < spillCount; i++) {
-      @SuppressWarnings("resource")
-      BatchGroup batch = batchGroups.pollFirst();
-      assert batch != null : "Encountered a null batch during merge and spill operation";
-      batchesToSpill.add(batch);
-      spillSize += batch.getDataSize();
+      batchesToSpill.add(batchGroups.pollFirst());
```
Ben-Zvi: In case there was only one batch group, but we bumped spillCount up to 2 (line 1332 above), then pollFirst() would return a null at the second time ?

paul-rogers:
Fixed code review comments. Rebased on latest master.

PR #762, opened 2017-02-24
DRILL-5114: Rationalize use of Logback logging in unit tests

paul-rogers:
Renamed logback.xml file used for testing to logback-test.xml as per
the Logback documentation. Made logging less detailed in the test
version to reduce verbose output.

### DIFF
--- a/exec/java-exec/src/test/resources/logback-test.xml
+++ b/exec/java-exec/src/test/resources/logback-test.xml
@@ -32,12 +32,13 @@
   </appender>
 
   <logger name="org.apache.drill" additivity="false">
-    <level value="debug" />
+    <level value="error" />
     <appender-ref ref="SOCKET" />
+    <appender-ref ref="STDOUT" />
   </logger>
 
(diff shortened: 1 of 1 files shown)


sudheeshkatkam:
Some of us rely on Lilith (SOCKET ref) to view "debug" logs while running unit tests, any reason to change to "error" level?

exec/java-exec/src/test/resources/logback-test.xml
```diff
@@ -32,12 +32,13 @@
   </appender>
 
   <logger name="org.apache.drill" additivity="false">
-    <level value="debug" />
+    <level value="error" />
     <appender-ref ref="SOCKET" />
+    <appender-ref ref="STDOUT" />
```
jinfengni: Why do we want to put all the log messages in STDOUT?

paul-rogers:
Can't achieve different log levels for different appenders. To favor Lilith, cannot reduce the level for console. So, need to come up with an alternative solution. Closing again until that is sorted out.

PR #763, opened 2017-02-24
DRILL-5208: Finding path to java executable should be deterministic

paul-rogers:
See DRILL-5208 for background. Instead of using “find” to locate the
java command, we use the any information available, resorting to find
only if the “usual suspects” fails. The result is that we use the JDK
java when available, instead of randomly choosing JDK or JRE java.

### DIFF
--- a/distribution/src/resources/drill-config.sh
+++ b/distribution/src/resources/drill-config.sh
@@ -347,10 +347,17 @@ case "`uname`" in
 CYGWIN*) is_cygwin=true;;
 esac
 
+if $is_cygwin; then
+  JAVA_BIN="java.exe"
+else
+  JAVA_BIN="java"
+fi
+
 # Test for or find JAVA_HOME
+
 if [ -z "$JAVA_HOME" ]; then
-  if [ -e `which java` ]; then
-    SOURCE=`which java`
+  SOURCE=`which java`
+  if [ -e $SOURCE ]; then
     while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
       DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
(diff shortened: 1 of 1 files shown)


distribution/src/resources/drill-config.sh
```diff
@@ -367,12 +375,18 @@ if [ -z "$JAVA_HOME" ]; then
 fi
 
 # Now, verify that 'java' binary exists and is suitable for Drill.
-if $is_cygwin; then
-  JAVA_BIN="java.exe"
-else
-  JAVA_BIN="java"
+# If we started with `which java` above, use that path (after replacing
+# symlinks. If we started with JAVA_HOME, try in bin. Doing so handles
```
arina-ielchiieva: Close the brace.

distribution/src/resources/drill-config.sh
```diff
@@ -359,6 +366,7 @@ if [ -z "$JAVA_HOME" ]; then
       [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
     done
     JAVA_HOME="$( cd -P "$( dirname "$SOURCE" )" && cd .. && pwd )"
+    JAVA = $SOURCE/$JAVA_BIN
```
arina-ielchiieva: $SOURCE already includes bin/java, so concatenating with $JAVA_BIN will give you:
`/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.111-1.b15.el7_2.x86_64/jre/bin/java/java`.
I guess you should concatenate $JAVA_BIN to $SOURCE dirname.

paul-rogers:
Adressed code review comments and rebased onto latest master.

sudheeshkatkam:
Please squash the commits as well.

arina-ielchiieva:
+1, looks good.

PR #764, opened 2017-02-27
DRILL-5301: Server metadata API

laurentgo:
Add a Server metadata API to the User protocol, to query server support of various SQL features.

Add support to the client (DrillClient) to query this information.
    
Add support to the JDBC driver to query this information, if the server supports the new API, or fallback to the previous behaviour (rely on Avatica defaults) otherwise.

Add support to the Server metadata API to the C++ client if available. If the API is not supported to the server, fallback to the previous hard-coded values.


protocol/src/main/protobuf/User.proto
```diff
@@ -439,6 +439,211 @@ message CreatePreparedStatementResp {
 }
 
 /*
+ * Request message for getting server metadata
+ */
+message GetServerMetaReq {
+}
+
+enum CollateSupport {
```
sudheeshkatkam: For all enums below, use UNKNOWN as the zero-th value to avoid the problem [mentioned here](http://androiddevblog.com/protocol-buffers-pitfall-adding-enum-values/)?

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -1722,8 +1774,68 @@ void DrillClientImpl::shutdownSocket(){
     DRILL_MT_LOG(DRILL_LOG(LOG_TRACE) << "Socket shutdown" << std::endl;)
 }
 
+namespace { // anonymous
```
sudheeshkatkam: empty block?

laurentgo: no, it's an anonymous namespace: the actual name is generated at compilation time, and it prevents symbols to leak from outside object file, or clash with other symbols (if not declared static).

protocol/src/main/protobuf/User.proto
laurentgo: done

sudheeshkatkam:
+1

Is this PR inclusive of comments in #613 (for DRILL-4994, DRILL-4730)?

laurentgo:
yes, pull requests have been kept in sync

jinfengni:
Which JIRA should the last commit (Refactor DrillCursor) belong to, DRILL-4994, 4730, OR 5301?

laurentgo:
The cursor change belongs to DRILL-4994

PR #765, opened 2017-02-28
DRILL-5293: Change the seed of the hash function for distribution

Ben-Zvi:
The fix is to use a different _seed_ for the hash function in the cases the hash is computed for distribution (i.e., when calling the method `getHashExpression(List<DistributionField> fields, RelDataType rowType)` ).
   The implementation -- added a third parameter _seed_ to createCall() (on line 102 in HashPrelUtil.java), which makes it use the "with seed" versions of the generated code functions (before that it was using the versions that had a built-in seed of zero).
   Then made all the calling places pass in a seed -- zero in the case of a Hash Table , or a large fixed prime ( 1301011 ) for distribution.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/HashPrelUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/HashPrelUtil.java
@@ -24,6 +24,7 @@ import org.apache.drill.common.expression.ExpressionPosition;
 import org.apache.drill.common.expression.FieldReference;
 import org.apache.drill.common.expression.FunctionCall;
 import org.apache.drill.common.expression.LogicalExpression;
+import org.apache.drill.common.expression.ValueExpressions;
 import org.apache.drill.exec.planner.physical.DrillDistributionTrait.DistributionField;
 
 import java.util.ArrayList;
@@ -36,6 +37,7 @@ public class HashPrelUtil {
 
   public static final String HASH_EXPR_NAME = "E_X_P_R_H_A_S_H_F_I_E_L_D";
 
+  public static final int DIST_SEED = 1301011; // distribution seed
   /**
    * Interface for creating different forms of hash expression types.
    * @param <T>
@@ -72,8 +74,9 @@ public class HashPrelUtil {
    */
   public static <T> T createHashBasedPartitionExpression(
       List<T> distFields,
+      T seed,
       HashExpressionCreatorHelper<T> helper) {
-    return createHashExpression(distFields, helper, true /*for distribution always hash as double*/);
+    return createHashExpression(distFields, seed, helper, true /*for distribution always hash as double*/);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/HashPrelUtil.java
```diff
@@ -89,14 +91,15 @@ public LogicalExpression createCall(String funcName, List<LogicalExpression> inp
...
+      T seed,
       HashExpressionCreatorHelper<T> helper,
       boolean hashAsDouble) {
 
     assert inputExprs.size() > 0;
 
     final String functionName = hashAsDouble ? HASH32_DOUBLE_FUNCTION_NAME : HASH32_FUNCTION_NAME;
 
-    T func = helper.createCall(functionName,  ImmutableList.of(inputExprs.get(0)));
+    T func = helper.createCall(functionName,  ImmutableList.of(inputExprs.get(0), seed ));
```
amansinha100: Could you confirm if all the function names in function template and data types that are covered in the Hash32Functions are also present in the Hash32FunctionsWithSeed ?

Ben-Zvi: Must be, as two code lines below there is another call to createCall() with three parameters (the last of type T). Also looked -- the files (of one kind) were copied (from the other one).

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/HashPrelUtil.java
```diff
@@ -129,11 +133,12 @@ public static LogicalExpression getHashExpression(List<DistributionField> fields
       return new FieldReference(HASH_EXPR_NAME);
     }
 
+    final LogicalExpression distSeed = ValueExpressions.getInt(1301011); // distribution seed
```
Ben-Zvi: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/InsertLocalExchangeVisitor.java
```diff
@@ -119,7 +118,8 @@ public Prel visitExchange(ExchangePrel prel, Void value) throws RuntimeException
       }
 
       outputFieldNames.add(HashPrelUtil.HASH_EXPR_NAME);
-      updatedExpr.add(HashPrelUtil.createHashBasedPartitionExpression(distFieldRefs, hashHelper));
+      final RexNode distSeed = rexBuilder.makeExactLiteral(BigDecimal.valueOf(1301011)); // distribution seed
```
Ben-Zvi: Done.

Ben-Zvi:
@kkhatua Actually any number (other than zero) would work here and completely change the hash result; I just went a little extra by picking a large prime.

amansinha100:
+1.   Pls squash commits before it is merged.

PR #767, opened 2017-03-03
DRILL-5226: Managed external sort fixes

paul-rogers:
* Memory leak in managed sort if OOM during sv2 allocation
* "Record batch sizer" does not include overhead for variable-sized
vectors
* Paranoid double-checking of merge batch sizes to prevent OOM when the
sizes differ from expectations
* Revised logging

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/CopierHolder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/CopierHolder.java
@@ -176,7 +176,7 @@ public class CopierHolder {
     private int targetRecordCount;
     private int copyCount;
     private int batchCount;
-    private long batchSize;
+    private long estBatchSize;
 
     /**
      * Creates a merger with an temporary output container.
@@ -238,7 +238,7 @@ public class CopierHolder {
         batchCount++;
         logger.trace("Took {} us to merge {} records", t, count);
         long size = holder.allocator.getAllocatedMemory() - start;
-        batchSize = Math.max(batchSize, size);
+        estBatchSize = Math.max(estBatchSize, size);
       } else {
         logger.trace("copier returned 0 records");
       }
@@ -308,8 +308,15 @@ public class CopierHolder {
       return batchCount;
     }
 
-    public long getBatchSize() {
-      return batchSize;
+    /**
+     * Gets the estimated batch size, in bytes. Use for estimating the memory
+     * needed to process the batches that this operator created.
+     * @return the size of the largest batch created by this operation,
+     * in bytes
+     */
+
+    public long getEstBatchSize() {
+      return estBatchSize;
     }
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/CopierHolder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/CopierHolder.java
@@ -176,6 +176,7 @@ public class CopierHolder {
     private int targetRecordCount;
     private int copyCount;
     private int batchCount;
+    private long batchSize;
 
     /**
      * Creates a merger with an temporary output container.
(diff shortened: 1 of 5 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -1333,8 +1339,43 @@ private void spillFromMemory() {
...
+    // Must always spill at least 2, even if this creates an over-size
+    // spill file. But, if this is a final consolidation, we may have only
+    // a single batch.
+
+    mergeCount = Math.max(mergeCount, 2);
+    mergeCount = Math.min(mergeCount, spilledRuns.size());
+
+    // Do the actual spill.
+
+    mergeAndSpill(spilledRuns, mergeCount);
```
Ben-Zvi: Just a comment: So we always merge the _FIRST_ mergeCount runs. So if (one of) the first run has some crazy "oversized" batch, we'd repeatedly merge a small number of runs, as that bad batch may be preserved on.
Alternatively - select the runs with the smaller "max batch"es, hence getting more runs to merge.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
```diff
@@ -1333,8 +1339,43 @@ private void spillFromMemory() {
...
+  private void mergeRuns(int targetCount) {
+
+    // Determine the number of runs to merge. The count should be the
+    // target count. However, to prevent possible memory overrun, we
+    // double-check with actual spill batch size and only spill as much
+    // as fits in the merge memory pool.
+
+    int mergeCount = 0;
+    long mergeSize = 0;
+    for (SpilledRun batch : spilledRuns) {
```
Ben-Zvi: Rename "batch" to "run" ?

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -253,7 +253,7 @@ public int getPayloadByteCount() {
       // If 1 or more values, then the last value is set to
       // the offset of the next value, which is the same as
       // the length of existing values.
-      return a.get(count-1);
+      return a.get(count-1) + offsetVector.getPayloadByteCount();
```
Ben-Zvi: Don't understand this new addition; doesn't the get(count-1) already reports all the memory used by this varchar, including the overhead ?  Maybe need to explain in the comment.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/CopierHolder.java
```diff
@@ -303,5 +307,9 @@ public int getRecordCount() {
     public int getBatchCount() {
       return batchCount;
     }
+
+    public long getBatchSize() {
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
paul-rogers: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/ExternalSortBatch.java
paul-rogers: The logic is a bit more complex. We use the size of the largest row as our row size estimate. But, when we spill, we also consider the actual batch sizes. Earlier versions of the code were more sensitive to the row size estimate, more recent revisions have evolved to favor actual sizes in many cases.

For "first generation" batches, we use the actual "payload" amount of data in each batch, sum that amount, and stop when we have enough data to fill a spill file. Thus, first generation spills are immune to the problem you describe.

For merge/spills, we guess a target number of batches based on the target spill batch size. If the above logic works, then we know (roughly) the size of the spilled batches -- because we created them. As a sanity check, new code above sums the actual batch sizes in the files to merge to ensure that they fit in memory, choosing a smaller-than-requested number of batches if the merge batches turn out to be a bit larger than expected.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
paul-rogers: Added a comment. Basically, for planning purposes, the "payload" bytes for a Varchar vector include the bytes taken to hold the text data PLUS the bytes needed by the offset vector that points to those bytes. When averaged, it means a Varchar column width is 4 + avg. byte width.

amansinha100:
+1

PR #768, opened 2017-03-03
DRILL-5313: Fix build failure in C++ client

sohami:


contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -662,8 +662,9 @@ DrillClientQueryResult* DrillClientImpl::ExecuteQuery(const PreparedStatement& p
 }
 
 static void updateLikeFilter(exec::user::LikeFilter& likeFilter, const std::string& pattern) {
-	likeFilter.set_pattern(pattern);
-	likeFilter.set_escape(meta::DrillMetadata::s_searchEscapeString);
+    likeFilter.set_pattern(pattern);
+    exec::user::ServerMeta srvrMetaData = meta::DrillMetadata::s_defaultServerMeta;
```
laurentgo: it's a workaround but it should use the actual escape character and not the default one.

sohami:
Thanks for the actual fix. I will close this pull request.

PR #769, opened 2017-03-03
DRILL-5313: Fix compilation issue in C++ connector

laurentgo:
DRILL-5301 and DRILL-5167 have conflicting changes, which causes
the C++ connector to not compile: the static symbol for the search
escape string has been removed as the server might use a different one.

Fix the issue by using the current search escape string (injected from the
meta to the internal drill client when querying metadata).


sudheeshkatkam:
+1

sudheeshkatkam:
Built, and sanity tested using querySubmitter test application.

PR #770, opened 2017-03-03
DRILL-5311: Check handshake result in C++ connector

laurentgo:
In C++ client connector, DrillClientImpl::recvHandshake always
return success, even in case of connection error (like a tcp
timeout issue). Only on WIN32 platform would the error code be
checked.

Remove the restriction to only check on WIN32, plus add some logging.


sudheeshkatkam:
+1

PR #772, opened 2017-03-07
DRILL-5316: Check drillbits size before we attempt to access the vector element

superbstreak:
…we attempt to access the vector element

### DIFF
--- a/contrib/native/client/src/clientlib/drillClientImpl.cpp
+++ b/contrib/native/client/src/clientlib/drillClientImpl.cpp
@@ -86,6 +86,9 @@ connectionStatus_t DrillClientImpl::connect(const char* connStr, DrillUserProper
         std::vector<std::string> drillbits;
         int err = zook.getAllDrillbits(hostPortStr, drillbits);
         if(!err){
+            if (drillbits.empty()){
+                return handleConnError(CONN_FAILURE, getMessage(ERR_CONN_ZKNODBIT));
+            }
             Utils::shuffle(drillbits);
             exec::DrillbitEndpoint endpoint;
             err = zook.getEndPoint(drillbits[drillbits.size() -1], endpoint);// get the last one in the list
@@ -2140,6 +2143,9 @@ connectionStatus_t PooledDrillClientImpl::connect(const char* connStr, DrillUser
         std::vector<std::string> drillbits;
         int err = zook.getAllDrillbits(hostPortStr, drillbits);
         if(!err){
+            if (drillbits.empty()){
+                return handleConnError(CONN_FAILURE, getMessage(ERR_CONN_ZKNODBIT));
+            }
             Utils::shuffle(drillbits);
             // The original shuffled order is maintained if we shuffle first and then add any missing elements
             Utils::add(m_drillbits, drillbits);


contrib/native/client/src/clientlib/zookeeperClient.cpp
```diff
@@ -138,6 +138,11 @@ int ZookeeperClient::getAllDrillbits(const std::string& connectStr, std::vector<
             DRILL_MT_LOG(DRILL_LOG(LOG_TRACE) << "\t Unshuffled Drillbit id: " << drillbits[i] << std::endl;)
         }
     }
+    else{
```
laurentgo: I'm not sure the check should be in the zookeeper client, but instead in the drill client: the zookeeper client is just returning the content of the zookeeper tree, and there's no error from the zookeeper client point of view (so no need to disconnect/reconnect).

sohami: Agreed. Should be handled in caller (i.e. DrillClient). If the returned vector size is zero then we should check that in DrillClient and close client connection with error as `ERR_CONN_ZKNODBIT`. Something like below:

`return handleConnError(CONN_INVALID_INPUT, getMessage(ERR_CONN_ZKNODBIT, pathToDrill.c_str()));`

superbstreak: Thanks both. Yea make sense, I'll make the change.

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -86,6 +86,9 @@ connectionStatus_t DrillClientImpl::connect(const char* connStr, DrillUserProper
         std::vector<std::string> drillbits;
         int err = zook.getAllDrillbits(hostPortStr, drillbits);
         if(!err){
+            if (drillbits.empty()){
+                return handleConnError(CONN_INVALID_INPUT, getMessage(ERR_CONN_ZKNODBIT));
```
superbstreak: double check CONN_INVALID_INPUT usage.

superbstreak:
@sohami please review.

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -2143,6 +2146,9 @@ connectionStatus_t PooledDrillClientImpl::connect(const char* connStr, DrillUser
             Utils::shuffle(drillbits);
             // The original shuffled order is maintained if we shuffle first and then add any missing elements
             Utils::add(m_drillbits, drillbits);
+            if (m_drillbits.empty()){
+                return handleConnError(CONN_FAILURE, getMessage(ERR_CONN_ZKNODBIT));
```
laurentgo: can be done before the shuffle (it should not change the number of elements)

superbstreak: Thanks, @laurentgo! A new ticket has been filed for the suggestion: [DRILL-5335](https://issues.apache.org/jira/browse/DRILL-5335)

sohami: Since we are not removing the offline nodes from m_drillbits then I think we should return connection error before shuffle. Let's say on first client connection we get all the active node from zookeeper and store it in m_drillbits. Then all the nodes went dead or offline. In the next connection request, zookeeper will return zero drillbits but since m_drillbits is not empty we will still try to connect and fail later. 

Instead I think zero drillbits returned from zookeeper is a good indication that we won't be able to connect to any other node already present inside m_drillbits and should fail there itself ?

sohami:
Thanks @superbstreak for the change. 
LGTM. +1 from my side.

amansinha100:
+1

PR #775, opened 2017-03-07
DRILL-5326: Unit tests failures related to the SERVER_METADTA

vdiravka:
- adding of the sql type name for the "GENERIC_OBJECT";
- changing "NullCollation" in the "ServerMetaProvider" to the correct default value;

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/metadata/ServerMetaProvider.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/metadata/ServerMetaProvider.java
@@ -76,7 +76,7 @@ public class ServerMetaProvider {
       .setReadOnly(false)
       .setGroupBySupport(GroupBySupport.GB_UNRELATED)
       .setLikeEscapeClauseSupported(true)
-      .setNullCollation(NullCollation.NC_AT_END)
+      .setNullCollation(NullCollation.NC_HIGH)
       .setSchemaTerm("schema")
       .setSearchEscapeString("\\")
       .setTableTerm("table")
--- a/common/src/main/java/org/apache/drill/common/types/Types.java
+++ b/common/src/main/java/org/apache/drill/common/types/Types.java
@@ -151,6 +151,7 @@ public class Types {
       case LATE:            return "ANY";
       case NULL:            return "NULL";
       case UNION:           return "UNION";
+      case GENERIC_OBJECT:  return "JAVA_OBJECT";
 
       // Internal types not actually used at level of SQL types(?):
 
@@ -198,6 +199,7 @@ public class Types {
       case "TIMESTAMP":                     return java.sql.Types.TIMESTAMP;
       case "TINYINT":                       return java.sql.Types.TINYINT;
       case "UNION":                         return java.sql.Types.OTHER;
+      case "JAVA_OBJECT":                   return java.sql.Types.JAVA_OBJECT;
       default:
         // TODO:  This isn't really an unsupported-operation/-type case; this
         //   is an unexpected, code-out-of-sync-with-itself case, so use an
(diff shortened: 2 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/work/metadata/ServerMetaProvider.java
```diff
@@ -76,7 +76,7 @@
       .setReadOnly(false)
       .setGroupBySupport(GroupBySupport.GB_UNRELATED)
       .setLikeEscapeClauseSupported(true)
-      .setNullCollation(NullCollation.NC_AT_END)
+      .setNullCollation(NullCollation.NC_HIGH)
```
jinfengni: I'm not completely sure why we should change from NC_AT_END to NC_HIGH, in stead of NC_AT_END.  I thought Drill is using ASC as default ordering, and NULLS LAST as default null collation for ASC. This is consistent to what Oracle [1] and Postgres [2] is doing : ASC /NULL LAST is the default option.  

1. http://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqlj13658.html
2. https://www.postgresql.org/docs/9.4/static/queries-order.html

zfong: @jinfengni  - See @vdiravka's comments in the Jira.  The Drill documentation at https://drill.apache.org/docs/order-by-clause/#usage-notes says NULLs sort highest.  If the doc is wrong, then we should fix the doc.

jinfengni: My understanding is NULL collation should be specified together with ASC/DESC.  ASC/NULL LAST as the default option essentially implies NC_HIGH. However, I'm not sure if we just specify NC_HIGH alone, or ASC/NC_AT_END combined.

paul-rogers: When sorting in Drill, the detailed sort spec is set in the {{ExternalSort}} operator definition. This thing is a bit complex. One can control sort order (ASC, DESC) and nulls position (LOW, HIGH, UNSPECIFIED.)

Data sorts according to ASC, DESC.
Nulls sort as follows:

HIGH: last if ASC, first if DESC
LOW: first if ASC, last if DESC
UNSPECIFIED: always high

If the planner has no way of setting the nulls ordering from a SQL query, then the value is UNSPECIFIED, which means nulls always sort last as Jinfeng explained.

vdiravka: @jinfengni Agree with Paul. The difference between NC_HIGH and NC_AT_END for the DESC case. I checked that Drill uses NC_HIGH for sorting. Please, see it at [jira](https://issues.apache.org/jira/browse/DRILL-5326?focusedCommentId=15900411&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15900411), since github doesn't show formatted code correctly in the comments.

laurentgo: I'm not sure if the last sentence is correct. If user specified DESC, and null collation is unspecified, then null should sort first, no (if always HIGH). NC_HIGH seems then the correct value (unless the default can be changed in the planner configuration).

jinfengni: If only specify NC here (no sort order), then NC_HIGH looks more reasonable. The change looks fine to me.

vdiravka:
@laurentgo I changed the names for the server meta method.

laurentgo:
LGTM

jinfengni:
@laurentgo , @vdiravka , thanks. I'll run regression & merge.

+1

PR #776, opened 2017-03-08
DRILL-5165: limitRel to return correct rows for limit all case

chunhui-shi:


exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLimitWithExchanges.java
```diff
@@ -125,6 +125,15 @@ public void testLimitImpactExchange() throws Exception {
     }
   }
 
+  @Test
+  public void TestLimitAllOnParquet() throws Exception {
```
jinfengni: Have you tried this new testcase without the change above? Will it successful or fail?

chunhui-shi:
It failed without the fix.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLimitWithExchanges.java
chunhui-shi: update the tested query. I did not notice the query in unit test is slightly different than the one I used in sqlline.

jinfengni:
+1

PR #777, opened 2017-03-10
DRILL-5330: NPE in FunctionImplementationRegistry

paul-rogers:
Fixes:

* DRILL-5330: NPE in
FunctionImplementationRegistry.functionReplacement()
* DRILL-5331:
NPE in FunctionImplementationRegistry.findDrillFunction() if dynamic
UDFs disabled

For DRILL-5331, we leverage an existing session option to determine if
DUDFs are enabled. If not, we skip the DUDF registry check.

For DRILL-5330, we use an existing option validator rather than
accessing the raw option directly.

Then, both options cached on setup rather than repeatedly resolved in
each function lookup.

Also includes a bit of code cleanup in the class in question.

The result is that the code now works when used in a sub-operator unit
test.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
@@ -83,17 +83,30 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
   private boolean deleteTmpDir = false;
   private File tmpDir;
   private List<PluggableFunctionRegistry> pluggableFuncRegistries = Lists.newArrayList();
-  private OptionManager optionManager = null;
+  private final OptionManager optionManager;
+  private final boolean useDynamicUdfs;
 
-  @Deprecated @VisibleForTesting
-  public FunctionImplementationRegistry(DrillConfig config){
-    this(config, ClassPathScanner.fromPrescan(config));
+  @VisibleForTesting
+  public FunctionImplementationRegistry(DrillConfig config) {
+    this(config, ClassPathScanner.fromPrescan(config), null);
   }
 
-  public FunctionImplementationRegistry(DrillConfig config, ScanResult classpathScan){
+  public FunctionImplementationRegistry(DrillConfig config, ScanResult classpathScan) {
+    this(config, classpathScan, null);
+  }
+
+  public FunctionImplementationRegistry(DrillConfig config, ScanResult classpathScan, OptionManager optionManager) {
     Stopwatch w = Stopwatch.createStarted();
 
     logger.debug("Generating function registry.");
+    this.optionManager = optionManager;
+
+    // Unit tests fail if dynamic UDFs are turned on AND the test happens
+    // to access an undefined function. Since we want a reasonable failure
+    // rather than a crash, we provide a boot-time option, set only by
+    // tests, to disable DUDF lookup.
+
+    useDynamicUdfs = ! config.getBoolean(ExecConstants.UDF_DISABLE_DYNAMIC);
     localFunctionRegistry = new LocalFunctionRegistry(classpathScan);
 
     Set<Class<? extends PluggableFunctionRegistry>> registryClasses =
@@ -123,11 +136,6 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
     this.localUdfDir = getLocalUdfDir(config);
   }
 
-  public FunctionImplementationRegistry(DrillConfig config, ScanResult classpathScan, OptionManager optionManager) {
-    this(config, classpathScan);
-    this.optionManager = optionManager;
-  }
-
   /**
    * Register functions in given operator table.
    * @param operatorTable operator table
@@ -142,7 +150,7 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
   }
 
   /**
-   * First attempts to finds the Drill function implementation that matches the name, arg types and return type.
+   * First attempts to find the Drill function implementation that matches the name, arg types and return type.
    * If exact function implementation was not found,
    * syncs local function registry with remote function registry if needed
    * and tries to find function implementation one more time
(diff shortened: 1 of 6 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
```diff
@@ -160,7 +168,7 @@ public DrillFuncHolder findDrillFunction(FunctionResolver functionResolver, Func
     FunctionResolver exactResolver = FunctionResolverFactory.getExactResolver(functionCall);
     DrillFuncHolder holder = exactResolver.getBestMatch(functions, functionCall);
 
-    if (holder == null) {
+    if (holder == null && useDynamicUdfs) {
```
arina-ielchiieva: 1. When dynamic udfs are enabled, we first try to find matching function using exact function resolver `exactResolver.getBestMatch` and if match if not found check if remote and local function registries are in sync and then try to find matching function using provided function resolver `functionResolver.getBestMatch`. 
So when dynamic udfs are disabled, we should not try to find matching function  using exact function resolver, we need to use provided one and skip function registry sync check,

2. Regarding new bootstap option, since we already have system / session option `ExecConstants.USE_DYNAMIC_UDFS`, I suggest we use it during this check, first we won't have two similar options, second if we re-write method as suggested above (point 1), users who disabled dynamic udfs using system / session option will have performance benefit since we won't double-check function existence. Probably we should have done this when `ExecConstants.USE_DYNAMIC_UDFS` was introduced in first place.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestSimpleFunctions.java
```diff
@@ -62,10 +62,11 @@
 
 public class TestSimpleFunctions extends ExecTest {
   //private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(TestSimpleFunctions.class);
-  private final DrillConfig c = DrillConfig.create();
+//  private final DrillConfig c = DrillConfig.create();
```
arina-ielchiieva: Why we need to comment this? If it's not need, can we just remove it?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestSimpleFunctions.java
```diff
@@ -133,7 +134,7 @@ public void resolveHash(DrillConfig config, LogicalExpression arg, TypeProtos.Ma
                                     FunctionImplementationRegistry registry) throws JClassAlreadyExistsException, IOException {
     final List<LogicalExpression> args = new ArrayList<>();
     args.add(arg);
-    final String[] registeredNames = { "hash" };
+//    final String[] registeredNames = { "hash" };
```
arina-ielchiieva: Why we need to comment this? If it's not need, can we just remove it?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestSimpleFunctions.java
paul-rogers: I can remove it. I initially commented it just to see if things would still work.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestSimpleFunctions.java
paul-rogers: Same reason. Removed this and the previous commented-out code.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
arina-ielchiieva: 1. Since you have mentioned I remembered one more issue with FunctionImplementationRegistry, it can access only system options, so using `ExecConstants.USE_DYNAMIC_UDFS` won't work properly since it can be set at session level as well. I guess using bootsrap option you introduced is OK for now. Regarding your suggestion to have single option OFF, READ_ONLY and ON to handle the various cases (I love this idea!), we can try to implement this the scope of MVCC (I'll add this point to the document).

2. Even with boostrap option we need to update `findDrillFunction` to use provided function resolver when dynamic udfs are turned off (more details in my first comment). For example, `findDrillFunction` should can be re-written the following way (please optimize if needed):
```java
public DrillFuncHolder findDrillFunction(FunctionResolver functionResolver, FunctionCall functionCall) {
    AtomicLong version = new AtomicLong();
    String newFunctionName = functionReplacement(functionCall);
    List<DrillFuncHolder> functions = localFunctionRegistry.getMethods(newFunctionName, version);
    if (!useDynamicUdfs) {
       return functionResolver.getBestMatch(functions, functionCall);
    }
    FunctionResolver exactResolver = FunctionResolverFactory.getExactResolver(functionCall);
    DrillFuncHolder holder = exactResolver.getBestMatch(functions, functionCall);

    if (holder == null) {
      syncWithRemoteRegistry(version.get());
      List<DrillFuncHolder> updatedFunctions = localFunctionRegistry.getMethods(newFunctionName, version);
      holder = functionResolver.getBestMatch(updatedFunctions, functionCall);
    }

    return holder;
  }
```
3. Also changes should be done in `findExactMatchingDrillFunction` method to take into account boostrap option as well. For example (please optimize if needed):
```java
  public DrillFuncHolder findExactMatchingDrillFunction(String name, List<MajorType> argTypes, MajorType returnType) {
    if (useDynamicUdfs) {
       return findExactMatchingDrillFunction(name, argTypes, returnType, true);
    }
    return findExactMatchingDrillFunction(name, argTypes, returnType, false);
  }
```

paul-rogers: Ah, now I see what's happening (I hope...) I pushed another commit that makes the suggested changes. 

I wonder, do we have any unit tests for the ambiguous-function case? The unit tests passed with both the original and this new version, so I wonder if we have a hole in our test coverage?

arina-ielchiieva:
+1, looks good. Please squash the commits and put read-to-commit label in Jira when done. Thank you!

PR #778, opened 2017-03-10
DRILL-5344: Priority queue copier fails with an empty batch

paul-rogers:
DRILL-5344:  External sort priority queue copier fails with an empty
batch

Unit tests showed that the “priority queue copier” does not handle an
empty batch. This has not been an issue because code elsewhere in the
sort specifically works around this issue. This fix resolves the issue
at the source to avoid the need for future work-arounds.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/PriorityQueueCopierTemplate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/PriorityQueueCopierTemplate.java
@@ -57,9 +57,12 @@ public abstract class PriorityQueueCopierTemplate implements PriorityQueueCopier
 
     queueSize = 0;
     for (int i = 0; i < size; i++) {
-      vector4.set(i, i, batchGroups.get(i).getNextIndex());
-      siftUp();
-      queueSize++;
+      int index = batchGroups.get(i).getNextIndex();
+      vector4.set(i, i, index);
+      if (index > -1) {
+        siftUp();
+        queueSize++;
+      }
     }
   }
 


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/PriorityQueueCopierTemplate.java
```diff
@@ -57,9 +57,12 @@ public void setup(FragmentContext context, BufferAllocator allocator, VectorAcce
 
     queueSize = 0;
     for (int i = 0; i < size; i++) {
-      vector4.set(i, i, batchGroups.get(i).getNextIndex());
-      siftUp();
-      queueSize++;
+      int index = batchGroups.get(i).getNextIndex();
+      vector4.set(i, i, index);
```
Ben-Zvi: If the record batch is empty, why set this (65535) value in the vector ?
Probably usually the empty batch is last, so queueSize would not be incremented; otherwise if the next iteration brings a non empty batch, then siftUp() may run into that 65535 .... (not sure ...)

paul-rogers: This is tricky, and likely to due conflicting design decisions.

It seems that the vector container retrieved by `get()` has a built-in iterator. Not a good design as it will cause havoc if two bits of code try to iterate at the same time. Sigh...

The iterator decides to return -1 when it reaches the end. That is OK, as long as code checks.

The original code did not check, it just used -1 (truncated to 16 bits, stored as a char, so 65535) as a row index.

The solution is to not add the batch to the "queue" when the iterator returns -1 in the first position.

Note that if the iterator has simply returned the count at the end, and we used the classic "index >= count" as the terminal condition, then none of this would be necessary.

PR #779, opened 2017-03-13
Indexr 0.3.0 drill 1.9.0

xsq0718:


paul-rogers:
Thanks for the contribution! Seems there has been no action on this PR for a while. I suspect this is due to several reasons. Let's see if we can move things along.

* In general, the Drill project requires that each PR be in response to a Drill JIRA ticket. Please create a JIRA ticket for this improvement. Then, please edit the title to say "DRILL-xxxx: description." See other PRs for examples.
* Would be great to describe the work that's been done. The description is best in the JIRA, but can appear as a comment here. (Once the JIRA ticket is in the title, the post will automatically be transferred to JIRA as a comment.) The more information you can provide, the easier it is for reviewers to understand the code. In many cases, you can simply provide links to material that, say, describes what "indexr" is.
* The PR has many conflicts with existing files. For example, the pom files have old versions. To fix this, perhaps rebase the work onto the latest master.
* The PR changes some Drill files (Parquet, etc.) outside the scope of "indexr". Is this intentional? Due to working with an old version of Drill?

Let's see if we can work through the issues so we can begin to review the actual code.

PR #780, opened 2017-03-13
DRILL-5349: Fix TestParquetWriter unit tests when synchronous parquet…

parthchandra:
… reader is used.

Seems like I removed some lines from the code that I should not have. This PR reinstates them.


amansinha100:
+1

PR #781, opened 2017-03-13
DRILL-5351: Minimize bounds checking in var len vectors for Parquet

parthchandra:
reader

Two changes in var len vectors: 
1) Instead of checking to see if we need to realloc for every setSafe call, let the write fail and catch the exception. The exception, though expensive, will happen very rarely.
2) Call fillEmpties only if there are empty values to fill. 
This saves a bunch of CPU on every setSafe call.

### DIFF
--- a/exec/vector/src/main/codegen/templates/VariableLengthVectors.java
+++ b/exec/vector/src/main/codegen/templates/VariableLengthVectors.java
@@ -502,11 +502,15 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements V
       assert index >= 0;
 
       final int currentOffset = offsetVector.getAccessor().get(index);
-      while (data.capacity() < currentOffset + bytes.length) {
-        reAlloc();
-      }
       offsetVector.getMutator().setSafe(index + 1, currentOffset + bytes.length);
-      data.setBytes(currentOffset, bytes, 0, bytes.length);
+      try {
+        data.setBytes(currentOffset, bytes, 0, bytes.length);
+      } catch (IndexOutOfBoundsException e) {
+        while (data.capacity() < currentOffset + bytes.length) {
+          reAlloc();
+        }
+        data.setBytes(currentOffset, bytes, 0, bytes.length);
+      }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -502,11 +502,15 @@ public void setSafe(int index, byte[] bytes) {
...
 
       final int currentOffset = offsetVector.getAccessor().get(index);
-      while (data.capacity() < currentOffset + bytes.length) {
-        reAlloc();
-      }
       offsetVector.getMutator().setSafe(index + 1, currentOffset + bytes.length);
-      data.setBytes(currentOffset, bytes, 0, bytes.length);
+      try {
+        data.setBytes(currentOffset, bytes, 0, bytes.length);
+      } catch (IndexOutOfBoundsException e) {
```
parthchandra: The check for a DrillBuf exceeding bounds is already being done once in Netty (which also throws the Exception). What we want to do  is to avoid having to do more than one bounds check for every vector write operation. By adding the suggested check, we simply move the check from every vector setSafe method to every Drillbuf write. This would possibly impact performance in other parts of the code. 
I particularly changed only var len vectors to address a hotspot in the Parquet reader performance, because as you are suggesting, the right way to fix is to address the low density batch problem at the same time as the vector overflow problem. Perhaps a longer discussion may be required here.

paul-rogers: The Netty/Drillbuf code is complex, so this does boil down to details... Yes, I agree that Netty does the bounds checks -- if we call Netty code. Consider this code in DrillBuf:

```
  @Override
  public ByteBuf setBytes(int index, ByteBuf src, int srcIndex, int length) {
    udle.setBytes(index + offset, src, srcIndex, length);
    return this;
  }

  public ByteBuf setBytes(int index, ByteBuffer src, int srcIndex, int length) {
    if (src.isDirect()) {
      checkIndex(index, length);
      PlatformDependent.copyMemory(PlatformDependent.directBufferAddress(src) + srcIndex, this.memoryAddress() + index,
          length);
  ...
```

We have one version that delegates to the "udle" which calls another buf which calls... all the way to a bounds check.

But, we have another method which cuts out the middleman and just does a direct memory copy. Given that, we could certainly add a version that does a bounds check and copy from heap into direct memory. Just use this Netty method:

```
class  PlatformDependent ...
    public static void copyMemory(byte[] src, int srcIndex, long dstAddr, long length) ...
```

So, something like this:

```
class Drillbuf ...
  public boolean setIfCapacity(int offset, byte data[], int len) {
    if (offset + len >= capacity()) { return false; }
    PlatformDependent.copyMemory(data, 0,
      PlatformDependent.directBufferAddress(src) + offset,
      len);
   return true;
}
```

Of course, this assumes an implementation of the underlying direct memory, but, as we saw, we are already doing something similar.

Would this work and perform as well as the exception-based approach?

paul-rogers: By the way, if the suggested change blows up the scope of this fix, then the original proposal is fine; we can always adjust it later if needed when we solve the low-density batch problem.

jinfengni:
+1

PR #782, opened 2017-03-14
DRILL-5352: Profile parser printing for multi fragments

paul-rogers:
Enhances the recently added ProfileParser to display run times for
queries that contain multiple fragments. (The original version handled
just a single fragment.)

Prints the query in “classic” mode if it is linear, or in the new
semi-indented mode if the query forms a tree.

Also cleans up formatting - removing spaces between parens.


amansinha100:
+1

PR #783, opened 2017-03-14
DRILL-5324: Provide simplified column reader/writer for tests

paul-rogers:
The new "sub-operator" unit test framework provides simple ways to create row sets in code. This PR includes the column accessor code:

* Interfaces for column accessors
* Template for generated implementations
* Base implementation used by the generated code
* Factory class to create the proper reader or writer given a major
type (type and cardinality)
* Utilities for generic access, type conversions, etc.

Many vector types can be mapped to an int for get and set. One key
exception are the decimal types: decimals, by definition, require a
different representation. In Java, that is `BigDecimal`. Added get, set
and setSafe accessors as required for each decimal type that uses
`BigDecimal` to hold data.

Work remains to be done on other complex types: intervals and so on.
This will be added incrementally as work proceeds.

The generated code builds on the `valueVectorTypes.tdd` file, adding
additional properties needed to generate the accessors.

The PR also includes a number of code cleanups done while reviewing
existing code. In particular `DecimalUtility` was very roughly
formatted and thus hard to follow.

Supports Drill’s interval types (INTERVAL, INTERVALDAY,
INTERVALYEAR) in the form of the Joda interval class.

Adds support for Map vectors. Maps are treated as nested tuples and are
expanded out to create a flattened row in the schema. The accessors
then access rows using the flattened column index or the combined name
(“a.b”).

Supports arrays via a writer interface that appends values as written,
and an indexed, random-access reader interface.

### DIFF
--- a/exec/jdbc-all/pom.xml
+++ b/exec/jdbc-all/pom.xml
@@ -143,6 +143,10 @@
           <artifactId>hazelcast</artifactId>
           <groupId>com.hazelcast</groupId>
         </exclusion>
+        <exclusion>
+	      <groupId>nl.basjes.parse.httpdlog</groupId>
+	      <artifactId>httpdlog-parser</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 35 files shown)


amansinha100:
@paul-rogers can you squash the commits for merging ?

paul-rogers:
Done. Also rebased on master.

amansinha100:
This PR combines the main changes for simplified column readers/writers with the code cleanup/hygiene changes which made it harder to review.  For future, please separate the hygiene changes into a separate PR.  

Rest LGTM.   +1

PR #784, opened 2017-03-14
DRILL-5355: Misc. code cleanup

paul-rogers:
Recent work in Drill has identified a number of cases where code can be
cleaned up: adding missing annotations, etc. These changes don't fit as part of a separate ticket and so are rolled up into a this "general
hygiene" PR.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/PrelVisualizerVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/PrelVisualizerVisitor.java
@@ -30,6 +30,17 @@ import org.apache.drill.exec.planner.physical.WriterPrel;
  * inspection. Insert this into code during development to see
  * the state of the tree at various points of interest during
  * the planning process.
+ * <p>
+ * Use this by inserting lines into our prel transforms to see
+ * what is happening. This is useful if you must understand the transforms,
+ * or change them. For example:
+ * <p>
+ * In file: {@link DefaultSqlHandler#convertToPrel()}:
+ * <pre><code>
+ * PrelVisualizerVisitor.print("Before EER", phyRelNode); // Debug only
+ * phyRelNode = ExcessiveExchangeIdentifier.removeExcessiveEchanges(phyRelNode, targetSliceSize);
+ * PrelVisualizerVisitor.print("After EER", phyRelNode); // Debug only
+ * <code></pre>
  */
 
 public class PrelVisualizerVisitor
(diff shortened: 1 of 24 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/PrelVisualizerVisitor.java
```diff
@@ -0,0 +1,217 @@
...
+import org.apache.drill.exec.planner.physical.WriterPrel;
+
+/**
+ * Debug-time class that prints a PRel tree to the console for
+ * inspection. Insert this into code during development to see
+ * the state of the tree at various points of interest during
+ * the planning process.
+ */
+
+public class PrelVisualizerVisitor
```
chunhui-shi: Would like to see a unit test for this class.

logical/src/main/java/org/apache/drill/common/expression/FieldReference.java
```diff
@@ -1,4 +1,4 @@
-/**
```
chunhui-shi: I have seen you changed '/**' to '/*' in several pull requests. What is the reason for that? I saw other Apache projects, e.g. Hive, use the previous style.

paul-rogers: In Java, comments of the form /** ... */ are Javadoc comments; they are copied into the Javadoc generated for the code. If we put the copyright in such comments, it will appear at the top of the Javadoc file for the class.

Most projects do no include a copyright notice at the top of each Javadoc file (at least not in the visible HTML shown to readers.) Our copyright notice is not formatted for Javadoc, so it is unlikely that our goal is to include the notice in Javadoc.

By contrast, the /* ... */ style comment is a simple code comment, not copied into Javadoc. So, to avoid showing the copyright in Javadoc, we use the plain-comment style instead.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/PrelVisualizerVisitor.java
paul-rogers: Hard to unit test this class... It's purpose is to visualize a "prel" at a various stages in the transform.

```
 PrelVisualizerVisitor.print("Before EER", phyRelNode); // Debug only
 phyRelNode = ExcessiveExchangeIdentifier.removeExcessiveEchanges(phyRelNode, targetSliceSize);
 PrelVisualizerVisitor.print("After EER", phyRelNode); // Debug only
```

I did, however, add a comment to the top of the class explaining how to use the class, showing the above example.

sudheeshkatkam:
+1

PR #786, opened 2017-03-17
DRILL-5359: Fix ClassCastException when Drill pushes down filter on t…

jinfengni:
…he output of flatten operator.

- Move findItemOrFlatten as a static method in DrillRelOptUtil.
- Exclude filter conditions if they contain item/flatten operator.


amansinha100:
+1

PR #787, opened 2017-03-19
DRILL-5319: Refactor "contexts" for unit testing

paul-rogers:
This PR is purely a refactoring: no functionality is added or changed.
The refactoring splits various context and related classes into a set
of new interfaces with needed for operator-level unit tests. The other,
Drillbit-related methods are left in the original interfaces. Most code
need not change.

The changes here allow operator-level unit tests to mock up the
exec-time methods so we can use them without firing up a Drillbit (or
using mocking libraries).

A later PR will provide the sub-operator test framework that uses this
refactoring.

Changes include:

* The OptionManager is split, with read-only methods moving to a new
OptionSet interface.
* The FragmentContext is split, with an exec-only FragmentExecContext
proving low-level methods.
* OperatorStats is split, with a new OperatorStatReceiver class
providing write-only support to operators.
* Several places that accepted an OperatorContext or FragmentContext,
but needed only an allocator, are changed to accept the allocator
directly.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentExecContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentExecContext.java
@@ -1,0 +1,131 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.ops;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.apache.drill.common.config.DrillConfig;
+import org.apache.drill.exec.exception.ClassTransformationException;
+import org.apache.drill.exec.expr.ClassGenerator;
+import org.apache.drill.exec.expr.CodeGenerator;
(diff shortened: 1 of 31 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentExecContext.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.drill.exec.expr.CodeGenerator;
+import org.apache.drill.exec.expr.fn.FunctionImplementationRegistry;
+import org.apache.drill.exec.server.options.OptionSet;
+import org.apache.drill.exec.testing.ExecutionControls;
+
+public interface FragmentExecContext {
+  FunctionImplementationRegistry getFunctionRegistry();
+  OptionSet getOptionSet();
+
+  <T> T getImplementationClass(final ClassGenerator<T> cg)
```
gparai: Good to have comments describing the interface methods

paul-rogers: Was preserving the terseness of the original code, but sure, added comments.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/OptionSet.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.server.options;
+
+/**
+ * Immutable set of options accessible by name or validator.
+ */
+
+public interface OptionSet {
```
paul-rogers: Good idea. If I may, I'd like to defer the actual renaming as doing so would impact a number of in-flight PRs. The suggestion of two interfaces is good. I'll do a follow-on PR to rename this one and split out the writer one.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/TypeValidators.java
```diff
@@ -90,19 +90,19 @@ public void validate(final OptionValue v, final OptionManager manager) {
   }
 
   public static class MinRangeDoubleValidator extends RangeDoubleValidator {
-    private final double min;
-    private final double max;
+//    private final double min;
```
paul-rogers: Fixed.

gparai:
Thanks for the detailed and clear comments.
+1 LGTM

arina-ielchiieva:
Please squash commits and resolve conflict.

paul-rogers:
Done.

PR #788, opened 2017-03-19
DRILL-5318: Sub-operator test fixture

paul-rogers:
This commit depends on:

* DRILL-5319
* DRILL-5323
* DRILL-5324

This PR cannot be accepted (or built) until the above are pulled and
this PR is rebased on top of them. The PR is issued now so that reviews
can be done in parallel.

Provides the following:

* A new OperatorFixture to set up all the objects needed to test at the
sub-operator level. This relies on the refactoring to create the
required interfaces.
* Pulls the config builder code out of the cluster fixture builder so
that configs can be build for sub-operator tests.
* Modifies the QueryBuilder test tool to run a query and get back one
of the new row set objects to allow direct inspection of data returned
from a query.
* Modifies the cluster fixture to create a JDBC connection to the test
cluster. (Use requires putting the Drill JDBC project on the test class
path since exec does not depend on JDBC.)

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/test/ConfigBuilder.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/test/ConfigBuilder.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ ******************************************************************************/
+package org.apache.drill.test;
+
+import java.util.Collection;
+import java.util.Properties;
+import java.util.Map.Entry;
+
+import org.apache.drill.common.config.DrillConfig;
+
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigValueFactory;
+
+/**
+ * Builds a {@link DrillConfig} for use in tests. Use this when a config
+ * is needed by itself, separate from an embedded Drillbit.
+ */
+public class ConfigBuilder {
+
+  protected String configResource;
+  protected Properties configProps;
+
+  /**
+   * Use the given configuration properties as overrides.
+   * @param configProps a collection of config properties
+   * @return this builder
+   * @see {@link #configProperty(String, Object)}
+   */
+
+  public ConfigBuilder configProps(Properties configProps) {
+    if (hasResource()) {
+      // Drill provides no constructor for this use case.
+      throw new IllegalArgumentException( "Cannot provide both a config resource and config properties.");
+    }
+    if (this.configProps == null) {
+      this.configProps = configProps;
+    } else {
+      this.configProps.putAll(configProps);
+    }
+    return this;
+  }
+
+  /**
+   * Use the given configuration file, stored as a resource, to initialize
+   * the Drill config. Note that the resource file should have the two
+   * following settings to work as a config for an embedded Drillbit:
+   * <pre><code>
+   * drill.exec.sys.store.provider.local.write : false,
+   * drill.exec.http.enabled : false
+   * </code></pre>
+   * It may be more convenient to add your settings to the default
+   * config settings with {@link #configProperty(String, Object)}.
+   * @param configResource path to the file that contains the
+   * config file to be read
+   * @return this builder
+   * @see {@link #configProperty(String, Object)}
+   */
+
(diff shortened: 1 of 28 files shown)


exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
```diff
@@ -271,6 +276,91 @@ public QuerySummary run() throws Exception {
...
+   */
+
+  public long singletonLong() throws RpcException {
+    RowSet rowSet = rowSet();
+    if (rowSet == null) {
+      throw new IllegalStateException("No rows returned");
+    }
+    RowSetReader reader = rowSet.reader();
+    reader.next();
+    long value = reader.column(0).getLong();
```
sohami: Can we please refactor the common portion in singletonLong and singletonInt ? Since I guess in future if we support retrieving more types then again it will be repeated?

exec/java-exec/src/test/java/org/apache/drill/test/QueryBuilder.java
```diff
@@ -271,6 +276,91 @@ public QuerySummary run() throws Exception {
...
+   */
+
+  public DirectRowSet rowSet() throws RpcException {
+
+    // Ignore all but the first non-empty batch.
+
+    QueryDataBatch dataBatch = null;
+    for (QueryDataBatch batch : results()) {
+      if (dataBatch == null  &&  batch.getHeader().getRowCount() != 0) {
+        dataBatch = batch;
```
sohami: Ok makes sense after reading this: _"The reason for the funny code is that we want to release memory for all but the first batch."_

sohami:
Thanks for the changes. LGTM. +1

exec/java-exec/src/test/java/org/apache/drill/test/ConfigBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+    if (hasResource()) {
+      return DrillConfig.create(configResource);
+    } else if (configProps != null) {
+      return DrillConfig.create(configProperties());
+    } else {
+      return DrillConfig.create();
+    }
+  }
+
+  private Properties configProperties() {
```
paul-rogers: Note that this has all been replaced by your recent addition...

sudheeshkatkam:
+1

sudheeshkatkam:
This does not compile. Please fix the errors and squash the commits into one.

paul-rogers:
Rebased onto DRILL-5318 so that this branch includes the DRILL-5318 code that this PR depends upon, as well as the additional code required for this PR. Fixed the merge errors that caused the compilation issue.

Please commit DRILL-5318 first. Then, commit this PR. Given that this PR is already rebased on 5318, this PR should merge cleanly into master.

sudheeshkatkam:
I am confused by your comment. Which should be committed first, 5318 or 5323? Your comments says 5318, but the commit message for 5318 and the fact that 5318 is based on 5323 suggests 5323 should be committed first?

Please format the commit message in 70fdc88 so it starts with "DRILL-XXXX: ..."

paul-rogers:
Please see the initial comment which identifies that this PR depends on several others. Two have been done, only DRILL-5323 remains. Please commit DRILL-5323 first, then this one.

Note that, per your request, this PR is rebased onto DRILL-5323. The history here includes all changes from DRILL-5323 plus additional changes specific to DRILL-5318. As I understand it, if you first commit DRILL-5323 the common changes will reside in master, and will be ignored when you then (as a second step) commit this PR.

paul-rogers:
Revised the commit header. Note that the commit messages also identifies that this PR is dependent on DRILL-5323, and so this PR must be committed after DRILL-5323.

Sorry for the confusion. The work was broken into multiple commits to ease the burden on code reviewers. Unfortunately, there is no free lunch, and the multiple commits puts a large burden on the submitter to keep all branches in sync, and on the committer to pull the PRs in the correct order.

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/test/RowSetTest.java
```diff
@@ -0,0 +1,400 @@
...
+  private static OperatorFixture fixture;
+
+  @BeforeClass
+  public static void setUpBeforeClass() throws Exception {
+    fixture = OperatorFixture.standardFixture();
+  }
+
+  @AfterClass
+  public static void tearDownAfterClass() throws Exception {
+    fixture.close();
```
jinfengni: Is this the only unit test added to DRILL-5318 / DRILL-5323 which are for unit test framework related enhancements? Looks like the new tests in this class are to test the new added concept such as "RowSet", "RowSetSchema", "FlattenedSchema". Do we have some example that shows how to use those concepts to do a operator / sub-operator testing, which I thought are the main purpose of the two PRs?

paul-rogers:
One of the key reasons for creating tests is that they show how to use an API. In this case, the associated tests show how to use the row set abstractions to create a schema, populate a row set and so on. Please also see the package-info.java file for Javadoc that provides an overview.

Please let me know if you want to use the test framework. If so, I'd like to work with you to figure out what additional documentation and/or examples would be useful.

This PR is a dependency on the "main show": a PR with refactored sort code and complete sort unit tests. That PR will be submitted this week now that all the prerequisites are in master. That will demonstrate now to use the framework to create a suite of unit tests. I can point to my personal repo if if it would be helpful to preview that code.

jinfengni:
It's fine to have example in a follow-up PR.

I was trying to see if I can use new operator/sub-operator test framework for the schema change task I'm doing. That requires to work with one or multiple RecordBatch (both data and operator).

PR #790, opened 2017-03-20
DRILL-5368: Fix memory leak issue in DrillClientImpl::processServerMetaResult

laurentgo:
Fix a small memory leak by doing local allocation instead since the
object doesn't escape the function.


sudheeshkatkam:
+1

PR #791, opened 2017-03-20
DRILL-5369: Add initializer for ServerMetaContext

laurentgo:
ServerMetaContext had no default constructor. The lack of it
might cause m_done to be set to true, same for other variables.

Add a default constructor to explicitly initialize its members.


superbstreak:
Thanks @laurentgo for the fix!

sudheeshkatkam:
+1

PR #792, opened 2017-03-22
DRILL-4971: Query encounters system error, when there aren't eval subexpressions of any function in boolean and/or expressions

vdiravka:
- New evaluated blocks for boolean operators should be with braces always, since they use labels.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -144,6 +144,9 @@ public class EvaluationVisitor {
     previousExpressions = mapStack.pop();
   }
 
+  /**
+   * Get a HoldingContainer for the expression if it had been already evaluated
+   */
   private HoldingContainer getPrevious(LogicalExpression expression, MappingSet mappingSet) {
     HoldingContainer previous = previousExpressions.get(new ExpressionHolder(expression, mappingSet));
     if (previous != null) {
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
```diff
@@ -671,8 +674,9 @@ private HoldingContainer visitBooleanAnd(BooleanOperator op,
       HoldingContainer out = generator.declare(op.getMajorType());
 
       JLabel label = generator.getEvalBlockLabel("AndOP");
-      JBlock eval = generator.getEvalBlock().block();  // enter into nested block
-      generator.nestEvalBlock(eval);
+      JBlock eval = new JBlock();
```
jinfengni: How about we add the following two methods in ClassGenerator.java: 

```java
  private JBlock createInnerBlock(BlockType type) {
    final JBlock currBlock = getBlock(type);
    final JBlock innerBlock = new JBlock();
    currBlock.add(innerBlock);
    return innerBlock;
  }

  protected JBlock createInnerEvalBlock() {
    return createInnerBlock(BlockType.EVAL);
  }
```

Then, replace `generator.getEvalBlock().block()` with `generator.createInnerEvalBlock()`  ?

vdiravka: It makes sense. 
Added. Thanks.

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestEvaluationVisitor.java
```diff
@@ -72,6 +73,37 @@ public void x() throws Exception{
     logger.debug(g.generateAndGet());
   }
 
+  @Test
+  public void testVisitBooleanOrWithoutFunctionsEvaluation() throws Exception {
```
jinfengni: How about move the new testcase into TestBugFixes? The new testcases uses SQL, while the existing one in TestEvaluationVisitor actually tests in a lower level.  Probably better to not mix them together.

vdiravka: I agree that new test cases use higher level, SQL. The first thought was TestExampleQueries class, but this class is overloaded.
TestBugFixes will be better. Thanks.

jinfengni:
+1

PR #793, opened 2017-03-22
DRILL-4678: Tune metadata by generating a dispatcher at runtime

Serhii-Harnyk:
Changes for rebasing to Calcite 1.4.0-drill-r20

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdRowCount.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdRowCount.java
@@ -1,4 +1,4 @@
-/*******************************************************************************
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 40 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdRowCount.java
```diff
@@ -14,35 +14,71 @@
...
     } else {
-      return super.getRowCount(rel);
+      return super.getRowCount(rel, mq);
     }
   }
 
   @Override
-  public Double getRowCount(Filter rel) {
-    return rel.getRows();
+  public Double getRowCount(Filter rel, RelMetadataQuery mq) {
```
Serhii-Harnyk: In RelMdRowCount.java added overloaded methods, which returns other result than getRowCount(RelNode rel, RelMetadataQuery mq) [1] as it was before. 
So in order to preserve previous Drill behavior, I override those methods.
[1] https://github.com/Serhii-Harnyk/incubator-calcite/blob/02848c99d75f2d0e00c219f3fa300fc8e7df26df/core/src/main/java/org/apache/calcite/rel/metadata/RelMdRowCount.java#L65

PR #794, opened 2017-03-23
DRILL-5375: Nested loop join: return correct result for left join

arina-ielchiieva:
With this fix nested loop join will correctly process INNER and LEFT joins with non-equality conditions.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/BatchReference.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/BatchReference.java
@@ -1,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.expr;
+
+import com.google.common.base.Preconditions;
+
+/**
+ * Holder class that contains batch naming, batch  and record index. Batch index is used when batch is hyper container.
+ * Used to distinguish batches in non-equi conditions during expression materialization.
+ * Mostly used for nested loop join which allows non equi-join.
+ *
+ * BatchReference instance can be created during batch initialization
+ * (ex: instance of {@link org.apache.drill.exec.record.AbstractRecordBatch})
+ * since naming of batches used won't change during data processing.
+ * Though information from batch reference will be used during schema build (i.e. once per OK_NEW_SCHEMA).
+ *
+ * Example:
+ * BatchReference{batchName='leftBatch', batchIndex='leftIndex', recordIndex='leftIndex'}
+ * BatchReference{batchName='rightContainer', batchIndex='rightBatchIndex', recordIndex='rightRecordIndexWithinBatch'}
+ *
+ */
+public final class BatchReference {
+
+  private final String batchName;
+
+  private final String batchIndex;
+
+  private final String recordIndex;
+
+  public BatchReference(String batchName, String recordIndex) {
+    // when batch index is not indicated, record index value will be set instead
+    this(batchName, recordIndex, recordIndex);
+  }
+
+  public BatchReference(String batchName, String batchIndex, String recordIndex) {
+    Preconditions.checkNotNull(batchName, "Batch name should not be null.");
+    Preconditions.checkNotNull(batchIndex, "Batch index should not be null.");
+    Preconditions.checkNotNull(recordIndex, "Record index should not be null.");
+    this.batchName = batchName;
+    this.batchIndex = batchIndex;
+    this.recordIndex = recordIndex;
+  }
+
+  public String getBatchName() {
+    return batchName;
+  }
+
+  public String getBatchIndex() {
+    return batchIndex;
+  }
+
+  public String getRecordIndex() {
+    return recordIndex;
+  }
+
+  @Override
+  public String toString() {
+    return "BatchReference{" +
+        "batchName='" + batchName + '\'' +
+        ", batchIndex='" + batchIndex + '\'' +
+        ", recordIndex='" + recordIndex + '\'' +
+        '}';
+  }
+}
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinBatch.java
@@ -19,9 +19,16 @@ package org.apache.drill.exec.physical.impl.join;
 
 import java.io.IOException;
 import java.util.LinkedList;
+import java.util.Map;
 
+import com.google.common.collect.ImmutableMap;
+import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
+import org.apache.drill.common.expression.ErrorCollector;
+import org.apache.drill.common.expression.ErrorCollectorImpl;
+import org.apache.drill.common.expression.LogicalExpression;
 import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.common.types.Types;
 import org.apache.drill.exec.compile.sig.GeneratorMapping;
 import org.apache.drill.exec.compile.sig.MappingSet;
 import org.apache.drill.exec.exception.ClassTransformationException;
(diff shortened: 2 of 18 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinBatch.java
```diff
@@ -214,26 +226,62 @@ private boolean hasMore(IterOutcome outcome) {
...
+
+    /*
+        Logical expression may contain fields from left and right batches. During code generation (materialization)
+        we need to indicate from which input field should be taken. Mapping sets can work with only one input at a time.
+        But non-equality expressions can be complex:
+          select t1.c1, t2.c1, t2.c2 from t1 inner join t2 on t1.c1 between t2.c1 and t2.c2
+        or even contain self join which can not be transformed into filter since OR clause is present
+          select *from t1 inner join t2 on t1.c1 >= t2.c1 or t1.c3 <> t1.c4
+
+        In this case logical expression can not be split according to input presence (like during equality joins
```
amansinha100: I am not sure the distinction is accurate.  Even for equality condition, you could have in your example above the columns coming from the same table.  i.e instead of  ...OR t1.c3 <> t1.c4  one could have OR t1.c3 = t1.c4 which is a valid condition.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillOptiq.java
```diff
@@ -70,27 +70,65 @@
...
+         to retrieve field from filed list by index in {@link #visitInputRef(RexInputRef)} method. Example:
+
+         Query: select t1.c1, t2.c1. t2.c2 from t1 inner join t2 on t1.c1 between t2.c1 and t2.c2
+
+         Input 1: $0
+         Input 2: $1, $2
+
+         Result: $0, $1, $2
+       */
+      for (RelNode input : inputs) {
```
arina-ielchiieva: I have only allowed to pass multiple inputs instead of one as it needed to non-equi joins plus I have merged all fields into one list from all inputs for performance improvement. You are correct it's Calcite that ensures fields numeration. I have just added comment explaining this behavior to ease code review and as some help for future developers.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/PlannerSettings.java
```diff
@@ -105,6 +103,29 @@
...
+      join will be optimized: left and right inputs will be flipped and LEFT join type will be changed to RIGHT one.
+      If query contains non-equi joins, after such optimization it will fail, since nested loop does not allow
+      RIGHT join. In this case if user accepts probability of non optimal performance, he may turn off join optimization.
+      Turning off join optimization, makes sense only if user are not sure that right output is less or equal to left,
+      otherwise join optimization can be left turned on.
+
+      Note: once hash and merge joins will allow non-equi join conditions,
+      the need to turn off join optimization may go away.
+   */
+  public static final BooleanValidator JOIN_OPTIMIZATION = new BooleanValidator("planner.enable_join_optimization", true);
```
arina-ielchiieva: JOIN_OPTIMIZATION enables two rules `DRILL_JOIN_TO_MULTIJOIN_RULE` and `DRILL_LOPT_OPTIMIZE_JOIN_RULE` which are applicable for any types of joins. That's why naming is quite broad, I believe these two rules are not only responsible for join swap but for all other join optimization techniques. In our use case, user may want to disable them when he doesn't won't join swap to be performed but there may other reasons. Though as I have noted, when we implement non-equality joins for hash and merge joins, we may remove this configuration parameter.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinBatch.java
arina-ielchiieva: The thing is that inequality join is not only join that has `t1.c3 <> t1.c4` but also the one that has `OR`.
For example, currently the following query `select * from t1 inner join t2 on t1.c1 = t2.c1 or t1.c2 = t2.c2` will fail which the following error: `UNSUPPORTED_OPERATION ERROR: This query cannot be planned possibly due to either a cartesian join or an inequality join`.

The main idea of my comment is that I don't bother if it's equality or inequality join, I just materialize the whole expression with fields from two inputs and to find out which input field is I add batch indication. If you want I can remove the comment, if it's confusing.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinTemplate.java
```diff
@@ -40,132 +41,133 @@
...
+    int nextRightRecordToProcess = tracker.getNextRightRecordToProcess();
+    int nextLeftRecordToProcess = tracker.getNextLeftRecordToProcess();
+    boolean rightRecordMatched = tracker.isRightRecordMatched();
+
+    outer:
+    // for every record in the left batch
+    for (; nextLeftRecordToProcess < leftRecordCount; nextLeftRecordToProcess++) {
+      // for every batch on the right
+      for (; nextRightBatchToProcess < rightCounts.size(); nextRightBatchToProcess++) {
+        int rightRecordCount = rightCounts.get(nextRightBatchToProcess);
```
amansinha100: For keeping track of the number of records on the right that are processed due to the O(n^2) join, we should have a  non-local metric which would be a Long.  Currently, it seems the query profile for NLJoin will be incorrect.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/BatchReference.java
```diff
@@ -0,0 +1,73 @@
...
+ * Holder class that contains batch naming, batch  and record index. Batch index is used when batch is hyper container.
+ * Used to distinguish batches in non-equi conditions during expression materialization.
+ * Mostly used for nested loop join which allows non equi-join.
+ *
+ * Example:
+ * BatchReference{batchName='leftBatch', batchIndex='leftIndex', recordIndex='leftIndex'}
+ * BatchReference{batchName='rightContainer', batchIndex='rightBatchIndex', recordIndex='rightRecordIndexWithinBatch'}
+ *
+ */
+public final class BatchReference {
```
amansinha100: Can you add some comments about when the batch reference should be set ? for every batch or once per OK_NEW_SCHEMA status ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillOptiq.java
amansinha100: I guess I am still unclear about the reason for merging the multiple lists into one.  Later, you will have to determine which input each $n reference belonged to, isn't it ?

amansinha100:
A general comment about filter evaluation: In HashJoin and MergeJoin we do implicit casting of one of the join columns to process the join condition (see invocations of JoinUtils.addLeastRestrictiveCasts()).  Since you are adding filter evaluation to NestedLoopJoin, I would think we need this there as well.  However, I am ok if you want to create a separate JIRA for doing that.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillOptiq.java
arina-ielchiieva: Yes, in `public LogicalExpression visitInputRef(RexInputRef inputRef)` we determine to which input field belongs to. Before that we had only one input thus we did simple get operation `input.getRowType().getFieldList().get(index)` but now we have two inputs so we have to get operation on one input and if field in not found try in the second.  I could iterate over two inputs and do get operation and once filed is found break the loop OR I could merge filed list in one and do simple get operation `fieldList.get(index)`. For performance reasons, I decided to merge filed lists in constructor and use them in `public LogicalExpression visitInputRef(RexInputRef inputRef)` rather than iterating over them for each field.

arina-ielchiieva:
Thanks for bringing up this point. I have done some investigation and found out that implicit casts for nested loop join are already included during materialization.
Join condition is transformed into FunctionCall [1] which is later on materialized using ExpressionTreeMaterializer [2]. 
`ExpressionTreeMaterializer.visitFunctionCall` includes section which implicit casts [3].
Actually these casts are more enhanced that during hash and merge joins.
For example, during hash and merge joins only casts between numeric types, date and timestamp, varchar and varbinary are supported, i.e.
join by int and varchar columns won't be performed. The following error will be returned: `Join only supports implicit casts between 1. Numeric data
 2. Varchar, Varbinary data 3. Date, Timestamp data Left type: INT, Right type: VARCHAR. Add explicit casts to avoid this error`. 
In our case nested loop join will be able to perform join by int and varchar columns without adding explicit casts.

[1] https://github.com/arina-ielchiieva/drill/blob/71628e70a525d9bd27b4f5f56259dce84c75154d/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/NestedLoopJoinPrel.java#L95
[2] https://github.com/arina-ielchiieva/drill/blob/71628e70a525d9bd27b4f5f56259dce84c75154d/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinBatch.java#L265
[3] https://github.com/apache/drill/blob/9411b26ece34ed8b2f498deea5e41f1901eb1013/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ExpressionTreeMaterializer.java#L362

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinTemplate.java
amansinha100: Does that mean the query profile correctly show NxM rows processed in the nested loop join ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/PlannerSettings.java
amansinha100: Ah, you added this option to enable/disable the *logical* join rules.  Since NestedLoopJoin is a physical join implementation, from the comments I interpreted that this was intended for the swapping of left and right inputs of the (physical) NL join, which is why I mentioned about hashjoin_swap option.   It seems to me that if there is an LEFT OUTER JOIN and condition is non-equality, then we should not allow changing to a Right Outer Join by flipping the left and right sides, since that would make the query fail.   What do you think ?
I suppose we could keep your boolean option for this PR and address the left outer join issue separately.

amansinha100:
A few follow-up comments which are non-blockers.  Overall LGTM.  +1

PR #795, opened 2017-03-24
DRILL-5089: Get only partial schemas of relevant storage plugins inst…

chunhui-shi:
…ead of all storages ahead.

1. For each query, rootSchema is empty, add schemas of a storage plugin only when needed -- when asked for a schemaPath.

2. Allow mock environments to provide dynamic schemas.

3. SchemaUtils.findSchema used in many Sql handlers now is handled by SqlConverter.catalog to allow expand schema dynamically

4. Temp table resolve also could not take a schema tree and assume it is complete.

NOTE: pom.xml in this pull request is pointing to temporary calcite versoin 'r20-test', this should change to 'r20' once this pull request is ready to commit

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
@@ -20,6 +20,7 @@ package org.apache.drill.exec.ops;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.calcite.schema.SchemaPlus;
 import org.apache.drill.common.AutoCloseables;
@@ -144,6 +145,25 @@ public class QueryContext implements AutoCloseable, OptimizerRulesContext, Schem
     return defaultSchema;
   }
 
+  public SchemaPlus getPartialDefaultSchema() {
+
+    final String sessionSchemaPath = session.getDefaultSchemaPath();
+    final List<String> schemaPathAsList = Lists.newArrayList(sessionSchemaPath.split("\\."));
+
+    final SchemaPlus rootSchema = schemaTreeProvider.createPartialRootSchema(getQueryUserName(),
+        this, schemaPathAsList.get(0));
+
+    final SchemaPlus defaultSchema = session.getDefaultSchema(rootSchema);
+
+    if (defaultSchema == null) {
+      return rootSchema;
+    }
+    return defaultSchema;
+  }
+
+  public void addNewRelevantSchema(Set<String> storages, SchemaPlus toExpandSchema) {
+    schemaTreeProvider.addPartialRootSchema(getQueryUserName(), this, storages, toExpandSchema);
+  }
   /**
    * Get root schema with schema owner as the user who issued the query that is managed by this QueryContext.
    * @return Root of the schema tree.
@@ -251,6 +271,10 @@ public class QueryContext implements AutoCloseable, OptimizerRulesContext, Schem
 
   public RemoteFunctionRegistry getRemoteFunctionRegistry() {
     return drillbitContext.getRemoteFunctionRegistry();
+  }
+
+  public SchemaTreeProvider getSchemaTreeProvider() {
+    return schemaTreeProvider;
   }
 
   @Override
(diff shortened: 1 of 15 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/SchemaTreeProvider.java
```diff
@@ -119,6 +127,74 @@ public SchemaPlus createRootSchema(SchemaConfig schemaConfig) {
...
+        try {
+          StoragePlugin plugin = dContext.getStorage().getPlugin(name);
+          if(plugin == null) {
+            logger.error("Got null for storage plugin of name: " + name);
+            continue;
+          }
+          plugin.registerSchemas(schemaConfig, rootSchema);
+          expandSecondLevelSchema(rootSchema);
+        }
+        catch (ExecutionSetupException ex) {
```
paul-rogers: Presumably this method adds storages based on need: we add a storage only if a query references it. If so, then we need to pass the exception along. If we don't, the user will simply be told that "foo.bar" does not exist when, in fact, perhaps it does exist but is configured wrong.

And, note that we do catch an IOException and translate it to a user exception below.

So, maybe

* Remove this try/catch block.
* Change the following catch from IOException to just Exception.

The result will be that we fail the query for all errors, not just IO errors. Again, this is the right choice if we are only loading the schemas needed by the query.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -144,6 +145,25 @@ public SchemaPlus getNewDefaultSchema() {
     return defaultSchema;
   }
 
+  public SchemaPlus getPartialDefaultSchema() {
```
paul-rogers: Javadoc to explain what a partial default schema might be?

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -144,6 +145,25 @@ public SchemaPlus getNewDefaultSchema() {
...
+    final SchemaPlus defaultSchema = session.getDefaultSchema(rootSchema);
+
+    if (defaultSchema == null) {
+      return rootSchema;
+    }
+    return defaultSchema;
+  }
+
+  public void addNewRelevantSchema(Set<String> storages, SchemaPlus toExpandSchema) {
+    schemaTreeProvider.addPartialRootSchema(getQueryUserName(), this, storages, toExpandSchema);
```
paul-rogers: Per a comment later on, find a good home for computing the effective user name (with impersonation enabled). That calculation should probably be done on the user session if the impersonation is for the life of the session, or on the query context if the impersonation is only for the single query. (Note that if it is only for a query, temp tables probably won't work the way we expect.)

But, doing the impersonation check in the schema tree provider is pushing that behavior too far down into the code.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -253,6 +273,10 @@ public RemoteFunctionRegistry getRemoteFunctionRegistry() {
     return drillbitContext.getRemoteFunctionRegistry();
   }
 
+  public SchemaTreeProvider getSchemaTreeProvider() {
```
paul-rogers: Javadoc comment to explain this?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillCatalogReader.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.drill.common.exceptions.UserException;
+import org.apache.drill.exec.ops.QueryContext;
+import org.apache.drill.exec.store.SchemaConfig;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Implementation of {@link org.apache.calcite.prepare.Prepare.CatalogReader}
```
paul-rogers: Maybe explain the purpose and not just the implementation? What part of query processing does this address? Does this resolve references to schemas and workspaces? Then the schema resolves the reference to a table name?

paul-rogers:
Is this PR ready for final review? Or, on hold waiting for something?

PR #796, opened 2017-03-25
DRILL-5365: DrillFileSystem setConf in constructor. DrillFileSystem c…

chunhui-shi:
…ould be created based on provided URI.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
@@ -89,22 +89,36 @@ public class DrillFileSystem extends FileSystem implements OpenFileTracker {
   }
 
   public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws IOException {
-    this.underlyingFs = FileSystem.get(fsConf);
+    this(fsConf, URI.create(fsConf.getRaw(FS_DEFAULT_NAME_KEY)), operatorStats);
+  }
+
+  public DrillFileSystem(Configuration fsConf, URI Uri, OperatorStats operatorStats) throws IOException {
+    this.underlyingFs = FileSystem.get(Uri, fsConf);
+    logger.trace("Configuration for the DrillFileSystem " + fsConf.getRaw(FS_DEFAULT_NAME_KEY) +
+        ", underlyingFs: " + this.underlyingFs.getUri());
     this.codecFactory = new CompressionCodecFactory(fsConf);
     this.operatorStats = operatorStats;
+    setConf(fsConf);
   }
 
   @Override
   public void setConf(Configuration conf) {
     // Guard against setConf(null) call that is called as part of superclass constructor (Configured) of the
     // DrillFileSystem, at which point underlyingFs is null.
-    if (conf != null && underlyingFs != null) {
-      underlyingFs.setConf(conf);
+    if(conf != null) {
+      super.setConf(conf);
+      if (underlyingFs != null && underlyingFs.getConf() == null) {
+        underlyingFs.setConf(conf);
+      }
+
     }
   }
 
   @Override
   public Configuration getConf() {
+    if (super.getConf() != null) {
+      return super.getConf();
+    }
     return underlyingFs.getConf();
   }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
@@ -74,6 +74,7 @@ public class FileSystemPlugin extends AbstractStoragePlugin{
           fsConf.set(s, config.config.get(s));
         }
       }
+      fsConf.set("fs.default.name", config.connection);
       fsConf.set(FileSystem.FS_DEFAULT_NAME_KEY, config.connection);
       fsConf.set("fs.classpath.impl", ClassPathFileSystem.class.getName());
       fsConf.set("fs.drill-local.impl", LocalSyncableFileSystem.class.getName());


paul-rogers:
Would be incredibly useful to provide a description of the problem and fix: either here in the PR or as a comment in the JIRA. Saves having the reviewer having to duplicate your debugging and diagnostic efforts. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -89,22 +89,36 @@ public DrillFileSystem(Configuration fsConf) throws IOException {
   }
 
   public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws IOException {
-    this.underlyingFs = FileSystem.get(fsConf);
+    this(fsConf, URI.create(fsConf.getRaw(FS_DEFAULT_NAME_KEY)), operatorStats);
```
paul-rogers: Actually, why do we need to do this step? The original code seems to do the right thing:

```
class FileSystem ...
  public static FileSystem get(Configuration conf) throws IOException {
    return get(getDefaultUri(conf), conf);
  }
  public static URI getDefaultUri(Configuration conf) {
    return URI.create(fixName(conf.get(FS_DEFAULT_NAME_KEY, DEFAULT_FS)));
  }
```

That is, the original code gets the file system using the URI stored in the config. Standard practice is that the caller must have set the file system property: that is how we tell a "file:///" system from an HDFS system, etc.

So, isn't the problem here with the caller?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -89,22 +89,36 @@ public DrillFileSystem(Configuration fsConf) throws IOException {
...
+  }
+
+  public DrillFileSystem(Configuration fsConf, URI Uri, OperatorStats operatorStats) throws IOException {
```
paul-rogers: This constructor is added to allow passing a URI. But, the rules for `FileSystem` (see above), is that the URI is stored in the config as a property. So, is this constructor needed? (One argument for using it is that the base class, `FileSystem`, does provide this form.)

Also, this constructor is new. But there is no code in this PR that calls this constructor. So, how does this new (unused) constructor fix the bug? Perhaps a bit of explanation would clear up the mystery...

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -89,22 +89,36 @@ public DrillFileSystem(Configuration fsConf) throws IOException {
...
+    this.underlyingFs = FileSystem.get(Uri, fsConf);
+    logger.trace("Configuration for the DrillFileSystem " + fsConf.getRaw(FS_DEFAULT_NAME_KEY) +
+        ", underlyingFs: " + this.underlyingFs.getUri());
     this.codecFactory = new CompressionCodecFactory(fsConf);
     this.operatorStats = operatorStats;
+    setConf(fsConf);
   }
 
   @Override
   public void setConf(Configuration conf) {
```
paul-rogers: This is existing code. But, the existence of this method is probably very poor design. This method says that we create a file system using one config, then change the config later. Does the class work if we start with, say, "file:///" then change to HDFS? The `setConf()` method should probably be removed and code in this class should say that the config is immutable once the `DrillFileSystem` is created. Else, all bets are off on semantics.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -89,22 +89,36 @@ public DrillFileSystem(Configuration fsConf) throws IOException {
...
+    this(fsConf, URI.create(fsConf.getRaw(FS_DEFAULT_NAME_KEY)), operatorStats);
+  }
+
+  public DrillFileSystem(Configuration fsConf, URI Uri, OperatorStats operatorStats) throws IOException {
+    this.underlyingFs = FileSystem.get(Uri, fsConf);
+    logger.trace("Configuration for the DrillFileSystem " + fsConf.getRaw(FS_DEFAULT_NAME_KEY) +
+        ", underlyingFs: " + this.underlyingFs.getUri());
     this.codecFactory = new CompressionCodecFactory(fsConf);
     this.operatorStats = operatorStats;
+    setConf(fsConf);
```
paul-rogers: I think this whole code block needs more thought. DrillFileSystem derives from the Hadoop `FileSystem` class, but we never invoke the `FileSystem` constructor. Why? Without proper init, can we guarantee that the methods defined in the base class actually work?

On the other hand, it seems we've reimplemented all base class methods, basically masking the base class implementation. If that is true, why set the config on the base class?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -89,22 +89,36 @@ public DrillFileSystem(Configuration fsConf) throws IOException {
...
+      if (underlyingFs != null && underlyingFs.getConf() == null) {
+        underlyingFs.setConf(conf);
+      }
+
     }
   }
 
   @Override
   public Configuration getConf() {
+    if (super.getConf() != null) {
```
paul-rogers: Shouldn't the base class config and the underlying FS config be the same? If they differ, how do we have any idea which is the correct one? In what situation would we want the base class config but not the underlying FS config?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
```diff
@@ -74,6 +74,7 @@ public FileSystemPlugin(FileSystemConfig config, DrillbitContext context, String
           fsConf.set(s, config.config.get(s));
         }
       }
+      fsConf.set("fs.default.name", config.connection);
```
paul-rogers: Please use `FileSystem.FS_DEFAULT_NAME_KEY`. It seems that `"fs.default.name"` is deprecated.

But, note, that is exactly what the next line of code does. So, perhaps there is more going on here than a first read suggests? Or, are we confusing ourselves about the meaning of the two properties?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
chunhui-shi: Yes, this can be removed

paul-rogers:
Is this PR ready for final review? Or, on hold waiting for something?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
ilooner: Doing a quick google it looks like you are correct. **fs.default.name** is deprecated and we should use **fs.defaultFS** instead https://stackoverflow.com/questions/30480847/difference-between-fs-defaultfs-and-fs-default-name?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa. Also hortonworks dropped **fs.default.name** from Ambari a long time ago so we should be safe https://issues.apache.org/jira/browse/AMBARI-2789 .

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: Removed this constructor

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: Removed the other constructor and am creating underlying filesystem directly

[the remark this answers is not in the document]
ilooner: This constructor is removed now, so the only thing getting this value is a logging statement now. I made the logging statement use getTrimmed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: Agreed. Unfortunately setConf() is a public method declared in hadoop's FileSystem class. Additionally setConf() is called with a null argument everytime DrillFileSystem is initialized because the FileSystem constructor calls the Configured(Configuration conf) constructor and passes null for the configuration.

What I can do is make the semantics of DrillFileSystem explicit by documenting that DrillFileSystem.setConf should never be called and by throwing an exception if DrillFileSystem.setConf is ever called after a DrillFileSystem object is constructed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: Agreed I will remove this call, and make setConf effectively a noop which will throw an exception if anyone calls it.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: No idea, this is extremely confusing. I will just return the conf on the underlying file system. I'll update the other PR if I observe any regressions when I do this.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
ilooner: Okay I think I have a better handle on this now. The original issue was that Drill's hive storage plugin had a configuration option of fs.default.name = file:// . Somehow when a hive table was dropped and then recreated with a ctas statement in drill, the CTAS statement picked up the fs.default.name configuration from the hive storage plugin and passed that on to DrillFileSystem. And apparently if both fs.default.name and fd.defaultFS are present with different values the value for fs.default.name wins even though it is deprecated. So the CTAS statement would end up creating the table on a drill node's local filesystem.

I believe the crux of this PR is to force "fs.default.name" to have to correct value in the event that a different value is defined in the HiveStorage plugin.

With that said, there are several questions. 

 1. How the heck does a property in the HiveStoragePlugin make it's way into the FileSystem configuration? I spent a good amount of time looking at the code and for the life of me I can't figure that out.
 2. The follow up to (1) is do we actually want that behavior? We can force fs.default.name to have the right value but what about other properties we might suck in from a HiveStoragePlugin configuration?
 3. If we don't want this behavior what would be the real fix?

In the face of all this ambiguity I think we should move forward with a minimal PR that forces fs.default.name to be correct now. We can have a follow up Jira that actually fixes the underlying problem of sucking in stray configs down the road if someone complains about it.

PR #797, opened 2017-03-25
DRILL-5286: No need to convert when the relNode and target candidate …

chunhui-shi:
…set are the same

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/SubsetTransformer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/SubsetTransformer.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.planner.physical;
 
+import com.google.common.collect.Sets;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.plan.ConventionTraitDef;
 import org.apache.calcite.plan.RelOptRule;
@@ -24,6 +25,8 @@ import org.apache.calcite.plan.RelOptRuleCall;
 import org.apache.calcite.plan.RelTrait;
 import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.plan.volcano.RelSubset;
+
+import java.util.Set;
 
 public abstract class SubsetTransformer<T extends RelNode, E extends Exception> {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SubsetTransformer.class);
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/SubsetTransformer.java
```diff
@@ -45,15 +48,20 @@ public RelTraitSet newTraitSet(RelTrait... traits) {
...
     if ( !(candidateSet instanceof RelSubset) ) {
       return false;
     }
 
     boolean transform = false;
+    Set<RelNode> transformedRels = Sets.newHashSet();
     for (RelNode rel : ((RelSubset)candidateSet).getRelList()) {
       if (isPhysical(rel)) {
         RelNode newRel = RelOptRule.convert(candidateSet, rel.getTraitSet().plus(Prel.DRILL_PHYSICAL));
+        if(transformedRels.contains(newRel)) {
```
paul-rogers: Does this work? A hash map depends on `hashCode()` and `equals()`. A random check of one `DrillRelNode`, `DrillFilterRelBase`, shows that the above two methods are not implemented. As a result, set membership may be based on object identity. But, surely each call to `convert()` will create a new object?

Or, does this rule apply to only certain kinds of nodes which do have `hashCode()` and `equals()` implemented? If so, perhaps add some comments to identify the subtree of nodes with which this code works.

Do we have a test case for this code? Or, did you step through it to see if it was doing what we want? If so, and it worked, then likely this code does work with a specific subset of nodes that have the needed methods.

chunhui-shi: this " if(transformedRels.contains(newRel))" is to check if the newRel is the _same_ object we got before, it is not to check a 'equal' node. So no hashCode or equal function required.

And yes, I did step into this code and it is doing what we want: for the same node(n) to be converted in below convertChild() and the same equivalent set(newRel), we don't need to run the rule again, since we are going to get the same result, and that result will be added to the same set. So we are very sure we could skip convertChild() for this pair of input(n, newRel) since this has been called before.

In some cases it can reduce the same rule from running 16 times to 4 times and saved hundreds ms.

paul-rogers: No doubt that time can be saved by avoiding repeated work. When I was tinkering with some Calcite rules I saw the same node being analyzed many, many times for even a simple query.

Since you want to use identify (not equality) to compare hash map entries, consider using `Sets.newIdentityHashSet()` in place of the equality-based `Sets.newHashSet()`.

More efficient would be if we knew the type of the `RelNode`s. If we did, and they all derive from a single Drill type, we could just add an `isTransformed` flag to avoid the set overhead.

chunhui-shi: Change to use newIdentityHashSet. 

And we could not simply mark a node is isTransformed, since there could be many rules transforming the same node. Maintaining a map of applied rules in relNode might not be a good idea since remembering what rule has been applied should not be the job of a relNode(set). So what about maintaining such map in planner? Not a good idea since the same node might need to be converted into different sets. So I would think leaving this part as is.

PR #800, opened 2017-03-26
DRILL-5385: Vector serializer fails to read saved SV2

paul-rogers:
Unit testing revealed that the VectorAccessorSerializable class claims
to serialize SV2s, but, in fact, does not. Actually, it writes them,
but does not read them, resulting in corrupted data on read.

Fortunately, no code appears to serialize sv2s at present. Still, it is
a bug and needs to be fixed.

First task is to add serialization code for the sv2.

That revealed that the recently-added code to save DrillBufs using a
shared buffer had a bug: it relied on the writer index to know how much
data is in the buffer. Turns out sv2 buffers don’t set this index. So,
new versions of the write function takes a write length.

Then, closer inspection of the read code revealed duplicated code. So,
DrillBuf allocation moved into a version of the read function that now
does reading and DrillBuf allocation.

Turns out that value vectors, but not SV2s, can be built from a
Drillbuf. Added a matching constructor to the SV2 class.

Finally, cleaned up the code a bit to make it easier to follow. Also
allowed test code to access the handy timer already present in the code.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorAccessibleSerializable.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorAccessibleSerializable.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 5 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorAccessibleSerializable.java
```diff
@@ -146,36 +157,24 @@ public void writeToStream(OutputStream output) throws IOException {
...
       batchDef.writeDelimitedTo(output);
 
       /* If we have a selection vector, dump it to file first */
-      if (svBuf != null) {
-        allocator.write(svBuf, output);
-        sv2.setBuffer(svBuf);
-        svBuf.release(); // sv2 now owns the buffer
-        sv2.setRecordCount(svCount);
+      if (svMode == BatchSchema.SelectionVectorMode.TWO_BYTE) {
+        recordCount = sv2.getCount();
```
Ben-Zvi: Ultra minor comment: When SV2 is read, the recordCount comes from the batchDef; here it is taken from the sv2 itself (they should be the same anyway ....)

PR #801, opened 2017-03-27
DRILL-5378: Put more information for schema change exception in hash …

jinfengni:
…join, hash agg and sort operator.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/exception/SchemaChangeException.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/exception/SchemaChangeException.java
@@ -18,6 +18,7 @@
 package org.apache.drill.exec.exception;
 
 import org.apache.drill.common.exceptions.DrillException;
+import org.apache.drill.exec.record.BatchSchema;
 
 public class SchemaChangeException extends DrillException{
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SchemaChangeException.class);
@@ -48,5 +49,18 @@ public class SchemaChangeException extends DrillException{
 
   public SchemaChangeException(String message, Throwable cause, Object...objects){
     super(String.format(message, objects), cause);
+  }
+
+  public static SchemaChangeException schemaChanged(String message, BatchSchema priorSchema, BatchSchema newSchema) {
+    final String errorMsg = new StringBuilder()
+        .append(message)
+        .append("\n")
+        .append("Prior schema : \n")
+        .append(priorSchema.toString())
+        .append("\n")
+        .append("New schema : \n")
+        .append(newSchema.toString())
+        .toString();
+    return new SchemaChangeException(errorMsg);
   }
 }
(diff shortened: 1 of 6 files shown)


amansinha100:
+1

jinfengni:
Agreed with @paul-rogers 's comment regarding overuse SchemaChangeException in the code. I actually planned to open a jira, if no one has been opened yet. We currently use SchemaChangeException in many situations, including real schema change or operational error for "something is wrong"

exec/java-exec/src/main/java/org/apache/drill/exec/exception/SchemaChangeException.java
```diff
@@ -49,4 +50,16 @@ public SchemaChangeException(String message, Object...objects){
   public SchemaChangeException(String message, Throwable cause, Object...objects){
     super(String.format(message, objects), cause);
   }
+
+  public static SchemaChangeException schemChanged(String message, BatchSchema priorSchema, BatchSchema newSchema) {
```
gparai: Please correct the typo: schemChanged to schemaChanged

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -278,7 +281,7 @@ public void setup(HashAggregate hashAggrConfig, HashTableConfig htConfig, Fragme
   }
 
   @Override
-  public AggOutcome doWork() {
+  public AggOutcome doWork() throws SchemaChangeException {
```
gparai: `doWork()` is used elsewhere as well. e.g. StreamingAggregator. Would it make sense to address other `doWork()` s in this PR. Better yet factor out another interface which the HashAggregator/StreamingAggregator implement - this would contain `public AggOutcome doWork() throws SchemaChangeException`

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -324,9 +328,9 @@ public AggOutcome doWork() {
                   logger.debug("Received new schema.  Batch has {} records.", incoming.getRecordCount());
                 }
 //                newSchema = true;
-                this.cleanup();
                 // TODO: new schema case needs to be handled appropriately
-                return AggOutcome.UPDATE_AGGREGATOR;
+                this.cleanup();
+                throw SchemaChangeException.schemChanged("Hash aggregate does not support schema changes", schema, incoming.getSchema());
```
gparai: Please correct to the renamed function here and elsewhere.

exec/java-exec/src/main/java/org/apache/drill/exec/exception/SchemaChangeException.java
jinfengni: Nice catch.  Modified it.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
jinfengni: You are right that Streaming aggregator can not handle schema change exception currently. 

In the current code since Drill's planner put sort operator before streaming aggregator, if the input has schema change, it would hit schema change exception in sort operator first, before come to streaming aggregator.

Anyway, I modified the code in StreamingAggregator, in case Drill planner generate a plan without sort for streaming aggregator.

I also simplified the change, since I realized we do not have to change in AggTemplate; the detailed error message could be produced in AggBatch.

jinfengni:
Addressed review comments. Pls let me know if you have any further comments. thx.

gparai:
Would it be a good idea to add tests for the same? Otherwise, LGTM.

jinfengni:
If we change the flow to handle schema change, it makes senses to add new tests. Here, we only changes the error message ; not how we handle scheme change. I'm not inclined to add new tests just to verify the error message, since error message might involve over time.

PR #802, opened 2017-03-29
DRILL-5394: Optimize query planning for MapR-DB tables by caching row…

ppadma:
… counts

During the planning phase, we fetch MapR DB table row count multiple times. This is an expensive operation and fetching it multiple times increases planning time considerably.  Implemented per query cache. Fetch table rowCount only once, cache it in hbaseScanSpec and reuse it. Copy the rowCount from old to new hbaseScanSpec when a new groupScan is created i.e. when the filter is pushed down into scan.

### DIFF
--- a/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushFilterIntoScan.java
+++ b/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushFilterIntoScan.java
@@ -184,8 +184,10 @@ public abstract class MapRDBPushFilterIntoScan extends StoragePluginOptimizerRul
       return; //no filter pushdown ==> No transformation.
     }
 
+    // Pass tableStats from old groupScan so we do not go and fetch stats (an expensive operation) again from MapR DB client.
     final BinaryTableGroupScan newGroupsScan = new BinaryTableGroupScan(groupScan.getUserName(), groupScan.getStoragePlugin(),
-                                                              groupScan.getFormatPlugin(), newScanSpec, groupScan.getColumns());
+                                                                        groupScan.getFormatPlugin(), newScanSpec, groupScan.getColumns(),
+                                                                        groupScan.getTableStats());
     newGroupsScan.setFilterPushedDown(true);
 
     final ScanPrel newScanPrel = ScanPrel.create(scan, filter.getTraitSet(), newGroupsScan, scan.getRowType());
--- a/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/binary/BinaryTableGroupScan.java
+++ b/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/binary/BinaryTableGroupScan.java
@@ -89,6 +89,15 @@ public class BinaryTableGroupScan extends MapRDBGroupScan implements DrillHBaseC
     init();
   }
 
+  public BinaryTableGroupScan(String userName, FileSystemPlugin storagePlugin,
+                              MapRDBFormatPlugin formatPlugin, HBaseScanSpec scanSpec,
+                              List<SchemaPath> columns, MapRDBTableStats tableStats) {
+    super(storagePlugin, formatPlugin, columns, userName);
+    this.hbaseScanSpec = scanSpec;
+    this.tableStats = tableStats;
+    init();
+  }
+
   /**
    * Private constructor, used for cloning.
    * @param that The HBaseGroupScan to clone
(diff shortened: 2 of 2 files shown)


Agirish:
Can confirm that with this fix, query times have improved (details will be updated in JIRA). Thanks for making this change.

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseScanSpec.java
```diff
@@ -76,6 +78,10 @@ public String getTableName() {
     return stopRow == null ? HConstants.EMPTY_START_ROW : stopRow;
   }
 
+  public long getRowCount() { return rowCount; }
+
+  public void setRowCount(long numRowCount) { rowCount = numRowCount; }
```
gparai: Should the ScanSpec be mutable since it serves a specification? I looked at other ScanSpecs but none seem to set ScanSpec members. Maybe all the members should be final and initialized via the constructor?

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/binary/BinaryTableGroupScan.java
```diff
@@ -115,8 +112,11 @@ private void init() {
     try (Admin admin = formatPlugin.getConnection().getAdmin();
          RegionLocator locator = formatPlugin.getConnection().getRegionLocator(tableName)) {
       hTableDesc = admin.getTableDescriptor(tableName);
-      tableStats = new MapRDBTableStats(getHBaseConf(), hbaseScanSpec.getTableName());
-
+      // Fetch rowCount only once and cache it in hbaseScanSpec.
+      if (hbaseScanSpec.getRowCount() == hbaseScanSpec.ROW_COUNT_UNKNOWN) {
+        MapRDBTableStats tableStats = new MapRDBTableStats(getHBaseConf(), hbaseScanSpec.getTableName());
+        hbaseScanSpec.setRowCount(tableStats.getNumRows());
```
gparai: This looks weird. We create the TableStats relying on some information from the ScanSpec and then proceed to modify the same ScanSpec with the information retrieved from TableStats. Please look at below comment as well regarding Spec mutability.

Should we instead just overload the MapRDBTableStats constructor to allow passing numRows - since that is what the existing constructor endup doing but makes a call to DB Client? So instead of populating the ScanSpec we populate the tableStats using this new constructor?

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/binary/BinaryTableGroupScan.java
```diff
@@ -174,7 +174,7 @@ public MapRDBSubScan getSpecificScan(int minorFragmentId) {
   @Override
   public ScanStats getScanStats() {
     //TODO: look at stats for this.
-    long rowCount = (long) ((hbaseScanSpec.getFilter() != null ? .5 : 1) * tableStats.getNumRows());
+    long rowCount = (long) ((hbaseScanSpec.getFilter() != null ? .5 : 1) * hbaseScanSpec.getRowCount());
```
gparai: This code change would not be required if we use the alternative approach?

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushFilterIntoScan.java
```diff
@@ -184,6 +184,8 @@ protected void doPushFilterIntoBinaryGroupScan(final RelOptRuleCall call,
       return; //no filter pushdown ==> No transformation.
     }
 
+    // Set rowCount in newScanSpec so we do not go and fetch rowCount (an expensive operation) again from MapR DB client.
+    newScanSpec.setRowCount(groupScan.getHBaseScanSpec().getRowCount());
```
gparai: Change the constructor for BinaryTableGroupScan to also pass in TableStats and add a getter for TableStats. This should be sufficient for passing around the rows.

ppadma:
Thanks for the review Gautam. Good suggestion. I changed the code. rowCount is cached in groupScan. Added a new constructor to pass the rowCount from old groupScan to new groupScan. Please review the new diffs.

gparai:
Further code refactoring may be worth the effort - We leave tableStats as-is and pass it around instead of using the new `rowCount`. This would cleanup the code further and would make it more robust to future changes in tableStats.

However, in the interest of time I would not hold up the pull request for the above code refactoring.
+1

ppadma:
ok, I changed it so we cache MapRDBTableStats, not just rowCount. Please review the new diffs.

gparai:
The new changes look good! Thanks for further refactoring the code. LGTM  +1

jinfengni:
+1

PR #803, opened 2017-03-29
DRILL-5395: Query on MapR-DB table fails with NPE due to an issue wit…

ppadma:
… counts
See [DRILL-5395](https://issues.apache.org/jira/browse/DRILL-5395) for details.


Agirish:
Can confirm that with this fix, the previously observed issue (NPE) is no longer reproducible. Thanks for making this change.

PR #214, opened 2015-10-21
DRILL-3963 Add Sequence file support.

amithadke:
Read key values  from sequence file as VarBinary.

PR #804, opened 2017-03-31
DRILL-5405: Add missing operator types

arina-ielchiieva:


arina-ielchiieva:
It's CoreOperatorType enum from UserBitShared.proto [1].

[1] https://github.com/apache/drill/blob/master/protocol/src/main/protobuf/UserBitShared.proto#L266

sudheeshkatkam:
From the [comments on 214](https://github.com/apache/drill/pull/214#discussion_r42687575), looks like this change in not required. I could not find the resolution/ discussion on the dev list, as mentioned in the comments. But since a new operator type was not added in the PR, this change is not necessary. -0 for this change.

arina-ielchiieva:
@sudheeshkatkam thanks for pointing out on the discussion. I will close this PR but I'll update Jira to reflect discussion points.

PR #805, opened 2017-04-03
Drill-4139: Exception while trying to prune partition. java.lang.UnsupportedOperationException: Unsupported type: BIT & Interval

vvysotskyi:
Add missing BIT partition pruning support.
Add missing Interval, VarBinary and Varchar with nulls partition pruning support.
Fix metadata serialization for fixed_len_byte_array and binary types.
Fix partition pruning for decimal type.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Iterator;
 
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
@@ -86,6 +87,7 @@ import static org.apache.drill.exec.store.parquet.MetadataVersion.Constants.V2;
 import static org.apache.drill.exec.store.parquet.MetadataVersion.Constants.V3;
 import static org.apache.drill.exec.store.parquet.MetadataVersion.Constants.V3_1;
 import static org.apache.drill.exec.store.parquet.MetadataVersion.Constants.V3_2;
+import static org.apache.drill.exec.store.parquet.MetadataVersion.Constants.V3_3;
 
 public class Metadata {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Metadata.class);
@@ -700,7 +702,8 @@ public class Metadata {
       @JsonSubTypes.Type(value = ParquetTableMetadata_v2.class, name = V2),
       @JsonSubTypes.Type(value = ParquetTableMetadata_v3.class, name = V3),
       @JsonSubTypes.Type(value = ParquetTableMetadata_v3.class, name = V3_1),
-      @JsonSubTypes.Type(value = ParquetTableMetadata_v3.class, name = V3_2)
+      @JsonSubTypes.Type(value = ParquetTableMetadata_v3.class, name = V3_2),
+      @JsonSubTypes.Type(value = ParquetTableMetadata_v3.class, name = V3_3)
       })
   public static abstract class ParquetTableMetadataBase {
 
@@ -756,7 +759,7 @@ public class Metadata {
 
     public abstract Long getNulls();
 
-    public abstract boolean hasSingleValue();
+    public abstract boolean hasSingleValue(long rowCount);
 
     public abstract Object getMinValue();
 
@@ -1054,8 +1057,36 @@ public class Metadata {
       return nulls;
     }
 
-    @Override public boolean hasSingleValue() {
-      return (max != null && min != null && max.equals(min));
+    /**
+     * Checks that the column chunk has a single value.
+     * Returns {@code true} if {@code min} and {@code max} are the same but not null
+     * and nulls count is 0 or equal to the rows count.
+     * <p>
+     * Returns {@code true} if {@code min} and {@code max} are null and the number of null values
+     * in the column chunk is equal to the rows count.
+     * <p>
+     * Comparison of nulls and rows count is needed for the cases:
+     * <ul>
+     * <li>column with primitive type has single value and null values</li>
+     *
+     * <li>column <b>with primitive type</b> has only null values, min/max couldn't be null,
+     * but column has single value</li>
+     * </ul>
+     *
+     * @param rowCount rows count in column chunk
+     * @return true if column has single value
+     */
+    @Override
+    public boolean hasSingleValue(long rowCount) {
+      if (nulls != null) {
+        if (min != null) {
+          // Objects.deepEquals() is used here, since min and max may be byte arrays
+          return Objects.deepEquals(min, max) && (nulls == 0 || nulls == rowCount);
+        } else {
+          return nulls == rowCount && max == null;
+        }
+      }
+      return false;
     }
 
     @Override public Object getMinValue() {
@@ -1363,9 +1394,24 @@ public class Metadata {
       return nulls;
     }
 
+    /**
+     * Checks that the column chunk has a single value.
+     * Returns {@code true} if {@code mxValue} is not null
+     * and nulls count is 0 or if nulls count is equal to the rows count.
+     * <p>
+     * Comparison of nulls and rows count is needed for the cases:
+     * <ul>
+     * <li>column with primitive type has single value and null values</li>
+     *
+     * <li>column <b>with binary type</b> has only null values, so column has single value</li>
+     * </ul>
+     *
+     * @param rowCount rows count in column chunk
+     * @return true if column has single value
+     */
     @Override
-    public boolean hasSingleValue() {
-      return (mxValue != null);
+    public boolean hasSingleValue(long rowCount) {
+      return (mxValue != null && nulls == 0) || nulls == rowCount;
     }
 
     @Override public Object getMinValue() {
@@ -1426,7 +1472,7 @@ public class Metadata {
 
   }
 
-  @JsonTypeName(V3_2)
+  @JsonTypeName(V3_3)
   public static class ParquetTableMetadata_v3 extends ParquetTableMetadataBase {
     @JsonProperty(value = "metadata_version", access = JsonProperty.Access.WRITE_ONLY) private String metadataVersion;
     /*
@@ -1753,9 +1799,36 @@ public class Metadata {
       return nulls;
     }
 
+    /**
+     * Checks that the column chunk has a single value.
+     * Returns {@code true} if {@code minValue} and {@code maxValue} are the same but not null
+     * and nulls count is 0 or equal to the rows count.
+     * <p>
+     * Returns {@code true} if {@code minValue} and {@code maxValue} are null and the number of null values
+     * in the column chunk is equal to the rows count.
+     * <p>
+     * Comparison of nulls and rows count is needed for the cases:
+     * <ul>
+     * <li>column with primitive type has single value and null values</li>
+     *
+     * <li>column <b>with primitive type</b> has only null values, min/max couldn't be null,
+     * but column has single value</li>
+     * </ul>
+     *
+     * @param rowCount rows count in column chunk
+     * @return true if column has single value
+     */
     @Override
-    public boolean hasSingleValue() {
-      return (minValue !=null && maxValue != null && minValue.equals(maxValue));
+    public boolean hasSingleValue(long rowCount) {
+      if (nulls != null) {
+        if (minValue != null) {
+          // Objects.deepEquals() is used here, since min and max may be byte arrays
+          return Objects.deepEquals(minValue, maxValue) && (nulls == 0 || nulls == rowCount);
+        } else {
+          return nulls == rowCount && maxValue == null;
+        }
+      }
+      return false;
     }
 
     @Override public Object getMinValue() {
@@ -1795,8 +1868,9 @@ public class Metadata {
         jgen.writeEndArray();
         if (value.minValue != null) {
           Object val;
-          if (value.primitiveType == PrimitiveTypeName.BINARY && value.minValue != null) {
-            val = new String(((Binary) value.minValue).getBytes());
+          if (value.primitiveType == PrimitiveTypeName.BINARY
+              || value.primitiveType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
+            val = ((Binary) value.minValue).getBytes();
           } else {
             val = value.minValue;
           }
@@ -1804,8 +1878,9 @@ public class Metadata {
         }
         if (value.maxValue != null) {
           Object val;
-          if (value.primitiveType == PrimitiveTypeName.BINARY && value.maxValue != null) {
-            val = new String(((Binary) value.maxValue).getBytes());
+          if (value.primitiveType == PrimitiveTypeName.BINARY
+              || value.primitiveType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
+            val = ((Binary) value.maxValue).getBytes();
           } else {
             val = value.maxValue;
           }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
@@ -76,6 +76,7 @@ import org.apache.drill.exec.store.schedule.CompleteWork;
 import org.apache.drill.exec.store.schedule.EndpointByteMap;
 import org.apache.drill.exec.store.schedule.EndpointByteMapImpl;
 import org.apache.drill.exec.util.ImpersonationUtil;
+import org.apache.drill.exec.vector.NullableBitVector;
 import org.apache.drill.exec.vector.NullableBigIntVector;
 import org.apache.drill.exec.vector.NullableDateVector;
 import org.apache.drill.exec.vector.NullableDecimal18Vector;
@@ -477,6 +478,12 @@ public class ParquetGroupScan extends AbstractFileGroupScan {
     String f = Path.getPathWithoutSchemeAndAuthority(new Path(file)).toString();
     MinorType type = getTypeForColumn(column).getMinorType();
     switch (type) {
+      case BIT: {
+        NullableBitVector bitVector = (NullableBitVector) v;
+        Boolean value = (Boolean) partitionValueMap.get(f).get(column);
+        bitVector.getMutator().setSafe(index, value ? 1 : 0);
+        return;
+      }
       case INT: {
         NullableIntVector intVector = (NullableIntVector) v;
         Integer value = (Integer) partitionValueMap.get(f).get(column);
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
@@ -36,6 +36,7 @@ import org.apache.drill.common.expression.SchemaPath;
 import org.apache.drill.common.expression.ValueExpressions;
 import org.apache.drill.common.logical.FormatPluginConfig;
 import org.apache.drill.common.logical.StoragePluginConfig;
+import org.apache.drill.common.types.TypeProtos;
 import org.apache.drill.common.types.TypeProtos.MajorType;
 import org.apache.drill.common.types.TypeProtos.MinorType;
 import org.apache.drill.common.types.Types;
@@ -75,6 +76,7 @@ import org.apache.drill.exec.store.schedule.AssignmentCreator;
 import org.apache.drill.exec.store.schedule.CompleteWork;
 import org.apache.drill.exec.store.schedule.EndpointByteMap;
 import org.apache.drill.exec.store.schedule.EndpointByteMapImpl;
+import org.apache.drill.exec.util.DecimalUtility;
 import org.apache.drill.exec.util.ImpersonationUtil;
 import org.apache.drill.exec.vector.NullableBitVector;
 import org.apache.drill.exec.vector.NullableBigIntVector;
@@ -83,6 +85,7 @@ import org.apache.drill.exec.vector.NullableDecimal18Vector;
 import org.apache.drill.exec.vector.NullableFloat4Vector;
 import org.apache.drill.exec.vector.NullableFloat8Vector;
 import org.apache.drill.exec.vector.NullableIntVector;
+import org.apache.drill.exec.vector.NullableIntervalVector;
 import org.apache.drill.exec.vector.NullableSmallIntVector;
 import org.apache.drill.exec.vector.NullableTimeStampVector;
 import org.apache.drill.exec.vector.NullableTimeVector;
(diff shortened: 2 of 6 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1668,9 +1666,16 @@ public ColumnMetadata_v3(String[] name, PrimitiveTypeName primitiveType, Object
...
+     * Checks that the column chunk has single value.
+     * Returns true if minValue and maxValue are null and
+     * column statistics was available.
+     *
+     * @return true if column has single value
+     */
     @Override
     public boolean hasSingleValue() {
-      return (minValue !=null && maxValue != null && minValue.equals(maxValue));
+      return minValue != null ? minValue.equals(maxValue) : (maxValue == null && isMetadataSet());
```
jinfengni: We probably need better definition of "column has single value", before go ahead and modify this function.  Cases we need consider:

1) minValue / maxValue != null and there is no "null" in the column.
2) minValue / maxValue != null and there is "null" in the column. 
3) minValue/maxValue == null, and there is "null" in the column

vvysotskyi: 1. It is clear that we should just compare minValue and maxValue.
2. In this case we should check that nulls == 0. Fixed that. Now for this case function returns false.
3. We should additionally check that nulls > 0. Fixed that.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1668,9 +1666,25 @@ public ColumnMetadata_v3(String[] name, PrimitiveTypeName primitiveType, Object
...
+     * and all column chunk values are not null.
+     * Returns true if minValue and maxValue are null and null values count in
+     * the column chunk is greater than 0.
+     *
+     * @return true if column has single value
+     */
     @Override
     public boolean hasSingleValue() {
-      return (minValue !=null && maxValue != null && minValue.equals(maxValue));
+      if (nulls != null) {
```
jinfengni: Should this applies to v2 and v1 of ColumnMetadata?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -508,7 +511,10 @@ public void populatePruningVector(ValueVector v, int index, SchemaPath column, S
         NullableVarBinaryVector varBinaryVector = (NullableVarBinaryVector) v;
         Object s = partitionValueMap.get(f).get(column);
         byte[] bytes;
-        if (s instanceof Binary) {
+        if (s == null) {
+          varBinaryVector.getMutator().setNull(index);
```
jinfengni: Should we apply this setNull() to all other types?

exec/java-exec/src/main/codegen/templates/CastVarBinaryInterval.java
```diff
@@ -0,0 +1,72 @@
...
+  @Output ${type.to}Holder out;
+
+  public void setup() {
+  }
+
+  public void eval() {
+    final byte[] buf = new byte[in.end - in.start];
+    in.buffer.getBytes(in.start, buf, 0, in.end - in.start);
+  <#if type.to == "Interval">
+    out.months = org.apache.drill.exec.store.parquet.ParquetReaderUtility.getIntFromLEBytes(buf, 0);
```
vvysotskyi: I made conclusion that Interval type encoded in this way when found that IntervalVector stores it's value in DrillBuf data as 3 int values.

But I had seen that there are a lot of other types encoded as you described, so you are right, thanks for the clarification.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
vvysotskyi: Yes, we should apply it to ColumnMetadata_v1, thanks.
In ColumnMetadata_v2 mxValue is set only in the case when min value is the same as the max value. So we could not be sure that if mxValue == null and nulls count is greater than zero, column has single value.

Also I have changed the result for the second case described in your previous comment. Statistics [1] for most parquet types use java primitive types to store min and max values, so min/max can not be null even if the table has null values. So I removed check nulls == 0.

[1] https://github.com/apache/parquet-mr/tree/e54ca615f213f5db6d34d9163c97eec98920d7a7/parquet-column/src/main/java/org/apache/parquet/column/statistics

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
vvysotskyi: We can apply this setNull() only for types that use BinaryStatistics [1] for the reason described in my previous comment. So we apply setNull() to VARBINARY, DECIMAL18, VARCHAR and INTERVAL types.

[1] https://github.com/apache/parquet-mr/blob/e54ca615f213f5db6d34d9163c97eec98920d7a7/parquet-column/src/main/java/org/apache/parquet/column/statistics/BinaryStatistics.java#L25

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -181,6 +185,71 @@ else if (parquetTableMetadata instanceof Metadata.ParquetTableMetadata_v2 &&
...
+   * the version where was changed the serialization of BINARY values.
+   *
+   * @param parquetTableMetadata the source of drill version
+   * @return true if metadata file was created by drill with version less that
+   * the version where was changed the serialization of BINARY values
+   */
+  public static boolean hasOldBinarySerialization(Metadata.ParquetTableMetadataBase parquetTableMetadata) {
+    String drillVersion = parquetTableMetadata.getDrillVersion();
+    return drillVersion == null
+            || new ComparableVersion(drillVersion).compareTo(new ComparableVersion("1.11.0-SNAPSHOT")) < 0;
```
arina-ielchiieva: I am not sure that using version with SNAPSHOT is correct.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -181,6 +185,71 @@ else if (parquetTableMetadata instanceof Metadata.ParquetTableMetadata_v2 &&
...
+            ((Metadata.ParquetTableMetadata_v2) parquetTableMetadata).columnTypeInfo.values()) {
+          if (columnTypeMetadata.primitiveType == PrimitiveTypeName.BINARY) {
+            names.add(Arrays.asList(columnTypeMetadata.name));
+          }
+        }
+      }
+      for (Metadata.ParquetFileMetadata file : parquetTableMetadata.getFiles()) {
+        // Drill has only ever written a single row group per file, only need to correct the statistics
+        // on the first row group
+        Metadata.RowGroupMetadata rowGroupMetadata = file.getRowGroups().get(0);
```
jinfengni: It's true that parquet files created by Drill have single RG per file. But the metadata file could be created from parquet files from other source. Such assumption may not be true.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -181,6 +185,71 @@ else if (parquetTableMetadata instanceof Metadata.ParquetTableMetadata_v2 &&
   }
 
   /**
+   * Checks that the metadata file was created by drill with version less than
+   * the version where was changed the serialization of BINARY values
```
jinfengni: Can you explain a bit more about why the serialization of BINARY values are wrong in prior version? Is it only happening in the metadata cache file?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -181,6 +185,71 @@ else if (parquetTableMetadata instanceof Metadata.ParquetTableMetadata_v2 &&
...
+   * and assigns byte arrays to min/max values obtained from the deserialized string.
+   *
+   * @param parquetTableMetadata table metadata that should be corrected
+   */
+  public static void correctBinaryInMetadataCache(Metadata.ParquetTableMetadataBase parquetTableMetadata) {
+    if (hasOldBinarySerialization(parquetTableMetadata)) {
+      Set<List<String>> names = Sets.newHashSet();
+      if (parquetTableMetadata instanceof Metadata.ParquetTableMetadata_v2) {
```
jinfengni: Any reason you only check v2 here?

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -398,10 +399,115 @@ public void testDrill4877() throws Exception {
 
   }
 
+  @Test // DRILL-4139
+  public void testBooleanPartitionPruning() throws Exception {
+    final String boolPartitionTable = "dfs_test.tmp.`interval_bool_partition`";
+    try {
+      test("create table %s partition by (col_bln) as " +
+            "select * from cp.`parquet/alltypes_required.parquet`", boolPartitionTable);
+      test("refresh table metadata %s", boolPartitionTable);
```
jinfengni: We probably want to cover the cases where metadata cache file is not created (run the query before call "refresh table metadata").

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
vvysotskyi: Added check for ParquetTableMetadata_v3, thanks. In class ColumnMetadata_v1 field `primitiveType` is serialized to the metadata cache file, so it is easier to check column type for this metadata version.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
vvysotskyi: Thanks, fixed it.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vvysotskyi: Thanks, modified tests to cover those cases.

paul-rogers:
@jinfengni, this PR has been sitting idle for a while. Anything else that needs to be done on this one?

jinfengni:
Thanks for the reminder, @paul-rogers . Somehow, I overlooked this PR. I thought I have reviewed and merged a PR from @vvysotskyi . But looks like I confused this one with another one I merged.

Anyway, I'll take another look at this PR.

vvysotskyi:
Resolved merge conflicts and rebased on latest master.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1008,8 +1008,24 @@ public void setMax(Object max) {
...
 
-    @Override public boolean hasSingleValue() {
-      return (max != null && min != null && max.equals(min));
+    /**
+     * Checks that the column chunk has single value.
+     * Returns true if min and max are the same, but not null.
+     * Returns true if min and max are null and the number of null values
+     * in the column chunk is greater than 0.
+     *
+     * @return true if column has single value
```
jinfengni: My understanding is hasSingleValue() return true if the column meta data shows only one single value.  A null value is also counted as a different value from other non-null value.

Therefore, for the case of  column has min != null && max !=null && min.equals(max) && nulls!=null && nulls > 0, it should return false. However, in both the implementation of v1 and v3, it would return true. 

That would actually lead to wrong query result.  A simple reproduce:

```
create table dfs.tmp.`t5/a` as select 100 as mykey from cp.`tpch/nation.parquet` union all select col_notexist from cp.`tpch/region.parquet`;

create table dfs.tmp.`t5/b` as select 200 as mykey from cp.`tpch/nation.parquet` union all select col_notexist from cp.`tpch/region.parquet`;
```

We got two files, each having one single unique non-null value, plus null values. Now query the two files:

```
select mykey from dfs.tmp.`t5` where mykey = 100;
+--------+
| mykey  |
+--------+
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
| null   |
| null   |
| null   |
| null   |
| null   |
+--------+
30 rows selected (0.246 seconds)

```
Apparently, those 5 nulls should not be returned. 

I applied the 3 commits in this PR on top of today's master branch.

```
select * from sys.version;
+------------------+-------------------------------------------+-------------------------------------------------------------------------------+----------------------------+-----------------+----------------------------+
|     version      |                 commit_id                 |                                commit_message                                 |        commit_time         |   build_email   |         build_time         |
+------------------+-------------------------------------------+-------------------------------------------------------------------------------+----------------------------+-----------------+----------------------------+
| 1.11.0-SNAPSHOT  | cad6e4dc950aa4a95ad20515ce5abd9c546d3e5d  | DRILL-4139: Fix loss of scale value for DECIMAL in parquet partition pruning  | 05.07.2017 @ 12:05:25 PDT  | jni@apache.org  | 05.07.2017 @ 12:06:07 PDT  |
+------------------+-------------------------------------------+-----
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1696,9 +1712,24 @@ public ColumnMetadata_v3(String[] name, PrimitiveTypeName primitiveType, Object
...
+     */
     @Override
     public boolean hasSingleValue() {
-      return (minValue !=null && maxValue != null && minValue.equals(maxValue));
+      if (nulls != null) {
+        if (minValue != null) {
+          return minValue.equals(maxValue);
```
jinfengni: Same comment as above. I think in both v1 & v3, we should check nulls = 0 ( there is no nulls in the column).

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -445,12 +448,148 @@ public void testMetadataCacheAbsolutePaths() throws Exception {
       test("drop table if exists %s", RELATIVE_PATHS_METADATA);
     }
   }
+  @Test // DRILL-4139
+  public void testBooleanPartitionPruning() throws Exception {
```
jinfengni: In addition to "=" comparison, can you please try with "is null" / "is not null" predicate, since one of this PR's objectives is to fix partition pruning dealing null values.  I saw another incorrect query result, although I'm not sure if it's a bug in the base code, or a bug in this PR.

```
create table dfs.tmp.`t6/a` as select col_notexist as mykey from cp.`tpch/region.parquet`;
create table dfs.tmp.`t6/b` as select 100 as mykey from cp.`tpch/region.parquet`;
select mykey from dfs.tmp.t6;
+--------+
| mykey  |
+--------+
| null   |
| null   |
| null   |
| null   |
| null   |
| 100    |
| 100    |
| 100    |
| 100    |
| 100    |
+--------+

select mykey from dfs.tmp.t6 where mykey is null;
+--------+
| mykey  |
+--------+
+--------+

``` 

Notice that the last query returns 0 row, where it should return 5 rows.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
vvysotskyi: Thanks for such detailed analysis. This bug is also reproduced on master version, but it appears due to the wrong result of this method. Added a comparison of `nulls` with rows count to fix this bug and allow to return true for the case when field has only nulls.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vvysotskyi: Thanks, fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -548,23 +567,57 @@ public void populatePruningVector(ValueVector v, int index, SchemaPath column, S
...
+   * Returns the sequence of bytes received from {@code Object source}.
+   *
+   * @param type   the column type
+   * @param source the source of the bytes sequence
+   * @return bytes sequence obtained from {@code Object source}
+   */
+  private byte[] getBytes(MinorType type, Object source) {
+    byte[] bytes;
+    if (source instanceof String) { // if the metadata was read from a JSON cache file it maybe a string type
+      bytes = Base64.decodeBase64(((String) source).getBytes());
```
paul-rogers: The data here should be assumed to be binary. Binary can have any characters, even JSON symbols. For example:

"value:" : "look a quote",\n"bogus" : { 10, "foo" }"

The above will make a mash of any JSON structure. Better to encode binary data as Base 64, which is designed to fit within a text format.

Note, however, that the casting is unfortunate. The caller knows the type. Better to have:

byte[] getBytes(String value) ...
byte[] getBytes(byte[] value) ...
byte[] getBytes(Binary value) ...
{code}

Seems we'd also need a to/from Base64 method so that, on read, we do:

* Get string value.
* Pass string value to Base64 decoder, get bytes
* Pass bytes to vector, using vector's code to understand byte meaning

On encoding:

* Get the vector value as a byte array
* Base64 encode the array
* Write it to JSON as a string

All that said, Strings need not be Base64 encoded if we have a JSON encoder that will escape JSON special characters. This works as long as the JSON file itself is UTF-8 encoded because then each Unicode string has a UTF-8 equivalent.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -181,6 +186,101 @@ else if (parquetTableMetadata instanceof Metadata.ParquetTableMetadata_v2 &&
   }
 
   /**
+   * Checks that the metadata file has version less than
+   * the version where was changed the serialization of BINARY values
+   * and assigns byte arrays to min/max values obtained from the deserialized string.
+   *
+   * @param parquetTableMetadata table metadata that should be corrected
+   */
```
paul-rogers: This is very hard to follow for those of us not familiar with the details of the metadata file format. Please include a comment in the PR conversation that shows an example of the old and new formats. That will make it easy to double-check that the logic makes sense.

paul-rogers:
What is the status of this PR? More work to do? Ready for another review?

vvysotskyi:
Currently, I am waiting when DRILL-5660 is merged. I will add a comment when it is ready for another review.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
vvysotskyi: `ColumnMetadata_v1`  class serializes / deserializes `primitiveType` field so it does not hard to check that field has `BINARY` type.
But `ColumnMetadata_v2` and `ColumnMetadata_v3` does not serialize this field. They have classes `ColumnTypeMetadata_v2` and `ColumnTypeMetadata_v3` respectively which contain information about the field name, primitive and original types. In this method, we are receiving fields with `BINARY` type taking information about the field type from `ColumnTypeMetadata_v2` or `ColumnTypeMetadata_v3`.

paul-rogers:
Thanks for addressing my questions. While waiting for DRILL-5660, I wonder if we can ask Salim to take a look?

arina-ielchiieva:
@paul-rogers, PR is already merged :)

paul-rogers:
OK. Since I'm not a Parquet expert, have asked Salim to take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1054,8 +1057,36 @@ public void setMax(Object max) {
...
+     *
+     * @param rowCount rows count in column chunk
+     * @return true if column has single value
+     */
+    @Override
+    public boolean hasSingleValue(long rowCount) {
+      if (nulls != null) {
+        if (min != null) {
+          // Objects.deepEquals() is used here, since min and max may be byte arrays
+          return Objects.deepEquals(min, max) && (nulls == 0 || nulls == rowCount);
```
sachouche: - if (min != null), then nulls cannot be equal to rowCount
- In this case, only nulls == 0 should be checked

vvysotskyi: Statistics [1] for most parquet types use java primitive types to store min and max values, so min/max can not be null even if the table has null values.

[1] https://github.com/apache/parquet-mr/tree/e54ca615f213f5db6d34d9163c97eec98920d7a7/parquet-column/src/main/java/org/apache/parquet/column/statistics

sachouche:
+1

jinfengni:
The revised patch addressed my earlier comments. LGTM. 

+1

ps: I did not look into the change of version number. @paul-rogers would have better idea.

paul-rogers:
Tried to commit this one. Failed in functional tests:

{noformat}
Data Verification Failures:
/root/drillAutomation/mapr/framework/resources/Functional/metadata_caching/data/varchar_partition6.q
/root/drillAutomation/mapr/framework/resources/Functional/parquet_storage/parquet_date/auto_partition/data/q2.q
/root/drillAutomation/mapr/framework/resources/Functional/metadata_caching/data/varchar_partition3.q

Plan Verification Failures:
/root/drillAutomation/mapr/framework/resources/Functional/metadata_caching/plan/varchar_partition8.q
/root/drillAutomation/mapr/framework/resources/Functional/metadata_caching/plan/varchar_partition2.q
/root/drillAutomation/mapr/framework/resources/Functional/parquet_storage/parquet_date/auto_partition/plan/q2.q
/root/drillAutomation/mapr/framework/resources/Functional/metadata_caching/plan/varchar_partition1.q
{noformat}

And in unit tests:

{noformat}
Failed tests: 
  TestHashAggrSpill.testHashAggrFailWithFallbackDisabed:165 null

Tests in error: 
  TestHashAggrSpill.testNoPredictHashAggrSpill:135->testSpill:110 » IllegalState
  TestHashAggrSpill.testHashAggrSecondaryTertiarySpill:147->testSpill:110 » IllegalState
  TestHashAggrSpill.testSimpleHashAggrSpill:123->testSpill:110 » IllegalState Cl...
  TestHashAggrSpill.testHashAggrSuccessWithFallbackEnabled:179->testSpill:110 » IllegalState
{noformat}

See build MERGE-170924-1 for details.

vvysotskyi:
@paul-rogers, test data should be regenerated (using the Drill version with this fix) before running functional tests since these tests use metadata cache files. 
Regarding unit test failures: I rebased onto the master and run tests on my machine and on the test cluster and for both cases, tests finished without errors. So, perhaps it was a random test failure.

paul-rogers:
I'm probably confused. Our goal with the Parquet metadata versioning was that new Drill versions can read old files. Here we have a case in which I was running with old versions of the file, using your new code. Why do I need to rebuild files? Is version backward compatibility already broken?

vvysotskyi:
This my bad habit to regenerate metadata left from the times before fix DRILL-5660. These functional test failures are the result of the mistake that I have made when resolving merge conflicts with DRILL-4264. After I made rebase, older 3.2 version still used in `ParquetReaderUtility` instead of 3.3. 
I fixed this bug and functional tests are pass without regenerating parquet metadata cache. Could you please take a look at this change in the last commit?

PR #807, opened 2017-04-05
DRILL-5415: Improve Fixture Builder to configure client properties an…

sohami:
…d keep collection type properties for server

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
@@ -28,6 +28,8 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
 
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigValueFactory;
 import org.apache.commons.io.FileUtils;
 import org.apache.drill.BaseTestQuery;
 import org.apache.drill.DrillTestWrapper.TestServices;
@@ -133,6 +135,7 @@ public class ClusterFixture implements AutoCloseable {
   private boolean usesZk;
   private boolean preserveLocalFiles;
   private boolean isLocal;
+  private Properties clientProps;
 
   /**
    * Temporary directories created for this test cluster.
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -203,7 +219,15 @@ private void createConfig(FixtureBuilder builder) throws Exception {
     if (builder.configResource != null) {
       config = DrillConfig.create(builder.configResource);
     } else if (builder.configProps != null) {
-      config = DrillConfig.create(configProperties(builder.configProps));
+      // Create Config object out of all the String type configuration.
```
paul-rogers: Suggestion to keep things tidy. Keep all properties (simple and complex) in the builder's `configProps` object. Then, in `configProperties`, do the mapping as required to create the `DrillConfig`. If a value is a collection, use the new code to set the value. Else, convert the value using `toString()`.

exec/java-exec/src/test/java/org/apache/drill/test/FixtureBuilder.java
```diff
@@ -117,7 +122,27 @@ public FixtureBuilder configProperty(String key, Object value) {
     if (configProps == null) {
       configProps = defaultProps();
     }
-    configProps.put(key, value.toString());
+
+    if(value instanceof Collection<?>) {
```
paul-rogers: Can we just leave the collection in the single `configProps` object, and do the mapping when creating the final config? That is, for the DrillConfig, make the collection test when building DrillConfig. See comment above.

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
sohami: Fixed

exec/java-exec/src/test/java/org/apache/drill/test/FixtureBuilder.java
sohami: Fixed

amansinha100:
+1

PR #808, opened 2017-04-06
DRILL-5325: Unit tests for the managed sort

paul-rogers:
Uses the sub-operator test framework (DRILL-5318), including the test
row set abstraction (DRILL-5323) to enable unit testing of the
“managed” external sort. This PR allows early review of the code, but
cannot be pulled until the dependencies (mentioned above) are pulled.

Refactors the external sort code into small chunks that can be unit
tested, then “wraps” that code in tests for all interesting data types,
record batch sizes, and so on.

Refactors some of the operator definitions to more easily allow
programmatic setup in the unit tests.

Fixes a number of bugs discovered by the unit tests. The biggest
changes were in the new code: the code that computes spilling and
merging based on memory levels.

Otherwise, although GitHub will show many files change, most of the
changes are simply moving blocks of code around to create smaller units
that can be tested independently.

Includes a refactoring of the code that does spilling, along with a
complete set of low-level unit tests.

Since this PR is dependent on others, it cannot (yet) be built on its
own. The PR will be rebased once the dependencies are in master so that
it will build prior to commit to master.


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
```diff
@@ -0,0 +1,506 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.xsort.managed;
+
+public class SortMemoryManager {
```
Ben-Zvi: Could this class be generalized for all spilling operators (and then customized for Sort, e.g., merge related parts.) ?  There are some parts with "fuzzy math" (e.g. batch size estimates; memory needed) ; Hash Agg is also doing its own fuzzy math - should all these be done the same way ? 
(The Hash Agg tried to be more conservative - like assuming 3 batches per a partition, though in theory 2 would do; currently this is controlled with an option)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortImpl.java
```diff
@@ -0,0 +1,495 @@
...
+  private SortResults mergeSpilledRuns() {
+    logger.debug("Starting consolidate phase. Batches = {}, Records = {}, Memory = {}, In-memory batches {}, spilled runs {}",
+                 metrics.getInputBatchCount(), metrics.getInputRowCount(),
+                 allocator.getAllocatedMemory(),
+                 bufferedBatches.size(), spilledRuns.size());
+
+    // Consolidate batches to a number that can be merged in
+    // a single last pass.
+
+    while (consolidateBatches()) {
```
paul-rogers: Done. Ends up needing a label, which some folks dislike...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortImpl.java
```diff
@@ -0,0 +1,495 @@
...
+    RuntimeException ex = null;
+    try {
+      spilledRuns.close();
+    } catch (RuntimeException e) {
+      ex = e;
+    }
+    try {
+      bufferedBatches.close();
+    } catch (RuntimeException e) {
+      ex = e;
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
paul-rogers: Not sure how transferable the logic here is. This is, essentially, a mathematical model of how memory is used by the sort/merge algorithm. Each operator needs its own model, else we're back to using made up logic that does not map to the actual code behavior.

For the most part, the numbers and algorithms are no longer "fuzzy." Instead, they reflect actual code behavior as determined first by inspection, then relentless QA testing. Any places where the estimates don't match reality have resulted in bugs filed against the sort.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
```diff
@@ -0,0 +1,506 @@
...
+   * <p>
+   * Under normal circumstances, the amount of memory available is much
+   * larger than the input, spill or merge batch sizes. The primary question
+   * is to determine how many input batches we can buffer during the load
+   * phase, and how many spill batches we can merge during the merge
+   * phase.
+   *
+   * @param batchSize the overall size of the current batch received from
+   * upstream
+   * @param batchRowWidth the width in bytes (including overhead) of each
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
```diff
@@ -0,0 +1,506 @@
...
+      bufferMemoryLimit = 0;
+      potentialOverflow = true;
+    }
+
+    // Sanity check
+
+    minNeeds = 2 * expectedSpillBatchSize + expectedMergeBatchSize;
+    if (minNeeds > memoryLimit) {
+      ExternalSortBatch.logger.warn("Potential memory overflow during merge phase! " +
+          "Minumum needed = {} bytes, actual available = {} bytes",
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
```diff
@@ -0,0 +1,506 @@
...
+  public long freeMemory(long allocatedBytes) {
+    return memoryLimit - allocatedBytes;
+  }
+
+  public long getMergeMemoryLimit() { return mergeMemoryLimit; }
+  public int getSpillBatchRowCount() { return spillBatchRowCount; }
+  public int getMergeBatchRowCount() { return mergeBatchRowCount; }
+
+  // Primarily for testing
+
```
paul-rogers: Fixed.

exec/memory/base/src/main/java/org/apache/drill/exec/memory/BaseAllocator.java
```diff
@@ -793,20 +797,20 @@ public static boolean isDebug() {
...
-   */
-
-  private byte ioBuffer[];
-
   public byte[] getIOBuffer() {
     if (ioBuffer == null) {
+      // Length chosen to the smallest size that maximizes
+      // disk I/O performance. Smaller sizes slow I/O. Larger
+      // sizes provide no increase in performance.
+      // Revisit from time to time.
```
paul-rogers: Fixed. It is a constant because no one can detect changes to performance of the number from outside Drill. To come up with the number required intense, focused tests.

paul-rogers:
Rebased onto latest master and resolved merge conflicts.

arina-ielchiieva:
+1

arina-ielchiieva:
Revoke +1 since there are failures in unit tests:

> Failed tests: 
  TestExternalSortExec.testSortSpec:175 expected:<1000000> but was:<20000000>

PR #811, opened 2017-04-09
DRILL-5423: Refactor ScanBatch to allow unit testing record readers

paul-rogers:
Refactors ScanBatch to allow unit testing of record reader
implementations, especially the “writer” classes.

See JIRA for details.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/OperatorExecContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/OperatorExecContext.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.ops;
+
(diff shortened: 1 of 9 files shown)


paul-rogers:
Rebased on latest master.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -323,11 +318,31 @@ public TypedFieldId getValueVectorId(SchemaPath path) {
     return container.getValueAccessorById(clazz, ids);
   }
 
-  private class Mutator implements OutputMutator {
+  public static class Mutator implements OutputMutator {
```
gparai: Maybe add comments about `Mutator` since the class is `public static`?

exec/java-exec/src/main/java/org/apache/drill/exec/ops/OperatorExecContext.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package org.apache.drill.exec.ops;
+
+import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.testing.ExecutionControls;
+
+import io.netty.buffer.DrillBuf;
+
+/**
+ * Narrowed version of the {@link OpeartorContext} used to create an
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/OperatorUtilities.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.drill.exec.physical.base.PhysicalOperator;
+
+public class OperatorUtilities {
+
+  private OperatorUtilities() { }
+
+  public static int getChildCount(PhysicalOperator popConfig) {
+    Iterator<PhysicalOperator> iter = popConfig.iterator();
+    int i = 0;
```
paul-rogers: Fixed. This code was actually just moved from its old home in OperatorContext, but went ahead and did the rename as it seems like a good idea.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
paul-rogers: Done. Also added a `@VisibleForTesting` annotation as well.

gparai:
Thanks for making the changes Paul. 

LGTM +1

PR #812, opened 2017-04-10
DRILL-5213: Prepared statement for actual query is missing the query text

vdiravka:


exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
arina-ielchiieva: As far as I  understood, `final RunQuery runQuery` now is not used in `QueryManager`, it can be removed from class.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -149,7 +148,7 @@
   private final boolean queuingEnabled;
 
 
-  private String queryText;
+  private String queryText = "";
```
arina-ielchiieva: May be it's better to check queryText nullability in QueryManager.java then setting empty string as default?

vdiravka: Done. Even we can not set the filed if it is null.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -540,6 +539,9 @@ private void runPreparedStatement(final PreparedStatementHandle preparedStatemen
     final String sql = serverState.getSqlQuery();
     logger.info("Prepared statement query for QueryId {} : {}", queryId, sql);
     runSQL(sql);
+
+    // keep a query text of the preparedStatement for the queryManager
+    queryText = sql;
```
vdiravka: Deleted

arina-ielchiieva:
+1

PR #815, opened 2017-04-11
DRILL-5424: Fix IOBE for reverse function

vvysotskyi:


exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java
```diff
@@ -273,4 +273,13 @@ public void testSplit() throws Exception {
...
 
+  @Test
+  public void testReverse() throws Exception {
+    testBuilder()
+      .sqlQuery("select reverse(reverse(n_comment)) words from cp.`tpch/nation.parquet`")
+      .unOrdered()
+      .sqlBaselineQuery("select n_comment words from cp.`tpch/nation.parquet`")
+      .build()
+      .run();
+  }
```
arina-ielchiieva: Unfortunately this unit test does not ensure that reverse function works correctly.
Let's replace this unit test with two unit tests:
1. test reverse result for one row like in `testSplit()` test.
2. test that reverse doesn't fail for table with several long varchars.

vvysotskyi: Thanks, added these both tests.

arina-ielchiieva:
+1

PR #817, opened 2017-04-13
DRILL-5429: Improve query performance for MapR DB JSON Tables

ppadma:


contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushFilterIntoScan.java
```diff
@@ -137,11 +137,13 @@ protected void doPushFilterIntoJsonGroupScan(RelOptRuleCall call,
...
     }
 
+    // Pass tableStats from old groupScan so we do not go and fetch stats (an expensive operation) again from MapR DB client.
     final JsonTableGroupScan newGroupsScan = new JsonTableGroupScan(groupScan.getUserName(),
                                                                     groupScan.getStoragePlugin(),
                                                                     groupScan.getFormatPlugin(),
                                                                     newScanSpec,
-                                                                    groupScan.getColumns());
+                                                                    groupScan.getColumns(),
+                                                                    groupScan.getTableStats());
```
gparai: We should try to use `clone()` here. All we are doing is copying stuff from one groupscan to another. `JsonTableGroupScan` already has a clone which clones everything except columns.

`@Override 
public GroupScan clone(List<SchemaPath> columns) 
{ JsonTableGroupScan newScan = new JsonTableGroupScan(this); 
newScan.columns = columns; return newScan; }`
We can create another which would clone everything except scanSpec. This can be used to pass in the `newScanSpec` generated here. Doing this would also clone the `regionsToScan` saving the call to `init()`.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonTableGroupScan.java
```diff
@@ -106,7 +115,11 @@ private void init() {
       Configuration conf = new Configuration();
       Table t = MapRDB.getTable(scanSpec.getTableName());
       TabletInfo[] tabletInfos = t.getTabletInfos(scanSpec.getCondition());
-      tableStats = new MapRDBTableStats(conf, scanSpec.getTableName());
+
+      // Fetch tableStats only once and cache it.
+      if (tableStats == null) {
```
gparai: This can probably be removed if we call clone(). However, it may be a useful check if we end up calling it from some other code-paths. Maybe add some logging to ensure we are not recreating the tableStats?

ppadma:
I changed as per your suggestion to use clone. We should recompute regionsToScan as it depends upon scanSpec. We can skip init by copying table and tabletInfo from old scan. Also, we can skip getting tableStats altogether as rowCount can be obtained from tabletInfo. 
Please review new diffs.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonTableGroupScan.java
```diff
@@ -100,30 +104,46 @@ public GroupScan clone(List<SchemaPath> columns) {
     return newScan;
   }
 
+  public JsonTableGroupScan clone(JsonScanSpec scanSpec) {
+    JsonTableGroupScan newScan = new JsonTableGroupScan(this);
+    newScan.scanSpec = scanSpec;
+    newScan.computeRegionsToScan();
+    return newScan;
+  }
+
```
gparai: Please add comments describing the function

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonTableGroupScan.java
```diff
@@ -100,30 +104,46 @@ public GroupScan clone(List<SchemaPath> columns) {
...
-        }
-        foundStartRegion = true;
-        regionsToScan.put(new TabletFragmentInfo(tabletInfoImpl), tabletInfo.getLocations()[0]);
-        if (!isNullOrEmpty(scanSpec.getStopRow())
-            && tabletInfoImpl.containsRow(scanSpec.getStopRow())) {
-          break;
-        }
+        totalRowCount += tabletInfo.getEstimatedNumRows();
       }
+
```
gparai: Please add your explanation as a comment 
> We should recompute regionsToScan as it depends upon scanSpec

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonTableGroupScan.java
```diff
@@ -100,30 +104,46 @@ public GroupScan clone(List<SchemaPath> columns) {
...
-      TabletInfo[] tabletInfos = t.getTabletInfos(scanSpec.getCondition());
-      tableStats = new MapRDBTableStats(conf, scanSpec.getTableName());
 
-      boolean foundStartRegion = false;
-      regionsToScan = new TreeMap<TabletFragmentInfo, String>();
+      // Fetch table and tabletInfo only once and cache.
+      table = MapRDB.getTable(scanSpec.getTableName());
+      tabletInfos = table.getTabletInfos(scanSpec.getCondition());
+
+      // Calculate totalRowCount for the table
```
gparai: Please add a comment explaining why we compute the totalRowCount like so? 
`totalRowCount += tabletInfo.getEstimatedNumRows();`

parthchandra:
+1

PR #819, opened 2017-04-25
DRILL-5419: Calculate return string length for literals & some string…

arina-ielchiieva:
… functions

1. Revisited calculation logic for string literals and some string functions (cast, upper, lower, initcap, reverse, concat, concat operator, rpad, lpad, case statement, coalesce, first_value, last_value, lag, lead).
2. Deprecated width and changed it to precision for string types.
3. Synchronized return type length calculation logic between limit 0 and regular queries.
4. Performed minor refactoring, added appropriate unit tests.


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/TypeInferenceUtils.java
```diff
@@ -394,31 +414,37 @@ public RelDataType inferReturnType(SqlOperatorBinding opBinding) {
   }
 
   private static class DrillConcatSqlReturnTypeInference implements SqlReturnTypeInference {
-    private static final DrillConcatSqlReturnTypeInference INSTANCE = new DrillConcatSqlReturnTypeInference();
```
jinfengni: For the various SqlReturnTypeInference, seems the same logic is applied here and in DrillConcatFuncHolder.  If we choose not to expand "FunctionScope" and model it as ReturnTypeInference, then we probably will be able to abstract and share the common code. That would save effort in the future to maintain two almost identical pieces of codes.

protocol/src/main/protobuf/Types.proto
```diff
@@ -70,8 +70,8 @@ enum MinorType {
 message MajorType {
   optional MinorType minor_type = 1;
   optional DataMode mode = 2;
-  optional int32 width = 3; // optional width for fixed size values.
-  optional int32 precision = 4; // used for decimal types
+  optional int32 width = 3;
+  optional int32 precision = 4; // used for decimal types or as optional length for fixed size values
```
jinfengni: optional length for "fixed" size value?

jinfengni:
2 additional comments.

1. Maybe we need include the length field in MaterializedField.toString() for varchar type if length is set. 
2. Will a varchar(5) vs varchar(10) be regarded as schema change? The underneath value vector would be same. Not sure if we may see more schema change exception. I tried to come up with some example, but could not find one. It may be worthwhile to think it over.

common/src/main/java/org/apache/drill/common/types/Types.java
```diff
@@ -27,10 +27,14 @@
 import org.apache.drill.common.types.TypeProtos.MinorType;
 
 import com.google.protobuf.TextFormat;
+import org.apache.drill.common.util.CoreDecimalUtility;
 
 public class Types {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Types.class);
 
+  public static final int MAX_VARCHAR_LENGTH = 65536;
```
jinfengni: Oracle seems to allow up to 4000 bytes for varchar [1].  The number Drill allows seems to be bigger than 4k.

This magic number probably comes from Calcite, if I understand correctly.  In RDBMS, if a string goes beyond the limit, people is recommended to use LOB/BLOB in stead. 

1. https://docs.oracle.com/cd/B28359_01/server.111/b28320/limits001.htm#i287903

common/src/main/java/org/apache/drill/common/types/Types.java
```diff
@@ -636,43 +658,63 @@ public static String toString(final MajorType type) {
...
-    if (type == MinorType.VARBINARY || type == MinorType.VARCHAR) {
-      return 65536;
-    }
-
     if (majorType.hasPrecision()) {
       return majorType.getPrecision();
     }
 
-    return 0;
+    return isScalarStringType(majorType) ? MAX_VARCHAR_LENGTH : UNDEFINED;
```
jinfengni: I believe what this PR tries to do is exactly as what you describe : derive length based on metadata. Such metadata comes from the function used , or the string literal.  For example, if you have a cast(some_input as varchar(20)), you will know the output type's max length is 20 (depending on the input, the output may only have values up to 15 characters).   

The max length is different from what could be obtained from {{RecordBatchSizer}}.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/TestHiveStorage.java
```diff
@@ -510,6 +517,38 @@ public void testIgnoreSkipHeaderFooterForSequencefile() throws Exception {
         .go();
   }
 
+  @Test
```
paul-rogers: Seems to be no test for the Decimal precision mapping. That may be why the issue mentioned in the code slipped through: there is no unit test to catch it...

exec/java-exec/src/main/java/org/apache/drill/exec/expr/annotations/FunctionTemplate.java
```diff
@@ -106,15 +106,21 @@
     DECIMAL_ADD_SCALE,
     DECIMAL_SET_SCALE,
     DECIMAL_ZERO_SCALE,
-    SC_BOOLEAN_OPERATOR
+    SC_BOOLEAN_OPERATOR,
```
paul-rogers: At the very least, call it something other than "scope." The word "scope" has a very precise meaning in CS: it is the set of symbols visible to an execution context. I must admit I'm struggling to see how, say, "CONCAT" fits this. Thanks to Jinfeng for explaining that FunctionScope has evolve to mean something like "FunctionHints"...

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillConcatFuncHolder.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.common.types.Types;
+
+import java.util.List;
+
+/**
+ * Function holder for functions with function scope set as
+ * {@link org.apache.drill.exec.expr.annotations.FunctionTemplate.FunctionScope#CONCAT}.
+ */
+public class DrillConcatFuncHolder extends DrillSimpleFuncHolder {
```
paul-rogers: As Jinfeng noted, we are extending a concept meant for one purpose and using it for another. We may not be stretching it to the breaking point yet, but I can see that coming.

What we really have is the idea of a function definition (what we call a "holder.") That definition has many aspects: signature, type inference, etc. We are using a single class with all properties and using inheritance to share properties. (Here, we are using inheritance for any concat-style function.)

But, what happens for a function that acts like concat in one area, but like something else in another? We have to copy & paste code to create a mixture class.

Better would be to hold each property as a separate class. That is, decorate the function definition with its types, with its cast rules, with its type inference rules, and so on, each expressed as a class.

By this:
```
public interface TypeInference {
  MajorType returnType();
}
public interface CastInference {
  Something inferCast();
}

Then, create instances that implement each rule set. Finally, create function rules that are collection of property rules:

FunctionDefn = new FunctionDefnBuilder()
  .returnType( concatReturnType )
  .cast( somethingCastRules )
  .name( "org.apache.drill.some.package.MyFunc" )
  .build();
```
The important thing isn't the specific classes. These are just by way of illustrating the general idea; which can (and probably would) be adjusted to fit.

Again, we may not need this quite yet. But, once we see copy & paste of method bodies, we'll know we need some refactoring.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillStringCastFuncHolder.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public class DrillStringCastFuncHolder extends DrillSimpleFuncHolder {
+
+  public DrillStringCastFuncHolder(FunctionAttributes functionAttributes, FunctionInitializer initializer) {
+    super(functionAttributes, initializer);
+  }
+
+  /**
+   * Defines function return type and sets cast length as type precision
+   * if cast length is simple long expression.
```
paul-rogers: Why just long?

In general, if we know the input type, we know the maximum output string width for that type (for all types except, say, Map, List, Repeated, VarBinary.) In many cases, the actual string won't be the maximum, but at least we set an upper bound far less than the default of 64K.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillStringLeftRightFuncHolder.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.drill.common.expression.ValueExpressions;
+import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.common.types.Types;
+import org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers;
+
+import java.util.List;
+
+/**
+ * Function holder for functions with function scope set as
+ * {@link org.apache.drill.exec.expr.annotations.FunctionTemplate.FunctionScope#STRING_LEFT_RIGHT}.
```
paul-rogers: I really appreciate the effort you put into adding Javadoc! Makes reviews SO much easier!

Maybe explain what STRING_LEFT_RIGHT means? Not entirely obvious to the casual observer...

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillStringLeftRightFuncHolder.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public class DrillStringLeftRightFuncHolder extends DrillSimpleFuncHolder  {
+
+  public DrillStringLeftRightFuncHolder(FunctionAttributes functionAttributes, FunctionInitializer initializer) {
+    super(functionAttributes, initializer);
+  }
+
+  /**
+   * Defines function return type and calculates output precision.
```
paul-rogers: Nit. That you've added comments is already a huge win. To make the generated Javadoc appear nicely, and to help us Eclipse users (Eclipse generated formatted Javadoc when we hover the cursor over the class or method name), please remember to include any needed HTML formatting tags, such as <p> between paragraphs, <br> at non-paragraph newlines -- you know the drill. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillStringLeftRightFuncHolder.java
```diff
@@ -0,0 +1,72 @@
...
+   *
+   * Target length calculation logic for left and right functions is the same,
+   * they substring string the same way, just from different sides of the string.
+   *
+   * <b>left(source, length)</b>
+   * <b>right(source, length)</b>
+   *
+   * <ul>If length is positive, target length is given length.</ul>
```
paul-rogers: In think you mean:
```
<ul>
<li>If length...</li>
<li>...</li>
</ul>
```

common/src/main/java/org/apache/drill/common/types/Types.java
jinfengni: The way this PR calculates the maximum length is not from "guess". The metadata information leveraged in this PR is in the "query" itself ( the function appearing in the query, or the string literal in the query). It does not apply to all the queries; only when the query provides enough information Drill will calculate the maximum length. In all other cases, Drill will perform same as before.

The maximum length (L1) calculated in this PR is different different from what you describe in the case of scanner (L2).  L2 <= L1. In the example of scanner with no metadata available, there would be no schema change; the precision is unset in type field, same as before. 
 
@arina-ielchiieva can correct me if my understanding is wrong.

arina-ielchiieva:
@jinfengni, @paul-rogers made changes after CR, Please review when possible.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/TestHiveStorage.java
jinfengni: I think it's fine to leave the Decimal related issues alone, as long as this PR does not introduce new issue to decimal functions.

jinfengni:
My understanding is Drill would hit a schema change between column with type of varchar(5) and varchar(10).  However, the length calculated in this PR is from the query metadeta (function parameter, string literal), which will not change across different batches. In that sense, we should not see schema change exception, just because of varchar length.

The following code shows that varchar(5) is different from varchar(10), also different from varchar without setting length. 

```java
   TypeProtos.MajorType varchar5 = TypeProtos.MajorType.newBuilder()
        .setMinorType(TypeProtos.MinorType.VARCHAR)
        .setMode(TypeProtos.DataMode.OPTIONAL)
        .setPrecision(5)
        .build();
    TypeProtos.MajorType varchar10 = TypeProtos.MajorType.newBuilder()
        .setMinorType(TypeProtos.MinorType.VARCHAR)
        .setMode(TypeProtos.DataMode.OPTIONAL)
        .setPrecision(10)
        .build();
    TypeProtos.MajorType varcharNoLength = TypeProtos.MajorType.newBuilder()
        .setMinorType(TypeProtos.MinorType.VARCHAR)
        .setMode(TypeProtos.DataMode.OPTIONAL)
        .build();

    System.out.println(varchar5.equals(varchar10) ? "varchar5 equals varchr10" : "varchar5 NOT equals varchr10" );
    System.out.println(varchar5.equals(varcharNoLength) ? "varchar5 equals varchrNoLength" : "varchar5 NOT equals varchrNoLength" );
```

output:
```
varchar5 NOT equals varchr10
varchar5 NOT equals varchrNoLength
```

jinfengni:
Regarding Paul's suggestion of using sampling (either 1st batch, or `n` batches), if the sampled length is returned to client as metadata for query result set, it could cause quite big problems for client. If the sampled max length is 5, client expects to see varchar up to 5 chars. If a new batch arrives with varchar(10), it would either crash the client, or make the client show incorrect result.  AFAIK, that's exactly what happened when Sean was working on 'LIMIT 0' optimization, and if there is an inconsistency of the type returned between 'LIMIT 0' and one returned from a regular query.

arina-ielchiieva:
@jinfengni 
1. Agree with the point
> However, the length calculated in this PR is from the query metadeta (function parameter, string literal), which will not change across different batches. In that sense, we should not see schema change exception, just because of varchar length.

2. I wanted to clarify regarding schema change, if I understood correctly from the code, two schemas are compared using `BatchSchema#equals` method [1] and when it compares two major types, precision and scale are not taken into account [2]. Am I correct?

[1] https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/record/BatchSchema.java#L101
[2] https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/record/BatchSchema.java#L144

jinfengni:
+1

The revised commit looks good to me. Thanks for taking effort to separate the return type inference logic from function scope logic. You may consider adding brief description of that change, plus the  refactoring of part of function code-gen logic (ValueReference etc).

arina-ielchiieva:
Squashed the commits and updated commit message to include brief description of the changes.

paul-rogers:
Three comments.

1. The equals() method for BatchSchema and MaterializedField should consider all fields, including precision. If it does not, that is likely a bug from when those fields were added. If we need a more general rule, then provide isSameType(), isSameMajorType() methods to be clearer.

2. It would be *VERY* useful to have a spec for this work. Trying to reverse engineer a spec from code changes is quite difficult. It will be hard for future maintainers to keep things working correctly without a spec to explain.

3. If/when we can extract schema hints from data, then it seems a simple process to feed that information into the system created here. So, we can tackle that as a separate discussion.

jinfengni:
I agreed with @paul-rogers  that BatchSchema and MaterializedField should consider all fields (I once briefly mentioned to @paul-rogers I feel there might be bug in the equals/hashCode methods). 

Regarding Paul's 3 point, if we can get the schema infor (varchar length) from data, i.e. the underneath storage plugin (RDBMS) provides such information, when Drill gets the batch from recordReader, then Arina's patch will work the same way as as the input is a string literal. In that sense, the work in this patch is complimentary to how we know the table column's varchar length.

PR #820, opened 2017-04-26
DRILL-5391: CTAS: make folder and file permission configurable

arina-ielchiieva:
1. Added new configuration option exec.persistent_table.umask which defaults to 002. User can modify this option on session or system level. If umask was set incorrectly, default umask will be used (002) and error will be logged.
2. Refactored `StorageStrategy` to use umask.
3. Added appropriate unit tests.


jinfengni:
+1

PR #821, opened 2017-04-27
DRILL-5450: Fix initcap function to convert upper case characters cor…

arina-ielchiieva:
…rectly

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctionHelpers.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctionHelpers.java
@@ -1,4 +1,4 @@
-/*******************************************************************************
+/*
 
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
@@ -15,7 +15,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- ******************************************************************************/
+*/
 package org.apache.drill.exec.expr.fn.impl;
 
 import io.netty.buffer.DrillBuf;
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctionHelpers.java
```diff
@@ -169,7 +169,7 @@ public static void initCap(int start, int end, DrillBuf inBuf, DrillBuf outBuf)
         if (currentByte >= 0x30 && currentByte <= 0x39) { // 0-9
           // noop
         } else if (currentByte >= 0x41 && currentByte <= 0x5A) { // A-Z
-          currentByte -= 0x20; // Lowercase this character
+          currentByte += 0x20; // Lowercase this character
```
paul-rogers: currentByte = Character.toLowerCase( currentByte )

The above handles all the Unicode complexity -- no need for us to reimplement it here.

A concern might be performance. Try calling the above 10K times in a loop and this function 10K times. Is there a difference in cost?

Ben-Zvi: toLowerCase() is implemented as a big switch statement for the Unicode, so very little cost ....

arina-ielchiieva: I did not notice any significant difference in performance, so will replace to `Character` methods.

arina-ielchiieva:
@paul-rogers 
I have changed customer lower / upper implementation in favor of `Character` methods.
Made changes in lower, upper and initcap functions.
Please review when possible.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctionHelpers.java
```diff
@@ -144,41 +144,28 @@ public static int varTypesToInt(final int start, final int end, DrillBuf buffer)
...
-        if (currentByte >= 0x30 && currentByte <= 0x39) { // 0-9
-          // noop
-        } else if (currentByte >= 0x41 && currentByte <= 0x5A) { // A-Z
-          currentByte -= 0x20; // Lowercase this character
-        } else if (currentByte >= 0x61 && currentByte <= 0x7A) { // a-z
-          // noop
-        } else { // whitespace
-          capNext = true;
-        }
+      int currentByte = inBuf.getByte(id);
```
paul-rogers: This code works only for ASCII, but not for UTF-8. UTF-8 is a multi-byte code that requires special encoding/decoding to convert to Unicode characters. Without that encoding, this method won't work for Cyrillic, Greek or any other character set with upper/lower distinctions.

Since this method never worked, it is probably OK to make it a bit less broken than before: at least now it works for ASCII. Please add unit tests below, then file a JIRA, for the fact that this function does not work with UTF-8 despite the fact that Drill claims it supports UTF-8.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java
```diff
@@ -308,4 +308,58 @@ public void testReverseLongVarChars() throws Exception {
       FileUtils.deleteQuietly(path);
     }
   }
+
+  @Test
+  public void testLower() throws Exception {
+    testBuilder()
+        .sqlQuery("select\n" +
+            "lower('ABC') col_upper,\n" +
+            "lower('abc') col_lower,\n" +
```
paul-rogers: Please add tests for Greek and Cyrillic. Our source encoding is UTF-8, so you can enter the characters directly. Or, if that does not work, you can instead use the Java Unicode encoding: U1234.

If the tests fail because of parsing of SQL, please file a bug. If they fail because the function above does not support UTF-8, please file a different bug.

In either case, you can then comment out the test cases and add a comment that says that they fail due to DRILL-xxxx, whatever your bug number turns out to be.

arina-ielchiieva: Created Jira DRILL-5477 and added appropriate unit test which is ignored for now.

parthchandra:
+1 so we can commit this. But Paul is right. This could be a lot better.

PR #824, opened 2017-05-03
DRILL-3867: Store relative paths in metadata file

vdiravka:
For easier review this PR consists of two commits, which can be squashed into one.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
@@ -27,7 +27,6 @@ import java.util.Iterator;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
-import com.google.common.base.Joiner;
 import org.apache.drill.common.expression.SchemaPath;
 import org.apache.drill.common.util.DrillVersionInfo;
 import org.apache.drill.exec.store.AbstractRecordReader;
@@ -158,13 +157,13 @@ public class Metadata {
   }
 
   /**
-   * Create the parquet metadata files for the directory at the given path, and for any subdirectories.
-   * Result metadata cache files and returned Pair of metadata contain relative paths.
+   * Create the parquet metadata files for the directory at the given path and for any subdirectories.
+   * Metadata cache files written to the disk contain relative paths. Returned Pair of metadata contains absolute paths.
    *
    * @param path to the directory of the parquet table
-   * @return Pair of parquet metadata. The left one is a parquet metadata for the table (metadata for every
-   *         parquet file). The right one of the Pair is a metadata for all subdirectories (if they are present
-   *         and there are no any parquet files in the {@code path} directory).
+   * @return Pair of parquet metadata. The left one is a parquet metadata for the table. The right one of the Pair is
+   *         a metadata for all subdirectories (if they are present and there are no any parquet files in the
+   *         {@code path} directory).
    * @throws IOException if parquet metadata can't be serialized and written to the json file
    */
   private Pair<ParquetTableMetadata_v3, ParquetTableMetadataDirs>
@@ -182,18 +181,10 @@ public class Metadata {
 
     for (final FileStatus file : fs.listStatus(p, new DrillPathFilter())) {
       if (file.isDirectory()) {
-        String subdirectoryName = file.getPath().getName();
         ParquetTableMetadata_v3 subTableMetadata = (createMetaFilesRecursively(file.getPath().toString())).getLeft();
-        for (ParquetFileMetadata_v3 pfm_v3 : subTableMetadata.files) {
-          // Construction of the relative file path by adding subdirectory name and inner relative file path
-          String relativePath = Joiner.on("/").join(subdirectoryName, pfm_v3.getPath());
-          metaDataList.add(new ParquetFileMetadata_v3(relativePath, pfm_v3.length, pfm_v3.rowGroups));
-        }
-        for (String directory : subTableMetadata.getDirectories()) {
-          String relativeDirPath = Joiner.on("/").join(subdirectoryName, directory);
-          directoryList.add(relativeDirPath);
-        }
-        directoryList.add(subdirectoryName);
+        metaDataList.addAll(subTableMetadata.files);
+        directoryList.addAll(subTableMetadata.directories);
+        directoryList.add(file.getPath().toString());
         // Merge the schema from the child level into the current level
         //TODO: We need a merge method that merges two colums with the same name but different types
         columnTypeInfoSet.putAll(subTableMetadata.columnTypeInfo);
@@ -204,7 +195,7 @@ public class Metadata {
     ParquetTableMetadata_v3 parquetTableMetadata = new ParquetTableMetadata_v3(DrillVersionInfo.getVersion());
     if (childFiles.size() > 0) {
       List<ParquetFileMetadata_v3 > childFilesMetadata =
-          getParquetFileMetadata_v3(parquetTableMetadata, childFiles, false);
+          getParquetFileMetadata_v3(parquetTableMetadata, childFiles);
       metaDataList.addAll(childFilesMetadata);
       // Note that we do not need to merge the columnInfo at this point. The columnInfo is already added
       // to the parquetTableMetadata.
@@ -218,16 +209,21 @@ public class Metadata {
     }
     parquetTableMetadata.columnTypeInfo.putAll(columnTypeInfoSet);
 
-    for (String oldname : OLD_METADATA_FILENAMES) {
-      fs.delete(new Path(p, oldname), false);
+    for (String oldName : OLD_METADATA_FILENAMES) {
+      fs.delete(new Path(p, oldName), false);
     }
-    writeFile(parquetTableMetadata, new Path(p, METADATA_FILENAME));
+    //  relative paths in the metadata are only necessary for meta cache files.
+    ParquetTableMetadata_v3 metadataTableWithRelativePaths =
+        MetadataPathUtils.createMetadataWithRelativePaths(parquetTableMetadata, path);
+    writeFile(metadataTableWithRelativePaths, new Path(p, METADATA_FILENAME));
 
     if (directoryList.size() > 0 && childFiles.size() == 0) {
-      ParquetTableMetadataDirs parquetTableMetadataDirs = new ParquetTableMetadataDirs(directoryList);
-      writeFile(parquetTableMetadataDirs, new Path(p, METADATA_DIRECTORIES_FILENAME));
+      ParquetTableMetadataDirs parquetTableMetadataDirsRelativePaths =
+          new ParquetTableMetadataDirs(metadataTableWithRelativePaths.directories);
+      writeFile(parquetTableMetadataDirsRelativePaths, new Path(p, METADATA_DIRECTORIES_FILENAME));
       logger.info("Creating metadata files recursively took {} ms", timer.elapsed(TimeUnit.MILLISECONDS));
       timer.stop();
+      ParquetTableMetadataDirs parquetTableMetadataDirs = new ParquetTableMetadataDirs(directoryList);
       return Pair.of(parquetTableMetadata, parquetTableMetadataDirs);
     }
     List<String> emptyDirList = Lists.newArrayList();
@@ -266,7 +262,7 @@ public class Metadata {
   private ParquetTableMetadata_v3 getParquetTableMetadata(List<FileStatus> fileStatuses)
       throws IOException {
     ParquetTableMetadata_v3 tableMetadata = new ParquetTableMetadata_v3();
-    List<ParquetFileMetadata_v3> fileMetadataList = getParquetFileMetadata_v3(tableMetadata, fileStatuses, true);
+    List<ParquetFileMetadata_v3> fileMetadataList = getParquetFileMetadata_v3(tableMetadata, fileStatuses);
     tableMetadata.files = fileMetadataList;
     tableMetadata.directories = new ArrayList<String>();
     return tableMetadata;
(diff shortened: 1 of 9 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -526,6 +534,48 @@ private void writeFile(ParquetTableMetadataDirs parquetTableMetadataDirs, Path p
   }
 
   /**
+   * Serializer for ParquetPath. Writes the path relative to the root path
+   */
```
paul-rogers: Why compute the relative during serialization? A more common approach is to store the relative paths in our internal data structures and serialize that relative value.

Doing things this way is awkward: we need to define a serializer and deserializer unnecessarily.

Also, see comment later on backward compatibility.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -526,6 +534,48 @@ private void writeFile(ParquetTableMetadataDirs parquetTableMetadataDirs, Path p
...
+
+    ParquetPathSerializer(String rootPath) {
+      super(ParquetPath.class);
+      this.rootPath = rootPath;
+    }
+
+    @Override
+    public void serialize(ParquetPath parquetPath, JsonGenerator jsonGenerator, SerializerProvider serializerProvider) throws IOException, JsonGenerationException {
+      Preconditions.checkState(parquetPath.getFullPath().startsWith(rootPath), String.format("Path %s is not a subpath of %s", parquetPath.getFullPath(), rootPath));
+      String relativePath = parquetPath.getFullPath().replaceFirst(rootPath, "");
```
paul-rogers: Java defines a Path abstraction that will compute relative paths.

An [old fashioned way](http://stackoverflow.com/questions/204784/how-to-construct-a-relative-path-in-java-from-two-absolute-paths-or-urls):
```
String relative = new File(base).toURI().relativize(new File(path).toURI()).getPath();
```

From the same post, the Java 7 way:
```
        Path pathAbsolute = Paths.get("/var/data/stuff/xyz.dat");
        Path pathBase = Paths.get("/var/data");
        Path pathRelative = pathBase.relativize(pathAbsolute);
```

Hadoop's path may have something similar.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -680,7 +731,7 @@ private boolean tableModified(List<String> directories, Path metaFilePath,
   }
 
   public static abstract class ParquetFileMetadata {
-    @JsonIgnore public abstract String getPath();
+    @JsonIgnore public abstract ParquetPath getParquetPath();
```
paul-rogers: Doing this changes the on-disk format for the metadata file, doesn't it? If we do that, we need to introduce a new file version. Since metadata is expensive, we'd have to be able to read the existing file format. I don't see code for any of that.

To address the issue, we can instead leave the field as a string. Treat the string as either relative or absolute. This should be easy to detect: "/this/is/absolute", "but/this/is/relative".

Then, create a method to set the paths. When setting paths, convert them to relative. When retrieving them, give a base directory. Convert relative (new) paths to absolute, leave (old) absolute paths unchanged. This is exactly how browsers handle URLs, OS's handle paths and so on. Classic approach.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1029,14 +1099,14 @@ public ParquetTableMetadata_v2(String drillVersion) {
     }
 
     public ParquetTableMetadata_v2(ParquetTableMetadataBase parquetTable,
-        List<ParquetFileMetadata_v2> files, List<String> directories, String drillVersion) {
+        List<ParquetFileMetadata_v2> files, List<ParquetPath> directories, String drillVersion) {
```
paul-rogers: If this is left as-is we need a new metadata file version. See above.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -398,6 +398,23 @@ public void testDrill4877() throws Exception {
...
+    test(String.format("refresh table metadata %s", tableName));
+    checkForMetadataFile(tableName);
+    File srcFile = new File(getDfsTestTmpSchemaLocation(), tableName);
+    File dstFile = new File(getDfsTestTmpSchemaLocation(), newTableName);
+    FileUtils.moveDirectory(srcFile, dstFile);
+    Assert.assertFalse("Cache file was not moved successfully", srcFile.exists());
+    int rowCount = testSql(String.format("select * from %s", newTableName));
+    Assert.assertEquals(50, rowCount);
+  }
+
```
paul-rogers: The tests here don't verify opening a version 1.10 metadata file with this 1.11 change. What happens? Can we read old files?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -179,10 +182,18 @@ private Metadata(FileSystem fs, ParquetFormatConfig formatConfig) {
...
     for (final FileStatus file : fs.listStatus(p, new DrillPathFilter())) {
       if (file.isDirectory()) {
+        String subdirectoryName = file.getPath().getName();
         ParquetTableMetadata_v3 subTableMetadata = (createMetaFilesRecursively(file.getPath().toString())).getLeft();
-        metaDataList.addAll(subTableMetadata.files);
-        directoryList.addAll(subTableMetadata.directories);
-        directoryList.add(file.getPath().toString());
+        for (ParquetFileMetadata_v3 pfm_v3 : subTableMetadata.files) {
+          // Construction of the relative file path by adding subdirectory name and inner relative file path
+          String relativePath = Joiner.on("/").join(subdirectoryName, pfm_v3.getPath());
```
paul-rogers: `Path.mergePaths()`?

We really don't want to work with paths as strings: such code is hard to test and maintain.

If we need new Path operations (such as merging relative paths), I suggest we create a `PathUtils` class to hold the operations. Then, create unit tests to check all the various conditions: empty head, empty tail, neither empty, etc.

Also, in general, we would work with path names as `Path` objects: the job of the `Path` class is do properly implement file path operations, just as the job of the older `File` and newer `Path` classes in Java is to handle OS paths.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -264,15 +275,18 @@ private ParquetTableMetadata_v3 getParquetTableMetadata(List<FileStatus> fileSta
...
+   * @param fileStatuses list of the parquet files statuses
+   * @param absolutePathInMetadata true if result metadata files should contain absolute paths, false for relative paths.
+   *                               Relative paths in the metadata are only necessary while creating meta cache files.
+   * @return list of the parquet file metadata (parquet metadata for every file)
    * @throws IOException
    */
-  private List<ParquetFileMetadata_v3> getParquetFileMetadata_v3(
-      ParquetTableMetadata_v3 parquetTableMetadata_v3, List<FileStatus> fileStatuses) throws IOException {
+  private List<ParquetFileMetadata_v3> getParquetFileMetadata_v3(ParquetTableMetadata_v3 parquetTableMetadata_v3,
+      List<FileStatus> fileStatuses, boolean absolutePathInMetadata) throws IOException {
```
paul-rogers: Is this really needed? Or, is it an attempt to answer my earlier concern about compatibility?

Only newer Drill instances will create metadata. If we want relative paths, then we should always use relative paths. No need to pass along a flag.

On the other hand, if we are saying that the root call is absolute (as seen in the code earlier), but subdirectories are relative, then doesn't the presence of even one absolute directory name make the whole feature invalid?

Perhaps some more background explanation in the PR comments (or even a design spec) might shed some light on what we are trying to accomplish here. Very hard to simply reverse engineer a design from code changes...

Also, below, we have a method to convert relative paths to absolute in bulk. Should we do the same here? Always gather data in absolute form, then convert it to relative just before serializing?

I wasn't sure why we are converting paths from relative to absolute. If we are doing that because we use absolute paths internally, then it is OK to gather absolute paths here. Convert the to relative just before writing if that is easier.

Here, I'm referring to the note about the "proposed alternative solution".

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -748,6 +771,22 @@ public ParquetTableMetadataDirs(List<String> directories) {
       return directories;
     }
 
+    /** If directories list contains relative paths, update it to absolute ones
+     * @param baseDir base parent directory
+     */
+    @JsonIgnore public void updateRelativePaths(Path baseDir) {
+      if (!directories.isEmpty()) {
+        // It is enough to check the first path to decide if updating needed
+        if (!new Path(directories.get(0)).isAbsolute()) {
```
paul-rogers: This is getting a bit silly, converting to/from String and Path. The `directories` list should contain Path elements. In general, we should never work with Path as strings except if we need to serialize as in Jackson.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -748,6 +771,22 @@ public ParquetTableMetadataDirs(List<String> directories) {
       return directories;
     }
 
+    /** If directories list contains relative paths, update it to absolute ones
```
paul-rogers: Explanation of why we do this? Do we store data in relative form, but convert it to absolute form internally?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1413,6 +1452,31 @@ public ColumnTypeMetadata_v3 getColumnTypeInfo(String[] name) {
       return directories;
     }
 
+    /** If directories list and file metadata list contain relative paths, update it to absolute ones
+     * @param baseDir base parent directory
+     */
+    @JsonIgnore public void updateRelativePaths(Path baseDir) {
```
paul-rogers: How is this method different from the previous one? Looks like they are doing a bit of the same thing...

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
paul-rogers: Will they stay absolute? Or, are they rebuilt from time to time?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
vdiravka: It is possible to replace String with Path for directories paths due to implementing custom `JsonSerializer<Path>` and `JsonDeserializer<Path>`. But then it will be necessary to convert every `Path` from lists back into `String`, because a String paths are used in a lot of places: `FileSelection`, `Metadata`, `ParquetGroupScan`, `ReadEntryWithPath`, `FileWork`, `FormatSelection`, `FormatPlugin`,  `PartitionLocation`and so on.

I am totally agree with replacing `String` with `Path` requirement. But it should be done not only for parquet and in context of separate jira. I am going to create it.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
vdiravka: I combined the general code for this two methods and created a separate helper methods.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vdiravka: There is no requirement for them to use absolute paths. After this fix they can be upgraded to use relative paths. (I'm going to open separate jira for it). Therefore a new test case for metadata cache files with absolute paths was added.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
paul-rogers: Thanks for the explanation.

exec/java-exec/src/test/java/org/apache/drill/BaseTestQuery.java
```diff
@@ -593,4 +610,49 @@ private void convert(List<QueryDataBatch> batches) throws SchemaChangeException
...
+  protected static void copyDirectoryIntoTempSpace(String resourcesDir, String destinationSubDir) throws IOException {
+    Path destination = destinationSubDir != null ? new Path(getDfsTestTmpSchemaLocation(), destinationSubDir)
+        : new Path(getDfsTestTmpSchemaLocation());
+    fs.copyFromLocalFile(
+        new Path(replaceWorkingPathInString(resourcesDir)),
+        destination);
+  }
+
+  /**
+   * Metadata cache files include full paths to the files that have been scanned.
```
paul-rogers: For older files with the marker, should we just replace the marker to be relative and take advantage of this improvement? Can that be done without having to edit the old files?

exec/java-exec/src/test/java/org/apache/drill/BaseTestQuery.java
```diff
@@ -593,4 +610,49 @@ private void convert(List<QueryDataBatch> batches) throws SchemaChangeException
...
+   *
+   * There is no way to generate a metadata cache file with absolute paths that
+   * will be guaranteed to be available on an arbitrary test machine.
+   *
```
paul-rogers: Very small suggestion: Javadoc is HTML-formatted. Insert a <p> between paragraphs.

PR #826, opened 2017-05-05
DRILL-5379: Set Hdfs Block Size based on Parquet Block Size

ppadma:
Provide an option to specify blocksize during file creation.
This will help create parquet files with single block on HDFS, helping improve
performance when we read those files.
See DRILL-5379 for details.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -160,6 +160,9 @@ public interface ExecConstants {
   OptionValidator OUTPUT_FORMAT_VALIDATOR = new StringValidator(OUTPUT_FORMAT_OPTION, "parquet");
   String PARQUET_BLOCK_SIZE = "store.parquet.block-size";
   OptionValidator PARQUET_BLOCK_SIZE_VALIDATOR = new LongValidator(PARQUET_BLOCK_SIZE, 512*1024*1024);
+  String PARQUET_WRITER_USE_SINGLE_FS_BLOCK = "store.parquet.writer.use_single_fs_block";
+  OptionValidator PARQUET_WRITER_USE_SINGLE_FS_BLOCK_VALIDATOR = new BooleanValidator(
+    PARQUET_WRITER_USE_SINGLE_FS_BLOCK, false);
   String PARQUET_PAGE_SIZE = "store.parquet.page-size";
   OptionValidator PARQUET_PAGE_SIZE_VALIDATOR = new LongValidator(PARQUET_PAGE_SIZE, 1024*1024);
   String PARQUET_DICT_PAGE_SIZE = "store.parquet.dictionary.page-size";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.store.parquet;
 
+import static java.lang.Math.ceil;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
 
@@ -77,6 +78,7 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
   private static final int MINIMUM_BUFFER_SIZE = 64 * 1024;
   private static final int MINIMUM_RECORD_COUNT_FOR_CHECK = 100;
   private static final int MAXIMUM_RECORD_COUNT_FOR_CHECK = 10000;
+  private static final int BLOCKSIZE_MULTIPLE = 64 * 1024;
 
   public static final String DRILL_VERSION_PROPERTY = "drill.version";
   public static final String WRITER_VERSION_PROPERTY = "drill-writer.version";
@@ -89,6 +91,7 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
   private int pageSize;
   private int dictionaryPageSize;
   private boolean enableDictionary = false;
+  private boolean useSingleFSBlock = false;
   private CompressionCodecName codec = CompressionCodecName.SNAPPY;
   private WriterVersion writerVersion = WriterVersion.PARQUET_1_0;
   private CodecFactory codecFactory;
@@ -156,6 +159,12 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
     }
 
     enableDictionary = Boolean.parseBoolean(writerOptions.get(ExecConstants.PARQUET_WRITER_ENABLE_DICTIONARY_ENCODING));
+    useSingleFSBlock = Boolean.parseBoolean(writerOptions.get(ExecConstants.PARQUET_WRITER_USE_SINGLE_FS_BLOCK));
+
+    if (useSingleFSBlock) {
+      // Round up blockSize to multiple of 64K.
+      blockSize = (int)ceil((double)blockSize/BLOCKSIZE_MULTIPLE) * BLOCKSIZE_MULTIPLE;
+    }
   }
 
   private boolean containsComplexVectors(BatchSchema schema) {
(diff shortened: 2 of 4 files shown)


ppadma:
@parthchandra can you review this ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -380,14 +384,21 @@ public void endRecord() throws IOException {
 
       // since ParquetFileWriter will overwrite empty output file (append is not supported)
       // we need to re-apply file permission
-      parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE);
+      if (useConfiguredBlockSize) {
```
parthchandra: The API `ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE)` will cause the Parquet file writer to set the file block size to the greater of the configured files system block size or 128 MB (the ParquetWriter's row group size). 
Drill's Parquet writer will use the block size specified in Drill's options to create a new Parquet row group when the limit is reached (See `ParquetRecodWriter.checkBlockSizeReached()` ). If you set Drill's Parquet block size to the larger of the configured file system block size or 128 MB, you will get the row group to match the HDFS block size. 
Which is what the current code does.
Isn't this what the original JIRA wanted?

ppadma: What we are doing is create parquet file as single block without changing the file system default block size.  For ex. default Parquet block size is 512MB and if file system block size is 128MB, we create single file with 4 blocks on filesystem, which can get distributed on different nodes, not good for performance. If we change Parquet block size to 128MB (to match with file system block size), for same amount of data, we end up creating 4 files, one block each, which is not good either. 

JIRA wanted single HDFS block per Parquet file that is larger than file system block size , without changing file system block size.  They had file system block size configured as 128MB. Lowering parquet block size (from default value of 512MB) to match with file system block size is creating too many files for them. For some other reasons, they are not able to change file system block size.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -380,14 +384,21 @@ public void endRecord() throws IOException {
...
+        // Round up blockSize to multiple of 64K.
+        long writeBlockSize = ((long) ceil((double)blockSize/BLOCKSIZE_MULTIPLE)) * BLOCKSIZE_MULTIPLE;
```
parthchandra: This is not quite consistent with the use of block size in the `checkBlockSizeReached` function. You want to use the same size in both places.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -380,14 +384,21 @@ public void endRecord() throws IOException {
...
+        // Passing writeBlockSize creates files with this blockSize instead of filesystem default blockSize.
+        // Currently, this is supported only by filesystems included in
+        // BLOCK_FS_SCHEMES (ParquetFileWriter.java in parquet-mr), which includes HDFS.
+        // For other filesystems, it uses default blockSize configured for the file system.
+        parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE, writeBlockSize, 0);
+      } else {
+        parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE);
```
parthchandra: This breaks previous behaviour. If a user had previously changed the parquet block size, they will not get Parquet RowGrougs of that size, but will get the default 512MB

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -160,6 +160,9 @@
   OptionValidator OUTPUT_FORMAT_VALIDATOR = new StringValidator(OUTPUT_FORMAT_OPTION, "parquet");
   String PARQUET_BLOCK_SIZE = "store.parquet.block-size";
   OptionValidator PARQUET_BLOCK_SIZE_VALIDATOR = new LongValidator(PARQUET_BLOCK_SIZE, 512*1024*1024);
+  String PARQUET_WRITER_USE_CONFIGURED_BLOCK_SIZE = "store.parquet.writer.use_configured_block-size";
```
kkhatua: Making my contribution to the review thread:
`match_filesystem_block-size`
The value would be the logical inverse of the original flag.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
ppadma: No. It does not. If user had previously changed the parquet block size, that will be used in checkBlockSizeReached as before. So, parquet rowGroup size will be same as configured parquet blockSize (not default). This option only changes the behavior of how we write that parquet row group onto disk i.e. single block or not.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
ppadma: Underlying FS API expects blockSize specified during create to be multiple of 64K.  So, I rounded it up.  This will be greater than (to nearest 64K) or equal to what is used in checkBlockSizeReached.  Should I change checkBlockSizeReached to do the same ? I was trying to avoid any changes to existing behavior/code.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
ppadma: ok, I went with "store.parquet.writer.use_single_fs_block".  Please review updated changes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
parthchandra: Right, it doesn't. The changing of the option name (and semantics) make this much clearer.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -380,14 +384,21 @@ public void endRecord() throws IOException {
 
       // since ParquetFileWriter will overwrite empty output file (append is not supported)
       // we need to re-apply file permission
-      parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE);
+      if (useSingleFSBlock) {
+        // Round up blockSize to multiple of 64K.
+        long writeBlockSize = ((long) ceil((double)blockSize/BLOCKSIZE_MULTIPLE)) * BLOCKSIZE_MULTIPLE;
```
parthchandra: Would it be better to change the blockSize itself so that checkBlockSize also gets it if the option is set.

ppadma: ok. I made the change. Please review updated diffs.

parthchandra:
+1

kkhatua:
@ppadma , Khurram [~khfaraaz] and I were looking at the details in the PR and it's not very clear what new behavior does the PR allow. If you need to specify the block-size as described in the [comment ](https://issues.apache.org/jira/browse/DRILL-5379?focusedCommentId=15981366&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15981366)by @fmethot , doesn't Drill already do that? I thought Drill implicitly creates files with a single row-group anyway. 

My understanding of the JIRA's problem statement was that if the Parquet block-size (i.e. the rowgroup size) is set to a large value that exceeds the HDFS block size, using the flag would allow Drill to ignore the larger value in the options and write with a parquet-blocksize that matches the target HDFS location. So, I could have {{store.parquet.block-size=1073741824}} (i.e. 1GB), but when writing an output worth 512MB, instead of 1 file... Drill would read the HDFS block-size (say 128GB) and apply that as the parquet-block-size and write 4 files. 

@fmethot is that what you were looking for? An **automatic scaling down** of the parquet file's size to match (and be contained within) the HDFS block size?

ppadma:
@kkhatua  HDFS allows specifying block size during file creation which overrides the default file system block size. With this PR, we can have single HDFS block per Parquet file that can be larger than file system block size , without changing file system default block size. I know it is confusing. But, it is possible to create file with  block size that is different from file system default block size.

PR #827, opened 2017-05-05
DRILL-5481: Persist profiles in-memory only with a max capacity

kkhatua:
Refactored NoWriteLocalStore to  InMemoryPersistentStore with the ability to maintain a max capacity
Updated Default values, and added Zookeeper's property to ExecConstants

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -107,8 +107,9 @@ public interface ExecConstants {
   String SYS_STORE_PROVIDER_CLASS = "drill.exec.sys.store.provider.class";
   String SYS_STORE_PROVIDER_ZK_BLOBROOT = "drill.exec.sys.store.provider.zk.blobroot";
   String SYS_STORE_PROVIDER_LOCAL_PATH = "drill.exec.sys.store.provider.local.path";
-  String SYS_STORE_PROVIDER_LOCAL_INMEMORY_WRITE = "drill.exec.sys.store.provider.local.inmemory.write";
-  String SYS_STORE_CAPACITY_PROFILES = "drill.exec.sys.store.capacity.profiles";
+  String SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE = "drill.exec.sys.store.provider.local.enable.write";
+  String SYS_STORE_PROFILES_INMEMORY = "drill.exec.sys.store.profiles.inmemory";
+  String SYS_STORE_PROFILES_CAPACITY = "drill.exec.sys.store.profiles.capacity";
   String IMPERSONATION_ENABLED = "drill.exec.impersonation.enabled";
   String IMPERSONATION_MAX_CHAINED_USER_HOPS = "drill.exec.impersonation.max_chained_user_hops";
   String AUTHENTICATION_MECHANISMS = "drill.exec.security.auth.mechanisms";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -105,8 +105,10 @@ public interface ExecConstants {
   String HTTP_TRUSTSTORE_PATH = "javax.net.ssl.trustStore";
   String HTTP_TRUSTSTORE_PASSWORD = "javax.net.ssl.trustStorePassword";
   String SYS_STORE_PROVIDER_CLASS = "drill.exec.sys.store.provider.class";
+  String SYS_STORE_PROVIDER_ZK_BLOBROOT = "drill.exec.sys.store.provider.zk.blobroot";
   String SYS_STORE_PROVIDER_LOCAL_PATH = "drill.exec.sys.store.provider.local.path";
-  String SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE = "drill.exec.sys.store.provider.local.write";
+  String SYS_STORE_PROVIDER_LOCAL_INMEMORY_WRITE = "drill.exec.sys.store.provider.local.inmemory.write";
+  String SYS_STORE_CAPACITY_PROFILES = "drill.exec.sys.store.capacity.profiles";
   String IMPERSONATION_ENABLED = "drill.exec.impersonation.enabled";
   String IMPERSONATION_MAX_CHAINED_USER_HOPS = "drill.exec.impersonation.max_chained_user_hops";
   String AUTHENTICATION_MECHANISMS = "drill.exec.security.auth.mechanisms";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/PersistentStoreConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/PersistentStoreConfig.java
@@ -38,11 +38,17 @@ public class PersistentStoreConfig<V> {
   private final String name;
   private final InstanceSerializer<V> valueSerializer;
   private final PersistentStoreMode mode;
+  private final int maxCapacity;
 
-  protected PersistentStoreConfig(String name, InstanceSerializer<V> valueSerializer, PersistentStoreMode mode) {
+  protected PersistentStoreConfig(String name, InstanceSerializer<V> valueSerializer, PersistentStoreMode mode, int maxCapacity) {
     this.name = name;
     this.valueSerializer = valueSerializer;
     this.mode = mode;
+    this.maxCapacity = maxCapacity;
+  }
+
+  public int getMaxCapacity() {
+    return maxCapacity;
   }
 
   public PersistentStoreMode getMode() {
@@ -85,6 +91,7 @@ public class PersistentStoreConfig<V> {
     private String name;
     private InstanceSerializer<V> serializer;
     private PersistentStoreMode mode = PersistentStoreMode.PERSISTENT;
+    private int maxCapacity = Integer.MAX_VALUE;
 
     protected StoreConfigBuilder(InstanceSerializer<V> serializer) {
       super();
@@ -106,9 +113,14 @@ public class PersistentStoreConfig<V> {
       return this;
     }
 
+    public StoreConfigBuilder<V> setMaxCapacity(int maxCapacity) {
+      this.maxCapacity = maxCapacity;
+      return this;
+    }
+
     public PersistentStoreConfig<V> build(){
       Preconditions.checkNotNull(name);
-      return new PersistentStoreConfig<>(name, serializer, mode);
+      return new PersistentStoreConfig<>(name, serializer, mode, maxCapacity);
     }
   }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/PersistentStoreConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/PersistentStoreConfig.java
@@ -38,13 +38,19 @@ public class PersistentStoreConfig<V> {
   private final String name;
   private final InstanceSerializer<V> valueSerializer;
   private final PersistentStoreMode mode;
+  private final boolean inMemory;
   private final int maxCapacity;
 
-  protected PersistentStoreConfig(String name, InstanceSerializer<V> valueSerializer, PersistentStoreMode mode, int maxCapacity) {
+  protected PersistentStoreConfig(String name, InstanceSerializer<V> valueSerializer, PersistentStoreMode mode, boolean inMemory, int maxCapacity) {
     this.name = name;
     this.valueSerializer = valueSerializer;
     this.mode = mode;
+    this.inMemory = inMemory;
     this.maxCapacity = maxCapacity;
+  }
+
+  public boolean isInMemory() {
+    return inMemory;
   }
 
   public int getMaxCapacity() {
@@ -91,6 +97,7 @@ public class PersistentStoreConfig<V> {
     private String name;
     private InstanceSerializer<V> serializer;
     private PersistentStoreMode mode = PersistentStoreMode.PERSISTENT;
+    private boolean inMemory = false;
     private int maxCapacity = Integer.MAX_VALUE;
 
     protected StoreConfigBuilder(InstanceSerializer<V> serializer) {
@@ -113,6 +120,11 @@ public class PersistentStoreConfig<V> {
       return this;
     }
 
+    public StoreConfigBuilder<V> setInMemory() {
+      this.inMemory = true;
+      return this;
+    }
+
     public StoreConfigBuilder<V> setMaxCapacity(int maxCapacity) {
       this.maxCapacity = maxCapacity;
       return this;
@@ -120,7 +132,7 @@ public class PersistentStoreConfig<V> {
 
     public PersistentStoreConfig<V> build(){
       Preconditions.checkNotNull(name);
-      return new PersistentStoreConfig<>(name, serializer, mode, maxCapacity);
+      return new PersistentStoreConfig<>(name, serializer, mode, inMemory, maxCapacity);
     }
   }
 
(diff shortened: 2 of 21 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/provider/LocalPersistentStoreProvider.java
```diff
@@ -27,28 +27,30 @@
...
-//  private static final Logger logger = LoggerFactory.getLogger(LocalPersistentStoreProvider.class);
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LocalPersistentStoreProvider.class);
 
   private final Path path;
   private final DrillFileSystem fs;
   // This flag is used in testing. Ideally, tests should use a specific PersistentStoreProvider that knows
   // how to handle this flag.
-  private final boolean enableWrite;
+  private final boolean useEphemeral;
+  private int maxCapacity;
```
ppadma: can we name this maxEphemeralCapacity or maxEphemeralProfiles ?

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -105,8 +105,10 @@
   String HTTP_TRUSTSTORE_PATH = "javax.net.ssl.trustStore";
   String HTTP_TRUSTSTORE_PASSWORD = "javax.net.ssl.trustStorePassword";
   String SYS_STORE_PROVIDER_CLASS = "drill.exec.sys.store.provider.class";
+  String SYS_STORE_PROVIDER_ZK_BLOBROOT = "drill.exec.sys.store.provider.zk.blobroot";
   String SYS_STORE_PROVIDER_LOCAL_PATH = "drill.exec.sys.store.provider.local.path";
-  String SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE = "drill.exec.sys.store.provider.local.write";
+  String SYS_STORE_PROVIDER_EPHEMERAL_WRITE = "drill.exec.sys.store.provider.ephemeral.write";
+  String SYS_STORE_PROVIDER_EPHEMERAL_CAPACITY = "drill.exec.sys.store.provider.ephemeral.capacity";
```
ppadma: How about ephemeral.maxProfiles instead of capacity ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
```diff
@@ -15,28 +15,42 @@
...
   private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
   private final AutoCloseableLock readLock = new AutoCloseableLock(readWriteLock.readLock());
   private final AutoCloseableLock writeLock = new AutoCloseableLock(readWriteLock.writeLock());
-  private final ConcurrentMap<String, V> store = Maps.newConcurrentMap();
+  private ConcurrentMap<String, V> store;
   private int version = -1;
+  private int maxCapacity;
+  private AtomicInteger currentSize = new AtomicInteger();
+
+  public EphemeralPersistentStore(final PersistentStoreConfig<V> config, int capacity) throws StoreException {
```
ppadma: seems like you don't need the config ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
```diff
@@ -15,28 +15,42 @@
...
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(EphemeralPersistentStore.class);
+
   private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
   private final AutoCloseableLock readLock = new AutoCloseableLock(readWriteLock.readLock());
   private final AutoCloseableLock writeLock = new AutoCloseableLock(readWriteLock.writeLock());
-  private final ConcurrentMap<String, V> store = Maps.newConcurrentMap();
+  private ConcurrentMap<String, V> store;
   private int version = -1;
+  private int maxCapacity;
+  private AtomicInteger currentSize = new AtomicInteger();
```
ppadma: I understand why we need this (instead of using store.size()).  But, please add a comment.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -105,8 +105,10 @@
   String HTTP_TRUSTSTORE_PATH = "javax.net.ssl.trustStore";
   String HTTP_TRUSTSTORE_PASSWORD = "javax.net.ssl.trustStorePassword";
   String SYS_STORE_PROVIDER_CLASS = "drill.exec.sys.store.provider.class";
+  String SYS_STORE_PROVIDER_ZK_BLOBROOT = "drill.exec.sys.store.provider.zk.blobroot";
   String SYS_STORE_PROVIDER_LOCAL_PATH = "drill.exec.sys.store.provider.local.path";
-  String SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE = "drill.exec.sys.store.provider.local.write";
+  String SYS_STORE_PROVIDER_EPHEMERAL_WRITE = "drill.exec.sys.store.provider.ephemeral.write";
```
sudheeshkatkam: This will break backward compatibility.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
```diff
@@ -15,28 +15,42 @@
...
 import org.apache.drill.exec.exception.VersionMismatchException;
 import org.apache.drill.exec.store.sys.BasePersistentStore;
+import org.apache.drill.exec.store.sys.PersistentStoreConfig;
 import org.apache.drill.exec.store.sys.PersistentStoreMode;
-import org.apache.drill.exec.store.sys.store.DataChangeVersion;
 
-public class NoWriteLocalStore<V> extends BasePersistentStore<V> {
+import com.google.common.collect.Iterables;
+
+public class EphemeralPersistentStore<V> extends BasePersistentStore<V> {
```
sudheeshkatkam: Please undo the oxymoron, and remove the EPHEMERAL mode to a PersistentStore. These are confusing.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
sudheeshkatkam: store.size() and currentSize have synchronization problems below!

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/provider/LocalPersistentStoreProvider.java
```diff
@@ -57,14 +59,14 @@ public LocalPersistentStoreProvider(final DrillConfig config) throws StoreExcept
   }
 
   @Override
-  public <V> PersistentStore<V> getOrCreateStore(PersistentStoreConfig<V> storeConfig) {
+  public <V> PersistentStore<V> getOrCreateStore(PersistentStoreConfig<V> storeConfig) throws StoreException {
     switch(storeConfig.getMode()){
     case BLOB_PERSISTENT:
     case PERSISTENT:
-      if (enableWrite) {
+      if (!useEphemeral) {
```
sudheeshkatkam: This is changing the default behavior, as enableWrite/useEphemeral is true by default.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
sudheeshkatkam: Never mind.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/provider/LocalPersistentStoreProvider.java
ppadma: yes, ephemeral write should be false by default. For backward compatibility, we should probably just stick with local write i.e. trigger ephemeral write when local write is false.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
kkhatua: There is already a writelock. However, with a SkipList, the call of using size() is expensive. To minimize this computation, we have an atomic counter.  With the writeLock, if I hit the max "virtual" capacity, I am simply inserting the latest entry and popping out the last (oldest entry). Unlike Concurrent Queues, ConcurrentSkipListMap allows natural ordering of the keys (which is required for serving via the WebUI).

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
kkhatua: Point taken! 👍 
Will rename it _InMemoryPersistentStore_

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
kkhatua: The Ephemeral store can also be used for non-profile related usage. Trying to keep it generic

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
```diff
@@ -93,6 +107,14 @@ public void put(final String key, final V value, final DataChangeVersion dataCha
         throw new VersionMismatchException("Version mismatch detected", dataChangeVersion.getVersion());
       }
       store.put(key, value);
+      currentSize.incrementAndGet();
+
+      if (currentSize.get() > maxCapacity) {
+        //Pop Out Oldest
+        ((ConcurrentSkipListMap<String, V>) store).pollLastEntry();
```
kkhatua: The natural ordering of the profiles actually has the 'smallest' key being the latest. Hence, we pop out the oldest entry by getting the 'largest'/last entry

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/provider/LocalPersistentStoreProvider.java
kkhatua: 👍  Ok.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -136,9 +136,10 @@ drill.exec: {
   },
   sys.store.provider: {
     class: "org.apache.drill.exec.store.sys.store.provider.ZookeeperPersistentStoreProvider",
-    local: {
-      path: "/tmp/drill",
-      write: true
+    local.path: "/tmp/drill",
+    ephemeral: {
+      write: true,
```
kkhatua: This should have been false as _ephemeral.write_ is inverse of _local.write_

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/provider/LocalPersistentStoreProvider.java
kkhatua: I've actually refactored the internal unit tests that leveraged this flag. It doesn't seem to be used outside of the tests.
I think the enableWrite is a little misleading. We already have a skipWritingProfiles flag introduced. Do you really think there is a backward compatibility that would be broken, considering this was hidden away in a 'test' package?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/EphemeralPersistentStore.java
kkhatua: That's true. Originally, I thought I might need it. Left it as is, assuming that future enhancements would not change the signature. 
I'll remove it.
👍

kkhatua:
@ppadma, @sudheeshkatkam 
Updated most of the changes except the flag itself. Waiting for a quick review and final comment on whether I can change the flag from _drill.exec.sys.store.provider.local.write_ to _drill.exec.sys.store.provider.local.inmemory.write_ .

kkhatua:
@sudheeshkatkam & @ppadma ... please review and bless the PR if everything looks fine.!

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/InMemoryPersistentStore.java
```diff
@@ -15,28 +15,40 @@
...
   private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
   private final AutoCloseableLock readLock = new AutoCloseableLock(readWriteLock.readLock());
   private final AutoCloseableLock writeLock = new AutoCloseableLock(readWriteLock.writeLock());
-  private final ConcurrentMap<String, V> store = Maps.newConcurrentMap();
+  private ConcurrentMap<String, V> store;
   private int version = -1;
+  private int maxCapacity;
+  private AtomicInteger currentSize = new AtomicInteger();
+
+  public InMemoryPersistentStore(int maximumCapacity) throws StoreException {
```
sudheeshkatkam: Does not throw?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/InMemoryPersistentStore.java
```diff
@@ -93,6 +105,14 @@ public void put(final String key, final V value, final DataChangeVersion dataCha
         throw new VersionMismatchException("Version mismatch detected", dataChangeVersion.getVersion());
       }
       store.put(key, value);
+      currentSize.incrementAndGet();
+
+      if (currentSize.get() > maxCapacity) {
+        //Pop Out Oldest
+        ((ConcurrentSkipListMap<String, V>) store).pollLastEntry();
```
sudheeshkatkam: Rather than cast, declare store as this type?

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -138,7 +138,10 @@ drill.exec: {
     class: "org.apache.drill.exec.store.sys.store.provider.ZookeeperPersistentStoreProvider",
     local: {
       path: "/tmp/drill",
-      write: true
+      inmemory: {
+        write: false,
+        capacity: 1000
```
sudheeshkatkam: Remove this line, otherwise `if (config.hasPath(ExecConstants.SYS_STORE_CAPACITY_PROFILES))` is always true.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/PersistentStoreConfig.java
```diff
@@ -106,9 +113,14 @@ protected StoreConfigBuilder(InstanceSerializer<V> serializer) {
       return this;
     }
 
+    public StoreConfigBuilder<V> setMaxCapacity(int maxCapacity){
```
sudheeshkatkam: `) {`

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/InMemoryPersistentStore.java
```diff
@@ -15,28 +15,40 @@
...
+import com.google.common.collect.Iterables;
+
+public class InMemoryPersistentStore<V> extends BasePersistentStore<V> {
+  // private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(InMemoryPersistentStore.class);
+
   private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
   private final AutoCloseableLock readLock = new AutoCloseableLock(readWriteLock.readLock());
   private final AutoCloseableLock writeLock = new AutoCloseableLock(readWriteLock.writeLock());
-  private final ConcurrentMap<String, V> store = Maps.newConcurrentMap();
+  private ConcurrentMap<String, V> store;
```
sudheeshkatkam: final

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/InMemoryPersistentStore.java
```diff
@@ -15,28 +15,40 @@
...
   private int version = -1;
+  private int maxCapacity;
```
sudheeshkatkam: final, here and below

kkhatua:
Closing this PR in favor of [another](https://github.com/apache/drill/pull/834/)

PR #829, opened 2017-05-10
DRILL-5485: Remove WebServer dependency on DrillClient

sohami:


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -0,0 +1,190 @@
...
+
+import java.net.SocketAddress;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class WebUserConnection extends AbstractUserClientConnectionWrapper implements ConnectionThrottle {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(WebUserConnection.class);
+
+  /**
```
sudheeshkatkam: class level doc?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/auth/DrillUserPrincipal.java
```diff
@@ -18,43 +18,37 @@
...
 
-import java.io.IOException;
 import java.security.Principal;
 import java.util.List;
 
+
 /**
  * Captures Drill user credentials and resources in a session.
  */
 public class DrillUserPrincipal implements Principal, AutoCloseable {
```
sohami: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/work/prepare/PreparedStatementProvider.java
```diff
@@ -59,20 +43,29 @@
...
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
 
-import io.netty.buffer.ByteBuf;
-import io.netty.channel.ChannelFuture;
+import static org.apache.drill.exec.ExecConstants.CREATE_PREPARE_STATEMENT_TIMEOUT_MILLIS;
+import static org.apache.drill.exec.proto.UserProtos.RequestStatus.*;
```
sohami: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/AbstractUserClientConnectionWrapper.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.drill.exec.proto.GeneralRPCProtos;
+import org.apache.drill.exec.proto.UserBitShared.DrillPBError;
+import org.apache.drill.exec.proto.UserBitShared.QueryId;
+import org.apache.drill.exec.proto.UserBitShared.QueryResult;
+import org.apache.drill.exec.proto.helper.QueryIdHelper;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public abstract class AbstractUserClientConnectionWrapper implements UserClientConnection {
```
sohami: Renamed.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -219,12 +232,43 @@ public void sessionDestroyed(HttpSessionEvent se) {
...
+
+        // Clear all the custom attributes set as part of session
+        clearSessionCustomAttributes(session);
       }
     });
 
     return new SessionHandler(sessionManager);
   }
 
+  private void clearSessionCustomAttributes(HttpSession session) {
```
sohami: WebUserConnection is a class which is encapsulating all the resources. I had the logic of close and cleanup in it earlier. But I Didn't set that as attribute of Http session object since in that case it will be reset for each request. Since there are not many resources which we are setting inside HttpSession and each of them are really a separate property for the session in itself, I guess keeping them separated should be fine.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/AbstractUserClientConnectionWrapper.java
```diff
@@ -0,0 +1,101 @@
...
+        exception = new UserRemoteException(error);
+        latch.countDown();
+        break;
+      case CANCELED:
+      case COMPLETED:
+        Preconditions.checkState(result.getErrorCount() == 0);
+        latch.countDown();
+        break;
+      default:
+        logger.error("Query with QueryId: {} is in unexpected state: {}", queryId, state);
```
sudheeshkatkam: That maybe an issue as well.

AFAIK [DRILL-2498](https://github.com/apache/drill/commit/1d9d82b001810605e3f94ab3a5517dc0ed739715#diff-158c887d198393117d3a1bbc42114a8b) ensures that only the final state is sent to client using `sendResult`; this is the terminal message from server to client for that query. So if that message is wrong, the query is in an illegal state.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/UserClientConnection.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  UserSession getSession();
+
+  /**
+   * Send query result outcome to client. Outcome is returned through <code>listener</code>
+   *
+   * @param listener
+   * @param result
+   */
+  void sendResult(RpcOutcomeListener<GeneralRPCProtos.Ack> listener, UserBitShared.QueryResult result);
```
sudheeshkatkam: Not fixed?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
sudheeshkatkam: (I somehow managed to delete this in my comment..) The life cycle of those resources could be placed together in one class, as resources are being initialized in one place but closed in different places.

sohami: Moved all the resources inside another class `WebSessionResources` which now manages their lifecycle.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/AbstractUserClientConnectionWrapper.java
sohami: As discussed I have opened a new [DRILL-5558](https://issues.apache.org/jira/browse/DRILL-5558) for this to be addressed in a separated PR since it's existing behavior in DrillClient as well.

jinfengni:
Seems @sohami has addressed @sudheeshkatkam 's comments. I'm going to merge this PR.

+1

PR #830, opened 2017-05-11
DRILL-5498: Improve handling of CSV column headers

paul-rogers:
See DRILL-5498 for details.

Replaced the repeated varchar reader for reading columns with a purpose
built column parser. Implemented rules to recover from invalid column
headers.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
@@ -1,0 +1,274 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.store.easy.text.compliant;
+
+import java.nio.BufferOverflowException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import com.google.common.base.Charsets;
+
+/**
+ * Text output that implements a header reader/parser.
+ * The caller parses out the characters of each header;
+ * this class assembles UTF-8 bytes into Unicode characters,
+ * fixes invalid characters (those not legal for SQL symbols),
+ * and maps duplicate names to unique names.
+ * <p>
+ * That is, this class is as permissive as possible with file
+ * headers to avoid spurious query failures for trivial reasons.
+ */
+
+// Note: this class uses Java heap strings and the usual Java
+// convenience classes. Since we do heavy Unicode string operations,
+// and read a single row, there is no good reason to try to use
+// value vectors and direct memory for this task.
+
+public class HeaderBuilder extends TextOutput {
+
+  /**
+   * Maximum Drill symbol length, as enforced for headers.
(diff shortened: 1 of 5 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+
+    public HeaderError(int colIndex, String msg) {
+      super("Column " + (colIndex + 1) + ": " + msg);
+    }
+
+  }
+
+  public List<String> headers = new ArrayList<>();
+  public ByteBuffer currentField = ByteBuffer.allocate(MAX_HEADER_LEN);
+
```
paul-rogers: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+      }
+    }
+    return header;
+  }
+
+  /**
+   * Given an invalid header, rewrite it to replace illegal characters
+   * with valid ones. The header won't be what the user specified,
+   * but it will be a valid SQL identifier. This solution avoids failing
+   * queries due to corrupted of invalid header data.
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+   *
+   * @param header the original header
+   * @return the rewritten header, valid for SQL
+   */
+
+  private String rewriteHeader(String header) {
+    StringBuilder buf = new StringBuilder();
```
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+    } else {
+      if (Character.isDigit(ch)) {
+          buf.append("col_");
+          buf.append(ch);
+      } else {
+
+        // For the strange case of only one character, format
+        // the same as an empty header.
+
+        if (header.length() == 1) {
```
paul-rogers: Good catch. Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+          return "column_" + (headers.size() + 1);
```
paul-rogers: Fixed, created the constants. "col_" is used to replace a non-alpha leading character: $foo --> col_foo. "column_" is used for an anonymous column: "column_3".

The symbol limit is not really a hard limit since names are stored as strings. The extra few characters shouldn't be a problem. A related issue is that we limit field width to 1024 bytes. But, in UTF-8, up to four bytes make up a character, so the byte limit results in a lower character limit. Should not hit too many files in practice.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/HeaderBuilder.java
```diff
@@ -0,0 +1,249 @@
...
+
+  @Override
+  public void finishRecord() {
+    if (headers.isEmpty()) {
+      throw new HeaderError("The file must define at least one header.");
+    }
+
+    // Force headers to be unique.
+
+    Set<String> idents = new HashSet<String>();
```
paul-rogers: Fixed.

pom.xml
```diff
@@ -309,8 +309,8 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
+          <target>1.8</target>
```
sohami: is this change needed ? It's failing on my setup. This will use JDK 1.8 but my environment is setup for 1.7

paul-rogers: Argh... Checked in a private change. Mockito won't run tests unless compiled with JDK 7. Eclipse won't run tests unless I recompile for JDK 8. Reverted.

sohami:
+1. LGTM

sudheeshkatkam:
+1

Please squash the commits.

paul-rogers:
Commits squashed.

PR #831, opened 2017-05-11
DRILL-5432: Added pcap-format support

Vlad-Storona:
See DRILL-5432 for details.


exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Packet.java
```diff
@@ -0,0 +1,371 @@
...
+      if (isRoutingV6) {
+        srcPos = src ? PacketConstants.IP6_SRC_OFFSET + 96 : PacketConstants.IP6_DST_OFFSET + 96;
+      } else {
+        srcPos = src ? PacketConstants.IP6_SRC_OFFSET : PacketConstants.IP6_DST_OFFSET;
+      }
+    } else if (isPPPoV6Packet()) {
+      ipBuffer = new byte[16];
+      srcPos = src ? PacketConstants.IP6_SRC_OFFSET + PacketConstants.PPPoV6_IP_OFFSET : PacketConstants.IP6_DST_OFFSET + PacketConstants.PPPoV6_IP_OFFSET;
+    } else {
+      return null;
```
parthchandra: What does a null value returned from here mean? An unsupported packet type? If so, caller should throw an UnsupportedXYZ exception, perhaps?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,295 @@
...
+          if (packet.isTcpPacket()) {
+            setIntegerColumnValue(packet.getSequenceNumber(), pci);
+          }
+          break;
+        case "packet_length":
+          setIntegerColumnValue(packet.getPacketLength(), pci);
+          break;
+        case "data":
+          if (packet.getData() != null) {
+            setStringColumnValue(parseBytesToASCII(packet.getData()), pci);
```
parthchandra: We want to print this in ASCII? Is the packet data always printable as ASCII characters?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Murmur128.java
```diff
@@ -0,0 +1,161 @@
+/*
```
parthchandra: We already have a Murmur Hash implementation for use with Direct buffers. (see `org.apache.drill.exec.expr.fn.impl.MurmurHash3` ). Can we merge these two and have only one copy?

parthchandra:
It seems to me that others (Ted, Charles) also contributed to this. Would be nice if the PR acknowledged their contribution.
Also, we should document limitations and what is not (yet) supported. 
Perhaps a README?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Murmur128.java
tdunning: I tried to merge them, but the assumptions of working on unsafe code were very difficult to remove. I also had a difficult time figuring out a useful API.

Happy to look at specific suggestions, but the way that the existing implementation always iterates through unsafe pointers meant that there is likely to be nearly no shared code. Merging by simply moving my methods to the existing implementation is certainly doable.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
tdunning: No. I don't want to always do this. I want the ASCII version of the packet to be an optional field that you can pull, likewise with the binary version.

exec/java-exec/src/test/java/org/apache/drill/exec/store/pcap/TestPcapDecoder.java
```diff
@@ -0,0 +1,230 @@
...
+      for (int i = 0; i < 1000e6 / (29208 - 24) + 1; i++) {
+        // might be faster to keep this open and rewind each time, but
+        // that is hard to do with a resource, especially if it comes
+        // from the class path instead of files.
+        try (InputStream in = Resources.getResource("store/pcap/tcp-2.pcap").openStream()) {
+          ConcatPcap.copy(first, in, out);
+        }
+        first = false;
+      }
+      System.out.printf("Created file is %.1f MB\n", bigFile.length() / 1e6);
```
tdunning: Sure.  This good and easy to delete.

tdunning: I changed those methods to be called from a public static void main(). That allows them to be used to get information about speeds, but doesn't include their output in the test.

I think that addresses this comment.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Packet.java
```diff
@@ -0,0 +1,371 @@
...
+  public boolean isArpPacket() {
+    return protocol == PacketConstants.ARP_PROTOCOL;
+  }
+
+  @SuppressWarnings("WeakerAccess")
+  public boolean isIcmpPacket() {
+    return protocol == PacketConstants.ICMP_PROTOCOL;
+  }
+
+  public long getSessionHash() {
```
tdunning: This is very useful. It is the way that TCP sessions are reassembled (which is a big deal).

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Packet.java
tdunning: I don't think that an exception is warranted. There may be other packet types than IP.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
tdunning: WIth lazy decoding, there is no cost to having multiple versions of fields.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Packet.java
parthchandra: If there are unsupported types we will end up with a null value in the output. Is that what is desired? I tried with a pcap file from here: http://www.netresec.com/?page=MACCDC and verified that is so. That would be one of the reasons I feel we should at least document what might not be supported? Your call on this one. I am not an expert on the matter.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/dto/ColumnDto.java
```diff
@@ -0,0 +1,63 @@
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    ColumnDto columnDto = (ColumnDto) o;
+    return Objects.equals(columnName, columnDto.columnName) &&
```
parthchandra: OK, my mistake. The code is `Objects.equals` and not `Object.equals` so is in fact correct.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Murmur128.java
parthchandra: I was thinking of simply moving the code to the existing implementation.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,295 @@
...
+    try {
+
+      final Class<? extends ValueVector> clazz = TypeHelper.getValueVectorClass(
+          minorType, majorType.getMode());
+      ValueVector vector = output.addField(field, clazz);
+      vector.allocateNew();
+      return vector;
+
+    } catch (SchemaChangeException sce) {
+      throw new NullPointerException("The addition of this field is incompatible with this OutputMutator's capabilities");
```
paul-rogers: `NullPointerException` --> `IllegalStateException`

This is not really a null pointer, but it is an illegal state...

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,295 @@
...
+  @Override
+  public void setup(final OperatorContext context, final OutputMutator output) throws ExecutionSetupException {
+    this.output = output;
+  }
+
+  @Override
+  public int next() {
+    projectedCols = getProjectedColsIfItNull();
+    try {
+      return parsePcapFilesAndPutItToTable();
```
paul-rogers: Drill has certain protocols that are not entirely obvious, but that are needed here. Each call to `next()` produces a batch of rows. Drill requires that the batch contain no more than 64K rows. (Bad things happen if the batch has more.) So, the trick for a record reader is to read up to some maximum number of records into each batch, return, and wait to be called again for the next batch.

Depending on the size of the data, 64K may be to large and 8K or some other number may be better. Try to avoid causing any vector to be larger than 16 MB. (We're working on a project to enforce this, and the result will make it much easier to write a record reader. Too bad your project is just a bit too early to benefit.)

If I read this code correctly, each call to `next()` returns just one record, until EOF at which time it returns 0.

You'll find that the rest of Drill will be *very* slow with a batch size of 1 record. So, perhaps put this into a loop to read 8K or so rows into a single batch, then return the total record count. You should see a nice execution speed-up.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,295 @@
...
+    } catch (IOException io) {
+      throw new RuntimeException("Trouble with reading packets in file!");
```
paul-rogers: Drill's error message system can use work, but you can help. Rather than throw a generic Java exception, please throw a so-called `UserException`:

```
    throw UserException.dataReadError(io)
        .addContext("Trouble with reading packets in file!")
        .build();
```

Even better:

```
        .addContext("File name:", inputPath)
```

Where `inputPath` is the file name passed to the ctor that you save, and the above line is added to those above (you can have multiple `addContext()` calls.)

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,295 @@
...
+      final PcapTypes type = column.getColumnType();
+      TypeProtos.MinorType minorType = TYPES.get(type);
+
+      ProjectedColumnInfo pci = getProjectedColumnInfo(column, name, minorType);
+      pciBuilder.add(pci);
+    }
+    return pciBuilder.build();
+  }
+
+  private ProjectedColumnInfo getProjectedColumnInfo(final ColumnDto column,
```
paul-rogers: This part of reading is a bit complex. Again, we are working in simplifying it, but that work is not available yet.

You need to handle two cases:

```
SELECT * ...
SELECT `type`, `dst_ip`, ...
```

The first says to select all available columns, the other provides an explicit list of columns to select. The code earlier said that this plugin supports push-down, (that is, `supportsPushDown` returned true) so you've got to handle projection for both cases.

The code handle the second type: a list of columns. It does not seem to handle `SELECT *`, however.

I noticed, however, that the test case does include `SELECT *`. Maybe a comment or two about how that case is handled?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,295 @@
...
+          }
+          break;
+      }
+    }
+    return true;
+  }
+
+  private void setLongColumnValue(long data, ProjectedColumnInfo pci) {
+    ((NullableBigIntVector.Mutator) pci.vv.getMutator())
+        .setSafe(0, data);
```
paul-rogers: This works only for a single row. As noted above, you want to put more than one row per batch. So, you'll need to keep track of the row index and pass that as the first argument rather than 0.

Again, soon all this will be handled for you, you'll just be able to say:
```
    writer.column("dst_ip").setString(value);
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/Utils.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package org.apache.drill.exec.store.pcap;
+
+import com.google.common.primitives.Ints;
+import com.google.common.primitives.Shorts;
+
+public class Utils {
+
+  public static int getIntFileOrder(boolean byteOrder, final byte[] buf, final int offset) {
+    if (byteOrder) {
```
paul-rogers: Maybe an explanation of mapping byte order to booleans? true/false = which/which endian?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Packet.java
```diff
@@ -0,0 +1,371 @@
...
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+import static org.apache.drill.exec.store.pcap.Utils.convertInt;
+import static org.apache.drill.exec.store.pcap.Utils.convertShort;
+import static org.apache.drill.exec.store.pcap.Utils.getByte;
+import static org.apache.drill.exec.store.pcap.Utils.getIntFileOrder;
+import static org.apache.drill.exec.store.pcap.Utils.getShort;
+
+public class Packet {
```
paul-rogers: Would it have been possible to use one of the existing pcap Java libraries here? Four are listed [here](https://en.wikipedia.org/wiki/Pcap).

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,295 @@
...
+  private PacketDecoder getPacketDecoder() {
+    try {
+      return new PacketDecoder(in);
+    } catch (IOException io) {
+      throw new RuntimeException("File Not Found or some I/O issue");
+    }
+  }
+
+  private ImmutableList<ProjectedColumnInfo> getProjectedColsIfItNull() {
+    return projectedCols != null ? projectedCols : initCols(new Schema());
```
paul-rogers: Can this just be done in setup rather than checking the pointer on every call to `next()`?

parthchandra:
I would recommend moving this format plugin to the contrib directory, since this is a new and untested implementation.

tdunning:
What is the effective difference between the contrib director and where the
plugin already is?

What sort of testing do you think is necessary?

parthchandra:
The contrib directory is where we have, in the past, added new storage and format plugins that are new and may not have been sufficiently tested.
For this plugin, I think testing with pcap files from different sources would be useful. [1,2] are useful sources for data that will test boundary conditions. I tried on a file from [2] and got an NPE (didn't investigate the cause). A random sample of files from [1] worked very nicely indeed, though I didn't validate the output.
You might have already done this level of testing; if so, I will withdraw the suggestion.

[1] https://wiki.wireshark.org/SampleCaptures#Captures_used_in_Wireshark_testing
[2] http://www.netresec.com/?page=PcapFiles

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  private ImmutableList<ProjectedColumnInfo> initCols(final Schema schema) {
+    ImmutableList.Builder<ProjectedColumnInfo> pciBuilder = ImmutableList.builder();
+    ColumnDto column;
+
+    for (int i = 0; i < schema.getNumberOfColumns(); i++) {
+      column = schema.getColumnByIndex(i);
+
+      final String name = column.getColumnName();
```
paul-rogers: The column name passed in from Drill will have the case that the user typed. For example:

```
SELECT DST_IP ...
SELECT DsT_iP ...
SELECT dst_ip ...
```

Drill is supposed to be case insensitive: all the above should match the same "dst_ip" column.

But, the code later code a case-insensitive switch. So, a simple solution here is to do:
```
final String name = column.getColumnName().toLowerCase();
```

paul-rogers:
Trying to figure out where we are on this PR. Looking at the review comments, some were addressed (thanks!), others are open. The question is: what is the resolution of the open issues? Anything to fix? Is the comment based on a misunderstanding? Is the requested work beyond the scope of this particular PR? Is there a disagreement?

Perhaps respond to each open comment saying something like "will fix", "won't fix", "beyond scope" or something so we know if we're ready to move ahead with this PR, or if we are waiting for a few more improvements.

Thanks!

parthchandra:
Can we move this to contrib? I think we should get this one into the next release even as the rough edges (if any) are being smoothed out.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/Packet.java
dmitriyHavrilovich: This java libraries uses JNI to work with existing libpicap c library. So they are just wrappers above. This plugin uses only java code to process pcap.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/Utils.java
dmitriyHavrilovich: Comment will be added here, explaining this case

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
dmitriyHavrilovich: @paul-rogers 

> The code handle the second type: a list of columns. It does not seem to handle SELECT *, however.
Why you say that this case is not working ? Or I missing something ? Both cases you listed are operational

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapDrillTable.java
```diff
@@ -0,0 +1,73 @@
+/*
```
dmitriyHavrilovich: I think a link to a JIRA can be added. All needed information about already listed there.

Vlad-Storona:
As a result of project transferring problems were found out. In java-exec package exists file bootstrap-storage-plugin.json from which drill takes information about supported files formats. But in contrib package, there is no such file. If to transfer pcap-reader to contrib package and not to remove information from bootstrap-storage-plugin.json about pcap format, then there will be JsonMappingException. And if you remove this information from the config file, then drill will can`t find pcap files. Maybe I have not enough info/experience about drill. Maybe you can provide any solution how to handle this ?

paul-rogers:
Regarding bootstrap-storage-plugin.json, this is a generic issue. Drill must provide a solution that allows each module to provide its own file, as is done for drill-module.conf. I did not file a JIRA for this issue, feel free to file one.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapDrillTable.java
paul-rogers: The JIRA link is also good. Once the code is into Drill, it will be hard for future users to locate the corresponding JIRA. But, if you place a comment into a file somewhere, then future developers can find the JIRA.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/PcapRecordReader.java
paul-rogers: I stand corrected. Couldn't find that functionality when reviewing; should have downloaded the code and tried it.

paul-rogers:
Filed DRILL-5618 to describe the `bootstrap-storage-plugin.json` issue.

There are one or two remaining open comments. Will those be addressed, or should we do a final review based on the code as it stands now?

dmitriyHavrilovich:
We will not move this to contrib because of issue. Request for package-info.java will be satisfied. Also comment in utils will be added. If I did not miss something, that is all ?

paul-rogers:
Looking at the big picture, Drill should allow specialized plugins such as this one to exist as independent projects. Users should be able to download the plugin jar, add it to Drill, and go.

As we've discussed, Drill has a bit of work before we get there. We can't hold up this work waiting for a better solution.

So, please fix the two minor issues you identified. The code will then be ready for a final quick review and approval.

Later, once Drill provides the correct framework, I'd suggest that this code move into a separate Github repo to be maintained by experts in pcap. Frankly, most Drill developers are familiar with query engines, not pcap (or other specialized formats.)

The same is true, for example, of the "indexr" and TSDB plugins which are (slowly) working their way through the review process.

Summary: please add the package-info file and the comments in utils. We can then give approval.

Can we do this by, say, July 10? If so, we can likely get this PR into 1.11, if the Release Manager agrees.

dmitriyHavrilovich:
This is really good. Will do this immediately.

tdunning:
On Fri, Jun 30, 2017 at 9:55 AM, Paul Rogers <notifications@github.com>
wrote:

> Later, once Drill provides the correct framework, I'd suggest that this
> code move into a separate Github repo to be maintained by experts in pcap.
> Frankly, most Drill developers are familiar with query engines, not pcap
> (or other specialized formats.)
>
>
That only works if Drill has an autoloading capability that allows storage
formats to be loaded and authenticated easily.

paul-rogers:
> That only works if Drill has an autoloading capability that allows storage
> formats to be loaded and authenticated easily.

Why? To get pcap in 1.11, one must install a new Drill version, which requires a restart. Assuming that pcap were a separate project, you'd install a new jar, and restart. Neither provides auto loading.

Not clear on the authentication issue. If a plugin were a separate project, wouldn't that project have a way of certifying its jars?

Understanding this will help us figure out how to handle the growing set of specialized storage plugins...

tdunning:
On Fri, Jun 30, 2017 at 5:34 PM, Paul Rogers <notifications@github.com>
wrote:

> That only works if Drill has an autoloading capability that allows storage
> formats to be loaded and authenticated easily.
>
> Why? To get pcap in 1.11, one must install a new Drill version, which
> requires a restart. Assuming that pcap were a separate project, you'd
> install a new jar, and restart. Neither provides auto loading.
>
No. But if I have 1.11 already and decide that I want pcap (or any similar
data format), it would be nice if I could do the equivalent of pip (for
python) or install.packages("...") (for R) or mvn test (for Java) and get
whatever cool capability I like.

This may be described as "just install a new jar" but the convenience level
is a proven Big Deal (tm). It would even be possible to leverage Maven
central to make it happen, but there needs to be convenience sugar around
the process to make it consumable.

> Not clear on the authentication issue. If a plugin were a separate
> project, wouldn't that project have a way of certifying its jars?
>
Dunno. Not some random project.

If Drill requires it, then yes.  And I am saying that Drill should require
resolution back to some level of trust.

dmitriyHavrilovich:
@paul-rogers , is anything else we can do for this PR ?

paul-rogers:
Looks like we're good to go. Will be working on pulling the PR into Apache master today and will post any issues that arise as I run the code through the detailed tests used in that process.

paul-rogers:
Squashed commits and committed to Apache master. Congratulations!

PR #832, opened 2017-05-11
DRILL-5504: Vector validator to diagnose offset vector issues

paul-rogers:
Validates offset vectors in VarChar and repeated vectors. Validates the
special case of repeated VarChar vectors (two layers of offsets.)

Provides two new session variables to turn on validation. One enables
the existing operator (iterator) validation, the other adds vector
validation. This allows validation to occur in a “production” Drill
(without restarting Drill with assertions, as previously required.)

Unit tests validate the validator. Another test validates the
integration, but requires manual steps, so is ignored by default.

This version is first-cut: all work is done within a single class.
Allows back-porting to an earlier version to solve a specific issues. A
revision should move some of the work into generated code (or refactor
vectors to allow outside access), since offset vectors appear for each
subclass; not on a base class that would allow generic operations.


sudheeshkatkam:
+1

Please squash the commits.

paul-rogers:
Commits squashed.

paul-rogers:
Fixed typo in log message and rebased onto latest master.

PR #834, opened 2017-05-14
DRILL-5481: Allow to persist profiles in-memory only with a max capacity

kkhatua:
1. Introduced an InMemoryStoreProvider with the ability to maintain a max capacity
2. DrillbitContext now explicitly has a profileStoreProvider that, by default, re-uses the general PersistentStoreProvider, unless it is InMemory, which is when #1 is used.
2. Cleanly separated out QueryProfileStoreContext
3. Converted literal values to constants within ExecConstants
4. Updated drill-module.conf for default capacity


kkhatua:
@sudheeshkatkam , @ppadma 
Smaller commit that addresses the in-memory profile store requirement. This one does not require me to change any existing functionality or tests. I'll close the other [PR](https://github.com/apache/drill/pull/827)

PR #836, opened 2017-05-15
DRILL-5511: Additional UserException categories

paul-rogers:
Adds three kinds of internal errors:
1. General execution exception: something failed due to issues outside
Drill.
2. Internal error: something failed due to a likely code error.
3. Unspecified: something failed, but we don’t know why.


paul-rogers:
Withdrawn based on Parth's comments.

PR #839, opened 2017-05-16
DRILL-5516: Limit memory usage for Hbase reader

arina-ielchiieva:
To limit memory usage for Hbase reader we are adding max allowed allocated memory contant which will default to 64 mb. Thus batch size will be limited to 4000 (as before if memory limit does not exceed) or to number of records that are within max allowed memory limit. If first row in batch is larger than allowed default, it will be written in batch but batch will contain only this row.

### DIFF
--- a/contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
+++ b/contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -59,6 +59,10 @@ import com.google.common.collect.Sets;
 public class HBaseRecordReader extends AbstractRecordReader implements DrillHBaseConstants {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(HBaseRecordReader.class);
 
+  // batch should not exceed this value to avoid OOM on a busy system
+  private static final int MAX_ALLOCATED_MEMORY_PER_BATCH = 64 * 1024 * 1024; // 64 mb in bytes
+
+  // batch size should not exceed max allowed record count
   private static final int TARGET_RECORD_COUNT = 4000;
 
   private OutputMutator outputMutator;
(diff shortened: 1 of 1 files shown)


paul-rogers:
The right approach is not to simply allow HBase to use more memory. The right approach is to limit memory.

Fortunately, another project is underway to do just that. Let's collaborate. In the next week or so I'll do a PR for the framework to limit batch sizes in readers, along with an implementation for the "compliant" text readers.

Maybe you can use that framework to retrofit the HBase reader to also limit it's batch size. Basically, we limit the length of the longest vector to 16 MB.

The present patch, using unlimited memory, has all kinds of other problems -- the very problems we are trying to solve, so it is not helpful to move forward in one area, backward in another.

arina-ielchiieva:
This approach does not allow Hbase to use more memory, actually it limits memory usage.. Previously when batch size was limited to 4000 rows, we could have one batch using ~3 GB.

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
```diff
@@ -187,8 +189,8 @@ public int next() {
     }
 
     int rowCount = 0;
-    done:
-    for (; rowCount < TARGET_RECORD_COUNT; rowCount++) {
+    // if first row is larger than allowed max size in batch, it will be added anyway
+    do {
```
paul-rogers: Still need to monitor row count: row count cannot exceed 64K. So, loop termination is either exceeds memory limit OR reaches max row count. For row count, might as well use the original limit, unless we know enough to pick a better limit.

paul-rogers:
General suggestion, perhaps change the title to more clearly describe the fix. Maybe "Limit memory usage for Hive reader" or some such. I originally read "use max allowable memory" as perhaps meaning to use the full 10 GB that the allocator gives to each operator...

paul-rogers:
One more comment: any way to unit test this improvement?

arina-ielchiieva:
@paul-rogers 
1. Renamed PR as you suggested to better convey changes idea. Thank you for suggestion.
2. Included row count into batch size determination.  Thus batch size will be limited to max allowed rows (as before if memory limit does not exceed) or to number of records that are within max allowed memory limit.
3. Unfortunately did not add unit tests, Not sure if there is a way to test Hbase reader in isolation for true unit testing (I know you have been working to some changes in this area but I don't know the details).
But I have tested the fix manually, changing in code max rows size and memory allocation. Could not write Drill "unit tests" for this since it's not easy to mock `static final int fields` and adding such large data sets to test on real numbers may extend our unit tests running time. I guess in this case it's better to add such test to Functional tests suite. So if possible let's leave this change without unit tests for now.

sudheeshkatkam:
+1

PR #840, opened 2017-05-16
DRILL-5517: Size-aware set methods in value vectors

paul-rogers:
Please see DRILL-5517 for an explanation.

### DIFF
--- a/exec/vector/src/main/codegen/templates/FixedValueVectors.java
+++ b/exec/vector/src/main/codegen/templates/FixedValueVectors.java
@@ -157,7 +157,7 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements F
    * Note that the maximum number of values a vector can allocate is Integer.MAX_VALUE / value width.
    *
    * @param valueCount
-   * @throws org.apache.drill.exec.memory.OutOfMemoryException if it can't allocate the new buffer
+   * @throws OutOfMemoryException if it can't allocate the new buffer
    */
   @Override
   public void allocateNew(final int valueCount) {
@@ -637,7 +637,8 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements F
      * Set the value of a required or nullable vector. Enforces the value
      * and size limits.
      * @param index item to write
-     * @return true if the item was written, false if the index would
+     * @param value value to set
+     * @throws VectorOverflowException if the item was written, false if the index would
      * overfill the vector
      */
 
@@ -651,7 +652,8 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements F
     /**
      * Set the value of a repeated vector. Enforces only the size limit.
      * @param index item to write
-     * @return true if the item was written, false if the index would
+     * @param value value to set
+     * @throws VectorOverflowException if the item was written, false if the index would
      * overfill the vector
      */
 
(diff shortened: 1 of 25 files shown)


paul-rogers:
Rebased on master. Resolved merge conflict.

paul-rogers:
Reran pre-commit tests. One unit test fails, but in an area untouched by this commit. Indeed, this appears to be a random failure as the very same tests passes on my Mac and on my Linux development VM. See DRILL-5612.

exec/vector/src/main/codegen/templates/FixedValueVectors.java
```diff
@@ -546,238 +576,400 @@ public DateTime getObject(int index) {
...
-    public void set(int index, int months, int days, int milliseconds){
-      final int offsetIndex = index * ${type.width};
-      data.setInt(offsetIndex, months);
-      data.setInt((offsetIndex + ${minor.daysOffset}), days);
-      data.setInt((offsetIndex + ${minor.millisecondsOffset}), milliseconds);
+    /**
+     * Set the value of a required or nullable vector. Enforces the value
+     * and size limits.
+     * @param index item to write
+     * @return true if the item was written, false if the index would
```
amansinha100: Method returns a void but comment says the method returns a boolean..

exec/vector/src/main/java/org/apache/drill/exec/vector/VectorUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+public class VectorUtils {
+
+  /**
+   * Vectors cannot be any larger than the Netty memory allocation
+   * block size.
+   */
+
+  private static final int ABSOLUTE_MAX_SIZE = 16 * 1024 * 1024;
+  private static final int ABSOLUTE_MIN_SIZE = 16 * 1024;
```
amansinha100: It wasn't clear why the absolute_min_size is 16K .. that's not related to Netty I presume.  Couldn't it be much smaller ?

paul-rogers: This code allows a boot-time option to limit maximum vector size. I thought it would be handy for testing, but turned out not to be necessary, so I might remove it.

To prevent failures when the value is set wrong, I clamped the value to some minimum and maximum. The maximum is set by the Netty block size. The minimum I just made up: seems if we limit vector size to much smaller than 16 K we'd be unlikely to make progress.

Added a comment to clarify.

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -674,6 +764,14 @@ public void reset(){
       setCount = 0;
       <#if type.major = "VarLen">lastSet = -1;</#if>
     }
+
+    @Override
+    public void exchange(ValueVector.Mutator other) {
```
amansinha100: ok, thanks for the explanation.

exec/vector/src/main/codegen/templates/FixedValueVectors.java
amansinha100: What about this ?  The return type in Javadoc is not consistent with the return type here and a few other places in this file.

amansinha100:
Minor followup comment.   LGTM.  +1

exec/vector/src/main/codegen/templates/FixedValueVectors.java
paul-rogers: Fixed here and several other places.

PR #842, opened 2017-05-18
DRILL-5523: Revert if condition in UnionAllRecordBatch changed in DRI…

arina-ielchiieva:
…LL-5419

If condition in UnionAllRecordBatch which was changed to correct one in DRILL-5419 but this change caused the regression (DRILL-5521).
The root cause is not in change itself but rather in missing indication of child column in map types when it is null. There is a big project upcoming re-working json reader implementation, including map types (DRILL-4824) which will fix the problem.
So far reverting change to initial state so users won't see the regression till DRILL-4824 is merged.


arina-ielchiieva:
@vvysotskyi since you are working on DRILL-4824, please review pull request and confirm that issue will be fixed after DRILL-4824 is merged.

amansinha100:
+1

PR #843, opened 2017-05-23
DRILL-5533: Fix flag assignment in FunctionInitializer.checkInit() method

arina-ielchiieva:
Changes:
1. Fixed DCL in FunctionInitializer.checkInit() method (update flag parameter when function body is loaded).
2. Fixed ImportGrabber.getImports() method to return the list with imports.
3. Added unit tests for FunctionInitializer.
4. Minor refactoring (renamed methods, added javadoc).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -31,21 +31,18 @@ import org.codehaus.janino.Parser;
 import org.codehaus.janino.Scanner;
 import org.mortbay.util.IO;
 
-import com.google.common.collect.Maps;
-
 /**
  * To avoid the cost of initializing all functions up front,
- * this class contains all informations required to initializing a function when it is used.
+ * this class contains all information required to initializing a function when it is used.
  */
 public class FunctionInitializer {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FunctionInitializer.class);
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FunctionInitializer.class);
 
   private final String className;
   private final ClassLoader classLoader;
-  private Map<String, CompilationUnit> functionUnits = Maps.newHashMap();
   private Map<String, String> methods;
   private List<String> imports;
-  private volatile boolean ready;
+  private volatile boolean isLoaded;
 
   /**
    * @param className the fully qualified name of the class implementing the function
@@ -53,7 +50,6 @@ public class FunctionInitializer {
    *                    to prevent classpath collisions during loading an unloading jars
    */
   public FunctionInitializer(String className, ClassLoader classLoader) {
-    super();
     this.className = className;
     this.classLoader = classLoader;
   }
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
```diff
@@ -119,33 +117,35 @@ private void checkInit() {
...
-    String path = c.getName();
+  /**
+   * Using class name generates path to class source code (*.java),
+   * reads its content as string and parses it into {@link org.codehaus.janino.Java.CompilationUnit}.
+   *
+   * @param clazz function class
+   * @return compilation unit
+   * @throws IOException if did not find class or could not load it
+   */
+  public CompilationUnit convertToCompilationUnit(Class<?> clazz) throws IOException {
```
sudheeshkatkam: private?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
```diff
@@ -31,29 +31,25 @@
...
- * this class contains all informations required to initializing a function when it is used.
+ * this class contains all information required to initializing a function when it is used.
  */
 public class FunctionInitializer {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FunctionInitializer.class);
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FunctionInitializer.class);
 
   private final String className;
   private final ClassLoader classLoader;
-  private Map<String, CompilationUnit> functionUnits = Maps.newHashMap();
```
sudheeshkatkam: Is this map unused?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
```diff
@@ -74,41 +70,43 @@ public String getClassName() {
...
 
     synchronized (this) {
-      if (ready) {
+      if (isLoaded) {
         return;
       }
 
-      // get function body.
-
+      logger.debug("Getting function body for the {}", className);
```
sudheeshkatkam: trace

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
```diff
@@ -119,33 +117,35 @@ private void checkInit() {
...
     path = path.replaceFirst("\\$.*", "");
     path = path.replace(".", FileUtils.separator);
     path = "/" + path + ".java";
-    CompilationUnit cu = functionUnits.get(path);
-    if (cu != null) {
-      return cu;
-    }
 
-    try (InputStream is = c.getResourceAsStream(path)) {
+    logger.debug("Loading function code from the {}", path);
```
sudheeshkatkam: trace

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/ImportGrabber.java
```diff
@@ -63,9 +62,21 @@ public void traverseStaticImportOnDemandDeclaration(StaticImportOnDemandDeclarat
...
-    cu.getPackageMemberTypeDeclarations()[0].accept(visitor.finder.comprehensiveVisitor());
+  /**
+   * Creates list of imports that are present in compilation unit.
+   * For example:
+   * [import io.netty.buffer.DrillBuf;, import org.apache.drill.exec.expr.DrillSimpleFunc;]
+   *
+   * @param compilationUnit compilation unit
+   * @return list of imports
+   */
+  public static List<String> getImports(Java.CompilationUnit compilationUnit){
```
sudheeshkatkam: If the returned list was empty previously, what was the functionality previously?

And looks to me like `FunctionInitializer#imports` is unused. So should `ImportGrabber` be removed?

arina-ielchiieva: When writing code for UDFs, we recommend users to use the fully qualified class names instead of imports.
What if we enhance our UDFs generation code to use imports from function class? This way, fully qualified class names usage won't be required. Since I thought it could be used for future enhancement, I have left `ImportGrabber`.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
arina-ielchiieva: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionInitializer.java
arina-ielchiieva: It was used in `convertToCompilationUnit`. Before loading function body we checked if it was loaded before in `functionUnits` map.  If not - function body was loaded, if yes - previously loaded compilation unit was returned. With incorrectly implemented DCL such check prevented us from loading function body from disk each time, through threads were entering synchronized block anyway. With correctly implemented DCL we don't need to store this intermediate result. First thread will load all information about methods and imports, others will use this information.

arina-ielchiieva:
@sudheeshkatkam addressed all CR comments, please review. Thanks in advance!

sudheeshkatkam:
+1

PR #844, opened 2017-05-25
DRILL-5538: Create TopProject with validatedNodeType after PHYSICAL phase

arina-ielchiieva:
Details in DRILL-5538 description.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/TopProjectVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/TopProjectVisitor.java
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.planner.physical.visitor;
+
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.Project;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
+import org.apache.drill.exec.planner.common.DrillRelOptUtil;
+import org.apache.drill.exec.planner.physical.Prel;
+import org.apache.drill.exec.planner.physical.ProjectPrel;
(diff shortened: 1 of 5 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
```diff
@@ -180,9 +180,19 @@ public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin>
 
   PHYSICAL("Physical Planning") {
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
+      final RuleSet storageRules = getStorageRules(context, plugins, this);
+
+      final Builder<RelOptRule> newStorageRulesBuilder = ImmutableSet.builder();
+      for (final RelOptRule relOptRule : storageRules) {
+        // exclude ProjectRemoveRule to prevent Drill from losing column alias or displaying implicit columns
+        if (!(relOptRule instanceof ProjectRemoveRule)) {
```
jinfengni: For query  "select version as current_version from t;", why will ProjectRemoveRule think that the Project operator is a trivial project? I assume that the input/output's field name is different.

If there is bug in ProjectRemoveRule, it's better to fix that, in stead of removing it completely.

arina-ielchiieva: Actually, you are right there is a bug in ProjectRemoveRule in Calcite [1]. Jira was open at the beginning of this year: ProjectRemoveRule loses field names [1].

[1] https://issues.apache.org/jira/browse/CALCITE-1584

arina-ielchiieva:
@jinfengni updated to PR to create TopProject with validatedNodeType after PHYSICAL phase instead of removing ProjectRemoveRule completely.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/TopProjectVisitor.java
```diff
@@ -0,0 +1,139 @@
...
+import org.apache.drill.exec.planner.physical.ScreenPrel;
+import org.apache.drill.exec.planner.physical.WriterPrel;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Adds non-trivial top project to ensure the final output field names are preserved.
+ * Is added under the screen or writer operators.
```
jinfengni: You may add comments why we need such non-trivial project, due to Calcite's behavior of ProjectRemoveRule. Also, it might be worth mentioning that the added Project may not be sitting right under Screen/Writer operator in the physical plan, as we would add additional Projects, for case likes * column expansion, or partition by column processing.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -459,7 +468,13 @@ protected Prel convertToPrel(RelNode drel) throws RelConversionException, SqlUns
     /* The order of the following transformations is important */
 
     /*
-     * 0.) For select * from join query, we need insert project on top of scan and a top project just
+     * 0.)
+     * Add top project before screen operator or writer to ensure that final output column names are preserved.
+     */
+    phyRelNode = TopProjectVisitor.insertTopProject(phyRelNode, validatedRowType);
```
jinfengni: Make sense to me that you put this Visitor handling first, after physical transform() call, before these existing visitor handling.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/TopProjectVisitor.java
arina-ielchiieva: Added comments and put ready-to-commit label.

PR #845, opened 2017-05-25
DRILL-5537: Display columns alias for queries with sum() when RDBMS s…

arina-ielchiieva:
…torage plugin is enabled

For sum() queries DrillConvertSumToSumZero rule is applied. But during converting to new aggregate call, this call was created with name set to null, therefore column alias was lost when RDBMS storage plugin was enabled. RDBMS storage plugin was adding new rule during PHYSICAL phase - ReduceProjectRule. Since Project stage was removed alias was lost.

Changes:
1. Added old call aggregate name during new call aggregate creation in DrillConvertSumToSumZero  rule.
2. Replaced deprecated AggregateCall constructor to `AggregateCall.create`.
3. Minor refactoring.


jinfengni:
+1

LGTM.

PR #846, opened 2017-05-26
DRILL-5544: Out of heap running CTAS against text delimited

vdiravka:
- Since parquet version of PageWriter cann't allow to use direct memory for allocating ByteBuffers,
  this PR introduces other version of PageWriter and PageWriteStore. 
  See more: https://issues.apache.org/jira/browse/PARQUET-1006


exec/java-exec/src/main/java/org/apache/parquet/hadoop/ParquetColumnChunkPageWriteStore.java
```diff
@@ -0,0 +1,269 @@
+/*
```
paul-rogers: Compared this file to the Parquet original (thanks for providing the file name) using [this tool](https://www.diffchecker.com/diff), referencing the diagram [here](https://parquet.apache.org/documentation/latest/). The changes made in the Drill copy seem reasonable.

I do have questions, however, about the approach to writing. (See below.) Seems overly memory intensive. But, this is an issue with the Parquet original, not about this "port" of the file.

exec/java-exec/src/main/java/org/apache/parquet/hadoop/ParquetColumnChunkPageWriteStore.java
```diff
@@ -0,0 +1,269 @@
...
+
+    @Override
+    public void writePage(BytesInput bytes,
+                          int valueCount,
+                          Statistics statistics,
+                          Encoding rlEncoding,
+                          Encoding dlEncoding,
+                          Encoding valuesEncoding) throws IOException {
+      long uncompressedSize = bytes.size();
+      if (uncompressedSize > Integer.MAX_VALUE) {
```
paul-rogers: If the uncompressed size is anywhere near `Integer.MAX_VALUE` (2 GB), then we have a resource problem. Double that for the compressed version and we're talking about serious memory usage. And, this is per column.

What is needed to ensure that each page (see [this diagram](https://parquet.apache.org/documentation/latest/)) is kept to a reasonable size? Does Drill already chose a reasonable size? Is it dependent on the (widely varying) input batch size?

paul-rogers:
Another critical issue here. Direct memory is not a limitless resource, unfortunately. Allocating memory larger than 16 MB causes memory fragmentation as the allocation must come from the system, but all free memory is cached in Netty in 16 MB chunks. So, if the buffer size here is larger than 16 MB, the operation may fail with an OOM.

If this operation must buffer in memory, and must use larger than 16 MB, then you need something like what the original code provided. That buffered writer can be backed by a chain of 16 MB direct memory blocks.

exec/java-exec/src/main/java/org/apache/parquet/hadoop/ParquetColumnChunkPageWriteStore.java
```diff
@@ -0,0 +1,269 @@
...
+import org.apache.parquet.hadoop.CodecFactory.BytesCompressor;
+import org.apache.parquet.io.ParquetEncodingException;
+import org.apache.parquet.schema.MessageType;
+import org.apache.parquet.bytes.ByteBufferAllocator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This is a copy of ColumnChunkPageWriteStore from parquet library except of OutputStream that is used here.
+ * Using of CapacityByteArrayOutputStream allows to use different ByteBuffer allocators.
```
paul-rogers: I'm not sure that 20 GB is "OK" to write to a file: whether on the heap or otherwise. But, if we accept the poor design that needs this much, then I agree it is better to have uncontrolled use of direct memory than the heap. But, in either case, uncontrolled memory use is a very bad idea.

Note that the only buffering needed is 4K: the disk block size. Experiments with spilling show that there is zero benefit to buffering above 32K.

So, for 10 runs * 32K = 320K. This is FAR less than the 20 GB proposed, and plenty for heap, to avoid the cost of copying data into, then out of, direct memory.

Further, what is wrong with Java's own `BufferedOutputStream`? Was this tried? It works well for simple cases, not sure about more complex cases.

Is the problem that Parquet must somehow buffer an entire page so it can "backpatch" the first bytes after writing all the rest? If so, that part of Parquet design is incompatible with a write-once file system such as HDFS: it forces unnecessary load on the memory system. (Note, however, than an updatable file system, such as Linux or MFS, would allow us to write data to disk, then go back and write the header, without buffering.)

vdiravka: I meant a special case regarding 20Gb. Usually 40 threads can be used in sufficiently powerful machines with enough RAM space. But we can control the memory usage by reducing number of threads per node to prevent OOM while buffering data.

512Mb is the default `store.parquet.block-size`, but according [Drill doc](https://drill.apache.org/docs/parquet-format/#configuring-the-size-of-parquet-files) it is recommended to set this value to the number of bytes less than or equal to the block size of MFS, HDFS, or the file system. Then only 320K will be buffered in memory. 

Nothing wrong with `BufferedOutputStream`. Original `ColumnChunkPageWriter` uses it, but together with `ConcatenatingByteArrayCollector` (which can buffer data only in `List<byte[]>`). I suppose it is possible to upgrade the last class to use `ByteBuffers`, but the result will be similar to `CapacityByteArrayOutputStream`.

As I understand, data buffering allows to write into the disk entire block without going back and writing anything else, ideally one row group per one fs block. Looks like it is suitable for HDFS as well.

paul-rogers:
Chatted with Parth who mentioned that Parquet page sizes are typically on the order of 1MB, maybe 8 MB, but 16 MB is too large.

The concern expressed in earlier comments was that if we buffer, say, 256 MB of data per file, and we're doing many parallel writes, we will use up too much memory.

But, if we buffer only one page at a time, and we control page size to be some amount on the order of 1-2 MB, then even with 100 threads, we're still using only 200 MB, say, which is fine.

In this case, the direct memory solution is fine. (But please check performance.)

However, if we are running out of memory, I wonder if we are not controlling page size and letting them get too large? Did you happen to check the size of the pages we are writing?

If the pages are too big, let's file another JIRA ticket to fix that problem so that we have a complete solution.

Once we confirm that we are writing small pages (or file that JIRA if not), I'll change my vote from +0 to +1.

vdiravka:
@paul-rogers As I mentioned in my previous comment the page size can't greatly exceed 1Mb (default value of page-size option in Drill). And I checked it -- almost every time the page size is much less than 1 MB.

The data, which are buffered - all pages within one row group. And when buffered data exceeds the block-size then the row group will be written to the disk and flushed from the stream buffer.
Which is what the current code does.

I compared of creating a large parquet tables with current Drill master version and version of Drill with my fix and received the same performance. Also I found the same time of passing the Drill's tests

The branch is rebased to the last Drill master version.

paul-rogers:
Given that the page sizes are small (~ 1MB), then the solution simply transfers a modest amount of memory from heap to direct. So, LGTM.

+1

vdiravka:
Commits are squashed into one.

PR #847, opened 2017-05-31
Drill 5545: Update POM to add support for running findbugs

parthchandra:
Also fix some issues reported by findbugs in the Async Parquet Reader


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -462,7 +431,10 @@ public Void call() throws IOException {
           readStatus.setDiskScanTime(timeToRead);
           assert (totalValuesRead <= totalValuesCount);
         }
-        synchronized (queue) {
+        // FindBugs reports this is a possible bug, but it is not. You do need the synchronized block
```
parthchandra: I'd intended to suppress the warning initially, but then got rid of it by using a different object to synchronize on. Fixed the comment to not refer to FindBugs though.

PR #849, opened 2017-06-05
DRILL-5568: Include hadoop-common jars inside drill-jdbc-all.jar

sohami:
More details on this PR is in [JIRA](https://issues.apache.org/jira/browse/DRILL-5568)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.rpc.security;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.CommonConfigurationKeys;
+
+
+public class SecurityConfiguration extends Configuration {
+  //private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SecurityConfiguration.class);
+
+  public SecurityConfiguration() {
+    super();
+    updateGroupMapping();
+  }
+
+  /**
+   * Update the Group Mapping class name to add namespace prefix retrieved from System Property. This is needed since
+   * in drill-jdbc-all jar we are packaging hadoop dependencies under that namespace. This will help application
+   * using this jar as driver to avoid conflict with it's own hadoop dependency if any.
+   */
+  private void updateGroupMapping() {
+    final String originalClassName = get(CommonConfigurationKeys.HADOOP_SECURITY_GROUP_MAPPING);
+    final String profilePrefix = System.getProperty("namespacePrefix");
+    final String updatedClassName = (profilePrefix != null) ? (profilePrefix + originalClassName)
+                                                            : originalClassName;
+    set(CommonConfigurationKeys.HADOOP_SECURITY_GROUP_MAPPING, updatedClassName);
+  }
+}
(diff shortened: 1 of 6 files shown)


sohami:
+ @paul-rogers - To help review this PR in case JIRA didn't sent out the email.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /**
+   * Update the Group Mapping class name to add namespace prefix retrieved from System Property. This is needed since
+   * in drill-jdbc-all jar we are packaging hadoop dependencies under that namespace. This will help application
+   * using this jar as driver to avoid conflict with it's own hadoop dependency if any.
+   */
+  private void updateGroupMapping() {
+    final String originalClassName = get(CommonConfigurationKeys.HADOOP_SECURITY_GROUP_MAPPING);
+    final String profilePrefix = System.getProperty("namespacePrefix");
```
paul-rogers: Would recommend a name that is a bit less generic. Maybe "drill.security.namespacePrefix"?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+    final String updatedClassName = (profilePrefix != null) ? (profilePrefix + originalClassName)
+                                                            : originalClassName;
+    set(CommonConfigurationKeys.HADOOP_SECURITY_GROUP_MAPPING, updatedClassName);
```
paul-rogers:
```
if (! Strings.isNullOrEmpty(profilePrefix)) {
  set(CommonConfigurationKeys.HADOOP_SECURITY_GROUP_MAPPING, profilePrefix.trim() + originalClassName);
}
```
?
Handles the case where the prefix is missing, or exists, but is empty. Or, perhaps I'm being overly cautions about stray spaces, empty properties, and not changing properties when not needed...

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+
+  public SecurityConfiguration() {
+    super();
+    updateGroupMapping();
+  }
+
+  /**
+   * Update the Group Mapping class name to add namespace prefix retrieved from System Property. This is needed since
+   * in drill-jdbc-all jar we are packaging hadoop dependencies under that namespace. This will help application
+   * using this jar as driver to avoid conflict with it's own hadoop dependency if any.
```
paul-rogers: Maybe add a comment that the property is needed only when Hadoop classes are relocated. In a normal build, the property is not set and Hadoop classes are used normally.

exec/jdbc-all/pom.xml
```diff
@@ -483,6 +518,269 @@
 
   <profiles>
     <profile>
+      <id>default</id>
```
paul-rogers: Not real happy about the amount of redundancy. But, I suppose Maven offers no alternative...

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillFactory.java
```diff
@@ -37,6 +38,24 @@
   protected final int major;
   protected final int minor;
 
+  static {
+    Properties prop = new Properties();
```
paul-rogers: Actually, I think this code can be moved into `SecurityConfiguration.java`.

Java handles resources by looking in the class path, which is formed from all jars. A long as the resource file is somewhere in the top level of some jar, this code will find it. So, despite what I said in person the other day, the code need not be here to work.

That said, the property file DOES have to be in the JDBC-all package to avoid having multiple files of the same name in the class path.

Please try it out to see if the code works in `SecurityConfiguration.java`.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
sohami: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
sohami: Updated

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/SecurityConfiguration.java
sohami: Changed to use` Strings.isNullOrEmpty`. I am already trimming the value of prefix befire setting it inside system property.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillFactory.java
sohami: Moved it to SecurityConfiguration.java. All unit tests are also passing and the application is also running fine.

exec/jdbc-all/pom.xml
sohami: Yes didn't found any other way to set the default value.

PR #850, opened 2017-06-05
DRILL-5541: C++ Client Crashes During Simple "Man in the Middle" Atta…

superbstreak:
…ck Test with Exploitable Write AV


parthchandra:
+1

PR #853, opened 2017-06-15
DRILL-5130: Implement DrillValuesRel and ValuePrel as Calcite Values sub-classes

arina-ielchiieva:
1. Factor out common logic for DrillValuesRel and ValuesPrel into DrillValuesRelBase.
2. Revisit explainTerms to write full tuples content so consequently Values nodes with the same row type and count but different values are considered to be different.
3. Minor refactoring.

Details in Jira DRILL-5130.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillValuesRelBase.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillValuesRelBase.java
@@ -1,0 +1,251 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to you under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.planner.common;
+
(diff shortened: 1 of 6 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillValuesRelBase.java
```diff
@@ -0,0 +1,44 @@
...
+
+
+import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.plan.RelTraitSet;
+import org.apache.calcite.rel.AbstractRelNode;
+import org.apache.calcite.rel.RelWriter;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.drill.common.JSONOptions;
+
+public abstract class DrillValuesRelBase extends AbstractRelNode {
```
arina-ielchiieva: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillValuesRelBase.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.calcite.plan.RelTraitSet;
+import org.apache.calcite.rel.AbstractRelNode;
+import org.apache.calcite.rel.RelWriter;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.drill.common.JSONOptions;
+
+/**
+ * Base class for logical and physical Values implemented in Drill.
+ */
+public abstract class DrillValuesRelBase extends AbstractRelNode {
```
jinfengni: Calcite has an abstract class Values. Neither of DrillValueRel or ValuePrel extends from that Calcite class. Will it help solve the problem if we extend from Calcite's class?

https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/rel/core/Values.java

arina-ielchiieva:
@jinfengni I have implemented DrillValuesRel and ValuesPrel as Calcite Values sub-classes. Also added second constructor to avoid unnecessary tuples parsing into json during copying or logical to physical values conversion. Please review the changes when possible.

arina-ielchiieva:
Merged into master with commit id 33682be99e719dce9cb326e2835ebc4ae434104a

PR #856, opened 2017-06-19

kkhatua:
We submit a timeout-managing task that sleeps for the stipulated period before trying to cancel the query (JDBC Statement). 
It might be worth having a similar feature as a System/Session variable so that the same can be achieved via SQLLine or REST APIs, but is beyond the scope of this JIRA's PR.


kkhatua:
Existing negative Unit tests are failing. Look like we never had proper unit tests. Will add them and update the PR.

PR #857, opened 2017-06-20
DRILL-5599: Notify StatusHandler that batch sending has failed even if channel is still open

arina-ielchiieva:
Details in [DRILL-5599](https://issues.apache.org/jira/browse/DRILL-5599)

### DIFF
--- a/exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
+++ b/exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -47,7 +47,7 @@ class RequestIdMap {
   private final IntObjectHashMap<RpcOutcome<?>> map;
 
   public RequestIdMap() {
-    map = new IntObjectHashMap<RpcOutcome<?>>();
+    map = new IntObjectHashMap<>();
   }
 
   void channelClosed(Throwable ex) {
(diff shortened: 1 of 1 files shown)


exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
```diff
@@ -111,13 +111,16 @@ public RpcListener(RpcOutcomeListener<T> handler, Class<T> clazz, int coordinati
...
-        if (future.channel().isActive()) {
-          throw new RpcException("Future failed");
-        } else {
-          setException(new ChannelClosedException());
+        try {
+          removeFromMap(coordinationId);
+        } finally {
+          final Throwable cause = future.cause();
+          if (future.channel().isActive()) {
+            setException(cause == null ? new RpcException("Future has failed") : cause);
```
ppadma: Can we pick a better message ?  "Future has failed" does not convey anything useful to the end user.

exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
```diff
@@ -47,7 +47,7 @@
   private final IntObjectHashMap<RpcOutcome<?>> map;
 
   public RequestIdMap() {
-    map = new IntObjectHashMap<RpcOutcome<?>>();
+    map = new IntObjectHashMap<>();
   }
```
ppadma: why do we need this change ?

exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
ppadma: overall, LGTM.  Some minor comments.

exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
paul-rogers: This is code cleanup. Java can infer the type of the parameterized class without us telling it. Seems a similar cleanup was made a few lines below.

exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
paul-rogers: I think the point is that the message is not sent to the end user, or written to the log. Rather, it is letting the listener know that the send failed. There is a reason for the failure. Ideally the `cause` is the real reason.

Still it would be good if we were to log such failures. So, maybe "Unknown channel failure" or some such might be helpful. Probably the developer will search the code to find where we raise that exception, so the message just has to be unique...

arina-ielchiieva:
Reworded error message. @paul-rogers and @ppadma thanks for code review!

PR #859, opened 2017-06-21
DRILL-2478: Validating values assigned to SYSTEM/SESSION configuratio…

ppadma:
…n parameters

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -166,7 +166,7 @@ public interface ExecConstants {
   String DEFAULT_TEMPORARY_WORKSPACE = "drill.exec.default_temporary_workspace";
 
   String OUTPUT_FORMAT_OPTION = "store.format";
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -166,7 +166,7 @@
   String DEFAULT_TEMPORARY_WORKSPACE = "drill.exec.default_temporary_workspace";
 
   String OUTPUT_FORMAT_OPTION = "store.format";
-  OptionValidator OUTPUT_FORMAT_VALIDATOR = new StringValidator(OUTPUT_FORMAT_OPTION, "parquet");
+  OptionValidator OUTPUT_FORMAT_VALIDATOR = new EnumeratedStringValidator(OUTPUT_FORMAT_OPTION, "parquet", "parquet", "json", "psv", "csv", "tsv", "csvh");
```
paul-rogers: Is this the right approach? This validator means that anyone who adds an output file format must modify this file and rebuild all of Drill in order to use that format by default. Not sure we want such tight binding.

Seems a better idea would be to register all known output formats at runtime, validate the option against that list, warn if the format is unknown, and default to Parquet.

ppadma:
closing this pull request. Need to figure out a better way to do this like Paul suggested.

PR #860, opened 2017-06-23
DRILL-5601: Rollup of external sort fixes and improvements

paul-rogers:
- DRILL-5513: Managed External Sort : OOM error during the merge phase
- DRILL-5519: Sort fails to spill and results in an OOM
- DRILL-5522: OOM during the merge and spill process of the managed external sort
- DRILL-5594: Excessive buffer reallocations during merge phase of external sort
- DRILL-5597: Incorrect "bits" vector allocation in nullable vectors allocateNew()
- DRILL-5602: Repeated List Vector fails to initialize the offset vector

All of the bugs have to do with handling low-memory conditions, and with correctly estimating the sizes of vectors, even when those vectors come from the spill file or from an exchange. Hence, the changes for all of the above issues are interrelated.

# Vector Sizing

Testing revealed that Drill has three different memory layouts for vectors:

* Buffer-per-vector as created by the scan operator.
* Buffer-per-batch as created by the exchange receiver.
* Buffer-per-top-level-vector as created when reading a spill file

As it turns out, each layout produces radically different memory allocations for the same data. In general, the buffer-per-vector is the most compact, followed by the buffer-per-top-vector, with the buffer-per-batch the most variable. The reason is simple: the larger the allocation, the larger the internal fragmentation due to unused average 25% of memory in the power-of-two rounded buffers.

The original record batch sizer computed batch size by summing vectors. This worked for the first case above, but was far off for the other two cases. The estimation error caused the sort to incorrectly predict memory needs and thus triggered OOM errors.

To solve this, the implementation no longer looks at vectors to infer size. Instead, it reaches down through layers of abstraction to the "ledgers" that Drill uses to track memory assignment. In the buffer-per-batch case, computing size by vector ignores the very large 25% of memory that is unused. On the other hand, when sizing by ledger, the sizer learns that all vectors are backed by the same ledger, and asks the ledger for its size.

Since vector sizes are not reliable, the portion of the sizer based on that size was deprecated. Instead, the sizer computes actual data size per column, and computes total size per batch. The previous "gross" vector size (which was the buffer size) was dropped in favor of the "net" size.

# Repeated Map Vector Offset Vector

Another memory estimation error occurred with repeated map vectors. Let us first consider non-repeated maps. A "map" in Drill is really a "struct": just a collection of columns that happen to have nested names. That is, Map "m" may have members "a" and "b", referenced as "m.a" and "m.b". But, the vectors for a and b are implemented the same a any top-level vector. As a result, map vectors have no size; they are just containers. The previous "sizer" ignored maps for this reason.

Repeated map vectors are different. Not only are they containers, they are also arrays, and thus have an offset vector. That offset vector must be considered when computing overall batch size. The sizer needs a place to hold the size information. So, the sizer was extended to store a “ColumnSize” instance even for maps. This instance holds the offset vector size for repeated maps.

# Vector Allocation

Drill provides (at least) two tools for allocating buffers for vectors. One is the plain “allocateNew()” method which allocates memory for 4096 entries. The “allocateNew(valueCount)” variant takes additional arguments (here, just value count) needed to determine buffer size. For fixed-width vectors, we need only row count. For Varchar vectors, we need row count and average column width. And so on.

If we use the “basic” method, we end up doubling the buffers multiple times. Assume we start with 4096 entries. If the output needs 64K records, we double buffers from 4K to 8K, 16K, 32K and eventually 64K, copying the data every time.

Worse, some vectors allocate very little memory on initial allocation, causing an extreme number of doublings.

Turning on vector logging during sort tests revealed a very large number of doublings. (As it turns out, Project has the same problem, but that is an exercise for later.)

The solution is to use the “allocateNew()” variant that takes the desired data size. We know the record count, but where do we get average Varchar size or average number of entires per repeated column?

The answer is: from the record batch sizer. So, this PR contains code to pull information from the sizer, organize it as needed to allocate vectors, then use the existing methods to do the allocation. The new “SmartAllocationHelper” class performs this task. It is “Smart” in contrast to the existing “AllocationHelper” which assumes column widths and array sizes.

The “SmartAllocationHelper” is organized around hints that can be associated with each column. For efficiency, these hints should be part of the column metadata. But, for now, to prove the concept, the hints are stored in a hash table indexed by column name. If this approach to vector sizing works, the class will evolve to combine schema and size information into a single column description object, perhaps shared with the record batch sizer.

Code changes require passing the “smart allocation helper” into the sort components that do vector allocation. These components are designed to be independently unit tested. To ease testing, components also allow the helper to be null, in which case they continue to allocate vectors the “old fashioned” way: perfectly fine for testing.

Because the allocation helper depends on observed column sizes, the helper is discarded and recreated each time a significant change occurs in incoming batch size.

# Data, Expected and Worst-Case Batch Sizes

The sort code contains a “memory manager” that plans sort operator memory use. As discussed above, the sort must handle three distinct batch layouts. This revealed the complexity of dealing with the internal fragmentation that comes from Drill’s power-of-two vector allocations.

As it turns out, we must consider three distinct sizes for the input batch, spill batch and the merge batch:

* The data size (which determines size on disk)
* The “expected” size (which assumes, on average, 25% internal fragmentation, used when estimating the total size of a collection of batches)
* The “worst” case size (which assumes 50% internal fragmentation, used when computing memory needs for the single, large, output batches)

Several tests happened to hit the worst-case size, pointing out the need for the three estimates.

Some rather complex code is needed to convert between the sizes depending on where we start.

* If we start with data size (from the record batch sizer, say), we have to expand the size to get the expected and worst-case size.
* If we start from a memory budget (space available for a merge batch, say), we have to shrink that memory estimate to get the data estimate.

Since the sort operates based on row counts, to go from a memory budget to a row count:

* Determine the memory budget
* Choose whether this represents an expected or worst case batch size
* Work backward to the data size, assuming internal fragmentation
* Divide the batch size by average row width to get the row count

Sometimes we have to take a round trip if we are in tight memory.

# Sort Memory Management

Efforts continue to refine the algorithms used to manage memory within the sort.

A problem seen in several tests is that the incoming batch is so large that we can’t keep two in memory as (thought to be) needed for spilling. The “isSpillNeeded” method was changed to long a message if this case occurs as this problem indicates an issue with the scan operator and/or too little memory provided to the sort. That is, it indicates a configuration issue, not a sort bug.

(The two-batch restriction is a bug: see DRILL-5635.)

# Low Memory Revisions

Several revisions were made to the code that handles low-memory situations to better reflect the various buffer layouts and worst-case buffer sizes. Basically, if memory is plentiful, spill and merge batch sizes are of fixed sizes to tune performance.

But, if memory is short, we have to reduce spill and output batch sizes to fit the memory available. (Alas, we cannot do anything about input batch sizes.) In this case, we have to work up from available memory to sizes, subject to a variety of constraints such as the need to put at least one record per batch, and so on.

This code is quite messy; it is hard to explain in the abstract other than as “come up with the best memory layout possible given the constraints provided.” See the code comments for details.

Additional logging was added for the various ways we can be short on memory:

* Records and/or input batches too large for available sort memory during loading
* Records and spill batches too large for available memory during merge

Since these conditions may lead to query failure, detailed logging was added to help diagnose memory-related issues so we can differentiate between those due to bugs and those due to an unhappy collision of low sort memory and large input batches (or rows.)

# Revised Merge Phase Planning

The merge phase occurs when spill files exist on disk, buffered batches occur in memory, and we must merge all of these to produce output batches. Output batches should be large due to Drill’s large per-batch overhead. (More rows per batch amortizes the per-batch cost over more rows.)

The existing merge phase calculations omitted certain obscure cases, such as having a single merge file, and anticipating the extra memory needed to conduct a spill. Again, the code is complex, so see the detailed comments.

# Sort Spill Batch Size

It finally dawned on the code author that there is no advantage at all to having large spill *batches*, though there is advantage to having large spill *files*. Spill batch size was greatly reduced from 8 MB to 1 MB, which helped with tests that torture the sort with very low memory allocations.

# Unit Tests

The sort code allows detailed unit tests. These tests were adjusted for changes made above. New cases were added as needed to reproduce additional issues found in system tests.

# Vector Size Methods

The “getPayloadByteCount” method was revised to take a value count. It turned out that using the value count from the mutator did not always work for deserialized vectors.

The “getAllocatedByteCount” method was removed; replaced with the “getLedgers” method to estimate actual memory allocations, as described earlier.

# Minor Changes

Some minor fixes include:

* Consider power-of-two rounding when estimating the size needed for the in-memory sort SV4.
* Log more detail for spilled batches.
* Pass the “SmartAllocationHelper” into sort components that do vector allocation.
* Moved the “EmptyResults” class within its SortImpl parent file to clean up the code a bit.
* The sort “cheats” a bit by giving itself a 10% memory allocation margin for rounding errors. Logged this adjustment to aid debugging.
* Random code and comment cleanup.
* Minor bug in the profile parser, used for tests, was fixed.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
@@ -19,19 +19,29 @@ package org.apache.drill.exec.physical.impl.spill;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Set;
 
+import org.apache.drill.common.types.TypeProtos.DataMode;
 import org.apache.drill.common.types.TypeProtos.MinorType;
 import org.apache.drill.exec.expr.TypeHelper;
+import org.apache.drill.exec.memory.AllocationManager.BufferLedger;
 import org.apache.drill.exec.memory.BaseAllocator;
 import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.RecordBatch;
+import org.apache.drill.exec.record.VectorInitializer;
 import org.apache.drill.exec.record.VectorAccessible;
 import org.apache.drill.exec.record.VectorWrapper;
 import org.apache.drill.exec.record.selection.SelectionVector2;
+import org.apache.drill.exec.vector.UInt4Vector;
 import org.apache.drill.exec.vector.ValueVector;
 import org.apache.drill.exec.vector.complex.AbstractMapVector;
+import org.apache.drill.exec.vector.complex.RepeatedListVector;
+import org.apache.drill.exec.vector.complex.RepeatedMapVector;
+import org.apache.drill.exec.vector.complex.RepeatedValueVector;
 import org.apache.drill.exec.vector.VariableWidthVector;
+
+import com.google.common.collect.Sets;
 
 /**
  * Given a record batch or vector container, determines the actual memory
@@ -39,25 +49,21 @@ import org.apache.drill.exec.vector.VariableWidthVector;
  */
 
 public class RecordBatchSizer {
-//  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(RecordBatchSizer.class);
 
   /**
    * Column size information.
    */
   public static class ColumnSize {
+    public final String prefix;
     public final MaterializedField metadata;
 
     /**
-     * Assumed size from Drill metadata.
+     * Assumed size from Drill metadata. Note that this information is
+     * 100% bogus. Do not use it.
      */
 
+    @Deprecated
     public int stdSize;
-
-    /**
-     * Actual memory consumed by all the vectors associated with this column.
-     */
-
-    public int totalSize;
 
     /**
      * Actual average column width as determined from actual memory use. This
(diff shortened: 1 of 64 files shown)


paul-rogers:
# Additional Revisions

The code contains an optimization for short queries. If the query has just one batch then the sort need not copy that batch to produce the output. Instead, the sort simply passes along the input batch along with an SV2. This commit fixes some issues with transferring that single, buffered, input batch to the sort’s output container.

While reproducing problems found by QA, it turned out to be useful to be able to parse SQL statements from a file in the context of the test framework. Added test methods for this purpose.

The “AllocationHelper” class turns out to throw an exception if asked to allocate a vector with zero elements. Since a zero-size can occasionally come from the record batch sizer (via the “smart allocation helper”, special code was added to handle this case.

Backed out some debugging code that accidentally appeared in the original PR.

With this set of improvements, the revised, managed sort has become more stable and reliable than the original version. As a result, this commit enables the managed sort by default.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -247,27 +249,26 @@ public void copyEntry(int toIndex, ValueVector from, int fromIndex) {
...
     }
+    // If 1 or more values, then the last value is set to
+    // the offset of the next value, which is the same as
+    // the length of existing values.
+    // In addition to the actual data bytes, we must also
+    // include the "overhead" bytes: the offset vector entries
+    // that accompany each column value. Thus, total payload
+    // size is consumed text bytes + consumed offset vector
+    // bytes.
+    return offsetVector.getAccessor().get(valueCount) +
```
Ben-Zvi: Should this be **(valueCount - 1)** ??

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/BaseRepeatedValueVector.java
```diff
@@ -210,13 +212,15 @@ protected void replaceDataVector(ValueVector v) {
   }
 
   @Override
-  public int getAllocatedByteCount() {
-    return offsets.getAllocatedByteCount() + vector.getAllocatedByteCount();
+  public void getLedgers(Set<BufferLedger> ledgers) {
```
Ben-Zvi: This method looks more like a "set" than a "get" .....

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/ListVector.java
```diff
@@ -323,12 +325,15 @@ public void setValueCount(int valueCount) {
   }
 
   @Override
-  public int getAllocatedByteCount() {
-    return offsets.getAllocatedByteCount() + bits.getAllocatedByteCount() + super.getAllocatedByteCount();
+  public void getLedgers(Set<BufferLedger> ledgers) {
```
Ben-Zvi: Same comment - set vs. get

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/RepeatedListVector.java
```diff
@@ -435,20 +438,18 @@ public void copyEntry(int toIndex, ValueVector from, int fromIndex) {
     copyFromSafe(fromIndex, toIndex, (RepeatedListVector) from);
   }
 
-  @Override
-  public int getAllocatedByteCount() {
-    return delegate.getAllocatedByteCount();
+  public void getLedgers(Set<BufferLedger> ledgers) {
```
Ben-Zvi: again ....

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/RepeatedMapVector.java
```diff
@@ -598,7 +604,8 @@ public void clear() {
   }
 
   @Override
-  public int getAllocatedByteCount() {
-    return super.getAllocatedByteCount( ) + offsets.getAllocatedByteCount();
+  public void getLedgers(Set<BufferLedger> ledgers) {
```
Ben-Zvi: again ....

exec/java-exec/src/test/java/org/apache/drill/test/ClientFixture.java
```diff
@@ -231,4 +236,120 @@ public void setControls(String controls) {
...
+          break;
+        }
+        buf.append((char) c);
+        if (c == '"' || c == '\'' || c == '`') {
+          int quote = c;
+          boolean escape = false;
+          for (;;) {
+            c = in.read();
+            if (c == -1) {
+              break outer;
```
Ben-Zvi: Note: This code would return a truncated SQL Statement (non matched quote), as appears in the input.
Option: Throw some user exception here.

exec/vector/src/main/java/org/apache/drill/exec/vector/AllocationHelper.java
```diff
@@ -26,14 +26,16 @@ public static void allocate(ValueVector v, int valueCount, int bytesPerValue) {
     allocate(v, valueCount, bytesPerValue, 5);
   }
 
-  public static void allocatePrecomputedChildCount(ValueVector v, int valueCount, int bytesPerValue, int childValCount){
-    if(v instanceof FixedWidthVector) {
+  public static void allocatePrecomputedChildCount(ValueVector v, int valueCount, int bytesPerValue, int childValCount) {
```
Ben-Zvi: Discouraged: single letter parameter/field names. Even **"vv"** would be easier to search, if needed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/SpillSet.java
```diff
@@ -413,7 +419,12 @@ public SpillSet(DrillConfig config, FragmentHandle handle, PhysicalOperator popC
       fileManager = new HadoopFileManager(spillFs);
     }
 
-    spillDirName = String.format("%s_%s_%s-%s_minor%s", QueryIdHelper.getQueryId(handle.getQueryId()),
+    // If provided with a prefix to identify the Drillbit, prepend that to the
+    // spill directory.
+
+    spillDirName = String.format("%s%s_%s_%s-%s-%s",
+        ep == null ? "" : ep.getAddress() + "-" + ep.getUserPort() + "/",
```
Ben-Zvi: Any chance the address would be "" ?  Then the directory name would start with a "-" which could cause some trouble (probably only for access from the shell, thus OK).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
```diff
@@ -114,15 +127,25 @@ public SortConfig(DrillConfig config) {
     } else {
       mSortBatchSize = Character.MAX_VALUE;
     }
+
+    // Disable the single-batch shortcut; primarily for testing
+
+    if (config.hasPath(ExecConstants.EXTERNAL_SORT_OPTIMIZE_SINGLE_BATCH)) {
+      enableSingleBatchShortcut = config.getBoolean(ExecConstants.EXTERNAL_SORT_OPTIMIZE_SINGLE_BATCH);
+    } else {
```
Ben-Zvi: Minor comment - can save 3 LOC by initializing to "true" (and dropping the "final") , then no need for the else { }

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/PriorityQueueCopierWrapper.java
```diff
@@ -82,7 +84,7 @@ private PriorityQueueCopier newCopier(VectorAccessible batch) {
     ClassGenerator<PriorityQueueCopier> g = cg.getRoot();
     cg.plainJavaCapable(true);
     // Uncomment out this line to debug the generated code.
-//    cg.saveCodeForDebugging(true);
+    cg.saveCodeForDebugging(true);
```
paul-rogers: I guess the fiddle-factor is getting to high; time to implement your suggestion of having options to enable this stuff. I'll try to add that in some future PR.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -239,9 +239,7 @@ drill.exec: {
     external: {
       // Drill uses the managed External Sort Batch by default.
       // Set this to true to use the legacy, unmanaged version.
-      // Disabled in the intial commit, to be enabled after
-      // tests are committed.
-      disable_managed: true,
+      disable_managed: false,
```
paul-rogers: Done later.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -70,52 +72,90 @@
...
+     */
+
     public final int valueCount;
-    public final int entryCount;
-    public final int dataSize;
-    public final int estElementCount;
+
+    /**
+     * Number of times the entries of this column appears. If this is a
+     * scalar, the entry count is the same as the value count. If this
```
paul-rogers: The comment is poorly worded. Meant "array of scalar" vs. "Array of variable-width". There is also "array of array of scalar | variable width"... Please check if the revised comment makes this any clearer. (It is hard to explain without first explaining the whole column data model...)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
paul-rogers: Yes. But the "final" provides useful information: the value can't change once set. Once we accept "final" then we're stuck with the awkward code to initialize the value only once.

paul-rogers:
Fixed; slightly different than suggested, but similar. Didn't want to change this code too terribly much, so did the minimum change that would work.

paul-rogers:
Squashed commits.

paul-rogers:
Rebased onto master.

jinfengni:
+1

LGTM.

PR #861, opened 2017-06-27
DRILL-4722: Fix EqualityVisitor for interval day expressions with millis

vvysotskyi:
Please see the problem description in [DRILL-4722](https://issues.apache.org/jira/browse/DRILL-4722)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/EqualityVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/EqualityVisitor.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/test/java/org/apache/drill/TestFunctionsQuery.java
```diff
@@ -923,4 +923,40 @@ public void testRandom() throws Exception {
             .baselineValues(false)
             .go();
   }
+
```
arina-ielchiieva: Let's create new test class for `date_add` similar to `TestDateTruncFunctions` and test there all interval variations (year, month, hour etc).

vvysotskyi: Thanks, done

arina-ielchiieva:
+1, LGTM.

PR #862, opened 2017-06-28
DRILL-5420: ParquetAsyncPgReader goes into infinite loop during cleanup

kkhatua:
PageQueue is cleaned up using poll() instead of take(), which constantly gets interrupted and causes CPU churn.
During a columnReader shutdown, a flag is set so as to block any new page reading tasks from being submitted, before the queues are finally cleared and memory occupied by the pages released.
More details are in this JIRA comment : https://issues.apache.org/jira/browse/DRILL-5420?focusedCommentId=16066933&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16066933

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
@@ -130,7 +130,10 @@ class AsyncPageReader extends PageReader {
 
   @Override protected void init() throws IOException {
     super.init();
-    asyncPageRead.offer(threadPool.submit(new AsyncPageReaderTask(debugName, pageQueue)));
+    //Avoid Init if a shutdown is already in progress even if init() is called once
+    if (!parentColumnReader.isShuttingDown) {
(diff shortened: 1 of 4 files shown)


kkhatua:
@parthchandra Can you please review this patch?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -409,9 +414,10 @@ public Void call() throws IOException {
 
       DrillBuf pageData = null;
       timer.reset();
-      try {
+      parqProcessing: try {
```
parthchandra: Maybe just check isShuttingDown here and return? (breaks with labels with no outer for/while loops confuse me!)

kkhatua: Good point. I didn't realize that the `finally` block is empty. If there was something to be done, a labelled break would allow the finally to be executed.

kkhatua:
@parthchandra Please review the final change. I'll mark the JIRA with a ready-to-commit label if everything looks fine.

parthchandra:
LGTM.

PR #863, opened 2017-06-29
DRILL-4970: Prevent changing the negative value of input holder for cast functions

vvysotskyi:
Problem description in [DRILL-4970](https://issues.apache.org/jira/browse/DRILL-4970)


exec/java-exec/src/main/codegen/templates/CastFunctions.java
```diff
@@ -54,14 +56,14 @@ public void setup() {}
   public void eval() {
     <#if (type.from.startsWith("Float") && type.to.endsWith("Int"))>
     boolean sign = (in.value < 0);
-    in.value = java.lang.Math.abs(in.value);
-    ${type.native} fractional = in.value % 1;
+    ${type.native} value = java.lang.Math.abs(in.value);
+    ${type.native} fractional = value % 1;
```
paul-rogers: `out.value = Math.round(in.value)` ?

`Math.round()` rounds a double to the closest long; handling overly large and small values, which our code does not do. The Java method is likely to be much more efficient than this code.

Since the particular case is cast from `double` to `int`:

`out.value = Math.round((float) in.value)`

Returns an int.

I realize that this is existing code. But, it is broken. So, we might as well fix it the right way...

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
```diff
@@ -78,4 +81,71 @@ public void testToDateForTimeStamp() throws Exception {
         .build()
         .run();
   }
+
+  @Test // DRILL-4970
+  public void testCastNegativeFloatToInt() throws Exception {
+    try {
+      test("create table dfs_test.tmp.table_with_float as\n" +
```
paul-rogers: A full test would handle all conditions:

* 0
* +/- 0.4
* +/- 0.5
* INTEGER.MAX_VALUE as double
* INTEGER.MAX_VALUE + 0.4 as double
* INTEGER.MAX_VALUE + 0.5 as double
* INTEGER.MIN_VALUE as double
* INTEGER.MIN_VALUE - 0.4 as double
* INTEGER.MIN_VALUE - 0.5 as double
* DOUBLE.MAX_VALUE
* DOUBLE.MIN_VALUE

paul-rogers: Also, would be great if this can be unit tested directly, without a Drillbit. Use the `SubOperatorTest` framework if needed. But, since these are simple math functions, a basic JUnit test should be fine. Allows testing many more values without the complexity of a query.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
```diff
@@ -78,4 +81,71 @@ public void testToDateForTimeStamp() throws Exception {
...
+            .run();
+        }
+      }
+    } finally {
+      test("drop table if exists dfs_test.tmp.table_with_float");
+    }
+  }
+
+  @Test // DRILL-4970
+  public void testCastNegativeDecimalToVarChar() throws Exception {
```
paul-rogers: There are two tests here. Should we not test all cast functions to determine if others have similar issues: (short, int, long, float, double, decimal 9, decimal 18, etc.) --> (short, int, long, float, double, decimal 9, decimal 18, etc.)

This is a reason to do basic JUnit tests without the server: much easier to try all the boundary values for every type to every other type. That testing is tedious, yes. But, much less tedious using the functions directly.

exec/java-exec/src/main/codegen/templates/CastFunctions.java
vvysotskyi: Thanks for the pointing this. Drill uses `half away from zero` rounding mode, so method `Math.round()` wouldn't help. I rewrote current code and now it handles boundary cases in the same way as the `Math.round()`.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
vvysotskyi: The initial bug was caused by changing the input holder when negative values casting, so current test checks the correctness of the fix only.
Added tests to check cast functions with boundary values.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
vvysotskyi: The value of input holder was changing only for types that was tested. Cast function makes only explicit cast of the value for most of the types, so this bug couldn't appear for this types.

paul-rogers:
Not entirely thrilled with focusing on one bug without cleaning up the entire area, but that is another discussion.
+1

PR #865, opened 2017-07-03
DRILL-5634 - Add Crypto and Hash Functions

cgivre:
DRILL-5634: Crypto and Hash Functions for Drill
Please see #865 for an explanation.

This PR implements common hash and encryption methods including:

- MD2
- MD5
- SHA1
- SHA2 (SHA256, SHA384, SHA512)
- AES Encryption/Decryption


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,315 @@
+/**
```
arina-ielchiieva: Please include Apache header in a form of comment. Please correct here and in files below.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,315 @@
...
+
+    @FunctionTemplate(
+        name = "md5",
+        scope = FunctionTemplate.FunctionScope.SIMPLE,
+        nulls = FunctionTemplate.NullHandling.NULL_IF_NULL
+    )
+    public static class md5Function implements DrillSimpleFunc {
+
+        @Param
+        VarCharHolder raw_input;
```
arina-ielchiieva: Please use camel case here and throughout the code.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,315 @@
...
+
+        @Workspace
+        java.security.MessageDigest md;
+
+        @Override
+        public void setup() {
+            try {
+                md = java.security.MessageDigest.getInstance("MD5");
+            } catch( Exception e ) {
+            }
```
arina-ielchiieva: What will happen if we couldn't get MD5 instance? Shouldn't we throw an exception in this case?
We won't be able to provide correct result anyway.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,315 @@
...
+        DrillBuf buffer;
+
+        @Workspace
+        java.security.MessageDigest md;
+
+        @Override
+        public void setup() {
+            try {
+                md = java.security.MessageDigest.getInstance("MD5");
+            } catch( Exception e ) {
```
arina-ielchiieva: Please use proper formatting here and throughout the code [1].

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,315 @@
...
+            String output_string = "";
+
+            try {
+                byte[] bytesOfMessage = input.getBytes("UTF-8");
+                thedigest = md.digest(bytesOfMessage);
+                output_string = String.format("%032X", new java.math.BigInteger(1, thedigest));
+                output_string = output_string.toLowerCase();
+
+            } catch( Exception e ) {
+            }
```
arina-ielchiieva: 1. Please mind formatting.
2. If in case of error, we leave the defaults, then it's better  to at least log the exception.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,315 @@
...
+    static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CryptoFunctions.class);
+
+    private CryptoFunctions() {}
+
+    @FunctionTemplate(
+        name = "md5",
+        scope = FunctionTemplate.FunctionScope.SIMPLE,
+        nulls = FunctionTemplate.NullHandling.NULL_IF_NULL
+    )
+    public static class md5Function implements DrillSimpleFunc {
```
arina-ielchiieva: I guess according to java naming convention class names should start with letter in upper case.
Please correct here and in classes below.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoHelperFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+    {
+        MessageDigest sha = null;
+        try {
+            key = myKey.getBytes("UTF-8");
+            sha = MessageDigest.getInstance("SHA-1");
+            key = sha.digest(key);
+            key = Arrays.copyOf(key, 16);
+            secretKey = new SecretKeySpec(key, "AES");
+        }
+        catch (NoSuchAlgorithmException e) {
```
arina-ielchiieva: 1. Please use multi-catch.
2. Please remove e.printStackTrace() and log error instead here and in the code below.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCryptoFunctions.java
```diff
@@ -0,0 +1,91 @@
...
+            .sqlQuery(query)
+            .ordered()
+            .baselineColumns("md5_hash")
+            .baselineValues("ae2b1fca515949e5d54fb22b8ed95575")
+            .go();
+    }
+
+    @Test
+    public void testSHA1() throws Exception {
+        final String query = "select sha( 'testing' ) as sha_hash from (values(1))";
```
arina-ielchiieva: Please remove extra space here and in test below: `sha( 'testing' ) ` -> `sha('testing') `

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoHelperFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+import javax.xml.bind.DatatypeConverter;
+import java.io.UnsupportedEncodingException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+
+public class CryptoHelperFunctions {
+
+    private static SecretKeySpec secretKey;
+    private static byte[] key;
```
arina-ielchiieva: In concurrent environment this code will be prone to produce incorrect results, since each thread may have its own key. Saving state in static variables is not thread-safe.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoHelperFunctions.java
```diff
@@ -0,0 +1,85 @@
...
+package org.apache.drill.exec.expr.fn.impl;
+import javax.crypto.Cipher;
+import javax.crypto.spec.SecretKeySpec;
+import javax.xml.bind.DatatypeConverter;
+import java.io.UnsupportedEncodingException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+
+public class CryptoHelperFunctions {
```
arina-ielchiieva: Please add java doc describing the purpose of this class and also please add java doc to the methods below.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,315 @@
...
+import org.apache.drill.exec.expr.DrillSimpleFunc;
+import org.apache.drill.exec.expr.annotations.FunctionTemplate;
+import org.apache.drill.exec.expr.annotations.Output;
+import org.apache.drill.exec.expr.annotations.Param;
+import org.apache.drill.exec.expr.annotations.Workspace;
+import org.apache.drill.exec.expr.holders.VarCharHolder;
+
+import javax.inject.Inject;
+
+public class CryptoFunctions{
```
arina-ielchiieva: Please add java-doc to all the functions below (you have provided good description in Jira but its better if we'll have it in code).

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,345 @@
...
+
+    @Override
+    public void setup() {
+      key = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(rawKey.start, rawKey.end, rawKey.buffer);
+      java.security.MessageDigest sha = null;
+      try {
+        keyByteArray = key.getBytes("UTF-8");
+        sha = java.security.MessageDigest.getInstance("SHA-1");
+        keyByteArray = sha.digest(keyByteArray);
+        keyByteArray = java.util.Arrays.copyOf(keyByteArray, 16);
```
arina-ielchiieva: Might want to add comment here that we copy into array with length of 16 bytes since we need to ensure that input length is multiple of 16 bytes.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,345 @@
...
+    DrillBuf buffer;
+
+    @Workspace
+    String key;
+
+    @Workspace
+    SecretKeySpec secretKey;
+
+    @Workspace
+    byte[] keyByteArray;
```
cgivre: Fixed in latest revision.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,345 @@
...
```
cgivre: Fixed in latest version.  I was tweaking the code (and it was late) and originally all that was done in the `eval()` function.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,345 @@
...
+      } catch (Exception e) {
+        logger.debug(e.getMessage());
+      }
+    }
+
+    @Override
+    public void eval() {
+
+      String input = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(rawInput.start, rawInput.end, rawInput.buffer);
+      byte[] thedigest = null;
```
cgivre: Fixed in latest version.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,345 @@
...
+      String outputString = "";
+
+      try {
+        byte[] bytesOfMessage = input.getBytes("UTF-8");
+        thedigest = md.digest(bytesOfMessage);
+        outputString = String.format("%032X", new java.math.BigInteger(1, thedigest));
+        outputString = outputString.toLowerCase();
+
```
cgivre: Yes it can!

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
arina-ielchiieva: Sure, no problem. Please let me know when you push final changes so I'll be able to finish CR.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,395 @@
...
+    VarCharHolder rawInput;
+
+    @Output
+    VarCharHolder out;
+
+    @Inject
+    DrillBuf buffer;
+
+    @Workspace
+    java.security.MessageDigest md;
```
arina-ielchiieva: Unused.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,395 @@
...
```
arina-ielchiieva: Unused.

arina-ielchiieva:
+1, LGTM.

parthchandra:
Has anyone checked the export requirements for these functions? 
https://www.apache.org/dev/crypto.html

cgivre:
Hi Parth, 
I need to look at AES, but on the page you sent it specifically said that all the digests (MD5, SHA) have no restrictions. 
—C

parthchandra:
AES is specifically restricted. Can you check on commons-codec and see what it uses to implement AES support?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
+    @Override
+    public void setup() {
+      String key = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(rawKey.start, rawKey.end, rawKey.buffer);
+      java.security.MessageDigest sha = null;
+      try {
+        byte[] keyByteArray = key.getBytes("UTF-8");
+        sha = java.security.MessageDigest.getInstance("SHA-1");
+        keyByteArray = sha.digest(keyByteArray);
+        keyByteArray = java.util.Arrays.copyOf(keyByteArray, 16);
+        SecretKeySpec secretKey = new SecretKeySpec(keyByteArray, "AES");
```
arina-ielchiieva: Charles, do we need direct imports for `javax.crypto` package?

cgivre: @arina-ielchiieva I'm not sure.  Is there a style or functional reason to use the direct imports?

arina-ielchiieva: We don't need to use fully qualified class name when variable is declared in class (like for `Cipher`, import will be added automatically). For `SecretKeySpec` we need to use fully qualified class name since it's declared internally in method: `javax.crypto.spec.SecretKeySpec secretKey = new javax.crypto.spec.SecretKeySpec(keyByteArray, "AES");`

For example, Drill will create `ProjectorGen0` where it will add variables declared in class (thus imports will be added automatically). Also it will copy content of `setup` and `eval` methods, in this case imports won't be added automatically. We might need to enhance this mechanism so all imports UDFs class are added and user won't required to use fully qualified class names. The same problem be with `logger`.

Actually, I have run `TestCryptoFunctions` on your branch and `testAESEncrypt()` and `testAESDecrypt()` fail.
First with ` java.lang.Exception: org.apache.drill.common.exceptions.UserRemoteException: SYSTEM ERROR: CompileException: Line 55, Column 32: Cannot determine simple type name "SecretKeySpec"` and when I have fixed this problem with `Caused By (org.codehaus.commons.compiler.CompileException) Line 61, Column 16: Unknown variable or type "logger"`.

Note: it's a good practice (I even would say a requirement) to make sure that all unit test pass before publishing changes in Drill.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
+    @Inject
+    DrillBuf buffer;
+
+    @Workspace
+    Cipher cipher;
+
+    @Override
+    public void setup() {
+      String key = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(rawKey.start, rawKey.end, rawKey.buffer);
+      java.security.MessageDigest sha = null;
```
arina-ielchiieva: Variable initializer `null` is redundant.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
```
arina-ielchiieva: Variable initializer `null` is redundant.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
+import org.apache.drill.exec.expr.annotations.Param;
+import org.apache.drill.exec.expr.annotations.Workspace;
+import org.apache.drill.exec.expr.holders.VarCharHolder;
+
+import javax.crypto.Cipher;
+import javax.crypto.spec.SecretKeySpec;
+import javax.inject.Inject;
+
+public class CryptoFunctions {
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CryptoFunctions.class);
```
arina-ielchiieva: Can be removed since we don't need it anymore. See explanation below.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
+
+    @Override
+    public void eval() {
+
+      String input = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(rawInput.start, rawInput.end, rawInput.buffer);
+      String encryptedText = "";
+      try {
+        encryptedText = javax.xml.bind.DatatypeConverter.printBase64Binary(cipher.doFinal(input.getBytes("UTF-8")));
+      } catch (Exception e) {
+        logger.debug(e.getMessage());
```
arina-ielchiieva: Replace with comment explaining why we don't care about the exception.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
+        byte[] keyByteArray = key.getBytes("UTF-8");
+        sha = java.security.MessageDigest.getInstance("SHA-1");
+        keyByteArray = sha.digest(keyByteArray);
+        keyByteArray = java.util.Arrays.copyOf(keyByteArray, 16);
+        SecretKeySpec secretKey = new SecretKeySpec(keyByteArray, "AES");
+
+        cipher = Cipher.getInstance("AES/ECB/PKCS5Padding");
+        cipher.init(Cipher.DECRYPT_MODE, secretKey);
+      } catch (Exception e) {
+        logger.debug(e.getMessage());
```
arina-ielchiieva: Replace with comment explaining why we don't care about the exception.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
+    @Override
+    public void setup() {
+      String key = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(rawKey.start, rawKey.end, rawKey.buffer);
+      java.security.MessageDigest sha = null;
+      try {
+        byte[] keyByteArray = key.getBytes("UTF-8");
+        sha = java.security.MessageDigest.getInstance("SHA-1");
+        keyByteArray = sha.digest(keyByteArray);
+        keyByteArray = java.util.Arrays.copyOf(keyByteArray, 16);
+        SecretKeySpec secretKey = new SecretKeySpec(keyByteArray, "AES");
```
arina-ielchiieva: Use fully qualified class name.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
```diff
@@ -0,0 +1,389 @@
...
+
+    @Override
+    public void eval() {
+
+      String input = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(rawInput.start, rawInput.end, rawInput.buffer);
+      String decryptedText = "";
+      try {
+        decryptedText = new String(cipher.doFinal(javax.xml.bind.DatatypeConverter.parseBase64Binary(input)));
+      } catch (Exception e) {
+        logger.debug(e.getMessage());
```
arina-ielchiieva: Replace with comment explaining why re don't care about the exception.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
cgivre: Removed

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
cgivre: Done

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/CryptoFunctions.java
cgivre: Removed

arina-ielchiieva:
+1, checked the latest changes and ran unit tests.

PR #868, opened 2017-07-05
DRILL-5547:Linking config options with system option manager

dvjyothsna:
Drill has two different config systems each with its own namespace.First being the HOCON based boot time config system.This is a hierarchical system where the top layers override the bottom ones in the following order
 
 *  Java System Options
 *  distrib.conf
 *  drill-override.conf
 *  drill-module.conf
 
 These are the options that are set before the drill starts.But once drill starts System or session options can be modified using ALTER SYSTEM/SESSION.Even this system provides inheritance sytle in the following order
 
 *  Session options
 *  System options
 *  Hardcoded defaults
 
But system/session options have a validator and the validator has a hard coded default value for every option. In the current system validators are registered so that system/session options will always have a default value. So when a system/session options are not explicitly set or a user system/session option is null the hardcoded default was applied since it checks if the option value is null and returns the default set in the validator.But the config options set during boot time are never read and honored since there is no linkage between the two config systems.It is also evident that there are some places where there is some ad-hoc linkage between the two systems.For example, for the code gen compiler, config options are supposed to be read if the system option is not null.But as the validator provides the default values config options are never taken into consideration. The goal of the new system is to link both the systems in such a way that boot-time config options take precedence over the hard coded defaults set in the validator.All the options in the option validator i.e.c options from Exec constants, planner settings etc., are extracted and put into a new name space called drill.exec.options in the .conf file. The default values of the validators in the option validator are populated with the values in the boot-config.This way the values set in the boot time config system are honored.Any user who wish to change the option values in the
config should change the options under the name space drill.exec.options


exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -271,12 +320,20 @@ public static OptionValidator getValidator(final String name) {
...
     if (value != null) {
       return value;
     }
 
-    // otherwise, return default.
+    // otherwise, return default set in the validator.
     final OptionValidator validator = getValidator(name);
+    if (validator.getDefault() != null) {
+      if (!validator.getDefault().getValue().equals(validator.getDefault().getValue())) {
+                System.out.println("Config and hardcoded values are" + "\t" + validator.getDefault().getValue()
```
paul-rogers: We can't use println in production code. Instead, we can use the logger:

```
logger.trace("Config value: {}, hardcoded value: {},
 validator.getDefault().getValue(),
validator.getDefault().getValue());
```
The logger will call `toString()` on each argument for us.

Actually, this code seems to be temporary debugging code: checking the default value against hard-coded. Once the work is done, the hard-coded values should no longer exist.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -271,12 +320,20 @@ public static OptionValidator getValidator(final String name) {
   public OptionValue getOption(final String name) {
     // check local space (persistent store)
     final OptionValue value = options.get(name.toLowerCase());
+    OptionValue val;
```
paul-rogers: Unused?

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -48,11 +50,46 @@
...
+ *  drill-module.conf
+ *
+ *  These are the options that are set before the drill starts.But once drill starts System or session options can be
+ *  modified using ALTER SYSTEM/SESSION.Even this system provides inheritance sytle in the following order
+
+ *  Session options
+ *  System options
+ *  Hardcoded defaults
+
+ *  But system/session options have a validator and the validator has a hard coded default value for every option. In
```
paul-rogers: This description would be great as a comment for the pull request. Now, image that this code is committed to Drill, and someone reads this a year from now. By then, the comment will be describing ancient history.

We can see that this kind of comment should describe the system the way it is (or will be) after this change without need to reference history. For those who might be familiar with the old way, we often point them to the JIRA that describes the change.

Example:

"Default for system properties are externalized to the boot-time config file. See DRILL-5547. ..."

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +370,31 @@ public void deleteAllOptions(OptionType type) {
     }
   }
 
+  public void populateDefualtValues(Map<String, OptionValidator> VALIDATORS) {
+
+    // populate the options from the config
+    final Map<String, OptionValidator> tmp = new HashMap<>();
```
paul-rogers: No need to create a new map. Setting the default value does not change the key, so you can just update the fields of the value. (That is, we are not even replacing the value; we are just altering a field in the value.)

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +370,31 @@ public void deleteAllOptions(OptionType type) {
...
+    for (final Map.Entry<String, OptionValidator> entry : VALIDATORS.entrySet()) {
+
+      OptionValidator validator = entry.getValue();
+      final OptionValue.Kind kind = validator.getKind();
+      String name = entry.getValue().getOptionName();
```
paul-rogers: Given that we get the name from the validator, we can iterate over map values:

```
for (OptionValidator validator : validators.valueSet()) {
  ...
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +370,31 @@ public void deleteAllOptions(OptionType type) {
     }
   }
 
+  public void populateDefualtValues(Map<String, OptionValidator> VALIDATORS) {
```
paul-rogers: Nit: the `VALIDATORS` variable is not a constant here, so we should call it something like `validators`.

Or, since `VALIDATORS` is a static member, we can access it directly; no need to pass it into this method.

Moreover, since `VALIDATORS` is static, this whole method can be static. It should be called (once) from somewhere in, say, the Drillbit startup logic. But, since we can have multiple drillbits in the same process (when testing), we should perhaps have a flag so that initialization is done only on the first call, and synchronized so that two Drillbits don't try to do the same init at the same time.

Or, we might want to have a separate table per Drillbit. Maybe the items should become a non-static member of the SessionOptions class, if we have one SessionOptions per Drillbit. That way we can test strange scenarios, such as two Drillbits having different defaults (which is a bug, but we might want to test if we can catch that improper setup...)

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +370,31 @@ public void deleteAllOptions(OptionType type) {
...
+
+    // populate the options from the config
+    final Map<String, OptionValidator> tmp = new HashMap<>();
+    for (final Map.Entry<String, OptionValidator> entry : VALIDATORS.entrySet()) {
+
+      OptionValidator validator = entry.getValue();
+      final OptionValue.Kind kind = validator.getKind();
+      String name = entry.getValue().getOptionName();
+      OptionValue value;
+      String configPath = "drill.exec.options.";
```
paul-rogers: Let's pull the config path out to the top of the file as a constant. Better, as noted elsewhere, don't pass it in. Make it a constant in the base option validator class.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +370,31 @@ public void deleteAllOptions(OptionType type) {
...
+      String configName = configPath + name;
+      try {
+        value = validator.loadConfigDefault(bootConfig,name,configPath);
+        validator.setDefaultValue(value);
```
paul-rogers: Setting the default should be done in `loadConfigDefault`. Actually, since there is only one place to load the default from, the name `loadDefault` is fine.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +370,31 @@ public void deleteAllOptions(OptionType type) {
...
+        tmp.put(name, validator);
```
paul-rogers: As noted above, no need to alter the map as neither the key nor value object changed.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +370,31 @@ public void deleteAllOptions(OptionType type) {
...
+      } catch (ConfigException.Missing e) {
+        logger.error(e.getMessage(), e);
```
paul-rogers: Seems we'd want to throw a fatal exception here, such as `IllegalStateException` because this indicates, to the developers, that the developer forgot to externalize the config default when they created the validator.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/TypeValidators.java
```diff
@@ -139,6 +140,10 @@ public BooleanValidator(String name, boolean def) {
     public BooleanValidator(String name, boolean def, boolean isAdminOption) {
       super(name, Kind.BOOLEAN, OptionValue.createBoolean(OptionType.SYSTEM, name, def), isAdminOption);
     }
+
+    public OptionValue loadConfigDefault(DrillConfig bootConfig, String name, String configPath){
+      return OptionValue.createBoolean(OptionType.SYSTEM, name, bootConfig.getBoolean(configPath+name));
```
paul-rogers: Instead of `configPath+name` multiple times, if we turn `configPath` into a constant, and get `name` from the validator itself, we want:

```
...getBoolean(getConfigProperty());
...
private String getConfigProperty() {
    return CONFIG_PATH + getOptionName(); }
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Wrapper class for Extended Option Value
+   */
+  public static class ExtendedOptionValueWrapper {
+
+    public final String name;
+    public final Kind kind;
+    public final OptionType type;
+    public final String string_value;
```
paul-rogers: string_value --> value

Since this is a column name, might as well make it simple.

Also, can we include the data type? Each validator can return its type. For example, LongValidator can return either `Long.class` (kind of messy), or `OptionDataType.LONG` (which is a bit less messy). Then, convert the internal type to a nice display name, perhaps mapping to the DrillType. So, a long becomes `Bigint` or `BIGINT`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -0,0 +1,152 @@
...
+    for the same option based on the type
+   */
+  public Iterator<OptionValue> sortOptions(Iterator<OptionValue> options)
+  {
+    List<OptionValue> values = Lists.newArrayList(options);
+    List<OptionValue> optionValues = Lists.newArrayList();
+    List<OptionValue> optionValueList = Lists.newArrayList();
+    OptionValue value;
+    OptionType type;
+    OptionValue[] temp = new OptionValue[3];
```
paul-rogers: I think this may be a bit easier to implement with a map. Add each "inner" option to a map keyed by value. If you find an existing entry, choose the one with higher precedence and replace the map value if needed.

Then, iterate over the map values to get the final objects. Iteration over the values can be done in each call to the `next()` method of your iterator to avoid the need to build up a list of extended value objects.

exec/java-exec/src/main/java/org/apache/drill/exec/util/MemoryAllocationUtilities.java
```diff
@@ -58,7 +58,15 @@ public static void setupBufferedOpsMemoryAllocations(final PhysicalPlan plan, fi
     // if there are any sorts, compute the maximum allocation, and set it on them
     if (bufferedOpList.size() > 0) {
       final OptionManager optionManager = queryContext.getOptions();
-      final long maxWidthPerNode = optionManager.getOption(ExecConstants.MAX_WIDTH_PER_NODE_KEY).num_val;
+      final long maxWidthPerNode;
+      long maxWidth = optionManager.getOption(ExecConstants.MAX_WIDTH_PER_NODE);
```
paul-rogers: This code appears twice. Can we move it into a function so it appears once? Maybe on to a purpose-build Validator class?

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -338,3 +338,127 @@ drill.exec: {
 drill.jdbc: {
   batch_queue_throttling_threshold: 100
 }
+
+drill.exec.options:  {
```
paul-rogers: Comment to explain what this is? And to remind people how to use it: create a Validator. Create an entry below with the same name as the session/system option, with a default of a compatible type.

paul-rogers: Also, any chance we can alphabetize the list? Just run it through the `sort` command? Will make future human access just a bit easier...

kfaraaz:
@dvjyothsna Can you please add unit tests that cover your changes ?

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -314,6 +344,19 @@ public void deleteAllOptions(OptionType type) {
...
 
+  public void populateDefaultValues() {
+
+    // populate the options from the config
+    final Map<String, OptionValidator> tmp = new HashMap<>();
+    for (OptionValidator validator: VALIDATORS.values()) {
+      try {
+         validator.loadDefault(bootConfig);
+      } catch (ConfigException.Missing e) {
+        throw new IllegalStateException("Config Option is missing"+e+ validator.getOptionName());
```
paul-rogers: `...missing " + validator.getOptionName()`
The exception adds no real value here. Or:
`new IllegalStateException(..., e);`

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/TypeValidators.java
```diff
@@ -217,19 +236,52 @@ public void validate(final OptionValue v, final OptionSet manager) {
...
+    public int computeMaxWidth(double cpuLoadAverage, long maxWidth) {
+      // if maxwidth is already set return it
+      if (maxWidth != 0) {
+        return (int) maxWidth;
+      }
+      // else compute the value and return
+      else {
+        int maxWidthPerNode;
+        int availProc = Runtime.getRuntime().availableProcessors();
+        maxWidthPerNode = (int) Math.max(1, Math.min(availProc, Math.round(availProc * cpuLoadAverage)));
```
paul-rogers: `int maxWidthPerNode = ...`
or even just
`return (int) ...`

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.drill.exec.server.options.OptionValue.Kind;
+import org.apache.drill.exec.server.options.OptionValue.OptionType;
+import org.apache.drill.exec.server.options.OptionValue.OptionScope;
+
+
+/*
+ * To extend the original Option iterator to hide the implementation details and to return the row
+ * which takes precedence over others for an option.This is done by examining the type as the precendence
+ * order is session - system - default.All the values are represented as String instead of having multiple
+ * columns and the datatype is provided for type details.
```
paul-rogers: Maybe provide an example:
foo, bar, mumble
"fred", 10, SESSION
With the "foo" etc. replaced with actual column names...

exec/java-exec/src/main/java/org/apache/drill/exec/util/MemoryAllocationUtilities.java
```diff
@@ -58,7 +58,10 @@ public static void setupBufferedOpsMemoryAllocations(final PhysicalPlan plan, fi
     // if there are any sorts, compute the maximum allocation, and set it on them
     if (bufferedOpList.size() > 0) {
       final OptionManager optionManager = queryContext.getOptions();
-      final long maxWidthPerNode = optionManager.getOption(ExecConstants.MAX_WIDTH_PER_NODE_KEY).num_val;
+      final long maxWidthPerNode;
+      double cpu_load_average = optionManager.getOption(ExecConstants.CPU_LOAD_AVERAGE);
```
paul-rogers: `final double ...` and below.

exec/java-exec/src/main/java/org/apache/drill/exec/util/MemoryAllocationUtilities.java
```diff
@@ -58,7 +58,10 @@ public static void setupBufferedOpsMemoryAllocations(final PhysicalPlan plan, fi
...
+      final long maxWidth = optionManager.getOption(ExecConstants.MAX_WIDTH_PER_NODE);
+      maxWidthPerNode = ExecConstants.MAX_WIDTH_PER_NODE.computeMaxWidth(cpu_load_average,maxWidth);
```
paul-rogers: Move declaration here: `final long maxWidthPerNode = ...`

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -338,3 +338,142 @@ drill.exec: {
 drill.jdbc: {
   batch_queue_throttling_threshold: 100
 }
+
+
```
paul-rogers: Perhaps add a bit more info here:
"Defaults for all system and session options. Provided here for easy customization in Drill distributions in the drill-distrib.conf file. Note: users should not set these values in their drill-override.conf file; users should use ALTER SYSTEM and ALTER SESSION to set the options."

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestClassTransformation.java
```diff
@@ -68,7 +69,7 @@ public void testJaninoClassCompiler() throws Exception {
   @Test
   public void testJDKClassCompiler() throws Exception {
     logger.debug("Testing JDKClassCompiler");
-    sessionOptions.setOption(OptionValue.createString(OptionType.SESSION, ClassCompilerSelector.JAVA_COMPILER_OPTION, ClassCompilerSelector.CompilerPolicy.JDK.name()));
+    sessionOptions.setOption(OptionValue.createString(OptionType.SESSION, ClassCompilerSelector.JAVA_COMPILER_OPTION, ClassCompilerSelector.CompilerPolicy.JDK.name(), OptionScope.SESSION));
```
paul-rogers: Hmmm...

We should not need every caller that sets a session option to say that it is setting a session option. Instead, the implementation of `setOption` for the session option manager class should fill in the `OptionScope.SESSION` value.

exec/java-exec/src/test/java/org/apache/drill/test/TestConfigLinkage.java
```diff
@@ -0,0 +1,77 @@
...
+      String slice_target = client.queryBuilder().sql("SELECT val FROM sys.options2 where name='planner.slice_target'").singletonString();
+      assertEquals(slice_target,"20");
+    }
+  }
+
+  /* Test if config option takes precedence if config option is not set */
+  @Test
+  public void testConfigOption() throws Exception {
+    FixtureBuilder builder = ClusterFixture.builder()
+            .configProperty("drill.exec.options."+ExecConstants.SLICE_TARGET, 30);
```
paul-rogers: Maybe add a new method to the fixture builder:

`setOptionDefault(ExecConstants.SLICE_TARGET, Object value)`

Then, inside this method, append the prefix. For the prefix, refer to a constant defined somewhere for the value.

exec/java-exec/src/test/java/org/apache/drill/test/TestConfigLinkage.java
```diff
@@ -0,0 +1,77 @@
...
+  /* Test if altering system option takes precedence over config option */
+  @Test
+  public void testAlterSystem() throws Exception {
+    try (ClusterFixture cluster = ClusterFixture.standardCluster();
+         ClientFixture client = cluster.clientFixture()) {
+      client.queryBuilder().sql("ALTER SYSTEM SET `planner.affinity_factor` = 1.5").run();
+      String affinity_factor = client.queryBuilder().sql("SELECT val FROM sys.options2 where name='planner.affinity_factor'").singletonString();
+      assertEquals(affinity_factor,"1.5");
+    }
+  }
```
paul-rogers: Additional cases:

* both session and system options set: the session option takes precedence
* Query list of options using the new system table, with options in the four states (see below). Verify that the returned result is correct.
* Various cases for the max width per node. (Will be tricky, you'll have to get the CPU count for use in verifying the values.)

Four states: default, system only, session only, both system and session.

Any other code paths that are not exercised by the above set of tests?

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/FallbackOptionManager.java
```diff
@@ -144,4 +144,27 @@ public OptionList getOptionList() {
...
+  /**
+   * {@link FragmentOptionManager} and {@link SessionOptionManager} use {@link SystemOptionManager} as the fall back
+   * manager so for both FragmentOptionManager and SessionOptionManager fallback is the SystemOptionManager so it is
+   * returned. But in case of {@link QueryOptionManager}, it uses {@link SessionOptionManager} as the fallback manager
+   * and since SessionOptionManager uses SystemOptionManager as fallback, SystemOptionManager can be fetched from the
+   * SessionOptionManager.
+   */
+  public SystemOptionManager getSystemOptionManager() {
+    final SystemOptionManager systemOptionManager;
+    if(fallback instanceof SessionOptionManager) {
```
dvjyothsna: Yes, Correct a System Option Manager is not a fallback for session option manager. FragmentOptionManager and SessionOptionManager use SystemOptionManager as the fall back manager so for both FragmentOptionManager and SessionOptionManager fallback is the System Option Manager so it is returned. But in the case of  QueryOptionManager, it uses Session Option Manager as the fallback manager
 and since Session Option Manager uses System Option Manager as the fallback, we need to go one level deep and fetch System Option Manager from the Session Option Manager.

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestClassTransformation.java
dvjyothsna: Fallback option manager actually implements setOption and session option manager which is a subclass of the fallback option managerdoesn't override that. Moreover, we are creating the option value and passing it. So we need to have a new method to set option in session option manager. Or can we have a variable "type" and set it to OptionType and set it ?

ilooner:
@dvjyothsna @paul-rogers Tim here. When do you guys think this PR will go in? Seems like it didn't make it into this last batch of commits :(. I will be opening another PR soon that is dependent on this going in.

paul-rogers:
From a note sent by @dvjyothsna:

All the system/session option default values are externalized to the conf file. Now the options are looked up in the hierarchical order of SESSION - SYSTEM - CONFIG, the session being given the highest precedence. So, if the user doesn't set system/session option by issuing alter system/session, option manager fetches the value from the config.  Now we need not bother about getting the value from config separately if it is not in the system/session level. For example:

```
OptionValue value = sessionOptions.getOption(JAVA_COMPILER_OPTION);
policy = CompilerPolicy.valueOf((value != null)
    ? value.string_val.toUpperCase()
    : config.getString(JAVA_COMPILER_CONFIG)
```

In the above code snippet we try to get option value and if it is not set (if it is null) we try to get it from Config. With my code changes if the value is not set at system/session level, value from the config is picked automatically.For example, the above can be simplified to:

```
policy = CompilerPolicy.valueOf(sessionOptions.getOption(JAVA_COMPILER_OPTION));
```

I have already externalized all the existing options in ExecConstants, PlannerSettings, etc., 
to the drill-module.conf. Following are the steps to add a new system/session option.

Steps to add a new system/session option:

1. Add a validator in the ExecConstants.java or PlannerSettings.java or whichever is appropriate for the option. For example: 
```
           String NEW_OPTION_KEY = "drill.exec.new_option";
OptionValidator NEW_OPTION= new DoubleValidator(NEW_OPTION_KEY);
```
2. Add the validator from the above step to the  validators list  in `SystemOptionManager.java`.

3. Add the option in the conf file (for example, `drill-module.conf` or `drill-override.conf` file ) under the name space `drill.exec.options`. For  example:
```
	drill.exec.options : {
		drill : {
			exec : {
				new_option : 1.5
			}
		}
	 }
```

4. Set the default value for the option in the conf file as shown above.
 	
5. OptionManager loads default values from the conf file, User can override the default values by issuing `ALTER SYSTEM/SESSION`.

paul-rogers:
@ilooner, @Ben-Zvi, the PR is now in master. Have fun!

PR #869, opened 2017-07-06
DRILL-4511: Add unit test for "Table does not exist" situation in case of empty directory or incorrect table

KulykRoman:
- Added unit test for "Table does not exist" situation in case of empty directory or incorrect table.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
@@ -207,6 +207,34 @@ public class TestParquetMetadataCache extends PlanTestBase {
         .go();
   }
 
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -207,6 +207,31 @@ public void testAbsentPluginOrWorkspaceError() throws Exception {
         .go();
   }
 
+  @Test //DRILL-4511
+  public void testTableDoesNotExistError() throws Exception {
```
arina-ielchiieva: 1. Please split into two tests.
2. You can use to `ExecTest.getTempDir()` to get unique temporary directory.

KulykRoman:
Done.

arina-ielchiieva:
+1

PR #870, opened 2017-07-07
DRILL-5664: Enable security for Drill HiveStoragePlugin based on a co…

sohami:
…nfig parameter

         Change to enable/disable HiveStoragePlugin security configuration based on Drill's "security.storage_plugin.enabled" configuration. This will help to open secure channel between Drill's HiveClient and HiveMetastore

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/StoragePluginRegistryImpl.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/StoragePluginRegistryImpl.java
@@ -130,6 +130,8 @@ public class StoragePluginRegistryImpl implements StoragePluginRegistry {
 
   @SuppressWarnings("resource")
   private Map<String, StoragePlugin> createPlugins() throws DrillbitStartupException {
+    final boolean secureStoragePlugin = context.getConfig().getBoolean(ExecConstants.SECURITY_STORAGE_PLUGIN_ENABLED);
+
     try {
       /*
        * Check if the storage plugins system table has any entries. If not, load the boostrap-storage-plugin file into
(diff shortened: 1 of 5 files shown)


sohami:
More details in JIRA [DRILL-5664](https://issues.apache.org/jira/browse/DRILL-5664).

@parthchandra / @ppadma - Can you please help to review this change ?

ppadma:
I have a question.  The user can change the storage plugin configuration itself instead of enabling/disabling this option.  what is the use of this option ?

sohami:
@ppadma - 
1) For the ease of use. Once the security is being configured for Drill then it's better if all the config changes are done in single place (which is drill-override.conf) and then make sure that all communications are secure. With this option user doesn't have to explicitly go and edit the Storage Plugin configuration to enable security.
2) It will take care of enabling/disabling security both for active and inactive plugins.
3) This can be extended for handling security of all the storage plugin. Hence this config can be a single place to enable/disable security for all the plugins in one go rather than doing it manually for each and every plugin one by one.

exec/java-exec/src/main/java/org/apache/drill/exec/store/StoragePluginRegistryImpl.java
```diff
@@ -260,6 +289,10 @@ public StoragePlugin createOrUpdate(String name, StoragePluginConfig config, boo
 
       if (done) {
         if (persist) {
+          if (!config.isValidSecurityConfig(secureStoragePlugin)) {
+            logger.warn("Security setting for {} plugin is not recommended one. It should be set to: {}",
```
ppadma: This will cause confusion if we let the user update the storage plugin config ignoring the drill security config. I agree with Paul this can be messy, with things getting out of sync. We should discuss about this.

paul-rogers:
What is the status of this PR? Appears that there was substantial discussion about whether the approach here is appropriate for Apache Drill. Had the PR been updated with an alternative? Or, should this PR be closed pending a revised solution?

sohami:
closing this PR pending the revised version.

PR #872, opened 2017-07-11
DRILL-5665: planner.force_2phase_aggr only overrides small inputs

Ben-Zvi:
The planner option planner.force_2phase_aggr was created for testing, but may be used someday by a customer to allow spilling (currently single phase can not spill).
That option was implemented unconditionally, thus potentially violating other checks.

The code change is to only use this option to bypass a single check -- for small input tables.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/AggPruleBase.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/AggPruleBase.java
@@ -61,12 +61,11 @@ public abstract class AggPruleBase extends Prule {
   // currently won't generate a 2 phase plan.
   protected boolean create2PhasePlan(RelOptRuleCall call, DrillAggregateRel aggregate) {
     PlannerSettings settings = PrelUtil.getPlannerSettings(call.getPlanner());
-    if ( settings.isForce2phaseAggr() ) { // for testing - force 2 phase aggr
-      return true;
-    }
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/AggPruleBase.java
```diff
@@ -61,11 +61,9 @@ protected AggPruleBase(RelOptRuleOperand operand, String description) {
   // currently won't generate a 2 phase plan.
   protected boolean create2PhasePlan(RelOptRuleCall call, DrillAggregateRel aggregate) {
     PlannerSettings settings = PrelUtil.getPlannerSettings(call.getPlanner());
-    if ( settings.isForce2phaseAggr() ) { // for testing - force 2 phase aggr
-      return true;
-    }
     RelNode child = call.rel(0).getInputs().get(0);
-    boolean smallInput = child.getRows() < settings.getSliceTarget();
+    boolean smallInput = child.getRows() < settings.getSliceTarget() &&
```
amansinha100: Overloading smallInput makes it confusing.  Why not modify the if condition below to : 
    (... || (smallInput && !settings.isForce2phaseAggr()) 
This way the meaning of smallInput is not changed.

Ben-Zvi: Done

amansinha100:
+1

PR #874, opened 2017-07-12
DRILL-5663: Drillbit fails to start when only keystore path is provided without keystore password.

sindhurirayavaram:
When we configure keystore path without keystore password inside drill-override.conf for WebServer, then Drillbit fails to start. We should explicitly check for either both being present or both being absent. If any one of them is only present then we should throw startup exception for Drill.
In the code change, added checks for both keystorepath and keystorepasword.

### DIFF
--- a/distribution/src/resources/drill-override-example.conf
+++ b/distribution/src/resources/drill-override-example.conf
@@ -214,7 +214,7 @@ drill.exec: {
 }
 
 # Below SSL parameters need to be set for custom transport layer settings.
-javax.net.ssl {
+ssl{
   keyStore: "/keystore.file",
   keyStorePassword: "ks_passwd",
   trustStore: "/truststore.file",
--- a/distribution/src/resources/drill-override-example.conf
+++ b/distribution/src/resources/drill-override-example.conf
@@ -93,6 +93,13 @@ drill.exec: {
       credentials: true
     }
   },
+  # Below SSL parameters need to be set for custom transport layer settings.
+  ssl{
+    keyStorePath: "/keystore.file",
+    keyStorePassword: "ks_passwd",
+    trustStorePath: "/truststore.file",
+    trustStorePassword: "ts_passwd"
+  },
   functions: ["org.apache.drill.expr.fn.impl"],
   network: {
     start: 35000
@@ -213,11 +220,5 @@ drill.exec: {
   default_temporary_workspace: "dfs.tmp"
 }
 
-# Below SSL parameters need to be set for custom transport layer settings.
-ssl{
-  keyStore: "/keystore.file",
-  keyStorePassword: "ks_passwd",
-  trustStore: "/truststore.file",
-  trustStorePassword: "ts_passwd"
-}
+
 
--- a/distribution/src/resources/drill-override-example.conf
+++ b/distribution/src/resources/drill-override-example.conf
@@ -94,7 +94,7 @@ drill.exec: {
     }
   },
   # Below SSL parameters need to be set for custom transport layer settings.
-  ssl{
+  ssl: {
     keyStorePath: "/keystore.file",
     keyStorePassword: "ks_passwd",
     trustStorePath: "/truststore.file",
--- a/distribution/src/resources/drill-override-example.conf
+++ b/distribution/src/resources/drill-override-example.conf
@@ -95,9 +95,13 @@ drill.exec: {
   },
   # Below SSL parameters need to be set for custom transport layer settings.
   ssl: {
+    #If not provided then the default value is java system property javax.net.ssl.keyStore value
     keyStorePath: "/keystore.file",
+    #If not provided then the default value is java system property javax.net.ssl.keyStorePassword value
     keyStorePassword: "ks_passwd",
+    #If not provided then the default value is java system property javax.net.ssl.trustStore value
     trustStorePath: "/truststore.file",
+    #If not provided then the default value is java system property javax.net.ssl.trustStorePassword value
     trustStorePassword: "ts_passwd"
   },
   functions: ["org.apache.drill.expr.fn.impl"],
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -122,10 +122,10 @@ public interface ExecConstants {
   String HTTP_SESSION_MEMORY_RESERVATION = "drill.exec.http.session.memory.reservation";
   String HTTP_SESSION_MEMORY_MAXIMUM = "drill.exec.http.session.memory.maximum";
   String HTTP_SESSION_MAX_IDLE_SECS = "drill.exec.http.session_max_idle_secs";
-  String HTTP_KEYSTORE_PATH = "javax.net.ssl.keyStore";
-  String HTTP_KEYSTORE_PASSWORD = "javax.net.ssl.keyStorePassword";
-  String HTTP_TRUSTSTORE_PATH = "javax.net.ssl.trustStore";
-  String HTTP_TRUSTSTORE_PASSWORD = "javax.net.ssl.trustStorePassword";
+  String HTTP_KEYSTORE_PATH = "ssl.keyStorePath";
+  String HTTP_KEYSTORE_PASSWORD = "ssl.keyStorePassword";
+  String HTTP_TRUSTSTORE_PATH = "ssl.trustStorePath";
+  String HTTP_TRUSTSTORE_PASSWORD = "ssl.trustStorePassword";
   String SYS_STORE_PROVIDER_CLASS = "drill.exec.sys.store.provider.class";
   String SYS_STORE_PROVIDER_LOCAL_PATH = "drill.exec.sys.store.provider.local.path";
   String SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE = "drill.exec.sys.store.provider.local.write";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -122,10 +122,10 @@ public interface ExecConstants {
   String HTTP_SESSION_MEMORY_RESERVATION = "drill.exec.http.session.memory.reservation";
   String HTTP_SESSION_MEMORY_MAXIMUM = "drill.exec.http.session.memory.maximum";
   String HTTP_SESSION_MAX_IDLE_SECS = "drill.exec.http.session_max_idle_secs";
-  String HTTP_KEYSTORE_PATH = "ssl.keyStorePath";
-  String HTTP_KEYSTORE_PASSWORD = "ssl.keyStorePassword";
-  String HTTP_TRUSTSTORE_PATH = "ssl.trustStorePath";
-  String HTTP_TRUSTSTORE_PASSWORD = "ssl.trustStorePassword";
+  String HTTP_KEYSTORE_PATH = "drill.exec.ssl.keyStorePath";
+  String HTTP_KEYSTORE_PASSWORD = "drill.exec.ssl.keyStorePassword";
+  String HTTP_TRUSTSTORE_PATH = "drill.exec.ssl.trustStorePath";
+  String HTTP_TRUSTSTORE_PASSWORD = "drill.exec.ssl.trustStorePassword";
   String SYS_STORE_PROVIDER_CLASS = "drill.exec.sys.store.provider.class";
   String SYS_STORE_PROVIDER_LOCAL_PATH = "drill.exec.sys.store.provider.local.path";
   String SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE = "drill.exec.sys.store.provider.local.write";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec;
+
+import com.typesafe.config.Config;
+import org.apache.drill.common.exceptions.DrillException;
+
+public class SSLConfig {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SSLConfig.class);
+
+  private final String keystorePath;
+
+  private final String keystorePassword;
+
+  private final String truststorePassword;
+
+  private final String truststorePath;
+
+  public boolean isValid = false;
+
+  public SSLConfig(Config config) throws DrillException {
+
+    keystorePath = config.getString(ExecConstants.HTTP_KEYSTORE_PATH).trim();
+
+    keystorePassword = config.getString(ExecConstants.HTTP_KEYSTORE_PASSWORD).trim();
+
+    truststorePath = config.getString(ExecConstants.HTTP_TRUSTSTORE_PATH).trim();
+
+    truststorePassword = config.getString(ExecConstants.HTTP_TRUSTSTORE_PASSWORD).trim();
+
+    /*If keystorePath or keystorePassword is provided in the configuration file use that*/
+    if (!keystorePath.isEmpty() || !keystorePassword.isEmpty()) {
+      if (keystorePath.isEmpty() || keystorePassword.isEmpty()) {
+        throw new DrillException(" *.ssl.keyStorePath and/or *.ssl.keyStorePassword in the configuration file can't be empty.");
+      }
+      isValid = true;
+    }
+  }
+
+  public String getkeystorePath() {
+    return keystorePath;
+  }
+
+  public String getkeystorePassword() {
+    return keystorePassword;
+  }
+
+  public String gettruststorePath() {
+    return truststorePath;
+  }
+
+  public String gettruststorePassword() {
+    return truststorePassword;
+  }
+}
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
@@ -21,17 +21,15 @@ import com.typesafe.config.Config;
 import org.apache.drill.common.exceptions.DrillException;
 
 public class SSLConfig {
-  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SSLConfig.class);
 
   private final String keystorePath;
 
   private final String keystorePassword;
 
-  private final String truststorePassword;
-
   private final String truststorePath;
 
-  public boolean isValid = false;
+  private final String truststorePassword;
+
 
   public SSLConfig(Config config) throws DrillException {
 
@@ -45,26 +43,35 @@ public class SSLConfig {
 
     /*If keystorePath or keystorePassword is provided in the configuration file use that*/
     if (!keystorePath.isEmpty() || !keystorePassword.isEmpty()) {
-      if (keystorePath.isEmpty() || keystorePassword.isEmpty()) {
-        throw new DrillException(" *.ssl.keyStorePath and/or *.ssl.keyStorePassword in the configuration file can't be empty.");
+      if (keystorePath.isEmpty()) {
+        throw new DrillException(" *.ssl.keyStorePath in the configuration file is empty, but *.ssl.keyStorePassword is set");
       }
-      isValid = true;
+      else if (keystorePassword.isEmpty()){
+        throw new DrillException(" *.ssl.keyStorePassword in the configuration file is empty, but *.ssl.keyStorePath is set ");
+      }
+
     }
   }
 
-  public String getkeystorePath() {
+  public boolean isSslValid() { return !keystorePath.isEmpty() && !keystorePassword.isEmpty(); }
+
+  public String getKeyStorePath() {
     return keystorePath;
   }
 
-  public String getkeystorePassword() {
+  public String getKeyStorePassword() {
     return keystorePassword;
   }
 
-  public String gettruststorePath() {
+  public boolean hasTrustStorePath() { return !truststorePath.isEmpty(); }
+
+  public boolean hasTrustStorePassword() {return !truststorePassword.isEmpty(); }
+
+  public String getTrustStorePath() {
     return truststorePath;
   }
 
-  public String gettruststorePassword() {
+  public String getTrustStorePassword() {
     return truststorePassword;
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
@@ -53,25 +53,17 @@ public class SSLConfig {
     }
   }
 
-  public boolean isSslValid() { return !keystorePath.isEmpty() && !keystorePassword.isEmpty(); }
+  public boolean isSslValid() {  return !keystorePath.isEmpty() && !keystorePassword.isEmpty(); }
 
-  public String getKeyStorePath() {
-    return keystorePath;
-  }
+  public String getKeyStorePath() {  return keystorePath; }
 
-  public String getKeyStorePassword() {
-    return keystorePassword;
-  }
+  public String getKeyStorePassword() {  return keystorePassword; }
 
-  public boolean hasTrustStorePath() { return !truststorePath.isEmpty(); }
+  public boolean hasTrustStorePath() {  return !truststorePath.isEmpty(); }
 
-  public boolean hasTrustStorePassword() {return !truststorePassword.isEmpty(); }
+  public boolean hasTrustStorePassword() {  return !truststorePassword.isEmpty(); }
 
-  public String getTrustStorePath() {
-    return truststorePath;
-  }
+  public String getTrustStorePath() {  return truststorePath; }
 
-  public String getTrustStorePassword() {
-    return truststorePassword;
-  }
+  public String getTrustStorePassword() {  return truststorePassword; }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
@@ -146,9 +146,8 @@ public class WebServer implements AutoCloseable {
         serverConnector = createHttpsConnector();
       }
       catch(DrillException e){
-        throw new DrillbitStartupException(e.getMessage(),e);
+        throw new DrillbitStartupException(e.getMessage(), e);
       }
-
     } else {
       serverConnector = createHttpConnector();
     }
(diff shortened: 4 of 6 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -264,19 +265,26 @@ private ServerConnector createHttpsConnector() throws Exception {
...
-      sslContextFactory.setKeyStorePath(config.getString(ExecConstants.HTTP_KEYSTORE_PATH));
-      sslContextFactory.setKeyStorePassword(config.getString(ExecConstants.HTTP_KEYSTORE_PASSWORD));
-
-      // TrustStore and TrustStore password are optional
-      if (config.hasPath(ExecConstants.HTTP_TRUSTSTORE_PATH)) {
-        sslContextFactory.setTrustStorePath(config.getString(ExecConstants.HTTP_TRUSTSTORE_PATH));
-        if (config.hasPath(ExecConstants.HTTP_TRUSTSTORE_PASSWORD)) {
-          sslContextFactory.setTrustStorePassword(config.getString(ExecConstants.HTTP_TRUSTSTORE_PASSWORD));
-        }
+    final boolean hasPath = config.hasPath(ExecConstants.HTTP_KEYSTORE_PATH);
```
paul-rogers: To avoid this double-checking, standard practice is:

* Use a blank value to indicate the value is unset.
* Provide a default (blank) value in `drill-module.conf`

Then, the code can just be:

```
String path = config.getString(ExecConstants.HTTP_KEYSTORE_PATH).trim();
if (! path.isEmpty()) {
  // do stuff
```

Also, if the user is expected to set this key, please provide an example in `drill-override-example.conf`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -264,19 +265,26 @@ private ServerConnector createHttpsConnector() throws Exception {
...
+    final boolean hasPassword = config.hasPath(ExecConstants.HTTP_KEYSTORE_PASSWORD);
+
+    // Check if both keypath and password are present or not
+    if (hasPath && hasPassword) {
+      final String pathValue = config.getString(ExecConstants.HTTP_KEYSTORE_PATH);
+      final String passwordValue = config.getString(ExecConstants.HTTP_KEYSTORE_PASSWORD);
+
+      // checking if any one of them is null or empty
+      if (!Strings.isNullOrEmpty(pathValue) && !Strings.isNullOrEmpty(passwordValue)) {
```
paul-rogers: Path can't be null here, so `! pathValue.isEmpty()` works also.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -264,19 +265,26 @@ private ServerConnector createHttpsConnector() throws Exception {
...
+        sslContextFactory.setKeyStorePath(pathValue);
+        sslContextFactory.setKeyStorePassword(passwordValue);
+      }
+
+      // Throwing an exception if anyone of them is null or empty
+      else {
+        throw new DrillbitStartupException("keystorepath and/or keystorepassword can't be empty.");
```
paul-rogers: Please provide a bit of a clearer message. Also, provide the actual path name. This will be the critical message for folks that made mistake, the Drillbit won't start, and they have to figure out what's what.

Maybe,

```
"To enable web UI security, both " + ExecConstants.HTTP_KEYSTORE_PATH + " and " + ... + " are required.";
```

Also, this is just for the Web UI? In that case, maybe just disable the web UI, but don't fail the Drillbit.

```
"Web UI misconfiguration: web UI is disabled."
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -289,30 +297,19 @@ private ServerConnector createHttpsConnector() throws Exception {
       final DateTime now = DateTime.now();
 
       // Create builder for certificate attributes
-      final X500NameBuilder nameBuilder =
-          new X500NameBuilder(BCStyle.INSTANCE)
-              .addRDN(BCStyle.OU, "Apache Drill (auth-generated)")
-              .addRDN(BCStyle.O, "Apache Software Foundation (auto-generated)")
-              .addRDN(BCStyle.CN, workManager.getContext().getEndpoint().getAddress());
+      final X500NameBuilder nameBuilder = new X500NameBuilder(BCStyle.INSTANCE).addRDN(BCStyle.OU, "Apache Drill (auth-generated)").addRDN(BCStyle.O, "Apache Software Foundation (auto-generated)").addRDN(BCStyle.CN, workManager.getContext().getEndpoint().getAddress());
```
paul-rogers: Can we keep the previous one-argument-per-line, indented style? Much easier to read.

distribution/src/resources/drill-override-example.conf
```diff
@@ -214,7 +214,7 @@ drill.exec: {
 }
 
 # Below SSL parameters need to be set for custom transport layer settings.
-javax.net.ssl {
+ssl{
```
paul-rogers: `ssl: {` under `drill.exec`

distribution/src/resources/drill-override-example.conf
```diff
@@ -214,7 +214,7 @@ drill.exec: {
...
   keyStore: "/keystore.file",
```
paul-rogers: This is an existing property. Generally, a leading slash means an absolute path. So, do we expect the keystore to be in the Linux root directory? This would be very unusual. Or, do we mean "keystore.file" relative to some implied root directory? What is that root?

If we do not have a root, should we look relative to the $DRILL_CONF directory (normally /usr/drill or $DRILL_HOME/conf or given by the --site option.)

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -122,10 +122,10 @@
   String HTTP_SESSION_MEMORY_RESERVATION = "drill.exec.http.session.memory.reservation";
   String HTTP_SESSION_MEMORY_MAXIMUM = "drill.exec.http.session.memory.maximum";
   String HTTP_SESSION_MAX_IDLE_SECS = "drill.exec.http.session_max_idle_secs";
-  String HTTP_KEYSTORE_PATH = "javax.net.ssl.keyStore";
-  String HTTP_KEYSTORE_PASSWORD = "javax.net.ssl.keyStorePassword";
-  String HTTP_TRUSTSTORE_PATH = "javax.net.ssl.trustStore";
-  String HTTP_TRUSTSTORE_PASSWORD = "javax.net.ssl.trustStorePassword";
+  String HTTP_KEYSTORE_PATH = "ssl.keyStorePath";
```
paul-rogers: `drill.exec.ssl.keyStorePath` and below.

exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
```diff
@@ -0,0 +1,70 @@
...
+    keystorePassword = config.getString(ExecConstants.HTTP_KEYSTORE_PASSWORD).trim();
+
+    truststorePath = config.getString(ExecConstants.HTTP_TRUSTSTORE_PATH).trim();
+
+    truststorePassword = config.getString(ExecConstants.HTTP_TRUSTSTORE_PASSWORD).trim();
+
+    /*If keystorePath or keystorePassword is provided in the configuration file use that*/
+    if (!keystorePath.isEmpty() || !keystorePassword.isEmpty()) {
+      if (keystorePath.isEmpty() || keystorePassword.isEmpty()) {
+        throw new DrillException(" *.ssl.keyStorePath and/or *.ssl.keyStorePassword in the configuration file can't be empty.");
```
paul-rogers: For message, use the constant so the user has exact information:

```
...( ExecConstants.HTTP_KEYSTORE_PATH + " and/or " ...
```

Even better, help the user:

* Keystore path is provided, but password is missing.
* Password is provided, but keystore path is not set.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -139,7 +142,13 @@ public void start() throws Exception {
 
     final ServerConnector serverConnector;
     if (config.getBoolean(ExecConstants.HTTP_ENABLE_SSL)) {
-      serverConnector = createHttpsConnector();
+      try {
+        serverConnector = createHttpsConnector();
+      }
+      catch(DrillException e){
+        throw new DrillbitStartupException(e.getMessage(),e.getCause());
```
paul-rogers: `e.getCause()` --> `e` else you'll pass `null` to the startup exception.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -263,18 +272,17 @@ private ServerConnector createHttpsConnector() throws Exception {
     logger.info("Setting up HTTPS connector for web server");
 
     final SslContextFactory sslContextFactory = new SslContextFactory();
+    SSLConfig ssl = new SSLConfig(config);
 
-    if (config.hasPath(ExecConstants.HTTP_KEYSTORE_PATH) &&
-        !Strings.isNullOrEmpty(config.getString(ExecConstants.HTTP_KEYSTORE_PATH))) {
+    if(ssl.isValid){
```
paul-rogers: `if (ssl.isValid) {`

Here, `isValid` means `isSslEnabled`.

Better: `ssl.isEnabled()`

exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private final String keystorePath;
+
+  private final String keystorePassword;
+
+  private final String truststorePassword;
+
+  private final String truststorePath;
+
+  public boolean isValid = false;
```
paul-rogers: `final`

exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
```diff
@@ -0,0 +1,70 @@
...
+    truststorePath = config.getString(ExecConstants.HTTP_TRUSTSTORE_PATH).trim();
+
+    truststorePassword = config.getString(ExecConstants.HTTP_TRUSTSTORE_PASSWORD).trim();
+
+    /*If keystorePath or keystorePassword is provided in the configuration file use that*/
+    if (!keystorePath.isEmpty() || !keystorePassword.isEmpty()) {
+      if (keystorePath.isEmpty() || keystorePassword.isEmpty()) {
+        throw new DrillException(" *.ssl.keyStorePath and/or *.ssl.keyStorePassword in the configuration file can't be empty.");
+      }
+      isValid = true;
```
paul-rogers: Confused by this one. Don't we have three cases?

* Neither property set. No SSL.
* Both properties set, use SSL.
* Incorrectly set, exception, this object is not created.

So `isValid` in the sense of "is the data valid" is unnecessary: the object is not created if invalid. But, "isSslEnabled" is useful. Rather than exposing a member variable, perhaps expose a function `isSslEnabled()`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -263,18 +272,17 @@ private ServerConnector createHttpsConnector() throws Exception {
...
-
-      // TrustStore and TrustStore password are optional
-      if (config.hasPath(ExecConstants.HTTP_TRUSTSTORE_PATH)) {
-        sslContextFactory.setTrustStorePath(config.getString(ExecConstants.HTTP_TRUSTSTORE_PATH));
-        if (config.hasPath(ExecConstants.HTTP_TRUSTSTORE_PASSWORD)) {
-          sslContextFactory.setTrustStorePassword(config.getString(ExecConstants.HTTP_TRUSTSTORE_PASSWORD));
+
+      sslContextFactory.setKeyStorePath(ssl.getkeystorePath());
+      sslContextFactory.setKeyStorePassword(ssl.getkeystorePassword());
+      if(!ssl.gettruststorePath().isEmpty()){
```
paul-rogers: Better: `ssl.hasTrustStore()`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -341,6 +349,7 @@ private ServerConnector createHttpsConnector() throws Exception {
     sslConnector.setPort(config.getInt(ExecConstants.HTTP_PORT));
 
     return sslConnector;
+
```
paul-rogers: Remove blank line.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -338,3 +338,11 @@ drill.exec: {
...
   batch_queue_throttling_threshold: 100
 }
+javax.net.ssl.keyStore = ""
+ssl.keyStorePath = ${?javax.net.ssl.keyStore}
+javax.net.ssl.keyStorePassword = ""
+ssl.keyStorePassword = ${?javax.net.ssl.keyStorePassword}
+javax.net.ssl.trustStore = ""
+ssl.trustStorePath = ${?javax.net.ssl.trustStore}
+javax.net.ssl.trustStorePassword = ""
+ssl.trustStorePassword =  ${?javax.net.ssl.trustStorePassword}
```
paul-rogers: Add a comment about the `javax` versions being legacy and deprecated.

```
javax.net.ssl: {
  keyStorePath = ...
  ...
}
drill.exec.ssl {
  keyStorePath = "",
  ...
}
```

Better, move the `drill.exec` stuff under the existing `drill.exec` section.

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -0,0 +1,89 @@
...
+
+public class TestSSLConfig {
+
+
+  @Test
+  public void firstTest() throws Exception {
+
+    OperatorFixture.OperatorFixtureBuilder builder = OperatorFixture.builder();
+    builder.configBuilder().put(ExecConstants.HTTP_KEYSTORE_PASSWORD, "root");
+    builder.configBuilder().put(ExecConstants.HTTP_KEYSTORE_PATH, "");
```
paul-rogers: Not a big deal, but handy shortcut "fluent" style:

```
   OperatorFixture.OperatorFixtureBuilder builder = OperatorFixture.builder()
      .configBuilder().put(ExecConstants.HTTP_KEYSTORE_PASSWORD, "root")
      .configBuilder().put(ExecConstants.HTTP_KEYSTORE_PATH, "");
```

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -0,0 +1,89 @@
...
+    try (OperatorFixture fixture = builder.build()) {
+      DrillConfig config = fixture.config();
+      try {
+        SSLConfig sslv = new SSLConfig(config);
+        fail();
+      } catch (Exception e) {
```
paul-rogers: `// Expected`

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.Test;
+import static junit.framework.TestCase.fail;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class TestSSLConfig {
+
+
+  @Test
+  public void firstTest() throws Exception {
```
paul-rogers: Good to give these names that suggest the test case: `testMissingKeystorePath` for example.

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -0,0 +1,89 @@
...
+    builder.configBuilder().put(ExecConstants.HTTP_KEYSTORE_PATH, "/root");
+    try (OperatorFixture fixture = builder.build()) {
+      DrillConfig config = fixture.config();
+      SSLConfig sslv = new SSLConfig(config);
+      assertEquals("root", sslv.getkeystorePassword());
+      assertEquals("/root", sslv.getkeystorePath());
+    }
+  }
+}
+
```
paul-rogers: Delete extra lines

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -0,0 +1,89 @@
...
+      } catch (Exception e) {
+        assertTrue(e instanceof DrillException);
+      }
+    }
+  }
+
+  @Test
+  public void thirdTest() throws Exception {
+
+    OperatorFixture.OperatorFixtureBuilder builder = OperatorFixture.builder();
```
paul-rogers: Again, just a nit. Since all you need is the config, you can use the `ConfigBuilder` directly:

```
    DrillConfig config = new ConfigBuilder()
        .put(...)
        .build();
```

In this case, no need for the try/catch block and the operator fixture since we are not using the memory allocator.

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -0,0 +1,89 @@
...
+    builder.configBuilder().put(ExecConstants.HTTP_KEYSTORE_PASSWORD, "root");
+    builder.configBuilder().put(ExecConstants.HTTP_KEYSTORE_PATH, "/root");
+    try (OperatorFixture fixture = builder.build()) {
+      DrillConfig config = fixture.config();
+      SSLConfig sslv = new SSLConfig(config);
+      assertEquals("root", sslv.getkeystorePassword());
+      assertEquals("/root", sslv.getkeystorePath());
+    }
+  }
+}
```
paul-rogers: Do we need to test the trust store options? Are they subject to the same "neither or both" constraints?

exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public String getkeystorePassword() {
+    return keystorePassword;
+  }
+
+  public String gettruststorePath() {
+    return truststorePath;
+  }
+
+  public String gettruststorePassword() {
```
paul-rogers: `getTrustStorePassword` That is, use `functionCase`... Here and above.

exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+package org.apache.drill.exec;
+
+import com.typesafe.config.Config;
+import org.apache.drill.common.exceptions.DrillException;
+
+public class SSLConfig {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SSLConfig.class);
+
+  private final String keystorePath;
```
paul-rogers: We call this `keyStorePath` in the config, (upper-case the "S") so might as well use the same convention here.

exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec;
+
+import com.typesafe.config.Config;
+import org.apache.drill.common.exceptions.DrillException;
+
+public class SSLConfig {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SSLConfig.class);
```
paul-rogers: Not used, so can be deleted.

distribution/src/resources/drill-override-example.conf
sindhurirayavaram: Can you look into this paul? This an example I got. 
SSL certificate keystore location.
Default: conf/default.keystore.
ssl.keystore.path can be configured in 2 ways:
By using the relative path, for example, ssl.keystore.path=conf/default.keystore, where the root directory is <Worklight Root Directory>\server\.
By using the absolute path, for example, ssl.keystore.path=/opt/wl/default.keystore for UNIX, or C:\\wl\\default.keystore for Windows.

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
parthchandra: @paul-rogers  No, truststore path and truststore password are optional. The patch correctly handles the creation of the SSLContext by setting these params only if not empty.

sindhurirayavaram:
Hadoop uses different property names for configuring ssl properties. We can provide backward compatibility where we can still use the java properties in drill-override.conf.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -122,10 +122,10 @@
   String HTTP_SESSION_MEMORY_RESERVATION = "drill.exec.http.session.memory.reservation";
   String HTTP_SESSION_MEMORY_MAXIMUM = "drill.exec.http.session.memory.maximum";
   String HTTP_SESSION_MAX_IDLE_SECS = "drill.exec.http.session_max_idle_secs";
-  String HTTP_KEYSTORE_PATH = "javax.net.ssl.keyStore";
```
sindhurirayavaram: Using the _javax_ property name for Drill property might be confusing for users, since changing that value inside Drill configuration will not reflect on the actual System property. So to avoid that confusion this property was renamed taking example from [Hadoop](https://hadoop.apache.org/docs/r2.7.2/hadoop-mapreduce-client/hadoop-mapreduce-client-core/EncryptedShuffle.html) and [IBM](https://www.ibm.com/support/knowledgecenter/en/SSZH4A_5.0.6/com.ibm.worklight.help.doc/admin/r_ssl_certificate_keystore_setup.html). Though Hadoop provides separate parameter for client/server whereas Drill has only server side config, these properties can be renamed to include _.server_ if needed. But making this name change will not affect Drill users already using the older property since backward compatibility is provided here. It will work seamlessly for them. All the below scenarios are tested. 
1) Setting javax.ssl.* properties in drill-override.conf
2) Setting javax.ssl.* properties as system options with -D option.
3) Setting ssl.* properties in drill-override.conf

So if a user provide the property values only as system property, then both Drill and JSSE will still be consuming it and work as expected.

parthchandra: Hmm, not entirely convinced, but I'm fine as long as you have provided backward compatibility for existing users.

distribution/src/resources/drill-override-example.conf
```diff
@@ -93,6 +93,13 @@ drill.exec: {
       credentials: true
     }
   },
+  # Below SSL parameters need to be set for custom transport layer settings.
+  ssl{
+    keyStore: "/keystore.file",
```
parthchandra: The names in this example file are not consistent with the names uses in ExecConstants. keyStore and trustStore instead of keyStorePath, trustStorePath.

parthchandra:
LGTM. 
Are you going to put the fix from DRILL_5712 into a separate PR?

sindhurirayavaram:
Created a separate pull request for DRILL-5712.

distribution/src/resources/drill-override-example.conf
```diff
@@ -93,6 +93,13 @@ drill.exec: {
       credentials: true
     }
   },
+  # Below SSL parameters need to be set for custom transport layer settings.
+  ssl{
+    keyStorePath: "/keystore.file",
+    keyStorePassword: "ks_passwd",
+    trustStorePath: "/truststore.file",
+    trustStorePassword: "ts_passwd"
```
paul-rogers: ssl{ --> ssl: {

exec/java-exec/src/main/java/org/apache/drill/exec/SSLConfig.java
```diff
@@ -0,0 +1,77 @@
...
+
+  public boolean hasTrustStorePassword() {return !truststorePassword.isEmpty(); }
+
+  public String getTrustStorePath() {
+    return truststorePath;
+  }
+
+  public String getTrustStorePassword() {
+    return truststorePassword;
+  }
```
paul-rogers: Very minor suggestion: these simple returns can also be one-liners like the "has" methods.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -139,7 +142,13 @@ public void start() throws Exception {
 
     final ServerConnector serverConnector;
     if (config.getBoolean(ExecConstants.HTTP_ENABLE_SSL)) {
-      serverConnector = createHttpsConnector();
+      try {
+        serverConnector = createHttpsConnector();
+      }
+      catch(DrillException e){
```
paul-rogers: {code}
catch (DrillbitException e) {
{code}
(note spaces)

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -139,7 +142,13 @@ public void start() throws Exception {
...
+        throw new DrillbitStartupException(e.getMessage(),e);
```
paul-rogers: Space after comma.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -263,18 +272,17 @@ private ServerConnector createHttpsConnector() throws Exception {
     logger.info("Setting up HTTPS connector for web server");
 
     final SslContextFactory sslContextFactory = new SslContextFactory();
+    SSLConfig ssl = new SSLConfig(config);
 
-    if (config.hasPath(ExecConstants.HTTP_KEYSTORE_PATH) &&
-        !Strings.isNullOrEmpty(config.getString(ExecConstants.HTTP_KEYSTORE_PATH))) {
+    if(ssl.isSslValid()){
```
paul-rogers: {code}
if (...) {
{code}

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -52,6 +52,14 @@ drill.client: {
 drill.tmp-dir: "/tmp"
 drill.tmp-dir: ${?DRILL_TMP_DIR}
 
+javax.net.ssl :
+ {
+ keyStore = "",
+ keyStorePassword = "",
+ trustStore = "",
+ trustStorePassword = ""
+ },
```
paul-rogers: {code}
javax.net.ssl: {
  keyStore = "",
{code}

That is, adjust formatting of the header line and indent the values.

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -0,0 +1,108 @@
...
+    ConfigBuilder config = new ConfigBuilder();
+    config.put(ExecConstants.HTTP_TRUSTSTORE_PATH, "/root");
+    config.put(ExecConstants.HTTP_TRUSTSTORE_PASSWORD, "root");
+    SSLConfig sslv = new SSLConfig(config.build());
+    assertEquals(true, sslv.hasTrustStorePath());
+    assertEquals(true,sslv.hasTrustStorePassword());
+    assertEquals("/root",sslv.getTrustStorePath());
+    assertEquals("root",sslv.getTrustStorePassword());
+  }
+}
```
paul-rogers: Thanks for the tests!

Maybe delete the extra lines at the end of the file...

sindhurirayavaram:
Changes made and added another commit!

distribution/src/resources/drill-override-example.conf
```diff
@@ -93,6 +93,13 @@ drill.exec: {
       credentials: true
     }
   },
+  # Below SSL parameters need to be set for custom transport layer settings.
+  ssl: {
+    keyStorePath: "/keystore.file",
+    keyStorePassword: "ks_passwd",
+    trustStorePath: "/truststore.file",
+    trustStorePassword: "ts_passwd"
```
paul-rogers: Comments as suggested in previous comment?

PR #875, opened 2017-07-13
DRILL-5671   Set secure ACLs (Access Control List) for Drill ZK nodes in a secure cluster

bitblender:


exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKSecureACLProvider.java
```diff
@@ -0,0 +1,71 @@
...
+public class ZKSecureACLProvider implements ACLProvider {
+
+    static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ZKSecureACLProvider.class);
+    // Creator has full access
+    static ImmutableList<ACL> DEFAULT_ACL = new ImmutableList.Builder<ACL>()
+                                              .addAll(Ids.CREATOR_ALL_ACL.iterator())
+                                              .build();
+    // Creator has full access
+    // Everyone else has only read access
+    static ImmutableList<ACL> DRILL_CLUSTER_ACL = new ImmutableList.Builder<ACL>()
```
adityakishore: Please use java-doc style comment block and mention why and where this ACL is used.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -112,7 +112,8 @@ drill.exec: {
     retry: {
       count: 7200,
       delay: 500
-    }
+    },
+    use.secure_acl: false
```
adityakishore: This needs to be set to true in `/opt/mapr/drill/drill-1.11.0/conf/drill-distrib.conf` by `configure.sh` when security is enabled on a MapR cluster. Do we have a tracking work item for this?

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKSecureACLProvider.java
bitblender: Done

exec/java-exec/src/main/resources/drill-module.conf
bitblender: Opened https://maprdrill.atlassian.net/browse/MD-2278

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKSecureACLProvider.java
```diff
@@ -0,0 +1,80 @@
...
+        this.clusterName = clusterName;
+        this.drillZkRoot = drillZKRoot;
+        this.drillClusterPath = "/" + this.drillZkRoot + "/" +  this.clusterName ;
+    }
+
+    public List<ACL> getDefaultAcl() {
+        return DEFAULT_ACL;
+    }
+
+    public List<ACL> getAclForPath(String path) {
```
paul-rogers: This approach encodes the meaning of paths in just the path name. Seems fragile.

Alternatives:

* Register the secure (or insecure) paths so that the ZK cluster coordinator (not this class) decides on security.
* Get ACL base on type: `getSecureACL()` or `getPublicACL()`, and let the cluster coordinator define which is which.
* As a refinement of the first idea, provide some kind of config option to externalize the set of paths and their security.

For example, in Drill-on-YARN, the app master creates ZK entries to ensure that only one app master starts per cluster. We would not want to encode that path information here.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -39,6 +39,7 @@
   String ZK_TIMEOUT = "drill.exec.zk.timeout";
   String ZK_ROOT = "drill.exec.zk.root";
   String ZK_REFRESH = "drill.exec.zk.refresh";
+  String ZK_SECURE_ACL = "drill.exec.zk.use.secure_acl";
```
bitblender: I was going to make this "zk.use_secure_acl" but when I was looking for an example I found "bit.use.ip", so I modeled it on that. I will change this to "zk.apply_secure_acl"

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKACLProviderFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.curator.framework.imps.DefaultACLProvider;
+import org.apache.drill.common.config.DrillConfig;
+import org.apache.drill.exec.ExecConstants;
+
+
+public class ZKACLProviderFactory {
+
+    static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ZKACLProviderFactory.class);
+
+    public static ACLProvider getACLProvider(DrillConfig config, String clusterId, String zkRoot) {
```
bitblender: Will make that change.

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKACLProviderFactory.java
```diff
@@ -0,0 +1,44 @@
...
+        if (config.getBoolean(ExecConstants.ZK_SECURE_ACL)) {
+            if (config.getBoolean(ExecConstants.USER_AUTHENTICATION_ENABLED)){
+                logger.trace("ZKACLProviderFactory: Using secure ZK ACL");
+                return new ZKSecureACLProvider(clusterId, zkRoot);
+            } else {
+                logger.warn("ZKACLProviderFactory : Secure ZK ACL enabled but user authentication is disabled." +
```
chunhui-shi: USER_AUTHENTICATION_ENABLED means the drill's client is logging in drill as some authenticated users, but enabling zookeeper ACL means we are going to protect our znodes using ACL. so other user should not be able to modify our znodes, I think this is a valid scenario, meaning no warning needed.

exec/java-exec/src/main/resources/drill-module.conf
paul-rogers: Best not to put vendor-specific comments in the Apache Drill PR...

paul-rogers:
Looks like this patch was applied to an internal branch. Will there be another commit to address the issues raised here for the Apache Drill branch?

bitblender:
Yes. There will be a commit addressing the issues raised here.

paul-rogers:
Where are we on this PR? A note three weeks ago said that another commit would be coming; but haven't see it...

bitblender:
https://github.com/apache/drill/pull/1467 covers the changes required for DRILL-5671. Closing this PR.

PR #876, opened 2017-07-13
DRILL-5659: Fix error code checking in reading from socket

parthchandra:
Check if errorcode is set before checking if the error code is EINTR. Also, check for error first thing after making the read_some or write_some calls.
This fixes the case when we read (or write) only a part of the data in a call to read from or write to the socket.
Tested for the issue on Mac/Centos. Not tested on Windows. Not tested for the case with SASL encryption.


sohami:
+1 LGTM
Thanks a lot for catching and fixing this!. I will test it with SASL encryption case.

superbstreak:
Thanks, @parthchandra!

laurentgo:
I tested non-SASL case and confirm it fixes the issue. Unfortunately I don't have any handy SASL-ready setup for now.

Thanks @parthchandra

parthchandra:
Still haven't been able to confirm the SASL encryption case, but there is no reason why this fix would affect that. Merging it in.

PR #877, opened 2017-07-13
DRILL-5660: Drill 1.10 queries fail due to Parquet Metadata "corrupti…

vdiravka:
…on" from DRILL-3867

- bump up metadata file version to v3_1;
- ignoring of unknown metadata version (for example metadata generated from future versions of drill);
- adding removing "%20" symbols from path.


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1860,81 @@ private static String relativize(String baseDir, String childPath) {
...
+     * @throws IOException if parquet metadata can't be deserialized from the json file
+     */
+    public static boolean isVersionSupported(FileSystem fs, Path path) throws IOException {
+      ObjectMapper mapper = new ObjectMapper();
+      mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+      FSDataInputStream is = fs.open(path);
+
+      MetadataVersion metadataVersion = mapper.readValue(is, MetadataVersion.class);
+      Versions version = Versions.fromString(metadataVersion.textVersion);
+      if (!(version == null)) {
```
arina-ielchiieva: `version != null`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1860,81 @@ private static String relativize(String baseDir, String childPath) {
...
+    /**
+     * @param fs current file system
+     * @param path of metadata cache file
+     * @return true if metadata version is supported, false otherwise
+     * @throws IOException if parquet metadata can't be deserialized from the json file
+     */
+    public static boolean isVersionSupported(FileSystem fs, Path path) throws IOException {
+      ObjectMapper mapper = new ObjectMapper();
+      mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+      FSDataInputStream is = fs.open(path);
```
arina-ielchiieva: How the steam will be closed?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1860,81 @@ private static String relativize(String baseDir, String childPath) {
...
+
+      Versions(String version) {
+        this.version = version;
+      }
+
+      public String getVersion() {
+        return version;
+      }
+
+      public static Versions fromString(String version) {
```
arina-ielchiieva: Please add Java doc.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1860,81 @@ private static String relativize(String baseDir, String childPath) {
...
+   */
+  public static class MetadataVersion {
+    @JsonProperty("metadata_version")
+    public String textVersion;
+
+    /**
+     * Supported metadata versions.
+     * Note: keep them synchronized with {@link ParquetTableMetadataBase} versions
+     */
+    enum Versions {
```
arina-ielchiieva: It would be good to follow Enum naming convention [1]:
1. Enum naming is usually singular.
2. Enum types are usually in upper case.
3. There is no need to use Constants class. Enum types usually represents a fixed set of constants. You can use `getName()` to get String value.

[1] http://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1377,7 +1386,7 @@ public void serialize(ColumnMetadata_v2 value, JsonGenerator jgen, SerializerPro
    *
    * Difference between v3 and v2 : min/max, type_length, precision, scale, repetitionLevel, definitionLevel
    */
-  @JsonTypeName("v3") public static class ParquetTableMetadata_v3 extends ParquetTableMetadataBase {
+  @JsonTypeName("v3_1") public static class ParquetTableMetadata_v3 extends ParquetTableMetadataBase {
```
arina-ielchiieva: Can be replaced with `Version` enum (ex: `Version.V1.getName()`), the same in `ParquetTableMetadataBase` class?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -658,11 +657,21 @@ private boolean tableModified(List<String> directories, Path metaFilePath,
...
+   * Bump up metadata minor version if only metadata content is changed, but metadata structure is the same.
+   *
+   * Note: keep metadata versions synchronized with {@link MetadataVersion.Versions}
+   */
   @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "metadata_version")
   @JsonSubTypes({
-      @JsonSubTypes.Type(value = ParquetTableMetadata_v1.class, name="v1"),
-      @JsonSubTypes.Type(value = ParquetTableMetadata_v2.class, name="v2"),
-      @JsonSubTypes.Type(value = ParquetTableMetadata_v3.class, name="v3")
+      @JsonSubTypes.Type(value = ParquetTableMetadata_v1.class, name = MetadataVersion.Versions.Constants.V1),
```
arina-ielchiieva: Can be replaced with `Version.V1.getName()`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1860,81 @@ private static String relativize(String baseDir, String childPath) {
...
+
+      public static class Constants {
+        public static final String V1 = "v1";
+        public static final String V2 = "v2";
+        public static final String V3 = "v3";
+        public static final String V3_1 = "v3_1";
+      }
+    }
+
+    /**
```
arina-ielchiieva: Please add brief method description.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -557,7 +556,7 @@ private void readBlockMeta(String path,
     mapper.registerModule(serialModule);
     mapper.registerModule(module);
     mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-    FSDataInputStream is = fs.open(p);
+    FSDataInputStream is = fs.open(path);
```
arina-ielchiieva: How the stream will be closed?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1860,81 @@ private static String relativize(String baseDir, String childPath) {
...
+        this.version = version;
+      }
+
+      public String getVersion() {
+        return version;
+      }
+
+      public static Versions fromString(String version) {
+        for (Versions v : Versions.values()) {
+          if (v.version.equalsIgnoreCase(version)) {
```
arina-ielchiieva: Can be replaced with `v.getName().equalsIgnoreCase(version)`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -139,13 +139,13 @@ public static ParquetTableMetadata_v3 getParquetTableMetadata(FileSystem fs,
    * @return
    * @throws IOException
    */
-  public static ParquetTableMetadataBase readBlockMeta(FileSystem fs, String path, MetadataContext metaContext, ParquetFormatConfig formatConfig) throws IOException {
+  public static ParquetTableMetadataBase readBlockMeta(FileSystem fs, Path path, MetadataContext metaContext, ParquetFormatConfig formatConfig) throws IOException {
```
arina-ielchiieva: Please update java doc.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -537,13 +537,12 @@ private void writeFile(ParquetTableMetadataDirs parquetTableMetadataDirs, Path p
    * @return
    * @throws IOException
    */
-  private void readBlockMeta(String path,
+  private void readBlockMeta(Path path,
```
arina-ielchiieva: Please update java doc.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
paul-rogers: Common practice is to enclose this in a try-with-resources block, assuming that this stream derives from `AutoCloseable`. Else, we should add our own `try` block and close the file in `finally`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -658,11 +657,21 @@ private boolean tableModified(List<String> directories, Path metaFilePath,
     return false;
   }
 
+  /**
+   * Basic class for parquet metadata. Inheritors of this class are json serializable structures of
+   * different versions metadata cache files.
+   *
```
paul-rogers: Sorry, nit. This is Javadoc and is HTML formatted. Please insert <p> to separate paragraphs.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
paul-rogers: One very handy thing to do for each version constant is to list what changed, possibly including the JIRA number for more information:

```
/**
 * Version 3.1: Changes the xyz property.
 * File names stored as relative paths.
 * See DRILL-1234.
 */
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1860,81 @@ private static String relativize(String baseDir, String childPath) {
...
+      }
+    }
+
+    /**
+     * @param fs current file system
+     * @param path of metadata cache file
+     * @return true if metadata version is supported, false otherwise
+     * @throws IOException if parquet metadata can't be deserialized from the json file
+     */
+    public static boolean isVersionSupported(FileSystem fs, Path path) throws IOException {
```
paul-rogers: Not now, but in the future, separate the Jackson-serialized persistence class from the code which works with that data. The persistence class just has the fields for the data and simple related operations. Code that works with the data should be in a separate class.

paul-rogers: Here, we can see the reason for the separation. We now open each file twice: once to check the version, another time to deserialize if the version is OK. Better to just deserialize the file. There would be two cases.

* Minor change: the current deserializer can read the file. (This is the case for file version <= code version.) Can also be the case, as here, when the file version bumps without adding new fields.
* Major change: the deserialization fails with a Jackson exception. This tells us we cannot read the file because we don't recognize the format. This should only be the case when file version > code version.

In either case, we can attempt to deserialize the file:

* Deserialize file.
* If error occurs, we don't support the file format.
* If OK, but file version is newer than code version, we don't support.
* If OK, and file version = code version, this is the normal path.
* If OK, but file version < code version, then some special fix-up may be needed to convert the deserialized data to current format.

Given this, we don't really need a zillion version classes. We have one class that handles the logic for the current version. We have a deserializer class that handles the above (including any needed data updates.) And, we have the serialized data class.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -723,19 +723,20 @@ private void init(MetadataContext metaContext) throws IOException {
         // if querying a single file we can look up the metadata directly from the file
         metaPath = new Path(p, Metadata.METADATA_FILENAME);
       }
-      if (metaPath != null && fs.exists(metaPath)) {
+      if (metaPath != null && fs.exists(metaPath) && Metadata.MetadataVersion.isVersionSupported(fs, metaPath)) {
```
paul-rogers: Nit. No need to check `fs.exists` here as we, in effect, check this when trying to open the file. Handle the file not found exception instead.

Note that Drill is a multi-user system. Even with the `exists` check, there is still a race condition where the file can exist during the check, but be deleted by the time we try to open the file. So, since we have to check on open anyway, no need to check here.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -723,19 +723,20 @@ private void init(MetadataContext metaContext) throws IOException {
...
-        parquetTableMetadata = Metadata.readBlockMeta(fs, metaPath.toString(), metaContext, formatConfig);
+        parquetTableMetadata = Metadata.readBlockMeta(fs, metaPath, metaContext, formatConfig);
       } else {
         parquetTableMetadata = Metadata.getParquetTableMetadata(fs, p.toString(), formatConfig);
       }
     } else {
       Path p = Path.getPathWithoutSchemeAndAuthority(new Path(selectionRoot));
       metaPath = new Path(p, Metadata.METADATA_FILENAME);
-      if (fs.isDirectory(new Path(selectionRoot)) && fs.exists(metaPath)) {
+      if (fs.isDirectory(new Path(selectionRoot)) && fs.exists(metaPath)
```
paul-rogers: Can we factor out this code so that we do the checks in only one place? Something like a `boolean loadMetadata(...)` method that does all the checks mentioned above?

In general, repeated code = bad. Somebody has to test this code multiple times if it is repeated. Do the unit tests do so?

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -446,10 +447,25 @@ public void testMetadataCacheAbsolutePaths() throws Exception {
     }
   }
 
+  @Test
```
paul-rogers: Tests for the version checks? A bit tricky because you have to create a file and "predict" the next version.

For this, you need a way to get the current version from your `Version` enum. Then do some simple logic to try:

* Add .1 to version number: 3.1 becomes 3.2.
* Move to next whole version number: 3.1 becomes 4.

Then, create a file with that version and check how this version of the code handles a future version of the file.

To be thorough, also create a file that contains a field that Jackson cannot deserialize. Be sure we properly handle a "hard" version incompatibility as discussed earlier.

paul-rogers:
I wonder if you are very close to solving another long-running problem in Drill: Parquet metadata file corruption (or, at least, the corruption causing queries to fail.)

We have an issue that two Drillbits can write to metadata file concurrently and corrupt it. We also have a case where Drillbit A writes the file while Drillbit B reads it. In both cases, the reading Drillbit can't deserialize the file and "bad things happen."

We need code that can try to deserialize a file, notice that the JSON in the file does not fit our schema, and handle the deserialization exception. If we do that, it is a small step to ignore all forms of read errors: file-not-found, file truncation, etc.

In this case, we give an explicit version error if we can read the file and it has the wrong version. If we can't deserialize the file, we log an error about "file is of a newer version or corrupt."

After this, the corruption can still occur, but the user view is that queries run slow instead of just failing.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
vdiravka: Answered under Paul's comment.

vdiravka:
@paul-rogers could you please review additional commit. The main changes:

1. All IOExceptions are ignored while deserializing metadata cache files (JsonMappingException, JsonParseException...) with appropriate logging. To avoid of reading such corrupted or unsupported file again that status is stored in metadata context.

2. Creating the metadata context while expanding Selection from Metadata Cache (if it is `null`) allows to always detect the status of partition punning.

3. Two new test cases are added: `testCorruptedMetadataFile()` and `testFutureUnsupportedMetadataVersion()`.

4. Test data created in `@BeforeClass` in `TestParquetMetadataCache` will be removed after executing all tests. 

5. For  `testMetadataCacheAbsolutePaths()` and `testSpacesInMetadataCachePath()` tests  `testPlanMatchingPatterns()` is added.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/MetadataContext.java
```diff
@@ -41,6 +41,10 @@
 
   private PruneStatus pruneStatus = PruneStatus.NOT_STARTED;
 
+  // False values of these flags allow to avoid double reading of corrupted or unsupported metadata files
+  public boolean isMetaCacheFileCorrect = true;
+  public boolean isMetaDirsCacheFileCorrect = true;
```
paul-rogers: Nice!

How are these states cleared? Do we keep track of the file timestamp, and try again if the file is replaced/updated? Or, is there some kind of manual reset? Is the state only per query?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -132,23 +135,57 @@ public static ParquetTableMetadata_v3 getParquetTableMetadata(FileSystem fs,
...
+      Metadata metadata = new Metadata(fs, formatConfig);
+      try {
+        metadata.readBlockMeta(path, true, metaContext);
+        return metadata.parquetTableMetadataDirs;
+      } catch (IOException e) {
+        logger.error(e.toString());
+        metaContext.isMetaDirsCacheFileCorrect = false;
+      }
+    }
+    logger.warn("Ignoring corrupted metadata file. Query performance may be slow. Make sure the cache file" +
```
paul-rogers: Maybe move the message to constant so it does not have to be maintained in two places? And, maybe add the file or directory that is corrupt?

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -38,24 +40,31 @@
...
+  private static File dataDir1;
+  private static File dataDir2;
 
 
   @BeforeClass
   public static void copyData() throws Exception {
     // copy the data into the temporary location
     String tmpLocation = getDfsTestTmpSchemaLocation();
-    File dataDir1 = new File(tmpLocation + Path.SEPARATOR + tableName1);
+    dataDir1 = new File(tmpLocation + Path.SEPARATOR + tableName1);
```
arina-ielchiieva: Please replace with `new File (tmpLocation, tableName1)`

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -38,24 +40,31 @@
...
     dataDir1.mkdir();
     FileUtils.copyDirectory(new File(String.format(String.format("%s/multilevel/parquet", TEST_RES_PATH))),
         dataDir1);
 
-    File dataDir2 = new File(tmpLocation + Path.SEPARATOR + tableName2);
+    dataDir2 = new File(tmpLocation + Path.SEPARATOR + tableName2);
```
arina-ielchiieva: The same as above.

exec/java-exec/src/test/java/org/apache/drill/BaseTestQuery.java
```diff
@@ -639,14 +644,18 @@ protected static void copyDirectoryIntoTempSpace(String resourcesDir, String des
...
    */
-  protected static void copyMetaDataCacheToTempReplacingInternalPaths(String srcFileOnClassPath, String destFolderInTmp,
-      String metaFileName) throws IOException {
+  protected static void copyMetaDataCacheToTempWithReplacements(String srcFileOnClassPath,
+      String destFolderInTmp, String metaFileName, String customStringReplacement) throws IOException {
     String metadataFileContents = getFile(srcFileOnClassPath);
     Path rootMeta = new Path(dfsTestTmpSchemaLocation, destFolderInTmp);
     Path newMetaCache = new Path(rootMeta, metaFileName);
     FSDataOutputStream outSteam = fs.create(newMetaCache);
+    if (customStringReplacement!=null) {
```
arina-ielchiieva: `if (customStringReplacement!=null) {` -> `if (customStringReplacement != null) {`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -132,23 +135,57 @@ public static ParquetTableMetadata_v3 getParquetTableMetadata(FileSystem fs,
...
-    return metadata.parquetTableMetadata;
+  public static @Nullable ParquetTableMetadataBase readBlockMeta(FileSystem fs, Path path, MetadataContext metaContext,
+      ParquetFormatConfig formatConfig) {
+    if (metaContext.isMetaCacheFileCorrect) {
+      Metadata metadata = new Metadata(fs, formatConfig);
+      try {
+        metadata.readBlockMeta(path, false, metaContext);
+        return metadata.parquetTableMetadata;
+      } catch (IOException e) {
+        logger.error(e.toString());
```
arina-ielchiieva: `logger.error("Failed to read metadata", e)`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -132,23 +135,57 @@ public static ParquetTableMetadata_v3 getParquetTableMetadata(FileSystem fs,
...
+   */
+  public static @Nullable ParquetTableMetadataDirs readMetadataDirs(FileSystem fs, Path path,
+      MetadataContext metaContext, ParquetFormatConfig formatConfig) {
+    if (metaContext.isMetaDirsCacheFileCorrect) {
+      Metadata metadata = new Metadata(fs, formatConfig);
+      try {
+        metadata.readBlockMeta(path, true, metaContext);
+        return metadata.parquetTableMetadataDirs;
+      } catch (IOException e) {
+        logger.error(e.toString());
```
arina-ielchiieva: The same as above.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1922,73 @@ private static String relativize(String baseDir, String childPath) {
...
+      return relativeFilePath.toUri().getPath();
+    }
+  }
+
+  /**
+   * Supported metadata versions.
+   * <p>
+   * Note: keep them synchronized with {@link ParquetTableMetadataBase} versions
+   */
+  public static class MetadataVersion {
```
arina-ielchiieva: 1. Since this class is used in  other classes, I suggest we factor out as separate class. This class has too many inner classes.
2. It's not an enum, in this case name should be plural - `MetadataVersions`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1922,73 @@ private static String relativize(String baseDir, String childPath) {
...
+     * @return a negative integer, zero, or a positive integer as the
+     *         first argument is less than, equal to, or greater than the
+     *         second.
+     */
+    public static int compare(String metadataVersion1, String metadataVersion2) {
+      if (isVersionSupported(metadataVersion1) && isVersionSupported(metadataVersion2)) {
+        return Integer.compare(SUPPORTED_VERSIONS.indexOf(metadataVersion1), SUPPORTED_VERSIONS.indexOf(metadataVersion2));
+      } else {
+        // this is never reached
+        throw UserException.validationError()
```
arina-ielchiieva: Replace please with `DrillRuntimeException`, it's not user's fault if we try to compare with unsupported version.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -1851,9 +1922,73 @@ private static String relativize(String baseDir, String childPath) {
...
+    /**
+     * Helper compare method similar to {@link java.util.Comparator#compare}
+     *
+     * @param metadataVersion1 the first metadata version to be compared
+     * @param metadataVersion2 the second metadata version to be compared
+     * @return a negative integer, zero, or a positive integer as the
+     *         first argument is less than, equal to, or greater than the
+     *         second.
+     */
+    public static int compare(String metadataVersion1, String metadataVersion2) {
```
arina-ielchiieva: Here you compare two versions based on their index in array list. It's fragile approach. First of all developer should always add version to the end of  `SUPPORTED_VERSIONS` lists but how do you ensure that? At least, note should have been added. Though I think it's better to parse version string representation and compare it.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -132,25 +134,64 @@ public static ParquetTableMetadata_v3 getParquetTableMetadata(FileSystem fs,
...
   }
 
+  /**
+   * Checking whether metadata is corrupted
+   *
+   * @param metaContext metadata context
+   * @param path The path to the metadata file, located in the directory that contains the parquet files
+   * @return true if parquet metadata is corrupted, false otherwise
+   */
+  private static boolean isMetadataFilesCorrupted(MetadataContext metaContext, Path path) {
```
arina-ielchiieva: This method does not only check if metadata files are corrupt, it also warns the user that corrupted metadata files won't be read. Taking into account warn message, it's better to rename method to `ignoreReadingMetadata`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersions.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Supported metadata versions.
+ * <p>
+ * Note: keep them synchronized with {@link Metadata.ParquetTableMetadataBase} versions
+ */
+public class MetadataVersions {
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Metadata.class);
```
arina-ielchiieva: private

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -424,32 +432,121 @@ public void testMoveCache() throws Exception {
...
+      test("use dfs_test.tmp");
+      test("create table `%s` as select * from cp.`tpch/nation.parquet`", unsupportedMetadataVersion);
+      String lastVersion = Iterables.getLast(MetadataVersions.SUPPORTED_VERSIONS);
+      for (String supportedVersion : MetadataVersions.SUPPORTED_VERSIONS) {
+        if (MetadataVersions.compare(lastVersion, supportedVersion) < 0) {
+          lastVersion = supportedVersion;
+        }
+      }
+      // Get the future version, which is absent in MetadataVersions.SUPPORTED_VERSIONS list
+      String futureVersion = "v" + (Integer.parseInt("" + lastVersion.charAt(1)) + 1);
```
arina-ielchiieva: Please user `String.valueOf` instead of concatenating with empty string.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataVersion.java
```diff
@@ -0,0 +1,97 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+
+public class TestParquetMetadataVersion {
+
+  private static String TEST_EXCEPTION_MESSAGE = "Failure is expected while parsing the wrong metadata version";
+  private static String TEST_ASSERT_EQUALS_MESSAGE = "Parquet metadata version is parsed incorrectly";
+  private static String TEST_ASSERT_TRUE_MESSAGE = "Not expected exception is obtained while parsing parquet metadata version";
```
vdiravka: constants are removed. Thanks

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersions.java
```diff
@@ -0,0 +1,115 @@
...
+      Matcher matcher = PATTERN.matcher(metadataVersion);
+      if (!matcher.matches()) {
+        DrillRuntimeException.format("Could not parse metadata version '%s' using format '%s'", metadataVersion, FORMAT);
+      }
+      int majorVersion = Integer.parseInt(matcher.group(1));
+      int minorVersion = 0;
+      if (matcher.group(2) != null) {
+        minorVersion = Integer.parseInt(matcher.group(2));
+      }
+      return new Version(metadataVersion, majorVersion, minorVersion, 0, 0, "");
```
vdiravka: `MetadataVersion` is introduced.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersions.java
```diff
@@ -0,0 +1,115 @@
...
+      // this is never reached
+      throw new DrillRuntimeException(String.format("Unsupported metadata version. '%s' version can't be compared with '%s'",
+          metadataVersion1, metadataVersion2));
+    }
+  }
+
+  /**
+   * Parses a parquet metadata version string
+   */
+  public static class VersionParser {
```
vdiravka: `VersionParser` is deleted. 
`MetadataVersion` is introduced with inner class `Constants`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersions.java
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Version 3.1: Absolute paths of files and directories are replaced with relative ones.<br>
+   * See DRILL-3867
+   */
+  public static final String V3_1 = "v3.1";
+
+  /**
+   * All historical versions of the Drill metadata cache files
+   */
+  public static final List<String> SUPPORTED_VERSIONS = Lists.newArrayList(V1, V2, V3, V3_1);
```
vdiravka: `MetadataVersion` is created.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataVersion.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.drill.common.exceptions.DrillRuntimeException;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+
+public class TestParquetMetadataVersion {
+
+  private static String TEST_EXCEPTION_MESSAGE = "Failure is expected while parsing the wrong metadata version";
```
vdiravka: Constant messages are deleted.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vdiravka: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersion.java
```diff
@@ -0,0 +1,124 @@
...
+import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.Lists;
+import org.apache.drill.common.exceptions.DrillRuntimeException;
+
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+class MetadataVersion implements Comparable<MetadataVersion> {
```
arina-ielchiieva: public

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersion.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Override
+  public int hashCode() {
+    int result = major;
+    result = 31 * result + minor;
+    return result;
+  }
+
+  @Override
```
arina-ielchiieva: Please add Java doc describing how string representation is formed. Basically the part of information you have in class description but it's more relevant here.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataVersion.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @Test(expected = DrillRuntimeException.class)
+  public void testVersionWithoutFirstLetter() throws Exception {
+    String versionWithoutFirstLetter = "3.1";
+    try {
+      new MetadataVersion(versionWithoutFirstLetter);
+    } catch (DrillRuntimeException e) {
+      assertTrue("Not expected exception is obtained while parsing parquet metadata version",
+          e.getMessage().contains("Could not parse metadata"));
+      throw new DrillRuntimeException(e);
```
arina-ielchiieva: 1. `throw e`, no need to wrap it again in `DrillRuntimeException`
2. You can check that error message also contains corrupted version:
`contains(String.format("Could not parse metadata version %s", versionWithoutFirstLetter)));`

Please correct here and in tests below.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersion.java
arina-ielchiieva: `StringBuilder` will look much cleaner here:
``` 
StringBuilder builder = new StringBuilder();
builder.append("v").append(major);
if (minor != 0) {
    builder.append(".").append(minor);
}
return builder.toString();
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersion.java
```diff
@@ -0,0 +1,124 @@
...
+ * Supported metadata versions.
+ * <p>
+ * Note: keep them synchronized with {@link Metadata.ParquetTableMetadataBase} versions
+ */
+  public static class Constants {
+    /**
+     * Version 1: Introduces parquet file metadata caching.<br>
+     * See DRILL-2743
+     */
+    public static final String V1 = "v1";
```
arina-ielchiieva: As I have pointed out in the previous round of code review, we want to remove the burden from the developer on how parquet metadata version string representation should look like when adding new version. For this `MetadataVersion` class was introduced and `toString` method was overridden:
`public static final String V1 = new MetadataVersion(1, 0).toString();`

arina-ielchiieva: As turned out we these constants will be used as annotation attributes thus they MUST have constant expression.
In this case we need to make sure that developer has added correct version before Drill start up. Let's keep list of `MetadataVersion` instead of `String` thus we'll be able to validate version constant.
```
public static final List<String> SUPPORTED_VERSIONS = new ArrayList();

static {
  SUPPORTED_VERSIONS.add(new MetadataVersion(V1));
  ...

}
```

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -495,14 +495,14 @@ public void testFutureUnsupportedMetadataVersion() throws Exception {
...
-        if (new MetadataVersion(lastVersion).compareTo(new MetadataVersion(supportedVersion)) < 0) {
+      MetadataVersion lastVersion = Iterables.getLast(MetadataVersion.Constants.SUPPORTED_VERSIONS);
+      for (MetadataVersion supportedVersion : MetadataVersion.Constants.SUPPORTED_VERSIONS) {
+        if (lastVersion.compareTo(supportedVersion) < 0) {
           lastVersion = supportedVersion;
         }
       }
       // Get the future version, which is absent in MetadataVersion.SUPPORTED_VERSIONS list
-      String futureVersion = "v" + (Integer.parseInt(String.valueOf(lastVersion.charAt(1))) + 1);
+      String futureVersion = "v" + (Integer.parseInt(String.valueOf(lastVersion.toString().charAt(1))) + 1);
```
arina-ielchiieva: `String futureVersion = new MedataVersion(lastVersion.getMajor() + 1, 0).toString();`
Just add getters for major and minor versions to `MedataVersion` class first.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -495,14 +495,14 @@ public void testFutureUnsupportedMetadataVersion() throws Exception {
     try {
       test("use dfs_test.tmp");
       test("create table `%s` as select * from cp.`tpch/nation.parquet`", unsupportedMetadataVersion);
-      String lastVersion = Iterables.getLast(MetadataVersion.Constants.SUPPORTED_VERSIONS);
-      for (String supportedVersion : MetadataVersion.Constants.SUPPORTED_VERSIONS) {
-        if (new MetadataVersion(lastVersion).compareTo(new MetadataVersion(supportedVersion)) < 0) {
+      MetadataVersion lastVersion = Iterables.getLast(MetadataVersion.Constants.SUPPORTED_VERSIONS);
```
arina-ielchiieva: No need to this manually:
```
List<MetadataVersion> supportedVersions = new ArrayList(MetadataVersion.Constants.SUPPORTED_VERSIONS);
Collections.sort(supportedVersions);
MetadataVersion lastVersion = supportedVersions.get(supportedVersions.size() - 1);
```

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -424,32 +432,121 @@ public void testMoveCache() throws Exception {
...
+      String usedMetaPattern = "usedMetadataFile=false"; // ignoring metadata cache file
+      PlanTestBase.testPlanMatchingPatterns(query, new String[]{numFilesPattern, usedMetaPattern},
+          new String[] {"Filter"});
+    } finally {
+      test("drop table if exists %s", unsupportedMetadataVersion);
+    }
+  }
+
+  @Test
+  public void testCorruptedMetadataFile() throws Exception {
```
arina-ielchiieva: Please add tests for the empty metadata file and when at lest one metadata file is missing.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersion.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.drill.common.exceptions.DrillRuntimeException;
+
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+
+public class MetadataVersion implements Comparable<MetadataVersion> {
+
+  private static final String FORMAT = "v((?!0)\\d+)\\.?((?!0)\\d+)?";
```
paul-rogers: To avoid having to parse the "v", might as well just say that the version is a number. Define that we increment the major number (before the decimal point) if the JSON structure changes, we increment the minor number (after the decimal point) of the format is unchanged, but the semantics change (as in the absolute/relative path change.) In general, convert the version to a float and just do normal numeric comparison to determine if the file version is earlier, same as, or later than the code version.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vdiravka: Added `ImmutableSortedSet` in the last commit. It allows do not worry about order of elements.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vdiravka: This code was replaced since float metadata versions are introduced.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vdiravka: Tests were added. 
And two test cases were added for missing cache files: when root metadata cache file is missing and when inner metadata cache files are absent.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/MetadataVersions.java
```diff
@@ -0,0 +1,86 @@
...
+      return stringVersion.charAt(0) == 'v' ? Float.valueOf(stringVersion.substring(1)) : Float.valueOf(stringVersion);
+    } catch (Exception e) {
+      throw new DrillRuntimeException(String.format("Could not parse metadata version '%s'", stringVersion), e);
+    }
+  }
+
+  /**
+   * All historical versions of the Drill metadata cache files
+   */
+  public static final NavigableSet<Float> SUPPORTED_VERSIONS = ImmutableSortedSet.of(
```
arina-ielchiieva: You can use `SortedSet`, more abstract implementation.

arina-ielchiieva:
@vdiravka 
limiting minor version to 9 is breaking the whole concept of major and minor versions usage. Since when we reach minor max we'll need to bump up the major though structure changes did not occur.
If using float is not working for major and minor version (I believe you having trouble comparing 1.1 and 1.10) then let's leave our `MetadataVersion` class, just get rid of `v` in version name.

vdiravka:
@arina-ielchiieva Small fix was made to resolve some regression tests failings. The branch is rebased to the master version.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -268,11 +276,15 @@ private FileSelection expandIfNecessary(FileSelection selection) throws IOExcept
     // use the cacheFileRoot if provided (e.g after partition pruning)
     Path metaFilePath = new Path(cacheFileRoot != null ? cacheFileRoot : selectionRoot, Metadata.METADATA_FILENAME);
     if (!fs.exists(metaFilePath)) { // no metadata cache
+      MetadataContext metaContext = selection.getMetaContext();
```
arina-ielchiieva: It seems you set metadata context, when there can be the cases when it is still null? Also can you please explain why there are so many checks in the code if meta context is null, can we make sure it always exists?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -209,9 +211,10 @@ public ParquetGroupScan( //
     this.entries = Lists.newArrayList();
     if (fileSelection.getMetaContext() != null &&
         (fileSelection.getMetaContext().getPruneStatus() == PruneStatus.NOT_STARTED ||
-          fileSelection.getMetaContext().getPruneStatus() == PruneStatus.NOT_PRUNED)) {
-      // if pruning was not applicable or was attempted and nothing was pruned, initialize the
-      // entries with just the selection root instead of the fully expanded list to reduce overhead.
+          fileSelection.getMetaContext().getPruneStatus() == PruneStatus.NOT_PRUNED ||
+            fileSelection.getMetaContext().isMetadataFilesMissingOrCorrupted)) {
```
arina-ielchiieva: There are 4 checks in `if` statement, may be we can factor out them in separate method?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -519,17 +564,16 @@ private void writeFile(ParquetTableMetadataDirs parquetTableMetadataDirs, Path p
...
-  private void readBlockMeta(String path,
-      boolean dirsOnly,
-      MetadataContext metaContext) throws IOException {
+  private void readBlockMeta(Path path, boolean dirsOnly, MetadataContext metaContext) {
     Stopwatch timer = Stopwatch.createStarted();
-    Path p = new Path(path);
-    Path parentDir = Path.getPathWithoutSchemeAndAuthority(p.getParent()); // parent directory of the metadata file
-    String parentDirString = parentDir.toUri().toString(); // string representation for parent directory of the metadata file
+    Path parentDir = Path.getPathWithoutSchemeAndAuthority(path.getParent()); // parent directory of the metadata file
+    String parentDirString = parentDir.toUri().getPath(); // string representation for parent directory of the metadata file
```
arina-ielchiieva: `parentDirPath`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -1123,4 +1163,22 @@ public GroupScan applyFilter(LogicalExpression filterExpr, UdfUtilities udfUtili
     }
   }
 
+  /**
+   * Ignore expanding selection, if metadata is corrupted
```
arina-ielchiieva: Please add comment that you not only check should we ignore expanding selection but also setting `parquetTableMetadata` and `fileSet` and `selection.revertDirStatuses()`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
```diff
@@ -543,78 +584,72 @@ private void readBlockMeta(String path,
...
-      MetadataContext metaContext)
+  private boolean tableModified(List<String> directories, Path metaFilePath, Path parentDir, MetadataContext metaContext)
       throws IOException {
 
     Stopwatch timer = Stopwatch.createStarted();
 
-    if (metaContext != null) {
-      metaContext.setStatus(parentDir.toString());
-    }
+    metaContext.setStatus(parentDir.toString());
```
arina-ielchiieva: `parentDir.toUri().getPath()`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/MetadataContext.java
```diff
@@ -41,6 +42,8 @@
 
   private PruneStatus pruneStatus = PruneStatus.NOT_STARTED;
 
+  private boolean isMetadataCacheCorrupted;
+
   public MetadataContext() {
```
arina-ielchiieva: 1. We can remove this constructor, it exists by default or init `pruneStatus` and `dirModifCheckMap` in it.
2. Please move enum `PruneStatus` in the end of the class

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/MetadataContext.java
```diff
@@ -41,6 +42,8 @@
 
   private PruneStatus pruneStatus = PruneStatus.NOT_STARTED;
 
+  private boolean isMetadataCacheCorrupted;
```
arina-ielchiieva: `metadataCacheCorrupted`?

arina-ielchiieva:
+1, LGTM.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/MetadataContext.java
vdiravka: Looks better. Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/MetadataContext.java
vdiravka: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/Metadata.java
vdiravka: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetFormatConfig.java
```diff
@@ -24,9 +24,16 @@
 import com.fasterxml.jackson.annotation.JsonTypeName;
 
 @JsonTypeName("parquet") @JsonInclude(JsonInclude.Include.NON_DEFAULT)
-public class ParquetFormatConfig implements FormatPluginConfig{
+public class ParquetFormatConfig implements FormatPluginConfig {
 
-  public boolean autoCorrectCorruptDates = true;
+  private boolean autoCorrectCorruptDates = true;
```
arina-ielchiieva: 4 unit test have failed in `TestCorruptParquetDateCorrection` class.
For Json generation let's leave `autoCorrectCorruptDates` as public and `areCorruptDatesAutoCorrected()` method mark as `@JsonIgnore`.

vdiravka: Fixed.

PR #879, opened 2017-07-18
DRILL-5669: Add a configurable option for minimum memory allocation t…

Ben-Zvi:
…o buffered ops

Added the option "planner.memory.min_memory_per_buffered_op" to enforce a minimum memory allocation to all buffered ops instances (i.e., External Sort or Hash Agg (either phase)).

Default value setting is 40 MB (to cover cases of a large incoming batch, while the query's memory divided among all buffered ops results in a small share per each).

There is a small risk of total allocation by default (i.e. 40MB * num-ops) may exceed the query's total memory; but this happens anyway with "uncontrolled" ops like hash join.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -343,6 +343,15 @@ public interface ExecConstants {
       MAX_QUERY_MEMORY_PER_NODE_KEY, 1024 * 1024, Long.MAX_VALUE, 2 * 1024 * 1024 * 1024L);
 
   /**
+   * Minimum memory alocated to each buffered operator instance.
+   * <p/>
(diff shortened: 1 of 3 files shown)


paul-rogers:
Did we want to increase the default max memory per query from 2 GB to 4 GB as Aman suggested?

Ben-Zvi:
Increasing 2gb --> 4gb may have some visible impact on existing installations, so would need some more testing. This could add a delay to the 1.11 schedule ...

paul-rogers:
Increasing the minimum means each sort uses more memory than before. This means that the query uses more memory than the 2 GB the user has allotted. Drill cannot create new memory, so that memory is coming from somewhere. So, you are right that increasing memory has impact, but we've already incurred that impact by increasing the minimum (that's why we did it.)

So, I'm asking if we should increase the aggregate maximum to reflect that memory use is no longer constrained by the per-query maximum.

PR #880, opened 2017-07-19
DRILL-5678: Undefined behavior due to un-initialized values in Server…

superbstreak:
…MetaContext


parthchandra:
+1. Nice catch.

PR #881, opened 2017-07-20
DRILL-5083: status.getOutcome() return FAILURE if one of the batches …

KulykRoman:
…has STOP status (to avoid infinite loop in Merge Join).


KulykRoman:
@paul-rogers, I retested this situation and I was not able to reproduce infinite loop with this fix (but I can reproduce it without the fix). So this fix solves the issue.

PR #882, opened 2017-07-21
DRILL-4735: ConvertCountToDirectScan rule enhancements

arina-ielchiieva:
1. ConvertCountToDirectScan rule will be applicable for 2 or more COUNT aggregates.
To achieve this DynamicPojoRecordReader was added which accepts any number of columns,
on the contrary with PojoRecordReader which depends on class fields.
AbstractPojoRecordReader class was added to factor out common logic for these two readers.

2. ConvertCountToDirectScan will distinguish between missing, directory and implicit columns.
For missing columns count will be set 0, for implicit to the total records count
since implicit columns are based on files and there is no data without a file.
If directory column will be encountered, rule won't be applied.
CountsCollector class was introduced to encapsulate counts collection logic.

3. MetadataDirectGroupScan class was introduced to indicate to the user when metadata was used
during calculation and for which files it was applied.

Details in Jira [DRILL-4735](https://issues.apache.org/jira/browse/DRILL-4735).


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
```diff
@@ -85,109 +91,231 @@ protected ConvertCountToDirectScan(RelOptRuleOperand rule, String id) {
...
+   * For each aggregate call creates field with "count$" prefix and bigint type.
+   * Constructs record type for created fields.
+   *
+   * @param aggregateRel aggregate relation expression
+   * @return record type
+   */
+  private RelDataType constructDataType(DrillAggregateRel aggregateRel) {
+    List<RelDataTypeField> fields = new ArrayList<>();
+    for (int i = 0; i < aggregateRel.getAggCallList().size(); i++) {
+      RelDataTypeField field = new RelDataTypeFieldImpl("count$" + i, i, aggregateRel.getCluster().getTypeFactory().createSqlType(SqlTypeName.BIGINT));
```
jinfengni: The fieldname in RelDataTypeField "count$0", "count$1" is not very informative. I think we could either get the field name from the aggregator operator, or use "count$colname" to indicate what column's count it represents.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
```diff
@@ -85,109 +91,231 @@ protected ConvertCountToDirectScan(RelOptRuleOperand rule, String id) {
...
+     */
+    boolean collect(DrillAggregateRel agg, DrillScanRel scan, DrillProjectRel project) {
+      return calculateCounts(agg, scan, project) && !containsPartitionColumns();
+    }
+
+    /**
+     * @return list of counts
+     */
+    List<Long> getCounts() {
+      return counts;
```
jinfengni: Minor comment: The API of this class seems to assume user would call collect() and only it return true then call getCounts().  What will happen if user call getCounts() even when collect() return false? Seems user would get some partial information.  Probably return an empty collection in case requirements not meet.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
```diff
@@ -85,109 +91,231 @@ protected ConvertCountToDirectScan(RelOptRuleOperand rule, String id) {
...
+     * @return true if counts were collected, false otherwise
+     */
+    boolean collect(DrillAggregateRel agg, DrillScanRel scan, DrillProjectRel project) {
+      return calculateCounts(agg, scan, project) && !containsPartitionColumns();
+    }
+
+    /**
+     * @return list of counts
+     */
+    List<Long> getCounts() {
```
jinfengni: In stead of returning list of Long, can we return either a map from schemaPath to counts, or list of pair schemaPath, count?  Returning just the lists of counts make it hard to figure out which count corresponds to which aggregate function.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
```diff
@@ -85,109 +91,231 @@ protected ConvertCountToDirectScan(RelOptRuleOperand rule, String id) {
...
+     * </ol>
+     *
+     * @param agg aggregate relational expression
+     * @param scan scan relational expression
+     * @param project project relational expression
+     *
+     * @return true if counts were collected, false otherwise
+     */
+    boolean collect(DrillAggregateRel agg, DrillScanRel scan, DrillProjectRel project) {
+      return calculateCounts(agg, scan, project) && !containsPartitionColumns();
```
jinfengni: If user calls collect(agg, scan, project), the class should clear/reset all the internal collections like "columns", "counts".  Otherwise, what if user uses the same CountsCollector instance, and call collect() twice with different inputs?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
```diff
@@ -85,109 +91,231 @@ protected ConvertCountToDirectScan(RelOptRuleOperand rule, String id) {
...
 
-    PojoRecordReader<CountQueryResult> reader = new PojoRecordReader<CountQueryResult>(CountQueryResult.class,
-        Collections.singleton(res).iterator());
+    /**
+     * Checks if stores list of columns contains file system partition columns.
+     *
+     * @return true if contains partition columns, false otherwise
+     */
+    private boolean containsPartitionColumns() {
+      final ColumnExplorer columnExplorer = new ColumnExplorer(settings.getOptions(), columns);
```
jinfengni: Can ColumnExplorer provide an API to tell if one particular column is the partition column? Seems to me that with such API, we could exit from the for loop in line 238, in stead of first building column list, then call this method to check if the column lists contain a file system partition column.

exec/java-exec/src/main/java/org/apache/drill/exec/store/direct/MetadataDirectGroupScan.java
```diff
@@ -0,0 +1,86 @@
...
+   * <p>
+   * Example: [usedMetadata = true, files = [/tmp/0_0_0.parquet], numFiles = 1]
+   * </p>
+   *
+   * @return string representation of group scan data
+   */
+  @Override
+  public String getDigest() {
+    StringBuilder builder = new StringBuilder();
+    builder.append("usedMetadata = true, ");
```
jinfengni: This "useMetadata=true" seems to be redundant, since it's for MetadataDirectGS.

arina-ielchiieva:
New PR was opened - #900

PR #883, opened 2017-07-24
DRILL-5685: Provide a way to set common environment variable between …

sohami:
…sqlline and Drillbit differently.


sohami:
@paul-rogers - For review

PR #884, opened 2017-07-25
DRILL-5686: Warning for sasl.max_wrapped_size contain incorrect syntax

sohami:


sohami:
@bitblender - Can you please help review this change ?

PR #891, opened 2017-08-02
DRILL-5699: Drill Web UI Page Source Has Links To External Sites

sindhurirayavaram:
When there is no internet, Drill Web UI fails to load the external CDNs and will not load the result page. This change, uses the maven download plugin to download the javascript and css files during build time. First this tries to load the CDNs from the internet. If a failure occurs, it uses the local downloaded Javascript and css files.


arina-ielchiieva:
@sindhurirayavaram, thanks for the PR! I have two questions though:
1. Will Drill build fail if we could not download these the javascript and css files during build time?
2. If the answer to the first question is no than why we try to download the javascript and css files from the internet if we have already loaded them locally when loading the result page?

arina-ielchiieva:
I have checked and build will fail: `Failed to execute goal com.googlecode.maven-download-plugin:download-maven-plugin:1.3.0:wget (add-jquery-as-resource) on project drill-java-exec: IO Error: Could not get content -> [Help 1]`, so I guess we just can refer to static resources.

sindhurirayavaram:
@arina-ielchiieva Thanks for the comments! Yes drillbit fails if it can't download them during build time. These are pretty large fails. Is it good to add them in the source tree? 
To answer your second question! Even if we have local files it's better to check the browser's cache first. If that fails, we can load them locally.

arina-ielchiieva:
> Even if we have local files it's better to check the browser's cache first. If that fails, we can load them locally.
But why? javascript and css files have stable versions. If we download them during build time, they won't change after that.

arina-ielchiieva:
Actually you are right, there are benefits of using CDN first [1].

[1] https://stackoverflow.com/questions/1014203/best-way-to-use-googles-hosted-jquery-but-fall-back-to-my-hosted-library-on-go

arina-ielchiieva:
Regarding build failure, it will fail only if we did not download these files before. If we did, it will take them from the cache:
`[INFO] --- download-maven-plugin:1.3.0:wget (add-datatables-css-as-resource) @ drill-java-exec ---
[INFO] Got from cache: .m2\repository\.cache\download-maven-plugin\dataTables.colVis.css_050ee51b8fd84b79ae0240ce840501c6
`
Basically we need internet only when we build the project for the first time, the next time having internet access is not necessary,  So I guess we are fine here.

+1, LGTM.

arina-ielchiieva:
@sindhurirayavaram, one more question about browser compatibility:
will this work in Chrome, Mozilla, Safari etc?
For example, quick search shows that there are can be issues with different browsers when using `link onerror` [1].

[1] https://stackoverflow.com/questions/30171270/link-onerror-do-not-work-in-ie

sindhurirayavaram:
We are trying to load jquery, datatables libraries here. The jquery library is already available in the resources which generic page is using. We are including generic in the result page. So there is no need to load that again. Also, Colvis has been retired, so it's better to keep it in the resources page. So, I am changing this to include all of the files in the resources folder and checking for external cdn while loading jquery in the generic page.

sindhurirayavaram:
@arina! I am planning to keep all the files in the local resources file! I have mentioned the reasons above. Yes, onload should be used for IE.

arina-ielchiieva:
@sindhurirayavaram I am a little confused with your latest changes. I thought the main purpose of this PR is not to include js and css files into Drill code as was discussed in https://github.com/apache/drill/pull/663.
I thought your first round of changes was correct, though we just needed to make sure, they work in different browsers, so basically I thought that all is left to do is to add `onload` so changes can be run in IE. 
Though maybe I am missing something ...

sindhurirayavaram:
@arina-ielchiieva After the first commit, I had a detailed discussion with @parthchandra. There are two reasons two include all the js and css files locally
1) The biggest library among them, the jquery is already included in the resources file. Even libraries like d3 are there locally.
2) One of the js and css libraries, Colvis by datatables is retired. We have to keep a local copy of that incase the url becomes extinct in the future. Check this please [Colvis](https://datatables.net/extensions/colvis/) 
 So we decided to just include all the js and css libraries locally. I updated the libraries to their min versions and some of them to their latest versions. I made a change in generic page where we are loading google external cdn first and if its failing falling back to the local library. Please let me know if there are any changes needed here. Thank you.

exec/java-exec/src/main/resources/rest/generic.ftl
```diff
@@ -27,7 +27,9 @@
 
       <link href="/static/css/bootstrap.min.css" rel="stylesheet">
 
-      <script src="/static/js/jquery.min.js"></script>
+      <script type="text/javascript" language="javascript" src="//ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js"></script>
```
arina-ielchiieva: `https://ajax.googleapis.com/ajax/libs/jquery/3.2.1/jquery.min.js`?

sindhurirayavaram: According to this [reference](https://stackoverflow.com/questions/550038/is-it-valid-to-replace-http-with-in-a-script-src-http), it shouldn't be a problem. I will change it though.

arina-ielchiieva: Oh, if this valid, please don't change.

arina-ielchiieva:
@parthchandra could you please do final review?

parthchandra:
+1

PR #892, opened 2017-08-02
DRILL-5645: negation of expression causes null pointer exception

josiahyan:
Fixes DRILL-5645, and another test case that got caught up in the change due to type issues.


arina-ielchiieva:
+1, LGTM.

josiahyan:
Yay! First commit!

Thanks for the review!

PR #894, opened 2017-08-03
DRILL-5701: drill.connections.rpc.<user/control/data>.<encrypted/unen…

sohami:
…crypted> metric not behaving correctly

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/security/TestUserBitKerberos.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/security/TestUserBitKerberos.java
@@ -23,7 +23,10 @@ import org.apache.drill.BaseTestQuery;
 import org.apache.drill.common.config.DrillProperties;
 import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.exec.ExecConstants;
+import org.apache.drill.exec.rpc.control.ControlRpcMetrics;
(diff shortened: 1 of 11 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/security/TestUserBitKerberos.java
```diff
@@ -137,6 +142,41 @@ public Void run() throws Exception {
...
+
+    // Run few queries using the new client
+    testBuilder()
+        .sqlQuery("SELECT session_user FROM (SELECT * FROM sys.drillbits LIMIT 1)")
+        .unOrdered()
+        .baselineColumns("session_user")
+        .baselineValues(krbHelper.CLIENT_SHORT_NAME)
+        .go();
+
+    // Check encrypted counters value
```
bitblender: Please explain the logic behind the different count values in the assertion. Same for other tests as well.

sohami: Updated the test description for explaining the logic behind different count values. Also rebase PR on latest apache master. This test will change based on the [PR for DRILL-5721](https://github.com/apache/drill/pull/919) since for local fragment's on Foreman node status update will happen locally and hence there won't be any Control Connection created from a Drillbit to itself. I will update the test based on which commit makes in first.

PR #897, opened 2017-08-07
Drill-5703 Added Syntax Highlighting and Limited Autocomplete on Query Page

cgivre:
This is a clean PR for #893 . This PR replaces the textarea on the query page with the Ace Code editor. (https://ace.c9.io)


parthchandra:
In general, I'm not in favor of including large javascript files into the code base, except where the JS might be getting deprecated or is not freely downloadable from the web. Alternatively, we could download the required js files as part of the build step.
See this commit (https://github.com/apache/drill/pull/891/commits/3779e8337692edcdbf0d2f1a1e0afbf1b6e5f255)

parthchandra:
I also feel we need to fix the pagination of results before we make improvements to the editor.

cgivre:
@parthchandra What do you want to fix in the pagination?  IMHO, the query page and the storage plugin editor both are in dire need of syntax highlighting.  I was going to work on that next.  In the meantime, I can include the JS files in the build.

arina-ielchiieva:
IMHO, Drill Web UI servers more as admin console where query profiles can be seen, storage plugins updated as well as system options etc. We might think of disabling Query page allowing users run queries using REST API, sqlline, JDBC, ODBC only and do not spend efforts on enhancing Query page.

cgivre:
Hi Arina, 
I’d also like to add syntax highlighting to the storage plugin page as well.  IMHO, it’s really easy to make a small mistake, missing a comma or quote or something, and it can be really difficult to find. 

I’d also respectfully disagree that we shouldn’t improve the UI.  IMHO, the Drill UI doesn’t need to be super fancy, but with a few enhancements, it could be made to be much more useful than it currently is and hopefully this could help improve adoption. 

— C

arina-ielchiieva:
Well, I did not mean we should not enhance Web UI. Storage plugin page enhancement would be definitely a huge plus. It's just I have a feeling that we might consider splitting admin part and query running part.

arina-ielchiieva:
@parthchandra I remember once I heard that number of rows sent to Web UI in Drill is limited. From the quick glance I did not see such limitation in code. May I assume it's not true?

parthchandra:
I might be mistaken then. But that would make things worse. The UI will wait to return all the records from a query that could return a million records. 
@cgivre that is what I mean by pagination. A query returning more than a screenful of records should return results page by page, and it should be cancelable. Without that, the query UI is a dangerous place.
Take a look at SQLPad (http://rickbergfalk.github.io/sqlpad/installation-and-administration/) maybe we can integrate that for the query UI?

kkhatua:
@cgivre  I might be digressing, but I'm wondering if the formatter is editable. So, for example, functions or keywords that are supported in Drill specifically could be highlighted correctly, while unknown functions/keywords are marked as an error.

@parthchandra I agree that we need to support pagination, but I think the SQL syntax highlighting needn't be tied with that feature.

cgivre:
@kkhatua I think the formatter is editable.  Truthfully, my main goal is to implement JSON syntax highlighting on the plugin config pages.  I've spent a decent amount of time debugging JSON because of a missed comma or something.

kkhatua:
@cgivre can we close this PR if the #1043 met the objectives?

cgivre:
Yes, it can be closed.  Nice work on the syntax highlighting!

kkhatua:
Thanks, @cgivre ! I don't have permissions to close the PR. Perhaps you can close it instead.

PR #890, opened 2017-07-31
DRILL-5698 Escape version number period separator, this captures version numbers…

darrenbrien:
… with 4 5 or 6 in them, like 1.8.0_144

PR #898, opened 2017-08-07
Fixed bug with java version detection

KrysKruk:
e.g. `java version 1.8.0_144` was failing

pobrane


KrysKruk:
Duplicate of #890

PR #900, opened 2017-08-09
DRILL-4735: ConvertCountToDirectScan rule enhancements

arina-ielchiieva:
1. ConvertCountToDirectScan rule will be applicable for 2 or more COUNT aggregates.
To achieve this DynamicPojoRecordReader was added which accepts any number of columns,
on the contrary with PojoRecordReader which depends on class fields.
AbstractPojoRecordReader class was added to factor out common logic for these two readers.

2. ConvertCountToDirectScan will distinguish between missing, directory and implicit columns.
For missing columns count will be set 0, for implicit to the total records count
since implicit columns are based on files and there is no data without a file.
If directory column will be encountered, rule won't be applied.

3. MetadataDirectGroupScan class was introduced to indicate which files statistics for used.

Details in Jira [DRILL-4735](https://issues.apache.org/jira/browse/DRILL-4735).


arina-ielchiieva:
@jinfengni addressed code review comments. Please review when possible.

jinfengni:
+1

LGTM. Thanks for the PR!

PR #901, opened 2017-08-09
DRILL-5709: Provide a value vector method to convert a vector to nullable

paul-rogers:
Please see DRILL-5709 for an explanation and example.

### DIFF
--- a/exec/vector/src/main/java/org/apache/drill/exec/vector/BaseValueVector.java
+++ b/exec/vector/src/main/java/org/apache/drill/exec/vector/BaseValueVector.java
@@ -19,6 +19,7 @@ package org.apache.drill.exec.vector;
 
 import io.netty.buffer.DrillBuf;
 
+import java.util.Collections;
 import java.util.Iterator;
 
 import com.google.common.base.Preconditions;
@@ -116,7 +117,7 @@ public abstract class BaseValueVector implements ValueVector {
 
   @Override
   public Iterator<ValueVector> iterator() {
-    return Iterators.emptyIterator();
+    return Collections.emptyIterator();
   }
 
   public static boolean checkBufRefs(final ValueVector vv) {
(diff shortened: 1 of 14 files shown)


exec/vector/src/main/java/org/apache/drill/exec/vector/BaseValueVector.java
```diff
@@ -133,5 +134,21 @@ public static boolean checkBufRefs(final ValueVector vv) {
...
+
+  public static void fillBitsVector(UInt1Vector bits, int valueCount) {
+
+    // Create a new bits vector, all values non-null
+
+    bits.allocateNew(valueCount);
+    UInt1Vector.Mutator bitsMutator = bits.getMutator();
+    for (int i = 0; i < valueCount; i++) {
+      bitsMutator.set(i, 1);
+    }
```
Ben-Zvi: This loop may be a bit expensive; Is there a way to mark the whole bitmap with '1's "at once" ?  A la "memset()".  
For example, keep a constant bitmap (of max length - 64K) someplace, and here just dup/copy it whole.

exec/vector/src/main/java/org/apache/drill/exec/vector/BaseValueVector.java
```diff
@@ -133,5 +134,21 @@ public static boolean checkBufRefs(final ValueVector vv) {
...
+  }
```
Ben-Zvi: And need to add:
<code>
bitsMutator.setValueCount(valueCount);
</code>

exec/vector/src/main/java/org/apache/drill/exec/vector/BaseValueVector.java
paul-rogers: Sadly, Netty's `PlatformDependent` provides no equivalent of `memset`. Some tricks we could do, if tests show we have a performance issue:

* Use `setLong()` to write the value 0x01010101010101L every 8 values, with bytes for the last < 8 values.
* Create a heap buffer of some length n, filled with 0x01, and do a copy from that buffer to the direct memory, as many times as needed.

Since both of these will make the code much more complex, let's measure to see the cost of the simple solution before we try them.

Also, at a higher level, we should not even need this trick. A properly-designed value vector hierarchy would treat a non-nullable vector as a nullable vector that always returns `false` for `isNull()`. This is what the new readers do and it vastly simplifies the code.

Sometimes it helps to go back to basics and remember Abstractions 101: well-designed abstractions are our friend.

exec/vector/src/main/java/org/apache/drill/exec/vector/BaseValueVector.java
paul-rogers: Fixed.

arina-ielchiieva:

paul-rogers:
@arina-ielchiieva, conflicts are resolved.

amansinha100:
+1

PR #902, opened 2017-08-10
DRILL-5714: Fix NPE when mapr-db plugin is used in table function

arina-ielchiieva:


arina-ielchiieva:
@paul-rogers, thansk for the code review. 

Here is short description, hope it would be helpful:
The main change is in `MapRDBSubScan` class. The problem was that mapr-db plugin was trying to get format plugin from the default list of format plugins (which is created at Drill start up) but since table function uses custom format it was getting null value. For this case we need to get format plugin from `StoragePluginRegistry` which will create format plugin if it is absent in the list of default ones. For example, the same approach is used in `EasySubScan` class which is used for text plugin.

paul-rogers:
Thanks for the explanation. Looks good.
+1 (again)

PR #903, opened 2017-08-10
DRILL-5712: Update the pom files with dependency exclusions for commons-codec

sindhurirayavaram:
In java-exec, we are adding a dependency for commons-codec of version 1.10. Other dependencies like hadoop-common, parquet-column etc are trying to download different versions for common codec. Exclusions should be added for common-codec in these dependencies.
This change has added exclusions in all the pom.xml files.

### DIFF
--- a/contrib/format-maprdb/pom.xml
+++ b/contrib/format-maprdb/pom.xml
@@ -71,10 +71,22 @@
     <dependency>
       <groupId>com.mapr.hadoop</groupId>
       <artifactId>maprfs</artifactId>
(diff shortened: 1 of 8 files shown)


paul-rogers:
@sindhurirayavaram, please rebase this onto the latest master; looks like some stray commits have found their way into this PR.

paul-rogers:
@sindhurirayavaram, will you be able to rebase this & squash commits? Then I can merge it into master. Thanks!

PR #904, opened 2017-08-11
DRILL-5717: change some date time test cases with specific timezone or Local

weijietong:
Some date time test cases is US Local .This will cause developers from other areas to fail. Fix is to specify the explicit Local to the test cases.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/ExecTest.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/ExecTest.java
@@ -19,11 +19,13 @@ package org.apache.drill.exec;
 
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.io.Files;
+import mockit.Mock;
+import mockit.MockUp;
 import mockit.NonStrictExpectations;
-import org.apache.commons.io.FileUtils;
 import org.antlr.runtime.ANTLRStringStream;
 import org.antlr.runtime.CommonTokenStream;
 import org.antlr.runtime.RecognitionException;
+import org.apache.commons.io.FileUtils;
 import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.common.expression.LogicalExpression;
 import org.apache.drill.common.expression.parser.ExprLexer;
@@ -35,18 +37,20 @@ import org.apache.drill.exec.metrics.DrillMetrics;
 import org.apache.drill.exec.physical.impl.OperatorCreatorRegistry;
 import org.apache.drill.exec.planner.PhysicalPlanReaderTestFactory;
 import org.apache.drill.exec.server.DrillbitContext;
-import org.apache.drill.exec.server.options.OptionValue;
 import org.apache.drill.exec.server.options.SystemOptionManager;
 import org.apache.drill.exec.store.sys.store.provider.LocalPersistentStoreProvider;
 import org.apache.drill.exec.util.GuavaPatcher;
 import org.apache.drill.test.DrillTest;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
+import org.joda.time.DateTimeUtils;
+import org.joda.time.DateTimeZone;
 import org.junit.After;
 import org.junit.BeforeClass;
-
 import java.io.File;
 import java.io.IOException;
+import java.text.DateFormatSymbols;
+import java.util.Locale;
 
 
 public class ExecTest extends DrillTest {
@@ -120,4 +124,29 @@ public class ExecTest extends DrillTest {
     return ret.e;
   }
 
+  /**
+   * This utility is to mock the method DateTimeUtils.getDateFormatSymbols()
+   * to mock the current local as US.
+   */
+  public static void mockUsDateFormatSymbols() {
+    new MockUp<DateTimeUtils>() {
+      @Mock
+      public DateFormatSymbols getDateFormatSymbols(Locale locale) {
+        return new DateFormatSymbols(Locale.US);
+      }
+    };
+  }
+
+  /**
+   * This utility is to mock the method DateTimeZone.getDefault() to
+   * mock current timezone as UTC.
+   */
+  public static void mockUtcDateTimeZone() {
+    new MockUp<DateTimeZone>() {
+      @Mock
+      public DateTimeZone getDefault() {
+        return DateTimeZone.UTC;
+      }
+    };
+  }
 }
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
@@ -19,6 +19,7 @@ package org.apache.drill.exec.fn.impl;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+import mockit.integration.junit4.JMockit;
 import org.apache.drill.BaseTestQuery;
 import org.apache.drill.categories.SqlFunctionTest;
 import org.apache.drill.categories.UnlikelyTest;
@@ -26,11 +27,12 @@ import org.apache.drill.common.util.FileUtils;
 import org.joda.time.DateTime;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
-
+import org.junit.runner.RunWith;
 import java.math.BigDecimal;
 import java.util.List;
 import java.util.Map;
 
+@RunWith(JMockit.class)
 @Category({UnlikelyTest.class, SqlFunctionTest.class})
 public class TestCastFunctions extends BaseTestQuery {
 
@@ -77,16 +79,18 @@ public class TestCastFunctions extends BaseTestQuery {
 
   @Test // DRILL-3769
   public void testToDateForTimeStamp() throws Exception {
-    final String query = "select to_date(to_timestamp(-1)) as col \n" +
-        "from (values(1))";
+    mockUtcDateTimeZone();
+
+    final String query = "select to_date(to_timestamp(-1)) as col \n"
+      + "from (values(1))";
 
     testBuilder()
-        .sqlQuery(query)
-        .ordered()
-        .baselineColumns("col")
-        .baselineValues(new DateTime(1969, 12, 31, 0, 0))
-        .build()
-        .run();
+      .sqlQuery(query)
+      .ordered()
+      .baselineColumns("col")
+      .baselineValues(new DateTime(1969, 12, 31, 0, 0))
+      .build()
+      .run();
   }
 
   @Test
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestDateFunctions.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestDateFunctions.java
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- *
+ * <p>
  * http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -17,11 +17,9 @@
  */
 package org.apache.drill.exec.fn.impl;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
-
-import java.util.List;
-
+import com.google.common.base.Charsets;
+import com.google.common.io.Files;
+import mockit.integration.junit4.JMockit;
 import org.apache.drill.categories.SqlFunctionTest;
 import org.apache.drill.categories.UnlikelyTest;
 import org.apache.drill.common.util.FileUtils;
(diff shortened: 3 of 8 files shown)


vvysotskyi:
Some tests from `TestDateFunctions`, `TestDateConversions` and `TestConstantFolding` classes also fails with some other locales (I checked it for `zh_TW` locale). 
I think they should also be fixed in the border of this PR.
Are there any other tests that depend on the locale?

weijietong:
Yes, thanks for the notice ,have noticed those cases before ,the work is in process.some of that will be part of this PR. Others like sql_to_timestamp function need to discuss whether it should have a Local parameter or not set the specific UTC timezone when get the date.

weijietong:
update points : 

>
* remove commons-codec dependency for 1.4 vs 1.10 conflict
* fix to_date , to_timestamp ,to_time ,sql_to_date ,sql_to_timestamp, sql_to_time function

contrib/format-maprdb/pom.xml
```diff
@@ -75,6 +75,12 @@
     <dependency>
       <groupId>com.mapr.fs</groupId>
       <artifactId>mapr-hbase</artifactId>
+      <exclusions>
+        <exclusion>
+          <artifactId>commons-codec</artifactId>
+          <groupId>commons-codec</groupId>
+        </exclusion>
+      </exclusions>
```
vvysotskyi: I think it would be better to create another Jira and move these changes in the pom files there since it does not connect with the issue described in DRILL-5717.

exec/java-exec/src/main/codegen/templates/DateIntervalFunctionTemplates/DateToCharFunctions.java
```diff
@@ -65,7 +65,7 @@ public void setup() {
         byte[] buf = new byte[right.end - right.start];
         right.buffer.getBytes(right.start, buf, 0, right.end - right.start);
         String input = new String(buf, com.google.common.base.Charsets.UTF_8);
-        format = org.joda.time.format.DateTimeFormat.forPattern(input);
+        format = org.joda.time.format.DateTimeFormat.forPattern(input).withLocale(java.util.Locale.ENGLISH);
```
vvysotskyi: I don't think that it is the right solution. A table may contain a field with date strings which were created with the non-ENGLISH locale so for this case, the query will fail. 

We need to set locale to ENGLISH only for required unit tests.

weijietong:
@vvysotskyi sorry for the late update as have passed a busy week.
update points:
* move out the pom update
* recover the date related templates and update the related test cases by specify specific timezone and Local.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestDateFunctions.java
```diff
@@ -43,6 +46,11 @@
 public class TestDateFunctions extends PopUnitTestBase {
     static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(TestDateFunctions.class);
 
+    @BeforeClass
+    public static void setupLocal() {
+        Locale.setDefault(new Locale("en", "US"));
```
vvysotskyi: This change also affects other unit tests which will be executed after the tests in this class. 
So to avoid this, for each unit test which depends on the locale, we should:
1. preserve current locale
2. change locale to "en"
3. execute test (in the try block)
4. restore locale (in the finally block).

As the example, you may use test testConstantFolding_allTypes() from the class TestConstantFolding below.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewDateFunctions.java
```diff
@@ -51,9 +51,9 @@ public void testIsDate() throws Exception {
         .sqlQuery("select case when isdate(date1) then cast(date1 as date) else null end res1 from " + dateValues)
         .unOrdered()
         .baselineColumns("res1")
-        .baselineValues(new DateTime(Date.valueOf("1900-01-01").getTime()))
-        .baselineValues(new DateTime(Date.valueOf("3500-01-01").getTime()))
-        .baselineValues(new DateTime(Date.valueOf("2000-12-31").getTime()))
+        .baselineValues(new DateTime(1900,1,1,0,0))
+        .baselineValues(new DateTime(3500,1,1,0,0))
+        .baselineValues(new DateTime(2000,12,31,0,0))
```
vvysotskyi: Please add spaces after the comma.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
```diff
@@ -30,6 +31,11 @@
 
 public class TestCastFunctions extends BaseTestQuery {
 
+  @BeforeClass
+  public static void setupLocal() {
+    System.setProperty("user.timezone","Etc/GMT");
```
vvysotskyi: We should change the timezone in the same way as I proposed for the locale.
Also since we fix test failures connected with the timezone, please update Jira and pull request titles.

weijietong:
@vvysotskyi  please review the update ones

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestDateFunctions.java
```diff
@@ -135,11 +139,16 @@ public void testIntervalArithmetic() throws Exception {
 
     @Test
     public void testToChar() throws Exception {
-
-        String expectedResults[] = {(new LocalDate(2008, 2, 23)).toString("yyyy-MMM-dd"),
+        Locale defaultLocale = Locale.getDefault();
+        try{
+          Locale.setDefault(new Locale("en", "US"));
```
vvysotskyi: Please replace `new Locale("en", "US")` by `Locale.US`

weijietong:
@vvysotskyi has updated the mentioned parts. These [4ea36c3](https://github.com/apache/drill/commit/4ea36c3f18841b229aca5048b7d162ea16bfd5a1), [caa8b78](https://github.com/apache/drill/commit/caa8b78c5c31c44e59b5ff1bdf6f1900d14b1a1a) two links has been deeply thought about . I think they can't be solved using the same approach at current PR. They are too old to lost some function definitions like `casttimestamptz` . `VectorUtil.appendVectorAccessibleContent` 's date format does not retain time zone part.  So I suggest they should be kept there or removed.

weijietong:
@vvysotskyi thanks for your patient review . As you pointed out  , the timezone would not take effect when test cases run in parallel. I also tried the mock strategy. It will also fail in parallel. So I moved out that part of code to a new class called `TestCastFunctionsSpecTZCases` by specifying the timezone before the cluster starts up. It would be clear to the codes.

paul-rogers:
@vvysotskyi, can you take a look at the revisions and see if it looks OK?

vvysotskyi:
@weijietong which problems did you have when tried to mock `System.getProperty()` method?
As the example of nice mock of this method, you may use [this test](https://github.com/apache/drill/pull/936/files#diff-b053496903bff0dc154e42604b93b9efR60). Also besides the timezone, property with locale should be mocked.

weijietong:
@vvysotskyi  the problem with mocking strategy is that when I run the test case just with the single method `testToDateForTimeStamp` it will pass. but when I run the whole test cases of the `TestCastFunctions ` , the included `testToDateForTimeStamp` method will fail.

vvysotskyi:
@weijietong thanks for the explanation of your problem. I was able to reproduce it, but also I found working solution. This mock works correctly. The problem appears when unit test is run with other tests: [DateTimeZone](https://github.com/JodaOrg/joda-time/blob/ba95735daf79d00ce0928f30701d691f5e029d81/src/main/java/org/joda/time/DateTimeZone.java) class contains static field [cDefault](https://github.com/JodaOrg/joda-time/blob/ba95735daf79d00ce0928f30701d691f5e029d81/src/main/java/org/joda/time/DateTimeZone.java#L128) which is used to receive timezone in `testToDateForTimeStamp()` and in other tests. When timezone does not set to this field, but method [getDefault()](https://github.com/JodaOrg/joda-time/blob/ba95735daf79d00ce0928f30701d691f5e029d81/src/main/java/org/joda/time/DateTimeZone.java#L149) is called, value is taken from `System.getProperty()`. 
When the first test that calls this method does not have a mock, it sets a real value of timezone from `System.getProperty()`. Therefore our test uses a unmocked value from `cDefault`. 

So lets mock `DateTimeZone.getDefault()` method:
```
    new MockUp<DateTimeZone>() {
      @Mock
      public DateTimeZone getDefault() {
        return DateTimeZone.UTC;
      }
    };
```

weijietong:
@vvysotskyi thanks for your hard work, I have rewritten the corresponding method. But I have not rewritten the other Locate related cases. Since mockup is a white box testing , I payed some tedious work to explore the possible invocation path, maybe the same as you. I think current Local solution is also acceptable.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestDateFunctions.java
```diff
@@ -34,19 +33,22 @@
...
-
 import com.google.common.base.Charsets;
 import com.google.common.io.Files;
 import org.junit.experimental.categories.Category;
 
+import java.util.Locale;
+
+
 @Category({UnlikelyTest.class, SqlFunctionTest.class})
 public class TestDateFunctions extends PopUnitTestBase {
```
vvysotskyi: This class has 4 space indentation, but according to our convention, it should be with 2 spaces. Since you have made changes here, could you please make 2 spaces indentation?

logical/src/test/java/org/apache/drill/common/expression/fn/JodaDateValidatorTest.java
```diff
@@ -24,6 +24,7 @@
 import org.junit.BeforeClass;
 import org.junit.Test;
 
+import java.util.Locale;
```
vvysotskyi: Could you please revert changes, made in this class to the version, where you just added `.withLocale(Locale.US)` in `parseDateFromPostgres` method. My earlier comment was connected only with change, made in UDF function, but here, in the method that just used for unit tests, we can do that.

weijietong:
@vvysotskyi  thanks for your help! have done, please review.

weijietong:
@vvysotskyi any more advice ?

exec/java-exec/src/test/java/org/apache/drill/exec/fn/interp/TestConstantFolding.java
```diff
@@ -117,6 +123,13 @@ public void createFiles(int smallFileLines, int bigFileLines) throws Exception{
 
   @Test
   public void testConstantFolding_allTypes() throws Exception {
+    new MockUp<DateTimeUtils>() {
```
vvysotskyi: I meant to use the same method in all tests where Locale should be mocked. Please replace it.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/testing/TestDateConversions.java
```diff
@@ -225,4 +243,16 @@ public void testPostgresDateFormatError() throws Exception {
       throw e;
     }
   }
+
+  /**
+   * mock current locale to US
+   */
+  private void mockUSLocale() {
```
vvysotskyi: Please move this method into ExecTest class and make it public and static.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
```diff
@@ -77,16 +82,23 @@ public void testCastByConstantFolding() throws Exception {
 
   @Test // DRILL-3769
   public void testToDateForTimeStamp() throws Exception {
-    final String query = "select to_date(to_timestamp(-1)) as col \n" +
-        "from (values(1))";
+    new MockUp<DateTimeZone>() {
```
vvysotskyi: This mock is used twice. So let's also move the code into the separate method.

weijietong:
applied the review comments

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestDateFunctions.java
```diff
@@ -33,128 +33,140 @@
...
+                                "PT8647.200S",
+                                "P6M19DT86399.999S",
+                                "PT1715.714S"};
+
+    testCommon(expectedResults, "/functions/date/interval_arithmetic.json", "/test_simple_date.json");
+  }
+
+  @Test
+  public void testToChar() throws Exception {
+    new MockUp<DateTimeUtils>() {
```
vvysotskyi: Please replace this mock by method

exec/java-exec/src/test/java/org/apache/drill/exec/ExecTest.java
```diff
@@ -120,4 +124,27 @@ protected LogicalExpression parseExpr(String expr) throws RecognitionException {
...
+      public DateFormatSymbols getDateFormatSymbols(Locale locale) {
+        return new DateFormatSymbols(Locale.US);
+      }
+    };
+  }
+
+  /**
+   * mock current timezone to UTC
+   */
+  public static void mockUTCTimezone() {
```
vvysotskyi: Please rename this method to mockUtcDateTimeZone

exec/java-exec/src/test/java/org/apache/drill/exec/ExecTest.java
```diff
@@ -120,4 +124,27 @@ protected LogicalExpression parseExpr(String expr) throws RecognitionException {
     return ret.e;
   }
 
+  /**
+   * mock current locale to US
+   */
+  public static void mockUSLocale() {
```
vvysotskyi: Please rename this method to mockUsDateFormatSymbols

exec/java-exec/src/test/java/org/apache/drill/exec/ExecTest.java
```diff
@@ -120,4 +124,27 @@ protected LogicalExpression parseExpr(String expr) throws RecognitionException {
     return ret.e;
   }
 
+  /**
+   * mock current locale to US
```
vvysotskyi: Please start the sentence with a capital letter and add a description about the method which is mocked.

exec/java-exec/src/test/java/org/apache/drill/exec/ExecTest.java
```diff
@@ -120,4 +124,27 @@ protected LogicalExpression parseExpr(String expr) throws RecognitionException {
...
+    new MockUp<DateTimeUtils>() {
+      @Mock
+      public DateFormatSymbols getDateFormatSymbols(Locale locale) {
+        return new DateFormatSymbols(Locale.US);
+      }
+    };
+  }
+
+  /**
+   * mock current timezone to UTC
```
vvysotskyi: Please start the sentence with a capital letter and add a description about the method which is mocked.

weijietong:
done

vvysotskyi:
@weijietong, thanks for the pull request, +1

arina-ielchiieva:
+1, LGTM.

PR #906, opened 2017-08-11
DRILL-5546: Handle schema change exception failure caused by empty in…

jinfengni:
…put or empty batche.

1. Modify ScanBatch's logic when it iterates list of RecordReader.
   1) Skip RecordReader if it returns 0 row && present same schema. A new schema (by calling Mutator.isNewSchema() ) means either a new top level field is added, or a field in a nested field is added, or an existing field type is changed.
   2) Implicit columns are added and populated only when the input is not empty, i.e. the batch contains > 0 row or rowCount == 0 && new schema.
   3) ScanBatch will return NONE directly (called as "fast NONE"), if all its RecordReaders haver empty input and thus are skipped, in stead of returing OK_NEW_SCHEMA first.

2. Modify IteratorValidatorBatchIterator to allow
   1) fast NONE ( before seeing a OK_NEW_SCHEMA)
   2) batch with empty list of columns.

2. Modify JsonRecordReader when it get 0 row. Do not insert a nullable-int column for 0 row input. Together with ScanBatch, Drill will skip empty json files.

3. Modify binary operators such as join, union to handle fast none for either one side or both sides. Abstract the logic in AbstractBinaryRecordBatch, except for MergeJoin as its implementation is quite different from others.

4. Fix and refactor union all operator.
  1) Correct union operator hanndling 0 input rows. Previously, it will ignore inputs with 0 row and put nullable-int into output schema, which causes various of schema change issue in down-stream operator. The new behavior is to take schema with 0 into account
  in determining the output schema, in the same way with > 0 input rows. By doing that, we ensure Union operator will not behave like a schema-lossy operator.
  2) Add a UnionInputIterator to simplify the logic to iterate the left/right inputs, removing significant chunk of duplicate codes in previous implementation.
  The new union all operator reduces the code size into half, comparing the old one.

5. Introduce UntypedNullVector to handle convertFromJson() function, when the input batch contains 0 row.
  Problem: The function convertFromJSon() is different from other regular functions in that it only knows the output schema after evaluation is performed. When input has 0 row, Drill essentially does not have
  a way to know the output type, and previously will assume Map type. That works under the assumption other operators like Union would ignore batch with 0 row, which is no longer
  the case in the current implementation.
  Solution: Use MinorType.NULL at the output type for convertFromJSON() when input contains 0 row. The new UntypedNullVector is used to represent a column with MinorType.NULL.

6. HBaseGroupScan convert star column into list of row_key and column family. HBaseRecordReader should reject column star since it expectes star has been converted somewhere else.
  In HBase a column family always has map type, and a non-rowkey column always has nullable varbinary type, this ensures that HBaseRecordReader across different HBase regions will have the same top level schema, even if the region is
  empty or prune all the rows due to filter pushdown optimization. In other words, we will not see different top level schema from different HBaseRecordReader for the same table.
  However, such change will not be able to handle hard schema change : c1 exists in cf1 in one region, but not in another region. Further work is required to handle hard schema change.

7. Modify scan cost estimation when the query involves * column. This is to remove the planning randomness since previously two different operators could have same cost.

8. Add a new flag 'outputProj' to Project operator, to indicate if Project is for the query's final output. Such Project is added by TopProjectVisitor, to handle fast NONE when all the inputs to the query are empty
and are skipped.
  1) column star is replaced with empty list
  2) regular column reference is replaced with nullable-int column
  3) An expression will go through ExpressionTreeMaterializer, and use the type of materialized expression as the output type
  4) Return an OK_NEW_SCHEMA with the schema using the above logic, then return a NONE to down-stream operator.

9. Add unit test to test operators handling empty input.

10. Add unit test to test query when inputs are all empty.


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -152,97 +157,75 @@ public void kill(boolean sendUpstream) {
...
-            // We're on the last reader, and it has no (more) rows.
-            currentReader.close();
-            releaseAssets();
-            done = true;  // have any future call to next() return NONE
-
-            if (mutator.isNewSchema()) {
-              // This last reader has a new schema (e.g., we have a zero-row
-              // file or other source).  (Note that some sources have a non-
-              // null/non-trivial schema even when there are no rows.)
+          injector.injectChecked(context.getExecutionControls(), "next-allocate", OutOfMemoryException.class);
```
paul-rogers: From the PR description:

> 1. Skip RecordReader if it returns 0 row && present same schema. A new schema (by calling Mutator.isNewSchema() ) means either a new top level field is added, or a field in a nested field is added, or an existing field type is changed.

The code, however, adds an additional condition: if implicit fields change. (But, as noted below, that should never occur in practice.)

What happens on the first reader? There is no schema, so *any* schema is a new schema. Suppose the file is JSON and the schema is built on the fly. Does the code handle the case that we have no schema (first reader), and that reader adds no columns?

Or, according to the logic that the downstream wants to know the schema, even if there are no records, do we send an empty schema (schema with no columns) downstream, because that is an accurate representation of an empty JSON file?

What happens in the case of an empty JSON file following a non-empty file? In this case, do we consider the empty schema as a schema change relative to a non-empty change?

In short, can we generalize this first rule a bit?

> 2. Implicit columns are added and populated only when the input is not empty, i.e. the batch contains > 0 row or rowCount == 0 && new schema.

How does this interact with a scan batch that has only one file, and that file is empty? Would we return the empty schema downstream? With the implicit columns?

> 3. ScanBatch will return NONE directly (called as "fast NONE"), if all its RecordReaders haver empty input and thus are skipped, in stead of returing OK_NEW_SCHEMA first.


This is just a bit ambiguous. If the reader is JSON, then an empty file has an empty schema (for reasons cited above.)

But, if the input is CSV, then we *always* have a schema. If the file has column headers, then we know that the schema is, say, (a, b, c) because those are the headers. Or, if the file has no headers, the schema is always the `columns` array. So, should we send that schema downstream? If so, should it include the implicit columns?

This, in fact, raises another issue (out of scope for this PR): if we return an empty batch with non-empty schema, we have no place to attach the implicit columns that will allow the user to figure out that, say, "foo.csv" is empty.

On the other hand, if we say that an empty CSV file has no schema, then we can skip that file. The same might be true of JSON. What about Parquet? We'd have a schema even if there are no rows. Same with JDBC. Should we return this schema, even if the data is empty?

Finally, do we need special handling for "null" files: a file that no longer exists on disk and so has a completely undefined schema? An undefined schema is different than an empty schema. Undefined = "we just don't know what the schema might be." Empty = "we know there is a schema and that schema is empty." A missing CSV file has an undefined schema. We could argue that a 0-length (or, a one-byte file with the only content being a newline) is a valid file, but it is a file with no columns, hence an empty schema.

Or, should we have a rule that says something like:

1. If a file is missing, or has an undefined schema, skip it.
2. If a file has a schema (empty or not) and no rows, then:
  * If this is the first file, remember the schema and skip to the next.
  * If this file follows a non-empty file, trigger a schema change.
4. If a file has a rows, then:
  * If the previous schema was empty, ignore that empty schema.
  * If the previous schema was non-empty, and different, return OK_NEW_SCHEMA
  * If the previous schema was the same, return OK.
5. At EOF:
  * If the previous schema was empty, and the previous file had no rows, return that schema.
  * If the previous schema was empty, or no non-empty files, return NONE.

We can reduce these to a simpler set of rules:

* We ignore all empty or missing schemas.
* We ignore empty files unless they are the only file(s) or they have a schema different than the next non-empty file.

Other rules are possible. One is simply to ignore all 0-length files regardless of whether the schema is the same or different than surrounding files. If all files return 0 records, just return NONE. (These are the rules adopted, as of now, by the new scan operator; though we will change them to match the rules here one we pin them down.)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+      final ValueVector vv = container.addOrGet(outputField, callBack);
+      allocationVectors.add(vv);
+
+    }
+
+    doAlloc(0);
+    container.buildSchema(SelectionVectorMode.NONE);
+    wasNone = true;
+    return IterOutcome.OK_NEW_SCHEMA;
+  }
```
paul-rogers: The above issues aside, a "mental execution" of the "fast none" code path suggests it should work.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/unit/TestEmptyBatchMiniPlan.java
```diff
@@ -0,0 +1,453 @@
...
+
+    new MiniPlanTestBuilder()
+        .root(batch)
+        .expectSchema(expectedSchema)
+        .expectZeroRow(true)
+        .go();
+  }
+
+  @Test
+  public void testOutputProjectEmpty() throws Exception {
```
paul-rogers: This test does a great job of handling a wide variety of expressions.

But, looking at the code under test, it does not exercise the following code paths:

* `SELECT *`
* `SELECT foo.*, bar.*`
* Error in the expression tree materializer
* Map column references: `myMap.innerMap.foo`
* Array references: `columns[5]`

exec/java-exec/src/main/java/org/apache/drill/exec/physical/config/Project.java
```diff
@@ -35,9 +35,19 @@
...
 
   private final List<NamedExpression> exprs;
+  /**
+   * {@link org.apache.drill.exec.planner.physical.ProjectPrel for the meaning of flag 'outputProj'}
+   */
+  private boolean outputProj = false;
 
   @JsonCreator
-  public Project(@JsonProperty("exprs") List<NamedExpression> exprs, @JsonProperty("child") PhysicalOperator child) {
+  public Project(@JsonProperty("exprs") List<NamedExpression> exprs, @JsonProperty("child") PhysicalOperator child, @JsonProperty("outputproj") boolean outputProj) {
```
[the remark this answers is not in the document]
jinfengni: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/config/Project.java
```diff
@@ -46,14 +56,18 @@ public Project(@JsonProperty("exprs") List<NamedExpression> exprs, @JsonProperty
     return exprs;
   }
 
+  public boolean isOutputProj() {
```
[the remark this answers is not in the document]
jinfengni: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -152,97 +157,75 @@ public void kill(boolean sendUpstream) {
...
-      try {
-        injector.injectChecked(context.getExecutionControls(), "next-allocate", OutOfMemoryException.class);
-
-        currentReader.allocate(mutator.fieldVectorMap());
-      } catch (OutOfMemoryException e) {
-        clearFieldVectorMap();
-        throw UserException.memoryError(e).build(logger);
-      }
-      while ((recordCount = currentReader.next()) == 0) {
+      while (true) {
```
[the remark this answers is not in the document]
jinfengni: It's not entirely true that the original logic allocates a batch once per call to next().  There are two places where allocate were called. The first one happens outside of the while loop [1], while the second happens inside the loop for the case where we want to skip certain record readers [2].

The new logic essentially combine the two into one, by putting the first one inside the loop as well. In that sense, it's not doing a completely job than the previous logic.

Regarding where the buffers are allocated if there are two passes,  each value vector releases its own buffer by calling clear(), before it allocate new buffer.  That's why the previous/new logic did not run into memory leak problem. 
 

1. https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java#L175
2. https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java#L214

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
-    for (ValueVector v : allocationVectors) {
-      try {
-        AllocationHelper.allocateNew(v, current.getRecordCount());
-      } catch (OutOfMemoryException ex) {
-        return false;
-      }
+  @SuppressWarnings("resource")
+  private IterOutcome doWork(RecordBatch inputBatch, boolean newSchema) throws ClassTransformationException, IOException, SchemaChangeException {
+    if (inputBatch.getSchema().getFieldCount() != container.getSchema().getFieldCount()) {
+      // wrong.
```
paul-rogers: If it is wrong, do we want to do something? Throw an `IllegalStateException`?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
-            .setMode(DataMode.OPTIONAL)
-            .build();
-
-        MaterializedField outputField = MaterializedField.create(colName, majorType);
-        ValueVector vv = container.addOrGet(outputField, callBack);
-        allocationVectors.add(vv);
-      }
+    for (final ValueVector v : this.allocationVectors) {
+      AllocationHelper.allocateNew(v, inputBatch.getRecordCount());
+    }
```
paul-rogers: This is a common operation. Maybe move it into `VectorAccessibleUtilities` or some such. The difference here is that `allocationVectors` (no need for `this.`) is a list of vectors, not an iterator over vector wrappers.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
 
-      container.buildSchema(BatchSchema.SelectionVectorMode.NONE);
+    recordCount = unionall.unionRecords(0, inputBatch.getRecordCount(), 0);
+    for (final ValueVector v : allocationVectors) {
+      final ValueVector.Mutator m = v.getMutator();
+      m.setValueCount(recordCount);
+    }
```
paul-rogers: Same comment as above. Setting value count is common. Here, we have a list of vectors, not vector wrappers.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -39,88 +35,107 @@
...
+
+      IterOutcome upstream = nextBatch.left;
+      RecordBatch incoming = nextBatch.right;
+
+      // skip batches with same schema as the previous one yet having 0 row.
+      if (upstream == IterOutcome.OK && incoming.getRecordCount() == 0) {
+        do {
+          for (final VectorWrapper<?> w : incoming) {
+            w.clear();
+          }
```
paul-rogers: `VectorAccessibleUtilities.clear(incoming)`

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -39,88 +35,107 @@
...
+    } else if (leftUpstream == IterOutcome.OK_NEW_SCHEMA && rightUpstream == IterOutcome.OK_NEW_SCHEMA) {
+      inferOutputFieldsBothSide(left.getSchema(), right.getSchema());
+    }
+
+    container.buildSchema(BatchSchema.SelectionVectorMode.NONE);
+
+    for (VectorWrapper vv: container) {
+      vv.getValueVector().allocateNew();
+      vv.getValueVector().getMutator().setValueCount(0);
+    }
```
paul-rogers:
```
VectorAccessibleUtilities.allocateVectors(container, 0)
VectorAccessibleUtilities.setValueCount(container,0)
```

Hides a bit of the implementation details to use common functions rather than copying the code each time...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
+    if (callBack.getSchemaChangedAndReset()) {
       return IterOutcome.OK_NEW_SCHEMA;
+    } else {
+      return IterOutcome.OK;
     }
+  }
+
+  private void createUnionAller(RecordBatch inputBatch) throws ClassTransformationException, IOException, SchemaChangeException {
+    transfers.clear();
+    allocationVectors.clear();;
```
paul-rogers: One semi-colon should be enough... :-)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
-          }
-
-          ValueVector vv = container.addOrGet(outputField, callBack);
-          allocationVectors.add(vv);
-          TypedFieldId fid = container.getValueVectorId(SchemaPath.getSimplePath(outputField.getPath()));
-          ValueVectorWriteExpression write = new ValueVectorWriteExpression(fid, expr, true);
-          cg.addExpr(write);
-        }
-      // Cast is necessary
+      } else if (vvIn.getField().getType().getMinorType() == TypeProtos.MinorType.NULL) {
```
paul-rogers: Why do we need to copy data to *rename* a column? The underlying buffer is identical regardless of name? Can't rename be done with a transfer? Or, is the original comment simply wrong?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
         if (collector.hasErrors()) {
           throw new SchemaChangeException(String.format("Failure while trying to materialize incoming schema.  Errors:\n %s.", collector.toErrorString()));
         }
 
         // If the inputs' DataMode is required and the outputs' DataMode is not required
         // cast to the one with the least restriction
-        if(vvIn.getField().getType().getMode() == DataMode.REQUIRED
-            && outputFields.get(index).getType().getMode() != DataMode.REQUIRED) {
-          expr = ExpressionTreeMaterializer.convertToNullableType(expr, vvIn.getField().getType().getMinorType(), context.getFunctionRegistry(), collector);
+        if(inField.getType().getMode() == TypeProtos.DataMode.REQUIRED
```
paul-rogers: See DRILL-5709, PR #901 for a faster way to do non-nullable to nullable conversions.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
-        // Both sides are non-empty
-        inferOutputFieldsBothSide();
-      } else if(!rightIsFinish) {
-        // Left side is non-empty
-        // While use left side's column names as output column names,
-        // use right side's column types as output column types.
-        inferOutputFieldsFromSingleSide(
-            leftSide.getRecordBatch().getSchema(),
-            rightSide.getRecordBatch().getSchema());
+  // The output table's column names always follow the left table,
```
paul-rogers: Please clarify just a bit more. We match columns by name, correct? So, the only name conflicts might be by case? "A" on the left, "a" on the right? So, is it more accurate to say that the case of column names always follow the left table?

Or, are the columns correlated by position, and so both the name and type follow the left table?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -39,88 +35,107 @@
...
 
-import com.google.common.collect.Lists;
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Stack;
 
-public class UnionAllRecordBatch extends AbstractRecordBatch<UnionAll> {
+public class UnionAllRecordBatch extends AbstractBinaryRecordBatch<UnionAll> {
```
paul-rogers: Very nice cleanup here! Thanks! The new code is much easier to follow.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
+//    outputFields = Lists.newArrayList();
+    final Iterator<MaterializedField> leftIter = leftSchema.iterator();
+    final Iterator<MaterializedField> rightIter = rightSchema.iterator();
+
+    int index = 1;
+    while (leftIter.hasNext() && rightIter.hasNext()) {
+      MaterializedField leftField  = leftIter.next();
+      MaterializedField rightField = rightIter.next();
+
+      if (hasSameTypeAndMode(leftField, rightField)) {
```
paul-rogers: The code here has the same "flavor" as the code in the code gen setup. Maybe add an explanation for why we need both flavors? What does the one do that the other does not do?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
     }
 
-    public List<MaterializedField> getOutputFields() {
-      if(outputFields == null) {
-        throw new NullPointerException("Output fields have not been inferred");
-      }
-
-      return outputFields;
-    }
+    assert !leftIter.hasNext() && ! rightIter.hasNext() : "Mis-match of column count should have been detected when validating sqlNode at planning";
```
paul-rogers: True? Even in the case of a `SELECT *`? Any other cases where the set of columns isn't known at plan time?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -152,97 +157,75 @@ public void kill(boolean sendUpstream) {
...
+          populateImplicitVectors();
+        }
 
-          // If all the files we have read so far are just empty, the schema is not useful
-          if (! hasReadNonEmptyFile) {
-            container.clear();
-            clearFieldVectorMap();
-            mutator.clear();
-          }
+        boolean isNewImplicitSchema = mutator.isNewSchema();
```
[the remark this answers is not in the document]
jinfengni: You are correct that implicit columns such as _filename_ etc are fixed, and only _dir0_, _dir1_ columns can change. 

Previously,  I thought in the same way, in terms of how Drill handling _a/b/c.csv_,  _a/b/d/e.csv_. That is,  dir2 only exists for the second file, but not for the first file, and hence could lead to a schema change. Turns out that there is code to align the dir list [1].  As a result, across the record readers, the schema for implicit vectors should remain identical. In other words, we do not have to consider implicit vectors in terms of schema change for scan batch.

In the revised patch, I dropped the isNewImplcitSchema logic, since the implicit schema is constant.  Also, I add the checking in ScanBatch's constructor, to make sure either the implicit column list is empty, or they should reman constant for all record readers, and throw setup exception if fails such check. 

1. https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/store/ColumnExplorer.java#L145-L147

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
jinfengni: This patch tries to decouple the logic of record reader and scanbatch:
 - Record reader is responsible to add vectors to batch (via Mutator), and populate data
 - ScanBatch is responsible to interpret the output of record reader, by checking rowCount && Mutator.isNewSchema() to decide whether return OK_NEW_SCHEMA, OK, or NONE. 

> What happens on the first reader? There is no schema, so any schema is a new schema. Suppose the file is JSON and the schema is built on the fly. Does the code handle the case that we have no schema (first reader), and that reader adds no columns?

It's not true "any schema is a new schema". If the first reader has no schema and adds no columns, then Mutator.isNewSchema() should return false. Mutator.isNewSchema() returns true only after the last call, one or more happens

- a new top level field is added, 
- a field in a nested field is added, 
- an existing field type is changed

You may argue a more appropriate way to represent an empty JSON file is an empty schema. However, such idea would lead to various schema conflicts in down-stream operator, if other scan thread has non-empty JSON files. This is exactly what happened before this patch. 

The proposal in this patch is to **ignore** empty JSON, since 1)rowCount=0, 2)no new column were added to batch. 
 - If all the record readers for a scan thread return with rowCount = 0, and produce no new schema, then this Scan thread should return 'NONE' directly, without returning OK_NEW_SCHEMA. 
 - If at least one of reader return either with >0 row, or new schema, then Scan thread will return batch with new schema.
- If all scan threads returns 'NONE' directly, implying the entire table does not have data/schema, this is what Project.handleFastNone() will deal with.

>But, if the input is CSV, then we always have a schema. If the file has column headers, then we know that the schema is, say, (a, b, c) because those are the headers. Or, if the file has no headers, the schema is always the columns array. So, should we send that schema downstream? If so, should it include the implicit columns?

If CSV always adds columns (either _a,b,c, or columns_), then ScanBatch will produce a batch with (a, b, c), or columns. It does not make sense to ignore those schema.  
  - In the case of file with header,  file with _a,b,c_ will lead to a batch with (a,b,c) while a file with _a,b,c,d_ will lead to a batch with (a,b,c,d). Those two files will cause a schema change, which is expected behavior.
  - In the case of file without header, all files will produce a batch with columns, which means there would be no schema change across different files, regardless whether they have row=0, or row > 0.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -152,97 +157,75 @@ public void kill(boolean sendUpstream) {
...
+        recordCount = currentReader.next();
+        Preconditions.checkArgument(recordCount >= 0,
+            "recordCount from RecordReader.next() should not be negative");
 
-              return IterOutcome.OK_NEW_SCHEMA;
-            }
-            return IterOutcome.NONE;
-          }
-          // At this point, the reader that hit its end is not the last reader.
+        boolean isNewRegularSchema = mutator.isNewSchema();
```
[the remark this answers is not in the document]
jinfengni: Since implicit schema is constant, we do not need to distinguish "regular" vs. "implicit".  The revised patch simply uses "isNewSchema".

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -152,97 +157,75 @@ public void kill(boolean sendUpstream) {
...
-            currentReader.allocate(mutator.fieldVectorMap());
-          } catch (OutOfMemoryException e) {
-            clearFieldVectorMap();
-            throw UserException.memoryError(e).build(logger);
+          if (isNewSchema) {
+            // current reader presents a new schema in mutator even though it has 0 row.
+            // This could happen when data sources have a non-trivial schema with 0 row.
+            container.buildSchema(SelectionVectorMode.NONE);
+            schema = container.getSchema();
+            if (readers.hasNext()) {
```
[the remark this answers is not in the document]
jinfengni: I'm neither clear nor convinced that using internal state (such as NO_READER) )would make the code clearer. But if you have such idea implemented in your patch, why don't we review such idea later on?

To some degree, everything we do is "temporary"; as we continue to improve Drill, nothing is implied to be "permanent".

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -152,97 +157,75 @@ public void kill(boolean sendUpstream) {
...
+              advanceNextReader();
+            } else {
+              done = true;  // indicates the follow-up next() call will return IterOutcome.NONE.
+            }
+            return IterOutcome.OK_NEW_SCHEMA;
+          } else { // not a new schema
+            if (readers.hasNext()) {
+              advanceNextReader();
+              continue; // skip reader returning 0 row and having same schema.
```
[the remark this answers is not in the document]
jinfengni: As explained above, both the original code and new code rely on value vector free existing buffer before allocate again.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractBinaryRecordBatch.java
```diff
@@ -0,0 +1,75 @@
...
+
+public abstract class AbstractBinaryRecordBatch<T extends PhysicalOperator> extends  AbstractRecordBatch<T> {
+  protected final RecordBatch left;
+  protected final RecordBatch right;
+
+  // state (IterOutcome) of the left input
+  protected IterOutcome leftUpstream = IterOutcome.NONE;
+
+  // state (IterOutcome) of the right input
+  protected IterOutcome rightUpstream = IterOutcome.NONE;
```
paul-rogers: Nice improvement!

Consider wrapping the two sides in a new class, say, BatchIterator, that holds the batch and iter outcome. Then, the code here is just a bit cleaner because we don't have to repeat variables for left and right.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractSingleRecordBatch.java
```diff
@@ -61,6 +63,10 @@ public IterOutcome innerNext() {
     }
     switch (upstream) {
     case NONE:
+      if (state == BatchState.FIRST) {
+        return handleFastNone();
```
paul-rogers: Again, consider a more descriptive name, such as `handleEmptyInput()`. Otherwise, everyone has to remember that "fast none" == "empty input".

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractSingleRecordBatch.java
```diff
@@ -125,4 +131,19 @@ public BatchSchema getSchema() {
...
+   *
+   * This behavior could be override in each individual operator, if the operator's semantics is to
+   * inject a batch with schema.
+   *
+   * @return IterOutcome.NONE.
+   */
+  protected IterOutcome handleFastNone() {
+    container.buildSchema(SelectionVectorMode.NONE);
+    return IterOutcome.NONE;
+  };
```
paul-rogers: Delete the semi-colon.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractSingleRecordBatch.java
```diff
@@ -125,4 +131,19 @@ public BatchSchema getSchema() {
 
   protected abstract boolean setupNewSchema() throws SchemaChangeException;
   protected abstract IterOutcome doWork();
+
+  /**
+   * Default behavior to handle fast NONE (incoming's first next() return NONE, in stead of OK_NEW_SCHEMA):
+   * FAST NONE could happen when the underneath Scan operators do not produce any batch with schema.
+   *
```
paul-rogers: Javadoc is HTML, so insert \<p>

exec/java-exec/src/test/java/org/apache/drill/TestEmptyBatchSql.java
```diff
@@ -0,0 +1,124 @@
...
+import com.google.common.collect.Lists;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.drill.common.expression.SchemaPath;
+import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.common.util.FileUtils;
+import org.junit.Test;
+
+import java.util.List;
+
+public class TestEmptyBatchSql extends  BaseTestQuery {
```
paul-rogers: Nice set of tests for JSON. Thanks!

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
```diff
@@ -0,0 +1,50 @@
...
+    throw new UnsupportedOperationException();
+  }
+
+  /*
+   * Reason for deprecation is that ValueHolders are potential scalar replacements
+   * and hence we don't want any methods to be invoked on them.
+   */
+  @Deprecated
+  public String toString(){
+    throw new UnsupportedOperationException();
```
paul-rogers: Why unsupported? According to Java, all classes should implement toString. Else, if we are in the debugger, and try to look at an instance of this class, the debugger will encounter an exception.

Also, it is not a good idea to deprecate a standard Java method.


Either leave the default implementation, or simply return `getClass().getSimpleName()`.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
```diff
@@ -0,0 +1,50 @@
...
+package org.apache.drill.exec.vector;
+
+import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.common.types.Types;
+import org.apache.drill.exec.expr.holders.ValueHolder;
+
+public class UntypedNullHolder implements ValueHolder {
+  public static final TypeProtos.MajorType TYPE = Types.optional(TypeProtos.MinorType.NULL);
+
+  public TypeProtos.MajorType getType() {return TYPE;}
```
paul-rogers: The typical Java convention is to list fields first, then methods, so maybe move this after the fields.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public static final int WIDTH = 0;
+
+  public int isSet = 1;
+
+  @Deprecated
+  public int hashCode(){
+    throw new UnsupportedOperationException();
+  }
```
paul-rogers: See note for `toString()`. This has the same issues.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
```diff
@@ -0,0 +1,50 @@
...
+   * Reason for deprecation is that ValueHolders are potential scalar replacements
+   * and hence we don't want any methods to be invoked on them.
+   */
+  @Deprecated
+  public String toString(){
+    throw new UnsupportedOperationException();
+  }
+
+
+
```
paul-rogers: Remove extra blank lines.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
```diff
@@ -0,0 +1,267 @@
...
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(UntypedNullVector.class);
+
+  /**
+   * Width of each fixed-width value.
+   */
+  public static final int VALUE_WIDTH = 0;
+
+  private final Accessor accessor = new Accessor();
+  private final Mutator mutator = new Mutator();
+  private int valueCount ;
```
paul-rogers: Remove space before ;

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
```diff
@@ -0,0 +1,267 @@
...
+  public FieldReader getReader() { throw new UnsupportedOperationException(); }
+
+  @Override
+  public int getBufferSizeFor(final int valueCount) {
+    return 0;
+  }
+
+  @Override
+  public int getValueCapacity(){
+    return Character.MAX_VALUE;
```
paul-rogers: Please use `ValueVector.MAX_ROW_COUNT`. It is a coincidence that the maximum row count happens to be (one greater than) `Character.MAX_VALUE`.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    public int isSet(int index) {
+      checkBounds(index);
+      return 0;
+    }
+
+    @Override
+    public Object getObject(int index) {
+      throw new UnsupportedOperationException();
```
paul-rogers: `return null`? Since the value is always null, shouldn't the Java Object value also be `null`?

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  /**
+   * UntypedNullVector.Mutator throws Exception for most of its mutate operations, except for the ones that set
+   * value counts.
+   *
+   */
+   public final class Mutator extends BaseMutator {
+
+    private Mutator() {};
```
paul-rogers: Drop trailing ;

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -152,97 +157,75 @@ public void kill(boolean sendUpstream) {
...
+                        // Skip to next loop iteration with next available reader.
+            } else {
+              releaseAssets(); // All data has been read. Release resource.
+              return IterOutcome.NONE;
+            }
+          }
+        } else { // recordCount > 0
+          if (isNewSchema) {
```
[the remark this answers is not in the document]
jinfengni: Removed one copy of shared code to handle two different cases in revised patch, though I did not use the idea of state. Hopefully, the new code looks cleaner.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -252,14 +235,28 @@ public IterOutcome next() {
...
+  private void clearFieldVectorMap() {
+    for (final ValueVector v : mutator.fieldVectorMap().values()) {
+      v.clear();
+    }
+  }
+
+  private void advanceNextReader() throws ExecutionSetupException {
+    currentReader = readers.next();
+    implicitValues = implicitColumns.hasNext() ? implicitColumns.next() : null;
+    currentReader.setup(oContext, mutator);
```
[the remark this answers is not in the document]
jinfengni: Nice catch. You are right we should not advance to next reader if we need pass a batch to downstream, since `reader.setup()` for new reader could modify both mutator and container.  This seems to be true for original ScanBatch code as well. 

In the revised batch,  I remove the code to get next reader in constructor, and put the code to get next reader in the beginning of loop when needed. In case of return a `OK_NEW_SCHEMA`, we will only get next reader in the next `next()` method call after `OK_NEW_SCHEMA` is returned.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -329,9 +326,11 @@ public TypedFieldId getValueVectorId(SchemaPath path) {
...
   @VisibleForTesting
   public static class Mutator implements OutputMutator {
-    /** Whether schema has changed since last inquiry (via #isNewSchema}).  Is
-     *  true before first inquiry. */
-    private boolean schemaChanged = true;
+    /** Flag keeping track whether top-level schema has changed since last inquiry (via #isNewSchema}).
+     * It's initialized to false, or reset to false after #isNewSchema or after #clear, until a new value vector
+     * or a value vector with different type is added to fieldVectorMap.
+     **/
+    private boolean schemaChanged;
```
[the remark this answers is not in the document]
jinfengni: No comments with respect to "flag vs counter". But the flag is in the existing code, and the patch I have simply 1) change the initial value of flag, 2) add comments.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
       }
     }
   }
+
+  /**
+   * handle FAST NONE specially when Project for query output. This happens when input returns a
```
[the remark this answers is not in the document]
jinfengni: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+   * FAST NONE directly ( input does not return any batch with schema/data).
+   *
```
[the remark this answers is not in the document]
jinfengni: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+   * Project operator has to return a batch with schema derived using the following 3 rules:
+   *  Case 1:  *  ==>  expand into an empty list of columns.
+   *  Case 2:  regular column reference ==> treat as nullable-int column
+   *  Case 3:  expressions => Call ExpressionTreeMaterialization over an empty vector contain.
```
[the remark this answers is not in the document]
jinfengni: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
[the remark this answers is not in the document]
jinfengni: I change to "handleNullInput" from "handleFastNone". 

The concept of "NullInput" captures the scenarios where input does not produce any batch, and returning NONE directly.  The concept of "empty result set" could mean the case where we have 0 row, but still we could have a batch. I think "empty result set" is not precisely what we want to cover here.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+   *           Once the expression is materialized without error, use the output type of materialized
+   *           expression.
+   * The batch is constructed with the above rules, and recordCount = 0.
+   * Returned with OK_NEW_SCHEMA to down-stream operator.
```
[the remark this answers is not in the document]
jinfengni: It could be Screen/Writer.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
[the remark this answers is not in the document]
jinfengni: I managed to find a way to share the existing code path. With the revised patch, both of the first scenario and second one would call same method to setup a new schema based on incoming batch. The only difference for the second scenario is we have to simulate a record batch with empty vector container, as the code for set up new schema requires it.  That is done by leveraging an existing class (with some extension).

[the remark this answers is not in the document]
jinfengni: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+   */
+  @Override
+  protected IterOutcome handleFastNone() {
+    if (! popConfig.isOutputProj()) {
+      return super.handleFastNone();
+    }
+
+    allocationVectors = new ArrayList<>();
+    final List<NamedExpression> exprs = getExpressionList();
+    final ErrorCollector collector = new ErrorCollectorImpl();
```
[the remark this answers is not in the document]
jinfengni: This code has been removed in revised patch.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+    VectorContainer fakedIncomingVC = new VectorContainer();
+
+    for (NamedExpression namedExpression : exprs) {
+      if (namedExpression.getExpr() instanceof SchemaPath) {
+        final NameSegment expr = ((SchemaPath) namedExpression.getExpr()).getRootSegment();
+        if (expr.getPath().contains(StarColumnHelper.STAR_COLUMN)) {
```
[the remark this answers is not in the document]
jinfengni: This code has been removed in revised patch.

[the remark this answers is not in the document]
jinfengni: This code has been removed in revised patch.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+          continue; // * would expand into an empty list.
+        } else {
+          final TypeProtos.MajorType majorType = TypeProtos.MajorType.newBuilder()
+              .setMinorType(MinorType.INT)
+              .setMode(TypeProtos.DataMode.OPTIONAL)
+              .build();
+
+          MaterializedField outputField = MaterializedField.create(namedExpression.getRef().getRootSegment().getPath(), majorType);
+          final ValueVector vv = container.addOrGet(outputField, callBack);
```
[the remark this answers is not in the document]
jinfengni: This code has been removed in revised patch.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -768,4 +765,73 @@ else if (exprHasPrefix && refHasPrefix) {
...
+          collector,
+          context.getFunctionRegistry(),
+          true,
+          unionTypeEnabled);
+
+      if (collector.hasErrors()) {
+        throw new IllegalArgumentException(String.format("Failure while trying to materialize expressions : %s.  Errors:\n %s.",
+            namedExpression.getExpr(),
+            collector.toErrorString()));
+      }
```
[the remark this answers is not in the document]
jinfengni: This code has been removed in revised patch.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/unit/TestEmptyBatchMiniPlan.java
jinfengni: Add unit case for *, foo.*, bar.*, map and array reference.

I did not cover the case of error in expression tree materializer. If you have an example to produce error in tree materializer, maybe you can share it here. On the other hand, as explained above, the code path is an existing code block.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: Why would you think the above the four lines of code are **_"copied"_** from somewhere else, while the two lines of code are NOT "copied"? You could say the two lines are shorter, but I could not see why the four lines are copied, while the two lines of code are NOT? 

With that said, thanks for pointing to this new util method. I was not aware of that until now. In the revised patch, I used the util method to reduce 4 lines code into 2 lines code.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: The revised patch will throw  IllegalArgumentException.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -39,88 +35,107 @@
...
+          if (!unionInputIterator.hasNext()) {
+            return IterOutcome.NONE;
+          }
+          nextBatch = unionInputIterator.next();
+          upstream = nextBatch.left;
+          incoming = nextBatch.right;
+        } while ((upstream == IterOutcome.OK) &&
+                incoming.getRecordCount() == 0);
+      }
+
```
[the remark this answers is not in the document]
jinfengni: The  loop does not have to handle `STOP` or `OOM` as they are handled in the `switch` statement.  Previously, the loop was intended to handle the same schema yet having row cases.

The revised patch adopted your suggestion, by putting everything in the loop.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: Sounds reasonable. I put two util method in `VectorUtil`, since `VectorAccessibleUtilities` probably should only take `VectorAccessible` related util method, as its comment stated.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: New util method in `VectorUtil`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: This logic actually was in the original UnionAll operator, I simply change the format. 

On the other hand, if there is a `new` and `faster` way for non-nullable to nullable conversions proposed in DRILL-5709, will it make more sense for DRILL-5709 to take care of that, since it's  probably not the scope of DRILL-5546, and more appropriate for DRILL-5709?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: I actually did not change the logic of existing union operator to resolve the output name/type.  Please see the original design doc of union all ( it should be in one of JIRA for union all support).

AFAIK, union operator uses position, in stead of names.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: If you are referring `createUnionAller` and `inferOutputFieldsBothSide`,  the majority of code actually are identical from the original one. AFASIK, the later one is to determine the output field name/type, while the first one is to generate code (may require implicit cast, depending on the future incoming batch's schema and output field type).  

1. https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java#L243-L244
2. https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java#L582-L585

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: This is also in the original code. (Seems git diff did not do a good job). I think it might need cleanup, but it's probably out of scope for this JIRA. 

1. https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java#L620

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: The comment is for the `else` branch, and should be moved a couple lines down. For rename a column, I do not think we are doing a copy; transfer should be good enough.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
jinfengni: I thought the more `;` , the better.  :-)

Removed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
```diff
@@ -130,562 +145,248 @@ public IterOutcome innerNext() {
...
+  private void inferOutputFieldsOneSide(final BatchSchema schema) {
+    for (MaterializedField field : schema) {
+      container.addOrGet(field, callBack);
     }
+  }
 
-    public RecordBatch getLeftRecordBatch() {
-      return leftSide.getRecordBatch();
-    }
+  private static boolean hasSameTypeAndMode(MaterializedField leftField, MaterializedField rightField) {
```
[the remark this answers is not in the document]
jinfengni: Move the method to `MaterializedField`.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillScanRel.java
```diff
@@ -160,12 +161,7 @@ public RelOptCost computeSelfCost(final RelOptPlanner planner, RelMetadataQuery
     final ScanStats stats = groupScan.getScanStats(settings);
     int columnCount = getRowType().getFieldCount();
     double ioCost = 0;
-    boolean isStarQuery = Iterables.tryFind(getRowType().getFieldNames(), new Predicate<String>() {
-      @Override
-      public boolean apply(String input) {
-        return Preconditions.checkNotNull(input).equals("*");
-      }
-    }).isPresent();
+    boolean isStarQuery = AbstractRecordReader.isStarQuery(columns);
```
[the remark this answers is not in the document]
jinfengni: Move to `org.apache.drill.exec.util.Utilities`.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ProjectPrel.java
```diff
@@ -35,18 +35,43 @@
...
+   * @param cluster
+   * @param traits traits of ProjectPrel node
+   * @param child  input
+   * @param exps   list of RexNode, representing expressions of projection.
+   * @param rowType output rowType of projection expression.
+   * @param outputProj true if ProjectPrel is inserted by {@link org.apache.drill.exec.planner.physical.visitor.TopProjectVisitor}
+   *                   Such top Project operator does the following processing, before the result was presented to Screen/Writer
+   *                   1) ensure final output field names are preserved,
+   *                   2) handle cases where input does not return any batch (a fast NONE) (see ProjectRecordBatch.handleFastNone() method)
+   *                   3) handle cases where expressions in upstream operator were evaluated to NULL type
```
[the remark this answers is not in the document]
jinfengni: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractBinaryRecordBatch.java
jinfengni: Better not wrap the two sides in a wrapper class, since the sub-class has to differentiate the two sides. (for instance in HashJoin, left is probe while right is build)

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractSingleRecordBatch.java
jinfengni: Changed to `handleNullInput`

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractSingleRecordBatch.java
jinfengni: done

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractSingleRecordBatch.java
jinfengni: done.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
jinfengni: Done.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
jinfengni: This applies to all the value holder class. See [1].

My understanding is it's for the benefit of potential scalar replacements in byte-code fixup during execution. We may argue we do not have to leverage scalar replacement with JDK8. But before we know for sure, it's better to keep it inline with the other value holder classes. 


1. https://github.com/apache/drill/blob/master/exec/vector/src/main/codegen/templates/ValueHolders.java#L97-L109

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
jinfengni: See response for `toString()`.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
jinfengni: done.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
jinfengni: done.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
jinfengni: Done.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
jinfengni: makes sense.  Changed to `return null`.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullVector.java
jinfengni: Done.

jinfengni:
@paul-rogers , thank you for your detailed review comments. I revised the patch based on your comments, and rebased on latest master. Please see the 2nd commit for the revised patch. Let me know if you have further comments.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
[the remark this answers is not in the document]
paul-rogers: Thanks for the explanation.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
paul-rogers: Makes sense. Thanks for the explanation.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/union/UnionAllRecordBatch.java
paul-rogers: Thanks for the explanation.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractBinaryRecordBatch.java
paul-rogers: Thanks for the explanation.

exec/vector/src/main/java/org/apache/drill/exec/vector/UntypedNullHolder.java
paul-rogers: Hmmm... Not entirely convinced, but let's let it go for now.

PR #907, opened 2017-08-16
DRILL-5697: Improve performance of filter operator for pattern matching

ppadma:
Implement character by character matching for simple patterns and fallback to regex for more complicated patterns. See DRILL-5697 for more details.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/RegexpUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/RegexpUtil.java
@@ -47,10 +47,49 @@ public class RegexpUtil {
       "[:alnum:]", "\\p{Alnum}"
   };
 
+  // type of pattern string.
+  public enum SqlPatternType {
+    STARTS_WITH, // Starts with a constant string followed by any string values (ABC%)
+    ENDS_WITH, // Ends with a constant string, starts with any string values (%ABC)
+    CONTAINS, // Contains a constant string, starts and ends with any string values (%ABC%)
+    CONSTANT, // Is a constant string (ABC)
+    COMPLEX // Not a simple pattern. Needs regex evaluation.
+  };
+
+  public static class SqlPatternInfo {
+    private final SqlPatternType patternType; // type of pattern
+
+    // simple pattern with like meta characters(% and _) and escape characters removed.
+    // Used for simple pattern matching.
+    private final String simplePatternString;
+
+    // javaPatternString used for regex pattern match.
+    private final String javaPatternString;
+
+    public SqlPatternInfo(final SqlPatternType patternType, final String javaPatternString, final String simplePatternString) {
+      this.patternType = patternType;
+      this.simplePatternString = simplePatternString;
+      this.javaPatternString = javaPatternString;
+    }
+
+    public SqlPatternType getPatternType() {
+      return patternType;
+    }
+
+    public String getSimplePatternString() {
+      return simplePatternString;
+    }
+
+    public String getJavaPatternString() {
+      return javaPatternString;
+    }
+
+  }
+
   /**
    * Translates a SQL LIKE pattern to Java regex pattern. No escape char.
    */
-  public static String sqlToRegexLike(String sqlPattern) {
+  public static SqlPatternInfo sqlToRegexLike(String sqlPattern) {
     return sqlToRegexLike(sqlPattern, (char)0);
   }
 
@@ -58,7 +97,7 @@ public class RegexpUtil {
    * Translates a SQL LIKE pattern to Java regex pattern, with optional
    * escape string.
    */
-  public static String sqlToRegexLike(
+  public static SqlPatternInfo sqlToRegexLike(
       String sqlPattern,
       CharSequence escapeStr) {
     final char escapeChar;
@@ -76,12 +115,22 @@ public class RegexpUtil {
   /**
    * Translates a SQL LIKE pattern to Java regex pattern.
    */
-  public static String sqlToRegexLike(
+  public static SqlPatternInfo sqlToRegexLike(
       String sqlPattern,
       char escapeChar) {
     int i;
     final int len = sqlPattern.length();
     final StringBuilder javaPattern = new StringBuilder(len + len);
+    final StringBuilder simplePattern = new StringBuilder(len);
+
+    // Figure out the pattern type and build simplePatternString
+    // as we are going through the sql pattern string
+    // to build java regex pattern string. This is better instead of using
+    // regex later for determining if a pattern is simple or not.
+    // Saves CPU cycles.
+    // Start with patternType as CONSTANT
+    SqlPatternType patternType = SqlPatternType.CONSTANT;
+
     for (i = 0; i < len; i++) {
       char c = sqlPattern.charAt(i);
       if (JAVA_REGEX_SPECIALS.indexOf(c) >= 0) {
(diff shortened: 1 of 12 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/RegexpUtil.java
```diff
@@ -76,12 +113,24 @@ public static String sqlToRegexLike(
...
+
+    // Figure out the pattern type and build simplePatternString
+    // as we are going through the sql pattern string
+    // to build java regex pattern string. This is better instead of using
+    // regex later for determining if a pattern is simple or not.
+    // Saves CPU cycles.
+    sqlPatternType patternType = sqlPatternType.NOT_SIMPLE;
+    boolean startsWith = false;
+    boolean endsWith = false;
+    boolean notSimple = false;
```
paul-rogers: These are not independent states. Probably better to use your enum, with the initial value as null (unknown).

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/RegexpUtil.java
```diff
@@ -96,20 +145,46 @@ public static String sqlToRegexLike(
...
             || (nextChar == escapeChar)) {
           javaPattern.append(nextChar);
+          simplePattern.append(nextChar);
           i++;
         } else {
           throw invalidEscapeSequence(sqlPattern, i);
         }
       } else if (c == '_') {
+        // if we find _, it is not simple pattern, we are looking for only %
+        notSimple = true;
```
paul-rogers: `type = NOT_SIMPLE`

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/RegexpUtil.java
paul-rogers: Or, since your enum represents terminal states, create a new enum with internal states. CONST_ONLY, WILDCARD, COMPLEX with transitions
```
initial: CONST_ONLY
all constant caracters, CONST_ONLY: -> CONST_ONLY
%, CONST_ONLY --> WILDCARD
any other special char, any state --> COMPLEX
%, WILDCARD --> COMPLEX
```
Or, even better, define a simple recursive decent parser in which states are encoded as methods rather than as state variables.
```
parseConstant() ...
- parseWildcard()
- parseComplex()

parseWildcard() ...
- parseComplex()

parseComplex() ...
```

Here, I'm ignoring the details of detecting abc, abc%, ab%c, %abc. These can also be represented as states with the resulting transitions.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/RegexpUtil.java
```diff
@@ -96,20 +145,46 @@ public static String sqlToRegexLike(
...
         javaPattern.append('.');
       } else if (c == '%') {
+        if (i == 0) {
+          // % at the start could potentially be one of the simple cases i.e. ENDS_WITH.
+          endsWith = true;
+        } else if (i == (len-1)) {
```
paul-rogers: A bit of a funky way to do this. Might was well actually wait to the end. This is why we need states (as an enum or via recursive descent.) At end:

If all constants: CONST
If one wildcard: one of the simple cases
Otherwise: COMPLEX

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/RegexpUtil.java
```diff
@@ -96,20 +145,46 @@ public static String sqlToRegexLike(
...
         javaPattern.append(".");
         javaPattern.append('*');
       } else {
         javaPattern.append(c);
+        simplePattern.append(c);
       }
     }
-    return javaPattern.toString();
+
+    if (!notSimple) {
```
paul-rogers: Yeah, the zillion-flags approach is too complex to follow. Really need a good-old state machine.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java
```diff
@@ -57,22 +57,120 @@ private StringFunctions() {}
...
     public void setup() {
-      matcher = java.util.regex.Pattern.compile(org.apache.drill.exec.expr.fn.impl.RegexpUtil.sqlToRegexLike( //
-          org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(pattern.start,  pattern.end,  pattern.buffer))).matcher("");
+      patternInfo = org.apache.drill.exec.expr.fn.impl.RegexpUtil.sqlToRegexLike(
+          org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(pattern.start, pattern.end, pattern.buffer));
       charSequenceWrapper = new org.apache.drill.exec.expr.fn.impl.CharSequenceWrapper();
-      matcher.reset(charSequenceWrapper);
+
+      // Use java regex and compile pattern only if it is not a simple pattern.
+      if (patternInfo.getPatternType() == org.apache.drill.exec.expr.fn.impl.RegexpUtil.sqlPatternType.NOT_SIMPLE) {
```
paul-rogers: You have an enum to describe the cases, and a class to capture the info. That is the perfect place to encode the information about how to process. In particular, the pattern class should act as a factory for a pattern executor: will create an instance of the class needed to do the work. That will also allow this stuff to be unit tested without needing all of Drill.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java
```diff
@@ -57,22 +57,120 @@ private StringFunctions() {}
...
+
+      // Not a simple case. Just use Java regex.
+      if (patternInfo.getPatternType() == org.apache.drill.exec.expr.fn.impl.RegexpUtil.sqlPatternType.NOT_SIMPLE) {
+        matcher.reset();
+        out.value = matcher.matches() ? 1 : 0;
+      }
+
+      // This is a simple pattern that ends with a constant string i.e. %ABC
+      // Compare the characters starting from end.
+      if (patternInfo.getPatternType() == org.apache.drill.exec.expr.fn.impl.RegexpUtil.sqlPatternType.ENDS_WITH) {
```
paul-rogers: Chains of ifs are rather old-school. Do a switch on the enum. And, do it in the pattern class so it can be unit tested.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java
```diff
@@ -85,23 +183,118 @@ public void eval() {
...
     @Override
     public void eval() {
       charSequenceWrapper.setBuffer(input.start, input.end, input.buffer);
       // Reusing same charSequenceWrapper, no need to pass it in.
       // This saves one method call since reset(CharSequence) calls reset()
-      matcher.reset();
-      out.value = matcher.matches()? 1:0;
+
+      // Not a simple case. Just use Java regex.
+      if (patternInfo.getPatternType() == org.apache.drill.exec.expr.fn.impl.RegexpUtil.sqlPatternType.NOT_SIMPLE) {
```
paul-rogers: We are doing a switch (actually, chain of ifs) per value. This is a tight inner loop. Far better to simply generate an instance of the proper class and call a single method to do the work.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java
```diff
@@ -157,6 +157,967 @@ public void testRegexpReplace() throws Exception {
   }
 
   @Test
+  public void testLikeStartsWith() throws Exception {
+
+    // all ASCII.
+    testBuilder()
```
paul-rogers: The regex parsing and execution code is becoming complex. Let's test it with a true unit test, not just a system-level test using a query. See the test frameworks available. We can also discuss in person.

ppadma:
@paul-rogers Paul, thanks a lot for the review. I made changes as per your comments. Please review updated diffs.

ppadma:
@kkhatua Kunal, we can add more patterns later if we want. For now, let us get the most simple cases done first.

PR #909, opened 2017-08-17
DRILL-4264: Allow field names to include dots

vvysotskyi:
1. Removed checking the field name for dots.
2. Replaced using `SchemaPath.getAsUnescapedPath()` method by `SchemaPath.getRootSegmentPath()` and `SchemaPathUtil.getMaterializedFieldFromSchemaPath()` where it is needed. 
3. Replaced using `MaterializedField.getPath()` and `MaterializedField.getLastName()` methods by `MaterializedField.getName()` method and checked the correctness of the behaviour.
4. Added tests


exec/java-exec/src/test/java/org/apache/drill/exec/record/TestMaterializedField.java
```diff
@@ -84,4 +89,22 @@ public void testClone() {
 
   }
 
+  @Test // DRILL-4264
+  public void testSchemaPathToMaterializedFieldConverting() {
```
paul-rogers: No test here for names with dots?

Also, still not convinced of the semantics: not sure we want to create a tree of maps given a nested path. Instead, we need a way to refer to an existing tree and either navigate down that tree to find existing maps, or create a map as needed when a node is missing.

Note that much work has been done in this area in PR #914.

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/RowSetSchema.java
```diff
@@ -83,7 +85,13 @@ private void updateStructure(int index, PhysicalSchema children) {
     public boolean isMap() { return mapSchema != null; }
     public PhysicalSchema mapSchema() { return mapSchema; }
     public MaterializedField field() { return field; }
-    public String fullName() { return fullName; }
+
+    /**
+     * Returns full name path of the column.
```
paul-rogers: This change is fine, but the code here is removed/replaced in PR #914. I believe that #914 correctly follows the guidelines you have created here, but feel free to take a look to check if I missed anything.

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/RowSetSchema.java
```diff
@@ -94,12 +102,20 @@ private void updateStructure(int index, PhysicalSchema children) {
...
+     * and returns the index of the column in the list.
+     *
+     * @param key   full name path of the column in the schema
+     * @param value value to be added to the list
+     * @return index of the column in the list
+     */
+    public int add(String[] key, T value) {
       int index = columns.size();
-      nameSpace.put(key, index);
+      nameSpace.put(SchemaPath.getCompoundPath(key).toExpr(), index);
```
paul-rogers: The goal here is that each name space corresponds to a tuple (a row or a map). So, we always want the the single child name as the key. Suppose we have the path a.b/c.d/e.f/g.h. (Using slashes as path separators here since dots are allowed in names. The name space might be for map "a.b". The key into the name space is "c.d", which resolves to a map. We then repeat the process down through the "c.d" and "e.f" maps.

In short, the original code was fine because the key is assumed to be a single name part that represents the child field relative to this map, not a fully qualified name.

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/RowSetSchema.java
```diff
@@ -182,8 +210,23 @@ public FlattenedSchema(NameSpace<LogicalColumn> cols, NameSpace<LogicalColumn> m
...
-    public int mapIndex(String name) { return maps.columnIndex(name); }
+
+    /**
+     * Returns {@code MaterializedField} instance from the schema
+     * with specified in param full name path of the column.
+     *
+     * @param name full name path of the column in the schema
+     * @return field from the schema
+     */
+    public MaterializedField map(String[] name) { return maps.column(name); }
```
paul-rogers: Not quite right; but don't worry about it; PR #914 replaces/fixes this file.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/TupleAccessor.java
```diff
@@ -48,9 +48,21 @@
...
     MaterializedField column(int index);
 
-    MaterializedField column(String name);
+    /**
+     * Returns {@code MaterializedField} instance from schema using the name path specified in param.
+     *
+     * @param name full name path of the column in the schema
+     * @return {@code MaterializedField} instance
+     */
+    MaterializedField column(String[] name);
```
paul-rogers: This is not right. As explained earlier, a tuple is a specialized map that allows both name and indexed access, but only to direct children. The name above should never be compound. Instead, the code must work the hierarchy itself (or we'd need to add a method to do so.)

Note that this has been revised/replaced in PR #914.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/TupleReader.java
```diff
@@ -27,7 +27,14 @@
...
   ColumnReader column(int colIndex);
-  ColumnReader column(String colName);
+
+  /**
+   * Returns column reader for the column with name path specified in param.
+   *
+   * @param colName full name path of the column in the schema
+   * @return column reader
+   */
+  ColumnReader column(String[] colName);
```
paul-rogers: Thanks for the comment! Again, this must be a single name part.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/TupleWriter.java
```diff
@@ -29,6 +29,13 @@
...
   ColumnWriter column(int colIndex);
-  ColumnWriter column(String colName);
+
+  /**
+   * Returns column writer for the column with name path specified in param.
+   *
+   * @param colName full name path of the column in the schema
+   * @return column writer
+   */
+  ColumnWriter column(String[] colName);
```
paul-rogers: And again.

logical/src/main/java/org/apache/drill/common/expression/PathSegment.java
```diff
@@ -17,11 +17,15 @@
  */
 package org.apache.drill.common.expression;
 
-public abstract class PathSegment{
+public abstract class PathSegment {
 
-  PathSegment child;
+  private PathSegment child;
```
paul-rogers: final?

logical/src/main/java/org/apache/drill/common/expression/SchemaPath.java
```diff
@@ -115,6 +112,33 @@ public static SchemaPath create(NamePart namePart) {
   }
 
   /**
+   * Parses input string and returns {@code SchemaPath} instance.
+   *
+   * @param expr input string to be parsed
+   * @return {@code SchemaPath} instance
+   */
+  public static SchemaPath parseFromString(String expr) {
```
paul-rogers: What are the rules for parsing a string containing dots? Containing []?

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/binary/CompareFunctionsProcessor.java
```diff
@@ -82,466 +45,37 @@ public static CompareFunctionsProcessor process(FunctionCall call, boolean nullC
...
-            PositionedByteRange br = new SimplePositionedMutableByteRange(bb.array(), 0, 9);
-            if (encodingType.endsWith("_OBD")) {
-              org.apache.hadoop.hbase.util.OrderedBytes.encodeFloat64(br,
-                  ((DoubleExpression)valueArg).getDouble(), Order.DESCENDING);
-              this.sortOrderAscending = false;
-            } else {
-              org.apache.hadoop.hbase.util.OrderedBytes.encodeFloat64(br,
-                  ((DoubleExpression)valueArg).getDouble(), Order.ASCENDING);
-            }
+  protected ByteBuf getByteBuf(LogicalExpression valueArg, String encodingType) {
```
paul-rogers: Sorry, it is not at all clear why we deleted the big wad of code. Nor is it clear why, when resolving functions, we start looking at byte buffers. Resolving function is a setup task; byte buffers are runtime tasks. Can you explain this a bit?

vvysotskyi: `org.apache.drill.exec.store.mapr.db.binary.CompareFunctionsProcessor` and `org.apache.drill.exec.store.hbase.CompareFunctionsProcessor` classes contain almost the same code, so I moved mutual code to the abstract class `AbstractCompareFunctionsProcessor`. 
`visitConvertExpression()` method in the `mapr.db.binary.CompareFunctionsProcessor` contains `UTF8_OB` and `UTF8_OBD` cases in the switch block, but this method in `hbase.CompareFunctionsProcessor` class does not contain them, so I added a default case to the method in the abstract class which calls `getByteBuf()` method which considers this difference between the methods.

As I understand, this code is used for **scan pruning** when used `convert_from` function, so we are using byte buffers there.

paul-rogers: Thanks!

logical/src/main/java/org/apache/drill/common/expression/SchemaPath.java
paul-rogers: Thanks for the explanation. Can you copy it into the Javadoc for this method? Will help others in the future.

paul-rogers:
Thanks much for the great work!
+1

amansinha100:
Thanks @vvysotskyi for the PR and @paul-rogers for reviewing the proposal and code.

amansinha100:
Merged in d105950a7a9fb2ff3acd072ee65a51ef1fca120e.  @vvysotskyi pls close the PR (for some reason github is not showing me the option to close).

PR #913, opened 2017-08-17
 - DRILL-5729 Fix Travis Build

ilooner-mapr:


ilooner-mapr:
@parthchandra

ilooner-mapr:
@vrozov Spoke with Parth. Could you take a look at this PR? There are other issues with the travis build. For example it doesn't even run the unit tests right now. But this is a first step to fix what is currently in the travis build from failing.

vrozov:
@ilooner-mapr Do you know why it fails on the Trusty (new default)? Going back to using Precise is OK for a while, but my guess that it will eventually become obsolete.

ilooner-mapr:
@vrozov It looks like the default jdk for Trusty on travis is jdk8. I'll check to see if explicitly telling travis to use jdk7 with Trusty is also a possible fix.

ilooner-mapr:
@vrozov I got it working with Trusty and explicitly configuring openjdk7. One thing to note is that Trusty was running out of memory doing the build with the default 4gb container size. So I had to add the **sudo: required** option in order to increase the vm size https://docs.travis-ci.com/user/reference/overview/ .

vrozov:
@ilooner-mapr I tested it on my branch and it works with the default jdk and without *sudo*. Add `MAVEN_OPTS="-Xms1G -Xmx1G"` before `mvn`.

Can you remove leading dash from the commit message as part of rebase.

.travis.yml
```diff
@@ -13,8 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-sudo: false
+sudo: required
 before_install: git fetch --unshallow
 language: java
+jdk:
+  - openjdk7
 install: mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || (cat mvn_install.log && false)
```
vrozov: @ilooner-mapr can you remove redirection to mvn_install.log as part of the same PR. It is not necessary.

.travis.yml
```diff
@@ -13,8 +13,10 @@
...
 script: mvn package -DskipTests=true
```
vrozov: `install` includes `package` goal and I don't think that javadoc or source plugin are enabled in the default profile, so `script` just repeats the `install`. Also, if the goal is to skip source jar generation, the proper property is maven.source.skip, I believe.

vrozov:
@parthchandra LGTM. There are questions that I have related to how travis-ci build is configured, but they are for pre-existing settings and can be addressed in a follow up PR if necessary.

.travis.yml
```diff
@@ -13,8 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-sudo: false
 before_install: git fetch --unshallow
 language: java
-install: mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dsource.skip=true > mvn_install.log || (cat mvn_install.log && false)
-script: mvn package -DskipTests=true
+jdk:
+  - openjdk7
```
vrozov: @ilooner-mapr Switch to jdk7 is not required, it works with jdk8 as well, but enforcing jdk7 helps with validating that the project is java 1.7 compliant, so I am OK with the change.

ilooner-mapr: Hmm good to know.

ilooner-mapr:
@vrozov I've fixed the source.skip switch to be maven.source.skip as you pointed out. For now I also made the **script** tag a noop since as you said the mvn package command is redundant at this point. The **script** tag can't be completely removed because then travis runs its own default command for **script** which causes failures.

ilooner:
@parthchandra I think the PR still needs a +1 from a committer. If Vlad's review is sufficient, could you give this a +1?

Thanks,
Tim

parthchandra:
+1

PR #914, opened 2017-08-18
DRILL-5657: Size-aware vector writer structure

paul-rogers:
This large PR provides another two levels of foundation for size-aware vector writers in the Drill record readers. It combines code from two previous PRS:

* PR 866 - DRILL-5657: Implement size-aware result set loader
* PR 887 - DRILL-5688: Add repeated map support to column accessors

The PR then goes on to integrate the two prior PRs and provide additional functionality.

Like the two previous PRs, this one is divided into commits to group the work.

1. Accessor layer
2. Row Set layer
3. Tuple and Column Model layer
4. Row Set Loader layer
5. Secondary changes

Much of the material below appears in Javadoc throughout the code. The material here is not meant to replace that documentation. Instead, it is meant to provide the “big picture”: placing the bits and pieces in context and pointing out interesting functionality to explore in each layer.

## Commit 1: The Accessor Layer

The first commit provides the core of the mechanism: the writers that put data into vectors, and the readers that retrieve that data. The version here is an evolution of the version provided in an earlier PR a few months ago.

### Overview of the Drill Vector Data Model

The code will make much more sense if we start with a review of Drill’s complex vector data model. Drill has 38+ data (“minor”) types as defined in the [proto buf](https://github.com/apache/drill/blob/master/protocol/src/main/protobuf/Types.proto) definition. Drill also has three cardinalities (“modes”) defined in the same file. The result is over 120+ different vector types. Then, when you add maps, repeated maps, lists and repeated lists, you rapidly get an explosion of types that the writer code must handle.

Vectors can be categorized along multiple dimensions:

* By data (minor) type
* By cardinality (mode)
* By fixed or variable width

A repeated map, a list, a repeated list and any array (repeated) scalar all are array-like. Nullable and required modes are identical (single values), but a nullable has an additional is-set (“bit”) vector.

A key contribution of this PR is the data model used to organize vectors.

* Both the top-level row, and a Drill map are “tuples” and are treated similarly in the model.
* All non-map, non-list (that is, scalar) data types are treated uniformly.
* All arrays (whether a list, a repeated list, a repeated map, or a repeated scalar) are treated uniformly.

### Accessor Data Model

The above leads to a very simple, JSON-like data model, introduced in this PR.

* A tuple reader or writer models a row. (Usually via a subclass.) Column are accessible by name or position.
* Every column is modeled as an object.
* The object can have an object type: scalar, tuple or array.
* An array has a single element type (but many run-time elements)
* A scalar can be nullable or not, and provides a uniform get/set interface.

This data model is similar to; but has important differences from, the prior, generated, readers and writers. 

The object layer is new: it is the simplest way to model the three “object types.” An app using this code would use just the leaf scalar readers and writers.

Although there is quite a bit of code change here to provide the new structure the core functionality of reading and writing to vectors has not changed much. And, this code has extensive unit tests, which should avoid the need to "mentally execute" each line of code.

See the classes in `org.apache.drill.exec.vector.accessor` for details. In particular, please see the `package-info.java` file in that package for more information.

As before, the top package provides a set of interfaces; the inner packages provide the implementation. The `ColumnAccessors.java` template generates the per-vector code. Warning: this template has become quite cryptic: the best bet for review is to generate the Java code and review that.

### Writer Performance

During previous review, we discussed ways to optimize writer performance. This PR has two improvements:

* Completely rework the writers to minimize code steps
* Rework the “column loaders” to eliminate them: instead of two additional method calls, the “loader” now uses the column writers directly.

Much behind-the-scenes rework was needed to accomplish the above.

Column readers, however, were left with their existing structure; we can apply the same optimizations to the readers in a later PR.

While doing the performance optimization, it became clear we can adopt a major simplification. Writers evolved from having three versions (nullable, required, repeated) to having a single version to handle all three cases. Some items of note:

* The writers bypass DrillBuf and the UDLE to needed writes to direct memory.
* The writers buffer the buffer address and implement a number of methods to synchronize that address when the buffer changes (on a new batch or during vector resize).
* Writing require a single bounds check. In most cases, the write is within bounds so the single check is all that is needed.
* If the write is out of bounds, then the writer determines the new vector size and performs the needed reallocation. To avoid multiple doublings, the writer computes the needed new size and allocates that size directly.
* Vector reallocation is improved to eliminate zeroing the new half of the buffer, data is left “garbage-filled.”
* If the vector would grow beyond 16 MB, then overflow is triggered, via a listener, which causes the buffer to be replaced. The write then continues.
* Offset vector updates are integrated into the writers using an `OffsetVectorWriter`. This writer caches the last write position so that each array write needs a single offset update, rather than the read and write as in previous code.
* The writers keep track of the “last write position” and perform “fill-empties” work if the new write position is more than one position behind the last write. All types now correctly support “fill-empties” (before, only nullable types did so reliably.)
* Null handling is done by an additional writer layer that wraps the underlying data writer. This avoids the need for a special nullable writer: the same nullable layer works for all data types.
* Array handling is done similarly: an array writer manages the offset vector and works the same for repeated scalars, repeated maps and (eventually) lists and repeated lists.

Since vector overflow is now incorporated directly into the writers, this PR backs out the `setScalar()` and related methods added to value vectors in a previous commit.

### Supporting Classes

A new `TupleMetadata` class is a superset of the existing `BatchSchema`, but also provides "classic" tuple-like access by position or name. Eventually, this will also hold additional information such as actual size and so on (information now laboriously rediscovered by the "record batch sizer.") Since the accessors use a "tuple" abstraction to model both rows and maps, the tuple metadata provides this same view. The top-most tuple models the row. Columns within the row can be maps, which have their own tuple schema, and so on.

`TupleNameSpace` moves locations (so it can be used in the vector package) but otherwise remains unchanged.

`DrillBuf` provides an experimental `putInt()` method that does bounds checking and sets a value, to minimize calls. This will probably move into the writer in a later PR.

This PR fixes DRILL-5690, a bug in repeated vectors that did not pass along Decimal scale and precision. See `RepeatedValueVectors.java`.

`MaterializedField` changes to add an `isEquivalent()` method to compare two fields, ignoring internal (`$offset$`, `$bits$`, etc.) vectors.

## Commit 2: Row Set Layer

The “Row Set” family of classes allow tests to quickly build and analyze vector containers in the form of a “row set” - a set of rows. In prior commits, the row set abstraction was the primary client of the accessors (readers and writers.) In this commit, much functionality shifts to be shared with the new “loader” abstraction intended or production code.

Key changes in response to the accessor changes:

* The reader and writer are moved to separate files.
* Row sets now use a common “model” to describe a vector tree (more below).
* Static factory methods were added to hide constructor complexity.
* The `RowSetBuilder` and `RowSetComparison` test tools added support for repeated maps.
* Code to handle generic object writing moved from the `RowSetBuilder` into the accessors.
* The old `RowSetSchema` evolved to become the `TupleMetadata` mentioned above.
* Tests were greatly enhanced to test all modes of all supported scalar types, as well as the new JSON-like structure.

As before the row set unit tests exercise the row set classes themselves, but also are the mechanism to exercise the accessor classes.

See the `org.apache.drill.test.rowSet` and its sub-packages for details.

## Commit 3: Tuple and Column Models

In the previous version, the row set classes had complex logic to figure out what kind of accessor to create for each vector. This became overly complex. In the previous PR, the row set "parses" a vector container to create "storage" objects that represent tuples and columns. A column can, itself, be a tuple. (Note: there is no need to model lists since lists are just vectors at this level of abstraction, so need no special handling.)

In this PR, that concept evolves again to become a “tuple model” and “column model”: a tree-structured representation of the structure of a Drill row. The model cleanly represents things like maps, repeated maps, list of repeated maps that contain arrays, etc.

Think of the model as the “backbone” to work with a batch vector tree. (“Tree” because maps and lists contain other vectors.) Ideally, the vectors themselves would provide this backbone. But, vectors are shared across operators, so that structure created by one operator would conflict with structure needed by another.

With this change, accessor creation is a simple matter of walking a tree to assemble the JSON-structure.

This structure is also used to create a batch's vectors from a schema using “visitor” classes, to allocate vectors based on metadata and other chores needed in the result set loader.

See `org.apache.drill.exec.physical.rowSet.model`. The top package contains interfaces. The two sub packages provide implementations for single batches and hyper batches. (Hyper batches are used only in the test “RowSet” abstractions at present.)

### Tuple and Column Metadata

Also in this commit is a new “extended” metadata model that replaces earlier versions in the row set and previous “loader” commits.

For reasons that will become clear in the next PR, the scan batch ends up doing quite a bit of semantic analysis to map from the select list and the table schema to the result schema. Drill provides a BatchSchema class that is useful, but limited in this context. The metadata layer helps to solve this problem by allowing fast access by both name and position, and allows the schema to grow dynamically.

The dynamic aspect appears in this PR as well: the “loader” API (see below) allows adding columns as the app proceeds; and that action internally adds new columns to the batch schema.

The metadata classes evolved from an earlier version in the row set abstractions. That code was extracted and extended to act as the foundation to create the metadata layer.

The metadata model is simple: it represents tuples (rows, maps) and columns. The metadata is intended to provide addition information beyond what exists in the `MaterializedField` class. In particular, if provides hints about VarChar width and array size. It also provides many useful methods to simplify working with columns.

Eventually, it would be good to combine the new metadata classes with the existing `BatchSchema` and `MaterializedSchema`. But, that is a big change that must wait for another day.

## Commit 4: The Result Set Loader Layer

We now finally get to the ultimate goal of this PR: to introduce a new “result set loader.” This abstraction is an evolution of the “Mutator” class in the scan batch. (The mutator manages a batch; some readers used the existing “writers” with the mutator, others used different techniques such as working directly with native memory.)

A result set loader loads a set of tuples (AKA records, rows) from any source (such as a record reader) into a set of record batches. The loader:

* Divides records into batches based on a maximum row count or a maximum vector size, whichever occurs first. (Later revisions may limit overall batch size.)
* Tracks the start and end of each batch.
* Tracks the start and end of each row.
* Provides column loaders to write each column value.
* Handles overflow when a vector becomes full, but the client still must finish writing the current row.

The original Mutator class divided up responsibilities:

* The Mutator handled the entire record batch
* An optional VectorContainerWriter writes each record

The result set loader follows this same general pattern.

* The result set loader handles the entire record batch (really, a series of batches that make up the entire result set: hence the name.)
* The TupleLoader class provides per-tuple services which mostly consists of access to the column loaders.
* A tuple schema defines the schema for the result set (see below.)

To hide this complexity from the client, a ResultSetLoader interface defines the public API. Then, a `ResultSetLoaderImpl` class implements the interface with all the gory details. Separate classes handle each column, the result set schema, and so on.

Recall that the overall goal of this PR is to handle memory fragmentation by limiting vectors to 16 MB in size. The Loader realizes that goal by silently handling vector overflow by moving the “overflow” row to a new “look-ahead” batch. The complete batch is then sent downstream, and the next read starts with the overflow row as the first row of the next batch.

This version is a bit simpler than the previous one because this one leverages the model, metadata and accessor layers for much of the work.

### Loader API

The actual loader API turns out to be quite simple: another important goal to allow third parties to more easily create custom “record readers.” For example:

```
 void writeABatch() {
   RowSetLoader writer = ...
   while (writer.start()) {
     writer.scalar(0).setInt(10);
     writer.scalar(1).setString("foo");
     ...
     writer.save();
   }
 }
```

Note that, in this version, the `writer` above is defined in the result set loader, but the writer is simply a tuple writer defined in the accessor layer. By the time we get to the `setInt()` or `setString()` methods, the code does a single method call to get into position to write the data. This eliminates the extra “column loader” layer of the prior version of the result set loader.

### Writer and Loader Integration

The column writers are low-level classes that interface between a consumer and a value vector. To create the tuple loader we need a way to bind the column writers to the result set loader. For this, we use a pair of listeners.

* A tuple writer listener handles requests to add a new column at run time.
* A column listener handles vector overflow.

In both cases, the app works with the low-level accessors, but the listener allows events to be fed back to the loader which handles the complex details of dynamic schema evolution and of vector overflow (in fact, it can handle both at the same time.)

To handle vector overflow, each “set” method:

* Does a single bounds check to determine if the value is in the range of the current buffer.
* If not, relocates the buffer, if possible.
* If the vector is full (reached the size limit), calls the listener to handle overflow.
* The listener, via the vector model described earlier, dispatches the request to the loader.
* The loader handles overflow, creating a new vector.
* The writer then writes the value into the new or existing buffer.

### Vector Overflow Logic

The heart of this abstraction is that last point: the ability to detect when a vector overflows, switch in a new vector, and continue writing. Several tricks are involved.

Suppose we have a row of five columns: a through e. The code writes a and b. Then, c overflows. The code can’t rewrite a and b. To handle this, the tuple loader:

* Creates a new, small set of vectors called the “overflow batch”
* Copies columns a and b from the current batch to the overflow batch.
* Writes column c to the overflow batch.
* Allows the client code to finish writing columns d and e (to the overflow batch).
* Reports to the client that the batch is full.

Note that the client is completely unaware that any of the above occurred: it just writes a row and asks if it can write another.

### Skipping Columns

The loader must also handle a reader, such as Parquet, that skips columns if they are null. There were bugs in Drill’s vectors for this case and temporary patches were made in a number of places to make this work. The trick also should work for arrays (a null array is allowed, Drill represents it as an empty array.) But, this code also was broken. For good measure, the code now also allows skipping non-null columns if a good “empty” value is available: 0 for numbers, blank for strings. This behavior is needed for the CSV reader; if a line is missing a field, the CSV reader treats it as an empty (not null) field.

### Result Set Schema

The tuple loader is designed to handle two kinds of tables: “early schema” (such as Parquet and CSV) define the schema up front. “Late schema” (such as JSON) discover the schema during reading. The tuple loader allows either form, and, in fact, uses the same mechanism.

Consumer of batches will, of course, want to know that the schema changed. Providing a simple flag is muddy: when should it be reset? A better solution is to provide a schema version which is incremented each time a column is added. (Columns cannot be removed or changed — at least not yet.)

### Internal Plumbing

The loader assembles the many pieces described earlier:

* The vectors themselves
* The writers (tuples, objects, arrays, scalars)
* The vector model (tree)
* Tuple and column metadata
* The loader logic for handling overflow, etc.

The solution is based on a strict separation of concerns via layering and loose coupling.

* The vector model is the “backbone” that coordinates the other pieces.
* Metadata provides instructions for building the vector model.
* The vector model builds the corresponding vector and writer.
* The writer listener sends events (see above) to the vector model.
* The vector model dispatches events to the loader via a “coordinator” interface.
* The loader performs batch, row and column operations via the vector model.

Please see the extensive Javadoc description in the `org.apache.drill.exec.physical.rowSet` package, especially in the `package-info.java` file in the `impl` sub-package.

### Result Vector Cache

Above we mentioned that the tuple loader allows schema changes on the fly. As the next PR will make more clear, downstream operators want a fixed set of vectors. To assist with this, the tuple loader uses a “result vector cache”. Let’s say a scanner reads two JSON files with the same schema. The first crates the schema and vectors. The second is obligated to use the same vectors. This is a royal pain. But, the vector cache does it automatically: when the tuple loader adds a new column, it checks if the vector already exists in the cache and reuses it. If not there, the cache adds it and returns it so that it is there the next time around.

### Unit Tests

All of the above is pretty thoroughly tested via unit tests. In fact, the unit tests are a good place to start (now I tell you!) in order to see how client code uses the various abstractions.

The bit of unit test structure that handled system options turned out to be wrong. Modified it to use the defaults defined in the system option manager, which required changing the visibility of the defaults table.

## Commit 5: Collateral Damage

The last commit contains various other changes, mostly reflecting the changes above.

Some unit tests were updated to use new features which become available in this PR. See TestFillEmpties and TestVectorLimits.

The `equals()` method in BatchSchema is badly broken. Cleaned it up some. But, didn’t want to change it too much in case anything depends on the current, broken, semantics. So, added a new `isEquivalent` method to provide the correct semantics. Added an `isEquivalent()` method to the MaterializedField as well that will ignore the “implementation” columns that hang off of types such as nullables, repeated, etc. That is, two repeated columns are identical if their type is identical, regardless of whether one has the “$offsets” child or not.

## Open Issues

This PR provides most of the required functionality. However, a few bits and pieces are still in progress and will appear in subsequent commits and/or PRs:

* Full testing of repeated maps
* Support for lists and repeated lists (will reuse structure from repeated maps)
* Column projection (was in an earlier version, new version needed for this structure)


paul-rogers:
Merged a set of enhancements & bug fixes. Compared to the first set of commits:

* Provides tools to print the sometimes-complex tuple model, schema and writer object used for complex data types.
* Sharpens up the specification of the client API used to write data, especially for obscure cases such as abandoning rows, vector overflow inside deep structures, etc.
* Revises the internal “WriterEvents” API used to pass events down the writer hierarchy in response to the API revisions.
* Fully handles the case where a client chooses to “abandon” and rewrite a row (perhaps based on a filter condition.)
* A couple of minor vector revisions. RepeatedMapVector allows building an instance from an existing offset vector (needed when harvesting batches with overflow.)
* Adds more Javadoc.
* Added many unit tests.
* Fixes in response to unit tests.
* Moved vector cache into result set loader layer.

Remaining open issues:

* Column projection
* Support for lists and repeated lists

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/package-info.java
```diff
@@ -0,0 +1,295 @@
...
+ * </dl>
+ *
+ * <h4>General Rules</h4>
+ *
+ * The above can be summarized into a smaller set of rules:
+ * <p>
+ * At the time of overflow on row n:
+ * <ul>
+ * <li>Create or clear the lookahead vector.</li>
+ * <li>Copy (last write position - n) values from row n in the old vector to 0
```
bitblender: Count of rows to be copied should be (last write position - n + 1).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/package-info.java
```diff
@@ -0,0 +1,295 @@
...
+ * <tr><td>n-1</td><td>X</td><td>X</td><td>X</td><td>X</td><td> </td><td> </td><td>-</td><td>-</td></tr>
+ * <tr><td>n  </td><td>X</td><td>!</td><td>O</td><td> </td><td>O</td><td> </td><td>O</td><td> </td></tr>
+ * </table>
+ * Here:
+ * <ul>
+ * <li>n-2, n-1, and n are rows. n is the overflow row.</li>
+ * <li>X indicates a value was written before overflow.</li>
+ * <li>Blank indicates no value was written in that row.</li>
+ * <li>! indicates the value that triggered overflow.</li>
+ * <li>- indicates a column that did not exist prior to overflow.</li>
```
bitblender: What does an 'O' value mean in the diagram above?

paul-rogers: Values added after overflow.
Fixed in latest commit.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/package-info.java
paul-rogers: Fixed.

paul-rogers:
The second group of five commits refines the result set loader mechanism.

### Model Layer Revision

Drill is unique for a query engine in that it handles structured types as first-class data types. For example, Drill supports maps (called “structs” by Hive and Impala), but also supports arrays of maps. Drill support simple scalars, and also arrays of scalars. Put these together and we have maps that contain arrays of maps that contain arrays of scalars. The result is a tree structure described in the Drill documentation as modeled on JSON.

The prior version used a “model” layer to construct internal structures that models the tree-structure of Drill’s data types. The model “reified” the structure into a set of objects. While this worked well, it added more complexity than necessary, especially when dynamically evolving a schema or working out how to handle scan-time projection.

This version retains the model layer, but as a series of algorithms that walk the vector container structure rather than as a separate data structure. The model layer still provides tools to build readers for “single” and “hyper” vectors, to extract a metadata schema from a set of vectors, to create writers for a “single” vector container, and so on.

Replacing the object structure with algorithms required changes to both the row set abstractions and the result set loader.

The key loss in this change is the set of “visitors” from the previous revision. The reified model allowed all visitors to use a common structure. The new solution still has visitors, but now they are ad-hoc, walking the container tree in different ways depending on whether the code can work with columns generically, or needs to deal with individual (single) vectors or hyper vectors.

### Revised Result Set Loader Column and Vector State

To understand the need for many of the changes in this commit, it helps to take a step back and remember what we’re trying to do. We want to write to vectors, but control the resulting batch size.

#### Background

Writing to vectors is easy if we deal only with flat rows and don’t worry about batch size:

* Vectors provide `Mutator` classes that write to single vectors.
* A set of “legacy” vector writers are available, and are used by some readers.
* Generated code uses `Mutator` and `Accessor` classes to work with vectors.
* The “Row Set” classes, used for testing, provides a refined column writer to populate batches.

The above are the easy parts. Some challenges include:

* None of the above limit batch memory, they only limit row count.
* Writing directly to vectors requires that the client deal with the complexity of tracking a common position across vectors.
* Drill’s tree structure makes everything more complex as positions must be tracked across multiple repetition levels (see below).

The “result set loader” (along with the column writers) provides a next level of completeness by tackling the vector memory size problem, for the entire set of Drill structured types.

#### Overflow and Rollover

The key trick is to handle vector “overflow” seamlessly shifting writes, mid-row, from a full batch, to a new “look-ahead” batch. The process of shifting data is called “rollover.”

To implement rollover, we need to work with two sets of vectors:

* The “active” set: the vectors “under” the writers and returned downstream.
* The “backup” set: holds the buffers not currently in use.

During an overflow event, buffers are shifted between the active and backup vectors:

* On overflow, the full buffers reside in the active vectors.
* After rollover, the full buffers reside in the backup vectors, and new buffers, now holding the in-flight row (called the “look-ahead” row), reside in the active vectors.
* When “harvesting” the full batch, the full buffers and look-ahead vectors are exchange, so the full buffers are back in the active vectors.
* When starting the next batch for writing, the look-ahead row is shifted from the backup vectors into the active vectors and the cycle repeats.

#### Column and Vector State

When writing without overflow, we need only one vector and so the usual Drill vector container is sufficient. With overflow, we have two sets of vectors, and must perform operations on them, so we need a place to store the data. This is the purpose of the “column state” and “vector state” classes.

Think of the overall result set loader structure a has having three key parts:

* Result set loader: manages the entire batch
* Column writers: accepts writes to vectors
* Column state: manages the vectors themselves.

The result set loader and column writers are part of the public API. Column state is an internal concept.

Drill is columnar. Each column in a query is represented by a column state object. This object tracks the state of the column: normal, overflow, etc.

Every column state must hold two vectors. The vector state object holds the active and backup vectors, and performs the required operations on these two vectors.

Vector state is separate from column state because some columns use multiple vectors. For example, an array of scalars has an offset vector and a data vector. This column is represented by a single column state object, but two vector state objects.

Maps are special: they have a column state for the map as a whole, along with a “tuple state” to represent the columns within the map. The row itself is a tuple, and so also uses the tuple state object.

Repeated maps have not just the tuple state, but also an offset vector that indexes into the tuple.

#### Schema Subtlety

Suppose a application uses the result set loader to write a batch up through overflow. Suppose that 1000 rows fit into the batch and the 1001st causes overflow. The above description handles this case.

Now suppose that on row 1001 the application adds a new column. Should the column appear within the batch that contains the first 1000 rows? One can argue either way. Consider this argument. If the application just used row counts, the application would have read 1000 rows and sent the batch downstream. Then, on the 1001st row (the first row of the next batch) the application would have added the new column.

Using this reasoning, we adopt a rule that a new column appears only in the batch that contains the row in which the column was added. So, in the overflow case, the new column does not appear in the first batch, but does appear in the second.

To make this work, the column state tracks schema version numbers and can hold onto columns that are not yet visible. The result set loader builds an “output” schema that contains only those columns that are visible. This means that the vector container sent downstream *must* be distinct from the structure used to hold vectors in the result set loader.

For this reason, the tuple and column states are the mechanism to hold the entire schema, while the output vector container is built to include only the visible columns.

This mechanism is even more necessary in the next commit which will add projection back into the result set loader mechanism.

The point is: the tuple and column states allow the result set loader to hold internally a larger set of columns than are made visible to the downstream operators via the output vector container.

### “Torture” Test

This commit adds a new result set loader test called the “torture test.” This single test combines all the complexities that the result set loader must handle:

* Deeply nested repeated structures
* Null values
* Missing values
* Omitted rows
* Vector Rollover

This test has lived up to its name, revealing several subtle bugs in each mechanism. These bugs lead to increased understanding of the requirements for the underlying mechanisms, which resulted in many subtle changes to the accessor and result set loader layers. Little new functionality was added, but the existing functionality was refined to handle many odd cases.

### Repetition Levels

The tests mostly revealed issues with managing offsets in the layers of repeated types. Drill has two kinds of nesting in its tree structure: structures and arrays. Structures just add a level to the name space. (For example, a top-level map introduces a name space, “m”, say, with its columns nested within that name space: “m.a”, “m.b”.) The other kind of nesting introduces a “repetition level.” Drill has four kinds of repetition levels:

* Rows within a row set (batch)
* Scalars within an array of scalars
* Maps (structs) within an array of structs (repeated map)
* Lists (not yet supported in this code)

Getting all the offset vectors, pointers, and rollover logic to work was non-trivial. A persistent source of errors is the fact that offsets written for a row are one head of the row itself. That is, row 10 writes its end offset in position 11, and so on. This requires many, many special cases.

### Rollover Refinement

When a vector overflows, data must be “rolled over” from the in-flight batch to a new one. Rollover is simple with “flat” rows. But, again, many subtle problems crop up when dealing with repetition levels and their corresponding offset vectors and nested indexes. The code here refines rollover handling with several new “writer events” to prepare for, and clean-up after rollover.

Rollover requires work at both the vector and writer level. Work is divided as follows:

* A “preRollover()” event in the writers prepares vectors for overflow by filling empties and setting value counts.
* The result set loader “column state” and “vector state” classes move data from one set of vectors to another.
* The “postRollover()” even in the writers resets writer indexes, addresses and other pointers to reflect the new data positions.

### Even More Javadoc

As the code has stabilized, it has become worthwhile describing the structure in detail in Javadoc, complete with “ASCII-art” to illustrate some of the concepts when working with Drill’s tree-structured vector model.

### More Tests

Additional lower-level “row set” tests were added for various bits of the code. Also, a couple of tools for printing structures were added to aid debugging. (It is easier to fix thing when we can visualize the complex data structures and vector contents.)

### Other Changes

A few nested classes have grown larger and were pulled out into their own files.

paul-rogers:
Rebased onto latest master.

paul-rogers:
This commit introduces a feature to limit memory consumed by a batch.

### Batch Size Limits

With this change, the code now has three overlapping limits:

* The traditional row-count limit.
* A maximum limit of 16 MB per vector.
* The new memory-per-batch limit.

### Overall Flow for Limiting Batch Memory Usage

The batch size limit builds on the work already done for overflow.

* The column metadata allows the client to specify allocation hints such as expected Varchar width and array cardinality.
* The result set loader allocates a batch using the hints and target row count.
* The result set loader measures the memory allocated above. This is the initial batch size.
* As the writers find the need to extend a vector, the writer calls a listener to ask if the extension is allowed, passing in the amount of growth expected.
* The result set loader adds the delta to the accumulated total, compares this against the size limit, and returns whether the resize is allowed.
* If the resize is not allowed, an overflow is triggered.

Note that the above reuses the overflow mechanism, allowing the size limit to be handled even if reached in the middle of a row.

### Implementation Details

To make the above work:

* A new batch size limit is added to the result set loader options.
* The batch size tracking code is added. This required a new method in the value vectors to report actual allocated memory.
* The scalar accessors are refactored to add in the batch size limitation without introducing duplicated code. Code moved from the template to base classes to factor out redundancy.
* General code clean-up in the vector limit found while doing the above work.
* Unit tests for the new mechanism.

paul-rogers:
This next commit reintroduces a projection feature. With this change, a client can:

* Define the set of columns to project
* Define the schema of the data source (table, file, etc.)
* Write columns according to the schema
* Harvest only the projected columns.

### Example

Here is a simple example adapted from `TestResultSetLoaderProjection`. First, declare the projection

```
    List<SchemaPath> selection = Lists.newArrayList(
        SchemaPath.getSimplePath("c"),
        SchemaPath.getSimplePath("b"),
        SchemaPath.getSimplePath("e"));
```

Then, declare the schema. (Here, we declare the schema up-front. Projection also works if the schema is defined as columns are discovered while creating a batch.)

```
    TupleMetadata schema = new SchemaBuilder()
        .add("a", MinorType.INT)
        .add("b", MinorType.INT)
        .add("c", MinorType.INT)
        .add("d", MinorType.INT)
        .buildSchema();
 ```

Then, use the options mechanisms to pass the information to the result set loader:

```
   ResultSetOptions options = new OptionBuilder()
        .setProjection(selection)
        .setSchema(schema)
        .build();
    ResultSetLoader rsLoader = new ResultSetLoaderImpl(fixture.allocator(), options);
```

Now, we can write the four columns in the data source:

```
    RowSetLoader rootWriter = rsLoader.writer();
    rsLoader.startBatch();
…
    rootWriter.start();
    rootWriter.scalar(“a”).setInt(10);
    rootWriter.scalar(“b”).setInt(20);
    rootWriter.scalar(“c”).setInt(30);
    rootWriter.scalar(“d”).setInt(40);
    rootWriter.save();

```

But, when we harvest the results, we get only the projected columns. Notice that “e” is projected, but does not exist in the table, and so is not projected to the output. A higher level of code will handle this case.

```
#: b, c
0: 20, 30
```

### Maps

Although the above example does not show the feature, the mechanism also handles maps and arrays of maps. The rules are:

* If the projection list includes specific map members (such as “m.b”), then project only those map members.
* If the projection list includes just the map name (such as “m”), then project all map members (such as “m.a” and “m.b”.)
* If the projection list does not include the map at all, then project neither the map nor any of its members.

### Implementation

The implementation builds on previous commits. The idea is that we create a “dummy” column and writer, but we do not create the underlying value vector. This allows the client to be blissfully ignorant of whether the column is projected or not. On the other hand, if the client wants to know if a column is projected (perhaps to optimize away certain operations), then the projection status is available in the column metadata.

#### Projection Set

Projection starts with a `ProjectionSet` abstraction. Each tuple (row, map) has a projection set. The projection set can be a set of names (`ProjectionSetImpl`) or a default (`NullProjectionSet`).

#### Result Set Loader

The result set loader is modified to check if a column is projected. If so, the code flow is the same as previously. If not, then the code will create the dummy vector state and dummy writers described above.

Adding support for non-projected columns involved the usual amount of refactoring and moving bits of code around to get a simple solution.

#### Accessor Factories

Prior versions had a `ColumnAccessorFactory` class that created both readers and writers. This commit splits the class into separate reader and writer factories. The writer factory now creates dummy writers if asked to create a writer when the backing vector is null. To make this easier, factory code that previously appeared in each writer has moved into the writer factory. (Note that readers don’t support projection: there is no need.)

#### Dummy Writers

The accessor layer is modified to create a set of dummy writers. Scalar writers have a wide (internal) interface. Dummy scalar writers simply ignore the unsupported operations. Dummy array and tuple writers are also provided.

#### Unit Test

The new `TestResultSetLoaderProjection` test exercises the new code. The new `DummyWriterTest` exercises the dummy writers.

**Comment by paul-rogers** [comment]
Rebased onto master to resolve conflict in `exec/jdbc-all/pom.xml`.

**Comment by paul-rogers** [comment]
Added final development stage to this PR. This is a minor implementation tweak. The result set loader internals for maps held redundant map vectors. In the final vector container, each map must be represented by a map (or repeated map) vector. But, because the result set loader handles projection and overflow, the set of columns that a map writer works with is a superset of those that appear in the output container. For this reason, there turns out to be no reason to maintain these "redundant" map vectors.

For repeated map vectors, we must maintain the offset vector. The code already did this; we just strip off the enclosing repeated map vector.

**Comment by paul-rogers** [comment]
Rebased on to master with conflicts resolved.

**Comment by paul-rogers** [comment]
Thank you, @bitblender, for your in-person review. I've pushed a commit that reflects your comments.

**Comment by paul-rogers** [comment]
@parthchandra, thank you for taking the time to review this large chunk of code. Thanks for noticing the tests; I see those as the only way to ensure that a change of this size actually works; expecting a human to "mentally execute" the code is not practical at this scale.

I did incorporate the change you suggested for grabbing the chunk size from `PooledByteBufAllocatorL` via a new static method in `AllocationManager`. Very good suggestion; thanks.

**Comment by paul-rogers** [comment]
Finally, a note on the fragmentation issue. As you noted, this is a subtle issue. It is true that Netty maintains a memory pool, based on binary allocations, that minimizes the normal kind of fragmentation that results from random sized allocations from a common pool.

The cost of the binary structure is _internal_ fragmentation. Today, Drill vectors have, on average, 25% internal fragmentation. This PR does not address this issue per-se, but sets us on the road toward a solution.

The key fragmentation issue that this PR _does_ deal with is that which occurs when allocations exceed the 16 MB (default) Netty block size. In that case, Netty does, in fact, go to the OS. The OS does a fine job of coalescing large blocks to prevent fragmentation. The problem, however, is that, over time, more and more memory resides in the Netty free list. Eventually, there simply is not enough memory left outside of Netty to service a jumbo (> 16MB) block. Drill gets an OOM error though Netty has many GB of memory free; just none available in the 32+ MB size we want.

We could force Netty to release unused memory. In fact, the original [JE-Malloc paper](https://people.freebsd.org/~jasone/jemalloc/bsdcan2006/jemalloc.pdf) (that you provided way back when, thanks) points out that the allocator should monitor its pools and release memory back to the system when a pool usage drops to zero. It does not appear that `PooledByteBufAllocatorL` implemented this feature, so the allocator never releases memory once it lands in the allocator's free list. We could certainly fix this; the JE-Malloc paper provides suggestions.

Still, however, we could end up with usage patterns in which some slice of memory is used from each chunk, blocking any chunk from being released to the OS, and thereby blocking a "jumbo" block allocation, again though much memory is free on the free list. This is yet another form of fragmentation.

Finally, as you point out, all of this assumes that we want to continue to allocate "jumbo" blocks. But, as we discovered in the managed sort work, and the hash agg spill work, Drill has two conflicting tendencies. On the one hand, "managed" operators wish to operate within a constrained memory footprint. (Which seems to often end up being on the order of 30 MB for the sort for various reasons.) If the scan operator, say, decides to allocate a batch that contains 32 MB vectors, then the sort can't accept even one of those batches an an OOM ensues.

So, rather than solve our memory fragmentation issues by mucking with Netty (force free of unused chunks, increase chunk size, etc.) The preferred solution is to live within a budget: both the constraints of the Netty chunk size *and* the constraints placed on Drill operator memory usage.

In short, we started by wanting to solve the fragmentation issue, but we realized that the best solution is to also solve the unlimited-batch-size issue, hence this PR.

**Comment by paul-rogers** [comment]
Pushed a commit with the changes suggested by Parth's review comments. @parthchandra, please review.

**Comment by paul-rogers** [comment]
Rebased on master. Squashed all previous commits; left code review revisions separate for now.

File: exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/BaseScalarWriter.java [hunk 3]
```diff
@@ -0,0 +1,264 @@
...
+   */
+
+  protected ColumnWriterListener listener;
+
+  /**
+   * Cached direct memory location of the start of data for the vector
+   * being written. Updated each time the buffer is reallocated.
+   */
+
+  protected long bufAddr;
```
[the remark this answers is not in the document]
parthchandra: I feel the buffers internal address should be kept private. Adding a method that the JVM can inline should take care of performance concerns. The main causes of performance degradation were bounds checking and zeroing of vectors and that would still be addressed.

parthchandra:
Just FYI, this (https://github.com/d-t-w/netty-buffers) seems to indicate that some of Netty's internal fragmentation issues seem to have been addressed since 4.0.37

paul-rogers:
Regarding the use of memory addresses. The only reason to do so is performance. To show the benefit of using addresses, I reran the `PerformanceTool` class to test the original code, the code using addresses, and a version that uses DrillBuf as @parthchandra suggested. I expected to see that using addresses was a winner. That's not at all what happened.

The code contains a class, `PerformanceTool` that compares the column writers with the original vector mutators. It loads a vector to 16 MB in size, repeated 300 times. The following are the run times, in ms.

Vector Type | Original  | New w/Address | New w/Drillbuf
------------ | --------  | ------------  | ------------- 
Required | 5703 | 4034 | 1461
Nullable | 12743 | 3645 | 3411
Repeated | 20430 | 7226 | 2669

Here:

* "Original" column uses the original int vector mutator class.
* "New w/Address" shows the same exercise, using the version of the vector writers based on a direct memory address.
* "New w/Drillbuf" shows the vector writers, but using the technique Parth suggested to create "unsafe" methods on the `Drillbuf` class.

The test is run with a pre-allocated vector (no double-and-copy operations). See `PerformanceTool` for details.

I have no explanation for why the `Drillbuf` version should be faster at all, let alone far faster; but I'll take it. The latest commit contains the code after this revision.

So, thank you Parth, you were right again with what turned out to be an outstanding performance boost.

exec/memory/base/src/main/java/io/netty/buffer/DrillBuf.java
```diff
@@ -882,4 +882,71 @@ public void print(StringBuilder sb, int indent, Verbosity verbosity) {
...
+
+  /**
+   * Write an integer to the buffer at the given byte index, without
+   * bounds checks.
+   *
+   * @param index byte (not int) index of the location to write
+   * @param value the value to write
+   */
+
+  public void unsafePutInt(int index, int value) {
```
bitblender: The first argument in these unsafePutXXX methods is an offset right? Should the 'index' be changed to an 'offset' ?

exec/vector/src/main/codegen/templates/ColumnAccessors.java
```diff
@@ -275,17 +273,17 @@ public boolean isNull() {
       final int offset = writeIndex(len);
       <#else>
       final int writeIndex = writeIndex();
-      <#assign putAddr = "bufAddr + writeIndex * VALUE_WIDTH">
+      <#assign putAddr = "writeIndex * VALUE_WIDTH">
```
bitblender: putAddr is not an address but an offset, right ? Should be renamed.

paul-rogers:
There has been discussion recently on Drill's goal of integrating with Arrow. The work to use the `Drillbuf` highlights how we can integrate with Arrow. Simply replace the `Drillbuf` usage with `ArrowBuf`, make the required changes in vector names, add a few methods to the Arrow API, and this entire mechanism can be easily ported over to use Arrow.

parthchandra:
I was hoping that we would just get better encapsulation without losing performance. This performance boost is rather serendipitous.
One possible explanation might be in the JVM optimizing away code paths in the benchmark itself. This means that we might not see the same performance gains in real life. @bitblender knows more about this than I do.

parthchandra:
+1. Ship it!

paul-rogers:
Thanks, Parth! Will make another commit to address small issues that Karthik pointed out. Let's hold off the actual commit until Drill 1.12 ships. I'll then commit the changes when we open things up again for Drill 1.13 changes.

exec/memory/base/src/main/java/io/netty/buffer/DrillBuf.java
paul-rogers: Fixed.

paul-rogers:
Addressed Karthik's comments, rebased on master, resolved conflicts, and squashed all but the conflict resolution commit.

PR #916, opened 2017-08-18
DRILL-5377: Five-digit year dates are displayed incorrectly via jdbc

vdiravka:


vdiravka:
@paul-rogers The similar manner is used for Time millis showing in Drill ([TimePrintMillis](https://github.com/apache/drill/blob/3e8b01d5b0d3013e3811913f0fd6028b22c1ac3f/exec/java-exec/src/main/java/org/apache/drill/exec/vector/accessor/sql/TimePrintMillis.java))
But you are right, using of the custom format for date-to-string conversion is better decision.

Not only test framework converts `Date` to `String`, but [sqlline](https://github.com/julianhyde/sqlline/blob/master/src/main/java/sqlline/Rows.java#L183) as well. So I am going to create an issue ticket for sqlline.

vdiravka:
I've Added TODO with reference to [SQLLine dateFormat, timeFormat, timestampFormat](https://github.com/julianhyde/sqlline/issues/66) issue. The branch is rebased to master version
@paul-rogers Please take take a look at this minor updates.

paul-rogers:
Back to my original question. The premise of this bug seems to be that we corrupt Parquet dates and convert perfectly valid 4-digit years into invalid 5-digit years. That is clearly a data corruption bug that should never occur. Why don't we fix that?

Given that we've accepted the data corruption, we need to display five-digit years which the Java classes for date and time don't support in `toString()`. The code uses `toString()` because it does not do correct formatting using the classes provided. That's the second bug. Date display should make use of format preferences provided by the user, not the default ones provided by `toString()`. So, that's bug number 2.

Now given the above two bugs, we introduce a third by creating ad-hoc, Drill-specific date/time classes, violating the JDBC standard, to display the corrupt five-digit years. So, no longer will Drill return the java.sql.Date class as specified by the standard, but rather our own subclass. How will this affect client code that relies on standard behavior?

I feel we are compounding error upon error. Can we go back and fix the original problem: that users might prefer that we don't corrupt dates in their data? That is, the problem is not so much that we don't format corrupt data correctly, but rather that we do, in fact, corrupt data.

vdiravka:
@paul-rogers 

There is no bug with corrupt Parquet dates, it was fixed in context of DRILL-4203. 

This commit fixes representing of the five digit year dates and doesn't change logic for the 4(3,2...) digit year dates. It is made in similar manner as TimePrintMillis.

But the best solution is to use necessary formatting. I am working on this, so this PR can be closed.

paul-rogers:
Other than dealing with corrupted dates from Parquet, what other uses of five digit years are expected? Few business processes project 8,000 years into the future. Scientific projects are likely to project far more than just 10,000 years into the future; they would need to handle billions of years (death of the sun) to trillions (heat death of the universe.) By contrast, 8000 years ago were civilizations that we no know only from a few archeological remains. Few business records extend that far back.

So, if not for corrupt Parquet dates, what is the use case for 5-digit years? Why is Drill the only tool needing such dates? If they were common, wouldn't `java.sql.Date`, SQL, the ISO standard and other mechanisms define the rules?

arina-ielchiieva:
It seems that this PR is not relevant after Calcite upgrade.
@vdiravka please confirm and close PR.

vdiravka:
@arina-ielchiieva You are right.
According to SQL spec after resolving [CALCITE-2055](https://issues.apache.org/jira/browse/CALCITE-2055) and Drill-Calcite upgrade Drill and Calcite don't support five digit years. 
Please find more details in jira description.

PR #917, opened 2017-08-20
Drill-5734: exclude the conflicted dependency commons-codec

weijietong:
exclude commons-codes from pom files:  contrib/format-maprdb/pom.xml , contrib/storage-hive/hive-exec-shade/pom.xml


paul-rogers:
Same as #903?

PR #920, opened 2017-08-23
DRILL-5737: Hash Agg uses more than the allocated memory under certai…

sohami:
…n low memory conditions

            Note: Provide a new config parameter HASHAGG_FALLBACK_ENABLED which is set to true by default. When 2 Phase
            HashAgg doesn't have enough memory to hold 2 partitions then based on this flag it either fallsback to old
            behavior of consuming unbounded memory or it fails the query.


sohami:
@Ben-Zvi - Please help to review this PR.

priteshm:
@Ben-Zvi can you please review for it to be included in the current sprint?

amansinha100:
+1

PR #921, opened 2017-08-25
DRILL-4286 Graceful shutdown of drillbit

dvjyothsna:
Following is the design document https://docs.google.com/document/d/1AauIwVCjsBKBSbbU6sQODErrEp7n3IwniMtgK4m09MQ/edit#heading=h.224lbmcaxu2d.

I will be doing minor modifications and adding more tests for a couple of days.

NOTE : PCAP_SUB_SCAN in UserBitShared.proto is missing. I have added that since I made some changes to protobuf and generated code. Once it is added as part of DRILL-5432 I will remove it from my commit.

TODOS: 
- Adding Cors support in WebUI.
- Post request to put the Drillbit in offline mode. (Estimate - 2 days)
- Allow port hunting for unit tests. Drillbit allows port hunting only in embedded mode. But port hunting is needed in scenarios where unit tests require running multiple drillbits on the same machine with Zookeeper. (Need some discussion)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/coord/local/LocalClusterCoordinator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/coord/local/LocalClusterCoordinator.java
@@ -18,6 +18,7 @@
 package org.apache.drill.exec.coord.local;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Map;
 import java.util.UUID;
@@ -33,6 +34,7 @@ import org.apache.drill.exec.coord.store.TransientStore;
 import org.apache.drill.exec.coord.store.TransientStoreConfig;
 import org.apache.drill.exec.coord.store.TransientStoreFactory;
 import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
+import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint.State;
 
 import com.google.common.collect.Maps;
 
@@ -69,9 +71,10 @@ public class LocalClusterCoordinator extends ClusterCoordinator {
   }
 
   @Override
-  public RegistrationHandle register(final DrillbitEndpoint data) {
+  public RegistrationHandle register( DrillbitEndpoint data) {
     logger.debug("Endpoint registered {}.", data);
-    final Handle h = new Handle();
+    final Handle h = new Handle(data);
+    data = data.toBuilder().setState(State.ONLINE).build();
     endpoints.put(h, data);
     return h;
   }
@@ -85,13 +88,62 @@ public class LocalClusterCoordinator extends ClusterCoordinator {
     endpoints.remove(handle);
   }
 
+  /**
+   * Update drillbit endpoint state. Drillbit advertises its
+   * state. State information is used during planning and initial
+   * client connection phases.
+   */
+  @Override
+  public RegistrationHandle update(RegistrationHandle handle, State state) {
+    DrillbitEndpoint endpoint = handle.getEndPoint();
+    endpoint = endpoint.toBuilder().setState(state).build();
+    handle.setEndPoint(endpoint);
+    endpoints.put(handle,endpoint);
+    return handle;
+  }
+
   @Override
   public Collection<DrillbitEndpoint> getAvailableEndpoints() {
     return endpoints.values();
   }
 
+  /**
+   * Get a collection of ONLINE Drillbit endpoints by excluding the drillbits
+   * that are in QUIESCENT state (drillbits shutting down). Primarily used by the planner
+   * to plan queries only on ONLINE drillbits and used by the client during initial connection
+   * phase to connect to a drillbit (foreman)
+   * @return A collection of ONLINE endpoints
+   */
+  @Override
+  public Collection<DrillbitEndpoint> getOnlineEndPoints() {
+    Collection<DrillbitEndpoint> runningEndPoints = new ArrayList<>();
+    for (DrillbitEndpoint endpoint: endpoints.values()){
+      if(isDrillbitInState(endpoint, State.ONLINE)) {
+        runningEndPoints.add(endpoint);
+      }
+    }
+    return runningEndPoints;
+  }
+
   private class Handle implements RegistrationHandle {
     private final UUID id = UUID.randomUUID();
+    private DrillbitEndpoint drillbitEndpoint;
+
+    /**
+     * Get the drillbit endpoint associated with the registration handle
+     * @return drillbit endpoint
+     */
+    public DrillbitEndpoint getEndPoint() {
+      return drillbitEndpoint;
+    }
+
+    public void setEndPoint(DrillbitEndpoint endpoint) {
+      this.drillbitEndpoint = endpoint;
+    }
+
+    private Handle(DrillbitEndpoint data) {
+      drillbitEndpoint = data;
+    }
 
     @Override
     public int hashCode() {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKClusterCoordinator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKClusterCoordinator.java
@@ -23,15 +23,16 @@ import static com.google.common.collect.Collections2.transform;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
+import java.util.ArrayList;
 import java.util.Set;
+import java.util.HashSet;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import com.google.common.collect.Lists;
+import org.apache.commons.collections.keyvalue.MultiKey;
 import org.apache.curator.RetryPolicy;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
@@ -54,6 +55,7 @@ import org.apache.drill.exec.coord.store.TransientStore;
 import org.apache.drill.exec.coord.store.TransientStoreConfig;
 import org.apache.drill.exec.coord.store.TransientStoreFactory;
 import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
+import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint.State;
 
 import com.google.common.base.Function;
 
@@ -70,7 +72,10 @@ public class ZKClusterCoordinator extends ClusterCoordinator {
   private final CountDownLatch initialConnection = new CountDownLatch(1);
   private final TransientStoreFactory factory;
   private ServiceCache<DrillbitEndpoint> serviceCache;
+  private DrillbitEndpoint endpoint;
 
+  // endpointsMap maps Multikey( comprises of endoint address and port) to Drillbit endpoints
+  private ConcurrentHashMap<MultiKey, DrillbitEndpoint> endpointsMap = new ConcurrentHashMap<MultiKey,DrillbitEndpoint>();
   private static final Pattern ZK_COMPLEX_STRING = Pattern.compile("(^.*?)/(.*)/([^/]*)$");
 
   public ZKClusterCoordinator(DrillConfig config) throws IOException{
@@ -169,9 +174,10 @@ public class ZKClusterCoordinator extends ClusterCoordinator {
   @Override
   public RegistrationHandle register(DrillbitEndpoint data) {
     try {
+      data = data.toBuilder().setState(State.ONLINE).build();
       ServiceInstance<DrillbitEndpoint> serviceInstance = newServiceInstance(data);
       discovery.registerService(serviceInstance);
-      return new ZKRegistrationHandle(serviceInstance.getId());
+      return new ZKRegistrationHandle(serviceInstance.getId(),data);
     } catch (Exception e) {
       throw propagate(e);
     }
@@ -200,11 +206,50 @@ public class ZKClusterCoordinator extends ClusterCoordinator {
     }
   }
 
+  /**
+   * Update drillbit endpoint state. Drillbit advertises its
+   * state in Zookeeper when a shutdown request of drillbit is
+   * triggered. State information is used during planning and
+   * initial client connection phases.
+   */
+  public RegistrationHandle update(RegistrationHandle handle, State state) {
+    ZKRegistrationHandle h = (ZKRegistrationHandle) handle;
+      try {
+        endpoint = h.endpoint.toBuilder().setState(state).build();
+        ServiceInstance<DrillbitEndpoint> serviceInstance = ServiceInstance.<DrillbitEndpoint>builder()
+                .name(serviceName)
+                .id(h.id)
+                .payload(endpoint).build();
+        discovery.updateService(serviceInstance);
+      } catch (Exception e) {
+        propagate(e);
+      }
+      return handle;
+  }
+
   @Override
   public Collection<DrillbitEndpoint> getAvailableEndpoints() {
     return this.endpoints;
   }
 
+  /*
+   * Get a collection of ONLINE Drillbit endpoints by excluding the drillbits
+   * that are in QUIESCENT state (drillbits shutting down). Primarily used by the planner
+   * to plan queries only on ONLINE drillbits and used by the client during initial connection
+   * phase to connect to a drillbit (foreman)
+   * @return A collection of ONLINE endpoints
+   */
+  @Override
+  public Collection<DrillbitEndpoint> getOnlineEndPoints() {
+    Collection<DrillbitEndpoint> runningEndPoints = new ArrayList<>();
+    for (DrillbitEndpoint endpoint: endpoints){
+      if(isDrillbitInState(endpoint, State.ONLINE)) {
+        runningEndPoints.add(endpoint);
+      }
+    }
+    logger.debug("Online endpoints in ZK are" + runningEndPoints.toString());
+    return runningEndPoints;
+  }
 
   @Override
   public DistributedSemaphore getSemaphore(String name, int maximumLeases) {
@@ -219,6 +264,7 @@ public class ZKClusterCoordinator extends ClusterCoordinator {
 
   private synchronized void updateEndpoints() {
     try {
+      // All active bits in the Zookeeper
       Collection<DrillbitEndpoint> newDrillbitSet =
       transform(discovery.queryForInstances(serviceName),
         new Function<ServiceInstance<DrillbitEndpoint>, DrillbitEndpoint>() {
@@ -229,27 +275,42 @@ public class ZKClusterCoordinator extends ClusterCoordinator {
         });
 
       // set of newly dead bits : original bits - new set of active bits.
-      Set<DrillbitEndpoint> unregisteredBits = new HashSet<>(endpoints);
-      unregisteredBits.removeAll(newDrillbitSet);
-
+      Set<DrillbitEndpoint> unregisteredBits = new HashSet<>();
       // Set of newly live bits : new set of active bits - original bits.
-      Set<DrillbitEndpoint> registeredBits = new HashSet<>(newDrillbitSet);
-      registeredBits.removeAll(endpoints);
+      Set<DrillbitEndpoint> registeredBits = new HashSet<>();
 
-      endpoints = newDrillbitSet;
 
+      // Updates the endpoints map if there is a change in state of the endpoint or with the addition
+      // of new drillbit endpoints. Registered endpoints is set to newly live drillbit endpoints.
+      for ( DrillbitEndpoint endpoint : newDrillbitSet) {
+        String endpointAddress = endpoint.getAddress();
+        int endpointPort = endpoint.getUserPort();
+        if (! endpointsMap.containsKey(new MultiKey(endpointAddress, endpointPort))) {
+          registeredBits.add(endpoint);
+        }
+        endpointsMap.put(new MultiKey(endpointAddress, endpointPort),endpoint);
+      }
+      // Remove all the endpoints that are newly dead
+      for ( MultiKey key: endpointsMap.keySet()) {
+        if(!newDrillbitSet.contains(endpointsMap.get(key))) {
+          unregisteredBits.add(endpointsMap.get(key));
+          endpointsMap.remove(key);
+        }
+      }
+      endpoints = endpointsMap.values();
       if (logger.isDebugEnabled()) {
         StringBuilder builder = new StringBuilder();
         builder.append("Active drillbit set changed.  Now includes ");
         builder.append(newDrillbitSet.size());
         builder.append(" total bits. New active drillbits:\n");
-        builder.append("Address | User Port | Control Port | Data Port | Version |\n");
+        builder.append("Address | User Port | Control Port | Data Port | Version | State\n");
         for (DrillbitEndpoint bit: newDrillbitSet) {
           builder.append(bit.getAddress()).append(" | ");
           builder.append(bit.getUserPort()).append(" | ");
           builder.append(bit.getControlPort()).append(" | ");
           builder.append(bit.getDataPort()).append(" | ");
           builder.append(bit.getVersion()).append(" |");
+          builder.append(bit.getState()).append(" | ");
           builder.append('\n');
         }
         logger.debug(builder.toString());
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -22,6 +22,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.drill.common.AutoCloseables;
 import org.apache.drill.common.StackTrace;
+import org.apache.drill.common.concurrent.ExtendedLatch;
 import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.common.map.CaseInsensitiveMap;
 import org.apache.drill.common.scanner.ClassPathScanner;
@@ -32,6 +33,7 @@ import org.apache.drill.exec.coord.ClusterCoordinator.RegistrationHandle;
 import org.apache.drill.exec.coord.zk.ZKClusterCoordinator;
 import org.apache.drill.exec.exception.DrillbitStartupException;
 import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
+import org.apache.drill.exec.server.DrillbitStateManager.DrillbitState;
 import org.apache.drill.exec.server.options.OptionDefinition;
 import org.apache.drill.exec.server.options.OptionValue;
 import org.apache.drill.exec.server.options.OptionValue.OptionScope;
@@ -48,6 +50,7 @@ import org.apache.drill.exec.util.GuavaPatcher;
 import org.apache.drill.exec.work.WorkManager;
 import org.apache.zookeeper.Environment;
 
+import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint.State;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Stopwatch;
 
@@ -77,6 +80,23 @@ public class Drillbit implements AutoCloseable {
   private final WorkManager manager;
   private final BootStrapContext context;
   private final WebServer webServer;
+  private final int gracePeriod;
+  private DrillbitStateManager stateManager;
+  private boolean quiescentMode;
+  private boolean forcefulShutdown = false;
+
+  public void setQuiescentMode(boolean quiescentMode) {
+    this.quiescentMode = quiescentMode;
+  }
+
+  public void setForcefulShutdown(boolean forcefulShutdown) {
+    this.forcefulShutdown = forcefulShutdown;
+  }
+
+  public RegistrationHandle getRegistrationHandle() {
+    return registrationHandle;
+  }
+
   private RegistrationHandle registrationHandle;
   private volatile StoragePluginRegistry storageRegistry;
   private final PersistentStoreProvider profileStoreProvider;
@@ -110,13 +130,15 @@ public class Drillbit implements AutoCloseable {
     final CaseInsensitiveMap<OptionDefinition> definitions,
     final RemoteServiceSet serviceSet,
     final ScanResult classpathScan) throws Exception {
+    gracePeriod = config.getInt(ExecConstants.GRACE_PERIOD);
     final Stopwatch w = Stopwatch.createStarted();
     logger.debug("Construction started.");
-    final boolean allowPortHunting = serviceSet != null;
+    boolean drillPortHunt = config.getBoolean(ExecConstants.DRILL_PORT_HUNT);
+    final boolean allowPortHunting = (serviceSet != null) || drillPortHunt;
     context = new BootStrapContext(config, definitions, classpathScan);
     manager = new WorkManager(context);
 
-    webServer = new WebServer(context, manager);
+    webServer = new WebServer(context, manager, this);
     boolean isDistributedMode = false;
     if (serviceSet != null) {
       coord = serviceSet.getCoordinator();
@@ -137,6 +159,7 @@ public class Drillbit implements AutoCloseable {
 
     engine = new ServiceEngine(manager, context, allowPortHunting, isDistributedMode);
 
+    stateManager = new DrillbitStateManager(DrillbitState.STARTUP);
     logger.info("Construction completed ({} ms).", w.elapsed(TimeUnit.MILLISECONDS));
   }
 
@@ -152,6 +175,7 @@ public class Drillbit implements AutoCloseable {
     final Stopwatch w = Stopwatch.createStarted();
     logger.debug("Startup begun.");
     coord.start(10000);
+    stateManager.setState(DrillbitState.ONLINE);
     storeProvider.start();
     if (profileStoreProvider != storeProvider) {
       profileStoreProvider.start();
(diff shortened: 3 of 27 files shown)


weijietong:
@dvjyothsna thanks for your contribution. Some objections please consider, the shutdown process should be quick and allow fault occurred. Google calls it lame duck model. If your cluster is large and has high concurrent queries,the grace model will let the shutdown process and the service unavailable time both too long.

dvjyothsna:
@weijietong Please refer to comments under JIRA - https://issues.apache.org/jira/browse/DRILL-4286

exec/java-exec/src/main/java/org/apache/drill/exec/coord/local/LocalClusterCoordinator.java
```diff
@@ -85,13 +88,62 @@ public void unregister(final RegistrationHandle handle) {
...
+   * that are in QUIESCENT state (drillbits shutting down). Primarily used by the planner
+   * to plan queries only on ONLINE drillbits and used by the client during initial connection
+   * phase to connect to a drillbit (foreman)
+   * @return A collection of ONLINE endpoints
+   */
+  @Override
+  public Collection<DrillbitEndpoint> getOnlineEndPoints() {
+    Collection<DrillbitEndpoint> runningEndPoints = new ArrayList<>();
+    for (DrillbitEndpoint endpoint: endpoints.values()){
+      if(endpoint.getState().equals(State.ONLINE)) {
```
sudheeshkatkam: This check should be:
`if(!endpoint.hasState() || endpoint.getState().equals(State.ONLINE)) {`
to not break backward compatibility (new client and old cluster of bits). So the assumption is that old server is ONLINE.

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKClusterCoordinator.java
```diff
@@ -200,11 +206,47 @@ public void unregister(RegistrationHandle handle) {
...
+   * that are in QUIESCENT state (drillbits shutting down). Primarily used by the planner
+   * to plan queries only on ONLINE drillbits and used by the client during initial connection
+   * phase to connect to a drillbit (foreman)
+   * @return A collection of ONLINE endpoints
+   */
+  @Override
+  public Collection<DrillbitEndpoint> getOnlineEndPoints() {
+    Collection<DrillbitEndpoint> runningEndPoints = new ArrayList<>();
+    for (DrillbitEndpoint endpoint: endpoints){
+      if(endpoint.getState().equals(State.ONLINE)) {
```
sudheeshkatkam: (more relevant in this case)

This check should be:
if(!endpoint.hasState() || endpoint.getState().equals(State.ONLINE)) {
to not break backward compatibility (new client and old cluster of bits). So the assumption is that old server is ONLINE.

arina-ielchiieva:
@dvjyothsna is there any update on the status for this PR? Should it be targeted for the 1.12.0 release?

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -165,32 +176,60 @@ public DrillbitContext getContext() {
...
     }
-
-    // Wait for at most 5 seconds or until the latch is released.
-    exitLatch.awaitUninterruptibly(5000);
+    // Wait uninterruptibly until all the queries and running fragments on that drillbit goes down
+    // to zero
+    if( forcceful_shutdown) {
+      exitLatch.awaitUninterruptibly(5000);
+    }
+    else {
```
bitblender: Inconsistent if-else formatting

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -251,6 +252,11 @@ public void run() {
     final String originalName = currentThread.getName();
     currentThread.setName(queryIdString + ":foreman");
 
+    try {
+      checkForemanState();
+    } catch (ForemanException e) {
+      addToEventQueue(QueryState.FAILED, new ForemanException("Query submission failed since foreman is shutting down"));
```
bitblender: why are you not just adding 'e' to the eventQueue ? why do you have to create a new ForemanException object?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -348,6 +354,21 @@ public void run() {
      */
   }
 
+  /*
+    Check if the foreman is ONLINE. If not dont accept any new queries.
+   */
+  public void checkForemanState() throws ForemanException{
+    DrillbitEndpoint foreman = drillbitContext.getEndpoint();
+    Collection<DrillbitEndpoint> dbs = drillbitContext.getAvailableBits();
```
bitblender: I would recommend refactoring this into a generic boolean isOnline(DrillbitEndpoint endpoint) in the DrillbitEndpoint class.

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -471,6 +471,22 @@ public void close() throws Exception {
   }
 
   /**
+   * Shutdown the drillbit given the name of the drillbit.
+   */
+  public void close_drillbit(final String drillbitname) throws Exception {
```
bitblender: camel case.

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -471,6 +471,22 @@ public void close() throws Exception {
...
+    Exception ex = null;
+    for (Drillbit bit : drillbits())
+    {
```
bitblender: placement of '{'

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -0,0 +1,323 @@
...
+    } catch (IOException e) {
+      fail(e.getMessage());
+    }
+  }
+
+  @AfterClass
+  public static void cleanUpTestData() {
+    try {
+      File testData = new File("/tmp/drill-test/gracefulShutdown/");
+      FileUtils.cleanDirectory(testData);
```
bitblender: This only cleans the contents of /tmp/drill-test/gracefulShutdown/ ? Will a subsequent run's setupFile() fail because the second user does not have permission to write into   /tmp/drill-test/gracefulShutdown/ ?

exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitStateManager.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public enum DrillbitState {
+    STARTUP, ONLINE, GRACE, DRAINING, OFFLINE, SHUTDOWN
+  }
+
+  public DrillbitState getState() {
+    return currentState;
+  }
+
+  private DrillbitState currentState;
```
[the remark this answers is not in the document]
dvjyothsna: currentState is updated only in the close() method and close() is synchronized.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
dvjyothsna: DrillbitEndPoint is protobuf generated code so I cannot modify that. Can I add isOnline() method in foreman or a new Class ?

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
dvjyothsna: Yes, that's true. But all other test cases also follow same approach. Is there any other way that I can do this?

exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitStateManager.java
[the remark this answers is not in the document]
bitblender: I think Drillbit.quiescentMode and Drillbit.forceful_shutdown also need NOT be volatile given the way they are used now. You don't have to enforce happens-before (by preventing re-ordering) here and even if these variables are volatile, the read of these variables in close() can anyway race with the setting of these variables in another thread doing a stop/gracefulShutdown. Let me know if I am missing anything.

That said, adding volatiles can only makes the code more correct (and slower). Since this code is not critical you can let it be as it is.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
bitblender: Maybe add it to the DrillbitContext class ?

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
bitblender: Tim's pending changes will make it easy to deal with but meanwhile you can use Files.createTempFile()

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
bitblender: I was thinking of encapsulating code from lines 360 to 367 into a boolean isOnline(), since all the values in that code are derived from the current DrillbitContext.  Then your code would be simplified to 
`
public void checkForemanState() throws ForemanException{
  if (!drillbitContext.isOnline()) {
      throw new ForemanException("Query submission failed since Foreman is shutting down.");
  }
}
`

exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitStateManager.java
```diff
@@ -0,0 +1,80 @@
...
+  public void setState(DrillbitState state) {
+    switch (state) {
+      case ONLINE:
+        if (currentState == DrillbitState.STARTUP) {
+          currentState = state;
+        } else {
+          throw new IllegalStateException("Cannot set drillbit to" + state + "from" + currentState);
```
paul-rogers: To avoid redundant code and to catch new states:

```
setState(State newState) {
  currentState = transitionTo(newState);
}

State transitionTo(State newState) {
  switch (newState) {
    case ONLINE:
      if (currentState == DrillbitState.STARTUP) {
        return newState;
      }
      break;
    ...
    default:
      break;
    }
  throw new IllegalStateException("Cannot set drillbit to" + state + "from" + currentState);
```

An alternative is to switch on current state to say, essentially "if we are in state X, we allow transition to states Y and Z." But, that may be overkill here.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -58,13 +63,170 @@
...
+
+  @SuppressWarnings("resource")
+  @GET
+  @Path("/state")
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response getDrillbitStatus(){
+    Collection<DrillbitInfo> drillbits = getClusterInfoJSON().getDrillbits();
+    Map<String, String> drillStatusMap = new HashMap<String ,String>();
+    for (DrillbitInfo drillbit : drillbits) {
+      drillStatusMap.put(drillbit.getAddress()+"-"+drillbit.getUserPort(),drillbit.getState());
```
paul-rogers: Small point: spaces around `+`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -58,13 +63,170 @@
...
+  @Produces(MediaType.APPLICATION_JSON)
+  public Response shutdownDrillbit() throws Exception {
+    Map<String, String> shutdownInfo = new HashMap<String, String>();
+    try {
+      new Thread(new Runnable() {
+        public void run() {
+          try {
+            drillbit.close();
+          } catch (Exception e) {
+            e.printStackTrace();
```
paul-rogers: Printing the stack trace is OK for debugging. Fo production, we need to log:

```
logger.error("Web request to shutdown drillbit failed", e);
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -58,13 +63,170 @@
...
+  public Response ShutdownForcefully() throws Exception {
+    Map<String, String> shutdownInfo = new HashMap<String, String>();
+    try {
+      new Thread(new Runnable() {
+        public void run() {
+          try {
+            drillbit.setForcefulShutdown(true);
+            drillbit.close();
+          } catch (Exception e) {
+            e.printStackTrace();
```
paul-rogers: See above.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -58,13 +63,170 @@
...
+  public Response drillbitToQuiescentMode() throws Exception {
+    Map<String, String> shutdownInfo = new HashMap<String, String>();
+    try {
+      new Thread(new Runnable() {
+        public void run() {
+          try {
+            drillbit.setQuiescentMode(true);
+            drillbit.close();
+          } catch (Exception e) {
+            e.printStackTrace();
```
paul-rogers: See above.

This is the third copy of this pattern. Can we have a single function that takes a shutdown mode? Does this mean the `Drillbit` class itself should have such a method:

```
public void shutdown(ShutdownOption option) {
...
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -58,13 +63,170 @@
...
+    }
+    catch (Exception e) {
+      logger.debug("Exception in putting drillbit to Quiescent mode",e);
+      shutdownInfo.put("response", "Error in putting drillbit to Quiescent mode");
+      return Response.ok()
+              .entity(shutdownInfo)
+              .header("Access-Control-Allow-Origin", "*")
+              .header("Access-Control-Allow-Methods", "GET, POST, DELETE, PUT")
+              .header("Access-Control-Allow-Credentials","true")
+              .allow("OPTIONS").build();
```
paul-rogers: Similarly, there are many copies of the above pattern. Can we do this in a function that takes, say, the entity as a parameter?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -315,7 +481,12 @@ public int compareTo(DrillbitInfo drillbitToCompare) {
 
       if (this.isVersionMatch() == drillbitToCompare.isVersionMatch()) {
         if (this.version.equals(drillbitToCompare.getVersion())) {
-          return this.address.compareTo(drillbitToCompare.getAddress());
+          {
+            if(this.address.equals(drillbitToCompare.getAddress())) {
+              return (this.controlPort.compareTo(drillbitToCompare.getControlPort()));
```
paul-rogers: Repeat of the `isSameDrillbit()` pattern noted earlier.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -114,11 +117,12 @@
    * @param context Bootstrap context.
    * @param workManager WorkManager instance.
    */
-  public WebServer(final BootStrapContext context, final WorkManager workManager) {
+  public WebServer(final BootStrapContext context, final WorkManager workManager, final Drillbit drillbit) {
```
paul-rogers: Do we really want to pass the entire Drillbit to the web server? Seems like this creates a coupling that is too tight.

What services are needed? Only the shutdown? Something else?

Can we define an interface just for those services, and let `Drillbit` implement that interface to avoid the tight coupling?

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -165,32 +169,59 @@ public DrillbitContext getContext() {
    *
    * <p>This is intended to be used by {@link org.apache.drill.exec.server.Drillbit#close()}.</p>
    */
-  public void waitToExit() {
+  public void waitToExit(Drillbit bit, boolean forcefulShutdown) {
     synchronized(this) {
-      if (queries.isEmpty() && runningFragments.isEmpty()) {
+      numOfRunningQueries = queries.size();
+      numOfRunningFragments = runningFragments.size();
```
paul-rogers: Here we update the counts inside a synchronized block. Do we access them outside the block? How do we synchronize reads? Or, can these be local variables inside this method?

Or, should we have an `updateQueryCount()` method that handles the updates? That method would be synchronized, maybe?

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -165,32 +169,59 @@ public DrillbitContext getContext() {
...
+      if ( queries.isEmpty() && runningFragments.isEmpty()) {
```
paul-rogers: We just updated two variables with counts, then we ignore them. Do we even need them?

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -165,32 +169,59 @@ public DrillbitContext getContext() {
...
+      logger.info("Draining " + queries +" queries and "+ runningFragments+" fragments.");
       exitLatch = new ExtendedLatch();
     }
-
-    // Wait for at most 5 seconds or until the latch is released.
-    exitLatch.awaitUninterruptibly(5000);
+    // Wait uninterruptibly until all the queries and running fragments on that drillbit goes down
+    // to zero
+    if( forcefulShutdown ) {
+      exitLatch.awaitUninterruptibly(5000);
```
paul-rogers: Why 5000? I guess that is the old magic number. Perhaps move this to a constant:

```
private final int FORCEFUL_SHUTDOWN_GRACE_MS = 5000;
```

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -165,32 +169,59 @@ public DrillbitContext getContext() {
...
     synchronized(this) {
       if (exitLatch != null) {
+        logger.info("Waiting for "+ queries.size() +" queries to complete before shutting down");
+        logger.info("Waiting for "+ runningFragments.size() +" running fragments to complete before shutting down");
+        if(runningFragments.size() > numOfRunningFragments|| queries.size() > numOfRunningQueries) {
+          logger.info("New Fragments or queries are added while drillbit is Shutting down");
+        }
         if (queries.isEmpty() && runningFragments.isEmpty()) {
+          // Both Queries and Running fragments are empty.
+          // So its safe for the drillbit to exit.
```
paul-rogers: As it turns out @ilooner is making significant changes to this area. Can you two coordinate?

We just need the @dvjyothsna is making to work long enough for the @ilooner changes replace them. But, Tim's changes need to support this new graceful shutdown model.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -165,32 +169,59 @@ public DrillbitContext getContext() {
...
       }
     }
   }
+  /**
+   *  Get the number of queries that are running on a drillbit.
+   *  Primarily used to monitor the number of running queries after a
+   *  shutdown request is triggered.
+   */
+  public Map<String, Integer> getRemainingQueries() {
+    synchronized (this) {
```
paul-rogers: Since the entire body is synchronized, put the `synchronized` on the method itself.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -368,6 +368,13 @@ drill.exec: {
     // planning and managing queries. Primarily for testing.
     cpus_per_node: 0,
   }
+  # Grace period is the amount of time where the drillbit accepts work after
+  # the shutdown request is triggered. The primary use of grace period is to
+  # avoid the race conditions caused by zookeeper delay in updating the state
+  # information of the drillbit that is shutting down. So, it is advisable
+  # to have a grace period that is atleast twice the amount of zookeeper
+  # refresh time.
+  grace_period : 10000
```
paul-rogers: Units? Just add a note to the end of your comment to say the units are ms. In the future, I've found it to be handy to put the units in the name: `grace_period_ms` for example.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -57,19 +57,19 @@
...
           </thead>
           <tbody>
             <#assign i = 1>
             <#list model.getDrillbits() as drillbit>
-              <tr>
+              <tr id="row-${i}">
                 <td>${i}</td>
-                <td>${drillbit.getAddress()}
-                  <#if drillbit.isCurrent()>
+                <td id="address" >${drillbit.getAddress()}<#if drillbit.isCurrent()>
```
paul-rogers: `"address" >` --> `"address">`

If-statement back to being on its own line?

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -57,19 +57,19 @@
...
                     <span class="label label-info">Current</span>
                   </#if>
                 </td>
-                <td>${drillbit.getUserPort()}</td>
+                <td id="port" >${drillbit.getUserPort()}</td>
```
paul-rogers: Can't us an id here; this is a list of many Drillbits and ids in HTML must be unique (that is, after all, what "identifier" means...)

If this is form formatting, use a class. If for identification, append a suffix: "id-1" or "id-xxx-xxx-xxx-xxx:xx".

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -78,6 +78,11 @@
                     ${drillbit.getVersion()}
                   </span>
                 </td>
+                <td id="status" >${drillbit.getState()}</td>
+                <td>
+                    <button type="button" id="shutdown" onClick="shutdown('${drillbit.getAddress()}',$(this));"> SHUTDOWN </button>
```
paul-rogers: Same id issue.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -78,6 +78,11 @@
...
+                </td>
+                <td id="queriesCount">  </td>
```
paul-rogers: Again.

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -471,6 +471,21 @@ public void close() throws Exception {
   }
 
   /**
+   * Shutdown the drillbit given the name of the drillbit.
+   */
+  public void closeDrillbit(final String drillbitName) throws Exception {
+    Exception ex = null;
+    for (Drillbit bit : drillbits()) {
+      if(bit.equals(bits.get(drillbitName))) {
```
paul-rogers: Space after `if`, here and below.

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -471,6 +471,21 @@ public void close() throws Exception {
...
+        bit.close();
+      }
+    }
+    if(ex != null) {
```
paul-rogers: Looks like `ex` is declared, but never set. Was the goal to capture exceptions from `bit.close()`?

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -0,0 +1,248 @@
...
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Collection;
+import java.util.Properties;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.fail;
+
+public class TestGracefulShutdown {
```
paul-rogers: Add `extends DrillTest` so we get the narration of tests as they run.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -0,0 +1,248 @@
...
+
+  static String testDirPath;
+  @BeforeClass
+  public static void setUpTestData() throws Exception {
+
+    final File testDir = getTempDir("graceful_shutdown");
```
paul-rogers: Will want to change this to integated with the temp dir changes that @ilooner is making for Drill 1.12.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -0,0 +1,248 @@
...
+          ClientFixture client = cluster.clientFixture()) {
+      Drillbit drillbit = cluster.drillbit("db2");
+      int grace_period = drillbit.getContext().getConfig().getInt("drill.exec.grace_period");
+      DrillbitEndpoint drillbitEndpoint =  drillbit.getRegistrationHandle().getEndPoint();
+      new Thread(new Runnable() {
+        public void run() {
+          try {
+            cluster.closeDrillbit("db2");
+          } catch (Exception e) {
+            e.printStackTrace();
```
paul-rogers: OK for debugging. For automated uses, replace with `fail()` here and in any other tests that use this same pattern.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+      }
+    }
+  }
+
+  /*
+   Test shutdown through RestApi
+   */
+  @Test
+  public void testRestApi() throws Exception {
```
paul-rogers: @ilooner provided a REST test fixture in a recent PR. Should this test use that?

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -0,0 +1,248 @@
...
+    return tempDir;
+  }
+
+  private static File getTempFile(int file_num, File testDir) throws Exception {
+    return File.createTempFile("employee"+file_num, ".json", testDir);
+  }
+
+  private static void setupFile(File testDir, int file_num) throws Exception {
+    File destFile = getTempFile(file_num, testDir);
+    try (PrintWriter out = new PrintWriter(new FileWriter(destFile))) {
```
paul-rogers: Can this test use one of our many existing data files? Maybe the `customer` table on the class path?

protocol/src/main/java/org/apache/drill/exec/proto/beans/RpcType.java
```diff
@@ -25,28 +25,8 @@
...
-    CATALOGS(18),
-    SCHEMAS(19),
-    TABLES(20),
-    COLUMNS(21),
-    PREPARED_STATEMENT(23),
-    SERVER_META(9),
-    QUERY_RESULT(10),
-    SASL_MESSAGE(24);
+    REQ_RECORD_BATCH(3),
+    SASL_MESSAGE(4);
```
paul-rogers: Is this your change? Why has the list of RPC types changed? Is this a merge issue?

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -135,8 +157,9 @@ public Drillbit(
       profileStoreProvider = storeProvider;
     }
 
-    engine = new ServiceEngine(manager, context, allowPortHunting, isDistributedMode);
+    engine = new ServiceEngine(manager, context, true, isDistributedMode);
```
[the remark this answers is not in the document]
dvjyothsna: This is needed during testing phase where we start multiple drillbits and shutdown some drillbits

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
dvjyothsna: This was the default one that existed before. That is the reason I didn't changed it.

exec/java-exec/src/main/resources/rest/index.ftl
dvjyothsna: Yes, true. But for every row I have a different ID ("row-1") and I'm accessing "port"(child) of that row.  This was the entire row is unique and need not assign  a new unique id for every element in that row. Am I missing something here?

exec/java-exec/src/main/resources/rest/index.ftl
dvjyothsna: I have tried that before. But it was giving some alignment issues.

exec/java-exec/src/main/java/org/apache/drill/exec/server/DrillbitContext.java
```diff
@@ -157,10 +157,29 @@ public DrillConfig getConfig() {
...
+    return coord.getOnlineEndPoints();
+  }
+  
+  public boolean isOnline(DrillbitEndpoint endpoint) { return endpoint.getState().equals(DrillbitEndpoint.State.ONLINE); }
+
+  public boolean isForemanOnline() {
+    DrillbitEndpoint foreman = getEndpoint();
+    Collection<DrillbitEndpoint> dbs = getAvailableBits();
+    for( DrillbitEndpoint db : dbs) {
+      if( db.getAddress().equals(foreman.getAddress()) && db.getUserPort() == foreman.getUserPort()) {
```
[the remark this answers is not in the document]
bitblender: Can you please explain the logic in isForemanOnline(). Why do you have to get the list of endpoints from ZK and then check for foreman in that list before making the isOnline test ? Why can't it be done on the foreman object? Is this to make sure that the state is updated in ZK before refusing to take queries ?
Why do you assume that the foreman is online if the foreman is not found in the list of endPoints? 
i.e. if it is not in the dbs list
why do you return true in that case ?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -58,13 +63,170 @@
...
+            .entity(drillStatusMap)
+            .header("Access-Control-Allow-Origin", "*")
+            .header("Access-Control-Allow-Methods", "GET, POST, DELETE, PUT")
+              .header("Access-Control-Allow-Credentials","true")
+            .allow("OPTIONS").build();
+  }
+
+  @SuppressWarnings("resource")
+  @GET
+  @Path("/graceperiod")
```
arina-ielchiieva: /gracePeriod

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
arina-ielchiieva: There are at least 7 usages in code all should be removed.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -315,7 +481,12 @@ public int compareTo(DrillbitInfo drillbitToCompare) {
 
       if (this.isVersionMatch() == drillbitToCompare.isVersionMatch()) {
         if (this.version.equals(drillbitToCompare.getVersion())) {
-          return this.address.compareTo(drillbitToCompare.getAddress());
+          {
+            if(this.address.equals(drillbitToCompare.getAddress())) {
```
arina-ielchiieva: `if (`.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -179,6 +184,98 @@
...
+                }
+                else {
+                    fillQueryCount(address,i);
+                    $("#row-"+i).find("#status").text(status_map[key]);
+                }
+            }
+          }
+      }
+      function fillQueryCount(address,row_id) {
+          port_num = 8047
```
arina-ielchiieva: Port number is not always 8047. It can be configured via `drill.exec.http.port`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
dvjyothsna: We compare if the remaining queries are going down or if new queries are added during shutdown.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
dvjyothsna: We compare if the remaining queries are going down or if new queries are added during shutdown.

protocol/src/main/java/org/apache/drill/exec/proto/beans/RpcType.java
dvjyothsna: This is not my change. This file is generated by protobuf. When I looked into the issue I found out that user.proto and BitData.proto, Bitcontrol.proto have same enum (RpcType) . That is causing the issue.

paul-rogers: The change seems to be that messages are dropped. That can't be good. The only diff that should show up here is the addition of your new state codes.

The other explanation is that master is wrong, which would be a bad state of affairs.

arina-ielchiieva:
@paul-rogers is right, there is something wrong with the protos @dvjyothsna has generated.
I took master branch and generated protos. No changes so these extra changes we see are wrong. Also I have created branch where I have added changes in protos from this PR [1] and from the diff it can be seen that no extra changes as well [2].

@dvjyothsna please address this issue.
If it helps:
I have installed protoc:
```
protoc --version
libprotoc 2.5.0
```
and generated using the following:
```
cd ~/git_repo/drill/protocol
mvn process-sources -P proto-compile
```
[1] https://github.com/arina-ielchiieva/drill/tree/DRILL-4386-proto
[2] https://github.com/arina-ielchiieva/drill/commit/07af545ab27b77e5f86002d4ce0f54b3dd371343

arina-ielchiieva:
@dvjyothsna  also please remove `e.printStackTrace();` from the code, I see 4 occurrences.

arina-ielchiieva:
Is there any ETA for the fix?

arina-ielchiieva:
+1, LGTM.

arina-ielchiieva:
There are unit test failures with this fix. Full output is attached into Jira. Please check and fix.
```
  TestMetadataProvider.columns:251 expected:<92> but was:<93>
  DrillSeparatePlanningTest.testMultiMinorFragmentComplexQuery:122->getResultsHelper:219 expected:<1> but was:<0>
  DrillSeparatePlanningTest.testSingleFragmentQuery:88->getResultsHelper:219 expected:<1> but was:<0>
  DrillSeparatePlanningTest.testMultiMinorFragmentSimpleQuery:105->getResultsHelper:219 expected:<1> but was:<0>
```

arina-ielchiieva:
@dvjyothsna all tests has passed. Though there is significant performance degradation.
Before your changes unit test run takes ~ 33 minutes, after ~ 1 hour 18 minutes. 
For example, exec package:
before `[INFO] exec/Java Execution Engine ......................... SUCCESS [17:14 min]`,
after `[INFO] exec/Java Execution Engine ......................... SUCCESS [56:45 min]`.
Since you did not add that many unit tests, could you please explain the reason of such performance degradation and ideally fix it.
Attached two unit tests outputs in Jira.

dvjyothsna:
This was due to grace_period during shutdown of drillbits. Fixed it by changing grace_period = 0 in .conf

paul-rogers:
Interesting issue! Let's think about this a bit.

In a production system, we do want the grace period; it is an essential part of the graceful shutdown procedure.

However, if we are doing a non-graceful shutdown, the grace is unneeded.

Also, if the cluster contains only one node (as in most unit tests), there is nothing to wait for, so the grace period is not needed. The same is true in an embedded Drillbit for Sqlline.

So, can we provide a solution that handles these cases rather than simply turning off the grace period always?

If using the local cluster coordinator, say, then no grace is needed. If using ZK, but there is only one Drillbit, no grace is needed. (There is a race condition, but may be OK.)

Or, if we detect we are embedded, no grace period.

Then, also, if we are doing a graceful shutdown, we need the grace. But, if we are doing a "classic" shutdown, no grace is needed.

The result should be that the grace period is used only in production servers, only when doing a graceful shutdown.

There are probably some details to simplify, but I hope the above communicates the idea. Can we make this work?

arina-ielchiieva:
Since we don't want to hold the release any more. @paul-rogers should we address this issue in the next release or move this PR to the next release? I am in the favor of the second.

paul-rogers:
My suggestion is to go ahead and commit this PR with Jyothsna's change to disable the "graceful" part of the graceful shutdown. The feature will be in the product and people can turn it on to play with it. Then, @dvjyothsna should file a JIRA for the improvements we discussed (enabling by default for true servers, disabling for embedded.)

PR #923, opened 2017-08-26
DRILL-5723: Added System Internal Options That can be Modified at Runtime

ilooner:
This implements Internal system options that are modifiable at runtime: https://issues.apache.org/jira/browse/DRILL-5723

Internal options do not show up in the **sys.options** table.

Internal options can be queried by:

```
select * from sys.internal_options;
```

or

```
select * from sys.internal_options_val;
```

Internal options can be updated in the same way as normal options:

```
alter system set `my.internal.prop` = 'yay'
```

Internal options do not show up in the following rest endpoints

 - **http://host:8047/options.json**
 - **http://host:8047/options**

Internal options do show up in the following rest endpoints

 - **http://host:8047/internal_options.json**
 - **http://host:8047/internal_options**

Internal options can be set in the same way as other options via the rest api

 - **http://host:8047/option?name=my.internal.prop&value=blah&kind=session**

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -116,6 +116,7 @@ public interface ExecConstants {
   String HTTP_ENABLE = "drill.exec.http.enabled";
   String HTTP_MAX_PROFILES = "drill.exec.http.max_profiles";
   String HTTP_PORT = "drill.exec.http.port";
+  String HTTP_PORT_HUNT = "drill.exec.http.porthunt";
   String HTTP_ENABLE_SSL = "drill.exec.http.ssl_enabled";
   String HTTP_CORS_ENABLED = "drill.exec.http.cors.enabled";
   String HTTP_CORS_ALLOWED_ORIGINS = "drill.exec.http.cors.allowedOrigins";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -19,6 +19,7 @@ package org.apache.drill.exec;
 
 import org.apache.drill.exec.physical.impl.common.HashTable;
 import org.apache.drill.exec.rpc.user.InboundImpersonationManager;
+import org.apache.drill.exec.server.options.OptionMetaData;
 import org.apache.drill.exec.server.options.OptionValidator;
 import org.apache.drill.exec.server.options.TypeValidators.BooleanValidator;
 import org.apache.drill.exec.server.options.TypeValidators.DoubleValidator;
@@ -423,13 +424,13 @@ public interface ExecConstants {
    * such as changing system options.
    */
   String ADMIN_USERS_KEY = "security.admin.users";
-  StringValidator ADMIN_USERS_VALIDATOR = new StringValidator(ADMIN_USERS_KEY, true);
+  StringValidator ADMIN_USERS_VALIDATOR = new StringValidator(ADMIN_USERS_KEY);
 
   /**
    * Option whose value is a comma separated list of admin usergroups.
    */
   String ADMIN_USER_GROUPS_KEY = "security.admin.user_groups";
-  StringValidator ADMIN_USER_GROUPS_VALIDATOR = new StringValidator(ADMIN_USER_GROUPS_KEY, true);
+  StringValidator ADMIN_USER_GROUPS_VALIDATOR = new StringValidator(ADMIN_USER_GROUPS_KEY);
   /**
    * Option whose value is a string representing list of inbound impersonation policies.
    *
@@ -472,8 +473,7 @@ public interface ExecConstants {
    * for any query.
    */
   String ENABLE_QUERY_PROFILE_OPTION = "exec.query_profile.save";
-  BooleanValidator ENABLE_QUERY_PROFILE_VALIDATOR = new BooleanValidator(
-      ENABLE_QUERY_PROFILE_OPTION);
+  BooleanValidator ENABLE_QUERY_PROFILE_VALIDATOR = new BooleanValidator(ENABLE_QUERY_PROFILE_OPTION);
 
   /**
    * Profiles are normally written after the last client message to reduce latency.
@@ -482,8 +482,7 @@ public interface ExecConstants {
    * verification.
    */
   String QUERY_PROFILE_DEBUG_OPTION = "exec.query_profile.debug_mode";
-  BooleanValidator QUERY_PROFILE_DEBUG_VALIDATOR = new BooleanValidator(
-      QUERY_PROFILE_DEBUG_OPTION);
+  BooleanValidator QUERY_PROFILE_DEBUG_VALIDATOR = new BooleanValidator(QUERY_PROFILE_DEBUG_OPTION);
 
   String USE_DYNAMIC_UDFS_KEY = "exec.udf.use_dynamic";
   BooleanValidator USE_DYNAMIC_UDFS = new BooleanValidator(USE_DYNAMIC_UDFS_KEY);
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -23,6 +23,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.drill.common.AutoCloseables;
 import org.apache.drill.common.StackTrace;
 import org.apache.drill.common.config.DrillConfig;
+import org.apache.drill.common.map.CaseInsensitiveMap;
 import org.apache.drill.common.scanner.ClassPathScanner;
 import org.apache.drill.common.scanner.persistence.ScanResult;
 import org.apache.drill.exec.ExecConstants;
@@ -31,10 +32,10 @@ import org.apache.drill.exec.coord.ClusterCoordinator.RegistrationHandle;
 import org.apache.drill.exec.coord.zk.ZKClusterCoordinator;
 import org.apache.drill.exec.exception.DrillbitStartupException;
 import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
-import org.apache.drill.exec.server.options.OptionManager;
+import org.apache.drill.exec.server.options.OptionDefinition;
 import org.apache.drill.exec.server.options.OptionValue;
-import org.apache.drill.exec.server.options.OptionValue.OptionType;
 import org.apache.drill.exec.server.options.OptionValue.OptionScope;
+import org.apache.drill.exec.server.options.SystemOptionManager;
 import org.apache.drill.exec.server.rest.WebServer;
 import org.apache.drill.exec.service.ServiceEngine;
 import org.apache.drill.exec.store.StoragePluginRegistry;
@@ -82,19 +83,36 @@ public class Drillbit implements AutoCloseable {
 
   @VisibleForTesting
   public Drillbit(
+    final DrillConfig config,
+    final RemoteServiceSet serviceSet) throws Exception {
+    this(config, SystemOptionManager.createDefaultOptionDefinitions(), serviceSet, ClassPathScanner.fromPrescan(config));
+  }
+
+  @VisibleForTesting
+  public Drillbit(
       final DrillConfig config,
+      final CaseInsensitiveMap<OptionDefinition> definitions,
       final RemoteServiceSet serviceSet) throws Exception {
-    this(config, serviceSet, ClassPathScanner.fromPrescan(config));
+    this(config, definitions, serviceSet, ClassPathScanner.fromPrescan(config));
   }
 
   public Drillbit(
-      final DrillConfig config,
-      final RemoteServiceSet serviceSet,
-      final ScanResult classpathScan) throws Exception {
+    final DrillConfig config,
+    final RemoteServiceSet serviceSet,
+    final ScanResult classpathScan) throws Exception {
+    this(config, SystemOptionManager.createDefaultOptionDefinitions(), serviceSet, classpathScan);
+  }
+
+  @VisibleForTesting
+  public Drillbit(
+    final DrillConfig config,
+    final CaseInsensitiveMap<OptionDefinition> definitions,
+    final RemoteServiceSet serviceSet,
+    final ScanResult classpathScan) throws Exception {
     final Stopwatch w = Stopwatch.createStarted();
     logger.debug("Construction started.");
     final boolean allowPortHunting = serviceSet != null;
-    context = new BootStrapContext(config, classpathScan);
+    context = new BootStrapContext(config, definitions, classpathScan);
     manager = new WorkManager(context);
 
     webServer = new WebServer(context, manager);
@@ -198,7 +216,7 @@ public class Drillbit implements AutoCloseable {
       return;
     }
 
-    final OptionManager optionManager = getContext().getOptionManager();
+    final SystemOptionManager optionManager = getContext().getOptionManager();
 
     // parse out the properties, validate, and then set them
     final String systemProps[] = allSystemProps.split(",");
@@ -219,16 +237,16 @@ public class Drillbit implements AutoCloseable {
       }
 
       final OptionValue defaultValue = optionManager.getOption(optionName);
+
       if (defaultValue == null) {
         throwInvalidSystemOption(systemProp, "does not specify a valid option name");
       }
-      if (defaultValue.type != OptionType.SYSTEM) {
+
+      if (!defaultValue.accessibleScopes.inScopeOf(OptionScope.SYSTEM)) {
         throwInvalidSystemOption(systemProp, "does not specify a SYSTEM option ");
       }
 
-      final OptionValue optionValue = OptionValue.createOption(
-          defaultValue.kind, OptionType.SYSTEM, optionName, optionString, OptionScope.SYSTEM);
-      optionManager.setOption(optionValue);
+      optionManager.setLocalOption(defaultValue.kind, optionName, optionString);
     }
   }
 
@@ -287,22 +305,28 @@ public class Drillbit implements AutoCloseable {
   }
 
   public static Drillbit start(final StartupOptions options) throws DrillbitStartupException {
-    return start(DrillConfig.create(options.getConfigLocation()), null);
+    return start(DrillConfig.create(options.getConfigLocation()), SystemOptionManager.createDefaultOptionDefinitions(), null);
   }
 
   public static Drillbit start(final DrillConfig config) throws DrillbitStartupException {
-    return start(config, null);
+    return start(config, SystemOptionManager.createDefaultOptionDefinitions(), null);
+  }
+
+  public static Drillbit start(final DrillConfig config, final RemoteServiceSet remoteServiceSet) throws DrillbitStartupException {
+    return start(config, SystemOptionManager.createDefaultOptionDefinitions(), remoteServiceSet);
   }
 
   @SuppressWarnings("resource")
-  public static Drillbit start(final DrillConfig config, final RemoteServiceSet remoteServiceSet)
+  @VisibleForTesting
+  public static Drillbit start(final DrillConfig config, final CaseInsensitiveMap<OptionDefinition> validators,
+                               final RemoteServiceSet remoteServiceSet)
       throws DrillbitStartupException {
     logger.debug("Starting new Drillbit.");
     // TODO: allow passing as a parameter
     ScanResult classpathScan = ClassPathScanner.fromPrescan(config);
     Drillbit bit;
     try {
-      bit = new Drillbit(config, remoteServiceSet, classpathScan);
+      bit = new Drillbit(config, validators, remoteServiceSet, classpathScan);
     } catch (final Exception ex) {
       throw new DrillbitStartupException("Failure while initializing values in Drillbit.", ex);
     }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/BaseOptionManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/BaseOptionManager.java
@@ -17,44 +17,108 @@
  */
 package org.apache.drill.exec.server.options;
 
-import org.apache.drill.exec.server.options.TypeValidators.BooleanValidator;
-import org.apache.drill.exec.server.options.TypeValidators.DoubleValidator;
-import org.apache.drill.exec.server.options.TypeValidators.LongValidator;
-import org.apache.drill.exec.server.options.TypeValidators.StringValidator;
+import org.apache.drill.common.exceptions.UserException;
 
-public abstract class BaseOptionManager implements OptionSet {
-//  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(BaseOptionManager.class);
+import java.util.Iterator;
 
-  /**
-   * Gets the current option value given a validator.
-   *
-   * @param validator the validator
-   * @return option value
-   * @throws IllegalArgumentException - if the validator is not found
-   */
-  private OptionValue getOptionSafe(OptionValidator validator)  {
-    OptionValue value = getOption(validator.getOptionName());
-    return value == null ? validator.getDefault() : value;
+/**
+ * This {@link OptionManager} implements some the basic methods and should be extended by concrete implementations.
+ */
+public abstract class BaseOptionManager extends BaseOptionSet implements OptionManager {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(BaseOptionManager.class);
+
+  @Override
+  public OptionList getInternalOptionList() {
+    return getAllOptionList(true);
   }
 
   @Override
-  public boolean getOption(BooleanValidator validator) {
-    return getOptionSafe(validator).bool_val;
+  public OptionList getPublicOptionList() {
+    return getAllOptionList(false);
   }
 
   @Override
-  public double getOption(DoubleValidator validator) {
-    return getOptionSafe(validator).float_val;
+  public void setLocalOption(final String name, final boolean value) {
+    setLocalOption(name, Boolean.valueOf(value));
   }
 
   @Override
-  public long getOption(LongValidator validator) {
-    return getOptionSafe(validator).num_val;
+  public void setLocalOption(final String name, final long value) {
+    setLocalOption(name, Long.valueOf(value));
   }
 
   @Override
-  public String getOption(StringValidator validator) {
-    return getOptionSafe(validator).string_val;
+  public void setLocalOption(final String name, final double value) {
+    setLocalOption(name, Double.valueOf(value));
   }
 
+  @Override
+  public void setLocalOption(final String name, final String value) {
+    setLocalOption(name, (Object) value);
+  }
+
+  @Override
+  public void setLocalOption(final String name, final Object value) {
+    final OptionDefinition definition = getOptionDefinition(name);
+    final OptionValidator validator = definition.getValidator();
+    final OptionMetaData metaData = definition.getMetaData();
+    final OptionValue.AccessibleScopes type = definition.getMetaData().getType();
+    final OptionValue.OptionScope scope = getScope();
+    checkOptionPermissions(name, type, scope);
+    final OptionValue optionValue = OptionValue.create(type, name, value, scope);
+    validator.validate(optionValue, metaData, this);
+    setLocalOptionHelper(optionValue);
+  }
+
+  @Override
+  public void setLocalOption(final OptionValue.Kind kind, final String name, final String valueStr) {
+    Object value;
+
+    switch (kind) {
+      case LONG:
+        value = Long.valueOf(valueStr);
+        break;
+      case DOUBLE:
+        value = Double.valueOf(valueStr);
+        break;
+      case STRING:
+        value = valueStr;
+        break;
+      case BOOLEAN:
+        value = Boolean.valueOf(valueStr);
+        break;
+      default:
+        throw new IllegalArgumentException(String.format("Unsupported kind %s", kind));
+    }
+
+    setLocalOption(name, value);
+  }
+
+  private static void checkOptionPermissions(String name, OptionValue.AccessibleScopes type, OptionValue.OptionScope scope) {
+    if (!type.inScopeOf(scope)) {
+      throw UserException.permissionError()
+        .message(String.format("Cannot change option %s in scope %s", name, scope))
+        .build(logger);
+    }
+  }
+
+  abstract protected void setLocalOptionHelper(OptionValue optionValue);
+
+  abstract protected OptionValue.OptionScope getScope();
+
+  private OptionList getAllOptionList(boolean internal)
+  {
+    Iterator<OptionValue> values = this.iterator();
+    OptionList optionList = new OptionList();
+
+    while (values.hasNext()) {
+      OptionValue value = values.next();
+
+      if (getOptionDefinition(value.getName()).getMetaData().isInternal() == internal) {
+        optionList.add(value);
+      }
+    }
+
+    return optionList;
+  }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/DrillConfigIterator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/DrillConfigIterator.java
@@ -21,7 +21,7 @@ import java.util.Iterator;
 import java.util.Map.Entry;
 
 import org.apache.drill.common.config.DrillConfig;
-import org.apache.drill.exec.server.options.OptionValue.OptionType;
+import org.apache.drill.exec.server.options.OptionValue.AccessibleScopes;
 import org.apache.drill.exec.server.options.OptionValue.OptionScope;
 
 import com.typesafe.config.ConfigValue;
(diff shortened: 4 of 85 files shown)


paul-rogers:
Please add a brief description in the PR title.

This PR has commits from other PRs, and has merge conflicts. If this PR depends on others, then please identify the commits to be reviewed, and which commits will be removed once this one is rebased onto master with those other commits.

ilooner:
@paul-rogers Applied your comments and cleaned up the PR, it should be ready for review now. If additional work needs to be done to make this easier to review please let me know.

Thanks,
Tim

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -70,11 +69,13 @@
  *  </p>
  */
 
-public class SystemOptionManager extends BaseOptionManager implements OptionManager, AutoCloseable {
+public class SystemOptionManager extends BaseOptionManager implements AutoCloseable {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SystemOptionManager.class);
 
-  private CaseInsensitiveMap<OptionValidator> VALIDATORS;
-  public void populateValidators() {
+  public static final CaseInsensitiveMap<OptionValidator> DEFAULT_VALIDATORS =
```
paul-rogers: This change reverses a change made by Jyothsna. While it seems reasonable to create a static table of validators, there is a subtle problem which Jyothsna's fix addressed.

Under normal operation, we have exactly one Drillbit per process and thus exactly one system option manager.

But, we have unit tests that create multiple Drillbits per process. This is the most convenient way to test concurrency issues in unit tests. In general, each Drillbit could define a different default value for the system/session option; perhaps to check forward/backward compatibility when the time comes to support rolling upgrades.

In this model, each Drillbit will have a slightly different set of defaults. Thus, the option validator table must be a member variable of the system option manager; it cannot be global.

Note also, in the existing tests with multiple Drillbits, a global table runs into a race condition during reading defaults from the config file if we use a single table. But, if we have a separate table per Drillbit, the race condition disappears.

paul-rogers: I wonder, was this an intentional regression, or has your code not been rebased onto the latest master (with Jythosna's changes) and so your commit reintroduces old code?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/StatusResources.java
```diff
@@ -50,75 +53,124 @@
 public class StatusResources {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(StatusResources.class);
 
+  public static final String PATH_STATUS_JSON = "/status.json";
+  public static final String PATH_STATUS = "/status";
+  public static final String PATH_OPTIONS_JSON = "/options.json";
+  public static final String PATH_INTERNAL_OPTIONS_JSON = "/internal_options.json";
+  public static final String PATH_OPTIONS = "/options";
+  public static final String PATH_INTERNAL_OPTIONS = "/internal_options";
```
paul-rogers: Nice improvement to externalize constants. Perhaps the constants can show the relationships between paths. For example:

```
public static final String REST_API_SUFFIX = ".json";
public static final String PATH_STATUS_UI = "/status";
public static final String PATH_STATUS_API = PATH_STATUS_UI + REST_API_SUFFIX;
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/StatusResources.java
```diff
@@ -50,75 +53,124 @@
...
+  {
     List<OptionWrapper> options = new LinkedList<>();
-    for (OptionValue option : work.getContext().getOptionManager()) {
+    OptionManager optionManager = work.getContext().getOptionManager();
+    OptionList optionList = internal ? optionManager.getInternalOptionList(): optionManager.getExternalOptionList();
+
+    for (OptionValue option : optionList) {
       options.add(new OptionWrapper(option.name, option.getValue(), option.type, option.kind));
     }
+
```
paul-rogers: FWIW, DRILL-5716 introduces name sorting to this page. (I can never find anything in the original unsorted list...)

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/OptionIterator.java
```diff
@@ -36,7 +36,7 @@
 //  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(OptionIterator.class);
 
   enum Mode {
-    BOOT, SYS_SESS, BOTH
+    BOOT, SYS_SESS_EXTERNAL, SYS_SESS_INTERNAL, BOTH
```
paul-rogers: Can we improve these names? Maybe `RUNTIME` for system/session option.

I'm not clear on the distinction between external and internal. Do we mean something like public and internal?

Does the external mean only public, while internal means public and private?

And, given that we have three options, `BOTH` seems a bit of a misnomer...

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -421,13 +422,13 @@
    * such as changing system options.
    */
   String ADMIN_USERS_KEY = "security.admin.users";
-  StringValidator ADMIN_USERS_VALIDATOR = new StringValidator(ADMIN_USERS_KEY, true);
+  StringValidator ADMIN_USERS_VALIDATOR = new StringValidator(ADMIN_USERS_KEY, new OptionValue.MetaData(true, false));
```
paul-rogers: We are adding a distinction between runtime options that the user can set vs. those that the user cannot set? This is a very bizarre distinction as runtime options are only for the user...

Actually, these are options that are "public" vs. those that exist, but are generally "hidden", right?

Another very useful addition, while we're in the code, is between the allowed "scope" of a setting. Today, all options can be set at the system and session (per user) level. But some options, such as the memory per query, or enabling queues, really only make sense at the system level. So, part of the new metadata should be to specify "session-only" options and forbid setting such options at the session level.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
paul-rogers: A compromise is to introduce a two-level table.

* Global table of option metadata. Perhaps evolve the validator to be that metadata. Or, better, define a new `OptionDefinition` class, an attribute of which is the validator.
* Per-Drillbit table of defaults resolved from the config system using the information from the definitions.

If we do this, we should clean up all the messy enums in this system. The current ones confuse everyone. We want:

* Definition: The allowed scope: session-only or both session and system.
* Visibility: public or hidden (this change)
* Data type: (long, double, string, etc.)

For a value:

* Scope: default, session or system.

Sometimes it is easier to fix the original problems than to keep working around an outdated design...

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -83,18 +86,20 @@
   @VisibleForTesting
   public Drillbit(
       final DrillConfig config,
+      final CaseInsensitiveMap<OptionValidator> validators,
       final RemoteServiceSet serviceSet) throws Exception {
-    this(config, serviceSet, ClassPathScanner.fromPrescan(config));
+    this(config, validators, serviceSet, ClassPathScanner.fromPrescan(config));
```
paul-rogers: Interesting... Why was this change needed?

To be honest, I'm not convinced that the set of inputs is an *input* to the Drillbit rather than an internal *attribute* of the Drillbit. That is, the Drillbit should be responsible for defining its set of valid options, not the code that houses the Drillbit.

exec/java-exec/src/test/java/org/apache/drill/exec/server/rest/StatusResourcesTest.java
```diff
@@ -0,0 +1,80 @@
...
+
+    try (ClusterFixture cluster = builder.build();
+         ClientFixture client = cluster.clientFixture()) {
+      RestClientFixture restClientFixture = cluster.restClientFixture();
+
+      Assert.assertNull(restClientFixture.getStatusOption(MOCK_PROPERTY));
+      StatusResources.OptionWrapper option = restClientFixture.getStatusInternalOption(MOCK_PROPERTY);
+      Assert.assertEquals("a", option.getValueAsString());
+
+      client.alterSystem(MOCK_PROPERTY, "c");
```
paul-rogers: This test shows the value in allowing "extended" options defined outside of the usual big table in `SystemOptionManager`.

In fact, we can take a step back and say that having a global table is always a bad idea: it introduces tight coupling. Tight coupling makes it much harder to do unit testing. For an open source tool such as Drill, which should allow extensions, tight coupling is counter-productive to the goal of extensibility.

So, I wonder, can we add a mechanism that allows other modules (and tests) to add options? Just brainstorming a bit...

* For code modules, provide a config setting that names a class that builds options. The Drillbit, during initialization, scans for these classes (the way it scans for, say, storage plugins), instantiates the class, and calls a method to define the options.
* The easiest way to define options is to add an `OptionBuilder` class, created by the Drillbit in the constructor, that is available up to some point in time, such as once the Drillbit starts running.

The API might be:

```
Drillbit bit = new Drillbit();
bit.optionBuilder().add(new OptionDefinition("my.option", ...));
bit.run();
```

The `add()` method can be called from tests (perhaps with added `ClusterFixture` support) and from the module-specific option builder.

Two issues in this model:

* Is there any reason to add options after the Drillbit starts? Doing so adds complexity.
* Do we have code in Drillbit startup that uses the option manager? If so, some careful sequencing would be needed to call this code only *after* the options are built.

exec/java-exec/src/test/java/org/apache/drill/exec/work/metadata/TestMetadataProvider.java
```diff
@@ -211,12 +217,15 @@ public void tablesWithTableNameFilter() throws Exception {
...
-    assertEquals(5, tables.size());
+    assertEquals(7, tables.size());
 
     verifyTable("sys", "boot", tables);
     verifyTable("sys", "memory", tables);
     verifyTable("sys", "options", tables);
+    verifyTable("sys", "options2", tables);
     verifyTable("sys", "version", tables);
+    verifyTable("sys", "internal_options", tables);
+    verifyTable("sys", "internal_options2", tables);
```
paul-rogers: Better name? Nobody is going to know what "options2" means... I certainly can't guess...

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
```diff
@@ -90,13 +93,13 @@
   public static class OperatorFixtureBuilder
   {
     ConfigBuilder configBuilder = new ConfigBuilder();
-    TestOptionSet options = new TestOptionSet();
+    TestOptionManager options = new TestOptionManager();
```
paul-rogers: This is a bit tricky. Some background.

Most of our existing "unit" (really, system) tests start a Drillbit in order to test anything.

We are trying (slowly!) to move to true unit tests. To do so, we need to introduce lose coupling between modules. The full option managers have dependencies on the Drillbit which sucks in all of Drill.

The goal of the `OptionSet` is to provide a read-only view of the options since accessing options is pretty simple; it does not have dependencies on the Drillbit.

We need to preserve this view to allow unit testing. We can do that either by ensuring that the option managers are stand-alone (have no runtime dependencies on the Drillbit or the DrillbitContext), or by continuing to do what this code does: create a test-time, read-only mock version that implements the option reader (`OptionSet`) interface.

Either are valid approaches; which do you think would work best?

exec/java-exec/src/test/java/org/apache/drill/test/RestClientFixture.java
```diff
@@ -0,0 +1,115 @@
...
+import org.apache.drill.exec.server.rest.StatusResources;
+
+import javax.annotation.Nullable;
+
+import static io.restassured.RestAssured.when;
+
+/**
+ * Represents a client for the Drill Rest API.
+ */
+public class RestClientFixture {
```
paul-rogers: Very nice addition!

exec/java-exec/src/test/java/org/apache/drill/test/RestClientFixture.java
```diff
@@ -0,0 +1,115 @@
...
+  public static final String REST_API_PORT = "drill.exec.http.port";
```
paul-rogers: Isn't there already an `ExecConstants` value for this option? See `ExecConstants.HTTP_PORT`.

exec/java-exec/src/test/java/org/apache/drill/test/RestClientFixture.java
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+
+  private RestClientFixture(ClusterFixture cluster) {
+    int port = cluster.config.getInt(REST_API_PORT);
+    String address = cluster.drillbits().iterator().next().getContext().getEndpoint().getAddress();
+
+    RestAssured.port = port;
+    RestAssured.basePath = "/";
+    RestAssured.baseURI = "http://" + address;
```
paul-rogers: Are these global variables? How will this work in Surefire, when multiple tests run concurrently, if someone uses this mechanism to, say, test HTTPS support in one test while this runs in another test?

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
ilooner: Agreed we should not be using global variables. The intended flow was the following:

- DEFAULT_VALIDATORS holds the default set of validators
- The system option manager is passed a set of validators when it is initialized
- The validators map is copied and stored in the validators variable on line 230. So each option manager has a separate copy of the validator map.

Looking at this again I agree that this is not ideal. The DEFAULT_VALIDATORS constant should be deleted and usages should be replaced with createDefaultValidators.

I also see another issue that was present before where the validators themselves are static, they should be explicitly copied, so we should probably add a clone method to the validators to do that.

ilooner: Not a regression see comment above.

ilooner: I don't completely follow this. Let's talk about this offline next week so that I can understand :). I can post the conclusion of the discussion on the PR for the community then.

exec/java-exec/src/test/java/org/apache/drill/exec/server/rest/StatusResourcesTest.java
ilooner: I added a method that allows the addition of option validators at startup, and those validators are passed to the SystemOption manager. I think this could be enhanced using the OptionBuilder mechanism you described.

- I don't think it is useful to add options after starting a Drillbit, but we could apply the OptionBuilder to add options at boot time.
- That should be possible. It was possible to add option validators to the SystemOptionManager at boot time.

exec/java-exec/src/test/java/org/apache/drill/test/RestClientFixture.java
ilooner: Yeah the rest-assured library uses global variables, which I don't like either. But it gives some nice additionally functionality for testing. The global configuration shouldn't impact us though because we run tests concurrently by forking separate jvms, and running tests sequentially within each jvm. If you see this being a larger issue later when we potentially run tests concurrently within a jvm, I can flip back to using the jersey-client which does not have this problem.

ilooner:
Thanks @paul-rogers I'll work on addressing the comments and talk to you next week about some questions I have.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
ilooner: Currently there are no internal validators defined, so in order to test the logic for internal validators I had to create a set of mock validators which included internal validators, and use those for testing. I have changed this so that the Drillbit constructors which take a set of validators are annotated to be visible for testing.

exec/java-exec/src/test/java/org/apache/drill/test/RestClientFixture.java
ilooner: I reverted to using the jersey client so global variables aren't used anymore.

ilooner:
So it looks like Drill's treatment of OptionType and OptionScope is inconsistent at present. To correct this I am doing the following:

- OptionType seems to be I'll defined. I think the intention is that it defines the levels in which an option can be configured. This should be a static property tied to an option (much like a validator). In reality it looks like the OptionType of an option can be changed depending on where it is set, which is incorrect behavior. I think this came about because it is the user's responsibility to define the OptionType for an option value which is in correct.
- OptionScope will define the level at which the option was set

To solidify the semantics I am doing the following:

- OptionManagers no longer allow an option value to be set. Only the following methods are exposed: ```
void setLocalOption(String name, boolean value);
void setLocalOption(String name, long value);
void setLocalOption(String name, double value);
void setLocalOption(String name, String value);
void setLocalOption(OptionValue.Kind kind, String name, String value);
``` Doing this prevents the possibility of defining an incorrect scope or type. The option type is now determined by the static option definition. The option scope is determined by the option manager on which the option was set.

- I have also cleaned up the OptionManager interface so that it does not require the option type to be passed to the delete methods. This helps clean up the semantics further because OptionType was being overloaded to define both where an option can be set and which option manager can set it. This is unclear because some options can be set exclusively at one level (e.g. only the session level) whereas other options can be set at multiple levels (e.g. system and session levels). It also introduces the possibility for error by allowing the user to provide the wrong type to the wrong option manager.

**Comment by ilooner** [comment]
@paul-rogers Finished applying comments, and cleanup. It's ready for review again now.

File: exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java [hunk 13]
```diff
@@ -346,17 +347,63 @@ public void deleteAllOptions(OptionType type) {
     }
   }
 
-  public void populateDefaultValues() {
-
+  public static CaseInsensitiveMap<OptionValue> populateDefualtValues(Map<String, OptionDefinition> definitions, DrillConfig bootConfig) {
```
paul-rogers: `populateDefualtValues` --> `populateDefaultValues` (that is "ua" --> "au").

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
```diff
@@ -102,10 +103,11 @@ public void testFix2967() throws Exception {
       test("select * from dfs_test.`%s/join/j1` j1 left outer join dfs_test.`%s/join/j2` j2 on (j1.c_varchar = j2.c_varchar)",
         TEST_RES_PATH, TEST_RES_PATH);
     } finally {
-      setSessionOption(PlannerSettings.BROADCAST.getOptionName(), String.valueOf(PlannerSettings.BROADCAST.getDefault().bool_val));
-      setSessionOption(PlannerSettings.HASHJOIN.getOptionName(), String.valueOf(PlannerSettings.HASHJOIN.getDefault().bool_val));
+      final OperatorFixture.TestOptionSet testOptionSet = new OperatorFixture.TestOptionSet();
+      setSessionOption(PlannerSettings.BROADCAST.getOptionName(), String.valueOf(testOptionSet.getDefault(PlannerSettings.BROADCAST.getOptionName()).bool_val));
```
paul-rogers: No need to fix now but I wonder:

* The `ClusterFixture` class has a method of form `setSystemOption(String name, Option value)` to avoid this kind of silly to/from String overhead. We might want to do the same for these "legacy" `BaseTestQuery` style tests.
* But, here we are setting system options to their default value. Why? Isn't that already the value? If not, wouldn't clearing (deleting) the option value be better?
* Or, has the test previously set a system option to other than the default value, and we are undoing that here? Seems it would be better to allow a test to say not to muck with the defaults. (Hard to do with `BaseTestQuery`, but there is a built-in provision for this in the `ClusterFixture`.)

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLimitWithExchanges.java
```diff
@@ -71,7 +72,8 @@ public void testPushLimitPastUnionExchange() throws Exception {
       final String[] expectedPlan5 = {"(?s)Limit\\(fetch=\\[1\\].*UnionExchange.*Limit\\(fetch=\\[1\\]\\).*Join"};
       testLimitHelper(sql5, expectedPlan5, excludedPlan, 1);
     } finally {
-      test("alter session set `planner.slice_target` = " + ExecConstants.SLICE_TARGET_OPTION.getDefault().getValue());
+      final OperatorFixture.TestOptionSet testOptionSet = new OperatorFixture.TestOptionSet();
+      test("alter session set `planner.slice_target` = " + testOptionSet.getDefault(ExecConstants.SLICE_TARGET).getValue());
```
paul-rogers: Here, seems it would be better to refer to the constant defining the option name rather than hard-coding it. (Yes, this is existing code, so OK to leave as is...)

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptions.java
```diff
@@ -56,7 +56,7 @@ public void checkChangedColumn() throws Exception {
     test("ALTER session SET `%s` = %d;", SLICE_TARGET,
       ExecConstants.SLICE_TARGET_DEFAULT);
     testBuilder()
-        .sqlQuery("SELECT status FROM sys.options WHERE name = '%s' AND type = 'SESSION'", SLICE_TARGET)
+        .sqlQuery("SELECT status FROM sys.options WHERE name = '%s' AND optionScope = 'SESSION'", SLICE_TARGET)
```
paul-rogers: Hmmm... I wonder about this. Aside from our own tests, might users have existing code or scripts that know about the existing schema? It is for this reason that Jyothsna added a new table for the enhanced option data, going out of her way to preserve the public schema of the existing tables. I wonder, should we do that here? Or, are we fortunate and no one depends on the current system table schema?

exec/java-exec/src/test/java/org/apache/drill/exec/testing/TestExceptionInjection.java
```diff
@@ -216,79 +216,77 @@ public void injectionOnSpecificBit() {
...
-    final DrillbitContext drillbitContext2 = drillbit2.getContext();
+      final DrillbitContext drillbitContext1 = drillbit1.getContext();
+      final DrillbitContext drillbitContext2 = drillbit2.getContext();
 
-    final UserSession session = UserSession.Builder.newBuilder()
-        .withCredentials(UserBitShared.UserCredentials.newBuilder().setUserName("foo").build())
-        .withUserProperties(UserProperties.getDefaultInstance())
-        .withOptionManager(drillbitContext1.getOptionManager())
-        .build();
+      final UserSession session = UserSession.Builder.newBuilder().withCredentials(UserBitShared.UserCredentials.newBuilder().setUserName("foo").build()).withUserProperties(UserProperties.getDefaultInstance()).withOptionManager(drillbitContext1.getOptionManager()).build();
```
paul-rogers: Did we want to combine all the method calls onto one line rather than the fluent, multi-line style originally in the code? An artifact of a code formatter?

exec/java-exec/src/test/java/org/apache/drill/exec/testing/TestPauseInjection.java
```diff
@@ -150,66 +150,61 @@ public void pauseOnSpecificBit() {
...
       }
-    }
 
-    {
-      final ExtendedLatch trigger = new ExtendedLatch(1);
-      final QueryContext queryContext = new QueryContext(session, drillbitContext2, QueryId.getDefaultInstance());
+      final DrillbitContext drillbitContext1 = drillbit1.getContext();
+      final DrillbitContext drillbitContext2 = drillbit2.getContext();
+
+      final UserSession session = UserSession.Builder.newBuilder().withCredentials(UserCredentials.newBuilder().setUserName("foo").build()).withUserProperties(UserProperties.getDefaultInstance()).withOptionManager(drillbitContext1.getOptionManager()).build();
```
paul-rogers: Same issue of one big line vs. multi-line.

exec/java-exec/src/test/java/org/apache/drill/exec/work/metadata/TestMetadataProvider.java
```diff
@@ -196,9 +200,12 @@ public void tablesWithSystemTableFilter() throws Exception {
     verifyTable("sys", "boot", tables);
     verifyTable("sys", "drillbits", tables);
     verifyTable("sys", "memory", tables);
-    verifyTable("sys", "options", tables);
+    verifyTable("sys", SystemTable.OPTION.getTableName(), tables);
+    verifyTable("sys", SystemTable.OPTION_VAL.getTableName(), tables);
     verifyTable("sys", "threads", tables);
     verifyTable("sys", "version", tables);
+    verifyTable("sys", SystemTable.INTERNAL_OPTIONS.getTableName(), tables);
+    verifyTable("sys", SystemTable.INTERNAL_OPTIONS_VAL.getTableName(), tables);
```
paul-rogers: Thanks for providing the constants. Nice improvement.

exec/java-exec/src/test/java/org/apache/drill/test/RestClientFixture.java
```diff
@@ -0,0 +1,117 @@
...
+      this.cluster = Preconditions.checkNotNull(cluster);
+    }
+
+    public RestClientFixture build() {
+      return new RestClientFixture(cluster);
+    }
+  }
+
+  private WebTarget baseTarget;
+  private Client client;
```
paul-rogers: `final` here and above?

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptions.java
ilooner: Yeah we are in a tough spot because **type** was not well defined previously, so the tests implicitly assumed **type** represented where an option was set. Now that we have settled on a definition for **type**, which is the set of scopes where an option can be set, we have deviated from the meaning **type** was given in the tests. One possible way out of this situation is to change the definition of **type** and **optionScope** again by swapping their meanings:

* **type**: Would become where an option was set.
* **optionScope**: Would become the set of scopes where an option could be set.

This would minimize the changes required to the unit tests. It's hard to say how it would impact other user's scripts though because **type** was treated inconsistently in the code base, so I'm not sure how someone could have used the **type** information productively except to write unit tests, which verified incorrect behavior.

Long story short, I'll swap the definitions of **type** and **optionScope**. I think that would minimize the impact on unit tests and users, but we cannot provide any guarantees for users who depended on type when it was inconsistently defined.

exec/java-exec/src/test/java/org/apache/drill/test/RestClientFixture.java
ilooner: done

exec/java-exec/src/test/java/org/apache/drill/exec/testing/TestPauseInjection.java
ilooner: done

exec/java-exec/src/test/java/org/apache/drill/exec/testing/TestExceptionInjection.java
ilooner: I'm a victim of autoformat. Fixed.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLimitWithExchanges.java
ilooner: done

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
ilooner: The test changed some of the options for the session. Since the same drillbit cluster is used for multiple tests, the changes need to be undone at the end of the test to avoid impacting other tests, which reuse the drill bit cluster. I basically kept things as they were before, but looking at it again I agree this is a pretty messy way of doing it. I think this could be cleaned up by doing:

```
ALTER SESSION RESET ALL;
```

To clear all the changes at the end of the test. Instead of passing an option value I'll also add methods for each type: 

```
setSessionOption(final String option, final boolean value)
setSessionOption(final String option, final long value)
setSessionOption(final String option, final double value)
setSessionOption(final String option, final String value)
```

ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/OptionValue.java
```diff
@@ -63,32 +88,32 @@
...
     return new OptionValue(Kind.STRING, type, name, null, val, null, null, scope);
   }
 
-  public static OptionValue createDouble(OptionType type, String name, double val, OptionScope scope) {
+  public static OptionValue create(OptionType type, String name, double val, OptionScope scope) {
     return new OptionValue(Kind.DOUBLE, type, name, null, null, null, val, scope);
   }
 
-  public static OptionValue createOption(Kind kind, OptionType type, String name, String val, OptionScope scope) {
+  public static OptionValue create(Kind kind, OptionType type, String name, String val, OptionScope scope) {
```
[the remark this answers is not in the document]
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/OptionMetaData.java
```diff
@@ -0,0 +1,68 @@
...
+package org.apache.drill.exec.server.options;
+
+/**
+ * Contains information about the scopes in which an option can be set, and an option's visibility.
+ */
+public class OptionMetaData {
+  public static final OptionMetaData DEFAULT = new OptionMetaData(OptionValue.OptionType.ALL, false, false);
+
+  private OptionValue.OptionType type;
+  private boolean adminOption;
```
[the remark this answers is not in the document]
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/OptionMetaData.java
```diff
@@ -0,0 +1,68 @@
...
+
+package org.apache.drill.exec.server.options;
+
+/**
+ * Contains information about the scopes in which an option can be set, and an option's visibility.
+ */
+public class OptionMetaData {
+  public static final OptionMetaData DEFAULT = new OptionMetaData(OptionValue.OptionType.ALL, false, false);
+
+  private OptionValue.OptionType type;
```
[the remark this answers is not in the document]
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/OptionManager.java
```diff
@@ -17,49 +17,97 @@
...
    */
   OptionList getOptionList();
+
+  /**
+   * Returns all the internal options contained in this option manager.
+   *
+   * @return All the internal options contained in this option manager.
+   */
+  @NotNull
+  OptionList getInternalOptionList();
```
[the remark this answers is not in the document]
ilooner: Internal and Local are not the same thing. Local means that the value for the option is stored in this option manager. Internal means that it is not visible with the rest of the options unless you look in the special sys.internal_options table. I will add this distinction to the javadoc.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/DrillConfigIterator.java
```diff
@@ -58,17 +58,17 @@ public OptionValue next() {
       OptionValue optionValue = null;
       switch(cv.valueType()) {
       case BOOLEAN:
-        optionValue = OptionValue.createBoolean(OptionType.BOOT, name, (Boolean) cv.unwrapped(), OptionScope.BOOT);
+        optionValue = OptionValue.create(OptionType.BOOT, name, (Boolean) cv.unwrapped(), OptionScope.BOOT);
```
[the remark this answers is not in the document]
ilooner: Agreed

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/BaseOptionManager.java
```diff
@@ -17,44 +17,84 @@
...
   }
 
+  @Override
+  public void setLocalOption(String name, String value) {
+    setLocalOption(OptionValue.Kind.STRING, name, value);
+  }
+
+  @Override
+  public void setLocalOption(OptionValue.Kind kind, String name, String value) {
+    final OptionDefinition definition = getOptionDefinition(name);
```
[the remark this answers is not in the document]
ilooner: Yeah it does. I've updated the javadoc to make this explicit.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/BaseOptionManager.java
```diff
@@ -17,44 +17,84 @@
...
-    return getOptionSafe(validator).bool_val;
+  public OptionList getPublicOptionList() {
+    return getAllOptionList(false);
   }
 
   @Override
-  public double getOption(DoubleValidator validator) {
-    return getOptionSafe(validator).float_val;
+  public void setLocalOption(String name, boolean value) {
+    setLocalOption(OptionValue.Kind.BOOLEAN, name, Boolean.toString(value));
```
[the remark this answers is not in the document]
ilooner: Done.

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptions.java
ilooner: Taking a note of what we discussed offline. 

Course of action is to just change the name for **type** since it is a bad name, and the semantics of **type** have been ill-defined. So no one could have relied on the values returned to them in the past anyway. I am changing the name to **accessibleScopes**.

ilooner:
@paul-rogers Finished apply review comments. PR is ready for review again.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/OptionManager.java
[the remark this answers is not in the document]
paul-rogers: Thanks for the explanation.

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptions.java
paul-rogers: Sounds good. Thanks.

exec/java-exec/src/test/java/org/apache/drill/exec/server/rest/StatusResourcesTest.java
paul-rogers: Thanks.

ilooner:
Thanks @paul-rogers squashed and ready to commit. Please merge this change before my other PR https://github.com/apache/drill/pull/940

PR #924, opened 2017-08-26
DRILL-5740: Ensure spill directories are unique

paul-rogers:
A recent change added the node name and port to make the spill path
unique. Turns out we need to add this information to the single spill
directory name. The previous change use the node ID as a parent
directory, which turns out not to work well in practice.


jinfengni:
+1

LGTM.

PR #925, opened 2017-08-29
DRILL-5749: solve foreman and netty threads deadlock

weijietong:
break the nest invocation of channelClosed method to avoid nested lock holding


weijietong:
anyone can review this PR ? more details about thread stacks see [DRLL-5749](https://issues.apache.org/jira/browse/DRILL-5749)

exec/rpc/src/main/java/org/apache/drill/exec/rpc/RequestIdMap.java
```diff
@@ -54,10 +52,14 @@ void channelClosed(Throwable ex) {
     isOpen.set(false);
     if (ex != null) {
       final RpcException e = RpcException.mapException(ex);
+      IntObjectHashMap<RpcOutcome<?>> clonedMap;
       synchronized (map) {
-        map.forEach(new SetExceptionProcedure(e));
+        clonedMap = map.clone();
         map.clear();
       }
+      if (clonedMap != null) {
```
paul-rogers: When would `clonedMap` be `null`?

weijietong: @paul-rogers clonedMap will not be null from the current code logic ,just for conservative safe. If it seems redundancy ,I will remove the null check.

paul-rogers: Please do. The if statement is a message to readers that clonedMap could be null and so we must try to sleuth out the conditions under which that occurs. Otherwise, I just go ahead and assume that Java is deterministic and that, once we set variable x in a single-threaded environment, it stays at that value until we change it...

sudheeshkatkam:
Nice catch @weijietong

Looks like the Netty thread, that is failing all outstanding requests to a certain endpoint, is itself trying to send a message (cancel fragment) to that endpoint. Any ideas on how to fix that?

Apart from Paul's comment, the current patch looks good to me as well. The above fix is a nice-to-have.

weijietong:
@paul-rogers have refactored the codes. @sudheeshkatkam nothing to fix ,once the netty thread got the re-spawned RPC connection ,it will send out the message.

paul-rogers:
Turns out that I cannot cleanly do the rebase. @weijietong, please rebase this branch onto the latest master, and squash your commits into a single commit.

PR #926, opened 2017-08-30
DRILL-5269 Make DirectSubScan Jackson JSON deserializable

vrozov:
@amansinha100 Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PhysicalPlanReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PhysicalPlanReader.java
@@ -35,12 +35,15 @@ import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
 import org.apache.drill.exec.record.MajorTypeSerDe;
 import org.apache.drill.exec.server.options.OptionList;
 import org.apache.drill.exec.store.StoragePluginRegistry;
+import org.apache.drill.exec.store.pojo.DynamicPojoRecordReader;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.InjectableValues;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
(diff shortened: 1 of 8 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordReader.java
```diff
@@ -26,6 +26,9 @@
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.vector.ValueVector;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+@JsonTypeInfo(use=JsonTypeInfo.Id.NAME, include=JsonTypeInfo.As.WRAPPER_OBJECT, property="type")
```
amansinha100: I am not a Jackson expert; since this annotation is in the core RecordReader interface, what impact does it have on all the implementers of this interface ?

vrozov: AFAIK, it will not impact implementations of the `RecordReader` interface. The annotation affects Jackson type inference when a physical operator has a reference to a `RecordReader` and Jackson needs to construct a concrete implementation of the `RecordReader`. Such information needs to be passed in any case and the annotation specifies JSON syntax used to pass the type information. To avoid the concern I'll move the annotation to the field declaration in `DirectSubScan`. The effect of such move is that it will be necessary to annotate `RecordReader` in every operator that may be passed as part of a fragment to a remote node for execution.

vrozov:
@paul-rogers I guess you refer to a system/integration test (execute the query provided in JIRA), not a unit test.

paul-rogers:
On the first comment, it sounds like the close is a placeholder, actual close is to be done after some refactoring? This is fine. Just a bit surprising to see a log message, but not actual close.

On the second, yes, Drill uses "unit test" to mean a system test run via JUnit. Can we create a JUnit-based system test that demonstrates the problem and fix? I'm guessing that we can do the test at the system level to avoid the difficulty, at present, of writing true unit tests against Drill code.

vrozov:
@paul-rogers Added unit test and adopted query to an existing parquet file.

vrozov:
rebased and squashed

PR #927, opened 2017-08-30
DRILL-5751: Fix unit tests to use local file system even if it is not…

arina-ielchiieva:
… set by default

Please refer to [DRILL-5751](https://issues.apache.org/jira/browse/DRILL-5751) for details.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/util/FileSystemUtilTestBase.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/util/FileSystemUtilTestBase.java
@@ -22,7 +22,6 @@ import org.apache.commons.io.FileUtils;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.junit.AfterClass;
 import org.junit.BeforeClass;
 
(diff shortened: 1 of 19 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/util/FileSystemUtilTestBase.java
```diff
@@ -64,7 +63,9 @@
   @BeforeClass
   public static void setup() throws IOException {
     // initialize file system
-    fs = FileSystem.get(new Configuration());
+    Configuration configuration = new Configuration();
+    configuration.set(FileSystem.FS_DEFAULT_NAME_KEY, FileSystem.DEFAULT_FS);
+    fs = FileSystem.get(configuration);
```
vvysotskyi: These three lines appear several times in the code, so may be it would be better to create a static method which returns `FileSystem` instance.

arina-ielchiieva: Agree. Added static method `ExecTest.getLocalFileSystem`.

vvysotskyi:
LGTM! +1.

amansinha100:
+1

PR #929, opened 2017-08-31
DRILL-5757: CONVERT_TO_JSON function is failed while using non-existe…

vdiravka:
…nce field as a parameter.


arina-ielchiieva:
LGTM, +1

PR #930, opened 2017-09-01
DRILL-5761: Disable Lilith ClassicMultiplexSocketAppender by default

vvysotskyi:


arina-ielchiieva:
LGTM. +1

jinfengni:
I have some concern about this PR.

AFASIK, some of Drill developers use Lilith to debug issues when running individual unit testcases.  If remove it,  does it mean people has to manually reverse this PR, in order to get log in  Lilith? 

If you shutdown Lilith application while running unit test suite, does it hang as well?

vvysotskyi:
Unit tests hang when Lilith application is not running, but when `ClassicMultiplexSocketAppender` in `logback.xml` is defined and used in the loggers.

priteshm:
@vvysotskyi is it possible to change the port number for Lilith to avoid the conflict?

vvysotskyi:
@priteshm, as I have mentioned in the Jira [DRILL-5761](https://issues.apache.org/jira/browse/DRILL-5761), port in Lilith UI cannot be changed (https://github.com/huxi/lilith/issues/10).

jinfengni:
Per this hive doc, seems it's possible to change hiveserver2 port #.

Can you try to change hivesever2# if the other one is hardcoded?

1. https://cwiki.apache.org/confluence/display/Hive/Setting+Up+HiveServer2

vvysotskyi:
Yes, we can change hiveserver2 port #, but I think that it would be better to disable Lilith by default since hiveserver2 port number may be used in the configs of other applications and it would be cumbersome to change all that configs.

jinfengni:
We have run unit in our test clusters and I never hear that people complained that unit tests failed due to Lilith.

vvysotskyi:
I have implemented the proposal of @vrozov to conditionally enable Lilith appender. I have used `ThresholdFilter` in Lilith appender, so it was not needed to change all places where it is used.
An example of how to enable Lilith was added to the Jira description.
@jinfengni, @vrozov could you please take a look?

jinfengni:
The proposal from @vrozov makes sense to me. As long as Lilith is not completely disabled, I'm fine with it.

common/src/test/resources/logback.xml
```diff
@@ -16,17 +16,22 @@
     <ReconnectionDelay>10000</ReconnectionDelay>
     <IncludeCallerData>true</IncludeCallerData>
     <RemoteHosts>${LILITH_HOSTNAME:-localhost}</RemoteHosts>
+    <!-- Disables Lilith ClassicMultiplexSocketAppender since its use causes hanging
```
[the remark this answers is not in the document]
vvysotskyi: I agree with you that it does not disable Lilith. The connection is established with HiveServer2 instead of Lilith server and unit tests hang when `MultiplexSendBytesService` calls `sendBytes()` method. 
I have updated changes, so now Lilith is disabled.

common/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,111 @@
...
+      <maxFileSize>100MB</maxFileSize>
+    </triggeringPolicy>
+    <encoder>
+      <pattern>%date{ISO8601} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+  -->
+
+  <logger name="org.apache.drill" additivity="false">
+    <level value="debug"/>
```
paul-rogers: Please put the test debug level at `ERROR`. I have a copy of `logback-test.xml` that I copy into each working branch that sets the level to `ERROR`. Then, specific tests set more detailed logging as needed. Otherwise, the console is bombarded with unwanted logging messages making it very hard to find those of interest.

common/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,111 @@
...
+    <level value="info"/>
+    <if condition='property("drill.lilith.enable").equalsIgnoreCase("true")'>
+      <then>
+        <appender-ref ref="SOCKET"/>
+      </then>
+    </if>
+<!--    <appender-ref ref="FILE" /> -->
+  </logger>
+
+  <logger name="com.mapr" additivity="false">
```
paul-rogers: Should this be in Apache Drill? Only the MapR profile pulls in MapR code... It is not clear that tests that happen to use the MapR profile want debug level logging from this subsystem. Better to use the `LogFixture` to set more detailed logging in those tests that need it.

common/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,111 @@
...
+    <if condition='property("drill.lilith.enable").equalsIgnoreCase("true")'>
+      <then>
+        <appender-ref ref="SOCKET"/>
+      </then>
+    </if>
+  </logger>
+  -->
+
+  <logger name="org.apache.hadoop" additivity="false">
+    <level value="info"/>
```
paul-rogers: See comment below about MapR. It is not clear that we always want info-level logging from Hadoop. Seems this should be enabled only in tests that want that detail to avoid cluttering output with unwanted messages.

common/src/test/resources/logback-test.xml
vvysotskyi: This logger contains only Lilith appender, and since we are using Lilith only for debugging separate tests, I think it would be better to left logging level here at `DEBUG`. In this case, everyone who uses Lilith won't need to change logback file. Console appender is used in the root logger and it has `ERROR` logging level.

common/src/test/resources/logback-test.xml
vvysotskyi: Initially, I assumed that it would be useful for those, who work with the mapr-format-plugin. I agree with you that we should delete this logger.

common/src/test/resources/logback-test.xml
vvysotskyi: Thanks for pointing this, already removed.

common/src/test/resources/logback-test.xml
paul-rogers: OK.

vrozov: Can this logger be defined between line 21 and 22 in a single if condition?

common/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,92 @@
...
+
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <!-- encoders are assigned the type
+         ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+
+  <!--
```
vrozov: Is it necessary to have RollingFileAppender defined for unit test, especially that it is commented anyway?

paul-rogers:
Holding of on the merge waiting for review by @vrozov.

common/src/test/resources/logback-test.xml
vrozov: +1 I think that developers are unlikely to use rolling file appender and can redirect console to a file if necessary.

vvysotskyi:
@paul-rogers, @vrozov, thanks for the review. Squashed commits and rebased onto master.

PR #931, opened 2017-09-01
DRILL-5763: Fix NPE during MapRDBSubScan deserialization

arina-ielchiieva:


vdiravka:
+1 LGTM

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBSubScan.java
```diff
@@ -71,14 +71,32 @@ public MapRDBSubScan(String userName, MapRDBFormatPlugin formatPlugin,
     this.tableType = tableType;
   }
 
+
+  @JsonProperty("formatPluginConfig")
```
amansinha100: There are trade-offs with serializing/deserializing more than the minimum needed.  One is the extra amount of data being sent as part of the plan fragment.  Another, somewhat more problematic is that the serialization brings in dependent classes and if the latter are not serialized properly then query fails.  Any thoughts about these ?  I understand the motivation for the change to fix DRILL-5714 and subsequently this Jira.  Just want to explore it further to see if the downsides can be avoided.

arina-ielchiieva: We need to serialize `formatPluginConfig` to be able to create `MapRDBFormatPlugin` during deserialization.
`MapRDBFormatPluginConfig` contains five fields only:
```
  public boolean allTextMode = false;
  public boolean enablePushdown = true;
  public boolean ignoreSchemaChange = false;
  public boolean readAllNumbersAsDouble = false;
  public boolean disableCountOptimization = false;
```
Regarding the change, DRILL-5714 fixed initially incorrect implementation logic in `MapRDBSubScan`. 
Current implementation is similar to `EasySubScan`.

amansinha100: Ok, yes since the fields are primitive types, they are not adding a dependency on other classes.  
Separately, we need to increase test coverage for these types of serialization/deserialization issues for the most commonly used storage and format plugins but that's a different discussion. 
I am good with this.  +1

PR #932, opened 2017-09-01
DRILL-5758: Fix for repeated columns; enable managed sort by default

paul-rogers:
Fix for DRILL-5443. Basically, the “record batch sizer” did not handle repeated columns correctly.

Enabled managed sort by default.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
@@ -59,10 +59,10 @@ public class RecordBatchSizer {
 
     /**
      * Assumed size from Drill metadata. Note that this information is
-     * 100% bogus. Do not use it.
+     * 100% bogus for variable-width columns. Do not use it for such
+     * columns.
      */
 
-    @Deprecated
     public int stdSize;
 
     /**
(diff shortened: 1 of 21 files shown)


paul-rogers:
@Ben-Zvi, added another fix. Please take another look.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
```diff
@@ -680,9 +680,11 @@ public MergeTask consolidateBatches(long allocMemory, int inMemCount, int spille
     // spilled run.
 
     // Maximum spill batches that fit into available memory.
+    // Use the maximum buffer size since spill batches seem to
+    // be read with almost 50% internal fragmentation.
 
     int memMergeLimit = (int) ((mergeMemoryLimit - allocMemory) /
-                                spillBatchSize.expectedBufferSize);
+                                spillBatchSize.maxBufferSize);
     memMergeLimit = Math.max(0, memMergeLimit);
```
Ben-Zvi: This line of code is not needed ( Math.max(....) ), as the code above (around line 675) ensures that mergeMemoryLimit >= allocMemory .

paul-rogers:
Added a fix for repeated columns that have a low cardinality. If, say, one row in ten has an array entry, then average cardinality (values per row) is 0.1. This was represented by an int, rounded to 0 and caused a zero-length vector to be allocated. Drill then tried to double the length, which resulted in 0, which was doubled again, and so on forever.

The fix has three parts:

* The "record batch sizer" uses floats to allow fractional cardinality.
* The vector initializer now works with fractional cardinality.
* If all else fails, if a fixed-width vector is asked to double from zero, it sizes the vector to 256 bytes.

paul-rogers:
Rebased onto latest master.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -74,53 +74,52 @@
...
-     * logic still applies, but the <tt>valueCount</tt> is the number of entries
-     * in the outer array, not the row count.
+     * Total number of elements for a repeated type, or 1 if this is
+     * a non-repeated type. That is, a batch of 100 rows may have an
+     * array with 10 elements per row. In this case, the element count
+     * is 1000.
      */
 
-    public int entryCount;
+    public final int elementCount;
```
amansinha100: Not related to elementCount per-se but I see that netBatchSize and accountedMemorySize are integers.  These could overflow depending on number of columns.  Should they be longs ?

amansinha100:
Changes LGTM.  +1

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
paul-rogers: Good point. However, a single batch of greater than 2 GB is far more than the sort can handle, so we'd not even get this far if the batch was this large.

Still, the point is valid, so a new commit changes batch size variables from int to long.

paul-rogers:
Rebased onto master.

paul-rogers:
Squashed commits.

PR #933, opened 2017-09-01
DRILL-5765: Json query profile is not shown on Web UI

sindhurirayavaram:
The "graph.js", used in the profiles page of Drill Web UI, uses $(window).load(function () {}). The ".load" function is deprecated from the Jquery after 1.8. In this change, I have changed the function to onload.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/static/js/graph.js
+++ b/exec/java-exec/src/main/resources/rest/static/js/graph.js
@@ -11,7 +11,7 @@
  *  language governing permissions and limitations under the License.
  */
 
-$(window).load(function () {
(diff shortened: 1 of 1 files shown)


arina-ielchiieva:
@sindhurirayavaram can you please explain why your previous changes caused the regression and how they are connected with `load` function. Thank you in advance.

sindhurirayavaram:
@arina-ielchiieva I updated the jquery to it's latest min version. The updated jquery library doesn't support load function anymore. It should be "onload". We were using load in graph.js which was failing. I updated that to "onload".

arina-ielchiieva:
Thanks for explanation. +1

PR #934, opened 2017-09-07
DRILL-3449 When Foreman node dies, the FragmentExecutor still tries to send status updates to Foreman

vrozov:


vrozov:
@paul-rogers Please review

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentStatusReporter.java
```diff
@@ -113,4 +120,14 @@ void fail(final UserException ex) {
     sendStatus(status);
   }
 
+  @Override
+  public void close()
+  {
+    final ControlTunnel tunnel = this.tunnel.getAndSet(null);
+    if (tunnel != null) {
+      logger.debug("Closing {}", this);
```
paul-rogers: We say we are closing the tunnel, but then don't do anything. Perhaps a comment to explain how this works?

vrozov: We are closing FragmentStatusReporter, not the `tunnel` that it references. The ControlTunnel is not Closable even though it has a reference to a resource that is Closable and should provide a way to release the resource it holds. Please let me know if a comment is required here, but I do plan to make ControlTunnel Closable. As it requires code refactoring not directly related to the JIRA/PR, I plan to do this in a separate PR.

vrozov: No, close() is not a placeholder. It closes FragmentStatusReporter and after the close, request to send status becomes no-op.

paul-rogers:
Thanks for the explanation.
+1

PR #935, opened 2017-09-07
DRILL-5766: Fix XSS vulnerabilities in Drill

arina-ielchiieva:
1. Bumped up freemarker version to 2.3.26-incubating.
2. Indicated default output format in Freemarker configuration (HTML).
3. Fixed Web UI bugs listed in DRILL-5346, DRILL-5341, DRILL-5339, DRILL-5338.

Details in [DRILL-5766](https://issues.apache.org/jira/browse/DRILL-5766).

### DIFF
--- a/exec/java-exec/src/main/resources/rest/profile/profile.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/profile.ftl
@@ -17,7 +17,7 @@
 <script>
     var globalconfig = {
         "queryid" : "${model.getQueryId()}",
-        "operators" : ${model.getOperatorsJSON()}
+        "operators" : ${model.getOperatorsJSON()?no_esc}
     };
 </script>
 </#macro>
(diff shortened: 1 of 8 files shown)


parthchandra:
+1. Looks good.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -162,8 +226,7 @@
       </div>
       <div id="fragment-overview" class="panel-collapse collapse">
         <div class="panel-body">
-          <svg id="fragment-overview-canvas" class="center-block"></svg>
```
kkhatua: @arina-ielchiieva  is there a reason we don't display this canvas any more? Doesn't seem related to the list of bugs fixed.

arina-ielchiieva: With this canvas, fragment overview was displayed weirdly. I have removed so the fragment overview looks the same as others.

arina-ielchiieva: Talked to Kunal, it turned out that we need this canvas to display graphic for fragments execution time, so I guess we need to return it back.

PR #937, opened 2017-09-07
DRILL-5002: Using hive's date functions on top of date column gives w…

vdiravka:
…rong results for local time-zone

### DIFF
--- a/contrib/storage-hive/core/src/main/codegen/templates/ObjectInspectorHelper.java
+++ b/contrib/storage-hive/core/src/main/codegen/templates/ObjectInspectorHelper.java
@@ -204,7 +204,11 @@ public class ObjectInspectorHelper {
           <#elseif entry.hiveType == "TIMESTAMP">
             JVar tsVar = jc._else().decl(m.directClass(java.sql.Timestamp.class.getCanonicalName()), "ts",
               castedOI.invoke("getPrimitiveJavaObject").arg(returnValue));
-            jc._else().assign(returnValueHolder.ref("value"), tsVar.invoke("getTime"));
+            // Bringing relative timestamp value without timezone info to timestamp value in UTC, since Drill keeps date-time values in UTC
+            JVar localDateTimeVar = jc._else().decl(m.directClass(org.joda.time.LocalDateTime.class.getCanonicalName()), "localDateTime",
+                JExpr._new(m.directClass(org.joda.time.LocalDateTime.class.getCanonicalName())).arg(tsVar));
+            jc._else().assign(returnValueHolder.ref("value"), localDateTimeVar.invoke("toDateTime")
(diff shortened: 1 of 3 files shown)


contrib/storage-hive/core/src/main/codegen/templates/ObjectInspectorHelper.java
```diff
@@ -204,7 +204,11 @@ public static JBlock getDrillObject(JCodeModel m, ObjectInspector oi,
           <#elseif entry.hiveType == "TIMESTAMP">
             JVar tsVar = jc._else().decl(m.directClass(java.sql.Timestamp.class.getCanonicalName()), "ts",
               castedOI.invoke("getPrimitiveJavaObject").arg(returnValue));
-            jc._else().assign(returnValueHolder.ref("value"), tsVar.invoke("getTime"));
+            // Bringing relative timestamp value without timezone info to timestamp value in UTC, since Drill keeps date-time values in UTC
```
paul-rogers: Drill does not actually keep values in UTC: Drill keeps values in the server time zone. This is why Drill dates, when shipped to a client with a different time zone, get corrupted.

paul-rogers:
The original description talks about data with local times. The TPC-H data has no TZ. Now, maybe we made one up in creating the Parquet files, but the original date just has dates without a tz.

The fundamental issue is that if we have a tz-less date, 1994-08-12, say, then this *cannot* be converted to a UTC timestamp. Which of the 23+ time zones would we use? How would the client and server agree on the arbitrary tz? This is like saying that I have a measurement in miles, but we can store distances only in km, so I'll take my length of 5 miles and store it as 5 km, remembering that I'm using km as an alias for miles. Does not make sense.

Your example uses `timestamp` constants. A timestamp is defined with a timezone, and so it fits Drill's model well. But, TPC dates don't have a timezone. See [the TPC-H spec](http://www.tpc.org/tpc_documents_current_versions/pdf/tpc-h_v2.17.2.pdf) which says:

> Date is a value whose external representation can be expressed as YYYY-MM-DD, where all characters are numeric. A date must be able to express any day within at least 14 consecutive years. There is no requirement specific to the internal representation of a date.

That is, TPC-H dates are not midnight on some date in some timezone, they are just dates. The cannot be converted to UTC. And so, they should not be subject to time zone shifting as tzs shift.

My point here is that Hive (according to the docs) implements functions correctly: using tz-less dates. Drill tries to convert to a (fake) UTC and use time-based functions on that data. This is, at best, a hack, and at worst, leads to great complexity and incorrect results.

That said, if all we have is km, and we can't do the miles-to-km conversion correctly, then we do need a way to know that a particular km value is actually miles. Similarly, using the current implementation, how will we know that a particular arbitrary-local-time-encoded-as-fake-UTC value really is local time vs. being an actual UTC time?

All that said, if you fix makes the current implementation work better, then it is a good improvement.

In the interests of moving ahead, let's table the basic discussion and just look at this one fix.

PR #938, opened 2017-09-09
DRILL-5694: Handle HashAgg OOM by spill and retry, plus perf improvement

Ben-Zvi:
The main change in this PR is adding a "_second way_" to handle memory pressure for the Hash Aggregate: Basically catch OOM failures when processing a new input row (during put() into the Hash Table), cleanup internally to allow a retry (of the put()) and return a new exception "**RetryAfterSpillException**". In such a case the caller spills some partition to free more memory, and retries inserting that new row.
   In addition, to reduce the risk of OOM when either creating the "Values Batch" (to match the "Keys Batch" in the Hash Table), or when allocating the Outgoing vectors (for the Values) -- there are new "_reserves_" -- one reserve for each of the two. A "_reserve_" is a memory amount subtracted from the memory-limit, which is added back to the limit just before it is needed, so hopefully preventing an OOM. After the allocation the code tries to restore that reserve (by subtracting from the limit, if possible). We always restore the "Outgoing Reserve" first; in case the "Values Batch" reserve runs empty just before calling put(), we skip the put() (just like an OOM there) and spill to free some memory (and restore that reserve).
   The old "_first way_" is still used. That is the code that predicts the memory needs, and triggers a spill if not enough memory is available. The spill code was separated into a new method called spillIfNeeded() which is used in two modes - either the old way (prediction), or (when called from the new OOM catch code) with a flag to force a spill, regardless of available memory. That flag is also used to reduce the priority of the "current partition" when choosing a partition to spill.

  A new testing option was added (**hashagg_use_memory_prediction**, default true) - by setting this to false the old "first way" is disabled. This allows stress testing of the OOM handling code (which may not be used under normal memory allocation).

  The HashTable put() code was re-written to cleanup partial changes in case of an OOM. And so the code around the call of put() to catch the new exception, spill and retry. Note that this works for 1st phase aggregation as well (return rows early).

For the estimates (in addition to the old "max batch size" estimate) - there is an estimate for the Values Batch, and one for for the Outgoing. These are used for restoring the "reserves". These estimates may be resized up in case actual allocations are bigger.

Other changes:
* Improved the "max batch size estimation" -- using the outgoing batch for getting the correct schema (instead of the input batch).
  The only information needed from the input batch is the "max average column size" (see change inRecordBatchSizer.java) to have a better estimate for VarChars.
  Also computed the size of those "no null" bigint columns added into the Values Batch when the aggregation is SUM, MIN or MAX (see changes in HashAggBatch.java and HashAggregator.java)
* Using a "plain Java" subclass for the HashTable  because "byte manipulation" breaks on the new template code (see ChainedHashTable.java)
* The three Configuration options where changed into System/Session options:   min_batches_per_partition , hashagg_max_memory , hashagg_num_partitions
* There was a potential memory leak in the HashTable BatchHolder ctor (vectors were added to the container only after the successful allocation, and the container was cleared in case of OOM. So in case of a partial allocation, the allocated part was no accessible). Also (Paul's suggestion) modified some vector templates to cleanup after any runtime error (including an OOM).
* Performance improvements: Eliminated the call to updateBatches() before each hash computation (instead used only when switching to a new SpilledRecordBatch); this was a big overhead.
   Also changed all the "setSafe" calls into "set" for the HashTable (those nanoseconds add up, specially when rehashing) - these bigint vectors need no resizing.
* Ignore "(spill) file not found" error while cleaning up.
* The unit tests were re-written in a more compact form. And a test with the new option (forcing the OOM code) was added (no memory prediction).

### DIFF
--- a/common/src/main/java/org/apache/drill/common/exceptions/RetryAfterSpillException.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/RetryAfterSpillException.java
@@ -1,0 +1,26 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.common.exceptions;
+
+/**
+ *  A special exception to be caught by caller, who is supposed to free memory by spilling and try again
+ *
+ */
+public class RetryAfterSpillException extends Exception {
+
+}
--- a/common/src/main/java/org/apache/drill/common/exceptions/UserException.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/UserException.java
@@ -22,6 +22,12 @@ import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
 import org.apache.drill.exec.proto.UserBitShared.DrillPBError;
 import org.slf4j.Logger;
 
+import java.io.File;
+import java.io.BufferedWriter;
+import java.io.FileWriter;
+import java.lang.management.ManagementFactory;
+
+import static java.lang.Thread.sleep;
 /**
  * Base class for all user exception. The goal is to separate out common error conditions where we can give users
  * useful feedback.
@@ -536,6 +542,36 @@ public class UserException extends DrillRuntimeException {
      * @return user exception
      */
     public UserException build(final Logger logger) {
+
+      // To allow for debugging:
+      //
+      //   A spinner code to make the execution stop here while the file '/tmp/drill/spin' exists
+      // Can be used to attach a debugger, use jstack, etc
+      // (do "clush -a touch /tmp/drill/spin" to turn this on across all the cluster nodes, and to
+      //  release the spinning threads do "clush -a rm /tmp/drill/spin")
+      // The processID of the spinning thread (along with the error message) should then be found
+      // in a file like  /tmp/drill/spin4148663301172491613.tmp
+      File spinFile = new File("/tmp/drill/spin");
+      if ( spinFile.exists() ) {
+        File tmpDir = new File("/tmp/drill");
+        File outErr = null;
+        try {
+          outErr = File.createTempFile("spin", ".tmp", tmpDir);
+          BufferedWriter bw = new BufferedWriter(new FileWriter(outErr));
+          bw.write("Spinning process: " + ManagementFactory.getRuntimeMXBean().getName()
+          /* After upgrading to JDK 9 - replace with: ProcessHandle.current().getPid() */);
+          bw.write("\nError cause: " +
+            (errorType == DrillPBError.ErrorType.SYSTEM ? ("SYSTEM ERROR: " + ErrorHelper.getRootMessage(cause)) : message));
+          bw.close();
+        } catch (Exception ex) {
+          logger.warn("Failed creating a spinner tmp message file: {}", ex);
+        }
+        while (spinFile.exists()) {
+          try { sleep(1_000); } catch (Exception ex) { /* ignore interruptions */ }
+        }
+        try { outErr.delete(); } catch (Exception ex) { } // cleanup - remove err msg file
+      }
+
       if (uex != null) {
         return uex;
       }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -93,18 +93,20 @@ public interface ExecConstants {
 
   // Hash Aggregate Options
 
-  String HASHAGG_NUM_PARTITIONS = "drill.exec.hashagg.num_partitions";
   String HASHAGG_NUM_PARTITIONS_KEY = "exec.hashagg.num_partitions";
   LongValidator HASHAGG_NUM_PARTITIONS_VALIDATOR = new RangeLongValidator(HASHAGG_NUM_PARTITIONS_KEY, 1, 128); // 1 means - no spilling
-  String HASHAGG_MAX_MEMORY = "drill.exec.hashagg.mem_limit";
   String HASHAGG_MAX_MEMORY_KEY = "exec.hashagg.mem_limit";
   LongValidator HASHAGG_MAX_MEMORY_VALIDATOR = new RangeLongValidator(HASHAGG_MAX_MEMORY_KEY, 0, Integer.MAX_VALUE);
   // min batches is used for tuning (each partition needs so many batches when planning the number of partitions,
   // or reserve this number when calculating whether the remaining available memory is too small and requires a spill.)
   // Low value may OOM (e.g., when incoming rows become wider), higher values use fewer partitions but are safer
-  String HASHAGG_MIN_BATCHES_PER_PARTITION = "drill.exec.hashagg.min_batches_per_partition";
-  String HASHAGG_MIN_BATCHES_PER_PARTITION_KEY = "drill.exec.hashagg.min_batches_per_partition";
-  LongValidator HASHAGG_MIN_BATCHES_PER_PARTITION_VALIDATOR = new RangeLongValidator(HASHAGG_MIN_BATCHES_PER_PARTITION_KEY, 2, 5);
+  String HASHAGG_MIN_BATCHES_PER_PARTITION_KEY = "exec.hashagg.min_batches_per_partition";
+  LongValidator HASHAGG_MIN_BATCHES_PER_PARTITION_VALIDATOR = new RangeLongValidator(HASHAGG_MIN_BATCHES_PER_PARTITION_KEY, 1, 5);
+  // Can be turned off mainly for testing. Memory prediction is used to decide on when to spill to disk; with this option off,
+  // spill would be triggered only by another mechanism -- "catch OOMs and then spill".
+  String HASHAGG_USE_MEMORY_PREDICTION_KEY = "exec.hashagg.use_memory_prediction";
+  BooleanValidator HASHAGG_USE_MEMORY_PREDICTION_VALIDATOR = new BooleanValidator(HASHAGG_USE_MEMORY_PREDICTION_KEY);
+
   String HASHAGG_SPILL_DIRS = "drill.exec.hashagg.spill.directories";
   String HASHAGG_SPILL_FILESYSTEM = "drill.exec.hashagg.spill.fs";
   String HASHAGG_FALLBACK_ENABLED_KEY = "drill.exec.hashagg.fallback.enabled";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
@@ -25,6 +25,7 @@ import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.expression.ErrorCollector;
 import org.apache.drill.common.expression.ErrorCollectorImpl;
+import org.apache.drill.common.expression.FunctionHolderExpression;
 import org.apache.drill.common.expression.IfExpression;
 import org.apache.drill.common.expression.LogicalExpression;
 import org.apache.drill.common.logical.data.NamedExpression;
@@ -251,6 +252,7 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
       groupByOutFieldIds[i] = container.add(vv);
     }
 
+    int extraNonNullColumns = 0; // each of SUM, MAX and MIN gets an extra bigint column
     for (i = 0; i < numAggrExprs; i++) {
       NamedExpression ne = popConfig.getAggrExprs().get(i);
       final LogicalExpression expr =
@@ -268,6 +270,10 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
         continue;
       }
 
+      if ( expr instanceof FunctionHolderExpression ) {
+         String funcName = ((FunctionHolderExpression) expr).getName();
+         if ( funcName.equals("sum") || funcName.equals("max") || funcName.equals("min") ) {extraNonNullColumns++;}
+      }
       final MaterializedField outputField = MaterializedField.create(ne.getRef().getAsNamePart().getName(), expr.getMajorType());
       @SuppressWarnings("resource")
       ValueVector vv = TypeHelper.getNewVector(outputField, oContext.getAllocator());
@@ -288,12 +294,11 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
         new HashTableConfig((int)context.getOptions().getOption(ExecConstants.MIN_HASH_TABLE_SIZE),
             HashTable.DEFAULT_LOAD_FACTOR, popConfig.getGroupByExprs(), null /* no probe exprs */, comparators);
 
-    agg.setup(popConfig, htConfig, context, this.stats,
-        oContext, incoming, this,
+    agg.setup(popConfig, htConfig, context, oContext, incoming, this,
         aggrExprs,
         cgInner.getWorkspaceTypes(),
         groupByOutFieldIds,
-        this.container);
+        this.container, extraNonNullColumns * 8 /* sizeof(BigInt) */);
 
     return agg;
   }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
@@ -27,7 +27,7 @@ import java.util.concurrent.TimeUnit;
 import javax.inject.Named;
 
 import com.google.common.base.Stopwatch;
-
+import org.apache.drill.common.exceptions.RetryAfterSpillException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.expression.ExpressionPosition;
 import org.apache.drill.common.expression.FieldReference;
@@ -59,8 +59,6 @@ import org.apache.drill.exec.physical.impl.spill.RecordBatchSizer;
 
 import org.apache.drill.exec.physical.impl.spill.SpillSet;
 import org.apache.drill.exec.planner.physical.AggPrelBase;
-
-import org.apache.drill.exec.proto.UserBitShared;
 
 import org.apache.drill.exec.record.MaterializedField;
 
@@ -109,14 +107,21 @@ public abstract class HashAggTemplate implements HashAggregator {
 
   private boolean isTwoPhase = false; // 1 phase or 2 phase aggr?
   private boolean is2ndPhase = false;
-  private boolean canSpill = true; // make it false in case can not spill
+  private boolean is1stPhase = false;
+  private boolean canSpill = true; // make it false in case can not spill/return-early
   private ChainedHashTable baseHashTable;
   private boolean earlyOutput = false; // when 1st phase returns a partition due to no memory
   private int earlyPartition = 0; // which partition to return early
-
-  private long memoryLimit; // max memory to be used by this oerator
-  private long estMaxBatchSize = 0; // used for adjusting #partitions
-  private long estRowWidth = 0;
+  private boolean retrySameIndex = false; // in case put failed during 1st phase - need to output early, then retry
+  private boolean useMemoryPrediction = false; // whether to use memory prediction to decide when to spill
+  private long estMaxBatchSize = 0; // used for adjusting #partitions and deciding when to spill
+  private long estRowWidth = 0; // the size of the internal "row" (keys + values + extra columns)
+  private long estValuesRowWidth = 0; // the size of the internal values ( values + extra )
+  private long estOutputRowWidth = 0; // the size of the output "row" (no extra columns)
+  private long estValuesBatchSize = 0; // used for "reserving" memory for the Values batch to overcome an OOM
+  private long estOutgoingAllocSize = 0; // used for "reserving" memory for the Outgoing Output Values to overcome an OOM
+  private long reserveValueBatchMemory; // keep "reserve memory" for Values Batch
+  private long reserveOutgoingMemory; // keep "reserve memory" for the Outgoing (Values only) output
   private int maxColumnWidth = VARIABLE_MIN_WIDTH_VALUE_SIZE; // to control memory allocation for varchars
   private long minBatchesPerPartition; // for tuning - num partitions and spill decision
   private long plannedBatches = 0; // account for planned, but not yet allocated batches
@@ -297,10 +302,7 @@ public abstract class HashAggTemplate implements HashAggregator {
   }
 
   @Override
-  public void setup(HashAggregate hashAggrConfig, HashTableConfig htConfig, FragmentContext context,
-                    OperatorStats stats, OperatorContext oContext, RecordBatch incoming, HashAggBatch outgoing,
-                    LogicalExpression[] valueExprs, List<TypedFieldId> valueFieldIds, TypedFieldId[] groupByOutFieldIds,
-                    VectorContainer outContainer) throws SchemaChangeException, IOException {
+  public void setup(HashAggregate hashAggrConfig, HashTableConfig htConfig, FragmentContext context, OperatorContext oContext, RecordBatch incoming, HashAggBatch outgoing, LogicalExpression[] valueExprs, List<TypedFieldId> valueFieldIds, TypedFieldId[] groupByOutFieldIds, VectorContainer outContainer, int extraRowBytes) throws SchemaChangeException, IOException {
 
     if (valueExprs == null || valueFieldIds == null) {
       throw new IllegalArgumentException("Invalid aggr value exprs or workspace variables.");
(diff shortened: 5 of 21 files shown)


common/src/main/java/org/apache/drill/common/exceptions/RetryAfterSpillException.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.apache.drill.common.exceptions;
+
+import org.apache.drill.exec.proto.UserBitShared;
+
+/**
+ *  A special exception to be caught by caller, who is supposed to free memory by spilling and try again
+ *
+ */
+public class RetryAfterSpillException extends UserException {
```
paul-rogers: If this exception is thrown and caught internally, it should not extend `UserException`. Instead it should extend the Java `RuntimeException`.

Better, since you know you must catch this, this should be a checked exception, extended from `Exception` and declared by the method that throws it.

`UserException` is purely for exceptions reported to the user.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -92,18 +92,20 @@
...
   LongValidator HASHAGG_MAX_MEMORY_VALIDATOR = new RangeLongValidator(HASHAGG_MAX_MEMORY_KEY, 0, Integer.MAX_VALUE);
   // min batches is used for tuning (each partition needs so many batches when planning the number of partitions,
   // or reserve this number when calculating whether the remaining available memory is too small and requires a spill.)
   // Low value may OOM (e.g., when incoming rows become wider), higher values use fewer partitions but are safer
-  String HASHAGG_MIN_BATCHES_PER_PARTITION = "drill.exec.hashagg.min_batches_per_partition";
-  String HASHAGG_MIN_BATCHES_PER_PARTITION_KEY = "drill.exec.hashagg.min_batches_per_partition";
-  LongValidator HASHAGG_MIN_BATCHES_PER_PARTITION_VALIDATOR = new RangeLongValidator(HASHAGG_MIN_BATCHES_PER_PARTITION_KEY, 2, 5);
+  String HASHAGG_MIN_BATCHES_PER_PARTITION_KEY = "exec.hashagg.min_batches_per_partition";
+  LongValidator HASHAGG_MIN_BATCHES_PER_PARTITION_VALIDATOR = new RangeLongValidator(HASHAGG_MIN_BATCHES_PER_PARTITION_KEY, 1, 5);
+  // Can be turns off mainly for testing. Memory prediction is used to decide on when to spill to disk; with this option off,
```
paul-rogers: turns --> turned

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -293,7 +299,7 @@ private HashAggregator createAggregatorInternal() throws SchemaChangeException,
         aggrExprs,
         cgInner.getWorkspaceTypes(),
         groupByOutFieldIds,
-        this.container);
+        this.container, extraNonNullColumns * 8 /* sizeof(BigInt) */);
```
paul-rogers: If the `BigInt` column is used to indicate nulls, then each value is of size 9. And, since, on average, each vector has 25% internal fragmentation. To account for this, perhaps assume that the average size is 12 or 13 bytes.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -109,14 +107,21 @@
...
+  private boolean is1stPhase = false;
+  private boolean canSpill = true; // make it false in case can not spill/return-early
   private ChainedHashTable baseHashTable;
   private boolean earlyOutput = false; // when 1st phase returns a partition due to no memory
   private int earlyPartition = 0; // which partition to return early
-
-  private long memoryLimit; // max memory to be used by this oerator
-  private long estMaxBatchSize = 0; // used for adjusting #partitions
-  private long estRowWidth = 0;
+  private boolean retrySameIndex = false; // in case put failed during 1st phase - need to output early, then retry
```
paul-rogers: As it turns out, unlike C++, Java is pretty good at initializing booleans to false and longs to 0. We only need to explicitly initialize values when the value should be other than 0/false/null.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -109,14 +107,21 @@
...
+  private boolean useMemoryPrediction = false; // whether to use memory prediction to decide when to spill
+  private long estMaxBatchSize = 0; // used for adjusting #partitions and deciding when to spill
+  private long estRowWidth = 0; // the size of the internal "row" (keys + values + extra columns)
+  private long estValuesRowWidth = 0; // the size of the internal values ( values + extra )
+  private long estOutputRowWidth = 0; // the size of the output "row" (no extra columns)
+  private long estValuesBatchSize = 0; // used for "reserving" memory for the Values batch to overcome an OOM
+  private long estOutgoingAllocSize = 0; // used for "reserving" memory for the Outgoing Output Values to overcome an OOM
+  private long reserveValueBatchMemory; // keep "reserve memory" for Values Batch
+  private long reserveOutgoingMemory; // keep "reserve memory" for the Outgoing (Values only) output
```
paul-rogers: Long lists of member variables are generally frowned upon. Can't unit test them. Too many states to keep in mind. Can these be grouped into a read-only config class (set up front, then never changed) vs, running estimates?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -297,10 +302,7 @@ public void outputRecordValues(@Named("htRowIdx") int htRowIdx, @Named("outRowId
   }
 
   @Override
-  public void setup(HashAggregate hashAggrConfig, HashTableConfig htConfig, FragmentContext context,
-                    OperatorStats stats, OperatorContext oContext, RecordBatch incoming, HashAggBatch outgoing,
-                    LogicalExpression[] valueExprs, List<TypedFieldId> valueFieldIds, TypedFieldId[] groupByOutFieldIds,
-                    VectorContainer outContainer) throws SchemaChangeException, IOException {
+  public void setup(HashAggregate hashAggrConfig, HashTableConfig htConfig, FragmentContext context, OperatorStats stats, OperatorContext oContext, RecordBatch incoming, HashAggBatch outgoing, LogicalExpression[] valueExprs, List<TypedFieldId> valueFieldIds, TypedFieldId[] groupByOutFieldIds, VectorContainer outContainer, int extraRowBytes) throws SchemaChangeException, IOException {
```
paul-rogers: Methods and constructors with many arguments are generally frowned upon as it suggests that a single class is trying to do too much: it is has too much internal coupling, performing tasks that should be broken apart. Can this single, huge, class be split into smaller, more focused, abstractions?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -382,19 +390,25 @@ private void delayedSetup() {
...
       updateEstMaxBatchSize(incoming);
     }
-    long memAvail = memoryLimit - allocator.getAllocatedMemory();
+    // create "reserved memory" and adjust the memory limit down
+    reserveValueBatchMemory = reserveOutgoingMemory = estValuesBatchSize ;
+    long newMemoryLimit = allocator.getLimit() - reserveValueBatchMemory - reserveOutgoingMemory ;
+    long memAvail = newMemoryLimit - allocator.getAllocatedMemory();
+    if ( memAvail <= 0 ) { throw new OutOfMemoryException("Too little memory available"); }
+    allocator.setLimit(newMemoryLimit);
+
```
paul-rogers: This code has grown to be incredibly complex with many, many paths through the various functions.

Tests are handy things. Do we have system-level unit tests that exercise each path through the code? Otherwise, as a reviewer, how can I be sure that each execution path does, in fact, work?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -500,22 +516,45 @@ private void initializeSetup(RecordBatch newIncoming) throws SchemaChangeExcepti
...
+      MaterializedField mr = vv.getField();
+      int fieldSize = vv instanceof VariableWidthVector ? maxColumnWidth :
+          TypeHelper.getSize(mr.getType());
+      estRowWidth += fieldSize;
+      estOutputRowWidth += fieldSize;
+      if ( fieldId < numGroupByOutFields ) { fieldId++; }
+      else { estValuesRowWidth += fieldSize; }
+    }
+    // multiply by the max number of rows in a batch to get the final estimated max size
+    estMaxBatchSize = Math.max(estRowWidth, estInputRowWidth) * MAX_BATCH_SIZE;
```
paul-rogers: Here, the output batch size is fixed based on the number of rows. Suppose we had a sort as the output of this operator, and the sort has a memory ceiling of x MB. Can the code here create batches larger than x/2 MB, meaning that that sort is forced to consume batches so large that it can't buffer two and spill?

In other words, is there an attempt here to control overall output batch memory use instead of just assuming that we always output `MAX_BATCH_SIZE` rows regardless of memory use?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -545,16 +584,19 @@ public AggOutcome doWork() {
...
         logger.debug("Starting outer loop of doWork()...");
       }
-      for (; underlyingIndex < currentBatchRecordCount; incIndex()) {
+      while (underlyingIndex < currentBatchRecordCount) {
         if (EXTRA_DEBUG_2) {
           logger.debug("Doing loop with values underlying {}, current {}", underlyingIndex, currentIndex);
         }
         checkGroupAndAggrValues(currentIndex);
+
+        if ( retrySameIndex ) { retrySameIndex = false; }  // need to retry this row (e.g. we had an OOM)
```
paul-rogers: I think Drill's coding style guidelines says no spaces after ( or before ).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -646,6 +687,46 @@ public AggOutcome doWork() {
...
+
+    reserveValueBatchMemory = 0;
+  }
+  /**
+   *   Use reserved outgoing output memory (if available) to try and preemp an OOM
+   */
+  private void useReservedOutgoingMemory() {
+    // try to preempt an OOM by using the reserved memory
+    long reservedMemory = reserveOutgoingMemory;
+    if ( reservedMemory > 0 ) { allocator.setLimit(allocator.getLimit() + reservedMemory); }
```
paul-rogers: Why is it necessary to change the allocator limit? The allocator limit should be fixed: it is the amount of memory given to this operator. Shouldn't the code use its own, internal, limits to make decisions? That is, if allocated memory + some expected use > a defined internal size, then spill?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -646,6 +687,46 @@ public AggOutcome doWork() {
...
+  }
+  /**
+   *  Restore the reserve memory (both)
+   *
+   */
+  private void restoreReservedMemory() {
+    if ( 0 == reserveOutgoingMemory ) { // always restore OutputValues first (needed for spilling)
+      long memAvail = allocator.getLimit() - allocator.getAllocatedMemory();
+      if ( memAvail > estOutgoingAllocSize) {
+        allocator.setLimit(allocator.getLimit() - estOutgoingAllocSize);
```
paul-rogers: Done this way, it is necessary for this reviewer to mentally track the current allocator limit. Can we ever subtract an amount twice? Add it twice?

Now, it seems we should not alter the allocator. But, if we do, shouldn't it be based on absolute amounts, not relative deltas?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -1335,7 +1470,7 @@ private void updateStats(HashTable[] htables) {
     }
     if ( rowsReturnedEarly > 0 ) {
       stats.setLongStat(Metric.SPILL_MB, // update stats - est. total MB returned early
-          (int) Math.round( rowsReturnedEarly * estRowWidth / 1024.0D / 1024.0));
+          (int) Math.round( rowsReturnedEarly * estOutputRowWidth / 1024.0D / 1024.0));
```
paul-rogers: This file is a template. This means, we copy *all* this code each time we generate a new class. How is doing so helping stability, customer value or performance? Should all this code be in a template that is copied on every query? Or, should it be refactored into a driver class, with only a very light wrapper appearing in the copied template?

As this code get ever more complex, it puts a strain on the Java code that must walk though this code and do method fixup, scalar replacements, etc. That work takes time. What value accrues to the user from doing this fixup on code that never changes  from one query to the next?

Filed [DRILL-5779](https://issues.apache.org/jira/browse/DRILL-5779) for this issue.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -1178,20 +1273,38 @@ private void checkGroupAndAggrValues(int incomingRowIdx) {
...
     // ==========================================
     // Insert the key columns into the hash table
     // ==========================================
+    boolean noReserveMem = reserveValueBatchMemory == 0;
     try {
+      if ( noReserveMem && canSpill ) { throw new RetryAfterSpillException();} // proactive spill, skip put()
+
       putStatus = htables[currentPartition].put(incomingRowIdx, htIdxHolder, hashCode);
+
+    } catch (RetryAfterSpillException re) {
```
paul-rogers: See above. Should be a checked exception declared by `put()`.

Also, why do we need to throw an exception before calling `put` only to catch it a couple of lines later?

If the spill code was in a method, rather than just inline, seems we could do:

```
if ( noReserveMem && canSpill ) { doSpill(); }
try {
  ...put(...)
} catch (...) {
   doSpill();
}
```

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -1178,20 +1273,38 @@ private void checkGroupAndAggrValues(int incomingRowIdx) {
...
+      if ( ! canSpill ) { throw new OutOfMemoryException(getOOMErrorMsg("Can not spill")); }
```
paul-rogers: This is the message sent to the log and user. Should we explain why we can't spill? And, what to do? Something like:

"Incoming batch too large and no in-memory partitions to spill. Increase memory assigned to the Hash Agg."

Replace the above wording with the actual reasons and fixes.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggregator.java
```diff
@@ -47,10 +47,7 @@
   // OK - batch returned, NONE - end of data, RESTART - call again
   public enum AggIterOutcome { AGG_OK, AGG_NONE, AGG_RESTART }
 
-  public abstract void setup(HashAggregate hashAggrConfig, HashTableConfig htConfig, FragmentContext context,
-                             OperatorStats stats, OperatorContext oContext, RecordBatch incoming, HashAggBatch outgoing,
-                             LogicalExpression[] valueExprs, List<TypedFieldId> valueFieldIds, TypedFieldId[] keyFieldIds,
-                             VectorContainer outContainer) throws SchemaChangeException, IOException, ClassTransformationException;
+  public abstract void setup(HashAggregate hashAggrConfig, HashTableConfig htConfig, FragmentContext context, OperatorStats stats, OperatorContext oContext, RecordBatch incoming, HashAggBatch outgoing, LogicalExpression[] valueExprs, List<TypedFieldId> valueFieldIds, TypedFieldId[] keyFieldIds, VectorContainer outContainer, int extraRowBytes) throws SchemaChangeException, IOException, ClassTransformationException;
```
paul-rogers: Not sure that putting all items on one big line is an improvement over the arg-per-line format previously.

Also, see note above: a large number of arguments suggest a muddy design with one class trying to do far too much.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTable.java
```diff
@@ -58,7 +59,7 @@
 
   public int getHashCode(int incomingRowIdx) throws SchemaChangeException;
 
-  public PutStatus put(int incomingRowIdx, IndexPointer htIdxHolder, int hashCode) throws SchemaChangeException;
+  public PutStatus put(int incomingRowIdx, IndexPointer htIdxHolder, int hashCode) throws SchemaChangeException, RetryAfterSpillException;
```
paul-rogers: At present, `RetryAfterSpillException` is unchecked, so it is not necessary to declare. But, change `RetryAfterSpillException` to extend `Exception` (checked) and this declaration then becomes useful.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -158,19 +158,17 @@ public BatchHolder(int idx) {
...
-          }
+          if (links != null) { links.clear();}
         }
       }
     }
 
     private void init(IntVector links, IntVector hashValues, int size) {
       for (int i = 0; i < size; i++) {
-        links.getMutator().setSafe(i, EMPTY_SLOT);
+        links.getMutator().set(i, EMPTY_SLOT);
```
paul-rogers: Is size ever less than the vector capacity()? Else, you can just ask the vector for its capacity.

The `links.getMutator()` call in an inner loop is inefficient.

Instead of a single function initializing two `IntVector`s with redundant code, can this be refactored to have a function that initializes one vector, that is called twice?

common/src/main/java/org/apache/drill/common/exceptions/RetryAfterSpillException.java
Ben-Zvi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
Ben-Zvi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
Ben-Zvi: Not sure .... seemed to work OK in some (limited) testing.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: This is more for code readability -- "by default, this flag was chosen to be false".

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Removed one argument "stats" - can be taken from the "oContext"

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Most of these estimates are for internal "worst case".  Only the "outgoing" one is for the outgoing batch (which is also for spilling - which is internal).
   Anyway all these estimates have nothing to do with _throttling_ the outgoing batch size; that logic was not changed from the original code (likely up to MAX_BATCH_SIZE). 
  Making such a change should be a separate project.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Yes indeed. The only attempt to force a code path is the new testing option *use_memory_prediction* which can disable the estimate based prediction (when to spill), hence forcing the code path that relies on an OOM (for hash table put() ) to take place (one unit test was added for that).
    Getting a full code coverage would be ideal, but hard.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Because the first uncontrolled memory allocation happens when inserting into the hash table (i.e. put()).  Given this "uncontrollability", better OOM there (which we can handle, by spilling and retrying). Now if all the memory was "given" in the limit, the put() may not OOM, but leave too little available memory to continue (i.e. to create a values batch, or an outgoing batch) -- these situations we can not handle.
   By subtracting from the limit a "reserve" for these two batches, we may force a put() OOM early (but that's OK). But we also ensure that the following two batches could be allocated. In some way this is similar to having multiple dedicated allocators, only simpler.
   Adding or subtracting is just an operation on a local field; no performance effect in any way. Also by using a single allocator we can handle cases like a "bump" in a batch size (which may exceed pre-allocation in a dedicated separate allocator).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Can never add twice, as the code only adds to an empty ( == 0 ) reserve.
And these are not "relative deltas", but the actual expected batch size (so that the following allocation would not OOM).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: The method getOOMErrorMsg() does all this explanation ...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTable.java
Ben-Zvi: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
Ben-Zvi: This init() method is not used .... looks like leftover old code

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Will wait for some future cleanup opportunity.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: So why does "or before" have spaces ? :-)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
Ben-Zvi: Work will be done later as part of DRILL-5779

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggregator.java
Ben-Zvi: That was one of the IDE's ideas ....
And simplification could be done as part of future cleanup work (like DRILL-5779)

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -536,6 +542,33 @@ public Builder pushContext(final String name, final double value) {
      * @return user exception
      */
     public UserException build(final Logger logger) {
+
+      // To allow for debugging:
+      //     A spinner code to make the execution stop here while the file '/tmp/drillspin' exists
```
paul-rogers: Would recommend `/tmp/drill/spin`. We already use `/tmp/drill` for other items, so this keep things tidy.

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -536,6 +542,33 @@ public Builder pushContext(final String name, final double value) {
...
+      // Can be used to attach a debugger, use jstack, etc
+      // The processID of the spinning thread should be in a file like /tmp/spin4148663301172491613.tmp
```
paul-rogers: Would also recommend `/tmp/drill/spin...`.

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -536,6 +542,33 @@ public Builder pushContext(final String name, final double value) {
...
+      // along with the error message.
+      File spinFile = new File("/tmp/drillspin");
```
paul-rogers: Should this be a config setting? Probably the config is not visible here, but can we set a static variable at start-up time? And, since this code will check the file system on every exception, should we have a config variable to turn on the check?

Feel free to tell me I'm being overly paranoid...

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
```diff
@@ -536,6 +542,33 @@ public Builder pushContext(final String name, final double value) {
...
+          bw.write("Spinning process: " + ManagementFactory.getRuntimeMXBean().getName()
+          /* After upgrading to JDK 9 - replace with: ProcessHandle.current().getPid() */);
+          bw.write("\nError cause: " +
+            (errorType == DrillPBError.ErrorType.SYSTEM ? ("SYSTEM ERROR: " + ErrorHelper.getRootMessage(cause)) : message));
+          bw.close();
+        } catch (Exception ex) {
+          logger.warn("Failed creating a spinner tmp message file: {}", ex);
+        }
+        while (spinFile.exists()) {
+          try { sleep(1_000); } catch (Exception ex) { /* ignore interruptions */ }
```
paul-rogers: What happens it the fragment executor tries to kill the query? Do we want the spinner to ignore that request here?

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
Ben-Zvi: Using a "flag file" instead of a config setting gives more flexibility; like no need to restart in order to turn this feature on/off, or can select to catch errors only in few nodes, and last -- can free the looping thread by deleting this "flag file". 
      I also plan on posting an announcement on the dev list about this new "feature", and see if there's any feedback.

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
Ben-Zvi: Does query killing cause a user exception ?

Ben-Zvi: Yes - if some non-blocked part tries to kill the query, the spinning parts would still be blocked - that may be by design, as debugging still goes on (until a user issues "clush -a rm /tmp/drill/spin" )

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
Ben-Zvi: Done

common/src/main/java/org/apache/drill/common/exceptions/UserException.java
Ben-Zvi: Done

paul-rogers:
@Ben-Zvi, unit tests failed with these errors:
```
Failed tests: 
  TestHashAggrSpill.testHashAggrFailWithFallbackDisabed:165 null

Tests in error: 
  TestHashAggrSpill.testNoPredictHashAggrSpill:135->testSpill:110 » IllegalState
  TestHashAggrSpill.testHashAggrSecondaryTertiarySpill:147->testSpill:110 » IllegalState
  TestHashAggrSpill.testSimpleHashAggrSpill:123->testSpill:110 » IllegalState Cl...
  TestHashAggrSpill.testHashAggrSuccessWithFallbackEnabled:179->testSpill:110 » IllegalState
```
Is this expected at this stage? Are these due to the test framework issue we discussed recently?

Ben-Zvi:
These tests passed for me (embedded, on the Mac) after squashing and rebasing yesterday.

Were these failures seen on the cluster ?

The test that "failed" was supposed to hit a failure internally (user error - not enough memory).



paul-rogers:
These failures were seen on the QA cluster. Look for for build "MERGE-070921-1".

The team just cleaned up a bunch of failing tests; I hesitate to introduce a commit that causes a new set to fail. Can you investigate please?

PR #939, opened 2017-09-11
DRILL-5550: Missing CSV column value set to null

prasadns14:
1) Changed the CSV datatype to nullable VARCHAR
2) Missing CSV column value set to null
3) Unit test added

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/FieldVarCharOutput.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/FieldVarCharOutput.java
@@ -32,7 +32,7 @@ import org.apache.drill.common.types.Types;
 import org.apache.drill.exec.exception.SchemaChangeException;
 import org.apache.drill.exec.physical.impl.OutputMutator;
 import org.apache.drill.exec.record.MaterializedField;
-import org.apache.drill.exec.vector.VarCharVector;
+import org.apache.drill.exec.vector.NullableVarCharVector;
 
 /**
  * Class is responsible for generating record batches for text file inputs. We generate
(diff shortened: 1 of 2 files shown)


prasadns14:
@paul-rogers Can you please review the PR?

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/FieldVarCharOutput.java
```diff
@@ -45,11 +45,11 @@
   static final String COL_NAME = "columns";
 
   // array of output vector
-  private final VarCharVector [] vectors;
+  private final NullableVarCharVector [] vectors;
```
paul-rogers: Let's think about this. CSV is defined as a fixed set of columns. For this reason, the original code used non-nullable values for each column.

The question would be, what about this fix means that defined columns should now become null? Are there other solutions to this bug?

Changing the column from required to nullable can have impact on the user experience and on the client: where code once expected required columns, now that code must change to handle nullable types.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/FieldVarCharOutput.java
```diff
@@ -211,10 +210,10 @@ public void finishRecord() {
       endField();
     }
 
-    // Fill in null (really empty) values.
+    // Fill in null values
 
     for (int i = 0; i < nullCols.length; i++) {
-      vectors[nullCols[i]].getMutator().setSafe(recordCount, nullValue, 0, 0);
+      vectors[nullCols[i]].getMutator().setNull(recordCount);
```
paul-rogers: Here we are combining null columns and data columns into a single array. Perhaps we should have an array of projected columns (non-null), and a separate array of null columns (optional varChar).

Note that "memory fragmentation" project has fixed this in the general case. In that solution, the solution is similar to that described above: projected columns are required, missing columns are nullable.

vdiravka:
@prasadns14 Could you please respond to Paul's review comments? [DRILL-6147](https://issues.apache.org/jira/browse/DRILL-6147) is resolved, so it can also be helpful to proceed with this issue.

arina-ielchiieva:
Closing PR since it's no longer valid to to new text reader introduction. Keeping the Jira open though since issue still persists.

PR #940, opened 2017-09-12
DRILL-5752 Speed Up Unit Tests add Test Categories

ilooner:
This PR does the following.

**Increased Concurrency:** Previously tests only executed in 2 simultaneous processes, now one process per core is used to execute tests. This PR also fixed a few issues that popped up when concurrency was increased.

**Test Categories:** Tests can now be executed according to their categories. There are three main categories of tests **Fast Tests**, **Slow Tests**, and "Unlikely Tests". **Fast Tests** are fast. **Slow Tests** are slow. **Unlikely Tests** are tests which test code that is unlikely to be changed.

All of the tests are run by default:

```
mvn -T 1C clean install
```

**Note:** -T 1C increased the number of threads used to compile classes, and allows us to run tests for sub modules in parallel. This command takes about 1h 22min to run on my laptop.

To run only fast tests you can do.

```
mvn -T 1C clean install -DexcludedGroups="org.apache.drill.categories.SlowTest,org.apache.drill.categories.UnlikelyTest"
```

This command takes 25min to run on my laptop.

In addition to **Fast** and **Secondary** tests there are more categories like:
- **JdbcTest**
- **ParquetTest**
- **HiveStorageTest**
- And many more

All of these categories are in the **drill-common** sub module in **org.apahce.drill.categories**. All the tests are marked with one or more of these categories, and we can use these categories to select the tests we run. To run all the **Fast** tests that belong to the **JdbcTest** category we can use the following command:

```
mvn -T 1C clean install -Dgroups="org.apache.drill.categories.JdbcTest" -DexcludedGroups="org.apache.drill.categories.SlowTest,org.apache.drill.categories.UnlikelyTest"
```

In order to run all the **Fast** AND **Secondary** JdbcTests you can use this command

```
mvn -T 1C clean install -Dgroups="org.apache.drill.categories.JdbcTest"
```


ilooner:
@paul-rogers This is an implementation of the Test Categories suggestion you made on the dev list. It's ready for review.

common/src/test/java/org/apache/drill/categories/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This package stores the JUnit test categories.
+ */
```
paul-rogers: Very cool. One thing that is not intuitively obvious here is the purpose and usage of the categories.

On the one hand, we already have modules for memory, vectors, Hive, etc. One can easily run just those tests by cd'ing into that module and running Maven from there.

So, one would imagine categories to provide another dimension, orthogonal to modules. For example, the classic "fast" vs. "slow." One could also imagine a core set of "smoke" tests that run quickly and do a quick validation. Then a set of standard tests that are more thorough. Then, very in-depth tests that either take a long time, or validate very specific bugs or regressions.

Can you add a bit of commentary to advise how you envision the categories to be used?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestCorruptParquetDateCorrection.java
```diff
@@ -59,6 +61,7 @@
  * Use of this option is assumed to be extremely unlikely, but it is included
  * for completeness.
  */
+@Category(ParquetTest.class)
```
paul-rogers: Secondary? Would not seem that this test need to be run on each build...

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -54,11 +56,13 @@
 import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
+import org.junit.experimental.categories.Category;
 import org.junit.rules.TemporaryFolder;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
 @RunWith(Parameterized.class)
+@Category({SecondaryTest.class, ParquetTest.class})
```
paul-rogers: On the other hand, the Parquet writer seems pretty important and we'd want to test it more often.

This is why, in my earlier comment, I asked about the purpose of categorization: those that are worth running frequently ("smoke tests") vs. those that give a more thorough test, but will typically not find issues.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSimpleExternalSort.java
```diff
@@ -148,10 +149,10 @@ public void sortOneKeyDescendingExternalSortLegacy() throws Throwable {
 
   private void sortOneKeyDescendingExternalSort(boolean testLegacy) throws Throwable {
     FixtureBuilder builder = ClusterFixture.builder()
-        .configProperty(ExecConstants.EXTERNAL_SORT_SPILL_THRESHOLD, 4)
-        .configProperty(ExecConstants.EXTERNAL_SORT_SPILL_GROUP_SIZE, 4)
-        .configProperty(ExecConstants.EXTERNAL_SORT_BATCH_LIMIT, 4)
-        .configProperty(ExecConstants.EXTERNAL_SORT_DISABLE_MANAGED, false)
+        .configProperty(OptionValidator.OPTION_DEFAULTS_ROOT + ExecConstants.EXTERNAL_SORT_SPILL_THRESHOLD, 4)
```
paul-rogers: These options are boot options, not runtime options, so no need for the prefix (unless someone changed something.)

exec/jdbc/src/test/java/org/apache/drill/jdbc/ConnectionTest.java
```diff
@@ -57,7 +61,9 @@ public static void setUpConnection() throws SQLException {
     // Connection--and other JDBC objects--on test method failure, but this test
     // class uses some objects across methods.)
     Driver.load();
-    connection = DriverManager.getConnection( "jdbc:drill:zk=local" );
+    Properties properties = new Properties();
+    properties.setProperty(OptionValidator.OPTION_DEFAULTS_ROOT + ExecConstants.CREATE_PREPARE_STATEMENT_TIMEOUT_MILLIS, "30000");
```
paul-rogers: I wonder, should we define a function for this pattern?

```
ExecConstants.bootDefaultFor(
    ExecConstants.CREATE_PREPARE_STATEMENT_TIMEOUT_MILLIS)
```

But, we use Java 7, so the static function must reside elsewhere. Actually, the function might exist; I seem to recall making the same comment in the PR that externalized the defaults.

paul-rogers: Also, since properties don't have to be strings, why is this property stored as a string when the value is an int?

common/src/test/java/org/apache/drill/categories/package-info.java
ilooner: Currently there are two broad categories: **Fast** and **Secondary**. The **Fast** tests are smoke tests. The **Secondary** tests are for slow tests. I can add a third **BugFix**category to represent tests for specific bugs.

The rest of the categories represent broad groups of tests. Some of those categories have tests which span multiple class files and multiple packages. An example of such a category is **OperatorTest**, **PlannerTest**, or **SqlTest**. These categories are helpful when you are making a change in a particular area since they allow you to run all the tests for that area easily, and also effectively serve as a form of documentation. For example it would have been helpful to have an <b>OptionsTest</b> category when I was making the option system changes, since it took me a while to find all the relevant test classes and I had to construct a long maven command to run the tests.

Also the structure of the tests as they are now is somewhat counter intuitive because tests can span different modules. For example consider the vector submodule. One would think all the relevant tests for vectors are in that submodule, but currently they are not. In fact there are no unit tests in the vector submodule and all relevant vector tests are in java-exec. Also the memory submodule has tests in it and a relevant test in the java-exec  module as well.

Some categories I agree are redundant like HiveStorageTest. Those tests have a clear one to one mapping to their submodule, however, I included a category for them as well for the sake of consistency. This way we can have a uniform mechanism for running subsets of tests instead of one mechanism for some groups of tests and another mechanism for other tests.

To summarize, I can add a **BugFixTest** category to add another level of distinction between **Secondary** tests. But I am compelled to keep the other test categories intact for the reasons outlined above. Let me know what you think.

exec/jdbc/src/test/java/org/apache/drill/jdbc/ConnectionTest.java
ilooner: I made ExecConstants a file class with a private constructor and added the static method there.

ilooner: The Properties class only allows strings to be set as values. The conversion to an integer is done somewhere inside the DriverManager.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
ilooner: To me the main distinction between smoke and secondary tests should be runtime. It's hard in my mind to distinguish based on importance, because the relevance of a test depends on the change being done. And you would end up running all the tests before finishing your change no matter what. The workflow I had in mind was the following:

  1. Make a change
  1. Run Fast tests for my category of interest (e.g. **OperatorTest**). Oops there's a failure
  1. Fix the failure
  1. Run Fast tests for category **OperatorTest** again. Oops another failure
  1. Fix the failure
  1. Run Fast tests for category **OperatorTest** again. Yay they pass
  1. Run all the tests
  1. Done

Let me know what you think.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSimpleExternalSort.java
ilooner: Done, thanks for catching.

common/src/test/java/org/apache/drill/categories/package-info.java
ilooner: Terminology update. I'm renaming **SecondaryTest** to **SlowTest** and instead of **BugFixTest** I will create a category called **UnlikelyTest**. **UnlikelyTest** will be a category for tests that excercise specific bug fixes, or are for tests that you're unlikely to break.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
ilooner: Please see Terminology update in comment above.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestCorruptParquetDateCorrection.java
ilooner: Done

paul-rogers:
Thanks for the explanations and corrections.
Please squash your commits to prepare for commit to master.
+1

ilooner:
Thanks @paul-rogers squashed and ready to commit

ilooner:
Please merge my other PR https://github.com/apache/drill/pull/923 before this one. Thanks!

ilooner:
Don't merge this. I need to test this on jenkins with the latest fixes in https://github.com/apache/drill/pull/945

ilooner:
@paul-rogers I have made a few other changes, could you review once more?

paul-rogers:
@ilooner, thanks for the improvements. Is this PR now ready for final review and commit?

Also, if all tests are enabled by default, is there a suggestion of a Maven command that developers should run normally in order to do a quick build and "smoke test"?

ilooner:
Yeah this is ready for final review and commit. To do a quick build and smoke test you can do

```
mvn -T 1C clean install -DexcludedGroups="org.apache.drill.categories.SlowTest,org.apache.drill.categories.UnlikelyTest"
```

PR #941, opened 2017-09-13
DRILL-5775: Expand * into list of column/column family for  MapR-DB b…

jinfengni:
…inary table groupscan.

 Move common code to HBaseUtils, so it's shared by both HBaseGroupScan and BinaryTableGroupScan
 The issue was exposed by existing unit tests in contrib/format-maprdb; no new unit tests added.


paul-rogers:
This fix is based on a change made in DRILL-5546. As described in DRILL-5830, the HBase part of that fix was a bit redundant. It turns out that HBase, via the project rel improvements in DRILL-5546, already expanded the wildcard into columns using the 'DrillHBaseTable.getRowType()' method. (You understand the details better than I...)

I wonder if we can do the same for MapR-DB binary? That way, we have just one common path for projection push down.

paul-rogers:
Is this PR obsolete after DRILL-5830 and the other DRILL-5775? If so can we close this one? Otherwise, if this one addresses a case not covered by the others, can we revise this PR to cover just that other case, on top of latest master? Thanks!

prasadns14:
Yes, this PR is now obsolete. All the cases are covered as part of DRILL-5830 and DRILL-5775

ilooner:
Since this PR is obsolete. I am closing it.

PR #942, opened 2017-09-14
DRILL-5781: Fix unit test failures to use tests config even if default config is available

vvysotskyi:
Please see [DRILL-5781](https://issues.apache.org/jira/browse/DRILL-5781) for details.

### DIFF
--- a/contrib/storage-hbase/src/test/resources/hbase-site.xml
+++ b/contrib/storage-hbase/src/test/resources/hbase-site.xml
@@ -66,15 +66,13 @@
     Default is 10.
     </description>
   </property>
-<!--
-   <property>
+  <property>
     <name>hbase.master.info.port</name>
     <value>-1</value>
     <description>The port for the hbase master web UI
     Set to -1 if you do not want the info server to run.
     </description>
   </property>
--->
   <property>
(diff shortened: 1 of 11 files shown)


contrib/storage-hbase/src/test/resources/hbase-site.xml
```diff
@@ -66,15 +66,13 @@
     Default is 10.
     </description>
   </property>
-<!--
-   <property>
+  <property>
```
arina-ielchiieva: Can you please explain why we needed to enable this property?

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/PathUtils.java
```diff
@@ -70,4 +72,14 @@ public static final String normalize(final String path) {
     return builder.toString();
   }
 
+  /**
+   * Creates and returns path with the protocol at the beginning from specified {@code url}.
+   */
```
arina-ielchiieva: Can you please add java doc with @param and @return?

arina-ielchiieva:
+1, LGTM.

PR #945, opened 2017-09-18
DRILL-5798 Fixed Unstable Tests

ilooner:
- Fixed unstable StatusResourcesTest
 - Fixed buggy port hunting for the WebService
 - Fixed bug in ClientFixture which does not assign the correct user port when port hunting is done for the user port
 - Fixed unstable hash agg tests


ilooner:
@paul-rogers PR https://github.com/apache/drill/pull/923 accidentally got merged before I finished fixing jenkins test issues. This PR fixes the tests that were failing on jenkins as a result of the PR, it also fixes a pre-existing issue with the ClientFixture that would make other tests flakey. The pre existing issue was that the the Drillbit would hunt for a user port, but the client would not be configured with port which was found by hunting, instead the client would be configured with the default user port. As a result, some tests could talk to a drill bit that was started in another test or a drill bit that is already running on the test machine.

paul-rogers:
@ilooner, thanks for the fixes. In general, would be great to avoid posting a PR until it is ready for prime time, that way we'll know when a fix is actually ready for review and commit. (Of course, there may be several rounds of comments and fixes...)

Does this fix also include a fix for the jdbc-all project which now exceeds the jar file size limit, probably because of a new library?

ilooner:
@paul-rogers Yeah it fixes jdbc-all. Sorry for the mix up this mistake mainly stemmed from me being ignorant of the jenkins build and being unaware of jenkins failures until the batch commit was being done. Will avoid this in the future.

ilooner:
@paul-rogers I had added a dependency that was being pulled into the jdbc-all jar. I excluded that dependency, but the size of the jar had still grown to be 100 kb over 29 Mb, so I bumped the allowed size to be 30 Mb. Thanks for teaching me about this :)

PR #947, opened 2017-09-18
DRILL-5800: Explicitly set locale to en_US on locale-dependent tests

xhochy:


xhochy:
Seeing that this is already covered by https://github.com/apache/drill/pull/904, will close.

PR #948, opened 2017-09-19
DRILL-5745: Corrected 'location' information in Drill web server

prasadns14:
Updated the 'location' information in Drill web server to-
1) consider if HTTPS is enabled
2) use the foreman address rather than "localhost"
3) use the set HTTP port

@paul-rogers, please review


arina-ielchiieva:
+1, LGTM.

PR #950, opened 2017-09-20
DRILL-5431: SSL Support

parthchandra:
Add support for SSL between Java/C++ clients and Drillbits.


contrib/native/client/src/include/drill/common.hpp
```diff
@@ -163,9 +170,13 @@ typedef enum{
 #define USERPROP_USERNAME "userName"
 #define USERPROP_PASSWORD "password"
 #define USERPROP_SCHEMA   "schema"
-#define USERPROP_USESSL   "useSSL"        // Not implemented yet
-#define USERPROP_FILEPATH "pemLocation"   // Not implemented yet
-#define USERPROP_FILENAME "pemFile"       // Not implemented yet
+#define USERPROP_USESSL   "enableTLS"
+#define USERPROP_TLSPROTOCOL "TLSProtocol" //TLS version
+#define USERPROP_CERTFILEPATH "certFilePath" // pem file path and name
+#define USERPROP_CERTPASSWORD "certPassword" // Password for certificate file
```
superbstreak: I think we can remove this to avoid confusion :)

parthchandra: Commented it out. I'm leaving it there so it can be used when we implement support for trust stores which may require passwords.

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -65,108 +66,70 @@ struct ToRpcType: public std::unary_function<google::protobuf::int32, exec::user
...
+    channelType_t type = ( props->isPropSet(USERPROP_USESSL) &&
+            props->getProp(USERPROP_USESSL, val) =="true") ?
+        CHANNEL_TYPE_SSLSTREAM :
+        CHANNEL_TYPE_SOCKET;
+
+    connectionStatus_t ret = CONN_SUCCESS;
+    m_pChannelContext = ChannelContextFactory::getChannelContext(type, props);
+    m_pChannel= ChannelFactory::getChannel(type, m_io_service, host, port);
+    m_pChannel->init(m_pChannelContext);
+    ret=m_pChannel->init(m_pChannelContext);
```
sohami: duplicate call to `m_pChannel->init(m_pChannelContext);`

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -65,108 +66,70 @@ struct ToRpcType: public std::unary_function<google::protobuf::int32, exec::user
...
-        if(ec){
-            return handleConnError(CONN_FAILURE, getMessage(ERR_CONN_FAILURE, host, port, ec.message().c_str()));
-        }
-
-    }catch(const std::exception & e){
-        // Handle case when the hostname cannot be resolved. "resolve" is hard-coded in boost asio resolver.resolve
-        if (!strcmp(e.what(), "resolve")) {
-            return handleConnError(CONN_HOSTNAME_RESOLUTION_ERROR, getMessage(ERR_CONN_EXCEPT, e.what()));
+connectionStatus_t DrillClientImpl::connect(const char* host, const char* port, DrillUserProperties* props){
+    if (this->m_bIsConnected || (this->m_pChannelContext!=NULL && this->m_pChannel!=NULL)) {
```
sohami: I think the second condition is not sufficient to say if there is already a valid connection. Instead there should be `isConnected()` method in channel which will `return m_state == CHANNEL_CONNECTED`

Also why we need both `this->m_bIsConnected` and second condition ?

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -0,0 +1,237 @@
...
+                    m_pSocket->protocolHandshake(useSystemConfig);
+                } catch (boost::system::system_error e) {
+                    status = handleError(CONN_HANDSHAKE_FAILED, e.what());
+                }
+                return status;
+            }
+
+            channelState_t m_state;
+            DrillClientError* m_pError;
+            bool m_ownIoService;
```
sohami: why is this flag needed ?

contrib/native/client/src/clientlib/wincert.ipp
```diff
@@ -0,0 +1,91 @@
...
+	};
+     
+    SSL_CTX * ctx = SSL_get_SSL_CTX(ssl);
+    X509_STORE *store = SSL_CTX_get_cert_store(ctx);
+
+	for(int i=0; i<4; i++){
+    hStore = CertOpenSystemStore(NULL, stores[i]);
+
+    if (!hStore)
+        return 1;
```
sohami: This means we will return with failure while opening any of the 4 system store. Should we instead try all 4 system stores and log the ones for which failure happened (by appending the names to string param suggested in above comment) but still succeed if anyone store was successfully opened ? 

But then I think we should also check if there is atleast one certificate which was added to X509 store out of these system store ?

contrib/native/client/src/include/drill/userProperties.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Drill{
+
+class DECLSPEC_DRILL_CLIENT DrillUserProperties{
+    public:
+        static const std::map<std::string, uint32_t> USER_PROPERTIES;
+
+        DrillUserProperties(){};
+
+        void setProperty( const std::string& propName, const std::string& propValue){
```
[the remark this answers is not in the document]
parthchandra: Not a bad idea, but we might have to stay with this for the moment. The implementation currently does not own the string that's passed in, leaving it to the caller to free the memory when appropriate. To convert always to lowercase during the set method, will require me to change the behavior, make a copy and then remember to free the memory in case it is overwritten. If we're doing that, then we might as well enhance the implementation to get/set type specific options (setBoolean, setInt, etc) and that would become so much work for not much gain.

sohami:
Thanks for the changes!
+1 LGTM.

arina-ielchiieva:
@parthchandra could you please resolve the conflicts?

parthchandra:
Squashed the review changes and rebased on master.

exec/java-exec/src/main/java/org/apache/drill/exec/ssl/SSLConfigClient.java
```diff
@@ -0,0 +1,273 @@
...
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLParameters;
+import javax.net.ssl.TrustManagerFactory;
+import java.util.Properties;
+
+public class SSLConfigClient extends SSLConfig {
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SSLConfigClient.class);
+
+  Properties properties;
```
arina-ielchiieva: private final?

exec/java-exec/src/main/java/org/apache/drill/exec/ssl/SSLConfigServer.java
```diff
@@ -0,0 +1,331 @@
...
+    if (value.isEmpty() && config.hasPath(name)) {
+      value = config.getString(name);
+    }
+    value = value.trim();
+    return value;
+  }
+
+  private String getHadoopConfigParam(String name) {
+    Preconditions.checkArgument(this.hadoopConfig != null);
+    String value = "";
```
arina-ielchiieva: Can be removed and initialized on the next line.

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -49,12 +64,19 @@ public void testMissingKeystorePassword() throws Exception {
...
+          .config(config.build())
+          .mode(SSLFactory.Mode.SERVER)
+          .initializeSSLContext(false)
+          .validateKeyStore(true)
+          .build();
       fail();
       //Expected
     } catch (Exception e) {
-      assertTrue(e instanceof DrillException);
+
```
arina-ielchiieva: So what is test is actually testing? Since we just fail and ignore the exception.

exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/security/TestUserBitSSL.java
```diff
@@ -0,0 +1,338 @@
...
+      tempFile1.deleteOnExit();
+      trustStorePath = tempFile1.getAbsolutePath();
+      //generate a truststore.
+      ts = KeyStore.getInstance(KeyStore.getDefaultType());
+      ts.load(null, password.toCharArray());
+      ts.setCertificateEntry("drillTest", certificate.cert());
+      // Store away the truststore.
+      FileOutputStream fos1 = new FileOutputStream(tempFile1);
+      ts.store(fos1, password.toCharArray());
+      fos1.close();
```
arina-ielchiieva: It's better to use try with resources thus stream will be definitely closed in case of exception.

distribution/pom.xml
```diff
@@ -97,8 +97,41 @@
...
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-client</artifactId>
+      <exclusions>
+       <exclusion>
+           <groupId>io.netty</groupId>
+           <artifactId>netty</artifactId>
+       </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
```
arina-ielchiieva: Not sure why we need to netty dependency with `test` scope in distribution pom...

exec/java-exec/pom.xml
```diff
@@ -589,6 +669,17 @@
...
   <build>
 
+    <extensions>
+      <!--
+      Include the os-maven-plugin to get os.detected.classifier
+      -->
+      <extension>
+        <groupId>kr.motd.maven</groupId>
+        <artifactId>os-maven-plugin</artifactId>
+        <version>1.4.0.Final</version>
```
arina-ielchiieva: Could not build Drill project on centos. Failed with exception:
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-shade-plugin:2.4.1:shade (default) on project drill-jdbc-all: Error creating shaded jar: The name "os.detected.release.like."centos"" is not legal for JDOM/XML elements: XML names cannot contain the character """. -> [Help 1]`

It seems we need to use `1.5.0.Final` version (example of similar problem - https://github.com/apache/beam/pull/2391).

contrib/native/client/src/clientlib/channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+}
+
+void ConnectionEndpoint::parseConnectString(){
+    boost::regex connStrExpr("(.*)=(((.*):([0-9]+),?)+)(/.+)?");
+    boost::cmatch matched;
+
+    if(boost::regex_match(m_connectString.c_str(), matched, connStrExpr)){
+        m_protocol.assign(matched[1].first, matched[1].second);
+        if(isDirectConnection()){
+            m_host.assign(matched[4].first, matched[4].second);
```
laurentgo: looks like you can also do `m_host = matched[4].str()`

contrib/native/client/src/clientlib/channel.cpp
```diff
@@ -0,0 +1,448 @@
...
+                             << std::endl;)
+    } else {
+        DRILL_MT_LOG(DRILL_LOG(LOG_DEBUG) << "Invalid connect string. Regexp did not match" << std::endl;)
+    }
+    return;
+}
+
+bool ConnectionEndpoint::isDirectConnection(){
+    assert(!m_protocol.empty());
+    return (!strcmp(m_protocol.c_str(), "local") || !strcmp(m_protocol.c_str(), "drillbit"));
```
laurentgo: let's not use strcmp if we can avoid it. `mprotocol != "local"` should work fine.

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -0,0 +1,236 @@
...
+            ChannelContext(DrillUserProperties* props):m_properties(props){};
+            virtual ~ChannelContext(){};
+            const DrillUserProperties* getUserProperties() const { return m_properties;}
+        protected:
+            DrillUserProperties* m_properties;
+    };
+
+    class SSLChannelContext: public ChannelContext{
+        public:
+            static boost::asio::ssl::context::method getTlsVersion(std::string version){
```
laurentgo: version should be a cref

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -0,0 +1,236 @@
...
+                if(version.empty()){
```
laurentgo: not necessary (since there's a catch-all)

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -0,0 +1,236 @@
...
+                    return boost::asio::ssl::context::tlsv12;
+                } else if (version == "tlsv12") {
+                    return boost::asio::ssl::context::tlsv12;
+                } else if (version == "tlsv11") {
+                    return boost::asio::ssl::context::tlsv11;
+                } else if (version == "sslv23") {
```
laurentgo: spec didn't mention support for ssl protocol. Shouldn't we disable support for those (since there are pretty insecure)? consensus is that tls1.2 should be used, so maybe it should be set as the mininum?

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -0,0 +1,236 @@
...
+            ~SSLChannelContext(){};
+            boost::asio::ssl::context& getSslContext(){ return m_SSLContext;}
+        private:
+            boost::asio::ssl::context m_SSLContext;
+    };
+
+    typedef ChannelContext ChannelContext_t; 
+    typedef SSLChannelContext SSLChannelContext_t; 
+
+    class ChannelContextFactory{
```
laurentgo: I wonder about having one factory for context and one for channel, vs one factory for a given channel type (to create both channel and socket)

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -65,108 +56,72 @@ struct ToRpcType: public std::unary_function<google::protobuf::int32, exec::user
...
-    DRILL_MT_LOG(DRILL_LOG(LOG_TRACE) << "Connecting to endpoint: " << host << ":" << port << std::endl;)
-    std::string serviceHost;
-    for (size_t i = 0; i < props->size(); i++) {
-        if (props->keyAt(i) == USERPROP_SERVICE_HOST) {
-            serviceHost = props->valueAt(i);
-        }
+    connectionStatus_t ret = CONN_SUCCESS;
+    m_pChannelContext = ChannelContextFactory::getChannelContext(type, props);
+    m_pChannel= ChannelFactory::getChannel(type, m_io_service, connStr);
+    ret=m_pChannel->init(m_pChannelContext);
```
laurentgo: I guess it could be passed at construct time (or even better, the channel creates its own context?)

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -250,7 +205,15 @@ void DrillClientImpl::doWriteToSocket(const char* dataPtr, size_t bytesToWrite,
     // Write all the bytes to socket. In case of error when all bytes are not successfully written
     // proper errorCode will be set.
     while(1) {
-        size_t bytesWritten = m_socket.write_some(boost::asio::buffer(dataPtr, bytesToWrite), errorCode);
+        size_t bytesWritten;
+        {
+            boost::lock_guard<boost::mutex> lock(m_channelMutex);
```
laurentgo: it seems weird that a mutex is now required. Before, the socket was existing, why it would not be the case for the channel?

contrib/native/client/src/clientlib/wincert.ipp
```diff
@@ -0,0 +1,98 @@
...
+#include <tchar.h>
+
+
+#pragma comment (lib, "crypt32.lib")
+#pragma comment (lib, "cryptui.lib")
+
+#define MY_ENCODING_TYPE  (PKCS_7_ASN_ENCODING | X509_ASN_ENCODING)
+
+inline
+int loadSystemTrustStore(const SSL *ssl, std::string& msg) {
```
laurentgo: maybe it should be a configurable callback where the drill client provides the certificate, and the callback returns if the certificate is valid or not? this way it would be easier to add support for system keystore support on mac or linux too...

contrib/native/client/src/include/drill/drillConfig.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+#ifndef DRILL_CONFIG_H
+#define DRILL_CONFIG_H
+
+#include "drill/common.hpp"
+#include <boost/thread.hpp>
+
+
+
+#if defined _WIN32 || defined __CYGWIN__
```
laurentgo: maybe should be in a common header since used in multiple places?

contrib/native/client/src/include/drill/drillConfig.hpp
```diff
@@ -0,0 +1,153 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#ifndef DRILL_CONFIG_H
+#define DRILL_CONFIG_H
+
+#include "drill/common.hpp"
+#include <boost/thread.hpp>
```
laurentgo: only using mutex I believe (defined in `<boost/thread/mutex.hpp>`)

contrib/native/client/src/clientlib/channel.hpp
parthchandra: I made it a const reference (or were you suggesting std::cref?)

contrib/native/client/src/clientlib/channel.hpp
parthchandra: Done

contrib/native/client/src/clientlib/channel.hpp
parthchandra: Agree that the outdated, insecure versions should not be supported. The Java version documents that we will support TLS 1, TLS 1.2, and TLS 1.2,  so I left those three in there with TLS1.2 as the default.

contrib/native/client/src/clientlib/channel.hpp
parthchandra: Done. That got rid of a bunch of code.

contrib/native/client/src/clientlib/wincert.ipp
parthchandra: And that is how boost does it too. But then checking for the certificate is complicated stuff, so it is easier to just load the certificates from the system store into openssl and let openssl deal with it.
Anyway, I think that this counts as an enhancement. I can take care of this when I add support for Mac Keychain.

contrib/native/client/src/include/drill/drillConfig.hpp
parthchandra: Done

contrib/native/client/src/include/drill/drillConfig.hpp
parthchandra: Done. It was already in common.hpp

contrib/native/client/src/clientlib/drillClientImpl.cpp
parthchandra: Oh this was found by Rob Wu. The problem occurs when the heartbeat timer has gone off and we are in the handler which is about to send off a heartbeat. Before the heartbeat is sent if the caller deletes DrillClientImpl (via DrillClient) then the channel may be closed and the pointer to the channel may be set to null causing the heartbeat send to crash. This did not occur previously because the socket was not a pointer and/or boost was able to handle it quite nicely.

contrib/native/client/src/clientlib/wincert.ipp
laurentgo: it looks like boost::asio support both loading a file and/or a verify callback:
- http://www.boost.org/doc/libs/1_47_0/doc/html/boost_asio/reference/ssl__context/set_verify_callback/overload1.html
- http://www.boost.org/doc/libs/1_47_0/doc/html/boost_asio/reference/ssl__context/load_verify_file.html

It seems you wouldn't even need to access the native handle when using these functions

contrib/native/client/src/clientlib/channel.hpp
laurentgo: sorry, it's abbreviation for const ref (I believe std::cref was introduced in C++11, which we are not using sadly, although there's a boost equivalent of course).

contrib/native/client/src/clientlib/wincert.ipp
parthchandra: I'm actually already using these methods (see `SSLStreamChannel::init()`). The verification callback implements validating the certificate. In our case we are using the boost provided rfc2818 verification method. The load verify file should point to the truststore containing the certificates in pem format. OpenSSL will read this file and load the certificate into its in-memory X509 certificate store.
The `loadSystemTrustStore` method reads the certificates from the Windows store (probably the registry) converts from the native store format into X509 and then loads it into the in-memory store. After that OpenSSL takes over and does the verification. 
For Keychain, we will have to do something similar. 
Writing our own certificate verification is going to be error prone, especially if you want to do rfc2818 verification. Not sure I'm up to it :(.

parthchandra:
@arina-ielchiieva  The `TestSSLConfig.testMissingKeystorePassword` test will fail if the exception is not thrown. The test is checking that the exception is in fact being thrown. Anyway, I added back the assertTrue check as well. It doesn't hurt to have it there.
Re the failing `TestUserBitSSL.testClientConfigHostnameVerification`. The test generates a self signed certificate with the local machine host name
```
      String fqdn = InetAddress.getLocalHost().getHostName();
      SelfSignedCertificate certificate = new SelfSignedCertificate(fqdn);
```
It then connects to the Drillbit which, one assumes, should have the same hostname so we can test that the certificate's hostname is being validated.  
I really don't know why this would fail unless you have your test machine hostname not set or set incorrectly. It should work for all cases where the hostname is the default localhost.
It does appear that this test might cause angst down the road and it might be best to disable it for the general case.

exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/security/TestUserBitSSL.java
```diff
@@ -201,6 +202,7 @@ public void testSSLQuery() throws Exception {
     test("SELECT * FROM cp.`region.json`");
   }
 
+  @Ignore
```
arina-ielchiieva: Parth, can you please add explanation (`@Ignore("some reason")`) why test is ignored (it would be really helpful in future).

exec/java-exec/src/test/java/org/apache/drill/exec/TestSSLConfig.java
```diff
@@ -76,7 +76,7 @@ public void testMissingKeystorePassword() throws Exception {
       fail();
       //Expected
     } catch (Exception e) {
-
+      assertTrue(e instanceof DrillException);
```
arina-ielchiieva: Thanks, for adding `assertTrue`.

exec/java-exec/src/test/java/org/apache/drill/exec/rpc/user/security/TestUserBitSSL.java
parthchandra: Will do.

parthchandra:
@laurentgo are you good with this? I'll open a new JIRA for the enhancements to the C++ client.

parthchandra:
Squashed and rebased on master again.

parthchandra:
Rebased again. Also updated protobuf files. 
If there are no further comments, I'll be merging this PR in.

paul-rogers:
Go for it. I won't do any more commits until you give the all-clear.

PR #952, opened 2017-09-21
DRILL-5438: Amazon S3 bucket can't be queried directly at root

xhochy:


xhochy:
It would be nice if someone could point me to the location where I should add a unit test for this.

paul-rogers:
Unit tests for S3 are difficult as we can't guarantee that each developer who runs the tests has visibility. @ilooner is adding test categories. Perhaps we can use one of those to create S3 tests that require some specific S3 setup, perhaps in a user-specific Maven properties file. Thoughts?

PR #954, opened 2017-09-22
DRILL-5803: Show the hostname for each minor fragment in operator table

kkhatua:
Based on DRILL-4909, list hostnames for each minor fragment within an operator table to identify any problematic host

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
@@ -73,8 +73,8 @@ public class ProfileWrapper {
     this.majorFragmentTallyTotal = tallyMajorFragmentCost(majors);
 
     final List<OperatorWrapper> ows = new ArrayList<>();
-    // temporary map to store (major_id, operator_id) -> [(op_profile, minor_id)]
-    final Map<ImmutablePair<Integer, Integer>, List<ImmutablePair<OperatorProfile, Integer>>> opmap = new HashMap<>();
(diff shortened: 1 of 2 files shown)


kkhatua:
@paul-rogers Can you please review this minor UI update?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -75,24 +75,31 @@ public ProfileWrapper(final QueryProfile profile) {
     final List<OperatorWrapper> ows = new ArrayList<>();
     // temporary map to store (major_id, operator_id) -> [(op_profile, minor_id)]
     final Map<ImmutablePair<Integer, Integer>, List<ImmutablePair<OperatorProfile, Integer>>> opmap = new HashMap<>();
+    // temporary map to store (major_id, operator_id) -> [minorFragHostname]
```
paul-rogers: Effectively, what this is doing is converting the opmap from a map of immutable pairs to a map of immutable triples. About here, we might want to restructure the map to point to a class which holds the two elements in the first pair, plus the host name.

Why? Simpler. Easier to keep class fields in sync than to keep two separate lists in sync.

kkhatua: +1 on this comment. However, I didn't what to make the PR look to radically alter what is already there. I'll incorporate the change. This should go nicely with a fix for DRILL-5802 and make life a bit easier

kkhatua:
@paul-rogers  Made the changes

paul-rogers:
The fixes are an improvement. In some future commit, an actual class would be better than nested immutable pairs. At present, the code has become a bit cryptic and Python-like. But, we can do that improvement later.
+1

PR #959, opened 2017-09-25
DRILL-5816: Hash function produces skewed results on String values wi…

sohami:
…th same leading prefix

            Note: Changing hash32 computation to use Murmur3.hash32 instead of int casted version of Murmur3.hash64

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/MurmurHash3.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/MurmurHash3.java
@@ -259,21 +259,20 @@ public final class MurmurHash3 extends DrillHash{
     return (int)lh1;
   }
 
-  public static long hash64(double val, long seed){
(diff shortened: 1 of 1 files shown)


sohami:
@amansinha100 - Please help to review.

paul-rogers:
The change looks benign. I wonder; do we have a test case that exercises the hash functions so we can be certain that the fix actually solves the original problem?

sohami:
@paul-rogers - I have added few tests and findings using hash32 and hash64 to compute the 32 bit hash codes in JIRA.

amansinha100:
lgtm.  +1.  @sohami could you pls check with @chunhui-shi if unit tests were added previously for  DRILL-4237 (another skew issue)?  If not, could you create a new JIRA to add such tests, which should include the data set for DRILL-4237, DRILL-4119, DRILL-5816 among others.

sohami:
I have created [DRILL-5287](https://issues.apache.org/jira/browse/DRILL-5827) for adding tests.

PR #962, opened 2017-09-26
DRILL-5820: Add support for libpam4j Pam Authenticator

sohami:


sohami:
@parthchandra - Please help to review

distribution/src/resources/drill-override-example.conf
```diff
@@ -133,6 +133,8 @@ drill.exec: {
   security.user.auth {
     enabled: false,
     packages += "org.apache.drill.exec.rpc.user.security",
+    # There are 2 implementations available "pam" using JPAM
+    # and "pam4j" using libpam4j
```
paul-rogers: As discussed in person, perhaps explain how the `impl` property here related to the implementation class. (Evidently via an annotation. So, maybe just name the annotation used.)

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/security/Pam4jUserAuthenticator.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Implement {@link org.apache.drill.exec.rpc.user.security.UserAuthenticator} based on Pluggable Authentication
+ * Module (PAM) configuration. Configure the PAM profiles using "drill.exec.security.user.auth.pam_profiles" BOOT
+ * option. Ex. value  <i>[ "login", "sudo" ]</i> (value is an array of strings).
+ */
+@UserAuthenticatorTemplate(type = "pam4j")
+public class Pam4jUserAuthenticator implements UserAuthenticator {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Pam4jUserAuthenticator.class);
+
+  private List<String> profiles;
```
paul-rogers: `final`?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/security/Pam4jUserAuthenticator.java
```diff
@@ -0,0 +1,81 @@
...
+        logger.error("PAM auth failed for user: {} against {} profile. Exception: {}", user, profile, ex.getMessage());
+        throw new UserAuthenticationException(String.format("PAM auth failed for user: %s using profile: %s",
+            user, profile));
+      } finally {
+        if (pam != null) {
+          pam.dispose();
+        }
+      }
+
+      if (!user.equals(unixUser.getUserName())) {
```
paul-rogers: `equals` or `equalsIgnoreCase`? That is, should "Fred" match "fred"?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/security/Pam4jUserAuthenticator.java
```diff
@@ -0,0 +1,81 @@
...
+        throw new UserAuthenticationException(String.format("Unexpected error from pam module. Input user %s is " +
+            "different from authenticated output user %s of pam module libpam4j", user, unixUser.getUserName()));
+      }
+
+      if (logger.isTraceEnabled()) {
```
paul-rogers: Can omit this if statement, `logger.trace()` will do teh right thing.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/security/PamUserAuthenticator.java
```diff
@@ -59,7 +59,8 @@ public void authenticate(String user, String password) throws UserAuthentication
     for (String pamProfile : profiles) {
       Pam pam = new Pam(pamProfile);
       if (!pam.authenticateSuccessful(user, password)) {
-        throw new UserAuthenticationException(String.format("PAM profile '%s' validation failed", pamProfile));
+        throw new UserAuthenticationException(String.format("PAM profile '%s' validation failed for user %s",
```
paul-rogers: The module above throws an exception. Here we get a `boolean`. Does this mean that the exception was mapped to `false`, then mapped back to an exception?

sohami: No the exceptions thrown here is directly mapped to Sasl Exception and sent back to client with AUTH_FAILED status.

distribution/src/resources/drill-override-example.conf
sohami: Updated the comment here to reflect the Annotation name `UserAuthenticatorTemplate`

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/security/Pam4jUserAuthenticator.java
sohami: cannot be final since it is initialized in `setup(..)` method

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/security/Pam4jUserAuthenticator.java
sohami: Removed.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/security/Pam4jUserAuthenticator.java
sohami: Some *Nix systems supports case sensitive usernames and some doesn't. Like Mac is case insensitive while linux is case sensitive. So I think username should be then treated as case sensitive. I will remove this check since looks like it's redundant and the pam module authenticate should take care of this.

PR #963, opened 2017-09-26

kkhatua:
Added an additional field in the UI allowing a user to specify the max number of profiles to display

### DIFF
--- a/exec/java-exec/src/main/resources/rest/profile/list.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/list.ftl
@@ -37,7 +37,42 @@
       <strong>No running queries.</strong>
     </div>
   </#if>
-  <h3>Completed Queries</h3>
+  <table width="100%">
+    <script type="text/javascript" language="javascript">
+    //Validate that the fetch number is valid
+    function checkMaxFetch() {
+      var maxFetch = document.forms["profileFetch"]["max"].value;
+      console.log("maxFetch: " + maxFetch);
+      if (isNaN(maxFetch) || (maxFetch < 1) || (maxFetch > 100000) ) {
+        alert("Invalid Entry: " + maxFetch + "\n" +
+               "Please enter a valid number of profiles to fetch (1 to 100000) ");
+        return false;
+      }
+      return true;
+    }
+    </script>
+    <tr>
+      <td><h3>Completed Queries</h3></td>
+      <td align="right">
+        <form name="profileFetch" action="/profiles" onsubmit="return checkMaxFetch();" method="get"><span title="Max number of profiles to list">Showing <b>${model.getFinishedQueries()?size}</b> profiles. Max: </span>
+        <input id="fetchMax" type="text" size="7" name="max" value="">
+        <input type="submit" value="Refresh">
+      </form></td>
+    </tr></table>
+    <!-- Placed after textbox to allow for DOM to contain "fetchMax" element -->
+    <script type="text/javascript" language="javascript">
+    //Get max fetched from URL for populating textbox
+    var maxFetched="${model.getMaxFetchedQueries()}";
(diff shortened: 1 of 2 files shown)


kkhatua:
@paul-rogers Opened a new PR as I accidentally closed #953 . This is ready for review.

exec/java-exec/src/main/resources/rest/profile/list.ftl
```diff
@@ -37,7 +37,15 @@
       <strong>No running queries.</strong>
     </div>
   </#if>
-  <h3>Completed Queries</h3>
+  <table width="100%">
+    <tr>
+      <td><h3>Completed Queries</h3></td>
+      <td align="right">
+        <form action="/profiles" method="get"><span title="Max number of profiles to list">Show </span>
+        <input id="fetchMax" type="text" size="7" name="max" value="${model.getFinishedQueries()?size}">
```
paul-rogers: Better! But, the number to show is the *target* number of profiles, not the *actual* number. If we show the actual number, then we get this:

* Show profiles with the target of 100.
* We actually have 10 profiles, so that is displayed in the "max" control.
* Two more profiles come in, now we have 12.
* Hit the "Go" button. Only 10 profiles appear because that was the prior count.
* Must manually type a larger number to see more.

What we want:

* Show profiles with the target of 100.
* We actually have 10 profiles, but the target is 100, so 100 is displayed in the "max" control.
* Two more profiles come in, now we have 12.
* Hit the "Go" button. All 12 profiles appear.

To make this work:

* Add to the model the *target* number: either the default of 100 or the number received via the max parameter.
* Display that model number in the "max" control.

exec/java-exec/src/main/resources/rest/profile/list.ftl
```diff
@@ -37,7 +37,15 @@
...
+        <input type="submit" value="Go">
+      </form></td>
+    </tr></table>
```
paul-rogers: BTW: we should validate the max parameter in the Java code. Force max into the range: 1 ... 500 (say).

That is, don't allow -1, 0 or 1000000 profiles.

kkhatua:
@paul-rogers please bless this
- Checks in place for non-numeric or out-of-range numeric inputs.
- Auto fill if the max requested range exceeds the actual number of profiles available. User would not need to re-enter a value, just hit `Go`

exec/java-exec/src/main/resources/rest/profile/list.ftl
```diff
@@ -37,7 +37,42 @@
...
+        return false;
+      }
+      return true;
+    }
+    </script>
+    <tr>
+      <td><h3>Completed Queries</h3></td>
+      <td align="right">
+        <form name="profileFetch" action="/profiles" onsubmit="return checkMaxFetch();" method="get"><span title="Max number of profiles to list">Showing <b>${model.getFinishedQueries()?size}</b> profiles. Max: </span>
+        <input id="fetchMax" type="text" size="7" name="max" value="">
```
paul-rogers: Here we want to show the current fetch max setting. This is either:

* The default number, if the user has nor provided a value.
* The last submitted value.

Let's say the default is 100. Then we'd want to first see:

Max: [  100] [Fetch]

The user decides to view just 20 by changing the number and clicking Fetch. Now we'd want to see:

Max: [   20] [Fetch]

Also, "Fetch" is a bit awkward. Maybe "Refresh" would be better.

kkhatua: Done. 
1. The keyword is now 'Refresh'.
2. The default fetch is populated into the textbox. Any subsequent changes are retained, so the user does not have to re-enter a value to trigger a reload. Only click on Refresh.

paul-rogers:
Thanks for the improvements.
LGTM
+1

PR #964, opened 2017-09-27
DRILL-5755 Reduced default number of batches kept in memory by TopN

ilooner:
This change prevents the TopN operator from consuming large amounts of memory by default. This is important since the TopN operator doesn't obey memory limits currently. This is not a true fix for the issue, but it will reduce the number of issues seen until a true fix is developed.


ilooner:
@Ben-Zvi @paul-rogers

PR #965, opened 2017-09-28
DRILL-5811 reduced repeated log messages further.

ilooner:
This reduces the number of messages printed for blocks that are on nodes on which there is no drillbit running. Now the number of messages printed is equal to the number of nodes on which there are blocks that don't have drillbits. For example if there are 6 data nodes and only 2 of those nodes have drill bits, then exactly 4 messages will be printed. This time around I have tested this on my own physical cluster.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/schedule/BlockMapBuilder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/schedule/BlockMapBuilder.java
@@ -104,12 +104,16 @@ public class BlockMapBuilder {
     @Override
     protected List<CompleteFileWork> runInner() throws Exception {
       final List<CompleteFileWork> work = Lists.newArrayList();
+
+      final Set<String> noDrillbitHosts = logger.isDebugEnabled() ? Sets.<String>newHashSet() : null;
+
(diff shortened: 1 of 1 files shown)


ilooner:
@paul-rogers

exec/java-exec/src/main/java/org/apache/drill/exec/store/schedule/BlockMapBuilder.java
```diff
@@ -104,12 +104,16 @@ public BlockMapReader(FileStatus status, boolean blockify) {
     @Override
     protected List<CompleteFileWork> runInner() throws Exception {
       final List<CompleteFileWork> work = Lists.newArrayList();
+
+      final Set<String> noDrillbitHosts = logger.isDebugEnabled() ? Sets.<String>newHashSet() : null;
```
vrozov: Consider moving `noDillbitHosts` to `BlockMapBuilder` class (use `Sets.newConcurrentHashSet()` in this case) as it does not seem to belong to `BlockMapReader`. With such change, other changes are not necessary and likely this will allow reducing repeated log messages even further. Drop `<Sring>` from `Sets.<String>newHashSet()`.

PR #967, opened 2017-09-29
DRILL-5564: Added finally block for stopWait() to avoid all situation…

KulykRoman:
…s where Drill able to miss stopWait() in case of exceptions (it can lead to assertions).


priteshm:
@parthchandra can you review this?

PR #968, opened 2017-09-30
DRILL-5830: Resolve regressions to MapR DB from DRILL-5546

paul-rogers:
DRILL-5546 fixed a wide variety of "empty batch" problems. But, it introduced a regression in the HBase and MapR-DB binary storage plugins. This PR refines the fix to resolve those regressions.

Prior to DRILL-5546, HBase provided a project push-down rule to expand wildcard columns. However, a bug in the push-down rule prevented proper function. DRILL-5546 fixed that rule. But, DRILL-5546 also explicitly expanded wildcards for HBase, which turned out to be redundant, and so is backed out in this PR.

Wildcard expansion in the HBase storage plugin was meant to overcome the schema change conflict that occurs with empty regions. In such regions, we get the row key and the column family as an empty map. In regions with data, we get the row key and a non-empty map for the column family. Examples:

* Empty: (row_key, cf{})
* Non-empty: (row_key, cf{col1, col2})

Where cf is a column family and col1, col2 are columns.

It turns out that the receivers were getting confused. The `RecordBatchLoader` class treated empty and non-empty maps as an identical schema. This was mentioned in  DRILL-5546:

> In HBase a column family always has map type, and a non-rowkey column always has nullable varbinary type, this ensures that HBaseRecordReader across different HBase regions will have the same top level schema, even if the region is empty or prune all the rows due to filter pushdown optimization. In other words, we will not see different top level schema from different HBaseRecordReader for the same table.

The problem is, a difference in map content really is a schema change, so we need to detect and report it. This PR makes that change.

Now, as it turns out, changes made by DRILL-5546 to the top-level project operator gracefully removes the empty batches (with empty maps), passing along just the non-empty batches (with non-empty maps.)

In short, this PR:

* Backs out the HBase-specific changes to DRILL-5546,
* Fixes the schema change issue in `RecordBatchLoader`,
* Adds unit tests for the fixes to `RecordBatchLoader`,
* Does a number of minor code cleanups.

The result is that the HBase problems that DRILL-5546 solved are still solved, but the regressions to MapR DB binary are also fixed.

We propose to modify the MapR DB binary storage plugin to do projection push-down the same way as is done in HBase, but that will be a separate PR.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchLoader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchLoader.java
@@ -19,12 +19,15 @@ package org.apache.drill.exec.record;
 
 import io.netty.buffer.DrillBuf;
 
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
 import org.apache.drill.common.StackTrace;
 import org.apache.drill.common.expression.SchemaPath;
+import org.apache.drill.common.map.CaseInsensitiveMap;
+import org.apache.drill.common.types.TypeProtos.MinorType;
 import org.apache.drill.exec.exception.SchemaChangeException;
 import org.apache.drill.exec.expr.TypeHelper;
 import org.apache.drill.exec.memory.BufferAllocator;
@@ -38,7 +41,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Maps;
 
 
 /**
@@ -85,7 +87,7 @@ public class RecordBatchLoader implements VectorAccessible, Iterable<VectorWrapp
     // the schema has changed since the previous call.
 
     // Set up to recognize previous fields that no longer exist.
-    final Map<String, ValueVector> oldFields = Maps.newHashMap();
+    final Map<String, ValueVector> oldFields = CaseInsensitiveMap.newHashMap();
     for(final VectorWrapper<?> wrapper : container) {
       final ValueVector vector = wrapper.getValueVector();
       oldFields.put(vector.getField().getName(), vector);
@@ -109,6 +111,17 @@ public class RecordBatchLoader implements VectorAccessible, Iterable<VectorWrapp
           vector.clear();
           schemaChanged = true;
           vector = TypeHelper.getNewVector(fieldDef, allocator);
+
+        // If the field is a map, check if the map schema changed.
+
+        } else if (vector.getField().getType().getMinorType() == MinorType.MAP  &&
+                   ! isSameSchema(vector.getField().getChildren(), field.getChildList())) {
+
+          // The map schema changed. Discard the old map and create a new one.
+
+          schemaChanged = true;
+          vector.clear();
+          vector = TypeHelper.getNewVector(fieldDef, allocator);
         }
 
         // Load the vector.
@@ -120,7 +133,6 @@ public class RecordBatchLoader implements VectorAccessible, Iterable<VectorWrapp
         bufOffset += field.getBufferLength();
         newVectors.add(vector);
       }
-
 
       // rebuild the schema.
       final SchemaBuilder builder = BatchSchema.newBuilder();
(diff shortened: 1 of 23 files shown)


exec/java-exec/src/test/java/org/apache/drill/test/QueryRowSetIterator.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.drill.exec.proto.UserBitShared.QueryId;
+import org.apache.drill.exec.proto.UserBitShared.QueryResult.QueryState;
+import org.apache.drill.exec.proto.helper.QueryIdHelper;
+import org.apache.drill.exec.record.RecordBatchLoader;
+import org.apache.drill.exec.record.VectorContainer;
+import org.apache.drill.exec.rpc.user.QueryDataBatch;
+import org.apache.drill.test.BufferingQueryEventListener.QueryEvent;
+import org.apache.drill.test.rowSet.DirectRowSet;
+
+public class QueryRowSetIterator implements Iterator<DirectRowSet>, Iterable<DirectRowSet> {
```
vdiravka: I have a compilation error here. Please implement remove() method from Iterator interface.

paul-rogers: Sorry, this is a Java 8 thing. In Java 8, `remove()` has a default implementation. In Java 7, the method must be implemented. Fixed.

jinfengni:
I'm not convinced that it's a good idea to back out the change to HBase specific changes made in DRILL-5546.

You are right that project push-down in planner ideally should do its job and push the list of columns. However, until planner could claim there is no issues at all ( just like what happened prior to DRILL-5546), execution still may face a columns list with "*". That's why in HBaseGroupScan we have verifyColumnsAndConvertStar, just in case planner's project push-down did not work in the way we want.  If as you suggested, such conversion in HBaseGroupScan is redundant, why would it cause regression (if planner's rule works as expected)? or is it your intention to still keep "*" in HBaseRecordReader? If that's your intention, I think we are going in the wrong direction. HBase has a unified schema at table level.  

I agree that the analysis of empty map {},  vs {a:varbinary}.  It's something we have to deal with. As a matter of fact, such scenarios does not have to come from empty batch. It could happen with two regions with >0 rows.  

For instance, regrion 1 has 10 rows, with cf1.c1 appears in only first 5 rows, while region 2 has 20 rows with cf1.c1 appears in every rows. For the following query:
select CF1 FROM table where some_condition_on_row_key;

if "some_condition_on_row_key" is pushed to hbase and prunes the first 5 rows, region1 will return a batch with 5 rows, but with cf1 as an empty map, while region2 will have map with cf1 as {c1:varbinary}.  

In that sense, DRILL-5546 exposes such issues, and force us to have a solution to handle empty map {} vs {a:varbinary}/

paul-rogers:
Rebased onto latest master.

paul-rogers:
@jinfengni, how unreliable is the project push-down rule? In general, it is hard to reason about software when parts are unreliable. Should every use of the projection push down rule implement a "just in case" fallback as was done for HBase? Should we add that to all storage plugins?

Or, can we fix the projection push-down rule to make it more reliable? It seems that one of your fixes did just that. An HBase query before your fix to the rule didn't expand the wildcard, but the same query, after your projection rule fix, does expand the wildcard.

What are the scenarios in which the rule does not work? Can we fix them in the planner?

jinfengni:
Project push-down rule may not work in the way we want in at least two cases: 1) there is a bug in the code that we are not aware of, 2) it would not push the list of columns for "select *", when the table is a dynamic table. For HBase, that's not an issue. But it's an issue for MapRDB binary table, as it's processed as a table on a file system, in stead of following the logic of HBase in planning time. I could not comment why MapRDB binary table was handled that way. But the reality is that you still will see column star in execution for MapRDB, if we do not keep the "redundant" code in execution (which essentially converts * into row_key/column family). 

If we keep column star in record reader's columns list, then we would end up with nullable-int vs map column. If we expand column star, then we may end up with empty map vs a non-empty map. To me, it's easier to handle the latter case. Plus it's something we have to deal with in case of HBase/JSON plugin.

paul-rogers:
Thanks for the explanation! Bugs in the code are always a concern, but I personally prefer to fix them rather than write more code that tries to work around potential, unknown bugs in other code. (Working around known bugs is, of course, part of the game.)

Sounds like we'll need to test the MapR-DB case for the conditions you identified; that will tell is if the HBase solution (which seems to work now) can also be applied to MapR DB.

vdiravka:
With separate table class for MapR-DB we can follow the same approach of expanding column star into list as for HBase plugin. For that reason I've created a [PR-973 (DRILL-5775)](https://github.com/apache/drill/pull/973) for it.

@jinfengni Please take a look at this PR as well.

exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestLoad.java
```diff
@@ -119,7 +122,283 @@ public void testLoadValueVector() throws Exception {
...
+      assertFalse(loadBatch(allocator, batchLoader, schema));
+
+      // Potential bug: see DRILL-5828
+
+      assertTrue(schema.isEquivalent(batchLoader.getSchema()));
+      batchLoader.getContainer().zeroVectors();
+    }
+
+    // Drop a column: a: INT
+    // Schema change: ?
```
arina-ielchiieva: If we drop column ,looks like we think schema change has occurred, right? Should we replace to `Schema change: Yes`?

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchLoader.java
```diff
@@ -150,6 +162,40 @@ public boolean load(RecordBatchDef def, DrillBuf buf) throws SchemaChangeExcepti
     return schemaChanged;
   }
 
+  private boolean checkMapSchema(Collection<MaterializedField> currentChildren,
```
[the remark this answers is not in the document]
paul-rogers: Renamed and added Javadoc comments to better explain the purpose.

exec/java-exec/src/main/java/org/apache/drill/exec/util/VectorUtil.java
```diff
@@ -42,7 +43,12 @@ public static void showVectorAccessibleContent(VectorAccessible va, final String
     System.out.println(rows + " row(s):");
     List<String> columns = Lists.newArrayList();
     for (VectorWrapper<?> vw : va) {
-      columns.add(vw.getValueVector().getField().getName());
+      MaterializedField field = vw.getValueVector().getField();
```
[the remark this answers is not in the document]
paul-rogers: Fixed.

exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestLoad.java
paul-rogers: Fixed.

PR #970, opened 2017-10-03
DRILL-5832: Migrate OperatorFixture to use SystemOptionManager rather than mock

paul-rogers:
The `OperatorFixture` provides structure for testing individual operators and other "sub-operator" bits of code. To do that, the framework provides mock network-free and server-free versions of the fragment context and operator context. As part of the mock, the `OperatorFixture` provides a mock version of the system option manager that provides a simple test-only implementation of an option set.

With the recent major changes to the system option manager, this mock implementation has drifted out of sync with the system option manager. Rather than upgrading the mock implementation, this change uses the system option manager directly – but configured for no ZK or file persistence of options.

The key reason for this change is that the system option manager has implemented a sophisticated way to handle option defaults; it is better to leverage that than to provide a mock implementation.

Work is divided into a series of commits. The first is just routine code cleanup.

### Options Manager Updates

The second commit introduces three changes to the options manager:

* Provide a test-only constructor for the system option manager that “persists” options to memory only.
* Provide classic `getType(String name)` methods for the option manager with types `Long`, `Double`, `Boolean` and `String`. This allows clients to provide only a name, not an option validator, to get an option value.
* In the `setLocalValue(String name, Object value)` method, allow values that are `int`s (mapped to `long`s) or `float`s (mapped to `double`s). This saves having to remember to write things like 10L or 1.0D in test code.
* Since we are now using the system option manager for the test framework, the `BaseOptionSet` class is no longer needed and was removed.

### Use System Option Manager for Tests

The next commit deals with replacing the ad-hoc `TestOptionSet` with the `SystemOptionManager` class.

* The `SystemOptionManager` is extended with a test-only constructor that uses an in-memory “persistent” store to avoid polluting the disk or ZK with test-only settings.
* The `OperatorFixtureBuilder` class is extended to gather a set of name/value pairs for options to set. These are set in the `OperatorFixture` once the system option manager is created. This mechanism uses the generic type mechanism to allow the values to be any valid Java type, rather than just strings. That is:

```
OperatorFixtureBuilder builder = …
builder.systemOption("myOption", 10);
builder.systemOption("yourOption", "yourValue");
…
```

### Rename FixtureBuilder

Drill now has two “fixtures”: `ClusterFixture` and `OperatorFixture`. The class `FixtureBuilder` is confusing: which fixture does it build? To resolve this, the class was renamed to `ClusterFixtureBuilder` to be consistent with other fixture builders.

### Use `ALTER SESSION RESET`

Many tests set a session option, then reset the option back to the defaults. Much code did this the hard way:

* Get the default value of the option from the code
* Set the session option using this default.

In removing the `TestOptionSet` class, we removed a method used to get the default. We need an alternative.

As it turns out, there is a cleaner way to accomplish the goal. Rather than forcing the option to the default value explicitly, we can just use the `ALTER SESSION RESET` command.

A new method, `resetSystemOption`, was added to `BaseTestQuery` and to `ClientFixture`. For tests that don’t use these classes, special one-off solutions were added.

This change has the added benefit of being more accurate. In the options system, an option set to the (presumed) default value is not the same as an unset option that automatically reverts to the default. Using `ALTER SESSION RESET` therefore puts the system back into the true "default" state where as explicitly setting the default does not. (And, trying to explicitly set the default can mask bugs, as described next.)

### `TestParquetWriter` and DRILL-5833

When testing the fixes, testing revealed a random failure in `TestParquetWriter` due to inconsistent resetting of session options. As it turns out, failure to reset the option to select the “new” Parquet reader caused the `testDecimal()` test sometimes use the “old” version. When the “old” version was used, it hit a long-standing bug in the way the code reads Decimal9 and Decimal18 types.

This bug was fixed, code visited during the investigation was cleaned up, the tests modified to always reset system options to the default value, and the failing decimal test setup to run both the “new” and “old” readers.

This particular change is somewhat complex, please see the writeup in [DRILL-5833](https://issues.apache.org/jira/browse/DRILL-5833) for details.


paul-rogers:
@ilooner, please review the option-related changes. @sachouche, can you review the Parquet parts?

ilooner:
@paul-rogers Some tests are failing on my machine

```
Failed tests: 
  TestMetadataProvider.tables:154 expected:<OK> but was:<FAILED>
  TestMetadataProvider.tablesWithTableNameFilter:219 expected:<OK> but was:<FAILED>
  TestMetadataProvider.tablesWithSystemTableFilter:191 expected:<OK> but was:<FAILED>
  TestMetadataProvider.tablesWithTableFilter:180 expected:<OK> but was:<FAILED>

Tests in error: 
  TestCustomUserAuthenticator.positiveUserAuth » UserRemote SYSTEM ERROR: URISyn...
  TestCustomUserAuthenticator.positiveUserAuthAfterNegativeUserAuth » UserRemote
  TestInfoSchema.selectFromAllTables » UserRemote SYSTEM ERROR: URISyntaxExcepti...
  TestViewSupport.infoSchemaWithView:355->BaseTestQuery.testRunAndReturn:356 » Rpc
  TestInfoSchemaFilterPushDown.testFilterPushdown_NonEqual » UserRemote SYSTEM E...
  TestParquetScan.testSuccessFile:60->BaseTestQuery.testRunAndReturn:356 » Rpc o...
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestSchemaChange.org.apache.drill.TestSchemaChange » Rpc CONNECTION : java.net...
  TestSchemaChange>DrillTest.finiDrillTest:106 » Interrupted sleep interrupted
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestSelectivity.org.apache.drill.TestSelectivity » Rpc CONNECTION : java.net.N...
  TestSelectivity>DrillTest.finiDrillTest:106 » Interrupted sleep interrupted
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestSelectWithOption.org.apache.drill.TestSelectWithOption » Rpc CONNECTION : ...
  TestSelectWithOption>DrillTest.finiDrillTest:106 » Interrupted sleep interrupt...
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestDynamicUDFSupport.setup:91->BaseTestQuery.updateTestCluster:148->BaseTestQuery.updateTestCluster:166 » Runtime
  TestDynamicUDFSupport>DrillTest.finiDrillTest:106 » Interrupted sleep interrup...
  TestTpchDistributed.tpch14 » User CONNECTION ERROR: Connection /192.168.0.80:4...
  TestTpchDistributed.tpch15:104->testDistributed:28->BaseTestQuery.test:402 » User
  TestTpchDistributed.tpch16:109->testDistributed:28->BaseTestQuery.test:402 » User
  TestTpchDistributed.tpch17:114->testDistributed:28->BaseTestQuery.test:402 » User
  TestTpchDistributed.tpch18:119->testDistributed:28->BaseTestQuery.test:402 » User
  TestTpchDistributed.tpch20:135->testDistributed:28->BaseTestQuery.test:402 » User
  TestTpchDistributed.tpch19_1:130->testDistributed:28->BaseTestQuery.test:402 » User
  TestTpchDistributedConcurrent.testConcurrentQueries:195 » ConcurrentModification

Tests run: 2335, Failures: 4, Errors: 29, Skipped: 138
```

Otherwise option related changes LGTM

ilooner:
@paul-rogers It looks like the unit test failures were caused by an unrelated issue. My Mac sometimes create a wifi*.log file in /tmp . And the presence of the file in /tmp causes some unit tests to fail. Everything LGTM.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
```diff
@@ -84,14 +82,7 @@
   private RecordReader<Void> recordReader;
   private DrillParquetRecordMaterializer recordMaterializer;
   private int recordCount;
-  private List<ValueVector> primitiveVectors;
   private OperatorContext operatorContext;
```
sachouche: Can you explain the reason you are removing the existing logic?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -184,25 +183,26 @@ public void testAllScalarTypes() throws Exception {
 
     try {
       // read all of the types with the complex reader
-      test(String.format("alter session set %s = true", ExecConstants.PARQUET_NEW_RECORD_READER));
+      alterSession(ExecConstants.PARQUET_NEW_RECORD_READER, true);
```
sachouche: Paul,

- I noticed the key PARQUET_NEW_RECORD_READER is erroneous
- There are currently two readers
   o The old one is used when nested data is used as it can handle all parquet use-cases
   o The new reader only deals with Flat parquet data types
- We might want to rename the keys as the new reader cannot always be substituted with the old one

paul-rogers: Good point. Here, I was not changing the semantics of the test; rather just ensuring that the tests are deterministic in doing what they already did. Presumably the test author knew which reader to use for which test. In fact, the fact that the tests now pass indicates that the proper reader is used (or, at least, the wrong reader is not used...)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
paul-rogers: Variable was not used.

ilooner:
@paul-rogers Some of the changes I am making on top of https://github.com/apache/drill/pull/978/ as part of DRILL-5730 will likely conflict with this change. When do you think this could make it in? It would be helpful to have it merged sooner to avoid more conflicts down the line :) .

paul-rogers:
There is a funny thing about the way Drill works. I can review your changes and commit them as soon as I provide a +1. My changes must wait until another committer to find time in their very busy schedules to consider this work. So, we'll likely commit yours first, I'll rebase mine on top of it, then wait for another committer to find time to consider it. The one exception would be if a non-committer can give this PR a +1 and a committer agrees to do a bulk commit this week.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/TestHiveStorage.java
```diff
@@ -74,10 +73,8 @@ public void testNativeScanWhenNoColumnIsRead() throws Exception {
           .baselineValues(200l)
           .go();
     } finally {
-      final OperatorFixture.TestOptionSet testOptionSet = new OperatorFixture.TestOptionSet();
-      test(String.format("alter session set `%s` = %s",
-          ExecConstants.HIVE_OPTIMIZE_SCAN_WITH_NATIVE_READERS,
-          Boolean.toString(testOptionSet.getDefault(ExecConstants.HIVE_OPTIMIZE_SCAN_WITH_NATIVE_READERS).bool_val)));
+      test(String.format("alter session reset `%s`",
```
ilooner: Minor nitpick. The String.format is not necessary here. BaseTestQuery has a test method that does the String.format for you.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
```diff
@@ -208,17 +208,17 @@ public boolean wasAllPartitionsPruned() {
...
-      return "";
-    }
-
-    final List<String> files = Lists.newArrayList();
-    for (final FileStatus status : statuses) {
-      files.add(status.getPath().toString());
-    }
-    return commonPathForFiles(files);
-  }
+//  private static String commonPath(final List<FileStatus> statuses) {
```
ilooner: Could we delete this instead of comment it out?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
```diff
@@ -331,21 +320,21 @@ public int next() {
...
-    int filled = 0;
-    for (final ValueVector v : primitiveVectors) {
-      filled = Math.max(filled, v.getAccessor().getValueCount() * 100 / v.getValueCapacity());
-      if (v instanceof VariableWidthVector) {
-        filled = Math.max(filled, ((VariableWidthVector) v).getCurrentSizeInBytes() * 100 / ((VariableWidthVector) v).getByteCapacity());
-      }
-      // TODO - need to re-enable this
-//      if (v instanceof RepeatedFixedWidthVector) {
-//        filled = Math.max(filled, ((RepeatedFixedWidthVector) v).getAccessor().getGroupCount() * 100)
+//  private int getPercentFilled() {
```
ilooner: Delete?

exec/java-exec/src/test/java/org/apache/drill/BaseTestQuery.java
```diff
@@ -371,6 +373,16 @@ public static void testNoResult(String query, Object... args) throws Exception {
     testNoResult(1, query, args);
   }
 
+  public static void alterSession(String key, Object value) throws Exception {
```
ilooner: There is the setSessionOption method on line 487 that already does something similar. Could we delete those methods and their usages in favor of this one?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriter.java
```diff
@@ -725,21 +724,24 @@ public void runTestAndValidate(String selection, String validationSelection, Str
     try {
       deleteTableIfExists(outputFile);
       test("use dfs_test.tmp");
-  //    test("ALTER SESSION SET `planner.add_producer_consumer` = false");
+//    test("ALTER SESSION SET `planner.add_producer_consumer` = false");
```
ilooner: Can we delete this?

ilooner:
@paul-rogers Will this be going in anytime soon? I'd like to fix the annoying

```
test(String.format("alter session..."))
```

statements in the tests after this goes in.

paul-rogers:
Addressed comments. Rebased on master. Resolved merge conflicts. Squashed commits.

@ilooner, @sachouche please do a quick final review to check for loose ends.

arina-ielchiieva:
+1, LGTM.

PR #971, opened 2017-10-03
Drill-5834 Add Networking Functions

cgivre:
This is a collection of Networking Functions to facilitate network analysis.  The functions include:

- **inet_aton(`<ip>`)**:  Converts an IPv4 address into an integer.
- **inet_ntoa( `<int>`)**:  Converts an integer IP into dotted decimal notation
- **in_network( `<ip>`,`<cidr>` )**: Returns true if the IP address is in the given CIDR block
- **address_count( `<cidr>` )**:  Returns the number of IPs in a given CIDR block
- **broadcast_address( `<cidr>` )**: Returns the broadcast address for a given CIDR block
- **netmask(`<cidr>` )**: Returns the netmask for a given CIDR block. 
- **low_address(`<cidr>`)**:  Returns the first address in a given CIDR block.
- **high_address(`<cidr>`)**:  Returns the last address in a given CIDR block.
- **url_encode( `<url>` )**:  Returns a URL encoded string.
- **url_decode( `<url>` )**:   Decodes a URL encoded string.
- **is_valid_IP(`<ip>`)**: Returns true if the IP is a valid IP address
- **is_private_ip(`<ip>`)**: Returns true if the IP is a private IPv4 address
- **is_valid_IPv4(`<ip>`)**: Returns true if the IP is a valid IPv4 address
- **is_valid_IPv6(`<ip>`)**: Returns true if the IP is a valid IPv6 address


exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNetworkFunctions.java
```diff
@@ -0,0 +1,93 @@
...
+
+import org.apache.drill.BaseTestQuery;
+import org.junit.Test;
+
+public class TestNetworkFunctions extends BaseTestQuery {
+
+  @Test
+  public void testInetAton() throws Exception {
+    final String query = "select inet_aton( '192.168.0.1') as inet from (values(1))";
+    testBuilder().sqlQuery(query).ordered().baselineColumns("inet").baselineValues( Long.parseLong("3232235521") ).go();
```
arina-ielchiieva: Please remove spaces...

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNetworkFunctions.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Test
+  public void testInetNtoa() throws Exception {
+    final String query = "select inet_ntoa( 3232235521 ) as inet from (values(1))";
```
arina-ielchiieva: spaces

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNetworkFunctions.java
```diff
@@ -0,0 +1,93 @@
...
+    testBuilder().sqlQuery(query).ordered().baselineColumns("inet").baselineValues("192.168.0.1").go();
+  }
+
+  @Test
+  public void testInNetwork() throws Exception {
+    final String query = "select in_network( '192.168.0.1', '192.168.0.0/28' ) as in_net FROM (values(1))";
```
arina-ielchiieva: What about negative test case?

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNetworkFunctions.java
```diff
@@ -0,0 +1,93 @@
...
+    final String query = "select getBroadcastAddress( '192.168.0.0/28' ) AS broadcastAddress FROM (values(1))";
+    testBuilder().sqlQuery(query).ordered().baselineColumns("broadcastAddress").baselineValues("192.168.0.15").go();
+  }
+  @Test
+  public void testNetmask() throws Exception {
+    final String query = "select getNetmask( '192.168.0.0/28' ) AS netmask FROM (values(1))";
+    testBuilder().sqlQuery(query).ordered().baselineColumns("netmask").baselineValues("255.255.255.240").go();
+  }
+  @Test
+  public void testFunctions() throws Exception {
```
arina-ielchiieva: Well, it is still good to split test into multiple for each function.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
```diff
@@ -0,0 +1,668 @@
...
+
+    @Inject
+    DrillBuf buffer;
+
+
+    public void setup() {
+    }
+
+
+    public void eval() {
```
arina-ielchiieva: May we use `InetAddress.html#isSiteLocalAddress` instead [1, 2]?

[1] http://docs.oracle.com/javase/1.5.0/docs/api/java/net/InetAddress.html#isSiteLocalAddress%28%29
[2] https://stackoverflow.com/questions/9729378/check-whether-the-ipaddress-is-in-private-range/9729432

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
```diff
@@ -0,0 +1,668 @@
...
```
arina-ielchiieva: For `ntoa` and `aton` may we use guava lib [1, 2]?

[1] https://google.github.io/guava/releases/19.0/api/docs/com/google/common/net/InetAddresses.html
[2] https://stackoverflow.com/questions/2241229/going-from-127-0-0-1-to-2130706433-and-back-again

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNetworkFunctions.java
```diff
@@ -0,0 +1,93 @@
+/**
```
arina-ielchiieva: Apache header should be in a form of comment.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
```diff
@@ -0,0 +1,668 @@
...
+
+  /**
+   * Returns true if the input string is a valid IP address
+   */
+  @FunctionTemplate(
+    name = "is_valid_IP",
+    scope = FunctionTemplate.FunctionScope.SIMPLE,
+    nulls = FunctionTemplate.NullHandling.NULL_IF_NULL
+  )
+  public static class IsValidIPFunction implements DrillSimpleFunc {
```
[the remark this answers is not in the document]
cgivre: That was my original plan, however the commons validator has a dependency on commons-logging which is banned in Drill.  I wasn't sure why that's the case, but if we can fix that somehow, I can reinstate the code using commons-validator.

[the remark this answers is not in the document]
arina-ielchiieva: Why it's banned in Drill? Can we exclude it, for example?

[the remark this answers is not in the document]
cgivre: I'm not really sure why it's banned.  I'd love to exclude it, but I couldn't seem to get Drill to build when I tried putting that exclusion in the java-exec pom file.  Any assistance would be greatly appreciated.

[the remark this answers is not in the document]
arina-ielchiieva: Also you may ask on dev list regarding `commons-logging` usage on Drill dev list.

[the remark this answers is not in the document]
arina-ielchiieva: We exclude `commons-logging` because Drill uses `logback` for logging.
Excluding it from `commons-validator` should work perfectly fine (I have tried).

```
    <dependency>
      <groupId>commons-validator</groupId>
      <artifactId>commons-validator</artifactId>
      <version>1.4.0</version>
      <exclusions>
      <exclusion>
        <groupId>commons-logging</groupId>
        <artifactId>commons-logging</artifactId>
      </exclusion>
      </exclusions>
    </dependency>
```

cgivre:
Thanks!  I’ll resubmit this weekend using commons-validator

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
cgivre: Is there a preference to using an existing library?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
arina-ielchiieva: Well, it depends but I would call it's Java way to reuse already defined logic from different libraries rather then repeating it. Of course if it's worse adding new lib. In our case logic for `ntoa` , `aton` and `private_ip` does not seem trivial. Did you write it yourself or copied from somewhere else? If first case according to which references algorithm was written? Were all cases considered, tested etc? If second, is source is well-trusted? Does source owner mind his code will be reused etc? Maybe it's better to use well-known lib instead which mostly likely has all these functions tested and followed common standards?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
```diff
@@ -0,0 +1,619 @@
...
+    DrillBuf buffer;
+
+
+    public void setup() {
+    }
+
+
+    public void eval() {
+      String ipString = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(inputTextA.start, inputTextA.end, inputTextA.buffer);
+      if( ipString == null || ipString.isEmpty() || ipString.length() == 0 ){
```
arina-ielchiieva: `isEmpty()` is the same as `length() == 0`. Please remove excess check.
Please fix here and in other functions

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
```diff
@@ -0,0 +1,619 @@
...
+    DrillBuf buffer;
+
+
+    public void setup() {
+    }
+
+
+    public void eval() {
+      String ipString = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.toStringFromUTF8(inputIP.start, inputIP.end, inputIP.buffer);
+      if( ipString == null || ipString.isEmpty() || ipString.length() == 0 ){
```
arina-ielchiieva: @cgivre regarding coding convention. I have mentioned this multiple time in your previous pull request and in this as well. You do fix some parts but still there are many places where code conventions are not followed. I suggest you follow guidelines from the link below [1], import formatter and format code using your IDE. It's really tedious for the code reviewer to point to all places where simple formatting are not observed which can be done by the developer with simple keyboard shortcut but following code conventions a common rule here and we all try to follow it!

cgivre:
I applied the code formatting template from the Drill site in IntelliJ and resubmitted.   Regarding converting an IP to an int, there are tons of examples of this online and when I implemented it, I reviewed a few of them and adapted it to Drill.   I don't remember which one I used at this point, but here is a good explanation of it. 
https://www.mkyong.com/java/java-convert-ip-address-to-decimal-number/

Other examples:
https://stackoverflow.com/questions/12057853/how-to-convert-string-ip-numbers-to-integer-in-java

If you'd prefer, I can resubmit using a pre-existing library. 

For `is_private`, it really isn't that complicated.  Assuming the IP is valid to start with, there are three cases:
1.  If the IP address starts with 192.168 it is private
2.  If the IP starts with 172. and the second octet is between 16-31 (inclusive) it is private
3.  If the IP starts with 10. it is private. 

Anything else is a public IP.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/NetworkFunctions.java
```diff
@@ -0,0 +1,566 @@
...
+
+      String[] ipAddressInArray = ipString.split("\\.");
+
+      int result = 0;
+
+      int[] octets = new int[3];
+
+      for (int i = 0; i < 3; i++) {
+        octets[i] = Integer.parseInt(ipAddressInArray[i]);
+        if (octets[i] > 255 || octets[i] < 0) {
```
arina-ielchiieva: Why do we need this `if`?  It seems it can be removed since the subsequent if will take precedence.

cgivre: I put that in there to detect invalid IP addresses.  I guess we can remove it.

arina-ielchiieva:
+1, LGTM.

paul-rogers:
Please squash commits.

arina-ielchiieva:
@cgivre under squash commits @paul-rogers meant that you should end up with one commit (now you have 13). You might consider using rebase and then do force push in your branch to overwrite this 13 commits. Also please make sure your commit message has proper naming: DRILL-XXX: <message>. In your case, `DRILL-5834: Add networking functions`.

PR #972, opened 2017-10-04
DRILL-5838: Fix MaprDB filter pushdown for the case of nested field (reg. of DRILL-4264)

vvysotskyi:
Please see [DRILL-4264|https://issues.apache.org/jira/browse/DRILL-5838] for the problem description.

`FieldPath.asPathString()` returns string in which field names may be quoted if `FieldPath` instance was created from the string, in which these fields were quoted. `SchemaPath.toExpr()` returns string were all field names are quoted.

- Used `FieldPath.asPathString()` method to receive string representation of field path for passing it into MapRDB filter.
- Fixed existing MaprDB unit tests to handle this case.

### DIFF
--- a/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonConditionBuilder.java
+++ b/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonConditionBuilder.java
@@ -20,9 +20,9 @@ package org.apache.drill.exec.store.mapr.db.json;
 import org.apache.drill.common.expression.BooleanOperator;
 import org.apache.drill.common.expression.FunctionCall;
 import org.apache.drill.common.expression.LogicalExpression;
-import org.apache.drill.common.expression.SchemaPath;
 import org.apache.drill.common.expression.visitors.AbstractExprVisitor;
 import org.apache.drill.exec.store.hbase.DrillHBaseConstants;
+import org.apache.drill.exec.store.mapr.db.util.FieldPathHelper;
 import org.ojai.Value;
 import org.ojai.store.QueryCondition;
 import org.ojai.store.QueryCondition.Op;
@@ -159,7 +159,7 @@ public class JsonConditionBuilder extends AbstractExprVisitor<JsonScanSpec, Void
   private JsonScanSpec createJsonScanSpec(FunctionCall call,
       CompareFunctionsProcessor processor) {
     String functionName = processor.getFunctionName();
-    String fieldPath = processor.getPath().asPathString(false);
+    String fieldPath = FieldPathHelper.schemaPathToFieldPath(processor.getPath()).asPathString();
     Value fieldValue = processor.getValue();
 
     QueryCondition cond = null;
--- a/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonConditionBuilder.java
+++ b/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonConditionBuilder.java
@@ -159,73 +159,73 @@ public class JsonConditionBuilder extends AbstractExprVisitor<JsonScanSpec, Void
   private JsonScanSpec createJsonScanSpec(FunctionCall call,
       CompareFunctionsProcessor processor) {
     String functionName = processor.getFunctionName();
-    SchemaPath field = processor.getPath();
+    String fieldPath = processor.getPath().asPathString(false);
     Value fieldValue = processor.getValue();
 
     QueryCondition cond = null;
     switch (functionName) {
     case "equal":
       cond = MapRDB.newCondition();
-      setIsCondition(cond, field.getRootSegmentPath(), Op.EQUAL, fieldValue);
+      setIsCondition(cond, fieldPath, Op.EQUAL, fieldValue);
       cond.build();
       break;
 
     case "not_equal":
       cond = MapRDB.newCondition();
-      setIsCondition(cond, field.getRootSegmentPath(), Op.NOT_EQUAL, fieldValue);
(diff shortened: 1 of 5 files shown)


logical/src/main/java/org/apache/drill/common/expression/PathSegment.java
```diff
@@ -151,6 +151,15 @@ public boolean isNamed() {
       return true;
     }
 
+    /**
+     * Checks that the path of this name segment is complex.
+     *
+     * @return true if the path of this name segment contains dots
+     */
+    public boolean isComplex() {
```
paul-rogers: Not sure I agree with this. We just did changes to enforce the rule that a path segment is represented by a distinct object. If a path segment can contain dots, then we don't need to represent parts as multiple parts. But, we went the multiple part route to allow dots in names.

How will we know if a dot in a name represents a complex (multi-part) name vs. a simple name that happens to contain a dot?

In short, this seems a bit of a hack and introduces undesirable ambiguity.

logical/src/main/java/org/apache/drill/common/expression/SchemaPath.java
```diff
@@ -264,6 +264,56 @@ public String getRootSegmentPath() {
...
+      throw new IllegalStateException("Drill doesn't currently support top level arrays");
+    }
+    NameSegment nameSegment = seg.getNameSegment();
+    writeQuoted(sb, nameSegment.getPath(), quoted || nameSegment.isComplex());
+
+    while ((seg = seg.getChild()) != null) {
+      if (seg.isNamed()) {
+        nameSegment = seg.getNameSegment();
+        sb.append('.');
+        writeQuoted(sb, nameSegment.getPath(), quoted || nameSegment.isComplex());
```
paul-rogers: If `isComplex()` is meant to indicate that a name must be quoted, then `requiresQuotes()` would be a better name. Some other names that require quotes:

{noformat}
names with spaces
names-with-dashes
anything/with+an*operator
$looks$like$internal$name
maybeEvenCaseSensitive
MAYBEeVENcASEsENSIVIVE
&so!on
...
{noformat}

Basically, anything that is not a symbol (initial alpha followed by any number of alphanumeric...)

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonConditionBuilder.java
```diff
@@ -159,73 +159,73 @@ private void setIsCondition(QueryCondition c,
   private JsonScanSpec createJsonScanSpec(FunctionCall call,
       CompareFunctionsProcessor processor) {
     String functionName = processor.getFunctionName();
-    SchemaPath field = processor.getPath();
+    String fieldPath = processor.getPath().asPathString(false);
```
paul-rogers: This needs an explanation. This is a `String`. The method is `asPathString()` which looks like we will take name parts ["a", "b.c", "d"] to produce "a.b.c.d". This cannot be right as either 1) it is represents a path to be split, or 2) represents a full name to match. But, I might have fields ["a.b", "c.d"] which also expands to "a.b.c.d". So, the matching part can't be right. The path split part can't be right since we cannot recover the name parts from the string.

In short, should we path the parsed path (collection of name parts) to the functions?

See comments for more info.

paul-rogers:
I think we have an opportunity here to take as step back and think a bit about design rather than just applying code patches.

Drill must handle a large variety of names:

{noformat}
this.kind
this kind
$stranger+stuff#
{noformat}

Drill supports paths of names using Drill's SQL syntax:

*part1* . *part2* . *part3*

Drill's SQL rules require quoting if the name does not follow SQL naming syntax or would be confused with a SQL reserved word (something we should fix):

{noformat}
fine.`needs.quote`.`also-needs+quotes`
{noformat}

So, we have three problems when dealing with external systems:

* How do we convert from the naming system of the external tool into Drill?
* How do we represent the names within Drill unambiguously?
* How do we give Drill's names back to the external system?

Drill must handle a variety of systems; Drill's internal naming system can't bend with the wind to mean one thing for MySQL, another for Parquet and something else for MapR DB. The conversion must be done at the interface.

Our convention is (or should be):

* External names, or at least those that are not valid Drill symbols, must be quoted in SQL.
* Internally, each name segment is stored as a `NamePart` object, with a collection of parts forming a `SchemaPath`.
* When calling back out to the external system, the interface (plugin) must convert from Drill format back out to the external system format.

Now, let's consider this change. In Drill code, we call Drill methods to convert Drill names to MapR DB format. While this just might work, it does not really follow the above rules.

What we want is a `String drillToMapRDName(SchemaPath path)` method that applies DB-specific rules for conversion at the point of calls into DB. All names within Drill (unless cached only in the DB storage plugin) must be in Drill's internal form.

This point would not be that important if Drill worked with just one external system (DB in this case). But, as Drill evolves, it must work with a variety of systems. We should not rely on methods on Drill classes to provide the proper name format for external systems: external systems must provide those rules. This is true even if it happens to work out, for now, that Drill's current methods happen to work.

Just to drive the point home. Here we are changing a Drill method to produce names in a format that DB needs. But, might some other external system interface depend on the old format? How will this converge?

Bottom line:

Do the conversion in DB-specific code and will work fine.

vvysotskyi:
@paul-rogers, Thanks for the so informative comments, I completely agree with the convention that you have proposed. 
The problem which this PR is addressed, partially connected to the third point:

> When calling back out to the external system, the interface (plugin) must convert from Drill format back out to the external system format.

MaprDB uses `FieldPath` class to represent the field. Since the string, that is received in the method `JsonConditionBuilder.createJsonScanSpec()` from `SchemaPath` instance will be parsed using `org.ojai.FieldPath.parseFrom()` method, to avoid any ambiguousness, I converted `SchemaPath` to `FieldPath` and then called `FieldPath.asPathString()`.

PR #974, opened 2017-10-05
DRILL-5839: Handle Empty Batches in Merge Receiver

ppadma:


exec/java-exec/src/main/java/org/apache/drill/exec/store/mock/MockRecordReader.java
```diff
@@ -52,7 +52,7 @@ public MockRecordReader(FragmentContext context, MockScanEntry config) {
 
   private int getEstimatedRecordSize(MockColumn[] types) {
     int x = 0;
-    for (int i = 0; i < types.length; i++) {
+    for (int i = 0; i < (types == null ? 0 : types.length); i++) {
```
paul-rogers: Nit: maybe use:

```
if (types == null) { return 0; }
// Original code here..
```

ppadma:
@paul-rogers Thank you Paul. I made the change and pushed the new diffs.

PR #975, opened 2017-10-05
DRILL-5743: Handling column family and column scan for hbase

prasadns14:
This PR handles the scenario where the projected column list contains both a column family and a column within the same family.

@paul-rogers please review

### DIFF
--- a/contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
+++ b/contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
@@ -94,9 +94,21 @@ public class HBaseRecordReader extends AbstractRecordReader implements DrillHBas
     setColumns(projectedColumns);
   }
 
+  /**
+   * Provides the projected columns information to the Hbase Scan instance. If the
+   * projected columns list contains a column family and also a column in the
+   * column family, only the column family is passed to the Scan instance.
+   *
+   * For example, if the projection list is {cf1, cf1.col1, cf2.col1} then we only
+   * pass {cf1, cf2.col1} to the Scan instance.
+   *
+   * @param columns collection of projected columns
+   * @return collection of projected column family names
+   */
   @Override
   protected Collection<SchemaPath> transformColumns(Collection<SchemaPath> columns) {
     Set<SchemaPath> transformed = Sets.newLinkedHashSet();
+    Set<String> completeFamilies = Sets.newHashSet();
     rowKeyOnly = true;
     if (!isStarQuery()) {
       for (SchemaPath column : columns) {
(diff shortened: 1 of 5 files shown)


contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
```diff
@@ -97,6 +97,7 @@ public HBaseRecordReader(Connection connection, HBaseSubScan.HBaseSubScanSpec su
   @Override
   protected Collection<SchemaPath> transformColumns(Collection<SchemaPath> columns) {
     Set<SchemaPath> transformed = Sets.newLinkedHashSet();
+    Set<String> completeFamilies = Sets.newHashSet();
```
paul-rogers: Do we need to worry about case sensitivity here?

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
```diff
@@ -109,11 +110,14 @@ public HBaseRecordReader(Connection connection, HBaseSubScan.HBaseSubScanSpec su
...
-        } else {
-          hbaseScan.addFamily(family);
+        if (!completeFamilies.contains(new String(family))) {
+          if (child != null && child.isNamed()) {
+            byte[] qualifier = child.getNameSegment().getPath().getBytes();
+            hbaseScan.addColumn(family, qualifier);
+          } else {
+            hbaseScan.addFamily(family);
+            completeFamilies.add(new String(family));
+          }
```
paul-rogers: This code would greatly benefit from a comment to explain what's happening. Would suggest a Javadoc comment for the function explaining the transform rules.

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
```diff
@@ -109,11 +110,14 @@ public HBaseRecordReader(Connection connection, HBaseSubScan.HBaseSubScanSpec su
...
         transformed.add(SchemaPath.getSimplePath(root.getPath()));
         PathSegment child = root.getChild();
-        if (child != null && child.isNamed()) {
-          byte[] qualifier = child.getNameSegment().getPath().getBytes();
-          hbaseScan.addColumn(family, qualifier);
-        } else {
-          hbaseScan.addFamily(family);
+        if (!completeFamilies.contains(new String(family))) {
+          if (child != null && child.isNamed()) {
+            byte[] qualifier = child.getNameSegment().getPath().getBytes();
```
paul-rogers: This assumes UTF-8 encoding for the name. Can we be sure that HBase always uses UTF-8 for its encoding? Or, does HBase only support ASCII names so that we need only the ASCII subset of UTF-8? What happens if the user puts a non-ASCII character into the name in this case?

contrib/storage-hbase/src/test/java/org/apache/drill/hbase/TestTableGenerator.java
```diff
@@ -133,6 +133,43 @@ public static void generateHBaseDataset1(Connection conn, Admin admin, TableName
...
+    BufferedMutator table = conn.getBufferedMutator(tableName);
+
+    Put p = new Put("a1".getBytes());
+    p.addColumn("f".getBytes(), "c1".getBytes(), "21".getBytes());
+    p.addColumn("f".getBytes(), "c2".getBytes(), "22".getBytes());
+    p.addColumn("f".getBytes(), "c3".getBytes(), "23".getBytes());
+    table.mutate(p);
+
+    p = new Put("a2".getBytes());
+    p.addColumn("f".getBytes(), "c1".getBytes(), "11".getBytes());
```
paul-rogers: Here, we are deciding to encode names as UTF-8. Is this a standard? Or, is it our own convention? Could we have used some other encoding? If we do, how do we tell the code above what encoding we chose?

contrib/storage-hbase/src/main/java/org/apache/drill/exec/store/hbase/HBaseRecordReader.java
prasadns14: I observed that the planner takes care of it. It returns a single column family if there are more than one column family with same name but different case.
I still made the change to make it case insensitive.

contrib/storage-hbase/src/test/java/org/apache/drill/hbase/TestTableGenerator.java
prasadns14: We currently assume encoding to be UTF-8. Support for different encoding can be addressed through DRILL-5825.

prasadns14:
@paul-rogers please review

PR #976, opened 2017-10-06
DRILL-5797: Choose parquet reader from read columns

dprofeta:
ParquetRecordReader is not able to read complex columns. However it is
able to read simple columns in a file containing complex
columns. Instead of looking at the file to choose the reader, we
now choose which reader to use based on what columns is asked and if
they are simple or not.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
@@ -46,6 +46,7 @@ import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.hadoop.CodecFactory;
 import org.apache.parquet.hadoop.ParquetFileReader;
 import org.apache.parquet.hadoop.metadata.ParquetMetadata;
+import org.apache.parquet.io.InvalidRecordException;
 import org.apache.parquet.schema.GroupType;
 import org.apache.parquet.schema.MessageType;
 import org.apache.parquet.schema.Type;
@@ -161,6 +162,12 @@ public class ParquetScanBatchCreator implements BatchCreator<ParquetRowGroupScan
   }
 
   private static boolean isComplex(ParquetMetadata footer, List<SchemaPath> columns) {
+    /*
+    ParquetRecordReader is not able to read any nested columns and is not able to handle repeated columns.
+    It only handles flat column and optional column.
+    If it is a wildcard query, we check every columns in the metadata.
+    If not, we only check the projected columns.
+    */
     if (Utilities.isStarQuery(columns)) {
       MessageType schema = footer.getFileMetaData().getSchema();
 
@@ -186,15 +193,14 @@ public class ParquetScanBatchCreator implements BatchCreator<ParquetRowGroupScan
   }
 
   private static boolean isColumnComplex(GroupType grouptype, SchemaPath column) {
-    PathSegment.NameSegment root = column.getRootSegment();
-    if (!grouptype.containsField(root.getPath().toLowerCase())) {
+    try {
+      Type type = grouptype.getType(column.getRootSegment().getPath().toLowerCase());
+      return type.isRepetition(Type.Repetition.REPEATED) || !type.isPrimitive();
+    }
+    catch(InvalidRecordException e) {
+      //if the type does exist, we will fill with null so it is a simple type.
       return false;
     }
-    Type type = grouptype.getType(root.getPath().toLowerCase());
-    if (type.isRepetition(Type.Repetition.REPEATED) || !type.isPrimitive()) {
-      return true;
-    }
-    return false;
   }
 
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
@@ -162,12 +162,16 @@ public class ParquetScanBatchCreator implements BatchCreator<ParquetRowGroupScan
   }
 
   private static boolean isComplex(ParquetMetadata footer, List<SchemaPath> columns) {
-    /*
-    ParquetRecordReader is not able to read any nested columns and is not able to handle repeated columns.
-    It only handles flat column and optional column.
-    If it is a wildcard query, we check every columns in the metadata.
-    If not, we only check the projected columns.
-    */
+    /**
+     * ParquetRecordReader is not able to read any nested columns and is not able to handle repeated columns.
+     * It only handles flat column and optional column.
+     * If it is a wildcard query, we check every columns in the metadata.
+     * If not, we only check the projected columns.
+     * We only check the first level columns because :
+     *   - if we need a.b, it means a is a complex type, no need to check b as we don't handle complex type.
+     *   - if we need a[10], a is repeated, ie its repetiton level is greater than 0
+     *   - if we need a, it is at the first level of the schema.
+     */
     MessageType schema = footer.getFileMetaData().getSchema();
     if (Utilities.isStarQuery(columns)) {
       for (Type type : schema.getFields()) {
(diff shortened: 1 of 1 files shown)


dprofeta:
@sachouche Can you review it?

sachouche:
Sure!


Regards,

Salim

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
```diff
@@ -156,18 +160,39 @@ public ScanBatch getBatch(FragmentContext context, ParquetRowGroupScan rowGroupS
     return new ScanBatch(rowGroupScan, context, oContext, readers, implicitColumns);
   }
 
-  private static boolean isComplex(ParquetMetadata footer) {
-    MessageType schema = footer.getFileMetaData().getSchema();
+  private static boolean isComplex(ParquetMetadata footer, List<SchemaPath> columns) {
+    if (Utilities.isStarQuery(columns)) {
```
paul-rogers: Perhaps a comment with some explanation? If wildcard query, we query all columns, so check if any of them are complex. If project list, then check only the projected columns.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
```diff
@@ -156,18 +160,39 @@ public ScanBatch getBatch(FragmentContext context, ParquetRowGroupScan rowGroupS
...
+    }
+  }
+
+  private static boolean isColumnComplex(GroupType grouptype, SchemaPath column) {
+    PathSegment.NameSegment root = column.getRootSegment();
+    if (!grouptype.containsField(root.getPath().toLowerCase())) {
+      return false;
+    }
+    Type type = grouptype.getType(root.getPath().toLowerCase());
+    if (type.isRepetition(Type.Repetition.REPEATED) || !type.isPrimitive()) {
```
paul-rogers: `return type.isRepetition(Type.Repetition.REPEATED) || !type.isPrimitive();`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
```diff
@@ -156,18 +160,39 @@ public ScanBatch getBatch(FragmentContext context, ParquetRowGroupScan rowGroupS
...
+      return false;
+    }
+  }
+
+  private static boolean isColumnComplex(GroupType grouptype, SchemaPath column) {
+    PathSegment.NameSegment root = column.getRootSegment();
+    if (!grouptype.containsField(root.getPath().toLowerCase())) {
+      return false;
+    }
+    Type type = grouptype.getType(root.getPath().toLowerCase());
```
paul-rogers: What are the semantics of `getType()`? Does it return null if there is no such type? If so, then we can retrieve the type and check if it is null. Otherwise, if it throws an exception, perhaps we can catch that. Either way, we avoid two searches for the same column and two conversions of the path to lower case.

Note also that a recent change deprecated `getPath()`. The preferred form is `getName()` so it is clear that we are getting a single name part. If the Parquet column is nested (a.b.c, say), then we have to explicitly handle the multiple name parts as Drill does support names with dots and one cannot simply concatenate or split a path to get a string. That is, `["a.b", "c"]` is two fields, `["a", "b", "c"]` is three, but both are represented as a full path as "a.b.c", creating an ambiguity.

dprofeta: ok, for `getType()`. It throws an exception so I will catch it.
I don't see any `getName()` in the `SchemaPath` / `PathSegment` class. Can you tell me which `getName()` you mean?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
dprofeta: Yes, sure. I wanted to check it in a loop first, but ParquetRecordReader doesn't handle any nested type, so the loop is not needed now. But I didn't refactor enough.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetScanBatchCreator.java
dprofeta: Added in a new commit.

dprofeta:
here is the updated PR.
Yes, I also wanted to add group without repetition. It is only a matter of naming so it should not be hard but when I tested, the fast reader was not able to handle it.

dprofeta:
I updated the javadoc with Paul remarks.

priteshm:
@sachouche can you please take a final look? If it looks good, maybe one of the committers can include this for the 1.12 release. @arina-ielchiieva ?

sachouche:
+1
looks good!

paul-rogers:
@dprofeta, tried to commit this PR, but ran into multiple functional test failures:

```
Execution Failures:
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex12.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex8.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex56.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex274.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex7.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex57.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex102.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex5.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex10.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex9.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex203.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex101.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex275.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex6.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex205.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex11.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex58.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex153.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex202.q
/root/drillAutomation/mapr/framework/resources/Functional/complex/parquet/complex151.q
```

The common failure stack trace seems to be:

```
    org.apache.drill.exec.store.parquet.columnreaders.ParquetRecordReader.handleException():272
    org.apache.drill.exec.store.parquet.columnreaders.ParquetRecordReader.setup():256
    org.apache.drill.exec.physical.impl.ScanBatch.getNextReaderIfHas():241
    org.apache.drill.exec.physical.impl.ScanBatch.next():167
...
```

sachouche:
Looking at the stack trace:
- The code definitely is initializing a column of type REPEATABLE
- The Fast Reader didn't expect this scenario so it used a default container (NullableVarBinary) for VL binary DT

Why this is happening?
- The code in ReadState::buildReader() is processing all selected columns
- This information is obtained from the ParquetSchema
- Looking at the code, this seems a case-sensitivity issue
- The ParquetSchema is case-insensitive whereas the Parquet GroupType is not
- Damien added a catch handler (column not found) to handle use-cases where we are projecting non-existing columns
- This basically is leading to an unforeseen use-case
- Assume column XYZ is complex
- User uses an alias (xyz)
- The new code will allow this column to pass and treat is as simple
- The ParquetSchema is being case insensitive will process this column
- and thus the exception in the test suite

Suggested Fix
- Create a map (key to-lower-case) and register all current row-group columns
- Use this map to locate a selected column type

paul-rogers:
Drill follows SQL rules and is case insensitive. If case sensitivity has snuck in somewhere (perhaps due to the use of `equals()` rather than `equalsIgnorCase()` or the use of a case-sensitive map), then we should fix that.

Note also that column aliases should not be visible to the Parquet reader.

priteshm:
@dprofeta will you be able to address the issues before the release?

arina-ielchiieva:
Closing this PR, since new one was opened - https://github.com/apache/drill/pull/1368.

PR #977, opened 2017-10-06
DRILL-5849: Add freemarker lib to dependencyManagement to ensure prop…

arina-ielchiieva:
…er version is used when resolving dependency version conflicts

Details in [DRILL-5849](https://issues.apache.org/jira/browse/DRILL-5849).

### DIFF
--- a/exec/java-exec/pom.xml
+++ b/exec/java-exec/pom.xml
@@ -224,7 +224,6 @@
     <dependency>
       <groupId>org.freemarker</groupId>
       <artifactId>freemarker</artifactId>
-      <version>${freemarker.version}</version>
     </dependency>
     <dependency>
       <groupId>net.hydromatic</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -678,6 +678,11 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
+        <groupId>org.freemarker</groupId>
+        <artifactId>freemarker</artifactId>
+        <version>${freemarker.version}</version>
+      </dependency>
+      <dependency>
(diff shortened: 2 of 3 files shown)


vrozov:
The proposed changes simplify the dependency management, but do not change the dependency resolution, AFAIK.

arina-ielchiieva:
@vrozov as far as I understand to manage dependency resolution, you can either exclude conflicting dependencies or use dependency management mechanism [1, 2]. I have used dependency management mechanism since it's more convenient and you don't have to track new conflicting dependencies. If you know it's not sufficient, please let us know what is the best solution.

[1] https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html
[2] http://techidiocy.com/maven-dependency-version-conflict-problem-and-resolution/

vrozov:
@arina-ielchiieva I refer to java-exec and fmpp modules dependencies resolution. Both AFAIK should not be affected by the change, so I suppose it is another module that has a transitive dependency on freemarker through java-exec and another 3rd party library that has the issue described in the JIRA. Please confirm and provide details on the module and the 3rd party library.

arina-ielchiieva:
@vrozov 
1. Change the the main pom.xml where I have added freemarker lib to the `dependencyManagement` section is the core change of this pull request. With this change dependency resolution is always done in favor of the freemarker version we expect.
2. Changes in other two pom files are just clean up. Since as we now have freemarker lib in `dependencyManagement` there is no need to indicate version there.
3. Regarding which modules do have transitive dependency on freemarker, right now there are two modules  - java-exec (jersey-mvc-freemarker lib) and fmpp (fmpp lib).

PR #978, opened 2017-10-06
DRILL-5842: Refactor and simplify the fragment, operator contexts for testing

paul-rogers:
Drill's execution engine has a "fragment context" that provides state for a fragment as a whole, and an "operator context" which provides state for a single operator. Historically, these have both been concrete classes that make generous references to the Drillbit context, and hence need a full Drill server in order to operate.

Drill has historically made extensive use of system-level testing: build the entire server and fire queries at it to test each component. Over time, we are augmenting that approach with unit tests: the ability to test each operator (or parts of an operator) in isolation.

Since each operator requires access to both the operator and fragment context, the fact that the contexts depend on the overall server creates a large barrier to unit testing. An earlier checkin started down the path of defining the contexts as interfaces that can have different run-time and test-time implementations to enable testing.

One solution has been to use JMockit or Mockito to mock up the operator and fragment contexts. This works, but is fragile. (Indeed, those tests failed spectacularly during the work for this PR until the mocks were updated with the new and changed methods.)

This PR refactors those interfaces: simplifying the operator context and introducing an interface for the fragment context. New code will use these new interfaces, while older code continues to use the concrete implementations. Over time, as operators are enhanced, they can be modified to allow unit-level testing.

### Context Revisions

The following changes were made to contexts:

* Rename `AbstractOperatorExecContext` to `BaseOperatorContext`.
* Introduce `BaseFragmentContext` as a common base class for test and production fragment contexts.
* Introduce `FragmentContextInterface` (sorry for the name) as the new abstraction to be (eventually) used by all operators so that they can use both test and production versions.
* Move methods from the concrete implementations to the base classes where they do not depend on Drillbit or network services.
* Retire earlier versions: `OperExecContext` and `OperExecContextImpl`.
* Add more services to `OperatorContext`, the interface for the operator context.

A result of these changes is that `OperatorContext` is now a one-stop-shop for services needed by operators. It now provides:

* Access to the fragment context: `getFragmentContext()`
* Access to the physical operator definition (AKA “physical operator”): `getOperatorDefn()`.

Because of these changes, we need no longer pass around the triple of fragment context, operator definition and operator context — something needed by the “managed” sort and upcoming PRs.

The above caused changes to consumers of the above classes. The funky `FragmentContextInterface` name was chosen to minimize such changes: code that uses the concrete `FragmentContext` did not change. (Renaming `FragmentContext` to `FragmentContextImpl` would have caused hundreds of changes…)

### Operator Fixture

The `OperatorFixture` class provides test-time implementations of the fragment and operator contexts. These classes were modified to extend the new base classes described above, and to include the new method on the context interfaces. This is not a final design, but it is a step toward the final design.

### Mocks

Added selected new methods to the JMockit mocks set up in `PhysicalOpUnitTestBase`. This is the crux of the problem that this change works to solve: mocks are not very stable and are very hard to debug. Having a test-time implementation is a better long-term solution. However, we are not yet in a a position to replace the mocks with the test-time alternatives; doing so will cause a cascade of other changes that would be too big for this one PR.

### Other Clean-up

Continued to shift code to use the write-only operator stats interface to allow easier operator context mapping. Changes here focused on the Drill file system to allow mocking up tests with scan batch.

Various minor code cleanups are also included.


priteshm:
@sohami  can you please review this?

jinfengni:
I understand the motivation of this PR is to refactor the various of "context" used in drill, so that we could get rid of JMock in test. I'm a bit confused after reading the statement that "we are not yet in a a position to replace the mocks with the test-time alternatives". I thought that's the purpose of this PR, yet it does not demonstrate the benefit, after going through big chunk of code change ?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/unit/MiniPlanUnitTestBase.java
```diff
@@ -360,7 +366,7 @@ public T columnsToRead(String ... columnsToRead) {
    */
   public class JsonScanBuilder extends ScanPopBuider<JsonScanBuilder> {
     List<String> jsonBatches = null;
-    List<String> inputPaths = Collections.EMPTY_LIST;
+    List<String> inputPaths = Collections.emptyList();
```
jinfengni: why replace a field with a function call, which eventually return that static field?

paul-rogers:
@jinfengni, sorry my description was a bit ambiguous. This PR has two goals:

1. Clean up the contexts based on what has been learned recently.
2. Evolve the contexts to allow "sub-operator" unit testing without mocks.

Drill has many tests that use mocks and these are unaffected by the change. The PR simply allows new tests to do sub-operator testing without mocks, when convenient.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/unit/MiniPlanUnitTestBase.java
paul-rogers: This is a subtle point. Using the constant creates the expression:

```
public static final List EMPTY_LIST = new EmptyList<>(); // Definition
List<String> inputPaths = EMPTY_LIST; // Original code
```

The above is not type-friendly: we are setting a typed list (`inputPaths`) to an untyped constant (`EMPTY_LIST`).

The revised code uses Java's parameterized methods to work around the type ambiguity:

```
public static final <T> List<T> emptyList() ... // Definition
List<String> inputPaths = Collections.emptyList(); // Type-safe assignment
```

Functionally, the two expressions are identical. But, the original was type-unsafe and generated a compiler warning. The new one is type-safe and resolves the warning.

sohami:
Looking into the PR and based on description it says below:

_A result of these changes is that OperatorContext is now a one-stop-shop for services needed by operators. It now provides:_

- Access to the fragment context: getFragmentContext()
- Access to the physical operator definition (AKA “physical operator”): getOperatorDefn().

_Because of these changes, we need no longer pass around the triple of fragment context, operator definition and operator context — something needed by the “managed” sort and upcoming PRs._

Based on my understanding it looks like existing interface and implementation `OperExecContextImpl` do provide access to `FragmentContext` and `PopConfig` (operator Defn). The instance of this context is created inside `ExternalSortBatch` which has access to FragmentContext and PopConfig already. Also `SortImpl` only takes in `OperExecContext`. I am little confused how the change avoided the need to pass around the triplet ?

paul-rogers:
@sohami, pushed a commit that addresses your PR comments.

Regarding the "one stop shop" comment. This PR removes the `OperExecContextImpl` class. That class was an earlier attempt to combine the three items into one. Additional experience showed that the existing operator context could do that task instead.

This PR did not change existing operators to avoid passing around multiple items. Instead, it allows new code (for the batch size limitation project) to pass just the operator context, and use that to obtain the other two items. The external sort had to be changed because it used the old `OperExecContextImpl` for that purpose, and so the code had to be updated when `OperExecContextImpl` was removed.

paul-rogers:
Rebased and squashed commits.

sohami:
+1 LGTM. 
Thanks for clarification.

arina-ielchiieva:
+1, LGTM.

PR #980, opened 2017-10-09
DRILL-5857: Fix NumberFormatException in Hive unit tests

vvysotskyi:
There is no unit test since with or without this change tests are passes and query plan does not change. 

The exception that has been appeared caught and wrote into the logs.
After the check on [this line](https://github.com/apache/drill/blob/3e8b01d5b0d3013e3811913f0fd6028b22c1ac3f/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveMetadataProvider.java#L90), correct stats was created calling [getStatsEstimateFromInputSplits()](https://github.com/apache/drill/blob/3e8b01d5b0d3013e3811913f0fd6028b22c1ac3f/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveMetadataProvider.java#L95) method.


arina-ielchiieva:
+1, LGTM.

PR #981, opened 2017-10-09
DRILL-5854: IllegalStateException when empty batch with valid schema …

ppadma:
…is received

Problem is that merge receiver is reading from the wrong sender when first batch is empty from one of the senders.
When first batch is empty, we are continuing without moving to the next sender.


priteshm:
@paul-rogers could you please review this one?

chunhui-shi:
+1

PR #982, opened 2017-10-09
DRILL-5859: Fixed queued time calculation

prasadns14:
@paul-rogers please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
@@ -455,9 +455,10 @@ public class Foreman implements Runnable {
           .build(logger);
     } catch (QueryQueueException e) {
       throw new ForemanSetupException(e.getMessage(), e);
+    } finally {
+      queryManager.markQueueWaitEndTime();
     }
     moveToState(QueryState.STARTING, null);
-    queryManager.markQueueWaitEndTime();
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/DistributedQueryQueue.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/DistributedQueryQueue.java
@@ -140,7 +140,7 @@ public class DistributedQueryQueue implements QueryQueue {
 
   private static class ConfigSet {
     private final long queueThreshold;
-    private final long queueTimeout;
+    private final int queueTimeout;
     private final int largeQueueSize;
     private final int smallQueueSize;
     private final double largeToSmallRatio;
@@ -149,7 +149,7 @@ public class DistributedQueryQueue implements QueryQueue {
 
     public ConfigSet(SystemOptionManager optionManager) {
       queueThreshold = optionManager.getOption(ExecConstants.QUEUE_THRESHOLD_SIZE);
-      queueTimeout = optionManager.getOption(ExecConstants.QUEUE_TIMEOUT);
+      queueTimeout = (int) optionManager.getOption(ExecConstants.QUEUE_TIMEOUT);
 
       // Option manager supports only long values, but we do not expect
       // more than 2 billion active queries, so queue size is stored as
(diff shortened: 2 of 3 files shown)


paul-rogers:
Thanks for the first-time commit! We like to have the JIRA number in the title; please go ahead and ecit the title so it looks like this:

DRILL-1234: Fixed queued time calculation

Use the actual Drill JIRA ticket number in place of "1234".

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/EmbeddedQueryQueue.java
```diff
@@ -93,7 +93,7 @@ public void release() {
     public String queueName() { return "local-queue"; }
   }
 
-  private final int queueTimeoutMs;
+  private final long queueTimeoutMs;
```
paul-rogers: It is not necessary for the timeout to be a long. Timeouts should be on the order of minutes, not decades, so an int is perfectly fine and represents the semantics just a bit better.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/QueryQueue.java
```diff
@@ -61,9 +61,9 @@
 
     private final QueryId queryId;
     private final String queueName;
-    private final int timeoutMs;
+    private final long timeoutMs;
```
paul-rogers: Same comment as above.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/DistributedQueryQueue.java
```diff
@@ -273,9 +273,8 @@ public QueueLease enqueue(QueryId queryId, double cost) throws QueryQueueExcepti
     }
 
     if (lease == null) {
-      int timeoutSecs = (int) Math.round(configSet.queueTimeout/1000.0);
-      logger.warn("Queue timeout: {} after {} seconds.", queueName, timeoutSecs);
-      throw new QueueTimeoutException(queryId, queueName, timeoutSecs);
+      logger.warn("Queue timeout: {} after {} ms.", queueName, configSet.queueTimeout);
```
paul-rogers: This was actually not wrong. Instead, it mapped the number from the internal ms units into more user-friendly units of seconds. That is, it is more useful to see:

Queue timeout: small after 60 seconds

Than:

Queue timeout: small after 60000 ms.

If we do want to use ms. perhaps format them with commas:

Queue timeout: small after 60,000 ms.

Or show both:

Queue timeout: small after 60,000 ms. (60 seconds)

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -455,9 +455,10 @@ private void admit(QueryWorkUnit work) throws ForemanSetupException {
           .build(logger);
     } catch (QueryQueueException e) {
       throw new ForemanSetupException(e.getMessage(), e);
+    } finally {
+      queryManager.markQueueWaitEndTime();
```
paul-rogers: Good!

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/EmbeddedQueryQueue.java
prasadns14: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/QueryQueue.java
prasadns14: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/DistributedQueryQueue.java
prasadns14: I made the change to be consistent with QueryTimeoutException which logs the time in ms.
I made the changes as suggested in both the places.

Queue timeout: small after 60,000 ms. (60 seconds)

prasadns14:
@paul-rogers Made the changes as suggested. please review

prasadns14:
Update the title.

PR #983, opened 2017-10-10
DRILL-5819: Default value of security.admin.user_groups and security.admin.users is true

bitblender:
The values for admin user/groups in the config file was incorrectly set to "true". 

These options have no meaningful static default values. At runtime, the process user/groups should be used as the default admin user/groups. New accessors have been added for these options and these have to be used instead of the usual option manager accessors. Dummy default strings are used in the config file, which when returned by the usual options manager, indicate that the user has not changed these options. 

This change implements the above-mentioned items and also adds a UI section to see the current admin user/groups. This UI section is only shown to authenticated admin users.


paul-rogers:
MD? Please remove that number from the PR title; we need only the Drill tickets here.

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
```diff
@@ -115,4 +124,59 @@ private void setOptHelper() throws Exception {
...
+    FixtureBuilder builder = ClusterFixture.builder();
+
+    try (ClusterFixture cluster = builder.build();
+         ClientFixture client = cluster.clientFixture()) {
+      OptionManager optionManager = cluster.drillbit().getContext().getOptionManager();
+
+      // Admin Users Tests
+      // config file should have the 'fake' default admin user and it should be returned
+      // by the option manager if the option has not been set by the user
+      String configAdminUser =  optionManager.getOption(ExecConstants.ADMIN_USERS_VALIDATOR);;
```
paul-rogers: one semicolon should be enough...

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
```diff
@@ -115,4 +124,59 @@ private void setOptHelper() throws Exception {
...
+      assertNotEquals(adminUser1, ExecConstants.ADMIN_USERS_VALIDATOR.DEFAULT_ADMIN_USERS);
+
+      // Change TEST_ADMIN_USER if necessary
+      String TEST_ADMIN_USER = "ronswanson";
+      if (adminUser1.equals(TEST_ADMIN_USER)) {
+        TEST_ADMIN_USER += "thefirst";
+      }
+      // Check if the admin option accessor honors a user-supplied values
+      String sql = String.format("ALTER SYSTEM SET `%s`='%s'", ExecConstants.ADMIN_USERS_KEY, TEST_ADMIN_USER);
+      client.queryBuilder().sql(sql).run();
```
paul-rogers: `client.alterSystem(key, value)`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -75,6 +80,29 @@ public ClusterInfo getClusterInfoJSON() {
...
+    // separate groups by comma + space
+    if (adminUsers != null) {
+      String[] groups = adminUsers.split(",");
+      adminUsers = DrillStringUtils.join(groups, ", ");
+    }
+
+    // separate groups by comma + space
+    if (adminUserGroups != null) {
+      String[] groups = adminUserGroups.split(",");
+      adminUserGroups = DrillStringUtils.join(groups, ", ");
```
paul-rogers: What if the user provided the list with spaces: "a, b,  c"? Actually, in the above, it is fine; HTML will compress the run of spaces to a single one for display...

But, what about where we do the real check? Do the unit tests cover this case?

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
```diff
@@ -115,4 +124,59 @@ private void setOptHelper() throws Exception {
       test(String.format("ALTER SYSTEM SET `%s` = %d;", ExecConstants.SLICE_TARGET, ExecConstants.SLICE_TARGET_DEFAULT));
     }
   }
+
+  @Test
+  public void testAdminUserOptions() throws Exception {
+    FixtureBuilder builder = ClusterFixture.builder();
+
+    try (ClusterFixture cluster = builder.build();
```
paul-rogers: No big deal, but if you set no options, then a shorthand is:

```
cluster = ClusterFixture.standardCluster();
```

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
```diff
@@ -115,4 +124,59 @@ private void setOptHelper() throws Exception {
...
+      assertEquals(configAdminUserGroups, ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.DEFAULT_ADMIN_USER_GROUPS);
+
+      // Option accessor should never return the 'fake' default from the config
+      String adminUserGroups1 = ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.getAdminUserGroups(optionManager);
+      assertNotEquals(adminUserGroups1, ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.DEFAULT_ADMIN_USER_GROUPS);
+
+      // Change TEST_ADMIN_USER_GROUPS if necessary
+      String TEST_ADMIN_USER_GROUPS = "yakshavers";
+      if (adminUserGroups1.equals(TEST_ADMIN_USER_GROUPS)) {
+        TEST_ADMIN_USER_GROUPS += ",wormracers";
```
paul-rogers: Usually ALL_CAPS is reserved for constants, which means we can't alter them as we do here... I suspect you mean `testAdminUserGroups`...

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
```diff
@@ -115,4 +124,59 @@ private void setOptHelper() throws Exception {
...
+      }
+      // Check if the admin option accessor honors a user-supplied values
+      sql = String.format("ALTER SYSTEM SET `%s`='%s'", ExecConstants.ADMIN_USER_GROUPS_KEY, TEST_ADMIN_USER_GROUPS);
+      client.queryBuilder().sql(sql).run();
+      String adminUserGroups2 = ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.getAdminUserGroups(optionManager);
+      assertEquals(adminUserGroups2, TEST_ADMIN_USER_GROUPS);
```
paul-rogers: Should the tests include actually checking if the user and/or group matches? This will check for problems such as spaces before/after the names, etc.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
bitblender: I pushing changes that handle ill-formatted user input

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
bitblender: I am pushing changes which add these tests.

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
bitblender: Done. It was a constant before I changed the code to handle the corner case.

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
bitblender: Done.

exec/java-exec/src/test/java/org/apache/drill/exec/server/TestOptionsAuthEnabled.java
bitblender: Done

common/src/main/java/org/apache/drill/common/util/DrillStringUtils.java
```diff
@@ -1,203 +1,258 @@
...
+   * where each individual string is separated by the provided separator.
+   * Copied from org.apache.commons.lang3.StringUtils (to avoid
+   * bringing in the entire package)
+   * @param array - array holding the strings to be concatenated
+   * @param separator - separator to be used in the concatenation
+   * @return The separator separated concatenated string
+   */
+
+  public static String join(Object[] array, String separator) {
+    return array == null?null:join(array, separator, 0, array.length);
```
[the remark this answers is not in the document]
bitblender: The original file has CRLF line termination. Hence the full diff.

exec/java-exec/src/test/java/org/apache/drill/test/ClientFixture.java
```diff
@@ -128,6 +128,21 @@ public void alterSystem(String key, Object value) {
   }
 
   /**
+   * Reset a system option
+   * @param key
+   */
+
+  public void resetSystem(String key) {
```
paul-rogers: These are added in PR #970, in [this file](https://github.com/apache/drill/pull/970/files#diff-e2de7c62cedd56fbabbf3cff0f0663a2).

But, this PR is likely to be committed before this one, so I'll deal with the merge conflict later.

PR #985, opened 2017-10-11
DRILL-5862 Update project parent pom xml to the latest ASF version

vrozov:


pom.xml
```diff
@@ -15,7 +15,8 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>14</version>
+    <version>18</version>
+    <relativePath/>
```
arina-ielchiieva: Do we need this element present in pom.xml?

vrozov: Yes, to make maven happy. Otherwise, it expects to find parent pom in a parent directory and if finds one (for example if there is an uber pom that builds multiple projects), maven will complain.

arina-ielchiieva: But it was happy before :) plus I don't see this element present in pom.xml of other Apache projects (Calcite, Camel). Does Drill structure differ?

vrozov: Try to add an empty pom to the parent of your drill repo :). I guess most of the projects assume that there will be no pom.xml file in a parent directory and it is not always the case, so it is better to follow maven guidelines (http://maven.apache.org/ref/3.0/maven-model/maven.html)

arina-ielchiieva: I have tried, it does fail ) Thanks for explanation.

arina-ielchiieva:
+1, LGTM.

pom.xml
```diff
@@ -15,7 +15,8 @@
   <parent>
     <groupId>org.apache</groupId>
     <artifactId>apache</artifactId>
-    <version>14</version>
+    <version>18</version>
```
paul-rogers: Never noticed this before. Which version is this? Why does Apache itself have a version? Can we include a comment about this version and the relative path for us newbies?

vrozov: @paul-rogers This is the latest version that majority of Apache projects use (if I remember correctly it was released in May 2016,  version 14 is quite old). Apache parent pom is a maven artifact that defines plugins and other common to ASF project maven parameters and it needs to be versioned. I'd suggest keeping pom file as is, the discussion from the PR is mirrored in the DRILL-5862 and the commit has reference to it.

PR #987, opened 2017-10-12
DRILL-5863: Sortable table incorrectly sorts fragments/time lexically

kkhatua:
The DataTables jQuery library sorts data based on the value of the element in a column.
However, since Drill publishes sortable items like fragment IDs and time durations as non-numeric text, the sorting is incorrect. 
This PR fixes the fragment and duration ordering based on their implicit numeric values (minor ID and millisecond representation, respectively).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
@@ -19,7 +19,9 @@ package org.apache.drill.exec.server.rest.profile;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import org.apache.drill.exec.proto.UserBitShared.MajorFragmentProfile;
 import org.apache.drill.exec.proto.UserBitShared.MinorFragmentProfile;
@@ -228,6 +230,8 @@ public class FragmentWrapper {
         Collections2.filter(major.getMinorFragmentProfileList(), Filters.missingOperatorsOrTimes));
 
     Collections.sort(complete, Comparators.minorId);
+
+    Map<String, String> attributeMap = new HashMap<String, String>(); //Reusing for different fragments
     for (final MinorFragmentProfile minor : complete) {
       final ArrayList<OperatorProfile> ops = new ArrayList<>(minor.getOperatorProfileList());
 
@@ -244,7 +248,8 @@ public class FragmentWrapper {
         biggestBatches = Math.max(biggestBatches, batches);
       }
 
-      builder.appendCell(new OperatorPathBuilder().setMajor(major).setMinor(minor).build());
+      attributeMap.put("data-order", String.valueOf(minor.getMinorFragmentId())); //Overwrite values from previous fragments
+      builder.appendCell(new OperatorPathBuilder().setMajor(major).setMinor(minor).build(), attributeMap);
       builder.appendCell(minor.getEndpoint().getAddress());
       builder.appendMillis(minor.getStartTime() - start);
       builder.appendMillis(minor.getEndTime() - start);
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -76,12 +77,14 @@ public String getId() {
   public String getContent() {
     TableBuilder builder = new TableBuilder(OPERATOR_COLUMNS, OPERATOR_COLUMNS_TOOLTIP, true);
 
+    Map<String, String> attributeMap = new HashMap<String, String>(); //Reusing for different fragments
     for (ImmutablePair<ImmutablePair<OperatorProfile, Integer>, String> ip : opsAndHosts) {
       int minor = ip.getLeft().getRight();
       OperatorProfile op = ip.getLeft().getLeft();
 
+      attributeMap.put("data-order", String.valueOf(minor)); //Overwrite values from previous fragments
```
paul-rogers: Not clear on how this works. If we are sorting on a string value, then the values must be zero padded so that 02 comes before 11. Otherwise, order will be 1, 10, 11, 2, 3.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -76,12 +77,14 @@ public String getId() {
...
       String path = new OperatorPathBuilder().setMajor(major).setMinor(minor).setOperator(op).build();
-      builder.appendCell(path);
+      builder.appendCell(path, attributeMap);
```
paul-rogers: Does `appendCell` copy the map contents? If not, if it simply holds onto the map, then all cells hold the same map, which is probably not what you want.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/FragmentWrapper.java
```diff
@@ -244,7 +248,8 @@ public String getContent() {
         biggestBatches = Math.max(biggestBatches, batches);
       }
 
-      builder.appendCell(new OperatorPathBuilder().setMajor(major).setMinor(minor).build());
+      attributeMap.put("data-order", String.valueOf(minor.getMinorFragmentId())); //Overwrite values from previous fragments
```
paul-rogers: See note below.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
kkhatua: The appendCell() method has been overloaded with to allow for a map, from which any set of attributes can be injected. 
https://github.com/kkhatua/drill/blob/783ca992f3bbb52ae606a8ca10745d6dd2ed0d4f/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/TableBuilder.java#L95
This allows for any future libraries that might need to inject their own set of attributes to be incorporated without the need for any significant modification to the TableBuilder.
So, each appendCell is provided its own unique map that is consumed during the construction of the `<td>` element

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
kkhatua: You are correct that the values must be zero padded. However, the padding by default is only up till achieving a width of two digits. The sequencing didn't get messed up as long as the number of minor fragments for each major fragment was less than 100.
For greater than 100, the sequencing broke the way you described it.
So, until this fix, you'll get the minor fragment ordered as: 
01-01-XX ..... 01-10-XX, 01-100-XX, ....
instead of 
01-01-XX ..... 01-10-XX, 01-11-XX, .... 01-10-XX, 01-11-XX, .... 01-100-XX, 01-101-XX

What we are doing is injecting the <td> element with an attribute that carries the actual numeric value. The dataTable library will discover and sort the elements of a column by the value of this attribute instead of the value that the element encapsulates.

paul-rogers:
Thanks for the explanations.
+1

PR #990, opened 2017-10-12
DRILL-5872: Workaround for invalid cost in physical plans

paul-rogers:
See DRILL-5872 for details.

Works around a bug in a storage plugin that produces NaN for a cost estimate, which then leads to profiles that can't be deserialized. This fix simply replaces the NaN in the profile with the maximum double value.

The real fix is that the storage plugin concerned should not produce NaN estimates.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/PhysicalPlan.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/PhysicalPlan.java
@@ -94,6 +94,14 @@ public class PhysicalPlan {
     for (final PhysicalOperator ops : getSortedOperators()) {
       totalCost += ops.getCost();
     }
+
+    // As it turns out, sometimes the total cost can be out of range.
+    // This throws off other code, so clamp the cost at the maximum
+    // double value.
+
+    if (Double.isNaN(totalCost)) {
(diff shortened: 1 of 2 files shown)


priteshm:
@amansinha100  or @arina-ielchiieva  can you please review this?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/PhysicalPlan.java
```diff
@@ -94,6 +94,14 @@ public double totalCost() {
     for (final PhysicalOperator ops : getSortedOperators()) {
       totalCost += ops.getCost();
     }
+
+    // As it turns out, sometimes the total cost can be out of range.
+    // This throws off other code, so clamp the cost at the maximum
+    // double value.
+
+    if (Double.isNaN(totalCost)) {
+      totalCost = Double.MAX_VALUE;
```
arina-ielchiieva: May be better to replace with 0 or negative number, like -1 (like the second option better). Won't be max double value confusing in this case? It's just nan is not a number and replace it with double max value.

arina-ielchiieva:
Is there a Jira for storage plugin fix?

paul-rogers:
Fixed in planner, so closing this.

PR #992, opened 2017-10-13
DRILL-5873: (C++ Client) Improve SASL error reporting.

parthchandra:
@sohami what do you think? I overrode the SASL_NOMECH error message. The message from the SASL library is very unfriendly.

### DIFF
--- a/contrib/native/client/src/clientlib/drillClientImpl.cpp
+++ b/contrib/native/client/src/clientlib/drillClientImpl.cpp
@@ -657,8 +657,11 @@ connectionStatus_t DrillClientImpl::handleAuthentication(const DrillUserProperti
         // Check the negotiated SSF value and change the handlers.
         if(m_encryptionCtxt.isEncryptionReqd()) {
             if(SASL_OK != m_saslAuthenticator->verifyAndUpdateSaslProps()) {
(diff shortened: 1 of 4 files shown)


superbstreak:
@parthchandra Thanks for adding this!

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -678,7 +678,10 @@ connectionStatus_t DrillClientImpl::handleAuthentication(const DrillUserProperti
         m_io_service.reset();
         return CONN_SUCCESS;
     } else {
-        logMsg << m_encryptionCtxt << ", Error: " << m_saslResultCode;
+        logMsg << m_encryptionCtxt
+               << ", Mechanism: " << m_saslAuthenticator->getAuthMechanismName()
+               << ", Error: " << m_saslResultCode
+               << ", Cause: " << m_saslAuthenticator->getErrorMessage(m_saslResultCode);
```
sohami: How about adding this cause message string in the [if condition](https://github.com/apache/drill/pull/992/files#diff-8e6df071d8ca863fcfa578892944c1dcR661) too ?

sohami:
+1 LGTM. Thanks a lot for the changes.

PR #993, opened 2017-10-13
DRILL-5874: NPE in AnonWebUserConnection.cleanupSession()

sohami:


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRestServer.java
```diff
@@ -204,9 +217,15 @@ public WebUserConnection provide() {
                 config.getLong(ExecConstants.HTTP_SESSION_MEMORY_RESERVATION),
                 config.getLong(ExecConstants.HTTP_SESSION_MEMORY_MAXIMUM));
 
+        // Create a dummy close future which is needed by Foreman only. Foreman uses this future to add a close
+        // listener to known about channel close event from underlying layer. We use this future to notify Foreman
+        // listeners when the Web connection between Web Client and WebServer is closed. This will help Foreman to cancel
+        // all the running queries for this Web Client.
```
paul-rogers: Classic Web communication is connectionless and stateless. For performance, some browsers use a keep-alive mechanism. But, this must be entirely optional. In this way, an HTTP connection is completely unlike an RPC connection.

Given that the HTTP connection will be immediately closed after submitting a query, do we really want to kill the query in response?

Or, are we using a web session concept that "closes" in response to a session timeout?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRestServer.java
```diff
@@ -204,9 +217,15 @@ public WebUserConnection provide() {
...
+        final ChannelPromise closeFuture = new DefaultChannelPromise(null, executor);
+
         // Create a WebSessionResource instance which owns the lifecycle of all the session resources.
-        // Set this instance as an attribute of HttpSession, since it will be used until session is destroyed.
-        webSessionResources = new WebSessionResources(sessionAllocator, remoteAddress, drillUserSession);
+        // Set this instance as an attribute of HttpSession, since it will be used until session is destroyed
+        webSessionResources = new WebSessionResources(sessionAllocator, remoteAddress, drillUserSession, closeFuture);
```
paul-rogers: The code suggests we are using a *session*, not *connection* as mentioned in the earlier comment.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRestServer.java
```diff
@@ -260,8 +282,14 @@ public WebUserConnection provide() {
         logger.trace("Failed to get the remote address of the http session request", ex);
       }
 
-      final WebSessionResources webSessionResources = new WebSessionResources(sessionAllocator,
-              remoteAddress, drillUserSession);
+      // Create a dummy close future which is needed by Foreman only. Foreman uses this future to add a close
+      // listener to known about channel close event from underlying layer. We use this future to notify Foreman
+      // listeners when the Web connection between Web Client and WebServer is closed. This will help Foreman to cancel
```
paul-rogers: Again: connection or session?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebSessionResources.java
```diff
@@ -68,16 +68,19 @@ public SocketAddress getRemoteAddress() {
...
-
     try {
       AutoCloseables.close(webUserSession, allocator);
     } catch (Exception ex) {
       logger.error("Failure while closing the session resources", ex);
     }
 
-    // Set the close future associated with this session.
+    // Notify all the listeners of this closeFuture for failure events so that listeners can do cleanup related to this
+    // WebSession. This will be called after every query execution by AnonymousWebUserConnection::cleanupSession and
```
paul-rogers: If a session is anonymous, is the session tied to a cookie (i.e. a browser?) If not, then does the session end on the close of the connection? If so, would we kill a query as soon as the web connection that submits the query is closed?

We might need a general comment about how this works (if one does not already exist.)

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -124,6 +124,13 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
     }
   }
 
+  /**
+   * Returns a DefaultChannelPromise which doesn't have reference to any actual channel but has an EventExecutor
+   * associated with it. In this case we use EventExecutor out of BitServer EventLoopGroup. Since there is no actual
+   * connection established using this class, hence the close event will never be fired by underlying layer and close
+   * future is set only when the WebSessionResources are closed.
```
paul-rogers: Can we take this comment up a level? This is a description of the mechanics, which is fine. But, what is the intent?

Are we using a web session as a pseudo-connection (as described in earlier comments)? Is this an adapter to make the session look like a connection?

exec/java-exec/src/test/java/org/apache/drill/exec/server/rest/WebSessionResourcesTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+  private WebSessionResources webSessionResources;
+
+  private boolean listenerComplete;
+
+  private CountDownLatch latch;
+
+  private EventExecutor executor;
+
+
```
paul-rogers: Maybe a comment to explain what this test is, in fact, testing? If the mechanism itself is explained elsewhere, simply reference that with a \@see or \@link tag.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRestServer.java
sohami: The query is killed when the HttpSession is invalidated either by logout or by timeout in case of authenticated user session. But for unauthenticated or anonymous user we don't kill the query after HttpSession timeout since we don't maintain any session related information. The query will keep running unless it is explicitly cancelled from /profile page.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRestServer.java
sohami: Will fix the comment. Sorry for confusion.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRestServer.java
sohami: Will fix the comment

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebSessionResources.java
sohami: From my understanding session is tied to a cookie not a connection. In case of anonymous session, we call close after completion of each query. But for authenticated session the close will be called when HttpSession is invalidated which is during logout or session timeout.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
sohami: The intent of putting this comment here was just as an FYI how the underlying CloseFuture is created which is returned by this function and stressting the fact that it doesn't have any actual physical connection associated with it.

WebUserConnection acts like an adapter between HttpSession and connection which Foreman expects with each submitted request. For authenticated user this adapter is used to get UserSession object which stores all the updated Drill session options changed over the time. 

For anonymous user this adapter is created fresh with each request. But as discussed in person I will make a change so that we can maintain session for anonymous user and have similar behavior like authenticated user. And when the HttpSession for unauthenticated user time's out then it will do clean up of all the web resources related to that session.

exec/java-exec/src/test/java/org/apache/drill/exec/server/rest/WebSessionResourcesTest.java
sohami: Sure will add a comment here. The purpose of this test was to validate that with Null channel in closeFuture which is used by WebSessionResources, during `WebSessionResources:close` it works as expected.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
sohami: Based on previous conversation the assumption was unsecure path also maintains cookie or session. On looking more this proved to be false and we don't maintain any session for unsecure case. This means that the requests submitted without authentication are stateless and once the connection is broken by closing the browser session then the query will still continue running until explicitly cancelled from profile page. I didn't found any way in Jetty to get notification about connection close event. And looks like that won't be much helpful right now since we send the response back only after the query is completed and we have entire result. I have opened another JIRA [DRILL-5897](https://issues.apache.org/jira/browse/DRILL-5897) to see if we can improve the way we send result back to client or make unsecure case to also use session concept such that we can cancel the in-flight queries based on when session is invalidated.

sohami:
Updated the PR with review comments and opened a JIRA based on in-person discussion.

PR #996, opened 2017-10-16
DRILL-5878: TableNotFound exception is being reported for a wrong sto…

HanumathRao:
…rage plugin.

@paul-rogers  @chunhui-shi Please review these changes. These changes are for reporting a meaningful error in case of wrong storage plugin is specified in the query. Currently Drill reports TableNotFound exception for
1) if no storage plugin exists.
2) if no table with the tablename exists.
These changes report SchemaNotFound exception for case 1) and TableNotFound exception for 2).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
@@ -77,6 +77,22 @@ public class SchemaUtilites {
     return findSchema(defaultSchema, schemaPathAsList);
   }
 
+  /**
+   * Utility function to get the commonPrefix schema between two supplied schemas.
+   * @param defaultSchema default schema
+   * @param schemaPath current schema path
+   * @param isCaseSensitive true if caseSensitive comparision is required.
+   * @return common prefix schemaPath
+   */
+  public static String getPrefixSchemaPath(final String defaultSchema, final String schemaPath, final boolean isCaseSensitive) {
+    if (!isCaseSensitive) {
+      return Strings.commonPrefix(defaultSchema.toLowerCase(), schemaPath.toLowerCase());
+    }
+    else {
+      return Strings.commonPrefix(defaultSchema, schemaPath);
+    }
+  }
+
   /** Utility method to search for schema path starting from the given <i>schema</i> reference */
   private static SchemaPlus searchSchemaTree(SchemaPlus schema, final List<String> schemaPath) {
     for (String schemaName : schemaPath) {
@@ -121,6 +137,9 @@ public class SchemaUtilites {
     return SCHEMA_PATH_JOINER.join(schemaPath);
   }
 
+  /** Utility method to get the schema path from one or more strings. */
+  public static String getSchemaPath(String s1, String s2, String... rest) { return SCHEMA_PATH_JOINER.join(s1,s2,rest); }
+
   /** Utility method to get the schema path as list for given schema instance. */
   public static List<String> getSchemaPathAsList(SchemaPlus schema) {
     if (isRootSchema(schema)) {
@@ -146,6 +165,16 @@ public class SchemaUtilites {
             givenSchemaPath)
         .addContext("Current default schema: ",
             isRootSchema(defaultSchema) ? "No default schema selected" : getSchemaPath(defaultSchema))
+        .build(logger);
+  }
+
+  /** Utility method to throw {@link UserException} with context information */
+  public static void throwSchemaNotFoundException(final String defaultSchema, final String givenSchemaPath) {
+    throw UserException.validationError()
+        .message("Schema [%s] is not valid with respect to either root schema or current default schema.",
+            givenSchemaPath)
+        .addContext("Current default schema: ",
+            defaultSchema.equals("") ? "No default schema selected" : defaultSchema)
         .build(logger);
   }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
@@ -109,7 +109,7 @@ public class SchemaUtilites {
    * @return true if the given <i>schema</i> is root schema. False otherwise.
    */
   public static boolean isRootSchema(SchemaPlus schema) {
-    return schema.getParentSchema() == null;
+    return schema == null || schema.getParentSchema() == null;
   }
 
   /**
@@ -136,9 +136,6 @@ public class SchemaUtilites {
   public static String getSchemaPath(List<String> schemaPath) {
     return SCHEMA_PATH_JOINER.join(schemaPath);
   }
-
-  /** Utility method to get the schema path from one or more strings. */
-  public static String getSchemaPath(String s1, String s2, String... rest) { return SCHEMA_PATH_JOINER.join(s1,s2,rest); }
 
   /** Utility method to get the schema path as list for given schema instance. */
   public static List<String> getSchemaPathAsList(SchemaPlus schema) {
@@ -169,13 +166,13 @@ public class SchemaUtilites {
   }
 
   /** Utility method to throw {@link UserException} with context information */
-  public static void throwSchemaNotFoundException(final String defaultSchema, final String givenSchemaPath) {
+  public static void throwSchemaNotFoundException(final SchemaPlus defaultSchema, final List<String> givenSchemaPath) {
     throw UserException.validationError()
-        .message("Schema [%s] is not valid with respect to either root schema or current default schema.",
-            givenSchemaPath)
-        .addContext("Current default schema: ",
-            defaultSchema.equals("") ? "No default schema selected" : defaultSchema)
-        .build(logger);
+            .message("Schema [%s] is not valid with respect to either root schema or current default schema.",
+                    givenSchemaPath)
+            .addContext("Current default schema: ",
+                    isRootSchema(defaultSchema) ? "No default schema selected" : getSchemaPath(defaultSchema))
+            .build(logger);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -481,6 +485,19 @@ public RelOptTableImpl getTable(final List<String> names) {
             .message("Temporary tables usage is disallowed. Used temporary table name: %s.", names)
             .build(logger);
       }
+
+      // Check the schema and throw a valid SchemaNotFound exception instead of TableNotFound exception.
```
arina-ielchiieva: Does it mean that Calcite instead of returning schema not found exception returns table not found exception?
Per my understanding this PR customizes Drill but what if we go different path and enhance Calcite (or may be this is already done in newer Calcite versions)?

HanumathRao: Thank you for the review. I agree that this should ideally be handled at Calcite layer. I also think that even after Calcite providing this functionality there should be some customization that needs to be done as we understand the context better than Calcite. Once the calcite fixes this issue then we can always change the code accordingly.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
```diff
@@ -149,6 +168,16 @@ public static void throwSchemaNotFoundException(final SchemaPlus defaultSchema,
         .build(logger);
   }
 
+  /** Utility method to throw {@link UserException} with context information */
+  public static void throwSchemaNotFoundException(final String defaultSchema, final String givenSchemaPath) {
+    throw UserException.validationError()
+        .message("Schema [%s] is not valid with respect to either root schema or current default schema.",
```
paul-rogers: Because names can contain dots, we must more careful format the schema. The schema should consists of a list (array) of name parts. For each name:

* If the name contains a dot, wrap it in back ticks: `` `a.b` ``
* Otherwise, use the name as is: `c`
* Concatenate the name parts with dots: `` `a.b`.c ``

Code to do this might already exist. @vvysotskyi may know.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
```diff
@@ -77,6 +77,22 @@ public static SchemaPlus findSchema(final SchemaPlus defaultSchema, final String
...
   }
 
+  /**
+   * Utility function to get the commonPrefix schema between two supplied schemas.
+   * @param defaultSchema default schema
+   * @param schemaPath current schema path
+   * @param isCaseSensitive true if caseSensitive comparision is required.
+   * @return common prefix schemaPath
+   */
+  public static String getPrefixSchemaPath(final String defaultSchema, final String schemaPath, final boolean isCaseSensitive) {
```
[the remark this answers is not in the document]
HanumathRao: @paul-rogers  Thanks for the review. For this API I am assuming the caller will supply the information whether it needs to be a case sensitive or case insensitive comparison. The caller for this function is passing the parserConfig case sensitive parameter.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
```diff
@@ -77,6 +77,22 @@ public static SchemaPlus findSchema(final SchemaPlus defaultSchema, final String
...
+    if (!isCaseSensitive) {
+      return Strings.commonPrefix(defaultSchema.toLowerCase(), schemaPath.toLowerCase());
```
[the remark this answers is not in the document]
HanumathRao: @paul-rogers  Thank you for this catch. I did try out some examples which were throwing schemanotfound exception. I have changed the code to handle it. getDefaultSchemaPath of session object returns the defaultSchema as String, hence I converted the SchemaPath to String by SchemaPathJOINER. As pointed out by your comment, this will loose some context information if the "." is part of the name. However, I also think that there is a bug in the defaultSchema code path as it is also loosing this information by converting it to String. For now I am not fixing that bug (which I am planning to fix in another JIRA) as I am not sure   about the changes in setting defaultSchema.

Coming on to the specific valid use case which was throwing schemaNotFound exception is
in dfs I have introduced "tmp.tmp1" in place of tmp.

And if I query select * from dfs.`tmp.tmp1`.`employee.json` limit 1; this query reports SchemaNotFound exception even though a workspace with name tmp.tmp1 exists.

This is currently working with new changes.

arina-ielchiieva:
@paul-rogers thanks for approving pull request but actually before merging these changes in Drill (since I am not still convinced we need them), I was expecting @HanumathRao to do investigation if newer Calcite versions handle this case, if yes, maybe postpone this changes till the Calcite upgrade or cherry-pick them in Drill Calcite (if urgent). If not, create Jira in Calcite project and provide fix, and maybe cherry-pick this very fix once done (again if urgent). Since if all these checks can be done in Calcite we don't need to customize Drill.

arina-ielchiieva:
Well, with the short-term solution I am afraid that eventually we'll forget to revert it when Calcite checks take over and we'll be checking for schema twice.

@HanumathRao before going any further with this pull request, could you please investigate the following:
1. Does newer Calcite versions handle this case?
2. If not, can this be handled in Calcite? If yes, please create Jira in Calcite. If possible create fix and check if Calcite community will accept it.

HanumathRao:
@arina-ielchiieva Thank your for the comments. There is some work that went into calcite to handle meaningful error messages. This is the checkin that has those changes.
https://github.com/apache/calcite/commit/5f9c019080c7231acaf3df80732d915351051d93#diff-0c11f3f4d738e3fa55968eb19f1c8050

It reports following errors when a table cannot be resolved.
{code}
select empid from "hr".emps;
Object 'EMPS' not found within 'hr'; did you mean 'emps'?
!error
{code}

However, I think the error logic should be customized to particular software(in this case DRILL) so as to report semantically meaningful error messages. Drill knows more about the context and hence can provide more customized error messages to the end user.

arina-ielchiieva:
So does it mean that schema validation will be done twice: first in Drill and then in Calcite? Will it influence query parsing performance?

HanumathRao:
@arina-ielchiieva I think this check shouldn't cause much of a performance impact as it is in parser code and also it is checked right now by the DRILL's custom overload of getTable. If this function throws an exception then it is caught and reported to the user. I think there can be improvement to not check this by every code path(i.e valid code path as well) and only check when super.getTable returns null, in this way we can report error only when we couldn't find a table. Please let me know if this approach seems reasonable, so that I can go ahead and change the code accordingly.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
arina-ielchiieva: Could you please factor out this logic in a separate method?

exec/java-exec/src/test/java/org/apache/drill/exec/store/dfs/TestFileSelection.java
```diff
@@ -63,4 +63,17 @@ public void testEmptyFolderThrowsTableNotFound() throws Exception {
...
+    final String query = String.format("select * from dfs1.`%s`", table);
+    try {
+      testNoResult(query);
+    } catch (Exception ex) {
+      final String pattern = String.format("[[dfs1]] is not valid with respect to either root schema or current default schema").toLowerCase();
+      final boolean isSchemaNotFound = ex.getMessage().toLowerCase().contains(pattern);
+      assertTrue(isSchemaNotFound);
+      throw ex;
+    }
+  }
```
arina-ielchiieva: Can you please add test case for the incorrect workspace, 
a. `select * from dfs.incorrect_wk.table;`
b. 
```
use dfs; 
select * from incorrect_wk.table;
```
I assume it will return incorrect schema exception as well?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
```diff
@@ -77,6 +77,22 @@ public static SchemaPlus findSchema(final SchemaPlus defaultSchema, final String
     return findSchema(defaultSchema, schemaPathAsList);
   }
 
+  /**
+   * Utility function to get the commonPrefix schema between two supplied schemas.
```
arina-ielchiieva: Could you please add example in Java doc?

exec/java-exec/src/test/java/org/apache/drill/exec/store/dfs/TestFileSelection.java
```diff
@@ -63,4 +63,17 @@ public void testEmptyFolderThrowsTableNotFound() throws Exception {
     }
   }
```
arina-ielchiieva: Maybe it makes sense to move new unit tests and `testEmptyFolderThrowsTableNotFound` into separate class which will test behavior when incorrect object is defined (i.e. schema, workspace, table)?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SchemaUtilites.java
HanumathRao: done.

HanumathRao:
@arina-ielchiieva Thank you for the review comments. I have modified the code accordingly. Please let me know if anything needs to be changed.

exec/java-exec/src/test/java/org/apache/drill/exec/store/dfs/TestFileSelection.java
HanumathRao: done

arina-ielchiieva:
@HanumathRao thanks for making the changes. 
+1, LGTM.

amansinha100:
Merged in 7a2fc87ee20f706d85cb5c90cc441e6b44b71592.  @HanumathRao  pls close the PR.

arina-ielchiieva:
@HanumathRao please close the pull request.

PR #997, opened 2017-10-17
DRILL-5882: C++ Client: [Threat Modeling] Drillbit may be spoofed by …

bitblender:
…an attacker and this may lead to data being written to the attacker's target instead of Drillbit


contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -595,6 +611,12 @@ connectionStatus_t DrillClientImpl::validateHandshake(DrillUserProperties* prope
 
     switch(this->m_handshakeStatus) {
         case exec::user::SUCCESS:
+            // Check if client needs auth/encryption and server is not requiring it
+            if(clientNeedsAuthentication(properties) || clientNeedsEncryption(properties)) {
```
parthchandra: Generally, all error messages come from errmsgs.cpp so we can localize them when we need to.

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -595,6 +611,12 @@ connectionStatus_t DrillClientImpl::validateHandshake(DrillUserProperties* prope
 
     switch(this->m_handshakeStatus) {
         case exec::user::SUCCESS:
+            // Check if client needs auth/encryption and server is not requiring it
```
parthchandra: Not too clear about the SASL flow, but I assume that if the server returning SUCCESS is sufficient to assume that there is no auth required by the server.

contrib/native/client/src/clientlib/saslAuthenticatorImpl.cpp
```diff
@@ -145,6 +145,8 @@ int SaslAuthenticatorImpl::init(const std::vector<std::string>& mechanisms, exec
             authMechanismToUse = value;
         }
     }
+    // clientNeedsAuth cannot be false if the code above picks an authMechanism
```
sohami: clientNeedsAuth --> clientNeedsAuthentication

contrib/native/client/src/clientlib/drillClientImpl.cpp
```diff
@@ -518,6 +518,22 @@ bool DrillClientImpl::clientNeedsEncryption(const DrillUserProperties* userPrope
...
+ * If the USERPROP_PASSWORD or USERPROP_AUTH_MECHANISM connection string properties are set,
+ * then it is implied that the client wants authentication.
+ */
+bool DrillClientImpl::clientNeedsAuthentication(const DrillUserProperties* userProperties) {
+    bool needsAuthentication = false;
+    if(!userProperties) {
+        return false;
+    }
+    needsAuthentication = userProperties->isPropSet(USERPROP_PASSWORD) ||
+        userProperties->isPropSet(USERPROP_AUTH_MECHANISM);
```
sohami: I think we should also check if the `password & auth parameter` value is not empty string.

contrib/native/client/src/clientlib/drillClientImpl.cpp
parthchandra: Also, do you need two different messages : auth required and encr required ?

bitblender: - Externalized the messages to errmsgs.cpp 
- Changed the error message to "Client needs a secure connection but server does not support... " to account for the case where auth and/or enc is required by the client but missing on the server

contrib/native/client/src/clientlib/drillClientImpl.cpp
bitblender: Yes. The control flow goes through the AUTH_REQUIRED case when the server requires auth.

parthchandra:
+1.

PR #998, opened 2017-10-18
DRILL-5887: Display process user/groups info in Drill UI

prasadns14:
Display process user and process user groups in Drill UI

@paul-rogers please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
@@ -38,6 +38,7 @@ import org.apache.drill.exec.server.DrillbitContext;
 import org.apache.drill.exec.server.rest.DrillRestServer.UserAuthEnabled;
 import org.apache.drill.exec.server.rest.auth.AuthDynamicFeature;
 import org.apache.drill.exec.server.rest.auth.DrillUserPrincipal;
+import org.apache.drill.exec.util.ImpersonationUtil;
 import org.apache.drill.exec.work.WorkManager;
 import org.apache.drill.exec.work.foreman.rm.DistributedQueryQueue;
 import org.apache.drill.exec.work.foreman.rm.DistributedQueryQueue.ZKQueueInfo;
@@ -85,22 +86,10 @@ public class DrillRoot {
     // For all other cases the user info need-not or should-not be displayed
     OptionManager optionManager = work.getContext().getOptionManager();
     final boolean isUserLoggedIn = AuthDynamicFeature.isUserLoggedIn(sc);
-    String adminUsers = isUserLoggedIn ?
-            ExecConstants.ADMIN_USERS_VALIDATOR.getAdminUsers(optionManager) : null;
-    String adminUserGroups = isUserLoggedIn ?
-            ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.getAdminUserGroups(optionManager) : null;
-
-    // separate groups by comma + space
-    if (adminUsers != null) {
-      String[] groups = adminUsers.split(",");
-      adminUsers = Joiner.on(", ").join(groups);
-    }
-
-    // separate groups by comma + space
-    if (adminUserGroups != null) {
-      String[] groups = adminUserGroups.split(",");
-      adminUserGroups = Joiner.on(", ").join(groups);
-    }
+    final String processUser = ImpersonationUtil.getProcessUserName();
+    final String processUserGroups = Joiner.on(", ").join(ImpersonationUtil.getProcessUserGroupNames());
+    String adminUsers = ExecConstants.ADMIN_USERS_VALIDATOR.getAdminUsers(optionManager);
+    String adminUserGroups = ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.getAdminUserGroups(optionManager);
 
     final boolean shouldShowUserInfo = isUserLoggedIn &&
             ((DrillUserPrincipal)sc.getUserPrincipal()).isAdminUser();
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -85,21 +86,33 @@ public ClusterInfo getClusterInfoJSON() {
     // For all other cases the user info need-not or should-not be displayed
     OptionManager optionManager = work.getContext().getOptionManager();
     final boolean isUserLoggedIn = AuthDynamicFeature.isUserLoggedIn(sc);
+    final String processUser = isUserLoggedIn ?
+            ImpersonationUtil.getProcessUserName() : null;
```
paul-rogers: Do we want to display the process user only if the user is logged in? Here, are we assuming logged in means authenticated? Why not display the process user otherwise?

Rather than null, should we just use a tag such as "<process user>"?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -85,21 +86,33 @@ public ClusterInfo getClusterInfoJSON() {
...
+    final String processUserGroups = isUserLoggedIn ?
+            Joiner.on(",").join(ImpersonationUtil.getProcessUserGroupNames()) : null;
```
paul-rogers: Are we guaranteed that the list contains no spaces around the commas? (When reading from ZK, the string is user entered, and the string might be "fred   ,   bob". Do we have to worry about that here?) If the string might contain extra spaces, we might want to use code that Karthik provided to split, strip spaces, and combine the names.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -85,21 +86,33 @@ public ClusterInfo getClusterInfoJSON() {
...
             ExecConstants.ADMIN_USERS_VALIDATOR.getAdminUsers(optionManager) : null;
     String adminUserGroups = isUserLoggedIn ?
             ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.getAdminUserGroups(optionManager) : null;
 
     // separate groups by comma + space
     if (adminUsers != null) {
-      String[] groups = adminUsers.split(",");
-      adminUsers = Joiner.on(", ").join(groups);
+      if (adminUsers.length() == 0) {
+        adminUsers = "<empty>";
```
paul-rogers: This string uses angle brackets, which have meaning to HTML. When we emit these strings into the template, do we call the Freemarker function to escape special HTML characters?

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -112,6 +112,14 @@
                         <td>Admin User Groups</td>
                         <td class="list-value">${model.getAdminUserGroups()}</td>
                       </tr>
+                      <tr>
+                        <td>Process User</td>
+                        <td class="list-value">${model.getProcessUser()}</td>
```
paul-rogers: Please look at the generated HTML to ensure that the angle brackets in "<none>" are escaped. Should happen [by default](http://freemarker.org/docs/dgui_misc_autoescaping.html) if we've set up Freemarker correctly.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
prasadns14: From security perspective it would be good to display process user only if an admin user is logged in. Removed the check here as the condition is handled in html.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
prasadns14: This list is obtained from hadoop.security.UserGroupInformation and is not the user option. hence we need not worry about the space

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
prasadns14: yes freemarker handles this

exec/java-exec/src/main/resources/rest/index.ftl
prasadns14: yes freemarker handles this

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -85,19 +86,24 @@ public ClusterInfo getClusterInfoJSON() {
...
-    String adminUserGroups = isUserLoggedIn ?
-            ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.getAdminUserGroups(optionManager) : null;
+    final String processUser = ImpersonationUtil.getProcessUserName();
+    final String processUserGroups = Joiner.on(", ").join(ImpersonationUtil.getProcessUserGroupNames());
+    String adminUsers = ExecConstants.ADMIN_USERS_VALIDATOR.getAdminUsers(optionManager);
+    String adminUserGroups = ExecConstants.ADMIN_USER_GROUPS_VALIDATOR.getAdminUserGroups(optionManager);
 
     // separate groups by comma + space
-    if (adminUsers != null) {
+    if (adminUsers.length() == 0) {
```
arina-ielchiieva: Frankly saying, I would suggest to do all these manipulations in freemarker. There are many built-in freemarker functions that can help (i.e. split, join, check for emptiness etc [1]).

[1] http://freemarker.org/docs/ref_builtins_string.html#ref_builtin_split

prasadns14: @arina-ielchiieva Thanks for the suggestion.
Any particular reason to move these manipulations to freemarker? Do we have any advantage moving this if block over to freemarker?

arina-ielchiieva: This would be adherence to the MVC pattern which separates model from the view. Model is generated on the server side while freemarker is responsible for representation layer.

prasadns14: Made the changes. @arina-ielchiieva please review

arina-ielchiieva:
+1, LGTM.

PR #999, opened 2017-10-18
DRILL-5881:Java Client: [Threat Modeling] Drillbit may be spoofed by …

sohami:
…an attacker and this may lead to data being written to the attacker's target instead of Drillbit


sohami:
@parthchandra - Please help to review this PR. I have added new unit tests for the change and made sure all the existing test's are also passing.

parthchandra:
+1. 
You need a newline at the end :)

parthchandra:
+1 to the second commit as well. This addresses the problem of the unit tests failing intermittently.

PR #1000, opened 2017-10-18
DRILL-5845: Columns returned by select with "ORDER BY" and "LIMIT" cl…

vdiravka:
…auses are not in correct order.


arina-ielchiieva:
+1, LGTM.

PR #1002, opened 2017-10-18
DRILL-5888: Remove dependency of SSLConfig on hadoop.security package…

parthchandra:
…. This allows jdbc-all-jar to be built without hadoop dependencies.

The original  code used Hadoop's SSLFactory.Mode enum. The same enum is now replicated in 
the SSLConfig class. 

@arina-ielchiieva  please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserClient.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserClient.java
@@ -128,7 +128,7 @@ public class UserClient
     this.allocator = allocator;
     this.supportComplexTypes = supportComplexTypes;
     try {
-      this.sslConfig = new SSLConfigBuilder().properties(properties).mode(SSLFactory.Mode.CLIENT)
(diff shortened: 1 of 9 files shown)


parthchandra:
Yes. The property is consumed by SSLConfig which uses that to construct a key name that is looked up in the hadoop config.

parthchandra:
Thanks Arina. I removed the unused imports.

arina-ielchiieva:
One more thing, I have noticed that `TestSSLConfig#testInvalidHadoopKeystore()` contains the following line 
```
String hadoopSSLFileProp = MessageFormat
         .format(HADOOP_SSL_CONF_TPL_KEY, SSLFactory.Mode.SERVER.toString().toLowerCase());		         .format(HADOOP_SSL_CONF_TPL_KEY, SSLFactory.Mode.SERVER.toString().toLowerCase());
```
Do we need to replace `SSLFactory.Mode.SERVER.toString()` here as well?

parthchandra:
Yes, I will remove it. Thanks !

arina-ielchiieva:
+1, LGTM.

PR #1003, opened 2017-10-19
DRILL-5893: Reverted the number of forked test processes back to 2

ilooner:
Reverting the number of forked test process back to a safe number, since the current default I added in  DRILL-5752 was too aggressive for build machines with a large number of cores. If you want to speed up the build you can do:

```
mvn clean install -DforkCount=1C
```


ilooner:
@arina-ielchiieva

arina-ielchiieva:
+1, LGTM.

PR #1004, opened 2017-10-20
DRILL-5876: Use openssl profile to include netty-tcnative dependency with the platform specific classifier

vrozov:


parthchandra:
+1 . Thanks Vlad. Maven awesomeness !

PR #1007, opened 2017-10-23
DRILL-5864 Selecting a non-existing field from a MapR-DB JSON table fai…

HanumathRao:
…ls with NPE

@amansinha100 @adityakishore 
These changes are about having a special handling in case if the query contains no fields that are present in the maprdb json table. I have looked at other readers which are having this special handling. Here are the readers which do the special handling now.
JSONRecordReader and ParquetRecordReader.

With out these changes the upstream operator are getting not populated vectors and hence they throw nullpointer exception.

### DIFF
--- a/contrib/format-maprdb/README.md
+++ b/contrib/format-maprdb/README.md
@@ -1,2 +1,8 @@
 drill-mapr-plugin
 =================
+By default all the tests in contrib/format-maprdb are disabled.
+To enable and run these tests please use -Pmapr profile to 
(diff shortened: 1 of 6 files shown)


adityakishore:
+1

amansinha100:
Merged in 125a9271d7cf0dfb30aac8e62447507ea0a7d6c9.  @HanumathRao pls close the PR (for some reason I don't have permission).

PR #1012, opened 2017-10-27
DRILL-5911: Upgrade esri-geometry-api version to 2.0.0 to avoid depen…

vvysotskyi:
…dency on org.json library

Please see [DRILL-5911](https://issues.apache.org/jira/browse/DRILL-5911) for details.

### DIFF
--- a/contrib/gis/pom.xml
+++ b/contrib/gis/pom.xml
@@ -40,7 +40,7 @@
 		<dependency>
 			<groupId>com.esri.geometry</groupId>
 			<artifactId>esri-geometry-api</artifactId>
-			<version>1.2.1</version>
+			<version>2.0.0</version>
 		</dependency>
 
 		<!-- Test dependencies -->


arina-ielchiieva:
+1, LGTM.

vrozov:
@vvysotskyi It is a major version change of the dependency, but except for the pom file changes, there are no other changes. It will be good to understand what triggered a major version change (do release notes cover that?) and how it may affect gis plugin. Also, there is another PR on the review related to gis, can this change be combined with the other changes?

vvysotskyi:
@vrozov As I understand from the [release notes](https://github.com/Esri/geometry-api-java/releases/tag/v2.0.0), the main reason for the major version change was changing the interface due to the removing org.json dependency. Gis storage plugin provides UDF functions, and they are covered by the existing unit tests, so since all tests are passed this change does not break drill-gis. 
I think this change should be merged in the current release, but I don't know when other PR will be merged. Also, another pull request isn't affected by changing esri-geometry-api library, ie this change does not break changes made there.

vrozov:
@vvysotskyi OK, thanks. Currently, gis unit tests only work with CSV geo-spacial data. It sounds that JSON format is also supported and is not covered by the gis unit tests. Can you confirm if JSON format is supported and if yes, whether it is affected by the major version upgrade or not. In case JSON format is supported, it will be good to cover it in unit tests as well (not as part of this PR).

julianhyde:
If I recall correctly -- we went through this because Calcite uses ESRI -- the only change was to remove org.json. It's a pretty important change because org.json is not compatible with Apache license. And I think org.json data structures were used in the API (e.g. as arguments and method return values) so obsoleting it would be an API change.

vvysotskyi:
@vrozov, If you meant [geojson](http://geojson.org) when saying about JSON format support, then my answer - not, it is not supported yet.

vrozov:
@vvysotskyi I refer to JSON format that ESRI supports by itself.

vvysotskyi:
@vrozov, sorry for the misunderstanding. Currently, it is also not supported.

vrozov:
OK. LGTM.

PR #1013, opened 2017-10-27
DRILL-5910: Logging exception when custom AuthenticatorFactory not found

vladimirtkach:
We are giving  drill another chance to use other available AuthenticatorFactory

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/ClientAuthenticatorProvider.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/ClientAuthenticatorProvider.java
@@ -57,17 +57,17 @@ public class ClientAuthenticatorProvider implements AuthenticatorProvider {
 
     // then, custom factories
     if (customFactories != null) {
-      try {
-        final String[] factories = customFactories.split(",");
-        for (final String factory : factories) {
+      final String[] factories = customFactories.split(",");
+      for (final String factory : factories) {
+        try {
           final Class<?> clazz = Class.forName(factory);
           if (AuthenticatorFactory.class.isAssignableFrom(clazz)) {
             final AuthenticatorFactory instance = (AuthenticatorFactory) clazz.newInstance();
             authFactories.put(instance.getSimpleName(), instance);
           }
+        } catch (final ReflectiveOperationException e) {
+          logger.error("Failed to create auth factory {}", factory, e);
         }
-      } catch (final ClassNotFoundException | IllegalAccessException | InstantiationException e) {
-        throw new DrillRuntimeException("Failed to create auth factory.", e);
       }
     }
 


arina-ielchiieva:
+1, LGTM.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/ClientAuthenticatorProvider.java
```diff
@@ -57,17 +57,17 @@ private ClientAuthenticatorProvider() {
...
-        for (final String factory : factories) {
+      final String[] factories = customFactories.split(",");
+      for (final String factory : factories) {
+        try {
           final Class<?> clazz = Class.forName(factory);
           if (AuthenticatorFactory.class.isAssignableFrom(clazz)) {
             final AuthenticatorFactory instance = (AuthenticatorFactory) clazz.newInstance();
             authFactories.put(instance.getSimpleName(), instance);
           }
+        } catch (final ClassNotFoundException | IllegalAccessException | InstantiationException e) {
```
vrozov: catch ReflectiveOperationException.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/ClientAuthenticatorProvider.java
```diff
@@ -57,17 +57,17 @@ private ClientAuthenticatorProvider() {
...
+          throw new DrillRuntimeException(String.format("Failed to create auth factory '%s'", factory), e);
```
vrozov: Not directly related to this PR, but should it continue with the default set of factories if one of custom factories can't be instantiated?

prasadns14: I don't think we should continue. When a custom factory is requested, in this case a custom authenticator factory it doesn't make sense to continue if it can't be instantiated.

vrozov: It will be good to explain why. In other systems (for example unix pam) usually, there is a best effort attempt to authenticate, meaning that if a provider can't be initialized or instantiated it is skipped, but the system is operational and users that authenticate against other providers can still use the system.

sohami: @vrozov  +1 - It makes sense to log the error and continue rather than failing because a custom factory cannot be instantiated. This is because client environment might be set to configure these custom factories but then for authentication it might be using other mechanism like Plain/Kerberos in which case connection should succeed.

sohami:
I tried with the repro step and based on that doesn't see `DrillRuntimeException` message in stack trace but since the PR is adding the factory name in same exception message, we will not see that in stack trace as well.

vladimirtkach:
@sohami I'm getting this stack
```
java.lang.ExceptionInInitializerError
	at org.apache.drill.exec.rpc.security.ClientAuthenticatorProvider.getInstance(ClientAuthenticatorProvider.java:45)
	at org.apache.drill.exec.rpc.user.UserClient.getAuthenticatorFactory(UserClient.java:405)
......
Caused by: org.apache.drill.common.exceptions.DrillRuntimeException: Failed to create auth factory 'org.apache.drill.exec.rpc.security.maprsasl.MapRSaslFactory'
	at org.apache.drill.exec.rpc.security.ClientAuthenticatorProvider.<init>(ClientAuthenticatorProvider.java:69)
	at org.apache.drill.exec.rpc.security.ClientAuthenticatorProvider.<init>(ClientAuthenticatorProvider.java:30)
	at org.apache.drill.exec.rpc.security.ClientAuthenticatorProvider$Holder.<clinit>(ClientAuthenticatorProvider.java:37)
	... 24 more
Caused by: java.lang.ClassNotFoundException: org.apache.drill.exec.rpc.security.maprsasl.MapRSaslFactory
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at org.apache.drill.exec.rpc.security.ClientAuthenticatorProvider.<init>(ClientAuthenticatorProvider.java:63)
	... 26 more
```

vrozov:
@vladimirtkach Who is the target of the message? If it is Java developers, the message makes sense. If it is for drill users or devops, the message is useless.

arina-ielchiieva:
@vrozov as you can see from the changes, Volodymyr just has added additional information to the error message. I've seen from the previous discussions that you suggest just to log this exception and move forward. Will this be acceptable?

vrozov:
@arina-ielchiieva I guess just logging an exception does not help unless it targets Java developers. Consider logging a message that has information regarding which authentication provider can't be instantiated/initialized and continue.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/security/ClientAuthenticatorProvider.java
vladimirtkach: a moot point, I think that three specific exceptions is better the one general.

vladimirtkach:
@vrozov made the changes, please review.

vrozov:
LGTM

PR #1015, opened 2017-10-28
DRILL-5899: Simple pattern matchers can work with DrillBuf directly

ppadma:
For the 4 simple patterns we have i.e. startsWith, endsWith, contains and constant,, we do not need the overhead of charSequenceWrapper. We can work with DrillBuf directly. This will save us from doing isAscii check and UTF8 decoding for each row.
UTF-8 encoding ensures that no UTF-8 character is a prefix of any other valid character. So, instead of decoding varChar from each row we are processing, encode the patternString once during setup and do raw byte comparison. 
This improved overall performance for filter operator by around 20%.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/AbstractSqlPatternMatcher.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/AbstractSqlPatternMatcher.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.expr.fn.impl;
+
+import com.google.common.base.Charsets;
+import org.apache.drill.common.exceptions.UserException;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CharsetEncoder;
+import static org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.logger;
+
+/**
+ * To get good performance for most commonly used pattern matches
+ * CONSTANT('ABC') {@link SqlPatternConstantMatcher}
+ * STARTSWITH('%ABC') {@link SqlPatternStartsWithMatcher}
+ * ENDSWITH('ABC%') {@link SqlPatternEndsWithMatcher }
+ * CONTAINS('%ABC%') {@link SqlPatternContainsMatcher}
+ * we have simple pattern matchers.
+ * Idea is to have our own implementation for simple pattern matchers so we can
+ * avoid heavy weight regex processing, skip UTF-8 decoding and char conversion.
+ * Instead, we encode the pattern string and do byte comparison against native memory.
+ * Overall, this approach
+ * gives us orders of magnitude performance improvement for simple pattern matches.
+ * Anything that is not simple is considered
+ * complex pattern and we use Java regex for complex pattern matches.
+ */
+
+public abstract class AbstractSqlPatternMatcher implements SqlPatternMatcher {
+  protected final String patternString;
+  protected final int patternLength;
+  protected final ByteBuffer patternByteBuffer;
+
+  public AbstractSqlPatternMatcher(String patternString) {
+    this.patternString = patternString;
+
+    final CharsetEncoder charsetEncoder = Charsets.UTF_8.newEncoder();
+    final CharBuffer patternCharBuffer = CharBuffer.wrap(patternString);
+
+    try {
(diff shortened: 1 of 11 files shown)


ppadma:
@sachouche @paul-rogers can you please review ?

sachouche:
Padma, DRILL-5889 is the wrong JIRA (sqlline loses RPC..); I think your JIRA is 5899


Regards,

Salim

ppadma:
@sachouche @paul-rogers Thanks for the review. I updated the PR with review comments. 

I made one more change.  Previously, I was copying native memory buffer into byte array and using it. Instead, if we go to native memory  directly, performance is significantly better. In fact,  it is 3 times faster :-)

Please review updated changes.

paul-rogers:
Second long list of detailed comments sent privately to keep dev list traffic down.

ppadma:
@paul-rogers Thanks a lot for the review. Updated the PR with code review comments. Please take a look. 

Overall, good improvement with this change. Here are the numbers.

select count(*) from `/Users/ppenumarthy/MAPRTECH/padma/testdata` where l_comment like '%a' 
1.4 sec vs 7 sec

select count(*) from `/Users/ppenumarthy/MAPRTECH/padma/testdata` where l_comment like '%a%' 
6.5 sec vs 13.5 sec

select count(*) from `/Users/ppenumarthy/MAPRTECH/padma/testdata` where l_comment like 'a%' 
1.4 sec vs 5.8 sec

select count(*) from `/Users/ppenumarthy/MAPRTECH/padma/testdata` where l_comment like 'a' 
1.1.65 sec vs 5.8 sec


I think for "contains", improvement is not as much as others, probably because of nested for loops. @sachouche changes on top of these changes can improve further.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/AbstractSqlPatternMatcher.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.google.common.base.Charsets;
+import org.apache.drill.common.exceptions.UserException;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.CharsetEncoder;
+import static org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.logger;
+
+// To get good performance for most commonly used pattern matches
```
paul-rogers: Javadoc?

```
/**
 * This is a Javadoc comment and appears in generated documentation.
 */
// This is a plain comment and does not appear in documentation.
```

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/AbstractSqlPatternMatcher.java
```diff
@@ -0,0 +1,61 @@
...
+// Idea is to have our own implementation for simple pattern matchers so we can
+// avoid heavy weight regex processing, skip UTF-8 decoding and char conversion.
+// Instead, we encode the pattern string and do byte comparison against native memory.
+// Overall, this approach
+// gives us orders of magnitude performance improvement for simple pattern matches.
+// Anything that is not simple is considered
+// complex pattern and we use Java regex for complex pattern matches.
+
+public abstract class AbstractSqlPatternMatcher implements SqlPatternMatcher {
+  final String patternString;
```
paul-rogers: `protected final`

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -17,37 +17,48 @@
...
+    // no match if input string length is less than pattern length
+    if (txtLength < patternLength) {
+      return 0;
+    }
+
+    outer:
+    for (int txtIndex = 0; txtIndex < txtLength; txtIndex++) {
+
+      // boundary check
+      if (txtIndex + patternLength > txtLength) {
```
paul-rogers: Better:
```
int end = txtLength - patternLength;
for (int txtIndex = 0; txtIndex < end; txtIndex++) {
```
And omit the boundary check on every iteration. That is, no reason to iterate past the last possible match, then use an if-statement to shorten the loop. Just shorten the loop.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternEndsWithMatcher.java
```diff
@@ -17,33 +17,30 @@
...
   }
 
   @Override
-  public int match() {
-    int txtIndex = charSequenceWrapper.length();
-    int patternIndex = patternLength;
-    boolean matchFound = true; // if pattern is empty string, we always match.
+  public int match(int start, int end, DrillBuf drillBuf) {
+
+    if ( (end - start) < patternLength) { // No match if input string length is less than pattern length.
```
paul-rogers: `( (end - start)` --> `(end - start`

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternEndsWithMatcher.java
```diff
@@ -17,33 +17,30 @@
...
+      return 0;
+    }
 
     // simplePattern string has meta characters i.e % and _ and escape characters removed.
     // so, we can just directly compare.
-    while (patternIndex > 0 && txtIndex > 0) {
-      if (charSequenceWrapper.charAt(--txtIndex) != patternString.charAt(--patternIndex)) {
-        matchFound = false;
-        break;
+    for (int index = 1; index <= patternLength; index++) {
```
paul-rogers:
```
int txtStart = end - patternLength;
if (txtStart < start) { return 0; }
for (int index = 0; index < patternLength; index++) {
   ... patternByteBuffer.get(index) ... drillBuf.getByte(txtStart + index) ...
```

ppadma:
@paul-rogers updated with latest review comments taken care of. Please take a look.

PR #1016, opened 2017-10-28
DRILL-5913:solve the mixed processing of same functions with same inputRefs but di…

weijietong:
`DrillReduceAggregatesRule` mix the processing of same functions with same inputRefs but different dataTypes.

The error info and related reproducible sample sql are [here](https://issues.apache.org/jira/browse/DRILL-5913) 

I will also try to concat the Calcite devs to make sure whether they agree to make the `RexBuilder.addAggCall` distinguish the same `AggregateCall`s with same inputRefs but different dataTypes.


weijietong:
@amansinha100 maybe you are familiar with this part of codes .  Could you give a review ? anyone else will also be welcome.

weijietong:
@vvysotskyi @amansinha100  could you take a look at this PR. I ever contact with @julianhyde . Since Calcite treats stddev stddev_samp input parameter data type as their original data type,no cast behavior happens at its` AggregateReduceFunctionsRule` implementation.So this error will not happen at Calcite. So this PR changes our Drill own `DrillReduceAggregatesRule` implementation.

weijietong:
@KulykRoman seems you are familiar with this part of codes . Could you also take look at this.

vvysotskyi:
@weijietong, could you please check that this bug is still reproduced on current master? I tried a query from the Jira description and it is finished successfully. I suppose it was fixed in the scope of Calcite upgrade.

weijietong:
@vvysotskyi  I have tested the jira issue sql on the current master and it passed .But another new test case:
```
  @Test
  public void testDRILL5913_t() throws Exception {
    test("select n_nationkey, stddev((case when ( bigint_col  ) >0 then 1 else 0 end)) * 1.0  as col1, avg((case when ( bigint_col) >0 then 1 else 0 end)) * 1.0 as col2 from " 
         + "(  select n_name,n_nationkey, sum( n_regionkey) as bigint_col from cp.`tpch/nation.parquet` group by n_name,n_nationkey ) t group by n_nationkey");

  }
```
will throw another Exception at Drill version 1.13 with Calcite version 1.15 but passed at current master. The exception message is:
```
Caused by: java.lang.AssertionError: Type mismatch:
rel rowtype:
RecordType(ANY n_nationkey, BIGINT $f1, BIGINT $f2, BIGINT NOT NULL $f3, BIGINT $f4) NOT NULL
equivRel rowtype:
RecordType(ANY n_nationkey, BIGINT $f1, BIGINT $f2, BIGINT NOT NULL $f3, BIGINT NOT NULL $f4) NOT NULL
```

All of the main reason is that DrillReduceAggregationRule.reduceAgg invoked RexBuilder.addAggCall method whose parameter aggCallMapping acts as a AggCall cache. The aggCallMapping cache only care about the call name not the data type. The current master code of Calcite does nothing about this part since I find this bug. I don't think I can exhaustive all the test cases to prove our current master implementation right. But it seems security to have my tuned part of codes (validating AggCall cache with data type) to the master to prevent any future possible issues.

vvysotskyi:
@weijietong, the root cause of this error may have another reason. Drill has its own rules to determine the return type for aggregate functions, and in the most cases, it differs from Calcite rules. I suppose in some places were used rules for Calcite, but should be used rules for Drill.

Regarding the creating new calls for the same agg call with the same inputs, I suppose it will be ineffective to create the new calls only because return types differ. The problem is that return type was chosen incorrectly.

weijietong:
I have noticed the return type inference codes. The return type is right. To current implementation, if there are two agg calls with the same inputs but truly with different data types,it will definitely go to choose wrong agg call for reuse and cause potential errors.

weijietong:
@vvysotskyi do you make sure that two same name aggs with the same input ref but different data types should never happen here?

vvysotskyi:
@weijietong, yes, they should never happen.
I think the bug is that was used `sum0` instead of `sum` aggregate function or vice versa. `sum0` has the non-nullable return type, but the return type of `sum` is nullable.

Could you please review `DrillReduceAggregatesRule` and check for this problem?

PR #1017, opened 2017-10-31
DRILL-5822: The query with "SELECT *" with "ORDER BY" clause and `planner.slice_t…

vdiravka:
…arget`=1 doesn't preserve column order

Issue: Columns ordering doesn't preserve for the star query with sorting when this is planned into multiple fragments.
Solution: The commit for DRILL-847 is oudated. There is no need to canonicalize the batch or container since RecordBatchLoader swallows the "schema change" for now if two batches have different column ordering.


vdiravka:
@paul-rogers Could you please review this PR? You can find a short description here or more detailed - in the jira ticket [DRILL-5822](https://issues.apache.org/jira/browse/DRILL-5822).

PR #1020, opened 2017-11-02
DRILL-5921: Display counter metrics in table

prasadns14:
Listed the counter metrics in a table
@arina-ielchiieva please review

### DIFF
--- a/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
+++ b/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
@@ -63,12 +63,6 @@
           <strong>No histograms.</strong>
         </div>
       </div>
-      <h3 id="meters">Meters</h3>
-      <div id="metersVal">
-        <div class="alert alert-info">
-          <strong>No meters.</strong>
-        </div>
-      </div>
       <h3 id="timers">Timers</h3>
       <div id="timersVal">
         <div class="alert alert-info">
@@ -161,7 +155,6 @@
         if(! $.isEmptyObject(metrics.timers)) createTable(metrics.timers, "timers");
         if(! $.isEmptyObject(metrics.histograms)) createTable(metrics.histograms, "histograms");
         if(! $.isEmptyObject(metrics.counters)) createCountersTable(metrics.counters);
-        if(! $.isEmptyObject(metrics.meters)) $("#metersVal").html(JSON.stringify(metrics.meters, null, 2));
       });
     };
 
--- a/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
+++ b/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
@@ -98,6 +98,22 @@
       });
     };
 
+    function createCountersTable(counters) {
+      $("#countersVal").html(function() {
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
```diff
@@ -138,21 +154,14 @@
...
-
     var update = function() {
       $.get("/status/metrics", function(metrics) {
         updateGauges(metrics.gauges);
         updateBars(metrics.gauges);
         if(! $.isEmptyObject(metrics.timers)) createTable(metrics.timers, "timers");
         if(! $.isEmptyObject(metrics.histograms)) createTable(metrics.histograms, "histograms");
-        updateOthers(metrics);
+        if(! $.isEmptyObject(metrics.counters)) createCountersTable(metrics.counters);
+        if(! $.isEmptyObject(metrics.meters)) $("#metersVal").html(JSON.stringify(metrics.meters, null, 2));
```
arina-ielchiieva: @prasadns14 
1. Please add two screenshots before and after the changes.
2. Can you please think of the way to make create table generic so can be used for timers, histograms and counters?
3. What about meters? How they are displayed right now? Maybe we need to display them in table as well?
Ideally, we can display all metrics in the same way.

prasadns14: @arina-ielchiieva, please review

1) Added three screenshots before (current behavior), snapshot1(reusing createTable function) and snapshot2 (new createCountersTable function)

2) The function "createTable" lists the keys for each metric class, which looks good for "timers" and "histograms" as they have many keys. But the "counters" metric has only one key "count" (same as guages metric which has only key value). So, I have a added a new function to just list the class and count value.

3) Currently we do not have any "meters" metric. So, I don't know how many keys "meters" metric will have. If we have multiple keys we can use createTable or else we may have to create a different function similar to createCountersTable. (depending on the key name)

I personally think if we have a single key, we shouldn't use the createTable function. We should just display the class and the only key value.

before
snapshot1
snapshot2

arina-ielchiieva: @prasadns14
1. Thanks for adding the screenshots.
2. Most of the code in `createTable` and `createCountersTable` coincide.  I suggested you make one function. For example, with three parameters, `createTable(metric, name, addReportingClass)`. When you don't need to add reporting class you'll call this method with false. Our goal here is generify existing methods rather then adding new specific with almost the same content.
3. If we don't have any meters, let's remove them.

prasadns14: @arina-ielchiieva,
I have considered reusing existing methods before deciding to have a separate method.
With the above suggestion, the table will now look as below-

drill.connections.rpc.control.encrypted    |  {count: 0}

'|' here is column delimiter. Do we want to display only the number in the second column or a key/value pair? 
I just wanted it to be consistent with the other metrics tables. (so I print value.count) 

Removed meters section.

arina-ielchiieva: Well, sounds good then, thanks for making the changes.

arina-ielchiieva:
+1, LGTM.

PR #1021, opened 2017-11-02
DRILL-5923: Display name for query state

prasadns14:
Defined display names for query state. 
@paul-rogers please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileUtil.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.server.rest.profile;
+
+import org.apache.drill.exec.proto.UserBitShared.QueryResult.QueryState;
+
+public class ProfileUtil {
+  // Display names for QueryState enum in UserBitShared.proto
+  private static final String[] queryStateDisplayNames = {
+    "Starting", // STARTING = 0
+    "Running", // RUNNING = 1
+    "Succeeded", // COMPLETED = 2
+    "Canceled", // CANCELED = 3
+    "Failed", // FAILED = 4
+    "CancellationRequested", // CANCELLATION_REQUESTED = 5
+    "Enqueued" // ENQUEUED = 6
+  };
+
+
+  /**
+   * Utility to return display name for query state
+   * @param queryState
+   * @return display string for query state
+   */
+  public final static String getQueryStateDisplayName(QueryState queryState) {
+    int queryStateOrdinal = queryState.getNumber();
+    if (queryStateOrdinal >= queryStateDisplayNames.length) {
+      return queryState.name();
+    } else {
+      return queryStateDisplayNames[queryStateOrdinal];
+    }
+  }
+}
--- a/exec/java-exec/src/main/resources/rest/profile/profile.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/profile.ftl
@@ -135,6 +135,7 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
 
   <#assign queueName = model.getProfile().getQueueName() />
   <#assign queued = queueName != "" && queueName != "-" />
+
   <div class="page-header"></div>
   <h3>Query Profile</h3>
   <div class="panel-group" id="query-profile-accordion">
@@ -162,7 +163,7 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
             </thead>
             <tbody>
               <tr>
-                  <td>${model.getProfile().getState().name()}</td>
+                  <td>${model.getQueryStateDisplayName()}</td>
                   <td>${model.getProfile().getForeman().getAddress()}</td>
                   <td>${model.getProfile().getTotalFragments()}</td>
      <#if queued>
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
@@ -63,7 +63,7 @@ import com.google.common.collect.Lists;
 @Path("/")
 @RolesAllowed(DrillUserPrincipal.AUTHENTICATED_ROLE)
 public class ProfileResources {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProfileResources.class);
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProfileResources.class);
 
   @Inject UserAuthEnabled authEnabled;
   @Inject WorkManager work;
@@ -217,7 +217,8 @@ public class ProfileResources {
             runningQueries.add(
                 new ProfileInfo(work.getContext().getConfig(),
                     runningEntry.getKey(), profile.getStart(), System.currentTimeMillis(),
-                    profile.getForeman().getAddress(), profile.getQuery(), profile.getState().name(),
+                    profile.getForeman().getAddress(), profile.getQuery(),
+                    ProfileUtil.getQueryStateDisplayName(profile.getState()),
                     profile.getUser(), profile.getTotalCost(), profile.getQueueName()));
           }
         } catch (Exception e) {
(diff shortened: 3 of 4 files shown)


exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -135,6 +135,8 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
 
   <#assign queueName = model.getProfile().getQueueName() />
   <#assign queued = queueName != "" && queueName != "-" />
+  <#assign queryStateDisplayName = ["Starting", "Running", "Succeeded", "Canceled", "Failed",
```
arina-ielchiieva: Why do we have need duplicated state display name array here and in `ProfileResources.java`. Is there a possibility to not duplicate it?

prasadns14: Query state is displayed in two places - list.ftl (accessed via ProfileResources.java) and profile.ftl (accessed via QueryProfile.proto). So, I had to duplicate the array once in ProfileResources.java & once in profile.ftl

arina-ielchiieva: Maybe you can create common freemarker function which list.ftl and profile.ftl will use to decode the state name thus we won't duplicate display names and always be sure they are in sync.

prasadns14: Yes, I could create a common freemarker function. But if the changes are not made in ProfileResources.java then the REST API /profile will still show the query state as "COMPLETED". It won't be in sync with the UI.

arina-ielchiieva: Prasad, maybe you can come up with different way to avoid display names duplication?

prasadns14:
@arina-ielchiieva, made changes to avoid display name duplication. Please review

arina-ielchiieva:
@prasadns14 as far as I understood, you made all these changes to replace `completed` with `succeeded`. What if you just make changes in State enum itself, refactor some code and thus no changes in rest part will be required? 
From UserBitShared.proto
```
	enum QueryState {
	  STARTING = 0; // query has been scheduled for execution. This is post-enqueued.
	  RUNNING = 1;
	  COMPLETED = 2; // query has completed successfully
	  CANCELED = 3; // query has been cancelled, and all cleanup is complete
	  FAILED = 4;
	  CANCELLATION_REQUESTED = 5; // cancellation has been requested, and is being processed
	  ENQUEUED = 6; // query has been enqueued. this is pre-starting.
	}
```
After the renaming, please don't forget to regenerate protobuf.

paul-rogers:
@arina-ielchiieva, @prasadns14, here is my two cents.

The names and numbers used in the protobuf definitions are part of Drill's public network API. This API is not versioned, so we can't really change it. If we changed the names, then, say, C-code or Java code that expects the old names will break. Being part of the public API, that code may not even be in the Drill source tree; perhaps someone has generated, say, a Python binding. So, can't change the public API.

For purely aesthetic reasons, the contributor wishes to change the message displayed in the UI. This is purely a UI decision (the user is not expected to map the display names to the Protobuf enums.) And, the display name is subject to change. Maybe other UIs want to use other names. Maybe we want to show icons, or abbreviate the names ("Fail", "OK", etc.) And, of course, what if the display name should have spaces other characters: "In Progress", "In Queue" or "Didn't Work!". Can't put those in enum names. You get the idea.

For this reason, the mapping from enum values to display names should be part of the UI, not the network protocol definition. The present change provides a UI-specific mapping from API Protobuf enum values to display strings, which seems like a good idea.

So, the key questions are:

* Should we use display strings other than the Protobuf constants (seems a good idea.)
* Should we do the mapping in Java or in Freemarker? (Java seems simpler.)

Thoughts?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/QueryStateDisplayName.java
```diff
@@ -0,0 +1,35 @@
+/**
```
arina-ielchiieva: Please use comment, not Java doc for header.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/QueryStateDisplayName.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.server.rest.profile;
+
+import org.apache.drill.exec.proto.UserBitShared.QueryResult.QueryState;
+
+public class QueryStateDisplayName {
```
arina-ielchiieva: Let's name this class more generically since in future we might wan't add some other helper methods as well. 
For example, `ProfileUtil`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/QueryStateDisplayName.java
```diff
@@ -0,0 +1,35 @@
...
+  // Values should correspond to the QueryState enum in UserBitShared.proto
```
arina-ielchiieva: Not quite right, value do not correspond to QueryState enum, it's rather a mapping to QueryState enum. To be more clear here I suggest you create a map with `QueryState` as a key and `String` as value:
```
STARTING -> Starting
COMPLETED -> Succeeded
```
Thus you'll be able to get desired representation value by `QueryState` and in case if value is null return some default, like `Unknown State`.

arina-ielchiieva:
@paul-rogers totally agree with your comments about public API (slipped from my mind when I was writing comment yesterday). I guess then it's fine to go with changes Prasad suggests, I just have left a couple of suggestions to make code clearer.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/QueryStateDisplayName.java
prasadns14: will fix it

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/QueryStateDisplayName.java
prasadns14: sure will give a generic name

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/QueryStateDisplayName.java
prasadns14: QueryState is an enum.
I feel having an array is simple. We could use the QueryState ordinal as index to fetch the display name from the array. I can add a check to check for array index overflow, and return some default value like "Unknown state".

prasadns14:
@arina-ielchiieva, please review

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/QueryStateDisplayName.java
arina-ielchiieva: @prasadns14 
1. When using array you depend on enum value order. If for some reason value order changes, your approach will be working incorrectly.
2. Having map will show which key is mapped to which value and it will be much easier to understand that, for example, completed is mapped to succeeded.

I recommend you use map approach.

prasadns14: @arina-ielchiieva 
yes, map will definitely make it to easier to visualize the mapping. 
Made the changes

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileUtil.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Map;
+
+import com.google.common.collect.Maps;
+
+public class ProfileUtil {
+  // Mapping query state names to display names
+  private static final Map<String, String> queryStateDisplayName;
+
+  static {
+    Map<String, String> displayNames = Maps.newHashMap();
```
arina-ielchiieva: 1. Please use `Map<QueryState, String>` since you're already receiving `QueryState` as in parameter in method.
Besides, it would guarantee you did not make mistake writing query state enum names.
2. `queryStateDisplayName` -> `queryStateDisplayNames`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileUtil.java
```diff
@@ -0,0 +1,57 @@
...
+
+
+  /**
+   * Utility to return display name for query state
+   * @param queryState
+   * @return display string for query state
+   */
+  public final static String getQueryStateDisplayName(QueryState queryState) {
+    String state = queryState.name();
+    if (queryStateDisplayName.containsKey(state)) {
```
arina-ielchiieva: This would be more optimal:
```
String state = queryStateDisplayNames.get(queryState);
if (state == null) {
   state = "Unknown State"
}
return state;
```

paul-rogers:
@arina-ielchiieva, it helps to think about the source of the enum. This is a Protobuf enum. The ordinal values cannot change; they are a contract between sender and receiver. We can add new ones, or retire old ones, but otherwise the values are frozen in time.

The array approach captures this reality. We could document the array better:
```
String displayNames[] = {
  "First Value", // FIRST_VALUE = 0
  "Second Value", // SECOND_VALUE = 1
  ...
};
```
We can also do a bounds check:
```
if (enumValue.ordinal() >= displayNames.length) {
  return enumValue.toString();
else
  return displayNames[enumValue.ordinal());
}
```
But, IMHO a map seems overkill for such a simple task. Yes, it works, but is unnecessary. As they say, "make it as simple as possible (but no simpler)."

arina-ielchiieva:
Well, I don't have strong preference here, we can use array, as long as Prasad makes it nicely documented as in your example rather then in one line.
```
String displayNames[] = {
  "First Value", // FIRST_VALUE = 0
  "Second Value", // SECOND_VALUE = 1
  ...
};
```

prasadns14:
@arina-ielchiieva, @paul-rogers 
Reverted to the array approach, also added documentation.

PR #1022, opened 2017-11-02
DRILL-5924: native-client: Support user-specified CXX_FLAGS

xhochy:
Also remove the redundant sets of `CMAKE_CXX_FLAGS`. They are set on a global scope so that there is no need to reset them more locally.


ilooner:
@parthchandra Please review.

PR #1023, opened 2017-11-03
DRILL-5922 Fixed Child Allocator Leak. DRILL-5926 Stabalize TestValueVector tests.

ilooner:
## DRILL-5922

 - QueryContext was never closed when the Foreman finished. So the query child allocator was never closed.
 - The PlanSplitter created a QueryContext temporarily to construct an RPC message but never closed it.
 - The waitForExit method was error prone. Changed it to use the standard condition variable pattern.

## DRILL-5926

The TestValueVector tests would run out of memory. I simple increased the MaxDirectMemorySize for the forked test processes in the pom to avoid this.

### DIFF
--- a/common/src/test/java/org/apache/drill/test/DrillTest.java
+++ b/common/src/test/java/org/apache/drill/test/DrillTest.java
@@ -54,7 +54,7 @@ public class DrillTest {
   static MemWatcher memWatcher;
   static String className;
 
-  @Rule public final TestRule TIMEOUT = TestTools.getTimeoutRule(100000);
+  @Rule public final TestRule TIMEOUT = TestTools.getTimeoutRule(100_000);
 
   @Rule public final TestLogReporter logOutcome = LOG_OUTCOME;
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
@@ -22,7 +22,6 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import org.apache.drill.common.SelfCleaningRunnable;
-import org.apache.drill.common.concurrent.ExtendedLatch;
 import org.apache.drill.exec.coord.ClusterCoordinator;
 import org.apache.drill.exec.metrics.DrillMetrics;
 import org.apache.drill.exec.proto.BitControl.FragmentStatus;
@@ -37,7 +36,6 @@ import org.apache.drill.exec.rpc.control.Controller;
 import org.apache.drill.exec.rpc.control.WorkEventBus;
 import org.apache.drill.exec.rpc.data.DataConnectionCreator;
 import org.apache.drill.exec.server.BootStrapContext;
-import org.apache.drill.exec.server.Drillbit;
 import org.apache.drill.exec.server.DrillbitContext;
 import org.apache.drill.exec.store.sys.PersistentStoreProvider;
 import org.apache.drill.exec.work.batch.ControlMessageHandler;
@@ -50,9 +48,12 @@ import org.apache.drill.exec.work.user.UserWorker;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Manages the running fragments in a Drillbit. Periodically requests run-time stats updates from fragments
@@ -61,12 +62,14 @@ import java.util.concurrent.Executor;
 public class WorkManager implements AutoCloseable {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(WorkManager.class);
 
+  public static final int EXIT_TIMEOUT_MS = 5000;
+
   /*
    * We use a {@see java.util.concurrent.ConcurrentHashMap} because it promises never to throw a
    * {@see java.util.ConcurrentModificationException}; we need that because the statusThread may
    * iterate over the map while other threads add FragmentExecutors via the {@see #WorkerBee}.
    */
-  private final Map<FragmentHandle, FragmentExecutor> runningFragments = new ConcurrentHashMap<>();
+  private final ConcurrentMap<FragmentHandle, FragmentExecutor> runningFragments = Maps.newConcurrentMap();
 
   private final ConcurrentMap<QueryId, Foreman> queries = Maps.newConcurrentMap();
 
@@ -79,8 +82,8 @@ public class WorkManager implements AutoCloseable {
   private final WorkEventBus workBus;
   private final Executor executor;
   private final StatusThread statusThread;
-  private long numOfRunningQueries;
-  private long numOfRunningFragments;
+  private final Lock isEmptyLock = new ReentrantLock();
+  private final Condition isEmptyCondition = isEmptyLock.newCondition();
 
   /**
    * How often the StatusThread collects statistics about running fragments.
@@ -162,51 +165,67 @@ public class WorkManager implements AutoCloseable {
     return dContext;
   }
 
-  private ExtendedLatch exitLatch = null; // used to wait to exit when things are still running
+  public void waitToExit(final boolean forcefulShutdown) {
+    isEmptyLock.lock();
 
-  /**
-   * Waits until it is safe to exit. Blocks until all currently running fragments have completed.
-   *
-   * <p>This is intended to be used by {@link org.apache.drill.exec.server.Drillbit#close()}.</p>
-   */
-  public void waitToExit(Drillbit bit, boolean forcefulShutdown) {
-    synchronized(this) {
-      numOfRunningQueries = queries.size();
-      numOfRunningFragments = runningFragments.size();
-      if ( queries.isEmpty() && runningFragments.isEmpty()) {
-        return;
+    try {
+      if (forcefulShutdown) {
+        final long startTime = System.currentTimeMillis();
+        final long endTime = startTime + EXIT_TIMEOUT_MS;
+        long currentTime;
+
+        while (!areQueriesAndFragmentsEmpty() && (currentTime = System.currentTimeMillis()) < endTime) {
+          try {
+            if (!isEmptyCondition.await(endTime - currentTime, TimeUnit.MILLISECONDS)) {
+              break;
+            }
+          } catch (InterruptedException e) {
+            logger.error("Interrupted while waiting to exit");
+          }
+        }
+
+        if (!areQueriesAndFragmentsEmpty()) {
+          logger.warn("Timed out after {} millis. Shutting down before all fragments and foremen " +
+            "have completed.", EXIT_TIMEOUT_MS);
+
+          for (QueryId queryId: queries.keySet()) {
+            logger.warn("Query {} is still running.", QueryIdHelper.getQueryId(queryId));
+          }
+
+          for (FragmentHandle fragmentHandle: runningFragments.keySet()) {
+            logger.warn("Fragment {} is still running.", QueryIdHelper.getQueryIdentifier(fragmentHandle));
+          }
+        }
+      } else {
+        while (!areQueriesAndFragmentsEmpty()) {
+          isEmptyCondition.awaitUninterruptibly();
+        }
       }
-      logger.info("Draining " + queries +" queries and "+ runningFragments+" fragments.");
-      exitLatch = new ExtendedLatch();
-    }
-    // Wait uninterruptibly until all the queries and running fragments on that drillbit goes down
-    // to zero
-    if( forcefulShutdown ) {
-      exitLatch.awaitUninterruptibly(5000);
-    } else {
-      exitLatch.awaitUninterruptibly();
+    } finally {
+      isEmptyLock.unlock();
     }
   }
 
+  private boolean areQueriesAndFragmentsEmpty() {
+    return queries.isEmpty() && runningFragments.isEmpty();
+  }
+
   /**
-   * If it is safe to exit, and the exitLatch is in use, signals it so that waitToExit() will
-   * unblock. Logs the number of pending fragments and queries that are running on that
-   * drillbit to track the progress of shutdown process.
+   * A thread calling the {@link #waitToExit(boolean)} method is notified when a foreman is retired.
    */
   private void indicateIfSafeToExit() {
-    synchronized(this) {
-      if (exitLatch != null) {
-        logger.info("Waiting for "+ queries.size() +" queries to complete before shutting down");
-        logger.info("Waiting for "+ runningFragments.size() +" running fragments to complete before shutting down");
-        if(runningFragments.size() > numOfRunningFragments|| queries.size() > numOfRunningQueries) {
-          logger.info("New Fragments or queries are added while drillbit is Shutting down");
-        }
-        if (queries.isEmpty() && runningFragments.isEmpty()) {
-          // Both Queries and Running fragments are empty.
-          // So its safe for the drillbit to exit.
-          exitLatch.countDown();
-        }
+    isEmptyLock.lock();
+    try {
+      logger.info("Waiting for "+ queries.size() +" queries to complete before shutting down");
+      logger.info("Waiting for "+ runningFragments.size() +" running fragments to complete before shutting down");
+
+      if (!areQueriesAndFragmentsEmpty()) {
+        logger.info("New Fragments or queries are added while drillbit is Shutting down");
+      } else {
+        isEmptyCondition.signal();
       }
+    } finally {
+      isEmptyLock.unlock();
     }
   }
   /**
@@ -256,9 +275,9 @@ public class WorkManager implements AutoCloseable {
 
       final QueryId queryId = foreman.getQueryId();
       final boolean wasRemoved = queries.remove(queryId, foreman);
+
       if (!wasRemoved) {
         logger.warn("Couldn't find retiring Foreman for query " + queryId);
-//        throw new IllegalStateException("Couldn't find retiring Foreman for query " + queryId);
       }
 
       indicateIfSafeToExit();
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
@@ -802,11 +802,16 @@ public class Foreman implements Runnable {
       fragmentsRunner.getBee().retireForeman(Foreman.this);
 
       try {
+        queryContext.close();
+      } catch (Exception e) {
+        logger.error("Unable to close query context for query {}", QueryIdHelper.getQueryId(queryId), e);
+      }
+
+      try {
         queryManager.close();
       } catch (final Exception e) {
         logger.warn("unable to close query manager", e);
       }
-
 
       queryStateProcessor.close();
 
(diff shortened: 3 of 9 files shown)


ilooner:
@Ben-Zvi @paul-rogers

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,51 @@ public DrillbitContext getContext() {
...
+   * This is intended to be used by {@link org.apache.drill.exec.server.Drillbit#close()}.
    */
   public void waitToExit() {
-    synchronized(this) {
-      if (queries.isEmpty() && runningFragments.isEmpty()) {
-        return;
+    final long startTime = System.currentTimeMillis();
+
+    try {
+      exitLock.lock();
```
vrozov: Usually, lock is called outside of try to avoid calling unlock in finally if lock was not successful.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,51 @@ public DrillbitContext getContext() {
...
+        if (queries.isEmpty() && runningFragments.isEmpty()) {
+          break;
+        }
+
+        try {
+          final boolean success = exitCondition.await(EXIT_TIMEOUT - diff, TimeUnit.MILLISECONDS);
+
+          if (!success) {
+            logger.warn("Timed out after %d millis while waiting to exit.", EXIT_TIMEOUT);
+            exitLock.lock();
```
vrozov: Why is this lock necessary?

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,51 @@ public DrillbitContext getContext() {
...
+          logger.warn("Interrupted while waiting to exit");
+          exitLock.lock();
+        }
       }
 
-      exitLatch = new ExtendedLatch();
-    }
+      if (!(queries.isEmpty() && runningFragments.isEmpty())) {
+        logger.warn("Timed out after %d millis. Shutting down before all fragments and foremen " +
+          "have completed.", EXIT_TIMEOUT);
```
Ben-Zvi: The "time out" time should be `diff`, not EXIT_TIMEOUT

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,51 @@ public DrillbitContext getContext() {
...
+        try {
+          final boolean success = exitCondition.await(EXIT_TIMEOUT - diff, TimeUnit.MILLISECONDS);
+
+          if (!success) {
+            logger.warn("Timed out after %d millis while waiting to exit.", EXIT_TIMEOUT);
+            exitLock.lock();
+          }
+        } catch (InterruptedException e) {
+          logger.warn("Interrupted while waiting to exit");
+          exitLock.lock();
```
Ben-Zvi: 1. Why calling lock() here ?
2. Should the code issue an error instead of a warning ?

pom.xml
```diff
@@ -442,7 +442,7 @@
               -Dorg.apache.drill.exec.server.Drillbit.system_options="org.apache.drill.exec.compile.ClassTransformer.scalar_replacement=on"
               -Ddrill.test.query.printing.silent=true
               -Ddrill.catastrophic_to_standard_out=true
-              -XX:MaxPermSize=512M -XX:MaxDirectMemorySize=3072M
+              -XX:MaxPermSize=512M -XX:MaxDirectMemorySize=4096M
```
Ben-Zvi: Did all the regression tests pass ?  This change could have some impact on other tests (e.g., if was expecting an OOM, now gone).

ilooner: All the unit tests passed locally on my laptop. They also passed on our build server along with all of the functional and TPCH tests.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: If the await times out, the condition variable does not automatically reacquire the lock for us, so we must do it manually.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: 1. If await is interrupted the condition variable does not acquire the lock for us, so we have to acquire it manually.
2. I'll change from warning to error.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: This should actually just break from the loop instead of print a message, since EXIT_TIMEOUT millis have elapsed in this case, and the Timed out message is already printed below. I've updated the PR with this change.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,49 @@ public DrillbitContext getContext() {
...
+        } catch (InterruptedException e) {
+          logger.error("Interrupted while waiting to exit");
+        }
       }
 
-      exitLatch = new ExtendedLatch();
-    }
+      if (!(queries.isEmpty() && runningFragments.isEmpty())) {
+        logger.warn("Timed out after %d millis. Shutting down before all fragments and foremen " +
+          "have completed.", EXIT_TIMEOUT);
```
paul-rogers: The logger has formatting, but does not use `String.format()` formats. Instead, it uses `{}`:

```
logger.warn("Timed out after {} millis...", EXIT_TIMEOUT);
```

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -828,8 +828,13 @@ public void close() throws Exception {
...
       }
 
-      // Remove the Foreman from the running query list.
-      bee.retireForeman(Foreman.this);
+      try {
+        queryContext.close();
+      } catch (Exception e) {
+        final String message = String.format("Unable to close query context for query {}",
+          QueryIdHelper.getQueryId(queryId));
+        logger.error(message, e);
```
paul-rogers:
```
logger.error("Unable...query {}", QueryIdHelper...);
```

The `{}` is a logger format pattern, not a `String.format()` pattern.

exec/java-exec/src/main/java/org/apache/drill/exec/work/user/PlanSplitter.java
```diff
@@ -79,6 +80,15 @@ public QueryPlanFragments planFragments(DrillbitContext dContext, QueryId queryI
       responseBuilder.setStatus(QueryState.FAILED);
       responseBuilder.setError(error);
     }
+
+    try {
+      queryContext.close();
+    } catch (Exception e) {
+      final String message = String.format("Error closing QueryContext when getting plan fragments for query %s",
+        QueryIdHelper.getQueryId(queryId));
+      logger.error(message, e);
```
paul-rogers: See above re log message formatting.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -59,12 +61,14 @@
 public class WorkManager implements AutoCloseable {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(WorkManager.class);
 
+  public static final long EXIT_TIMEOUT = 30000L;
```
paul-rogers: `30000L` --> `30_000L` for quicker reading.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,49 @@ public DrillbitContext getContext() {
...
   public void waitToExit() {
-    synchronized(this) {
-      if (queries.isEmpty() && runningFragments.isEmpty()) {
-        return;
+    final long startTime = System.currentTimeMillis();
+    exitLock.lock();
+
+    try {
+      long diff;
+      while ((diff = (System.currentTimeMillis() - startTime)) < EXIT_TIMEOUT) {
```
paul-rogers: Simpler:
```
final long endTime = System.currentTimeMillis() + EXIT_TIMEOUT;
...
while (System.currentTimeMillis() < endTime) {
...
```

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,49 @@ public DrillbitContext getContext() {
...
+        if (queries.isEmpty() && runningFragments.isEmpty()) {
+          break;
+        }
+
+        try {
+          final boolean success = exitCondition.await(EXIT_TIMEOUT - diff, TimeUnit.MILLISECONDS);
+
+          if (!success) {
+            break;
+          }
```
paul-rogers:
```
if (! exitCondition.await(...)) {
  break;
}
```

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/user/PlanSplitter.java
ilooner: Done

ilooner:
@arina-ielchiieva Please let me know if you have any additional comments.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,49 @@ public DrillbitContext getContext() {
...
-      if (queries.isEmpty() && runningFragments.isEmpty()) {
-        return;
+    final long startTime = System.currentTimeMillis();
+    final long endTime = startTime + EXIT_TIMEOUT;
+
+    exitLock.lock();
+
+    try {
+      long currentTime;
+      while ((currentTime = System.currentTimeMillis()) < endTime) {
```
vrozov: I'd recommend changing `while` condition to `queries` and `runningFragments` not empty check as a primary condition for exiting. Timeout (waitTime >0) is the secondary condition.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,49 @@ public DrillbitContext getContext() {
...
+          }
+        } catch (InterruptedException e) {
+          logger.error("Interrupted while waiting to exit");
+        }
       }
 
-      exitLatch = new ExtendedLatch();
-    }
+      if (!(queries.isEmpty() && runningFragments.isEmpty())) {
+        logger.warn("Timed out after {} millis. Shutting down before all fragments and foremen " +
```
vrozov: It may be good to add what queries and runningFragments are running to the log message.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -158,38 +165,49 @@ public DrillbitContext getContext() {
...
   private void indicateIfSafeToExit() {
-    synchronized(this) {
-      if (exitLatch != null) {
-        if (queries.isEmpty() && runningFragments.isEmpty()) {
-          exitLatch.countDown();
-        }
-      }
-    }
+    exitLock.lock();
+    exitCondition.signal();
```
vrozov: I'd recommend adding try/finally and checking the condition before signaling `exitCondition`. Consider renaming `exitCondition` to `isEmpty` or `isQueriesAndFragmentsEmpty`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
ilooner: *isEmpty* reminds me of a method name. I renamed it to *isEmptyCondition* instead.

ilooner:
Applied review comments

pom.xml
vrozov: It may be better to open a separate PR for DRILL-5926

ilooner: I could not properly test my PR because of a pre-existing sporadic test failure. This fix was necessary in order to test my changes. Errors with the builds should be fixed immediately, especially when they interfere with testing a change. So I will keep the fix in this PR.

vrozov: DRILL-5926 affects not only this PR but my PR  #1031 as well even though they are not related at all. At a minimum the change should be in a separate commit that refers to DRILL-5926 as it is a workaround and not an actual fix for the problem. In the future, the workaround may need to be reverted and if it is mixed with other changes, it will not be clear what needs to be reverted. Note that with other changes I would expect tests to require less memory, not more.

ilooner: I will move DRILL-5926 into a separate commit.

paul-rogers: This is a fix for a test that we don't really even need. We are attempting to verify that we can allocate a single value vector of 2 GB by doubling from a small amount. On the last allocation, we have a 1 GB vector doubling to 2 GB, so we temporarily need 3 GB.

But, vectors should never get this large. The batch size limitation project is doing work to limit vectors to the 16 MB Netty slab size.

For this reason, this is a workaround to a test for a feature that is actually a bug which we will be fixing in a later release.

ilooner: @paul-rogers I think there are a few choices available to us:

 1. Keep this change in a separate commit (already done) and revert it once the Batch Size Limitation work is done. The benefit is that the build will not be broken up until the Batch Size work is in. I can make a note on the Batch Size jira to revert this change as well.
 1. Remove this change and suffer with a broken build until the Batch Size work is complete. (not fun)
 1. Delete the offending tests instead as part of this PR.

I am in favor of option **1**. I am strongly against option **2** since ignoring build issues makes life much harder for everyone.

Please let me know what path forward you'd like to take.

paul-rogers: I'm in favor of 1 for now, then we can revert it, if needed, once the batch size work is available and we remove the offending test (option 3).

vrozov: I am in favor of 1 as well.

@paul-rogers One of the test that fails initially tries to allocate 2GB (without re-allocation). The test intermittently fails due to Pooled Allocator not releasing memory back to the system. I don't know if it is supposed to return memory back to the system when it is closed and whether it is supposed to be closed at all during the unit tests.

ilooner:
*Note for batch committer:* Please do not squash the two commits in this PR. Please see the discussion on **pom.xml** for details.

paul-rogers:
For commit, let's do this:

* With luck, Arina will commit two PRs this week that may conflict: PR #970, and PR #978.
* Tim should rebase this PR on top of those changes once they are committed.
* If the changes are non-trivial, they may require additional review. If so, Paul to do that early the week of the 13th.
* If there is time to squeeze in this commit before Arina starts the 1.12, release, Paul will do a one-off commit.
* Otherwise, since Arina wants to do the release during that week, this commit may have to wait until after that release.
* Once this commit is in, Paul to rebase the PR #914, as it changed some of the same files.

arina-ielchiieva:
@ilooner please resolve conflicts.

ilooner:
@arina-ielchiieva Spoke with @paul-rogers . This will likely introduce some conflicts for https://github.com/apache/drill/pull/921 . In an effort to lighten the workload for that contributor we'll delay merging this change. I'll remove the ready-to-commit label on the jira until the other PR is merged. After https://github.com/apache/drill/pull/921 is merged I'll rebase and mark this as read-to-commit once more.

ilooner:
@paul-rogers Can you take another pass at reviewing this? I have resolved conflicts with the graceful shutdown work. The graceful shutdown tests were also timing out on my laptop so I fixed them.

common/src/test/java/org/apache/drill/test/DrillTest.java
```diff
@@ -54,7 +54,7 @@
   static MemWatcher memWatcher;
   static String className;
 
-  @Rule public final TestRule TIMEOUT = TestTools.getTimeoutRule(100000);
+  @Rule public final TestRule TIMEOUT = TestTools.getTimeoutRule(180000);
```
paul-rogers: Two comments. One a nit: I find it hard to read 180000. Suggestion, use 180_000.

More substantial comment: 180K ms = 180 seconds = 3 minutes. This seems FAR too long for a default timeout.

I wonder, should we make this much smaller? Then, find the tests that need extra time and mark them specially? Maybe tag them with "SlowTests" and set a larger limit for just those tests?

This will cause tests to fail if they are too slow. That, in turn, will force us to either fix those tests, or label them as slow.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -61,12 +63,14 @@
 public class WorkManager implements AutoCloseable {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(WorkManager.class);
 
+  public static final long EXIT_TIMEOUT = 5_000L;
```
paul-rogers: Three nits:

1. We'll never wait decades, so this need not be a long.
2. Please encode units in the name: `EXIT_TIMEOUT_MS`
3. The underscore is great, but not entirely needed for four-digit numbers.

ilooner:
@paul-rogers Applied review comments. I'll squash commits and mark this as ready to commit

pom.xml
paul-rogers: Why not just retire the test. As explained above, no vector should ever allocate 2 GB unless we do some fixes to the Netty allocator. (Such a large vector will likely introduce issues in the network layer, but we don't have any user JIRAs to confirm this is, indeed, a problem.)

ilooner:
@paul-rogers For some reason this wan't included in a batch commit. After rebasing onto master I noticed some tests were failing sporadically so I've included a 1 line fix for DRILL-6003 in the last commit. Can you please take a look and +1 again? Hopefully the change will go in this time.

ilooner:
Travis build is flakey. Rerunning build.

PR #1025, opened 2017-11-07
DRILL-5936: Refactor MergingRecordBatch based on code inspection

vrozov:


priteshm:
@amansinha100 can you review this change?

amansinha100:
+1  with a minor comment.   In the commit message and JIRA it would be better to say 'code inspection'  instead of code review which may be interpreted to mean the normal code review process.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
```diff
@@ -177,11 +177,11 @@ public IterOutcome innerNext() {
     }
     boolean schemaChanged = false;
 
-    if (prevBatchWasFull) {
+    if (!prevBatchNotFull) {
```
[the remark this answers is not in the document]
vrozov: @amansinha100 Addressed minor comment, please review delta and I'll squash it.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
```diff
@@ -795,6 +788,8 @@ private void generateComparisons(final ClassGenerator<?> g, final VectorAccessib
    * @param node Reference to the next record to copy from the incoming batches
    */
   private boolean copyRecordToOutgoingBatch(final Node node) {
+    assert outgoingPosition < OUTGOING_BATCH_SIZE
```
amansinha100: Doing this assert check on each record would add some overhead.  Since the caller already is checking outgoingBatchHasSpace in the while loop, this check is not needed.  You could add a javadoc comment that caller is expected to do the validity check.

vrozov: I added the assert to avoid possible errors during further code refactoring. As it is an assert that will not affect performance in production and there is another assert already, I'd prefer to keep it.

amansinha100: Ok.  Updated changes lgtm.

PR #1026, opened 2017-11-07
DRILL-5919:  Add non-numeric support for JSON processing

vladimirtkach:
1. Added two session options store.json.reader.non_numeric_numbers and store.json.reader.non_numeric_numbers that allow to read/write NaN and Infinity as numbers. By default these options  are set to true.
2. Extended signature of convert_toJSON and convert_fromJSON functions by adding second optional parameter  that enables/disables read/write NaN and Infinity. By default it is set true.
3. Added unit tests with nan, infitity values for math and aggregate functions
4. Replaced JsonReader's constructors with builder.

### DIFF
--- a/contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoRecordReader.java
+++ b/contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoRecordReader.java
@@ -25,6 +25,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.collect.ImmutableList;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.common.expression.SchemaPath;
@@ -73,6 +74,7 @@ public class MongoRecordReader extends AbstractRecordReader {
   private final MongoStoragePlugin plugin;
 
   private final boolean enableAllTextMode;
+  private final boolean enableNanInf;
   private final boolean readNumbersAsDouble;
   private boolean unionEnabled;
   private final boolean isBsonRecordReader;
@@ -92,6 +94,7 @@ public class MongoRecordReader extends AbstractRecordReader {
 
     buildFilters(subScanSpec.getFilter(), mergedFilters);
     enableAllTextMode = fragmentContext.getOptions().getOption(ExecConstants.MONGO_ALL_TEXT_MODE).bool_val;
+    enableNanInf = fragmentContext.getOptions().getOption(ExecConstants.JSON_READER_NAN_INF_NUMBERS).bool_val;
     readNumbersAsDouble = fragmentContext.getOptions().getOption(ExecConstants.MONGO_READER_READ_NUMBERS_AS_DOUBLE).bool_val;
     isBsonRecordReader = fragmentContext.getOptions().getOption(ExecConstants.MONGO_BSON_RECORD_READER).bool_val;
     logger.debug("BsonRecordReader is enabled? " + isBsonRecordReader);
@@ -159,8 +162,12 @@ public class MongoRecordReader extends AbstractRecordReader {
           readNumbersAsDouble);
       logger.debug("Initialized BsonRecordReader. ");
     } else {
-      this.jsonReader = new JsonReader(fragmentContext.getManagedBuffer(), Lists.newArrayList(getColumns()),
-          enableAllTextMode, false, readNumbersAsDouble);
+      this.jsonReader = new JsonReader.Builder(fragmentContext.getManagedBuffer())
+          .schemaPathColumns(Lists.newArrayList(getColumns()))
+          .allTextMode(enableAllTextMode)
+          .readNumbersAsDouble(readNumbersAsDouble)
+          .enableNanInf(enableNanInf)
+          .build();
       logger.debug(" Intialized JsonRecordReader. ");
     }
   }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertFrom.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertFrom.java
@@ -30,6 +30,7 @@ import org.apache.drill.exec.expr.annotations.FunctionTemplate.NullHandling;
 import org.apache.drill.exec.expr.annotations.Output;
 import org.apache.drill.exec.expr.annotations.Param;
 import org.apache.drill.exec.expr.annotations.Workspace;
+import org.apache.drill.exec.expr.holders.BitHolder;
 import org.apache.drill.exec.expr.holders.VarBinaryHolder;
 import org.apache.drill.exec.expr.holders.VarCharHolder;
 import org.apache.drill.exec.vector.complex.writer.BaseWriter.ComplexWriter;
@@ -50,7 +51,9 @@ public class JsonConvertFrom {
     @Output ComplexWriter writer;
 
     public void setup(){
-      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader(buffer, false, false, false /* do not read numbers as doubles */);
+      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader.Builder(buffer)
+          .defaultSchemaPathColumns()
+          .build();
     }
 
     public void eval(){
@@ -76,7 +79,9 @@ public class JsonConvertFrom {
     @Output ComplexWriter writer;
 
     public void setup(){
-      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader(buffer, false, false, false /* do not read numbers as doubles */);
+      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader.Builder(buffer)
+          .defaultSchemaPathColumns()
+          .build();
     }
 
     public void eval(){
@@ -90,5 +95,6 @@ public class JsonConvertFrom {
       }
     }
   }
+
 
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertTo.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertTo.java
@@ -28,6 +28,7 @@ import org.apache.drill.exec.expr.annotations.FunctionTemplate.FunctionScope;
 import org.apache.drill.exec.expr.annotations.FunctionTemplate.NullHandling;
 import org.apache.drill.exec.expr.annotations.Output;
 import org.apache.drill.exec.expr.annotations.Param;
+import org.apache.drill.exec.expr.holders.BitHolder;
 import org.apache.drill.exec.expr.holders.VarBinaryHolder;
 import org.apache.drill.exec.vector.complex.reader.FieldReader;
 
@@ -104,4 +105,5 @@ public class JsonConvertTo {
       out.end = bytea.length;
     }
   }
+
 }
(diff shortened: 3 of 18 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestJsonNonNumerics.java
```diff
@@ -0,0 +1,167 @@
...
+      test("alter session reset `store.json.writer.non_numeric_numbers`");
+      FileUtils.deleteQuietly(file);
+    }
+  }
+
+  @Test
+  public void testConvertFromJsonFunction() throws Exception {
+    File file = new File(getTempDir(""), "nan_test.csv");
+    String csv = "col_0, {\"nan\":NaN}";
+    FileUtils.writeStringToFile(file, csv);
```
arina-ielchiieva: Should be in try-catch.

exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestJsonNonNumerics.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.*;
+
+public class TestJsonNonNumerics extends BaseTestQuery {
+
+  @Test
+  public void testNonNumericSelect() throws Exception {
+    File file = new File(getTempDir(""), "nan_test.json");
```
arina-ielchiieva: It's better to pass dir name as well, rather than emptiness. Ex: `getTempDir("test_nan")`

vladimirtkach:
@arina-ielchiieva made code changes according to your comments.

arina-ielchiieva:
Thanks, +1, LGTM.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertFrom.java
```diff
@@ -50,7 +51,7 @@ private JsonConvertFrom(){}
     @Output ComplexWriter writer;
 
     public void setup(){
-      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader(buffer, false, false, false /* do not read numbers as doubles */);
+      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader(buffer, false, false, false, false /* do not read numbers as doubles */);
```
paul-rogers: Here, the comment refers to the second-to-last item. Consider this:

```
false, // What is the first one?
false, // What is the second one?
false, // do not read numbers as doubles
false // Do not allow Nan, INF
```

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertFrom.java
```diff
@@ -76,7 +77,7 @@ public void eval(){
...
```
paul-rogers: See above. Why do we have two copies? Can we have a function that returns the reader using default configs?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertFrom.java
```diff
@@ -91,4 +92,60 @@ public void eval(){
...
+
+    @Param VarCharHolder in;
+    @Param BitHolder enableNonNumeric;
+    @Inject DrillBuf buffer;
+    @Workspace org.apache.drill.exec.vector.complex.fn.JsonReader jsonReader;
+
+    @Output ComplexWriter writer;
+
+    public void setup(){
+      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader(buffer, false, false, false,/* do not read numbers as doubles */
```
paul-rogers: See above.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertFrom.java
```diff
@@ -91,4 +92,60 @@ public void eval(){
...
+
+    @Param VarBinaryHolder in;
+    @Param BitHolder enableNonNumeric;
+    @Inject DrillBuf buffer;
+    @Workspace org.apache.drill.exec.vector.complex.fn.JsonReader jsonReader;
+
+    @Output ComplexWriter writer;
+
+    public void setup(){
+      jsonReader = new org.apache.drill.exec.vector.complex.fn.JsonReader(buffer, false, false, false, /* do not read numbers as doubles */
```
paul-rogers: See above. We really don't want all these duplicate copies.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/conv/JsonConvertTo.java
```diff
@@ -90,7 +91,71 @@ public void eval(){
...
+
+    public void setup(){
+    }
+
+    public void eval(){
+      out.start = 0;
+
+      java.io.ByteArrayOutputStream stream = new java.io.ByteArrayOutputStream();
+      try {
+        org.apache.drill.exec.vector.complex.fn.JsonWriter jsonWriter = new org.apache.drill.exec.vector.complex.fn.JsonWriter(stream, true, false,
```
paul-rogers: More copies.

contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoRecordReader.java
```diff
@@ -73,6 +73,7 @@
   private final MongoStoragePlugin plugin;
 
   private final boolean enableAllTextMode;
+  private final boolean enableNonNumericNumbers;
```
paul-rogers: Would recommend: `enableNanInf`.

"Non-numeric numbers" sounds like we might allow "foo" or "thirteen".

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -502,6 +502,8 @@ drill.exec.options: {
     store.format: "parquet",
     store.hive.optimize_scan_with_native_readers: false,
     store.json.all_text_mode: false,
+    store.json.writer.non_numeric_numbers: false,
+    store.json.reader.non_numeric_numbers: false,
```
paul-rogers: `allow_nan_inf`?

paul-rogers: Any reason these are not enabled by default?

paul-rogers: Have we tested all Drill's floating point methods to ensure that they correctly handle NaN and INF?

exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestJsonNonNumerics.java
```diff
@@ -0,0 +1,167 @@
...
+      String resultJson = FileUtils.readFileToString(resultFile);
+      int nanIndex = resultJson.indexOf("NaN");
+      assertFalse("`NaN` must not be enclosed with \"\" ", resultJson.charAt(nanIndex - 1) == '"');
+      assertFalse("`NaN` must not be enclosed with \"\" ", resultJson.charAt(nanIndex + "NaN".length()) == '"');
+      int infIndex = resultJson.indexOf("Infinity");
+      assertFalse("`Infinity` must not be enclosed with \"\" ", resultJson.charAt(infIndex - 1) == '"');
+      assertFalse("`Infinity` must not be enclosed with \"\" ", resultJson.charAt(infIndex + "Infinity".length()) == '"');
+    } finally {
+      test("drop table if exists dfs_test.tmp.`%s`", tableName);
+      test("alter session reset `store.json.reader.non_numeric_numbers`");
```
paul-rogers: Should be done in a finally block to ensure these get reset even if the test fails.

paul-rogers: Please consider using the new `alterSession()` and `resetSession()` methods once they are available.

paul-rogers: Consider referencing the constant for the symbol, rather than the option name itself:

```
resetSystem(ExecConstants.YOUR_OPTION_NAME);
```

paul-rogers:
Further, is the extra option to `convertFromJSON` really needed? Can't we just accept `NaN` and `Infinity` by default?

Consider. If the option is off by default, users without `NaN` or `Infinity` data will see no difference. But, users will this data will get an error and have to hunt down the option to make their data work.

If the option is on by default, users without `NaN` or `Infinity` data will see no difference. But, users will this data will also have their queries work by default.

So, seems no harm in making the `NaN` and `Infinity` support turned on by default.

exec/java-exec/src/main/resources/drill-module.conf
vladimirtkach: No, we don't and we should test them. I think we have to create separate jira for testing math functions, because code changes from this PR doesn't affect logic of any math function, and while testing math function there will be other issues not connected with with functionality (like BigInteger constructor doesn't accept NaN, Infinity and others)

vladimirtkach: thanks, allow_nan_inf definitely better.

vladimirtkach: I think we should stick to json standard and leave them switched off by default. If user get an exception we may show the option name which he want to switch on.

vladimirtkach:
what do you think about having two functions instead: convertFromJSON and convertFromJSON+some suffix. Second will be able to convert  NaN, Infinity

exec/java-exec/src/main/resources/drill-module.conf
paul-rogers: As noted in DRILL-5949, options like this should be part of the plugin config (as in CSV), not session options. For now, let's leave them as session options and I'll migrate them to plugin config options along with the others.

paul-rogers: See discussion below. If they are off, then queries that use NaN and Infinity will fail until the user turns them on. Queries that don't use NaN and Infinity won't care. So, what is the advantage to failing queries unnecessarily?

paul-rogers: OK. So, if the rest of Drill either does not support (or we don't know if it supports) NaN and Inf, should we introduce the change here that potentially leads to failures elsewhere?

Do we know if JDBC and ODBC support these values? (I suppose they do as they are features of Java's float/double primitives...)

contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoRecordReader.java
paul-rogers: The JSON parser we use does use the term "non-numeric numbers", but not sure we want to keep that naming in Drill itself.

paul-rogers:
On the two functions... Maybe just have one function that handles the Nan/Infinity case. As noted earlier, no matter what we do, JSON without these symbols will work. So, we need only consider JSON with the symbols. Either:

* The NaN/Infinity cases always work, or
* The NaN/Infinity cases work sometimes, fail others, depending on some option or argument.

I would vote for the first case: it is simpler. I just can't see how anyone would use Drill to valid JSON and would want a query to fail if it contains NaN or Infinity. Can you suggest a case where failing the query would be of help to the user?

vladimirtkach:
@paul-rogers Made changes, please review.

arina-ielchiieva:
@vladimirtkach thanks for adding tests for math functions. New changes look good to me.
@paul-rogers please take a look.

PR #1029, opened 2017-11-09
DRILL-5867: List profiles in pages rather than a long verbose listing

kkhatua:
Leverage existing DataTables libraries to paginate a long pre-fetched list of profiles for listing.
Added benefit of querying through that list (search field) is also available for a user
Minor change made to the display text for prefetching of profiles (DRILL-5259) so that it is not confused with what this commit adds to the UI.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
@@ -71,6 +71,9 @@ public class ProfileResources {
   @Inject SecurityContext sc;
 
   public static class ProfileInfo implements Comparable<ProfileInfo> {
+    private static final int QUERY_SNIPPET_MAX_CHAR = 150;
+    private static final int QUERY_SNIPPET_MAX_LINES = 8;
+
     public static final SimpleDateFormat format = new SimpleDateFormat("MM/dd/yyyy HH:mm:ss");
 
     private final String queryId;
@@ -93,7 +96,7 @@ public class ProfileResources {
       this.time = new Date(startTime);
       this.foreman = foreman;
       this.link = generateLink(drillConfig, foreman, queryId);
-      this.query = query.substring(0,  Math.min(query.length(), 150));
+      this.query = extractQuerySnippet(query);
       this.state = state;
       this.user = user;
       this.totalCost = totalCost;
@@ -127,6 +130,10 @@ public class ProfileResources {
 
     public String getForeman() { return foreman; }
 
+    public double getTotalCost() { return totalCost; }
+
+    public String getQueueName() { return queueName; }
+
     /**
      * Generates link which will return query profile in json representation.
      *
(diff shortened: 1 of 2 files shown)


kkhatua:
@arina-ielchiieva / @paul-rogers Please review this.

kkhatua:
Snapshot when rendering the defaults (10 per page) from a pre-loaded set of the latest 123 profiles
image
The query snippet is restricted to 8 lines at most and indicates if there is more to the query text with a trailing set of `...`

kkhatua:
Snapshot when testing with search filter for FAILED queries and navigating to page 2 of that list. Information about the number of filtered items, etc is also provided.
image

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -93,13 +96,35 @@ public ProfileInfo(DrillConfig drillConfig, String queryId, long startTime, long
...
       this.link = generateLink(drillConfig, foreman, queryId);
-      this.query = query.substring(0,  Math.min(query.length(), 150));
+      this.query = extractQuerySnippet(query);
       this.state = state;
       this.user = user;
       this.totalCost = totalCost;
       this.queueName = queueName;
     }
 
+    private String extractQuerySnippet(String queryText) {
```
arina-ielchiieva: 1. I usually place private method int he end of of the class.
2. We can add javadoc here explaining that first we limit original query size and if size fits but query has too many lines we limit it as well for better readability on Web UI.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -93,13 +96,35 @@ public ProfileInfo(DrillConfig drillConfig, String queryId, long startTime, long
...
+      //Extract upto max char limit as snippet
+      String sizeCappedQuerySnippet = queryText.substring(0,  Math.min(queryText.length(), QUERY_SNIPPET_MAX_CHAR));
+      //Trimming down based on line-count
+      if ( QUERY_SNIPPET_MAX_LINES < sizeCappedQuerySnippet.split(System.lineSeparator()).length ) {
```
arina-ielchiieva: 1. We can create variable for `sizeCappedQuerySnippet.split(System.lineSeparator())` so we do split only once.
2. Please remove spaces in `if` clause: `if ( QUERY_SNIPPET_MAX_LINES < sizeCappedQuerySnippet.split(System.lineSeparator()).length ) {` -> `if (QUERY_SNIPPET_MAX_LINES < splittedQuery.length) {` and in `if ( ++linesConstructed < QUERY_SNIPPET_MAX_LINES ) {` in the code below.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
```diff
@@ -93,13 +96,35 @@ public ProfileInfo(DrillConfig drillConfig, String queryId, long startTime, long
...
+        int linesConstructed = 0;
+        StringBuilder lineCappedQuerySnippet = new StringBuilder();
+        String[] queryParts = sizeCappedQuerySnippet.split(System.lineSeparator());
+        for (String qPart : queryParts) {
+          lineCappedQuerySnippet.append(qPart);
+          if ( ++linesConstructed < QUERY_SNIPPET_MAX_LINES ) {
+            lineCappedQuerySnippet.append(System.lineSeparator());
```
arina-ielchiieva: Do we want to append with new line or maybe space for better readability?

kkhatua: I wanted to preserve the original query format, hence, I'm applying new line instead of spaces. Besides, the snippet can also carry single-line comments, and putting them into 1 line will make it appear that anything following a `--` is all comments

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: +1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileResources.java
kkhatua: +1
Will do this

arina-ielchiieva:
+1, LGTM.

PR #1032, opened 2017-11-10
DRILL-5089: Dynamically load schema of storage plugin only when neede…

chunhui-shi:
…d for every query

For each query, loading all storage plugins and loading all workspaces under file system plugins is not needed.

This patch use DynamicRootSchema as the root schema for Drill. Which loads correspondent storage only when needed.

infoschema to read full schema information and load second level schema accordingly.

for workspaces under the same Filesyetm, no need to create FileSystem for each workspace.

use fs.access API to check permission which is available after HDFS 2.6 except for windows + local file system case.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
@@ -230,7 +230,12 @@ public class FragmentContext extends BaseFragmentContext implements AutoCloseabl
     return context;
   }
 
-  public SchemaPlus getRootSchema() {
+  /**
+   * This method is only used to construt InfoSchemaReader, it is for the reader to get full schema, so here we
+   * are going to return a fully initialized schema tree.
+   * @return root schema's plus
+   */
+  public SchemaPlus getFullRootSchema() {
     if (queryContext == null) {
       fail(new UnsupportedOperationException("Schema tree can only be created in root fragment. " +
           "This is a non-root fragment."));
@@ -248,7 +253,7 @@ public class FragmentContext extends BaseFragmentContext implements AutoCloseabl
         .setIgnoreAuthErrors(isImpersonationEnabled)
         .build();
 
-    return queryContext.getRootSchema(schemaConfig);
+    return queryContext.getFullRootSchema(schemaConfig);
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.planner.sql;
+
+import com.google.common.collect.ImmutableSortedSet;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+import org.apache.calcite.DataContext;
+import org.apache.calcite.jdbc.CalciteRootSchema;
+import org.apache.calcite.jdbc.CalciteSchema;
+
+import org.apache.calcite.linq4j.tree.Expression;
+import org.apache.calcite.linq4j.tree.Expressions;
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.calcite.schema.impl.AbstractSchema;
+import org.apache.calcite.util.BuiltInMethod;
+import org.apache.calcite.util.Compatible;
+import org.apache.drill.common.exceptions.ExecutionSetupException;
+import org.apache.drill.exec.store.SchemaConfig;
+import org.apache.drill.exec.store.StoragePlugin;
+import org.apache.drill.exec.store.StoragePluginRegistry;
+import org.apache.drill.exec.store.SubSchemaWrapper;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.NavigableSet;
+import java.util.Set;
+
+/**
+ * This class is to allow us loading schemas from storage plugins later when {@link #getSubSchema(String, boolean)}
+ * is called.
+ */
+public class DynamicRootSchema extends DynamicSchema
+    implements CalciteRootSchema {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DynamicRootSchema.class);
+
+  protected SchemaConfig schemaConfig;
+  protected StoragePluginRegistry storages;
+
+  public StoragePluginRegistry getSchemaFactories() {
+    return storages;
+  }
+
+  /** Creates a root schema. */
+  DynamicRootSchema(StoragePluginRegistry storages, SchemaConfig schemaConfig) {
+    super(null, new RootSchema(), "");
+    this.schemaConfig = schemaConfig;
+    this.storages = storages;
+  }
+
+  @Override
+  public CalciteSchema getSubSchema(String schemaName, boolean caseSensitive) {
+    CalciteSchema retSchema = getSubSchemaMap().get(schemaName);
+    if (retSchema != null) {
+      return retSchema;
+    }
+
+    loadSchemaFactory(schemaName, caseSensitive);
+    retSchema = getSubSchemaMap().get(schemaName);
+    return retSchema;
+  }
+
+  @Override
+  public NavigableSet<String> getTableNames() {
+    return Compatible.INSTANCE.navigableSet(ImmutableSortedSet.<String>of());
+  }
+
+  /**
+   * load schema factory(storage plugin) for schemaName
+   * @param schemaName
+   * @param caseSensitive
+   */
+  public void loadSchemaFactory(String schemaName, boolean caseSensitive) {
+    try {
+      SchemaPlus thisPlus = this.plus();
+      StoragePlugin plugin = getSchemaFactories().getPlugin(schemaName);
+      if (plugin != null) {
+        plugin.registerSchemas(schemaConfig, thisPlus);
+        return;
+      }
+
+      // Could not find the plugin of schemaName. The schemaName could be `dfs.tmp`, a 2nd level schema under 'dfs'
+      String[] paths = schemaName.split("\\.");
+      if (paths.length == 2) {
+        plugin = getSchemaFactories().getPlugin(paths[0]);
+        if (plugin == null) {
+          return;
+        }
+
+        // Found the storage plugin for first part(e.g. 'dfs') of schemaName (e.g. 'dfs.tmp')
+        // register schema for this storage plugin to 'this'.
+        plugin.registerSchemas(schemaConfig, thisPlus);
+
+        // Load second level schemas for this storage plugin
+        final SchemaPlus firstlevelSchema = thisPlus.getSubSchema(paths[0]);
+        final List<SchemaPlus> secondLevelSchemas = Lists.newArrayList();
+        for (String secondLevelSchemaName : firstlevelSchema.getSubSchemaNames()) {
+          secondLevelSchemas.add(firstlevelSchema.getSubSchema(secondLevelSchemaName));
+        }
+
+        for (SchemaPlus schema : secondLevelSchemas) {
+          org.apache.drill.exec.store.AbstractSchema drillSchema;
+          try {
+            drillSchema = schema.unwrap(org.apache.drill.exec.store.AbstractSchema.class);
+          } catch (ClassCastException e) {
+            throw new RuntimeException(String.format("Schema '%s' is not expected under root schema", schema.getName()));
+          }
+          SubSchemaWrapper wrapper = new SubSchemaWrapper(drillSchema);
+          thisPlus.add(wrapper.getName(), wrapper);
+        }
+      }
+    } catch(ExecutionSetupException | IOException ex) {
+      logger.warn("Failed to load schema for \"" + schemaName + "\"!", ex);
+    }
+  }
+
+  static class RootSchema extends AbstractSchema {
+    @Override public Expression getExpression(SchemaPlus parentSchema,
+                                              String name) {
+      return Expressions.call(
+          DataContext.ROOT,
+          BuiltInMethod.DATA_CONTEXT_GET_ROOT_SCHEMA.method);
+    }
+  }
+}
+
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicSchema.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicSchema.java
@@ -1,0 +1,58 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.planner.sql;
+
+import org.apache.calcite.jdbc.CalciteSchema;
+import org.apache.calcite.jdbc.SimpleCalciteSchema;
+import org.apache.calcite.schema.Schema;
+import org.apache.calcite.schema.SchemaPlus;
+import org.apache.drill.exec.store.SchemaConfig;
+import org.apache.drill.exec.store.StoragePluginRegistry;
+
+
+/**
+ * Unlike SimpleCalciteSchema, DynamicSchema could have an empty or partial schemaMap, but it could maintain a map of
+ * name->SchemaFactory, and only register schema when the corresponsdent name is requested.
+ */
+public class DynamicSchema extends SimpleCalciteSchema {
+
+  public DynamicSchema(CalciteSchema parent, Schema schema, String name) {
+    super(parent, schema, name);
+  }
+
+  @Override
+  public CalciteSchema getSubSchema(String schemaName, boolean caseSensitive) {
+    Schema s = schema.getSubSchema(schemaName);
+    if (s != null) {
+      return new DynamicSchema(this, s, schemaName);
+    }
+    CalciteSchema ret = getSubSchemaMap().get(schemaName);
+    return ret;
+  }
+
+  @Override
+  public SchemaPlus plus() {
+    return super.plus();
+  }
+
+  public static SchemaPlus createRootSchema(StoragePluginRegistry storages, SchemaConfig schemaConfig) {
+    DynamicRootSchema rootSchema = new DynamicRootSchema(storages, schemaConfig);
+    return rootSchema.plus();
+  }
+
+}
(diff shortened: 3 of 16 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+  @Override
+  public CalciteSchema getSubSchema(String schemaName, boolean caseSensitive) {
+    CalciteSchema retSchema = getSubSchemaMap().get(schemaName);
+
+    if (retSchema == null) {
+      loadSchemaFactory(schemaName, caseSensitive);
+    }
+
+    retSchema = getSubSchemaMap().get(schemaName);
+    return retSchema;
```
paul-rogers: if the original call returns non-null, we make the same call a second time. Better:
```
retSchema = ...
if (retSchema != null) { return retSchema; }
loadSchemaFactory(...)
return getSubSchemaMap()...
```

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  @Override
+  public NavigableSet<String> getTableNames() {
+    Set<String> pluginNames = Sets.newHashSet();
```
paul-rogers: Should be case insensitive?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+   * @param schemaName
+   * @param caseSensitive
+   */
+  public void loadSchemaFactory(String schemaName, boolean caseSensitive) {
+    try {
+      SchemaPlus thisPlus = this.plus();
+      StoragePlugin plugin = getSchemaFactories().getPlugin(schemaName);
+      if (plugin != null) {
+        plugin.registerSchemas(schemaConfig, thisPlus);
+      }
```
paul-rogers: If the name is `dfs.test`, we first look up the compound name, then the parts? Why? Do we put the compound names in the map? Or can we have one schema named "dfs.test" and another called `dfs`.`test`? Or, can this code be restructured a bit?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+      else {
+        //this schema name could be `dfs.tmp`, a 2nd level schema under 'dfs'
+        String[] paths = schemaName.split("\\.");
```
paul-rogers: Should this be done here in this simple way? How many other places do we do the same thing? Or, should we have a common function to split schema names so we can handle, way, escapes and other special cases that might come along?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+        if (paths.length == 2) {
```
paul-rogers: Do we support only 1- and 2-part names? Should we assert that the length <= 2?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+          if (secondLevelSchema != null) {
+            SchemaPlus secondlevel = secondLevelSchema.plus();
+            org.apache.drill.exec.store.AbstractSchema drillSchema =
+                secondlevel.unwrap(org.apache.drill.exec.store.AbstractSchema.class);
+            SubSchemaWrapper wrapper = new SubSchemaWrapper(drillSchema);
+            thisPlus.add(wrapper.getName(), wrapper);
+          }
+        }
+      }
+    } catch(ExecutionSetupException | IOException ex) {
```
paul-rogers: Really? Just ignore errors? Maybe add a comment inside the catch block to explain why.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+  }
+
+  static class RootSchema extends AbstractSchema {
+    RootSchema() {
+      super();
+    }
```
paul-rogers: Omit, this is a degenerate ctor that spells out what Java already does.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+import org.apache.drill.exec.store.StoragePlugin;
+import org.apache.drill.exec.store.StoragePluginRegistry;
+import org.apache.drill.exec.store.SubSchemaWrapper;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.NavigableSet;
+import java.util.Set;
+
+public class DynamicRootSchema extends DynamicSchema
```
paul-rogers: Class would benefit from general comments: purpose, references, structure.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+        plugin.registerSchemas(schemaConfig, thisPlus);
+      }
+      else {
+        //this schema name could be `dfs.tmp`, a 2nd level schema under 'dfs'
+        String[] paths = schemaName.split("\\.");
+        if (paths.length == 2) {
+          plugin = getSchemaFactories().getPlugin(paths[0]);
+          if (plugin != null) {
+            plugin.registerSchemas(schemaConfig, thisPlus);
+          }
```
paul-rogers: What if the plugin is null? Should we fail with an error, or just keep going?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,125 @@
...
+
+          final CalciteSchema secondLevelSchema = getSubSchemaMap().get(paths[0]).getSubSchema(paths[1], caseSensitive);
+          if (secondLevelSchema != null) {
+            SchemaPlus secondlevel = secondLevelSchema.plus();
+            org.apache.drill.exec.store.AbstractSchema drillSchema =
```
paul-rogers: Fully qualified name needed? Or, will an import work?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicSchema.java
```diff
@@ -0,0 +1,65 @@
...
+
+
+/**
+ * Unlike SimpleCalciteSchema, DynamicSchema could have an empty or partial schemaMap, but it could maintain a map of
+ * name->SchemaFactory, and only register schema when the corresponsdent name is requested.
+ */
+public class DynamicSchema extends SimpleCalciteSchema {
+
+  protected SchemaConfig schemaConfig;
+  protected StoragePluginRegistry storages;
```
paul-rogers: Seems that storages is never initialized or used except in the accessor.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicSchema.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.drill.exec.store.StoragePluginRegistry;
+
+
+/**
+ * Unlike SimpleCalciteSchema, DynamicSchema could have an empty or partial schemaMap, but it could maintain a map of
+ * name->SchemaFactory, and only register schema when the corresponsdent name is requested.
+ */
+public class DynamicSchema extends SimpleCalciteSchema {
+
+  protected SchemaConfig schemaConfig;
```
paul-rogers: Seems the schemaConfig is never set or used.

exec/java-exec/src/main/java/org/apache/drill/exec/store/SchemaTreeProvider.java
```diff
@@ -20,12 +20,13 @@
 import java.io.IOException;
 import java.util.List;
 
-import org.apache.calcite.jdbc.SimpleCalciteSchema;
+//import org.apache.calcite.jdbc.SimpleCalciteSchema;
```
paul-rogers: Remove

exec/java-exec/src/main/java/org/apache/drill/exec/store/SchemaTreeProvider.java
```diff
@@ -105,12 +106,36 @@ public SchemaPlus createRootSchema(final String userName, final SchemaConfigInfo
...
+
+  /**
+   * Return full root schema with schema owner as the given user.
+   *
+   * @param userName Name of the user who is accessing the storage sources.
+   * @param provider {@link SchemaConfigInfoProvider} instance
+   * @return Root of the schema tree.
+   */
+  public SchemaPlus createFullRootSchema(final String userName, final SchemaConfigInfoProvider provider) {
+    final String schemaUser = isImpersonationEnabled ? userName : ImpersonationUtil.getProcessUserName();
```
paul-rogers: Should this be factored out somewhere? Seems this magic stanza will be needed in many places: better to have one copy than several. Maybe a method in `ImpersonationUtil`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemSchemaFactory.java
```diff
@@ -52,9 +55,20 @@
 
   private List<WorkspaceSchemaFactory> factories;
   private String schemaName;
+  protected FileSystemPlugin plugin;
 
   public FileSystemSchemaFactory(String schemaName, List<WorkspaceSchemaFactory> factories) {
     super();
+    if (factories.size() > 0 ) {
+      this.plugin = factories.get(0).getPlugin();
```
paul-rogers: A comment would be helpful: what is special about plugin 0 that it should be the default? Is plugin 0 the default of some sort?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemSchemaFactory.java
```diff
@@ -52,9 +55,20 @@
...
+    }
+    this.schemaName = schemaName;
+    this.factories = factories;
+  }
+
+  public FileSystemSchemaFactory(FileSystemPlugin plugin, String schemaName, List<WorkspaceSchemaFactory> factories) {
+    super();
```
paul-rogers: Omit; Java does this by default.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemSchemaFactory.java
```diff
@@ -73,9 +87,10 @@ public void registerSchemas(SchemaConfig schemaConfig, SchemaPlus parent) throws
 
     public FileSystemSchema(String name, SchemaConfig schemaConfig) throws IOException {
       super(ImmutableList.<String>of(), name);
+      final DrillFileSystem fs = ImpersonationUtil.createFileSystem(schemaConfig.getUserName(), plugin.getFsConf());
```
paul-rogers: Comment to explain what's happening here?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemSchemaFactory.java
```diff
@@ -73,9 +87,10 @@ public void registerSchemas(SchemaConfig schemaConfig, SchemaPlus parent) throws
...
       for(WorkspaceSchemaFactory f :  factories){
-        if (f.accessible(schemaConfig.getUserName())) {
-          WorkspaceSchema s = f.createSchema(getSchemaPath(), schemaConfig);
+        WorkspaceSchema s = f.createSchema(getSchemaPath(), schemaConfig, fs);
+        if ( s != null) {
```
paul-rogers: Here we iterate over a list of workspace schema factories. For each, we resolve a schemaConfig against the file system.

Under what situations would we have multiple factories? Selecting from two distinct storage plugins?

Calcite tends to resolve the same things over and over. Will this method be called multiple times?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -150,14 +152,30 @@ public WorkspaceSchemaFactory(
...
-      // (see https://issues.apache.org/jira/browse/HDFS-6570). Drill currently relies on Hadoop 2.5.0 version of
-      // FileClient. TODO: Update this when DRILL-3749 is fixed.
-      fs.listStatus(wsPath);
+      /**
+       * For Windows local file system, fs.access ends up using DeprecatedRawLocalFileStatus which has
+       * TrustedInstaller as owner, and a member of Administrators group could not satisfy the permission.
+       * In this case, we will still use method listStatus.
+       * In other cases, we use access method since it is cheaper.
+       */
+      if (SystemUtils.IS_OS_WINDOWS && fs.getUri().getScheme().equalsIgnoreCase("file")) {
```
paul-rogers: HDFS probably defines a constant for "file". Should we reference that?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -175,6 +193,21 @@ public WorkspaceSchema createSchema(List<String> parentSchemaPath, SchemaConfig
     return new WorkspaceSchema(parentSchemaPath, schemaName, schemaConfig);
   }
 
+  public WorkspaceSchema createSchema(List<String> parentSchemaPath, SchemaConfig schemaConfig, DrillFileSystem fs) throws IOException {
+    if (!accessible(fs)) {
```
paul-rogers: Is returning null sufficient to tell the user that they don't have permission to do this operation?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -532,7 +572,10 @@ public boolean isMutable() {
     }
 
     public DrillFileSystem getFS() {
-      return fs;
+      if (this.fs == null) {
+        this.fs = ImpersonationUtil.createFileSystem(schemaConfig.getUserName(), fsConf);
+      }
+      return this.fs;
```
paul-rogers: No need for `this.fs`, just `fs` will do.

paul-rogers: This class caches the file system, which is good. The other classes in this PR do not; they create the fs as needed.

Does Calcite allow some kind of session state in which we can cache the fs for the query (plan) rather than creating it on the fly each time we need it?

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -229,7 +229,7 @@ public DrillbitContext getDrillbitContext() {
     return context;
   }
 
-  public SchemaPlus getRootSchema() {
+  public SchemaPlus getFullRootSchema() {
```
paul-rogers: Comment to explain what a "full root schema" is? Apparently, this is both the plugin config name and workspace combined?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemSchemaFactory.java
chunhui-shi: 'factories' is from storage plugin, so it will be identical when we don't update this storage plugin. This FileSystemSchema constructor will be called only once for a query if this FileSystem storage plugin is needed for this query.

exec/java-exec/src/main/java/org/apache/drill/exec/store/SchemaTreeProvider.java
chunhui-shi: not that many places, and need to pass in isImpersonationEnabled and userName if this line became a standalone method. will keep it as is for now.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
chunhui-shi: FileSystem in hdfs has a constant DEFAULT_FS "file:///", for now I will define our own.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
chunhui-shi: plugin name in drill is case sensitive.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
chunhui-shi: returning null then user could not even list this workspace, so they don't know the existence of this workspace at all. I think that is a good access control practice. 

If users expect to see a workspace but could not see it, then they need to figure out why by themselves.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -373,12 +402,12 @@ public String toString() {
...
     private final SchemaConfig schemaConfig;
-    private final DrillFileSystem fs;
+    private DrillFileSystem fs;
 
-    public WorkspaceSchema(List<String> parentSchemaPath, String wsName, SchemaConfig schemaConfig) throws IOException {
+    public WorkspaceSchema(List<String> parentSchemaPath, String wsName, SchemaConfig schemaConfig, DrillFileSystem fs) throws IOException {
       super(parentSchemaPath, wsName);
       this.schemaConfig = schemaConfig;
-      this.fs = ImpersonationUtil.createFileSystem(schemaConfig.getUserName(), fsConf);
+      this.fs = fs;
```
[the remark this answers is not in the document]
chunhui-shi: Now we pass in fs instead creating from inside of WorkspaceSchema.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -150,14 +152,30 @@ public WorkspaceSchemaFactory(
...
-      // (see https://issues.apache.org/jira/browse/HDFS-6570). Drill currently relies on Hadoop 2.5.0 version of
-      // FileClient. TODO: Update this when DRILL-3749 is fixed.
-      fs.listStatus(wsPath);
+      /**
+       * For Windows local file system, fs.access ends up using DeprecatedRawLocalFileStatus which has
+       * TrustedInstaller as owner, and a member of Administrators group could not satisfy the permission.
+       * In this case, we will still use method listStatus.
+       * In other cases, we use access method since it is cheaper.
+       */
+      if (SystemUtils.IS_OS_WINDOWS && fs.getUri().getScheme().equalsIgnoreCase(FileSystemSchemaFactory.LOCAL_FS_SCHEME)) {
```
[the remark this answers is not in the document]
chunhui-shi: Yes. it was tested in windows unit tests.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,140 @@
...
+          thisPlus.add(wrapper.getName(), wrapper);
+        }
+      }
+    } catch(ExecutionSetupException | IOException ex) {
+      logger.warn("Failed to load schema for \"" + schemaName + "\"!", ex);
+    }
+  }
+
+  static class RootSchema extends AbstractSchema {
+    @Override public Expression getExpression(SchemaPlus parentSchema,
```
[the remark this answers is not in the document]
chunhui-shi: This is copied from the RootSchema used in SimpleCalciteSchema which class is not public. getExpression is used in Calcite code not in our code.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DynamicRootSchema.java
```diff
@@ -0,0 +1,140 @@
...
+      String[] paths = schemaName.split("\\.");
+      if (paths.length == 2) {
+        plugin = getSchemaFactories().getPlugin(paths[0]);
+        if (plugin == null) {
+          return;
+        }
+
+        // we could find storage plugin for first part(e.g. 'dfs') of schemaName (e.g. 'dfs.tmp')
+        // register schema for this storage plugin to 'this'.
+        plugin.registerSchemas(schemaConfig, thisPlus);
```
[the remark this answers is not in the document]
chunhui-shi: we get to this place only when that split got an array of length 2 and after we check nullability of plugin, so 'plugin' in this line should not be null if this is your concern.

arina-ielchiieva:
+1

PR #1033, opened 2017-11-11
DRILL-5952: Implement "CREATE TABLE IF NOT EXISTS"

prasadns14:
1) Added <IF> <NOT> <EXISTS> support for CREATE TABLE
2) Added unit tests for the same

@paul-rogers please review

### DIFF
--- a/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
+++ b/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
@@ -184,14 +184,14 @@ SqlNode SqlCreateOrReplaceView() :
 }
 {
     <CREATE> { pos = getPos(); }
-    [ <OR> <REPLACE> { createViewType = "ORREPLACE"; } ]
+    [ <OR> <REPLACE> { createViewType = "OR_REPLACE"; } ]
     <VIEW>
     [
         <IF> <NOT> <EXISTS> {
-            if (createViewType == "ORREPLACE") {
+            if (createViewType == "OR_REPLACE") {
                 throw new ParseException("Create view statement cannot have both <OR REPLACE> and <IF NOT EXISTS> clause");
             }
-            createViewType = "IFNOTEXISTS";
+            createViewType = "IF_NOT_EXISTS";
         }
     ]
     viewName = CompoundIdentifier()
--- a/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
+++ b/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
@@ -172,26 +172,34 @@ SqlNodeList ParseRequiredFieldList(String relType) :
 
 /**
  * Parses a create view or replace existing view statement.
- *   CREATE [OR REPLACE] VIEW view_name [ (field1, field2 ...) ] AS select_statement
+ *   CREATE { [OR REPLACE] VIEW | VIEW [IF NOT EXISTS] | VIEW } view_name [ (field1, field2 ...) ] AS select_statement
  */
 SqlNode SqlCreateOrReplaceView() :
 {
     SqlParserPos pos;
-    boolean replaceView = false;
     SqlIdentifier viewName;
     SqlNode query;
     SqlNodeList fieldList;
+    String createViewType = "0";
 }
 {
     <CREATE> { pos = getPos(); }
-    [ <OR> <REPLACE> { replaceView = true; } ]
+    [ <OR> <REPLACE> { createViewType = "1"; } ]
     <VIEW>
+    [
+        <IF> <NOT> <EXISTS> {
+            if (createViewType == "1") {
+                throw new ParseException("Create view statement cannot have both <OR REPLACE> and <IF NOT EXISTS> clause");
+            }
+            createViewType = "2";
+        }
+    ]
     viewName = CompoundIdentifier()
     fieldList = ParseOptionalFieldList("View")
     <AS>
     query = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
     {
-        return new SqlCreateView(pos, viewName, fieldList, query, replaceView);
+        return new SqlCreateView(pos, viewName, fieldList, query, SqlLiteral.createExactNumeric(createViewType, getPos()));
     }
 }
 
@@ -215,11 +223,12 @@ SqlNode SqlDropView() :
 
 /**
  * Parses a CTAS or CTTAS statement.
- * CREATE [TEMPORARY] TABLE tblname [ (field1, field2, ...) ] AS select_statement.
+ * CREATE [TEMPORARY] TABLE [IF NOT EXISTS] tblname [ (field1, field2, ...) ] AS select_statement.
  */
 SqlNode SqlCreateTable() :
 {
     SqlParserPos pos;
+    boolean tableNonExistenceCheck = false;
     SqlIdentifier tblName;
     SqlNodeList fieldList;
     SqlNodeList partitionFieldList;
@@ -233,6 +242,7 @@ SqlNode SqlCreateTable() :
     <CREATE> { pos = getPos(); }
     ( <TEMPORARY> { isTemporary = true; } )?
     <TABLE>
+    ( <IF> <NOT> <EXISTS> { tableNonExistenceCheck = true; } )?
     tblName = CompoundIdentifier()
     fieldList = ParseOptionalFieldList("Table")
     (   <PARTITION> <BY>
@@ -241,8 +251,9 @@ SqlNode SqlCreateTable() :
     <AS>
     query = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
     {
-        return new SqlCreateTable(pos,tblName, fieldList, partitionFieldList, query,
-                                    SqlLiteral.createBoolean(isTemporary, getPos()));
+        return new SqlCreateTable(pos, tblName, fieldList, partitionFieldList, query,
+                                    SqlLiteral.createBoolean(isTemporary, getPos()),
+                                    SqlLiteral.createBoolean(tableNonExistenceCheck, getPos()));
     }
 }
 
--- a/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
+++ b/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
@@ -180,18 +180,18 @@ SqlNode SqlCreateOrReplaceView() :
     SqlIdentifier viewName;
     SqlNode query;
     SqlNodeList fieldList;
-    String createViewType = "0";
+    String createViewType = "SIMPLE";
 }
 {
     <CREATE> { pos = getPos(); }
-    [ <OR> <REPLACE> { createViewType = "1"; } ]
+    [ <OR> <REPLACE> { createViewType = "ORREPLACE"; } ]
     <VIEW>
     [
         <IF> <NOT> <EXISTS> {
-            if (createViewType == "1") {
+            if (createViewType == "ORREPLACE") {
                 throw new ParseException("Create view statement cannot have both <OR REPLACE> and <IF NOT EXISTS> clause");
             }
-            createViewType = "2";
+            createViewType = "IFNOTEXISTS";
         }
     ]
     viewName = CompoundIdentifier()
@@ -199,7 +199,7 @@ SqlNode SqlCreateOrReplaceView() :
     <AS>
     query = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
     {
-        return new SqlCreateView(pos, viewName, fieldList, query, SqlLiteral.createExactNumeric(createViewType, getPos()));
+        return new SqlCreateView(pos, viewName, fieldList, query, SqlLiteral.createCharString(createViewType, getPos()));
     }
 }
 
@@ -228,12 +228,12 @@ SqlNode SqlDropView() :
 SqlNode SqlCreateTable() :
 {
     SqlParserPos pos;
-    boolean tableNonExistenceCheck = false;
     SqlIdentifier tblName;
     SqlNodeList fieldList;
     SqlNodeList partitionFieldList;
     SqlNode query;
     boolean isTemporary = false;
+    boolean tableNonExistenceCheck = false;
 }
 {
     {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
@@ -382,6 +382,9 @@ public class SqlConverter {
    * @return The sql with a ^ character under the error
    */
   static String formatSQLParsingError(String sql, SqlParserPos pos) {
+    if (pos == null) {
+      return sql;
+    }
     StringBuilder sb = new StringBuilder();
     String[] lines = sql.split("\n");
     for (int i = 0; i < lines.length; i++) {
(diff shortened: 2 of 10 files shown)


exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -215,11 +215,12 @@ SqlNode SqlDropView() :
 
 /**
  * Parses a CTAS or CTTAS statement.
- * CREATE [TEMPORARY] TABLE tblname [ (field1, field2, ...) ] AS select_statement.
+ * CREATE [TEMPORARY] [IF NOT EXISTS] TABLE tblname [ (field1, field2, ...) ] AS select_statement.
```
Agirish: [IF NOT EXISTS] should be after TABLE?

prasadns14: Fixed it

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -313,6 +313,59 @@ public void createTableWithCustomUmask() throws Exception {
     }
   }
 
+  @Test // DRILL-5951
+  public void testCreateTableIfNotExistsWhenTableWithSameNameAlreadyExists() throws Exception{
+    final String newTblName = "createTableIfNotExistsWhenATableWithSameNameAlreadyExists";
+
+    try {
+      test(String.format("CREATE TABLE %s.%s AS SELECT * from cp.`region.json`", TEMP_SCHEMA, newTblName));
```
arina-ielchiieva: `test` method already contains `String.format`, you can use `test("CREATE TABLE %s.%s AS SELECT * from cp.`region.json`", TEMP_SCHEMA, newTblName)`

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -313,6 +313,59 @@ public void createTableWithCustomUmask() throws Exception {
...
+        String.format("CREATE TABLE IF NOT EXISTS %s.%s AS SELECT * FROM cp.`employee.json`", TEMP_SCHEMA, newTblName);
+
+      testBuilder()
+        .sqlQuery(ctasQuery)
+        .unOrdered()
+        .baselineColumns("ok", "summary")
+        .baselineValues(false, String.format("A table or view with given name [%s] already exists in schema [%s]", newTblName, TEMP_SCHEMA))
+        .go();
+    } finally {
+      test(String.format("DROP TABLE %s.%s", TEMP_SCHEMA, newTblName));
```
arina-ielchiieva: Please use here `drop table if exists`.  Since test may fail before creates the table and we don;t here one more exception not related to the problem.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -313,6 +313,59 @@ public void createTableWithCustomUmask() throws Exception {
...
+      test(String.format("DROP VIEW %s.%s", TEMP_SCHEMA, newTblName));
+    }
+  }
+
+  @Test // DRILL-5951
+  public void testCreateTableIfNotExistsWhenTableWithSameNameDoesNotExist() throws Exception{
+    final String newTblName = "createTableIfNotExistsWhenATableWithSameNameDoesNotExist";
+
+    try {
+      test(String.format("CREATE TABLE IF NOT EXISTS %s.%s AS SELECT * FROM cp.`employee.json`", TEMP_SCHEMA, newTblName));
```
arina-ielchiieva: You might want to check baseline here as well but only for `true` to confirm that table was actually created.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -241,8 +243,8 @@ SqlNode SqlCreateTable() :
     <AS>
     query = OrderedQueryOrExpr(ExprContext.ACCEPT_QUERY)
     {
-        return new SqlCreateTable(pos,tblName, fieldList, partitionFieldList, query,
-                                    SqlLiteral.createBoolean(isTemporary, getPos()));
+        return new SqlCreateTable(pos, SqlLiteral.createBoolean(tableNonExistenceCheck, getPos()), tblName, fieldList,
```
arina-ielchiieva: Not sure if it is a good practice to add new parameter in the beginning of constructor rather in the end.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/CreateTableHandler.java
```diff
@@ -83,7 +84,20 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ValidationException, RelConv
     final DrillConfig drillConfig = context.getConfig();
     final AbstractSchema drillSchema = resolveSchema(sqlCreateTable, config.getConverter().getDefaultSchema(), drillConfig);
 
-    checkDuplicatedObjectExistence(drillSchema, originalTableName, drillConfig, context.getSession());
+    String schemaPath = drillSchema.getFullSchemaName();
+
+    // Check duplicate object existence
+    boolean isTemporaryTable = context.getSession().isTemporaryTable(drillSchema, drillConfig, originalTableName);
```
arina-ielchiieva: 1. Please add unit tests for temp tables as well in `TestCTTAS` class.
2. Should we consider similar functionality for views?

prasadns14: 1) Added unit tests for temp tables
2) Implemented for views as well

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
prasadns14: fixed it

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
prasadns14: changed to use drop table if exists

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
prasadns14: if successful returns fragment & number of rows written. as long as no exception is thrown we can be sure the table is created. in sync with other test cases in TestCTAS

prasadns14:
@arina-ielchiieva, please review

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -215,11 +223,12 @@ SqlNode SqlDropView() :
 
 /**
  * Parses a CTAS or CTTAS statement.
- * CREATE [TEMPORARY] TABLE tblname [ (field1, field2, ...) ] AS select_statement.
+ * CREATE [TEMPORARY] TABLE [IF NOT EXISTS] tblname [ (field1, field2, ...) ] AS select_statement.
  */
 SqlNode SqlCreateTable() :
 {
     SqlParserPos pos;
+    boolean tableNonExistenceCheck = false;
```
arina-ielchiieva: Not critical but I am in favor of adding new parameters in the end. Of course if logically it is not required.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -172,26 +172,34 @@ SqlNodeList ParseRequiredFieldList(String relType) :
...
+ *   CREATE { [OR REPLACE] VIEW | VIEW [IF NOT EXISTS] | VIEW } view_name [ (field1, field2 ...) ] AS select_statement
  */
 SqlNode SqlCreateOrReplaceView() :
 {
     SqlParserPos pos;
-    boolean replaceView = false;
     SqlIdentifier viewName;
     SqlNode query;
     SqlNodeList fieldList;
+    String createViewType = "0";
```
arina-ielchiieva: Maybe use enum rather that 0, 1, 2 in string equivalent.  First it would be much clearer, second you won't need to add comments explaining what these numbers mean in `ViewHandler` class. Enum definition can be stored in `SqlCreateView.class`, to make it work in parser - just add imports in `vv_imports.ftl`.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -254,6 +255,66 @@ public void createTableWithCustomUmask() throws Exception {
...
+        String.format("CREATE TABLE IF NOT EXISTS %s.%s AS SELECT * FROM cp.`employee.json`", DFS_TMP_SCHEMA, newTblName);
+
+      testBuilder()
+        .sqlQuery(ctasQuery)
+        .unOrdered()
+        .baselineColumns("ok", "summary")
+        .baselineValues(false, String.format("A table or view with given name [%s] already exists in schema [%s]", newTblName, DFS_TMP_SCHEMA))
+        .go();
+    } finally {
+      test(String.format("DROP TABLE IF EXISTS %s.%s", DFS_TMP_SCHEMA, newTblName));
```
arina-ielchiieva: Please remove `String.format` here and other unit tests, `test()` method already incorporates it.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java
```diff
@@ -311,6 +311,75 @@ public void createViewWhenViewAlreadyExists() throws Exception {
...
   }
 
+  @Test // DRILL-5952
+  public void createViewIfNotExistsWhenTableAlreadyExists() throws Exception {
+    final String tableName = generateViewName();
+
+    try {
+      final String tableDef = "SELECT region_id, sales_city FROM cp.`region.json` ORDER BY `region_id` LIMIT 2";
+
+      test("CREATE TABLE %s.%s as %s", DFS_TMP_SCHEMA, tableName, tableDef);
```
arina-ielchiieva: There is not need to create table before creating the view just use select statement.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ViewHandler.java
```diff
@@ -78,47 +78,58 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ValidationException, RelConv
       final View view = new View(newViewName, viewSql, newViewRelNode.getRowType(),
           SchemaUtilites.getSchemaPathAsList(defaultSchema));
 
-      validateViewCreationPossibility(drillSchema, createView, context);
+      final String schemaPath = drillSchema.getFullSchemaName();
```
arina-ielchiieva: It's a good practice in Java to factor out large code pieces into separate methods. I did this before in this class but your changes incorporated all logic back... Maybe there is a way to keep this logic separately here and in CreateTableHandler.class?

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java
```diff
@@ -311,6 +311,75 @@ public void createViewWhenViewAlreadyExists() throws Exception {
...
+    final String viewName = generateViewName();
+
+    final String viewDef = "SELECT region_id, sales_city FROM cp.`region.json`";
+
+    // Try to create the view with both <or replace> and <if not exists> clause.
+    final String createViewSql = String.format("CREATE OR REPLACE VIEW IF NOT EXISTS %s.`%s` AS %s", DFS_TMP_SCHEMA, viewName, viewDef);
+
+    errorMsgTestHelper(createViewSql, "Create view statement cannot have both <OR REPLACE> and <IF NOT EXISTS> clause");
+  }
+
```
arina-ielchiieva: Please add unit test to check if view is created with `if not exists` clause.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -382,6 +382,9 @@ public int getInSubqueryThreshold() {
    * @return The sql with a ^ character under the error
    */
   static String formatSQLParsingError(String sql, SqlParserPos pos) {
+    if (pos == null) {
```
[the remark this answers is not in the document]
prasadns14: we throw parseException with custom messages at a couple of locations (parserImpls.ftl). Ex 
1) when the query contains '*' as one of the fields in create table/view statement 
2) when the create view query contains both replace and if-not-exists
unlike a parse error, the above two instances don't have the SqlParserPos set and will throw NPE

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ViewHandler.java
prasadns14: Separated the logic as before

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
prasadns14: Done

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
prasadns14: Moved the new parameter to the end

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java
prasadns14: Done

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java
prasadns14: This test specifically tests the behavior when we try to create a view with the same name as an existing table in the intended schema.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
prasadns14: SqlLiteral doesn't support creating enum objects. So, I am forced to use string in parserImpls.ftl but I have now defined an enum in SqlCreateView.class

prasadns14:
@arina-ielchiieva, please review

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlCreateView.java
```diff
@@ -47,20 +47,19 @@ public SqlCall createCall(SqlLiteral functionQualifier, SqlParserPos pos, SqlNod
...
   private SqlNodeList fieldList;
   private SqlNode query;
-  private boolean replaceView;
+  private SqlLiteral createViewType;
 
-  public SqlCreateView(SqlParserPos pos, SqlIdentifier viewName, SqlNodeList fieldList,
-      SqlNode query, SqlLiteral replaceView) {
-    this(pos, viewName, fieldList, query, replaceView.booleanValue());
+  public enum SqlCreateViewType {
+    SIMPLE, ORREPLACE, IFNOTEXISTS
```
arina-ielchiieva: Please use underscore to separate words: `OR_REPLACE`, `IF_NOT_EXISTS`.

prasadns14: done

prasadns14:
@arina-ielchiieva, please review

arina-ielchiieva:
+1, LGTM.

PR #1034, opened 2017-11-14
DRILL-5960: Adding asGeoJSON function

ChrisSandison:
Fixes #DRILL-5960

### DIFF
--- a/contrib/gis/src/main/java/org/apache/drill/exec/expr/fn/impl/gis/STAsGeoJSON.java
+++ b/contrib/gis/src/main/java/org/apache/drill/exec/expr/fn/impl/gis/STAsGeoJSON.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


contrib/gis/src/main/java/org/apache/drill/exec/expr/fn/impl/gis/STAsGeoJson.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.drill.exec.expr.DrillSimpleFunc;
+import org.apache.drill.exec.expr.annotations.FunctionTemplate;
+import org.apache.drill.exec.expr.annotations.Output;
+import org.apache.drill.exec.expr.annotations.Param;
+import org.apache.drill.exec.expr.holders.VarBinaryHolder;
+import org.apache.drill.exec.expr.holders.VarCharHolder;
+
+import io.netty.buffer.DrillBuf;
+
+@FunctionTemplate(name = "st_as_geo_json", scope = FunctionTemplate.FunctionScope.SIMPLE,
```
arina-ielchiieva: @ChrisSandison since you have changed function names here and did not not in unit tests below, they will fail. Please note, functions are cached when Drill is built so to make sure your changes took affect you need to build drill first and then run unit tests).

It seems I saw comment that suggested to keep function name as with compliance with previously created functions like (`st_astext`) which at some point makes sense. It looked in Calcite and they have geographical functions with the same naming convention as well, so I guess we should revert the previous name.

ChrisSandison: @arina-ielchiieva Ah, gotcha. I've been running `mvn install -e -rf :drill-gis` after changes to save on time. I'll run clean builds from now on

arina-ielchiieva:
Looks good, please squash the commits into one with commit message: `DRILL-5960: Added asGeoJSON function`.

arina-ielchiieva:
+1

PR #1037, opened 2017-11-15
DRILL-5968: Add support for empty service_host user property

superbstreak:


contrib/native/client/src/include/drill/userProperties.hpp
```diff
@@ -28,6 +28,20 @@ class DECLSPEC_DRILL_CLIENT DrillUserProperties{
...
+        
+        /// @brief Insert or update property value associate with the property key if the value is 
+        /// empty or key is undefined. This function is useful for setting default property.
+        /// 
+        /// @param in_propName              The property name.
+        /// @param in_propValue             The property value.
+        void SetDefaultProperty(const std::string& in_propName, const std::string& in_propValue)
+        {
+            // If the element value is empty (newly inserted or user defined), then update.
+            if (m_properties[in_propName].empty())
```
parthchandra: I would have preferred to use isPropSet to see if the property value is set. Operator[] for a map is intended only to add new values. I believe this might add a new empty element to the map.

contrib/native/client/src/include/drill/userProperties.hpp
```diff
@@ -28,6 +28,20 @@ class DECLSPEC_DRILL_CLIENT DrillUserProperties{
         static const std::map<std::string, uint32_t> USER_PROPERTIES;
 
         DrillUserProperties(){};
+        
+        /// @brief Insert or update property value associate with the property key if the value is 
+        /// empty or key is undefined. This function is useful for setting default property.
+        /// 
+        /// @param in_propName              The property name.
+        /// @param in_propValue             The property value.
+        void SetDefaultProperty(const std::string& in_propName, const std::string& in_propValue)
```
parthchandra: Follow naming convention setDefaultProperty instead of SetDefaultProperty

contrib/native/client/src/include/drill/userProperties.hpp
superbstreak: If the value doesn't exist, the original implementation with the setProperty(2) would also insert a new key. Not sure if that was intended or not.

contrib/native/client/src/include/drill/userProperties.hpp
superbstreak: gotcha thanks!

contrib/native/client/src/include/drill/userProperties.hpp
```diff
@@ -28,6 +28,18 @@ class DECLSPEC_DRILL_CLIENT DrillUserProperties{
...
         DrillUserProperties(){};
+        
+        /// @brief Update the property value associate with the property key if the value is 
+        /// empty.
+        /// 
+        /// @param in_propName              The property name.
+        /// @param in_propValue             The property value.
+        void setPropertyIfEmpty(const std::string& in_propName, const std::string& in_propValue){
+            // If the element value is empty, then update.
+            if (isPropSet(in_propName) && m_properties[in_propName].empty()){
```
parthchandra: Better check would be 
`if ( !isPropSet(in_propName) || m_properties[in_propName].empty() )`

superbstreak: certainly

superbstreak:
@parthchandra  We should get this in for 1.12?

parthchandra:
LGTM. (Yes, let's merge this in before 1.12.)

superbstreak:
Thanks, @parthchandra !

PR #1041, opened 2017-11-17
DRILL-5961: For long running queries (> 10 min) Drill may raise FragmentSetupException for completed/cancelled fragments

vrozov:


priteshm:
@parthchandra or @adityakishore since you'll have made changes to these files before, can you review the changes here?

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -277,7 +277,9 @@ public void startFragmentPendingRemote(final FragmentManager fragmentManager) {
         @Override
         protected void cleanup() {
           runningFragments.remove(fragmentHandle);
-          workBus.removeFragmentManager(fragmentHandle);
+          if (!fragmentManager.isCancelled()) {
+            workBus.removeFragmentManager(fragmentHandle, false);
```
parthchandra: Not sure why you don't want to cancel here.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/control/WorkEventBus.java
```diff
@@ -74,83 +70,42 @@ public void statusUpdate(final FragmentStatus status) {
...
       }
-
-      manager.cancel();
-      recentlyFinishedFragments.put(handle, 1);
-      managers.remove(handle);
       return true;
+    } else if (logger.isWarnEnabled()) {
+      logger.warn("Fragment {} manager is not found in the work bus.", QueryIdHelper.getQueryIdentifier(handle));
     }
+    return false;
```
parthchandra: Seems like this is redundant. Doesn't seem like we ever return false, and the callers never seem to check.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
vrozov: If cleanup is not called as a result of FragmentManager cancellation, it is part of the regular cleanup after run is complete.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/control/WorkEventBus.java
vrozov: ControlMessageHandler checks the result on line 196.

PR #1042, opened 2017-11-19
DRILL-5261: Expose REST endpoint in zookeeper

xhochy:


xhochy:
This still needs the protobufs to be re-compiled for C++, I would do this once someone signals me that I have taken the correct approach.

kkhatua:
@paul-rogers , since you've responded to the original JIRA, perhaps you can review this for @xhochy

kkhatua:
@xhochy I recently had a PR #1203  for DRILL-6289 and realized your PR's solution was a subset of that. I've already incorporated changes to the C++ client's protobuf as well. If it addresses the requirements you had for this JIRA, can you close the PR? If there is something missing, feel free to submit a new PR with the missing feature.

xhochy:
@kkhatua Thanks, I'll have a look at that PR.

PR #1043, opened 2017-11-20
DRILL-5981: Add Syntax Highlighting and Error Checking to Storage Plugin Config

cgivre:
This PR adds syntax highlighting to the storage plugin configuration page in the Drill web UI, as shown below.
screen shot 2017-11-20 at 16 05 27

In addition, and the real reason I'd like to see this included is that it also incorporates error checking, again as shown below:

screen shot 2017-11-20 at 16 05 39

I know there is some objection about including javascript libraries in Drill.  However, I work on a closed network and it is difficult if not impossible to build Drill from source if the libraries are not included.   I removed all JavaScript files except the bare minimum.


arina-ielchiieva:
@cgivre looks really nice, thanks for making the change! @kkhatua could you please take a look at this PR?

kkhatua:
Very nice, @cgivre ! I've been looking to do something similar for the SQL side as well, but with autocomplete support. Let's talk about that separately. We just need to make sure that there are no license/usage issues. Would be nice if we can leverage this into the SQL Editor as well. Let's talk over this offline.

I know I had a discussion recently with someone regarding validation of the JSON for the storage plugin, but that will be a stretch. Also, it seems like the library can recognize single line comments ( // ), which (I believe) is supported by Drill JSON config parser.

Can you pick a theme that helps visibly pop out the colors more than it currently is? Crimson or Eclipse themes look better, helping visualize.

Also, if the 'src-min-noconflict' is primarily to support the ace libraries and you don't want to risk renaming the library files, it's good to give it a more meaningful name (indicating that it contains AceJS libraries). 

Otherwise, LGTM
+1

kkhatua:
Oh.. just for a heads up (I guess you already knew this), I zeroed in on the 2 themes based on copy-pasting a sample storage plugin into this:
https://ace.c9.io/build/kitchen-sink.html

kkhatua:
@cgivre 
Just edited the last comment to remove the +1 [ Learnt that I should wait till there is a response to my asks first! :) ] . 
To summarize, here are my 3 asks:
1. Change the theme to **Crimson** or **Eclipse** since that's a more familiar theme and actually helps the colors stand out.
2. Change the `src-min-noconflict` directory to reflect the name of the library, whose files it contains. e.g. `aceJs`
3. Remove any non-mandatory files, e.g. `snippets` directory doesn't seems to be a requirement for the library. This would help in maintaining future updates to the library or debugging.

cgivre:
I made the requested fixes, and attempted to squash the commits into one, but failed on that last step. I actually renamed the js director `ace-code-editor`, which seemed descriptive to me.

kkhatua:
May be @arina-ielchiieva can help resolve squashing into one commit.
LGTM ... thanks for doing this!
+1

arina-ielchiieva:
I will but it's high time to learn how to do it :)

PR #1045, opened 2017-11-21
DRILL-5730 Test Mocking Improvements

ilooner:
## DRILL-5730

- Switched to using the interface for FragmentContext everywhere instead of passing around the concrete class.
- Minor refactoring of FragmentContext public methods
- Switched to using the OptionSet interface throughout the codebase instead of OptionManager
- Renamed **FragmentContext** to **FragmentContextImpl** and renamed **FragmentContextInterface** to **FragmentContext**.
- Removed JMockit from most unit tests in favor of Mockito. Unfortunately it cannot be removed some some of the unit tests which depend on it for mocking private method and static methods (which is functionality only JMockit provides). In the future we need to refactor the code so that these remaining tests can have JMockit removed completely.
- Refactored some tests to use a mock class of FragmentContext
- Some tests were using Mockito and JMockit when there was no need for a mocking framework

## Misc

- Removed commented out code and unused imports
- Removed unnecessary modifiers from methods in interfaces
- Fixed a bug in bootstrapcontext which leaked threads
- Fixed javadoc links that were broken

## Note

The vast majority of unit tests should be runnable from Eclipse except for the handful that still depend on JMockit


ilooner:
@paul-rogers

ilooner:
@paul-rogers have fixed merge conflicts. This is ready for review again.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/BaseOperatorContext.java
```diff
@@ -39,27 +39,29 @@
...
-  public BaseOperatorContext(FragmentContextInterface context, BufferAllocator allocator,
-               PhysicalOperator popConfig) {
+  public BaseOperatorContext(FragmentContext context, BufferAllocator allocator,
+                             PhysicalOperator popConfig,
+                             OperatorStatReceiver stats) {
     this.context = context;
     this.allocator = allocator;
     this.popConfig = popConfig;
     this.manager = new BufferManagerImpl(allocator);
+    this.statsWriter = stats;
```
paul-rogers: This member was removed in a recent commit. Since the test and production versions need different implementation, the derived classes manage the operator stats.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/BaseOperatorContext.java
```diff
@@ -161,16 +163,13 @@ public void close() {
     }
     try {
       if (fs != null) {
-        fs.close();
-        fs = null;
+          fs.close();
```
paul-rogers: Don't we like two space indentations?

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
  */
 package org.apache.drill.exec.ops;
 
+import java.io.IOException;
```
paul-rogers: Github is very confused by this double rename. I presume the new file just has the same contents, but a new name? Used an online diff checker to compare the two interfaces. Turns out there are important differences. Comments below.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
-   * @param fragment Fragment implementation.
-   * @param funcRegistry FunctionImplementationRegistry.
-   * @throws ExecutionSetupException
+   * Wait for ack that all outgoing batches have been sent
    */
-  public FragmentContext(final DrillbitContext dbContext, final PlanFragment fragment,
-      final FunctionImplementationRegistry funcRegistry) throws ExecutionSetupException {
-    this(dbContext, fragment, null, null, funcRegistry);
-  }
+  void waitForSendComplete();
```
paul-rogers: The goal of the interface was to define a self-contained set of services easily used in unit tests. Said another way, code that uses the interface controlled dependencies and can be unit tested without mocks.

Introducing these network-related APIs breaks that assumption. Now, any code can access network services, creating a huge API and requiring extensive mocking.

The suggestion here is to either:

1. Keep network (or server) related services only in the "impl" class (the design as of the commit that introduced this interface.)

2. Create a new "NetworkFragmentContext" that has the isolated service plus the network and server-related services.

In either case, code that uses the original interface can be easily unit tested. Code that uses the "network/server" flavor requires the full Drillbit or extensive mocking.

In general, it will only be exchange senders and receivers that need network services. All "internal" operators should be able to use the constrained version.

(Would sure be great if Drill were a project that valued design documents so all of this would not just be personal opinion or word of mouth...)

Doing the full review (see notes below), a better design might be:

1. Runtime fragment context (used by internal operators, makes it easy to unit test.)
2. Exchange fragment context (exposes network services, used only for exchange send and receive. Quite hard to unit test.)
3. Fragment builder context (used to create a fragment, but never used once the fragment is up and running.)

This three-part split will relieve the kitchen-sink design of the current classes while preserving the goal of the prior commit allow easy unit testing of internal (non-exchange) operators.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
-        .build(logger);
-    } catch(final Throwable e) {
-      throw new ExecutionSetupException("Failure while getting memory allocator for fragment.", e);
-    }
-
-    stats = new FragmentStats(allocator, fragment.getAssignment());
-    bufferManager = new BufferManagerImpl(this.allocator);
-    constantValueHolderCache = Maps.newHashMap();
-  }
+  FunctionImplementationRegistry getFunctionRegistry();
```
paul-rogers: OK to add as long as the registry can be created without ZK and network dependencies. (UDF introduced such dependencies, but, I believe, they can be turned off.)

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
-  public FragmentContext(DrillbitContext dbContext, PlanFragment fragment, UserClientConnection connection,
-      FunctionImplementationRegistry funcRegistry) throws ExecutionSetupException {
-    this(dbContext, fragment, null, connection, funcRegistry);
-  }
+  OptionSet getOptionSet();
 
-  public OptionManager getOptions() {
-    return fragmentOptions;
-  }
+  PhysicalPlanReader getPlanReader();
```
paul-rogers: Only in network version. Actually, probably should not be part of the fragment context itself, but rather part of the mechanism that boots up a fragment. (Once the fragment is running, we won't read the plan again.)

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
 
-  @Override
-  public OptionSet getOptionSet() {
-    return fragmentOptions;
-  }
+  ClusterCoordinator getClusterCoordinator();
```
paul-rogers: Introduces dependencies on cluster coordination; makes it hard to test operators that operate "below" the level of the cluster.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
 
-  public void setBuffers(final IncomingBuffers buffers) {
-    Preconditions.checkArgument(this.buffers == null, "Can only set buffers once.");
-    this.buffers = buffers;
-  }
+  AccountingDataTunnel getDataTunnel(final CoordinationProtos.DrillbitEndpoint endpoint);
```
paul-rogers: Requires tests mock up a network connection just to test, say, sort or project which don't care about network. Move to the "network/cluster" version of this interface.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
 
-  public void setExecutorState(final ExecutorState executorState) {
-    Preconditions.checkArgument(this.executorState == null, "ExecutorState can only be set once.");
-    this.executorState = executorState;
-  }
+  AccountingUserConnection getUserDataTunnel();
```
paul-rogers: Network related.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
 
-  public void fail(final Throwable cause) {
-    executorState.fail(cause);
-  }
+  void setBuffers(final IncomingBuffers buffers);
```
paul-rogers: Network related.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
 
-  public AccountingDataTunnel getDataTunnel(final DrillbitEndpoint endpoint) {
-    AccountingDataTunnel tunnel = tunnels.get(endpoint);
-    if (tunnel == null) {
-      tunnel = new AccountingDataTunnel(context.getDataConnectionsPool().getTunnel(endpoint), sendingAccountor, statusHandler);
-      tunnels.put(endpoint, tunnel);
-    }
-    return tunnel;
-  }
+  CoordinationProtos.DrillbitEndpoint getEndpoint();
```
paul-rogers: Above three are network related.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
```diff
@@ -17,483 +17,216 @@
...
-  }
+  Controller getController();
 
-  public OperatorContext newOperatorContext(PhysicalOperator popConfig, OperatorStats stats)
-      throws OutOfMemoryException {
-    OperatorContextImpl context = new OperatorContextImpl(popConfig, this, stats);
-    contexts.add(context);
-    return context;
-  }
+  OperatorCreatorRegistry getOperatorCreatorRegistry();
```
paul-rogers: Needed only when creating operators, not when executing a fragment. So, network/cluster related.

Rather than continue to tag each item; please review the original interface class where I did this analysis to decide which methods should be in the runtime interface and which are network/cluster/server related (and so I left them in the implementation class.)

Since we have no design doc, perhaps we can at least document these decisions and concepts in the class header comments. (I suppose I should have done that; rereading the comment it clearly lacks this level of detail.)

common/src/main/java/org/apache/drill/common/config/DrillConfig.java
```diff
@@ -45,13 +45,11 @@
...
   private final ImmutableList<String> startupArguments;
 
-  public static final boolean ON_OSX = System.getProperty("os.name").contains("OS X");
-
   @SuppressWarnings("restriction")
   private static final long MAX_DIRECT_MEMORY = sun.misc.VM.maxDirectMemory();
 
   @VisibleForTesting
-  public DrillConfig(Config config, boolean enableServerConfigs) {
+  public DrillConfig(Config config) {
```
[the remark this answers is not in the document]
ilooner: The flag is actually an unused argument. Maybe it did something at one point but the code that used it is long gone.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
```diff
@@ -117,7 +116,7 @@ public MaprDBJsonRecordReader(MapRDBSubScanSpec subScanSpec,
 
     disableCountOptimization = formatPluginConfig.disableCountOptimization();
     setColumns(projectedColumns);
-    unionEnabled = context.getOptions().getOption(ExecConstants.ENABLE_UNION_TYPE);
+    unionEnabled = context.getOptionSet().getOption(ExecConstants.ENABLE_UNION_TYPE);
```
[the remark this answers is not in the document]
ilooner: Done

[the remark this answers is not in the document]
ilooner: Sounds good

contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcBatchCreator.java
```diff
@@ -33,10 +32,10 @@
 public class JdbcBatchCreator implements BatchCreator<JdbcSubScan> {
   @Override
   public ScanBatch getBatch(FragmentContext context, JdbcSubScan config,
-      List<RecordBatch> children) throws ExecutionSetupException {
+                            List<RecordBatch> children) throws ExecutionSetupException {
     Preconditions.checkArgument(children.isEmpty());
     JdbcStoragePlugin plugin = config.getPlugin();
-    RecordReader reader = new JdbcRecordReader(context, plugin.getSource(), config.getSql(), plugin.getName());
+    RecordReader reader = new JdbcRecordReader(plugin.getSource(), config.getSql(), plugin.getName());
```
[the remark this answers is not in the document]
ilooner: Which file did I remove the POP config from? I can't seem to find it.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaRecordReader.java
```diff
@@ -66,13 +66,12 @@
...
-    OptionManager options = context.getOptions();
-    this.unionEnabled = options.getOption(ExecConstants.ENABLE_UNION_TYPE);
-    this.kafkaMsgReader = options.getOption(ExecConstants.KAFKA_RECORD_READER).string_val;
-    this.kafkaPollTimeOut = options.getOption(ExecConstants.KAFKA_POLL_TIMEOUT).num_val;
+    final OptionSet optionSet = context.getOptionSet();
+    this.enableAllTextMode = optionSet.getOption(ExecConstants.KAFKA_ALL_TEXT_MODE).bool_val;
+    this.readNumbersAsDouble = optionSet.getOption(ExecConstants.KAFKA_READER_READ_NUMBERS_AS_DOUBLE).bool_val;
+    this.unionEnabled = optionSet.getOption(ExecConstants.ENABLE_UNION_TYPE);
+    this.kafkaMsgReader = optionSet.getOption(ExecConstants.KAFKA_RECORD_READER).string_val;
+    this.kafkaPollTimeOut = optionSet.getOption(ExecConstants.KAFKA_POLL_TIMEOUT).num_val;
```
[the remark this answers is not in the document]
ilooner: done

contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoRecordReader.java
```diff
@@ -91,9 +90,9 @@ public MongoRecordReader(MongoSubScan.MongoSubScanSpec subScanSpec, List<SchemaP
         subScanSpec.getMinFilters(), subScanSpec.getMaxFilters());
 
     buildFilters(subScanSpec.getFilter(), mergedFilters);
-    enableAllTextMode = fragmentContext.getOptions().getOption(ExecConstants.MONGO_ALL_TEXT_MODE).bool_val;
-    readNumbersAsDouble = fragmentContext.getOptions().getOption(ExecConstants.MONGO_READER_READ_NUMBERS_AS_DOUBLE).bool_val;
-    isBsonRecordReader = fragmentContext.getOptions().getOption(ExecConstants.MONGO_BSON_RECORD_READER).bool_val;
+    enableAllTextMode = fragmentContext.getOptionSet().getOption(ExecConstants.MONGO_ALL_TEXT_MODE).bool_val;
+    readNumbersAsDouble = fragmentContext.getOptionSet().getOption(ExecConstants.MONGO_READER_READ_NUMBERS_AS_DOUBLE).bool_val;
+    isBsonRecordReader = fragmentContext.getOptionSet().getOption(ExecConstants.MONGO_BSON_RECORD_READER).bool_val;
```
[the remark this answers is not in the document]
ilooner: done

contrib/storage-mongo/src/main/java/org/apache/drill/exec/store/mongo/MongoRecordReader.java
```diff
@@ -144,13 +143,12 @@ private void init(MongoSubScan.MongoSubScanSpec subScanSpec) {
     }
     MongoClient client = plugin.getClient(addresses);
     MongoDatabase db = client.getDatabase(subScanSpec.getDbName());
-    this.unionEnabled = fragmentContext.getOptions().getOption(ExecConstants.ENABLE_UNION_TYPE);
+    this.unionEnabled = fragmentContext.getOptionSet().getOption(ExecConstants.ENABLE_UNION_TYPE);
```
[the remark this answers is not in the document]
ilooner: done

exec/java-exec/src/main/java/org/apache/drill/exec/ops/BaseOperatorContext.java
ilooner: fixed

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Thanks for the explanation. Sounds very reasonable to me. I will work on splitting things up into 3 interfaces. The FragmentContextImpl class could probably implement all three interfaces and be cast to the appropriate interface depending on where it is passed to.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: This is required since it's used by operators like join, hashagg, filter, and flatten. I've replaced FunctionImplementationRegistry with the FunctionLookupContext interface now. So it has no dependence on ZK or networking.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Yeah your right, it's only used by the FragmentExecutor. Removed it from the interface.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Removed

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Changed again. Moved this into ExecutorFragmentContext.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Changed again moved to ExecutorFragmentContext

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Moved to ExchangeFragmentContext

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Moved to ExchangeFragmentContext

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Moved to ExchangeFragmentContext

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Moved to ExecutorFragmentContext

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
ilooner: Moved to RootFragmentContext

ilooner:
@paul-rogers 

I have split the FragmentContext into Four interfaces. We discussed 3, but it felt like 4 was more appropriate since some methods didn't fit well in some of the contexts.

- **ExecutorFragmentContext**: This is the highest level context that exposes access to everything. This is used by the FragmentExecutor and is also passed to the BatchCreators and RootCreators. 
- **RootFragmentContext**: This is exposed to the root executors. It has a few methods only needed by the root, but doesn't have access to everything like the ExecutorFragmentContext.
- **ExchangeFragmentContext**: This contains the network functions, and is used by the senders and recievers.
- **FragmentContext**: This is the barebones context which has no network functions and only exposes the bare minimum needed by operators.

These contexts are interfaces that inherit each other in the following way:

ExecutorFragmentContext -> RootFragmentContext -> ExchangeFragmentContext -> FragmentContext

There is only one implementation of the interfaces **FragmentContextImpl**. The facade pattern is used to cast the context to the appropriate interface depending on where it's used.

Let me know if you'd like to change the hierarchy or organization. Once we finalize how we want to organize things I will add javadoc.

exec/java-exec/src/test/java/org/apache/drill/exec/store/TestAffinityCalculator.java
```diff
@@ -21,18 +21,14 @@
...
 
 import com.google.common.collect.ImmutableRangeMap;
 import com.google.common.collect.Range;
 
 public class TestAffinityCalculator extends ExecTest {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(TestAffinityCalculator.class);
-
-  String port = "1234";
-  final String path = "path";
+  private final String port = "1234";
```
paul-rogers: What happened to the blocks of code that were removed? Not used? Duplicate? Or, does the unused code suggest that this test is not actually testing what it should?

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
```diff
@@ -86,37 +117,35 @@
...
+    return new MockOperatorContext(context, allocator(), config);
+  }
+
   /**
    * Builds an operator fixture based on a set of config options and system/session
    * options.
    */
-
-  public static class OperatorFixtureBuilder
+  public static class Builder
```
paul-rogers: Your PR is accidentally undoing a name change from a prior commit. Please check if we have any other collisions in this or related files.

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
```diff
@@ -127,20 +156,26 @@ public OperatorFixture build() {
    * uses the same code generation mechanism as the full Drill, but
    * provide test-specific versions of various other services.
    */
-
-  public static class TestFragmentContext extends BaseFragmentContext {
-
+  public static class MockFragmentContext extends BaseFragmentContext {
```
paul-rogers: More collisions? Can't image you made this many changes here...

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
```diff
@@ -175,19 +210,189 @@ public DrillConfig getConfig() {
...
+      return null;
+    }
+
+    @Override
+    public void waitForSendComplete() {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public AccountingDataTunnel getDataTunnel(CoordinationProtos.DrillbitEndpoint endpoint) {
```
paul-rogers: We really don't want to expose this network-related stuff in most fragment contexts. The idea for the API I created was to separate out the operator-related stuff from the Drillbit/network stuff. Doing so minimizes coupling and mocking. Shame to see us going back the other direction...

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
```diff
@@ -351,8 +603,110 @@ public OperatorStats getStats() {
     }
   }
 
-  public OperatorContext operatorContext(PhysicalOperator config) {
-    return new TestOperatorContext(context, allocator(), config, stats);
+  public static class MockPhysicalOperator implements PhysicalOperator {
```
paul-rogers: Seems overkill. Most times, a test that needs an operator will provide one. The dummy one that was in this code (or maybe in my latest PR) handles those cases where we don't even have an operator (record batch) so any operator (definition) will do.

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
ilooner: Agreed. This is actually not exposed in the FragmentContext used by operators, it is only exposed in the **ExecutorFragmentContext** which is used by the FragmentExecutor and passed to the BatchCreators. Unfortunately we still need a simple mock of the ExecutorFragmentContext because batchCreators are used in PhysicalOpUnitTestBase. Furthermore the networking methods aren't even used in that test so we don't even have to provide functional mocks of them for that one test. Since PhysicalOpUnitTestBase is the only test that requires an ExecutorFragmentContext I will move this mock class into PhysicalOpUnitTestBase and will add a comment that this Mock class should not be used outside of PhysicalOpUnitTestBase. All the other tests are able to get away with FragmentContext, which does not have any networking methods. And eventually we should remove PhysicalOpUnitTestBase in favor of the Fixture classes you added, and we will then be able to delete this MockClass entirely.

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
ilooner: Thanks for catching this. This class is left over from some previous work and is not used. I've deleted it.

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
ilooner: I like prefixing Mock classes with **Mock** instead of **Test**.

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
ilooner: There are no collisions. Java has a nice syntax for handling names like this. Specifically you can use this name for the class **OperatorFixture.Builder** in all your variable declarations. This removes the redundancy of prefixing the name of an inner class with the name of the outer class.

paul-rogers: Yes, of course. But one of the changes was to pull this out to a new top-level class. In that case, "Builder" by itself is rather generic.

exec/java-exec/src/test/java/org/apache/drill/exec/store/TestAffinityCalculator.java
ilooner: - Two tests were blatantly commented out.
   - testSetEndpointBytes
   - testApplyAssignments
 - Several methods were unused according to IntelliJ and the java compiler.
   - buildRowGroups
   - buildEndpoints
   - buildBlockLocations2

After removing the unused code there is shockingly almost nothing left. There are basically no functioning unit tests for the AffinityCalculator.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/managed/TestSorter.java
```diff
@@ -59,7 +59,7 @@
 @Category(OperatorTest.class)
 public class TestSorter extends DrillTest {
 
-  public static OperatorFixture fixture;
+  public volatile static OperatorFixture fixture;
```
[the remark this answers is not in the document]
ilooner: I was running into an issue where the fixture variable was consistently null when the test ran. This seemed impossible and I hypothesized that JUnit is secretly using two threads, one to call the static initializers and then another to actually execute the test methods. So I made the variable volatile and the issue went away. I agree this is weird and it raises the question why other tests don't have the same issue. I'll remove the volatile for now, if the bizarre issue surfaces again, I will get more details and post them here.

exec/java-exec/src/test/java/org/apache/drill/PlanningBase.java
```diff
@@ -58,23 +55,24 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.Resources;
 
-public class PlanningBase extends ExecTest{
-  //private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PlanningBase.class);
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
```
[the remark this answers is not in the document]
ilooner: This test was previously using JMockit and I have simply replaced it with Mockito which is Eclipse friendly and a step forward for developers using Eclipse. So the amount of Mocking going on here is the same :) the only difference is that Eclipse users should now be able to run these tests. I have also completely removed mocking from some tests which no longer need it. As we take more steps to properly use interfaces for more classes, we can incrementally remove Mockito from even more tests. However, this is an incremental process and shouldn't be done all in one shot.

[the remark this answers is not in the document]
ilooner: https://issues.apache.org/jira/browse/DRILL-6097 will be the next step forward to removing our dependence on Mocking libraries :).

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/BaseRawBatchBuffer.java
```diff
@@ -172,7 +172,7 @@ public RawFragmentBatch getNext() throws IOException {
     } catch (final InterruptedException e) {
 
       // We expect that the interrupt means the fragment is canceled or failed, so we should kill this buffer
-      if (!context.shouldContinue()) {
+      if (!context.getExecutorState().shouldContinue()) {
```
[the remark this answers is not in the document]
ilooner: ExecutorState is an interface. I removed those methods from the FragmentContext interface because it felt redundant to replicate the ExecutorState interface methods in the FragmentContext interface.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/WriterRecordBatch.java
```diff
@@ -118,7 +118,7 @@ public IterOutcome innerNext() {
     } catch(IOException ex) {
       logger.error("Failure during query", ex);
       kill(false);
-      context.fail(ex);
+      context.getExecutorState().fail(ex);
```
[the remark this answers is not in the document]
ilooner: Sounds fun. Created https://issues.apache.org/jira/browse/DRILL-6098

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -288,7 +288,7 @@ public static void populateVector(final ValueVector vector, final DrillBuf manag
     }
   }
 
-  public static MajorType getMajorTypeFromHiveTypeInfo(final TypeInfo typeInfo, final OptionManager options) {
+  public static MajorType getMajorTypeFromHiveTypeInfo(final TypeInfo typeInfo, final OptionSet options) {
```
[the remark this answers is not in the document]
ilooner: I'll flip to using the OptionManager.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/managed/TestSorter.java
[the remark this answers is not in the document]
paul-rogers: Strange... So, if having volatile does solve a problem, just leave it in, but maybe add a comment that includes you note above so that we know what's what in the future.

exec/java-exec/src/test/java/org/apache/drill/PlanningBase.java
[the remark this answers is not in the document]
paul-rogers: Thanks!

exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java
ilooner: It's a nested class in the master branch currently https://github.com/apache/drill/blob/master/exec/java-exec/src/test/java/org/apache/drill/test/OperatorFixture.java . So I didn't change that in this PR. Did you want me to pull it out into a top level class as part of this PR?

ilooner:
Addressed review comments. Also in an effort to accelerate the review I removed Misc code changes like the deletion of unused code and variables from a few dozen classes. I will open a separate PR with those Misc code improvements.

@Ben-Zvi Please take a look.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaRecordReader.java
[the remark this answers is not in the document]
Ben-Zvi: Should we "complete" the cleanup - go over the code and convert all the getOption(..) to the getTYPE(..) ones ?  Also for standardizing -- should all the name definitions, like
{code}
public static final String ENABLE_UNION_TYPE_KEY = "exec.enable_union_type";
{code}
be changed as needed to have the suffix  **_KEY** ?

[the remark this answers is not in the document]
paul-rogers: Would be a good code cleanup task to control entropy. However, this is the kind of thing that can be done incrementally over time as the need arises.

Also, standardizing on session/system option names would be good: maybe `_KEY` for names, `_VALIDATOR` for validators.

ilooner:
Rebased and resolved conflicts @Ben-Zvi please take a look.

Ben-Zvi:
LGTM - +1 .
Does the new code also fixes DRILL-5170 ? If so, it should be changed to a duplicate of 5730.

ilooner:
Your right. I'll mark it as a duplicate.

priteshm:
@arina-ielchiieva - any chance this can be committed with the batch commits tomorrow? otherwise, @ilooner would need to rebase again!

PR #1048, opened 2017-11-22
DRILL-5987: Use one version of javassist

ilooner:
There were two different versions of javassist being pulled in by **msgpack** and **reflections**. This makes sure only one version is included.

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -63,12 +63,29 @@
       <groupId>org.msgpack</groupId>
       <artifactId>msgpack</artifactId>
       <version>0.6.6</version>
+      <exclusions>
+        <exclusion>
+          <groupId>org.javassist</groupId>
+          <artifactId>javassist</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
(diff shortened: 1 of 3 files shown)


ilooner:
@arina-ielchiieva

common/pom.xml
```diff
@@ -63,12 +63,29 @@
       <groupId>org.msgpack</groupId>
       <artifactId>msgpack</artifactId>
       <version>0.6.6</version>
+      <exclusions>
```
vrozov: Exclusions should not be necessary.

ilooner: Removed exclusion for msgpack. The exclusion of javassist for reflections is necessary because before javassist version 3.13 the groupId was **javassist** after version 3.13 the groupId of javassist was changed to **org.javassist**. I also noticed ./logical/pom.xml declared the same msgpack and reflections dependencies as drill common. So I removed the redundant dependency declaration and also made msgpack and reflections managed dependencies.

arina-ielchiieva:
+1, LGTM.

PR #1050, opened 2017-11-23
DRILL-5964: Do not allow queries to access paths outside the current …

parthchandra:
Added check to prevent users from accessing a path outside the current workspace. For backward compatibility this introduces a new parameter 'allowAccessOutsideWorkspace' in the dfs storage plugin configuration that allows the user to override the check and allow access outside the workspace. The default value for the parameter is false. Any existing storage plugin configurations which do not have the parameter specified will no longer be able to access paths outside the workspace.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
@@ -252,11 +252,15 @@ public class FileSelection {
     return builder.toString();
   }
 
-  public static FileSelection create(final DrillFileSystem fs, final String parent, final String path) throws IOException {
+  public static FileSelection create(final DrillFileSystem fs, final String parent, final String path,
+      final boolean allowAccessOutsideWorkspace) throws IOException {
     Stopwatch timer = Stopwatch.createStarted();
     boolean hasWildcard = path.contains(WILD_CARD);
 
     final Path combined = new Path(parent, removeLeadingSlash(path));
+    if (!allowAccessOutsideWorkspace) {
+      checkBackPaths(parent, combined.toString(), path);
+    }
     final FileStatus[] statuses = fs.globStatus(combined); // note: this would expand wildcards
     if (statuses == null) {
       return null;
@@ -359,13 +363,28 @@ public class FileSelection {
     }
   }
 
-  private static String removeLeadingSlash(String path) {
-    if (path.charAt(0) == '/') {
+  public static String removeLeadingSlash(String path) {
+    if (!path.isEmpty() && path.charAt(0) == '/') {
       String newPath = path.substring(1);
       return removeLeadingSlash(newPath);
     } else {
       return path;
     }
+  }
+
+  // Check if the path is a valid sub path under the parent after removing backpaths. Throw an exception if
+  // it is not
+  // We pass subpath in as a parameter only for the error message
+  public static boolean checkBackPaths(String parent, String combinedPath, String subpath) {
+    Preconditions.checkArgument(!parent.isEmpty());
+    Preconditions.checkArgument(!combinedPath.isEmpty());
+
+    if (!combinedPath.startsWith(parent)) {
+      StringBuilder msg = new StringBuilder();
+      msg.append("Invalid path : ").append(subpath).append(" takes you outside the workspace.");
+      throw new IllegalArgumentException(msg.toString());
+    }
+    return true;
   }
 
   public List<FileStatus> getFileStatuses() {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
@@ -259,7 +259,7 @@ public class FileSelection {
 
     final Path combined = new Path(parent, removeLeadingSlash(path));
     if (!allowAccessOutsideWorkspace) {
-      checkBackPaths(parent, combined.toString(), path);
+      checkBackPaths(new Path(parent).toUri().getPath(), combined.toUri().getPath(), path);
     }
     final FileStatus[] statuses = fs.globStatus(combined); // note: this would expand wildcards
     if (statuses == null) {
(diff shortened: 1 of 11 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
```diff
@@ -359,15 +363,30 @@ private static Path handleWildCard(final String root) {
...
+  public static String removeLeadingSlash(String path) {
+    if (!path.isEmpty() && path.charAt(0) == '/') {
       String newPath = path.substring(1);
       return removeLeadingSlash(newPath);
     } else {
       return path;
     }
   }
 
+  // Check if the path is a valid sub path under the parent after removing backpaths. Throw an exception if
```
arina-ielchiieva: Please use java doc.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
```diff
@@ -359,15 +363,30 @@ private static Path handleWildCard(final String root) {
...
+  // it is not
+  // We pass subpath in as a parameter only for the error message
+  public static boolean checkBackPaths(String parent, String combinedPath, String subpath) {
```
arina-ielchiieva: This method can return void, since it never will return false. I see you use this during testing but it can be tested with void type as well.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceConfig.java
```diff
@@ -30,18 +30,24 @@
...
 
   /** Default workspace is a root directory which supports read, but not write. */
-  public static final WorkspaceConfig DEFAULT = new WorkspaceConfig("/", false, null);
+  public static final WorkspaceConfig DEFAULT = new WorkspaceConfig("/", false, null, false);
 
   private final String location;
   private final boolean writable;
   private final String defaultInputFormat;
-
+  private final Boolean allowAccessOutsideWorkspace; // allow access outside the workspace by default. This
```
arina-ielchiieva: Consider using primitive, we don't want unnecessary boxing / unboxing.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
```diff
@@ -252,11 +252,15 @@ private static String buildPath(final String[] path, final int folderIndex) {
...
 
-  public static FileSelection create(final DrillFileSystem fs, final String parent, final String path) throws IOException {
+  public static FileSelection create(final DrillFileSystem fs, final String parent, final String path,
+      final boolean allowAccessOutsideWorkspace) throws IOException {
     Stopwatch timer = Stopwatch.createStarted();
     boolean hasWildcard = path.contains(WILD_CARD);
 
     final Path combined = new Path(parent, removeLeadingSlash(path));
+    if (!allowAccessOutsideWorkspace) {
+      checkBackPaths(parent, combined.toString(), path);
```
arina-ielchiieva: I usually void using `toString` for `Path`, consider using `combined.toUri().getPath()`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
```diff
@@ -359,15 +363,30 @@ private static Path handleWildCard(final String root) {
...
       return path;
     }
   }
 
+  // Check if the path is a valid sub path under the parent after removing backpaths. Throw an exception if
+  // it is not
+  // We pass subpath in as a parameter only for the error message
+  public static boolean checkBackPaths(String parent, String combinedPath, String subpath) {
+    Preconditions.checkArgument(!parent.isEmpty());
+    Preconditions.checkArgument(!combinedPath.isEmpty());
```
arina-ielchiieva: Please add message for pre-conditions so error message will be more clear.

exec/java-exec/src/test/java/org/apache/drill/exec/store/dfs/TestFileSelection.java
```diff
@@ -59,4 +61,53 @@ public void testEmptyFolderThrowsTableNotFound() throws Exception {
       throw ex;
     }
   }
+
+  @Test
+  public void testBackPath() throws Exception {
```
arina-ielchiieva: Please split into two tests.

parthchandra: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceConfig.java
arina-ielchiieva: As far I remember `@JsonProperty("allowAccessOutsideWorkspace") boolean allowAccessOutsideWorkspace` will set false by default if value is not present during deserialization. Could you please check?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceConfig.java
```diff
@@ -30,18 +30,25 @@
...
                          @JsonProperty("writable") boolean writable,
-                         @JsonProperty("defaultInputFormat") String defaultInputFormat) {
+                         @JsonProperty("defaultInputFormat") String defaultInputFormat,
+                         @JsonProperty("allowAccessOutsideWorkspace") Boolean allowAccessOutsideWorkspace
+      ) {
     this.location = location;
     this.writable = writable;
     this.defaultInputFormat = defaultInputFormat;
+    //this.allowAccessOutsideWorkspace = allowAccessOutsideWorkspace != null ? allowAccessOutsideWorkspace : false ;
+    this.allowAccessOutsideWorkspace = true;
```
arina-ielchiieva: It seems we should not always set true...

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceConfig.java
parthchandra: Yes it would, I believe. But we want the value to be `true` for backward compatibility. (This also addresses your next comment). So we need to know if the value is missing. Can only do that with a non primitive AFAIK.

arina-ielchiieva: 1. Can we adjust the variable to be false by default, i.e. rename it to `disallowAccessOutsideWorkspace`? Thus we'll be able to use primitive, right?
2. In the below code you always set `this.allowAccessOutsideWorkspace = true;`, block with `this.allowAccessOutsideWorkspace = allowAccessOutsideWorkspace != null ? allowAccessOutsideWorkspace : false ;` is commented. I guess this is a mistake.

parthchandra:
Changed the type of allowAccessOutsideWorkspace to boolean. This changes behavior. All existing dfs workspaces will also disallow access outside the workspace. Ideally, we should deprecate this parameter in a future release, but for the moment it allows existing users to access paths outside the workspace if they want to.

arina-ielchiieva:
+1, LGTM.

PR #1051, opened 2017-11-27
DRILL-5963: Query state process improvements

arina-ielchiieva:
1. Added two new query states: PREPARING (when foreman is initialized) and PLANNING (includes logical and / or physical planning).
2. Process of query enqueuing was made non-blocking. Once query was enqueued, fragments runner is called to submit fragments locally and remotely.
3. Ability to cancel query during planning and enqueued states was added.
4. Logic for submitting fragments was moved from Foreman to new class FragmentsRunner.
5. Major type in DrillFuncHolderExpr was cached for better performance.

Other details in [DRILL-5963](https://issues.apache.org/jira/browse/DRILL-5963)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
@@ -17,7 +17,6 @@
  */
 package org.apache.drill.exec.work.foreman;
 
-import com.codahale.metrics.Counter;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import com.google.protobuf.InvalidProtocolBufferException;
@@ -25,7 +24,6 @@ import io.netty.channel.ChannelFuture;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 import org.apache.drill.common.CatastrophicFailure;
-import org.apache.drill.common.EventProcessor;
 import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.logical.LogicalPlan;
@@ -33,7 +31,6 @@ import org.apache.drill.common.logical.PlanProperties.Generator.ResultMode;
 import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.exception.OptimizerException;
 import org.apache.drill.exec.exception.OutOfMemoryException;
-import org.apache.drill.exec.metrics.DrillMetrics;
 import org.apache.drill.exec.ops.QueryContext;
 import org.apache.drill.exec.opt.BasicOptimizer;
 import org.apache.drill.exec.physical.PhysicalPlan;
@@ -64,6 +61,8 @@ import org.apache.drill.exec.testing.ControlsInjectorFactory;
 import org.apache.drill.exec.util.Pointer;
 import org.apache.drill.exec.work.QueryWorkUnit;
 import org.apache.drill.exec.work.WorkManager.WorkerBee;
+import org.apache.drill.exec.work.foreman.rm.QueryQueue.QueueTimeoutException;
+import org.apache.drill.exec.work.foreman.rm.QueryQueue.QueryQueueException;
 import org.apache.drill.exec.work.foreman.rm.QueryResourceManager;
 import org.codehaus.jackson.map.ObjectMapper;
 
@@ -100,15 +99,6 @@ public class Foreman implements Runnable {
 
   private static final ObjectMapper MAPPER = new ObjectMapper();
 
-
-  private static final Counter planningQueries = DrillMetrics.getRegistry().counter("drill.queries.planning");
-  private static final Counter enqueuedQueries = DrillMetrics.getRegistry().counter("drill.queries.enqueued");
-  private static final Counter runningQueries = DrillMetrics.getRegistry().counter("drill.queries.running");
-  private static final Counter completedQueries = DrillMetrics.getRegistry().counter("drill.queries.completed");
-  private static final Counter succeededQueries = DrillMetrics.getRegistry().counter("drill.queries.succeeded");
-  private static final Counter failedQueries = DrillMetrics.getRegistry().counter("drill.queries.failed");
-  private static final Counter canceledQueries = DrillMetrics.getRegistry().counter("drill.queries.canceled");
-
   private final QueryId queryId;
   private final String queryIdString;
   private final RunQuery queryRequest;
@@ -116,18 +106,16 @@ public class Foreman implements Runnable {
   private final QueryManager queryManager; // handles lower-level details of query execution
   private final DrillbitContext drillbitContext;
   private final UserClientConnection initiatingClient; // used to send responses
-  private volatile QueryState state;
   private boolean resume = false;
   private final ProfileOption profileOption;
 
   private final QueryResourceManager queryRM;
 
   private final ResponseSendListener responseListener = new ResponseSendListener();
-  private final StateSwitch stateSwitch = new StateSwitch();
-  private final ForemanResult foremanResult = new ForemanResult();
   private final ConnectionClosedListener closeListener = new ConnectionClosedListener();
   private final ChannelFuture closeFuture;
   private final FragmentsRunner fragmentsRunner;
+  private final QueryStateProcessor queryStateProcessor;
 
   private String queryText;
 
@@ -146,40 +134,41 @@ public class Foreman implements Runnable {
     this.queryIdString = QueryIdHelper.getQueryId(queryId);
     this.queryRequest = queryRequest;
     this.drillbitContext = drillbitContext;
-
-    initiatingClient = connection;
-    closeFuture = initiatingClient.getChannelClosureFuture();
+    this.initiatingClient = connection;
+    this.closeFuture = initiatingClient.getChannelClosureFuture();
     closeFuture.addListener(closeListener);
 
-    queryContext = new QueryContext(connection.getSession(), drillbitContext, queryId);
-    queryManager = new QueryManager(queryId, queryRequest, drillbitContext.getStoreProvider(),
+    this.queryContext = new QueryContext(connection.getSession(), drillbitContext, queryId);
+    this.queryManager = new QueryManager(queryId, queryRequest, drillbitContext.getStoreProvider(),
         drillbitContext.getClusterCoordinator(), this);
-
-    recordNewState(QueryState.PREPARING);
-
-    queryRM = drillbitContext.getResourceManager().newQueryRM(this);
-    fragmentsRunner = new FragmentsRunner(bee, initiatingClient, drillbitContext, this);
-
-    profileOption = setProfileOption(queryContext.getOptions());
+    this.queryRM = drillbitContext.getResourceManager().newQueryRM(this);
+    this.fragmentsRunner = new FragmentsRunner(bee, initiatingClient, drillbitContext, this);
+    this.queryStateProcessor = new QueryStateProcessor(queryIdString, queryManager, drillbitContext, new ForemanResult());
+    this.profileOption = setProfileOption(queryContext.getOptions());
   }
 
-  private ProfileOption setProfileOption(OptionManager options) {
-    if (! options.getOption(ExecConstants.ENABLE_QUERY_PROFILE_VALIDATOR)) {
-      return ProfileOption.NONE;
-    }
-    if (options.getOption(ExecConstants.QUERY_PROFILE_DEBUG_VALIDATOR)) {
-      return ProfileOption.SYNC;
-    } else {
-      return ProfileOption.ASYNC;
-    }
+
+  /**
+   * @return query id
+   */
+  public QueryId getQueryId() {
+    return queryId;
   }
 
-  private class ConnectionClosedListener implements GenericFutureListener<Future<Void>> {
-    @Override
-    public void operationComplete(Future<Void> future) throws Exception {
-      cancel();
-    }
+  /**
+   * @return current query state
+   */
+  public QueryState getState() {
+    return queryStateProcessor.getState();
   }
+
+  /**
+   * @return sql query text of the query request
+   */
+  public String getQueryText() {
+    return queryText;
+  }
+
 
   /**
    * Get the QueryContext created for the query.
@@ -200,16 +189,21 @@ public class Foreman implements Runnable {
   }
 
   /**
-   * Cancel the query. Asynchronous -- it may take some time for all remote fragments to be terminated.
-   * For planning and enqueued states we cancel immediately since these states are done locally.
-   *
-   * Note this can be called from outside of run() on another thread, or after run() completes
+   * Cancel the query (move query in cancellation requested state).
+   * Query execution will be canceled once possible.
    */
   public void cancel() {
-    if (QueryState.PLANNING == state || QueryState.ENQUEUED == state) {
-      moveToState(QueryState.CANCELLATION_REQUESTED, null);
-    }
-    addToEventQueue(QueryState.CANCELLATION_REQUESTED, null);
+    queryStateProcessor.cancel();
+  }
+
+  /**
+   * Adds query status in the event queue to process it when foreman is ready.
+   *
+   * @param state new query state
+   * @param exception exception if failure has occurred
+   */
+  public void addToEventQueue(QueryState state, Exception exception) {
+    queryStateProcessor.addToEventQueue(state, exception);
   }
 
   /**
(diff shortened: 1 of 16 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/ThrottledResourceManager.java
```diff
@@ -314,13 +315,49 @@ public void exit() {
...
+        this.foreman = foreman;
+        this.queryCost = queryCost;
+      }
+
+      @Override
+      public void run() {
+        try {
+          queryQueue.enqueue(foreman.getQueryId(), queryCost);
+          foreman.moveToState(QueryState.STARTING, null);
+        } catch (Exception e) {
```
ppadma: Should we handle interruptedException (for cancel above) here ? I don't think we want to move Query to FAILED state by default.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/ThrottledResourceManager.java
```diff
@@ -288,8 +287,10 @@ public void setCost(double cost) {
     }
 
     @Override
-    public void admit() throws QueueTimeoutException, QueryQueueException {
-      lease = rm.queue().enqueue(foreman.getQueryId(), queryCost);
+    public void admit() {
+      QueueAcquirer queueAcquirer = new QueueAcquirer(rm.queue(), foreman, queryCost);
+      queueAcquirerThread = new Thread(queueAcquirer);
```
ppadma: General comment about the approach. We create minimum 2 threads for any query. With this, when queuing is enabled, we are creating minimum three threads. Instead, when a query is cancelled,  is it possible to interrupt the foreman thread itself if that is blocked ?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -426,48 +413,25 @@ private void runPhysicalPlan(final PhysicalPlan plan) throws ExecutionSetupExcep
...
-          .resourceError()
-          .message(e.getMessage())
-          .build(logger);
-    } catch (QueryQueueException e) {
-      throw new ForemanSetupException(e.getMessage(), e);
-    } finally {
-      queryManager.markQueueWaitEndTime();
-    }
-    moveToState(QueryState.STARTING, null);
+    planningQueries.dec();
```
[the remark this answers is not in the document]
paul-rogers: Agree. A better solution is to have a separate function for each new state to avoid big hairy switch statements from current state to desired state.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -873,105 +791,133 @@ public void close() throws Exception {
...
+            recordNewState(newState);
+            foremanResult.setFailed(exception);
+            foremanResult.close();
+            return;
+        }
+        break;
+      case ENQUEUED:
+        enqueuedQueries.dec();
+        queryManager.markQueueWaitEndTime();
+        switch (newState) {
```
paul-rogers: This is exactly the kind of double-switch statement that is hard to understand. Better, a method for each transition: `enqueue()`, `start()`, etc. Each transition is then easier to reason about. What are the valid from states? What are the accounting operations needed in each transition?

Even better would be to combine the logic for the state transition (adding the query to the queue, say) along with the state transition logic, since often the new state is the result not simply of a transition, but of some condition (such as, when queueing, moving to either ENQUEUED, RUNNING or FAILED depending on the queue result.)

Otherwise the logic has to be essentially duplicated.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+        drillbitContext.getWorkBus().addFragmentStatusListener(queryId, queryManager.getFragmentStatusListener());
+        drillbitContext.getClusterCoordinator().addDrillbitStatusListener(queryManager.getDrillbitStatusListener());
+
+        logger.debug("Submitting fragments to run.");
+        // set up the root fragment first so we'll have incoming buffers available.
+        setupRootFragment(rootPlanFragment, rootOperator);
+        setupNonRootFragments(planFragments);
+
+      } catch (ExecutionSetupException e) {
+        foreman.moveToState(QueryState.FAILED, e);
```
paul-rogers: Nothing is thrown here. We simply drop down into the running state.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+      }
+
+      foreman.moveToState(QueryState.RUNNING, null);
+      logger.debug("Fragments running.");
+    } finally {
+      foreman.startProcessingEvents();
```
paul-rogers: If we get here, we may have bypassed a state transition if an error occurred. Should the above two nested `try` blocks be collapsed so we are sure to catch all exceptions?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+      try {
+        drillbitContext.getWorkBus().addFragmentStatusListener(queryId, queryManager.getFragmentStatusListener());
+        drillbitContext.getClusterCoordinator().addDrillbitStatusListener(queryManager.getDrillbitStatusListener());
+
+        logger.debug("Submitting fragments to run.");
+        // set up the root fragment first so we'll have incoming buffers available.
+        setupRootFragment(rootPlanFragment, rootOperator);
+        setupNonRootFragments(planFragments);
+
+      } catch (ExecutionSetupException e) {
```
paul-rogers: We catch only `ExceptionSetupException`. But, Drill loves to throw unchecked exceptions. Should we handle those as well? Or, are they handled elsewhere? If elsewhere, why do we handle this one exception here?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+     * We will send a single message to each endpoint, regardless of how many fragments will be
+     * executed there. We need to start up the intermediate fragments first so that they will be
+     * ready once the leaf fragments start producing data. To satisfy both of these, we will
+     * make a pass through the fragments and put them into the remote maps according to their
+     * leaf/intermediate state, as well as their target drillbit. Also filter the leaf/intermediate
+     * fragments which are assigned to run on local Drillbit node (or Foreman node) into separate lists.
+     *
+     * This will help to schedule local
+     */
+    final Multimap<CoordinationProtos.DrillbitEndpoint, PlanFragment> remoteLeafFragmentMap = ArrayListMultimap.create();
```
paul-rogers: Any reason not to just import `DrillbitEndpoint` as in the original code?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+   * @param assignment the drillbit assigned to these fragments
+   * @param fragments the set of fragments
+   * @param latch the countdown latch used to track the requests to all endpoints
+   * @param fragmentSubmitFailures the submission failure counter used to track the requests to all endpoints
+   */
+  private void sendRemoteFragments(final CoordinationProtos.DrillbitEndpoint assignment, final Collection<PlanFragment> fragments,
+                                   final CountDownLatch latch, final FragmentSubmitFailures fragmentSubmitFailures) {
+    @SuppressWarnings("resource")
+    final Controller controller = drillbitContext.getController();
+    final BitControl.InitializeFragments.Builder fb = BitControl.InitializeFragments.newBuilder();
```
paul-rogers: Similarly, any reason to not just import `InitializeFragments` as in the original code?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+    for(final PlanFragment planFragment : fragments) {
+      fb.addFragment(planFragment);
+    }
+    final BitControl.InitializeFragments initFrags = fb.build();
+
+    logger.debug("Sending remote fragments to \nNode:\n{} \n\nData:\n{}", assignment, initFrags);
```
paul-rogers: No real readon for the first four newlines...

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+
+    // send remote intermediate fragments
+    for (final CoordinationProtos.DrillbitEndpoint ep : remoteFragmentMap.keySet()) {
+      sendRemoteFragments(ep, remoteFragmentMap.get(ep), endpointLatch, fragmentSubmitFailures);
+    }
+
+    final long timeout = RPC_WAIT_IN_MSECS_PER_FRAGMENT * numIntFragments;
+    if (numIntFragments > 0 && !endpointLatch.awaitUninterruptibly(timeout)) {
+      long numberRemaining = endpointLatch.getCount();
+      throw UserException.connectionError()
```
paul-rogers: Here is a good example of an unchecked exception not handled by the outer try/catch block.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+        if (endpoints.add(endpoint)) {
+          if (first) {
+            first = false;
+          } else {
+            sb.append(", ");
+          }
+          sb.append(endpoint.getAddress());
+        }
+      }
+      throw UserException.connectionError(submissionExceptions.get(0).rpcException)
```
paul-rogers: Another unchecked exception.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -0,0 +1,439 @@
...
+        bee.addFragmentRunner(fragmentExecutor);
+      } else {
+        // isIntermediate, store for incoming data.
+        final NonRootFragmentManager manager = new NonRootFragmentManager(fragment, fragmentExecutor, statusReporter);
+        drillbitContext.getWorkBus().addFragmentManager(manager);
+      }
+
+    } catch (final ExecutionSetupException ex) {
+      throw new ForemanException("Failed to create fragment context", ex);
+    } catch (final Exception ex) {
```
paul-rogers: Here we translate a generic exception (which might be a `UserException` into a `ForemanException`, which we don't catch in our try/catch block. (I know this code is original, but I had to do a detailed side-by-side comparison with that original code, which revealed our somewhat awkward exception handling.)

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
```diff
@@ -280,29 +280,29 @@ public void interrupted(final InterruptedException ex) {
...
         break;
-
       case COMPLETED:
       case CANCELED:
       case FAILED:
         try {
           runningProfileStore.remove(stringQueryId);
           inTransientStore = false;
-        } catch(final Exception e) {
+        } catch (final Exception e) {
```
paul-rogers: Below. What is an "estore"? Is that a typo?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/DefaultResourceManager.java
```diff
@@ -73,13 +75,20 @@ public void setCost(double cost) {
     @Override
     public void admit() {
       // No queueing by default
+      foreman.moveToState(UserBitShared.QueryResult.QueryState.STARTING, null);
```
paul-rogers: Would REALLY prefer that the resource manager not depend on the Foreman; makes unit testing impossible. The Foreman should depend on this class, but not the other way around.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/QueryResourceManager.java
```diff
@@ -38,22 +35,17 @@
...
 
   /**
-   * Admit the query into the cluster. Blocks until the query
-   * can run. (Later revisions may use a more thread-friendly
-   * approach.)
-   * @throws QueryQueueException if something goes wrong with the
-   * queue mechanism
-   * @throws QueueTimeoutException if the query timed out waiting to
-   * be admitted.
+   * Admit the query into the cluster. Starts enqueueing process in separate thread.
```
paul-rogers: If we want to do this, please provide an interface with the needed services. Also, if we want to do this, then the implementation should change. Maybe admit returns a status: ADMITTED, QUEUED, FAILED. (Or, FAILED can be in the form of an exception.)

The mechanism should be: pass in a handle to the query. If admit, tell the handle to run. If blocked, figure out how to receive the notification from ZK. Once the query is admitted, let it run. If a timeout occurs, tell the handle that the query failed.

With a handle, we can unit test the mechanism.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/ThrottledResourceManager.java
paul-rogers: Wait thread should not be started here. See above comments. Let's discuss via e-mail.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/rm/ThrottledResourceManager.java
paul-rogers: This has FAR to much visibility into the Foreman. See above.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -502,33 +438,82 @@ private void runFragment(List<PlanFragment> fragmentsList) throws ExecutionSetup
...
+    queryStateProcessor.moveToState(QueryState.ENQUEUED, null);
+
+    try {
+      queryRM.admit();
+      queryStateProcessor.moveToState(QueryState.STARTING, null);
+    } catch (QueueTimeoutException | QueryQueueException e) {
+      queryStateProcessor.moveToState(QueryState.FAILED, e);
+    } finally {
+      String queueName = queryRM.queueName();
+      queryManager.setQueueName(queueName == null ? "Unknown" : queueName);
```
paul-rogers: I'm going to violate our code review guidelines a bit and discuss design. Sorry.

The above is a good place to talk about state design. We have the Foreman taking actions and setting states. We have the query state processor taking some actions based on those states and trying to enforce valid state transitions. I think what we have is a clash of two systems.

The Foreman is a big bunch of code that handles the procedure of parsing, preparing and launching a query, all in a single thread. The query state processor appears to handle not just this, but events that arrive after the query is started.

I wonder rather than having the Foreman do the setup work and the query state processor try to follow the states, should we just bite the bullet and make the query state processor be the actual query owner and manager?

The query manager provides events: parse, plan, enqueue, launch, etc. It uses states to make sure events occur in the expected order.

All the Foreman does is create the environment, create the query manager, and call the set of methods that we currently expect of the Foreman (from parse to launch.)

In the future, if we can get queuing to be non-blocking (works by events, not synchronized calls), then the Foreman would do planning through enqueueing. The "accept query" action would invoke the launch action.

Then, the query manager not only enforces states, it also takes the actions required when an event arrives in a given state.

Now, I realize all of the code in Foreman is existing. But, we are doing refactoring, so now is the time to think through these issues.

arina-ielchiieva: @paul-rogers `QueryStateProcessor` is responsible to enforcing valid state transitions (directly if fragments were not sent out yet) or though queuing (if fragments were sent out), proper cancelling and failing, Also it is responsible for query counters. `Foreman` is responsible for execution part and does all the work. If we move remaining logic from `Foreman` into  `QueryStateProcessor`, we'll end up with the same huge unreadable class as `Foreman` was before. At this point of refactoring, I suggest we stop and if further refactoring is needed, create new Jira.

paul-rogers: Alright. A better design is to have a state machine with events, each state/event pair is tied to a function that checks/implements that transition. The code now is a bit of a muddle. But, since doing that is just an improvement, not a bug fix, I agree with considering such changes as a separate JIRA.

PR #1053, opened 2017-11-29
DRILL-5989 Travis Finally Runs Smoke Tests!!!

ilooner:
## DRILL-5989

Previously Travis would only do a build for each pull request. Now Travis will also run all the smoke tests every time a PR is opened or updated. Smoke Tests are defined as tests which do **not** belong to one of these categories:

- **org.apache.drill.categories.SlowTest**
- **org.apache.drill.categories.SecurityTest**
- **org.apache.drill.categories.UnlikelyTest**

In order to get the smoke tests to run on Travis I had to make the following fixes:

- Categorized Kafka storage plugin tests
- Categorized more long running tests as slow tests in order to avoid timing out on travis
- Fixed a bug with jdbc-all tests which would cause jdbc-all to fail when categories were excluded
- Fixed usage of surefire plugin in java-exec, and fixed a bug which caused **excludedGroups** not to be propagated to the java-exec project.
- Made the heap size of the forked test process configurable via a property. This was necessary because running the whole build requires about 4gb of memory for the test process, but 4gb is too large for the Travis container and the smoke tests only require 3.5gb
- Cached maven dependencies for the Travis build so that it runs faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,4 @@ cache:
   directories:
   - "$HOME/.m2"
 install: MAVEN_OPTS="-Xms1G -Xmx1G" mvn install --batch-mode -DskipTests=true -Dmaven.javadoc.skip=true -Dmaven.source.skip=true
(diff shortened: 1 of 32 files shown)


ilooner:
@arina-ielchiieva

arina-ielchiieva:
+1, LGTM.

PR #1056, opened 2017-11-29
DRILL-6000: Categorized graceful shutdown unit tests as SlowTests

ilooner:
Graceful shutdown unit tests were failing on Travis, and should not be run as part of the SmokeTests


arina-ielchiieva:
+1, LGTM.

PR #1057, opened 2017-11-29
DRILL-5993 Add Generic Copiers With Append Methods

ilooner:
## Motivation

Code generate copiers are not as fast as Generic copiers as benchmarked by Paul.

## Implementation

There are Generic Copiers which can take rows from Direct, Sv2, or Sv4 RecordBatches and append those rows to a destination VectorContainer.

## Misc Fixes

- The test framework did not create and configure a temp directory for codegen files in BaseTestQuery or ClusterFixture, so I fixed that.
- ExtendableHyperContainer did not build it's schema after batches were added, fixed that.
- Removed unused arguments in the RemovingRecordBatch.getGenerated4Copier method


ilooner:
@Ben-Zvi @paul-rogers

paul-rogers:
To answer the two questions:

1. The copier is used in multiple locations, some of which include selection vectors. Sort uses a copier to merge rows coming from multiple sorted batches. The SVR compresses out SVs. A filter will produce an SV2 which the SVR removes. An in-memory sort produces an SV4. But, because of the ways plans are generated, the hash join will never see a batch with an SV. (An SVR will be inserted, if needed, to remove the SV.)

2. We never write a batch using an SV. The SV is always a source indirection. Because we do indirection on the source side (and vectors are append only), there can be no SV on the destination side.

Note also that the {{VectorContainer}} class, despite it's API, knows nothing about SVs. The SV is tacked on separately by the {{RecordBatch}}. (This is a less-than-ideal design, but it is how things work at present.) FWIW, the test-oriented {{RowSet}} abstractions came about as wrappers around both the {{VectorContainer}} and SV to provide a unified view.

Because of how we do SVs, you'll need three copy methods: one for no SV, one for an SV2 and another for an SV4.

In the fullness of time, the new "column reader" and "column writer" abstractions will hide all this stuff, but it will take time before those tools come online.

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
```diff
@@ -353,6 +353,23 @@ public int getRecordCount() {
...
+   */
+  public void appendRow(VectorContainer srcContainer, int srcIndex) {
+    for (int vectorIndex = 0; vectorIndex < wrappers.size(); vectorIndex++) {
+      ValueVector destVector = wrappers.get(vectorIndex).getValueVector();
+      ValueVector srcVector = srcContainer.wrappers.get(vectorIndex).getValueVector();
+
+      destVector.copyEntry(recordCount, srcVector, srcIndex);
+    }
+
+    recordCount++;
```
paul-rogers: This is OK for a row-by-row copy. But, you'll get better performance if you optimize for the entire batch. Because you have no SV4, the source and dest batches are the same so the vectors can be preloaded into an array of vectors to avoid the vector wrapper lookup per column.

Plus, if the code is written per batch, you can go a step further: vectorize the operation. Copy all values for column 1, then all for column 2, and so on. (In this case, you only get each vector once, so sticking with the wrappers is fine.) By vectorizing, you may get the vectorized cache-locality benefit that Drill promises from its operations. Worth a try to see if you get any speed-up.

Ben-Zvi: The immediate need for appendRow() is to distribute rows from a single incoming batch into multiple other batches (for the Hash Join internal partitioning), based on the hash value of the key columns at each row. This would not work well with the second suggestion (vectorizing - column 1, column 2, etc.)

exec/java-exec/src/test/java/org/apache/drill/exec/record/TestVectorContainer.java
```diff
@@ -124,4 +132,52 @@ public void testContainerMerge() {
     leftIndirect.clear();
     right.clear();
   }
+
+  @Test
+  public void testAppendRow()
+  {
+    MaterializedField colA = MaterializedField.create("colA", Types.required(TypeProtos.MinorType.INT));
+    MaterializedField colB = MaterializedField.create("colB", Types.required(TypeProtos.MinorType.INT));
```
Ben-Zvi: Maybe add some "interesting" datatypes ?  Testing integers only may miss some issue.

ilooner: Added more interesting types. Currently RowSet classes don't support the Map data type. Paul asked me to look into adding support for this a while ago for DRILL-5870 . I'll update the test framework to support that in the next PR.

ilooner:
@paul-rogers since HashJoin does not need to support Selection Vectors maybe we can postpone adding the corresponding appendRow methods if and when they are needed. I suspect by the time anyone needs those methods we will have already migrated over to the new batch framework.

ilooner:
Please let me know if I should make any more changes.

ilooner:
@Ben-Zvi @paul-rogers As per our discussion I've added Paul's generic copiers and have provided a GenericCopier that works for RecordBatches that do not have selection vectors. I've enhanced the copier interface to include append methods.

ilooner:
Added back the appendRow method to VectorContainer as per our discussion. I am keeping the GenericCopiers since they may be useful in the future for bulk appends?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/CopierTemplate4.java
```diff
@@ -54,17 +52,33 @@ public int copyRecords(int index, int recordCount) throws SchemaChangeException
...
 
-    int outgoingPosition = 0;
-    for(int svIndex = index; svIndex < index + recordCount; svIndex++, outgoingPosition++){
+    return insertRecords(0, index, recordCount);
+  }
+
+  @Override
+  public int appendRecord(int index) throws SchemaChangeException {
+    return appendRecords(index, 1);
+  }
```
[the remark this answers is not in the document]
ilooner: Updated the code and made an implementation of appendRecord which doesn't use a for loop

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/CopierTemplate2.java
```diff
@@ -53,17 +51,32 @@ public int copyRecords(int index, int recordCount) throws SchemaChangeException
...
+
+    outgoing.setRecordCount(outgoingPosition);
     return outgoingPosition;
   }
 
-  public abstract void doSetup(@Named("context") FragmentContext context,
-                               @Named("incoming") RecordBatch incoming,
-                               @Named("outgoing") RecordBatch outgoing)
+  public abstract void doSetup(@Named("incoming") RecordBatch incoming,
+                               @Named("outgoing") VectorContainer outgoing)
```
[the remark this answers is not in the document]
ilooner: The copiers are only used in the SVRemover and TopN operator. I have replaced the code generated copiers in both now to use the GenericCopiers.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/Copier.java
```diff
@@ -19,13 +19,15 @@
...
 import org.apache.drill.exec.exception.SchemaChangeException;
-import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.record.RecordBatch;
+import org.apache.drill.exec.record.VectorContainer;
 
 public interface Copier {
-  public static TemplateClassDefinition<Copier> TEMPLATE_DEFINITION2 = new TemplateClassDefinition<Copier>(Copier.class, CopierTemplate2.class);
-  public static TemplateClassDefinition<Copier> TEMPLATE_DEFINITION4 = new TemplateClassDefinition<Copier>(Copier.class, CopierTemplate4.class);
+  TemplateClassDefinition<Copier> TEMPLATE_DEFINITION2 = new TemplateClassDefinition<Copier>(Copier.class, CopierTemplate2.class);
+  TemplateClassDefinition<Copier> TEMPLATE_DEFINITION4 = new TemplateClassDefinition<Copier>(Copier.class, CopierTemplate4.class);
```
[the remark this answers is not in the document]
ilooner: I will create a separate PR to do that since that change is unrelated to this PR

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
paul-rogers: I would not advise changing the meaning of -1; that is baked into the code in many places. Instead, before adding the first row, simply set the count to 0 when the target container is created.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/svremover/GenericCopierTest.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.svremover;
+
+import org.apache.drill.exec.memory.RootAllocator;
+import org.apache.drill.exec.record.BatchSchema;
+import org.apache.drill.test.rowSet.RowSet;
+import org.apache.drill.test.rowSet.RowSetBuilder;
+
+public class GenericCopierTest extends AbstractGenericCopierTest {
```
paul-rogers: One reason I was a bit hesitant to add this support earlier is that we need this to work for ALL Drill's data types, including the obscure ones. Clearly, this mechanism works for scalars. But, what about:

* Arrays of scalars
* Maps
* Arrays of maps
* Unions
* Lists of unions
* Lists of scalars
* Lists of maps
* Repeated lists of scalars
* Repeated lists of maps
* Repeated lists of repeated lists
* Repeated lists of unions

There are probably a few more that I'm overlooking. Until we test ALL of these, we can't be sure if the new code works as well as the old code.

ilooner: The generated code uses the copyFromSafe methods in code generated copiers. Your copyEntry method which is used by the GenericCopiers also uses the copyFromSafe method. So we should inherit all the same supported types. I think if all the unit tests, functional tests, and QA tests pass that will be a sufficient enough vote of confidence. Ideally we would have unit tests for all the operators (not just the copiers) which use the complex data types, but there is significant amount of work left to be able to effectively unit tests with complex data types and adding all the machinery to do so is out of the scope of this PR.

ilooner:
Addressed review comments and rebased.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/AbstractSV4Copier.java
```diff
@@ -17,32 +17,30 @@
...
 import org.apache.drill.exec.record.RecordBatch;
+import org.apache.drill.exec.record.VectorContainer;
 import org.apache.drill.exec.record.VectorWrapper;
 import org.apache.drill.exec.record.selection.SelectionVector4;
 import org.apache.drill.exec.vector.AllocationHelper;
+import org.apache.drill.exec.vector.ValueVector;
 
-public abstract class CopierTemplate4 implements Copier{
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CopierTemplate4.class);
+public abstract class AbstractSV4Copier implements Copier {
```
Ben-Zvi: The code for AbstractSV4Copier is nearly identical to the code of AbstractSV2Copier; I wonder if we could have combined the two somehow (like a super class ??)

ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/AbstractCopier.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.drill.exec.record.VectorWrapper;
+import org.apache.drill.exec.vector.AllocationHelper;
+import org.apache.drill.exec.vector.ValueVector;
+
+public abstract class AbstractCopier implements Copier {
+  protected ValueVector[] vvOut;
+  protected VectorContainer outgoing;
+
+  @Override
+  public void setup(RecordBatch incoming, VectorContainer outgoing) throws SchemaChangeException {
```
Ben-Zvi: The "incoming" parameter is not used anywhere .......

ilooner: It is not used by the AbstractCopier itself but it is used by the AbstractSV2Copier and AbstractSV4Copier which override the method.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/GenericSV4Copier.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.drill.exec.record.VectorContainer;
+import org.apache.drill.exec.record.VectorWrapper;
+import org.apache.drill.exec.vector.SchemaChangeCallBack;
+import org.apache.drill.exec.vector.ValueVector;
+
+public class GenericSV4Copier extends AbstractSV4Copier {
+
+  @Override
+  public void setup(RecordBatch incoming, VectorContainer outgoing) throws SchemaChangeException {
+    super.setup(incoming, outgoing);
```
Ben-Zvi: Can the setup() code in both GenericSV2Copier and GenericSV4Copier be moved up into the setup() method of AbstractCopier ?  Both codes are identical ....

exec/java-exec/src/main/java/org/apache/drill/exec/record/BatchSchema.java
```diff
@@ -60,6 +60,10 @@ public SelectionVectorMode getSelectionVectorMode() {
     return selectionVectorMode;
   }
 
+  public BatchSchema toMode(SelectionVectorMode mode) {
```
Ben-Zvi: Unused code ?

ilooner: removed

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/GenericSV4Copier.java
ilooner: The code is similar but not identical. For **GenericSV4Copier** **vvIn** is a 2D array of type **ValueVector[][]**. For **GenericSV2Copier** **vvIn** is a 1D array of type **ValueVector[]**. Since the types are different I can't move the code into the **AbstractCopier**. But I will move the setup code in **GenericSV2Copier** to **AbstractSV2Copier** since that code will likely be the same for different implementations. I'll also do the same thing with the **GenericSV4Copier** and **AbstractSV4Copier** classes.

ilooner:
For posterity I will post some of the benchmarking Paul did. I will then do some of my own micro benchmarking in a later comment.

++ Code Gen

 - For a query which has an execution time of **5578 ms** code generating the copiers takes about **2159 ms** (I don't have the original query). If we use Generic copiers we save all of this time.

++ Execution

 - The execution time of code generated copiers for the query of interest was **1302 ms / run**.
 - The execution time of generic copiers for the query of interest was **1186 ms / run**.

++ Conclusion

Generic copiers are faster than code generated copiers. The JVM has made big strides in the past few years and can now optimize code at runtime much better than us.

paul-rogers:
Thanks @ilooner for doing this fix!

A bit of background. A user provided a "short, fat" query that had, as I recall, only 6000 rows but something like 200 columns. The numbers I posted were for the cost of doing the code generation for the rather large code (200 columns to be copied), amortized against a very short query. As @ilooner said, in this special case, there are no advantages to grinding the code generator, the cache lookup mechanism, the compiler and the byte-code merge mechanism, when a simple function call does the same job.

One could argue that for the same query with 1 TB of data, the results may be different. I didn't run that test. But, even a cursory look at the two code paths shows that the copier path is still more efficient, just looking at per-row costs, than the generated code.

The reason is simple: the generated code to copy a value is a degenerate special case of code designed to fetch values, compute results, and create new columns, perhaps doing some casts along the way. The generated copy code is the fetch step and the save step with the compute part omitted.

For this reason, custom-built copier code will be more efficient than generic load-compute-save code.

Still, would be great for someone to run a test at scale to verify this reasoning.

arina-ielchiieva:
@ilooner please resolve conflicts.

ilooner:
@Ben-Zvi resolved conflicts. They were minor formatting conflicts. I will mark it as ready-to-commit again since there were no changes, please let me know if there are any issues.

PR #1058, opened 2017-11-29
DRILL-6002: Avoid memory copy from direct buffer to heap while spilling to local disk

vrozov:
@paul-rogers Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
@@ -17,15 +17,29 @@
  */
 package org.apache.drill.exec.cache;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
+import java.nio.channels.WritableByteChannel;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.metrics.DrillMetrics;
+import org.apache.drill.exec.proto.UserBitShared;
 import org.apache.drill.exec.record.VectorAccessible;
 import org.apache.drill.exec.record.VectorContainer;
 import org.apache.drill.exec.record.WritableBatch;
 import org.apache.drill.exec.record.selection.SelectionVector2;
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.Timer;
+
+import io.netty.buffer.DrillBuf;
+
+import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Serializes vector containers to an output stream or from
@@ -39,51 +53,84 @@ public class VectorSerializer {
    * objects to an output stream.
    */
 
-  public static class Writer {
+  public static class Writer implements Closeable
+  {
+    static final MetricRegistry metrics = DrillMetrics.getRegistry();
+    static final String WRITER_TIMER = MetricRegistry.name(VectorAccessibleSerializable.class, "writerTime");
 
-    private final OutputStream stream;
-    private final BufferAllocator allocator;
-    private boolean retain;
+    private final WritableByteChannel channel;
+    private final OutputStream output;
     private long timeNs;
+    private int bytesWritten;
 
-    public Writer(BufferAllocator allocator, OutputStream stream) {
-      this.allocator = allocator;
-      this.stream = stream;
+    private Writer(WritableByteChannel channel) {
+      this.channel = channel;
+      output = Channels.newOutputStream(channel);
     }
 
-    public Writer retain() {
-      retain = true;
-      return this;
-    }
-
-    public Writer write(VectorAccessible va) throws IOException {
+    public int write(VectorAccessible va) throws IOException {
       return write(va, null);
     }
 
     @SuppressWarnings("resource")
-    public Writer write(VectorAccessible va, SelectionVector2 sv2) throws IOException {
+    public int write(VectorAccessible va, SelectionVector2 sv2) throws IOException {
+      checkNotNull(va);
       WritableBatch batch = WritableBatch.getBatchNoHVWrap(
           va.getRecordCount(), va, sv2 != null);
-      return write(batch, sv2);
+      try {
+        return write(batch, sv2);
+      } finally {
+        batch.clear();
+      }
     }
 
-    public Writer write(WritableBatch batch, SelectionVector2 sv2) throws IOException {
-      VectorAccessibleSerializable vas;
-      if (sv2 == null) {
-        vas = new VectorAccessibleSerializable(batch, allocator);
-      } else {
-        vas = new VectorAccessibleSerializable(batch, sv2, allocator);
+    public int write(WritableBatch batch, SelectionVector2 sv2) throws IOException {
+      checkNotNull(batch);
+      checkNotNull(channel);
+      final Timer.Context timerContext = metrics.timer(WRITER_TIMER).time();
+
+      final DrillBuf[] incomingBuffers = batch.getBuffers();
+      final UserBitShared.RecordBatchDef batchDef = batch.getDef();
+      bytesWritten = batchDef.getSerializedSize();
+
+      /* Write the metadata to the file */
+      batchDef.writeDelimitedTo(output);
+
+      /* If we have a selection vector, dump it to file first */
+      if (sv2 != null) {
+        final int dataLength = sv2.getCount() * SelectionVector2.RECORD_SIZE;
+        ByteBuffer buffer = sv2.getBuffer(false).nioBuffer(0, dataLength);
+        while (buffer.remaining() > 0) {
+          bytesWritten += channel.write(buffer);
+        }
       }
-      if (retain) {
-        vas.writeToStreamAndRetain(stream);
-      } else {
-        vas.writeToStream(stream);
+
+      /* Dump the array of ByteBuf's associated with the value vectors */
+      for (DrillBuf buf : incomingBuffers) {
+        /* dump the buffer into the OutputStream */
+        ByteBuffer buffer = buf.nioBuffer();
+        while (buffer.remaining() > 0) {
+          bytesWritten += channel.write(buffer);
+        }
       }
-      timeNs += vas.getTimeNs();
-      return this;
+
+      timeNs += timerContext.stop();
+      return bytesWritten;
     }
 
-    public long timeNs() { return timeNs; }
+    @Override
+    public void close() throws IOException {
+      if (!channel.isOpen()) {
+        return;
+      }
+      channel.close();
+    }
+
+    public long time(TimeUnit unit) {
+      return unit.convert(timeNs, TimeUnit.NANOSECONDS);
+    }
+
+    public int getBytesWritten() { return bytesWritten; }
   }
 
   /**
@@ -111,8 +158,8 @@ public class VectorSerializer {
     public long timeNs() { return timeNs; }
   }
 
-  public static Writer writer(BufferAllocator allocator, OutputStream stream) {
-    return new Writer(allocator, stream);
+  public static Writer writer(WritableByteChannel channel) throws IOException {
+    return new Writer(channel);
   }
 
   public static Reader reader(BufferAllocator allocator, InputStream stream) {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/SpillSet.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/SpillSet.java
@@ -18,13 +18,15 @@
 package org.apache.drill.exec.physical.impl.spill;
 
 import java.io.BufferedInputStream;
-import java.io.BufferedOutputStream;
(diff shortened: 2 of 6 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
```diff
@@ -40,20 +52,18 @@
...
 
-    public Writer(BufferAllocator allocator, OutputStream stream) {
-      this.allocator = allocator;
-      this.stream = stream;
-    }
-
-    public Writer retain() {
-      retain = true;
-      return this;
+    private Writer(SpillSet spillSet, String path) throws IOException {
```
paul-rogers: Before this class was independent of the `SpillSet`. Now, it can be used only with that class. Not sure if this is an improvement.

The only thing it seems we need from `SpillSet` is to update the write byte count. Perhaps define an interface here with just the needed methods. Then, let `SpillSet` implement that interface. That gets what you need without introducing extra dependencies.

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
```diff
@@ -62,27 +72,65 @@ public Writer write(VectorAccessible va) throws IOException {
...
-        vas = new VectorAccessibleSerializable(batch, sv2, allocator);
-      }
-      if (retain) {
-        vas.writeToStreamAndRetain(stream);
-      } else {
-        vas.writeToStream(stream);
+      return write(batch, sv2, false);
+    }
+
+    public Writer write(WritableBatch batch, SelectionVector2 sv2, boolean retain) throws IOException {
```
paul-rogers: Retain is never actually used. That's why, in the earlier version, it was a flag that was not passed into the write method. Further, whether retain is not likely to be decided batch-by-batch, rather it is a general policy set at the writer level. So, if we don't like the `retain()` method, perhaps pass this option into the constructor rather than into each `write()` call.

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
```diff
@@ -62,27 +72,65 @@ public Writer write(VectorAccessible va) throws IOException {
...
-      timeNs += vas.getTimeNs();
       return this;
     }
 
+    public long close() throws IOException {
+      if (!channel.isOpen()) {
+        return 0;
+      }
+      long writeSize = spillSet.getPosition(channel);
+      spillSet.tallyWriteBytes(writeSize);
```
paul-rogers: This seems awkward. Before, this class was a utility that wrapped an existing write mechanism. Here, this class basically becomes the writer itself -- a fancy file writer that manages its own channel. That seems like a good evolution.

But, it seems that updating spill set metrics is not needed here. Instead, someone will call this close method. That method can first update the spill set metrics based on the total write bytes for this stream. That is, this writer can do the writing, manage the channel, and accumulate its own write bytes. The caller decides what to do with that info (add it to the spill set's metrics.)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/SpillSet.java
```diff
@@ -331,6 +337,43 @@ public long getReadBytes(InputStream inputStream) {
     }
   }
 
+  private static class WritableByteChannelImpl implements WritableByteChannel
+  {
+    private static final int TRANSFER_SIZE = 32 * 1024;
+
+    private OutputStream out;
+    private final byte buffer[] = new byte[TRANSFER_SIZE];
```
paul-rogers: Unfortunately, this is one of the problems that the current design was intended to solve. As we can see in the hash agg code; multiple of these objects are active in memory at any one time. But, no more than one is ever writing. By putting the buffer here, we create one per (potentially thousands) of retained channels. But, by putting the buffer on the allocator, we hold one buffer per operator.

Being the guy who wrote the current allocator code, I realize it is clunky. But, it is clunky because we favored memory frugality over elegance. (If there is a cleaner way to have one buffer per thread, I'm all for it!)

Isn't the purpose of this change to avoid memory buffers? Seems, for the above reasons, we're moving backwards...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/SpillSet.java
```diff
@@ -419,6 +462,13 @@ public SpillSet(DrillConfig config, FragmentHandle handle, PhysicalOperator popC
         operName, handle.getMajorFragmentId(), popConfig.getOperatorId(), handle.getMinorFragmentId());
   }
 
+  @VisibleForTesting
+  public SpillSet(String baseDir) {
```
paul-rogers: We're trying to use the Hadoop `Path` class as our preferred way to pass file and directory names around.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/BatchGroup.java
```diff
@@ -280,17 +278,14 @@ private void closeInputStream() throws IOException {
...
+    public void closeWriter() throws IOException {
+      if (writer == null) {
+        return;
       }
-      long writeSize = spillSet.getPosition(outputStream);
-      spillSet.tallyWriteBytes(writeSize);
-      outputStream.close();
-      outputStream = null;
+      long writeSize = writer.close();
+      long timeNs = writer.timeNs();
```
paul-rogers: See comment above. Updating spill set here decouples the writer from the spill set. The `BatchGroup` already knows about the spill set.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/BatchGroup.java
```diff
@@ -249,7 +247,7 @@ public void close() throws IOException {
         ex = e;
       }
       try {
-        closeOutputStream();
+        closeWriter();
```
paul-rogers: Just as a matter of record, this abstraction is rather awkward. It tries to be three things: 1) a writer for a spill file, 2) a reader for a spill file and 3) the tombstone for the spill file.

It would be much better to have separate reader and writer classes that come and go, with this class just representing the spill file over its lifetime. There was, however, no easy way to make that change and preserve existing code. Since I'd already changed more code than I really wanted, I left well enough alone. (See the "unmanaged" version for the original code, which was even more murky.)

Now that you've modified the Writer to be the batch writer, I wonder if we should revisit the issue rather than preserving the old, muddy, semantics of this class.

exec/java-exec/src/test/java/org/apache/drill/exec/cache/TestBatchSerialization.java
```diff
@@ -117,11 +117,12 @@ public void testNullableType(MinorType type) throws IOException {
   private void verifySerialize(SingleRowSet rowSet, SingleRowSet expected) throws IOException {
 
     File dir = OperatorFixture.getTempDir("serial");
-    File outFile = new File(dir, "serialze.dat");
-    try (OutputStream out = new BufferedOutputStream(new FileOutputStream(outFile))) {
-      VectorSerializer.writer(fixture.allocator(), out)
-        .write(rowSet.container(), rowSet.getSv2());
-    }
+    VectorSerializer.writer(new SpillSet(dir.getAbsolutePath()),  "serialize.dat")
```
paul-rogers: The `File dir` line should change to use Tim's new directory watcher feature.

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
```diff
@@ -62,27 +72,65 @@ public Writer write(VectorAccessible va) throws IOException {
...
+        for (DrillBuf buf : incomingBuffers) {
+          /* dump the buffer into the OutputStream */
+          channel.write(buf.nioBuffer());
+        }
+
+        timeNs += timerContext.stop();
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      } finally {
+        if (!retain) {
```
paul-rogers: Frankly, the caller should decide what to do with the SV2 and the batch. Much simpler to just do:

```
writer.write(batch, sv2);
batch.clear();
sv2.clear();
```

Or, provide a `writeAndClear()` method if we think it is worth pulling those two lines into the write method.

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
```diff
@@ -62,27 +72,65 @@ public Writer write(VectorAccessible va) throws IOException {
...
+        /* If we have a selection vector, dump it to file first */
+        if (sv2 != null) {
+          final int dataLength = sv2.getCount() * SelectionVector2.RECORD_SIZE;
+          channel.write(sv2.getBuffer(false).nioBuffer(0, dataLength));
+        }
+
+        /* Dump the array of ByteBuf's associated with the value vectors */
+        for (DrillBuf buf : incomingBuffers) {
+          /* dump the buffer into the OutputStream */
+          channel.write(buf.nioBuffer());
```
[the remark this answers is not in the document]
vrozov: I agree that we should only write necessary payload and avoid spilling unused buffers. Note that `channel.write()` writes only bytes written to a `ByteBuffer` (`ByteBuffer.remaining()`) and not the whole allocated buffer. Do you mean that there are bytes written to a buffer but they should not be spilled? In this case, I'd suggest limiting the scope of this PR to using `WritableByteChannel` to avoid memory copy from off-heap during spill to local files and handling extra bytes in a separate JIRA/PR.

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
vrozov: The `Writer` in `VectorSerializer` is used only for spilling.  I'd suggest moving `VectorSerializer` to the `spill`package to make it more explicit and keep the dependency on the `SpillSet`. I am not sure if there is a need for generic `Writer` that can handle multiple use cases. I think that `Writer` needs to be optimized to handle spill use case.

`SpillSet` is also needed to get `WritableByteChannel` and encapsulates what `Writer` uses from operators.

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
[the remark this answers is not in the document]
paul-rogers: I believe Drill uses the `writerIndex` of `AbstractByteBuf` to indicate the end of the written portion. So, we want to write that number of bytes. If that's what this code does, we're good.

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
paul-rogers: The `Writer` is a wrapper to make `VectorSerializer` easier to use. It is *presently* only used in spilling (that's the project that created it.) But, the intention is for it to be general purpose. By binding int to `SpillSet`, it then becomes the easy-to-use interface for spilling, but other code must use the clunkier prior interface.

Not a big issue; just trying to pass along the design intent when the code was created...

exec/java-exec/src/main/java/org/apache/drill/exec/cache/VectorSerializer.java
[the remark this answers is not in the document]
vrozov: I believe we're good as `buf.nioBuffer()` correctly sets ByteBuffer offset and length using `readerIndex()` and `readableBytes()`.

vrozov:
@paul-rogers Please review

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/SpillSet.java
```diff
@@ -107,7 +107,7 @@
      * nodes provide insufficient local disk space)
      */
 
-    private static final int TRANSFER_SIZE = 32 * 1024;
+    private static final int TRANSFER_SIZE = 1024 * 1024;
```
paul-rogers: Is a 1MB buffer excessive? The point of a buffer is to ensure we write in units of a disk block. For the local file system, experience showed no gain after 32K. In the MapR FS, each write is in units of 1 MB. Does Hadoop have a preferred size?

Given this variation, if we need large buffers, should we choose a buffer size based on the underlying file system? For example, is there a preferred size for S3?

32K didn't seem large enough to worry about, even if we had 1000 fragments busily spilling. But 1MB? 1000 * 1 MB = 1GB, which starts becoming significant, especially in light of our efforts to reduce heap usage. Should we worry?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/SpillSet.java
```diff
@@ -104,9 +107,17 @@
      * nodes provide insufficient local disk space)
      */
 
+    // The buffer size is calculated as LCM of the Hadoop internal checksum buffer (9 * checksum length), where
```
vrozov: @paul-rogers Changed `TRANSFER_SIZE` to 72K. It is calculated as LCM between Hadoop internal checksum buffer and MapR FS page size to make writes aligned on internal buffer boundaries.

PR #1060, opened 2017-12-04
DRILL-5846: Improve parquet performance for Flat Data Types

sachouche:
Performance improvements for the Parquet Scanner (Flat Data Types). The are two flags to control this performance enhancement (disabled by default):
Option I -
Config Name: store.parquet.flat.reader.bulk
Config Type  : boolean
Description   : Enables bulk processing to minimize memory checks and improve JVM HotSpot optimizations

Option II -
Config Name: scan.optimized.implicit.columns 
Config Type  : boolean
Description   : Memory optimization when storing duplicate value (implicit columns have duplicate values within a batch). Code profiling indicated this step represented one third of Parquet processing (when implicit columns are processed).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -311,6 +311,10 @@ public final class ExecConstants {
 
   public static final OptionValidator COMPILE_SCALAR_REPLACEMENT = new BooleanValidator("exec.compile.scalar_replacement");
 
+  // Controls whether to enable bulk parquet reader processing
+  public static final String PARQUET_FLAT_READER_BULK = "store.parquet.flat.reader.bulk";
+  public static final OptionValidator PARQUET_FLAT_READER_BULK_VALIDATOR = new BooleanValidator(PARQUET_FLAT_READER_BULK);
+
   public static final String JSON_ALL_TEXT_MODE = "store.json.all_text_mode";
   public static final BooleanValidator JSON_READER_ALL_TEXT_MODE_VALIDATOR = new BooleanValidator(JSON_ALL_TEXT_MODE);
   public static final BooleanValidator JSON_EXTENDED_TYPES = new BooleanValidator("store.json.extended_types");
@@ -321,6 +325,7 @@ public final class ExecConstants {
   public static final String JSON_READER_PRINT_INVALID_RECORDS_LINE_NOS_FLAG = "store.json.reader.print_skipped_invalid_record_number";
   public static final BooleanValidator JSON_READER_PRINT_INVALID_RECORDS_LINE_NOS_FLAG_VALIDATOR = new BooleanValidator(JSON_READER_PRINT_INVALID_RECORDS_LINE_NOS_FLAG);
   public static final DoubleValidator TEXT_ESTIMATED_ROW_SIZE = new RangeDoubleValidator("store.text.estimated_row_size_bytes", 1, Long.MAX_VALUE);
+
 
   /**
    * Json writer option for writing `NaN` and `Infinity` tokens as numbers (not enclosed with double quotes)
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
@@ -148,6 +148,7 @@ public class SystemOptionManager extends BaseOptionManager implements AutoClosea
       new OptionDefinition(ExecConstants.PARQUET_PAGEREADER_BUFFER_SIZE_VALIDATOR),
       new OptionDefinition(ExecConstants.PARQUET_PAGEREADER_USE_FADVISE_VALIDATOR),
       new OptionDefinition(ExecConstants.PARQUET_READER_INT96_AS_TIMESTAMP_VALIDATOR),
+      new OptionDefinition(ExecConstants.PARQUET_FLAT_READER_BULK_VALIDATOR),
       new OptionDefinition(ExecConstants.JSON_READER_ALL_TEXT_MODE_VALIDATOR),
       new OptionDefinition(ExecConstants.JSON_WRITER_NAN_INF_NUMBERS_VALIDATOR),
       new OptionDefinition(ExecConstants.JSON_READER_NAN_INF_NUMBERS_VALIDATOR),
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ParquetRecordReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ParquetRecordReader.java
@@ -90,6 +90,7 @@ public class ParquetRecordReader extends AbstractRecordReader {
   public boolean useFadvise;
   public boolean enforceTotalSize;
   public long readQueueSize;
+  public boolean useBulkReader;
 
   @SuppressWarnings("unused")
   private String name;
@@ -161,6 +162,7 @@ public class ParquetRecordReader extends AbstractRecordReader {
       ParquetMetadata footer,
       List<SchemaPath> columns,
       ParquetReaderUtility.DateCorruptionStatus dateCorruptionStatus) throws ExecutionSetupException {
+
     this.name = path;
     this.hadoopPath = new Path(path);
     this.fileSystem = fs;
@@ -171,20 +173,14 @@ public class ParquetRecordReader extends AbstractRecordReader {
     this.dateCorruptionStatus = dateCorruptionStatus;
     this.fragmentContext = fragmentContext;
     this.numRecordsToRead = numRecordsToRead;
-    useAsyncColReader =
-        fragmentContext.getOptions().getOption(ExecConstants.PARQUET_COLUMNREADER_ASYNC).bool_val;
-    useAsyncPageReader =
-        fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_ASYNC).bool_val;
-    useBufferedReader =
-        fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_USE_BUFFERED_READ).bool_val;
-    bufferedReadSize =
-        fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_BUFFER_SIZE).num_val.intValue();
-    useFadvise =
-        fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_USE_FADVISE).bool_val;
-    readQueueSize =
-        fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_QUEUE_SIZE).num_val;
-    enforceTotalSize =
-        fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_ENFORCETOTALSIZE).bool_val;
+    this.useAsyncColReader = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_COLUMNREADER_ASYNC).bool_val;
+    this.useAsyncPageReader = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_ASYNC).bool_val;
+    this.useBufferedReader = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_USE_BUFFERED_READ).bool_val;
+    this.bufferedReadSize = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_BUFFER_SIZE).num_val.intValue();
+    this.useFadvise = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_USE_FADVISE).bool_val;
+    this.readQueueSize = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_QUEUE_SIZE).num_val;
+    this.enforceTotalSize = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_PAGEREADER_ENFORCETOTALSIZE).bool_val;
+    this.useBulkReader = fragmentContext.getOptions().getOption(ExecConstants.PARQUET_FLAT_READER_BULK).bool_val;
 
     setColumns(columns);
   }
@@ -229,6 +225,13 @@ public class ParquetRecordReader extends AbstractRecordReader {
 
   public FragmentContext getFragmentContext() {
     return fragmentContext;
+  }
+
+  /**
+   * @return true if Parquet reader Bulk processing is enabled; false otherwise
+   */
+  public boolean useBulkReader() {
+    return useBulkReader;
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenAbstractEntryReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenAbstractEntryReader.java
@@ -1,0 +1,265 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.store.parquet.columnreaders;
+
+import java.nio.ByteBuffer;
+
+import org.apache.drill.exec.store.parquet.columnreaders.VarLenColumnBulkInput.ColumnPrecisionInfo;
+import org.apache.drill.exec.store.parquet.columnreaders.VarLenColumnBulkInput.PageDataInfo;
+
+import io.netty.buffer.DrillBuf;
+
+/** Abstract class for sub-classes implementing several algorithms for loading a Bulk Entry */
+abstract class VarLenAbstractEntryReader {
+
+  /** byte buffer used for buffering page data */
+  protected final ByteBuffer buffer;
+  /** Page Data Information */
+  protected final PageDataInfo pageInfo;
+  /** expected precision type: fixed or variable length */
+  protected final ColumnPrecisionInfo columnPrecInfo;
+  /** Bulk entry */
+  protected final VarLenColumnBulkEntry entry;
+
+  /**
+   * CTOR.
+   * @param buffer byte buffer for data buffering (within CPU cache)
+   * @param pageInfo page being processed information
+   * @param columnPrecInfo column precision information
+   * @param entry reusable bulk entry object
+   */
+  VarLenAbstractEntryReader(ByteBuffer buffer,
+    PageDataInfo pageInfo,
+    ColumnPrecisionInfo columnPrecInfo,
+    VarLenColumnBulkEntry entry) {
+
+    this.buffer = buffer;
+    this.pageInfo = pageInfo;
+    this.columnPrecInfo = columnPrecInfo;
+    this.entry = entry;
+  }
+
+  /**
+   * @param valuesToRead maximum values to read within the current page
+   * @return a bulk entry object
+   */
+  abstract VarLenColumnBulkEntry getEntry(int valsToReadWithinPage);
+
+  /**
+   * Indicates whether to use bulk processing
+   */
+  protected final boolean bulkProcess() {
+    return columnPrecInfo.bulkProcess;
+  }
+
+  /**
+   * Loads new data into the buffer if empty or the force flag is set.
+   *
+   * @param force flag to force loading new data into the buffer
+   */
+  protected final boolean load(boolean force) {
+
+    if (!force && buffer.hasRemaining()) {
+      return true; // NOOP
+    }
+
+    // We're here either because the buffer is empty or we want to force a new load operation.
+    // In the case of force, there might be unprocessed data (still in the buffer) which is fine
+    // since the caller updates the page data buffer's offset only for the data it has consumed; this
+    // means unread data will be loaded again but this time will be positioned in the beginning of the
+    // buffer. This can happen only for the last entry in the buffer when either of its length or value
+    // is incomplete.
+    buffer.clear();
+
+    final int bufferCapacity = VarLenBulkPageReader.BUFF_SZ;
+    final int remaining = remainingPageData();
+    final int toCopy = remaining > bufferCapacity ? bufferCapacity : remaining;
+
+    if (toCopy == 0) {
+      return false;
+    }
+
+    pageInfo.pageData.getBytes(pageInfo.pageDataOff, buffer.array(), buffer.position(), toCopy);
+
+    buffer.limit(toCopy);
+
+    // At this point the buffer position is 0 and its limit set to the number of bytes copied.
+
+    return true;
+  }
+
+  /**
+   * @return remaining data in current page
+   */
+  protected final int remainingPageData() {
+    return pageInfo.pageDataLen - pageInfo.pageDataOff;
+  }
+
+  /**
+   * @param buff source buffer
+   * @param pos start position
+   * @return an integer encoded as a low endian
+   */
+  protected final int getInt(final byte[] buff, final int pos) {
+    return DrillBuf.getInt(buff, pos);
+  }
+
+  /**
+   * Copy data; expects 8bytes PADDING for source and target
+   *
+   * @param src source buffer
+   * @param srcIndex source index
+   * @param dest destination buffer
+   * @param destIndex destination buffer
+   * @param length length to copy (in bytes)
+   */
+  static void vlCopy(byte[] src, int srcIndex, byte[] dest, int destIndex, int length) {
+    if (length <= 8) {
+      DrillBuf.putLong(src, srcIndex, dest, destIndex);
+    } else {
+      vlCopyGTLongWithPadding(src, srcIndex, dest, destIndex, length);
+    }
+  }
+
+  /**
+   * Copy data; no PADDING needed for source and target (though less efficient than the copy with
+   * padding as more comparisons need to be made).
+   *
+   * @param src source buffer
+   * @param srcIndex source index
+   * @param dest destination buffer
+   * @param destIndex destination buffer
+   * @param length length to copy (in bytes)
+   */
+  static void vlCopyNoPadding(byte[] src, int srcIndex, byte[] dest, int destIndex, int length) {
+    if (length <= 8) {
+      vlCopyLELongNoPadding(src, srcIndex, dest, destIndex, length);
+    } else {
+      vlCopyGTLongNoPadding(src, srcIndex, dest, destIndex, length);
+    }
+  }
+
+  private static void vlCopyGTLongWithPadding(byte[] src, int srcIndex, byte[] dest, int destIndex, int length) {
+    final int bulkCopyThreshold = 24;
+    if (length < bulkCopyThreshold) {
+      _vlCopyGTLongWithPadding(src, srcIndex, dest, destIndex, length);
+
+    } else {
+      System.arraycopy(src, srcIndex, dest, destIndex, length);
+    }
+  }
+
+  private static void _vlCopyGTLongWithPadding(byte[] src, int srcIndex, byte[] dest, int destIndex, int length) {
+    final int numLongEntries = length >> 3;
+    assert numLongEntries < 3;
+
+    final int remaining = length & 0x7;
+    int prevCopied      = 0;
+
+    if (numLongEntries == 1) {
+      DrillBuf.putLong(src, srcIndex, dest, destIndex);
+      prevCopied = 1 << 3;
+
+    } else {
+      DrillBuf.putLong(src, srcIndex, dest, destIndex);
+      DrillBuf.putLong(src, srcIndex + DrillBuf.LONG_NUM_BYTES, dest, destIndex + DrillBuf.LONG_NUM_BYTES);
+      prevCopied = 1 << 4;
+    }
+
+    if (remaining > 0) {
+      final int srcPos  = srcIndex  + prevCopied;
+      final int destPos = destIndex + prevCopied;
+
+      DrillBuf.putLong(src, srcPos, dest, destPos);
+    }
+  }
+
+  private static final void vlCopyLELongNoPadding(byte[] src, int srcIndex, byte[] dest, int destIndex, int length) {
+    if (length == 1) {
+      dest[destIndex] = src[srcIndex];
+
+    } else if (length == 2) {
+      DrillBuf.putShort(src, srcIndex, dest, destIndex);
+
+    } else if (length == 3) {
+      dest[destIndex] = src[srcIndex];
+      DrillBuf.putShort(src, srcIndex+1, dest, destIndex+1);
+
+    } else if (length == 4) {
+      DrillBuf.putInt(src, srcIndex, dest, destIndex);
+
+    } else if (length == 5) {
+      dest[destIndex] = src[srcIndex];
+      DrillBuf.putInt(src, srcIndex+1, dest, destIndex+1);
+
+    } else if (length == 6) {
+      DrillBuf.putShort(src, srcIndex, dest, destIndex);
+      DrillBuf.putInt(src, srcIndex+2, dest, destIndex+2);
+
+    } else if (length == 7) {
+      dest[destIndex] = src[srcIndex];
+      DrillBuf.putShort(src, srcIndex+1, dest, destIndex+1);
+      DrillBuf.putInt(src, srcIndex+3, dest, destIndex+3);
+
+    } else {
+      DrillBuf.putLong(src, srcIndex, dest, destIndex);
+    }
+  }
+
+  private static final void vlCopyGTLongNoPadding(byte[] src, int srcIndex, byte[] dest, int destIndex, int length) {
+    final int numLongEntries = length >> 3;
+    final int remaining      = length & 0x7;
+
+    if (numLongEntries == 1) {
+      DrillBuf.putLong(src, srcIndex, dest, destIndex);
+
+    } else if (numLongEntries == 2) {
+      DrillBuf.putLong(src, srcIndex, dest, destIndex);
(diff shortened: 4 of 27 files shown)


exec/vector/.externalToolBuilders/Maven_Ant_Builder.launch
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
paul-rogers: Did this file want to be committed? Looks like an Eclipse-specific launch script...

exec/vector/src/main/codegen/templates/FixedValueVectors.java
```diff
@@ -874,6 +880,46 @@ public void setSafe(int index, BigDecimal value) {
...
+    }
+
+    /**
+     * Copies the bulk input into this value vector and extends its capacity if necessary. The callback
+     * mechanism allows decoration as caller is invoked for each bulk entry.
+     *
+     * @param input bulk input
+     * @param callback a bulk input callback object (optional)
+     */
+    public <T extends VLBulkEntry> void setSafe(VLBulkInput<T> input, VLBulkInput.BulkInputCallback<T> callback) {
```
paul-rogers: Seems this code is unique to Parquet (in which data is columnar.) Should it be part of the vector itself? Or, should it be in the Parquet-specific vector writers?

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -296,26 +655,75 @@ public void transferTo(Nullable${minor.class}Vector target){
...
+    }
+    </#if>
     bits.splitAndTransferTo(startIndex, length, target.bits);
     values.splitAndTransferTo(startIndex, length, target.values);
     <#if type.major == "VarLen">
     target.mutator.lastSet = length - 1;
     </#if>
   }
 
+  public ${minor.class}Vector convertToRequiredVector(){
```
paul-rogers: Is this semantically valid?

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -456,55 +818,191 @@ public void get(int index, Nullable${minor.class}Holder holder){
...
+    }
+  }
+  </#if>
+
+//-----------------------------------------------------------------------------
+// Mutator inner classes
+//-----------------------------------------------------------------------------
+
+  /** Abstract mutator class */
+  public abstract class Mutator extends BaseDataValueVector.BaseMutator implements NullableVectorDefinitionSetter<#if type.major = "VarLen">, VariableWidthVector.VariableWidthMutator</#if> {
```
paul-rogers: We should discuss in person. We have two conflicting threads. My PR will render mutators obsolete. Yours will double-down on the mutators. I suspect the result will be confusion for the other developers.

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -804,6 +1406,310 @@ public void fromNotNullable(${minor.class}Vector srce) {
       setValueCount(valueCount);
     }
   }
+
+  <#if type.major == "VarLen" && minor.class == "VarChar">
```
paul-rogers: Given that is (and much other code) is for a single class, does it really need to be generated via a template?

paul-rogers:
This PR is a tough one. We generally like to avoid design discussions in PRs; but I'm going to violate that rule.

This PR is based on the premise that each vector has all the information needed to pull values from a `VLBulkEntry`. The bulk entry reads a given number of values. Presumably this is faster than having the source of the values set values one-by-one.

A separate project is confronting a limitation of Drill: that readers can create batches of very large sizes. Why? Readers choose to read x records. Records can be as small as a few bytes or as large as many MB. (Large records are common in document data sources.) Suppose we decide to read 4K records (a common size) and the records are 1 MB each. The batch needs 4 GB of data. But, the "managed" sort and hash agg operators often get only about 40 MB. Since 4 GB batch > 40 MB working memory, queries die.

If we knew that each record was y bytes, we could compute the record count, x = target batch size / record size. But, for most readers (including Parquet), Drill cannot know the record size until after the record is read. Hence, the entire premise of setting a fixed read size is unworkable. In short, unlike a classic relational DB, Drill cannot predict record sizes, Drill can only react to the size of records as they are read.

Suppose the design were modified so that vector V read either up to x values or until it reaches some maximum vector size. The problem then becomes that, if the record has two columns, V and W, they cannot coordinate. (Specifically, the inner loop in vector V should not have visibility into the state of vector W.) V may read, say, 16K values of 1K each and become full. (For reasons discussed elsewhere, 16 MB is a convenient maximum vector size.) But if column W is 4X wider, at 4K bytes, it must read the same 16K records already read by X, but will require 64 MB, exceeding Y's own size target by 4X.

Suppose the row has 1000 columns. (Unusual, but not unheard of.) Then, we have 1000 bulk readers all needing to coordinate. Maybe each one stays within its own size limit, growing to only 8 MB. But, the aggregate total of 1000 columns at 8 MB each is 8 GB, which exceeds the overall batch size target. (The actual target will be set by the planner, but low 10s of MB is a likely limit.)

To work around this, the bulk load mechanism must know the sizes of all columns it will read in a batch and compute a row count that causes the aggregate sum of values to be below both the per-vector and overall batch size targets. As discussed, since Drill cannot predict the future, and Parquet provides no per-row size metadata, this approach cannot work in the general case (though, with good guesses, it can work in, say, some percentage of "typical" queries -- the approach the Drill currently uses.)

Maybe the solution can read in small bursts, maybe 100 rows, to sense if any vector gets too large before doing another 100 values into each vector. Still with any value larger than 1 for the "burst count", it is easy to get back into the situation of an oversize batch if row sizes are not uniform. 

Since Parquet makes no guarantees that row sizes are uniform, this is a stochastic game. What are the chances that the next 100 (say) rows behave like the last 100? Any deviation leads to OOM errors and query failures. What percentage of the time will we allow queries to fail with OOM errors to go fast the rest of the time? 10% failure rate for a 100% speed improvement? Is there any correct failure rate other than 0%? This is a hard question. In fact, this is the very problem that the "batch size" project was asked to solve.

We are left with two choices. Either 1) exempt Parquet from row size limits, or 2) devise a new solution that allows size limits to be enforced, if only stochastically. Since we are entertaining this PR, then clearly we've decided to *not* address size issues in Parquet. But, since Parquet is our primary input format, we've decided not to address size issues for ~90% of queries. But, since we do, in fact, have a parallel project to limit batch sizes, perhaps we have not, in fact, exempted Parquet. This is quite the dilemma! 

For background, the "batch size" project, to be used for other readers, solves the problem by monitoring total batch size row by row. When a row causes the batch (or any individual vector) size to be exceeded, "rollover" occurs to create a new batch, and the full batch is sent downstream. Lots of complexity comes from the fact that overflow can occur in the middle of a rows, perhaps inside a deeply nested structure. The mechanism makes overflow transparent to the reader; the reader simply asks if it can write a row, writes it, then loops to check again.

It is understood that Parquet is columnar, and the row-by-row approach is perhaps inefficient. So, we need a columnar solution to the batch size problem. Perhaps this project can propose one.

The above description points out another concern. The PR says that this mechanism is for flat rows. Yet, we have yet another project to support nested tables (the so-called "complex types".) Is this PR consistent with that project, or will we need a new, complex type reader that uses, say, the "batch size" mechanism, while the "flat" reader uses this mechanism? Perhaps this project can propose a solution to that issue also.

Would appreciate direction from the PMC and commercial management about their intent and how these three now-conflicting projects are to be reconciled. Once that info is available, we can continue to determine if the code offered here realizes the goals set by project leadership.

paul-rogers:
Always a good idea to suggest an alternative in addition to identifying challenges. I wonder if the code can resolve the questions raised by taking a someone different approach:

1. Rather than vectors pulling in data, create a new abstraction that does so. It loads a heap buffer with values until either a) filling the buffer, or b) reaching a defined value limit.
2. Once all vector buffers are full, make a pass over all columns to determine how much will fit into the target vectors before overflow. (That is, if the buffer holds 1K values, but only 200 of column C fits before overflow, then 200 becomes the vector fill number.) Let's call this N.
3. Make another pass, copying N values from each buffer into the target vectors. If that leaves values in the buffers, hold those values for the next pass (by shifting them down or implementing a circular buffer.)

The above handles vector size limits. It has the advantage of turning "Drill can't predict the future" (can't know value lengths until they are read) into an exercise in examining the past: which values were read into buffers.

It is a bit harder to imagine how to honor batch size limits since the limit is computed over all vectors, not just one. There is the crude-but-effective approach in step 2:

2. In the pass above to compute N (the number of values to copy into vectors), go row by row, accumulating totals (as the "row set loader" does on actual writes.) Once the limit is set, back off by one to get N (the last count that completely fits within a row.)

Now, since the above requires row-by-row calculations, it requires the very work that this PR seeks to avoid (but does not do actual data transfers.)

The question, then, is this: is there a refinement to the above algorithm that can be done non-iteratively (without doing a row-by-row check?)

And, with all the extra machinery needed to honor limits, will we still see a significant speedup relative to a straightforward row-by-row read?

sachouche:
Before I reply to the provided comments I want first to thank both Parth and Paul for taking time to review this Pull Request.
@parthchandra Regarding the High Level Comments

- FS Access Manager
The attached document listed several possible improvements in terms of CPU, Partitioning (changes to the query plan), and IO layer. I didn't want to club all the changes in a single pull request; instead I started with the Pull Request with the highest priority (in terms of performance improvement) which is CPU related enhancements.

- Drill Memory Package
I understand where you're coming from; I have analyzed this task and came up with a solution which I felt was the cleanest. Let me take you through my thinking process and the alternative solutions that I have considered:
a) Why did I use this low level APIs?
The Bulk Parquet reader uses the Drill Buffer APIs to copy data chunks to the CPU cache; this means that memory sanity checks are always performed. During code profiling I have noticed the Hotspot was using too many instructions (byte based assembly was used); I have researched ways to improve this and found some Hadoop and Compression java code which utilized 64bit access instructions when manipulating byte arrays (e.g., [Snappy](https://github.com/airlift/aircompressor/blob/master/src/main/java/io/airlift/compress/snappy/SnappyRawCompressor.java)).  This makes a huge difference

b) Safety
As indicated previously, native data is always accessed through the Drill Buffer APIs (that is, accessed data is sanitized). During bulk processing, the new utility has additional checks which are triggered when assertions are enabled. This means all test-suite and QA tests (which run with assertions On) will be able to catch any faulty memory access; note also the checks are simpler since they only pertain to byte arrays boundary checks.

c) Abstraction
I agree that the Drill code base should follow a common code path when accessing native memory (minimize chances of introducing bugs), though I felt the Drill Memory package was the main abstraction since it was centralized and thus easier to test.

d) Alternatives
I have considered using the Netty io.netty.util.internal.PlatformDependent0 to access the Java intrinsics APIs though this class has a package scope and the wrapper public class io.netty.util.internal.PlatformDependent doesn't expose this low level functionality. Thus the work around would have been to create MemoryUtils class under the io.netty.util.internal package. I didn't think this was a cleaner approach than just using Java's intrinsic APIs directly (note also that Java9 will expose such intrinsics under a public JDK package instead of the sun.misc. Please let me know if you rather prefer the io.netty.util.internal route instead.

- Changes to the Vector Code
I had few discussions with Paul about the need to reconcile DRILL-5657; the agreement was to proceed with the current implementation and then apply refactoring (future releases) as my changes are backward compatible (new public APIs added). 

- Arrow Integration
I will create a task to analyze the impact of the Arrow Integration on the Parquet Reader.

- UserBitShared.proto
This class was updated to deal with the implicit columns optimization. I have added two fields to the NullableVarChar class to capture the logical number of rows and the fact this feature was active. This allowed me to use different Accessor and Mutator implementation. You and I had a discussion about this and the agreement was to:
a) The new optimization is configurable (current default is off); though we might change the default at some point
b) Detect C++ based clients and turn of this optimization (that is, the implicit columns enhancements); I didn't implement this task yet as I forgot the name of the property that you have indicated (this should be quite fast to implement though)

- Testing
Yes, QA helped me to run the test-suite with the two new options enabled. The new changes didn't modify any of the encoding/compression logic but I'll be happy to add any test(s) you deem necessary.. 

- Code Style
Thank you for the pointers; I'll stick to your feedback in future changes to ease the review process.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VLAbstractEntryReader.java
```diff
@@ -0,0 +1,215 @@
...
+    buffer.clear();
+
+    int remaining = remainingPageData();
+    int toCopy    = remaining > buffer.capacity() ? buffer.capacity() : remaining;
+
+    if (toCopy == 0) {
+      return false;
+    }
+
+    pageInfo.pageData.getBytes(pageInfo.pageDataOff, buffer.array(), buffer.position(), toCopy);
```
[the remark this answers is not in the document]
sachouche: @parthchandra I know this is counter intuitive but this path makes it faster and I had VTune to prove it; let me explain it:
- The logic is to fetch small chunks of data from main memory to the L1 cache (using 4k buffers) for processing
- This logic doesn't introduce any extra memory bandwidth penalty since in all cases data has to move from main memory to the CPU and back to main memory (if modified)
- Note that the CPU has optimizations that disable memory flush for private mutated data which fully fit in the cache
- This means the byte array read / write operations is always local (doesn't involve the memory bus)
- Another topic could be any extra CPU instructions we might be using.. usually this is not a problem as long as in the end our logic is avoiding CPU stalls, better utilizing the CPU pipelining, and using more optimal instructions (e.g., 64bit vs 8bit, SIMD, etc)

I would recommend using VTune as it clearly show cases the above assertions.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -471,8 +519,8 @@ public void close() throws Exception {
     container.clear();
     mutator.clear();
     if (currentReader != null) {
-      currentReader.close();
-    }
+    currentReader.close();
+  }
```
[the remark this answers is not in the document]
sachouche: done.

protocol/src/main/protobuf/UserBitShared.proto
```diff
@@ -148,6 +148,8 @@ message SerializedField {
   optional int32 value_count = 4;
   optional int32 var_byte_length = 5;
   optional int32 buffer_length = 7;
+  optional bool is_dup = 8;
+  optional int32 logical_value_count = 9;
```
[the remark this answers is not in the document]
sachouche: Just had an offline conversation with Paul and Kunal. The agreement is that we should introduce a versioning mechanism at the protocol level so that clients could advertise their version identifier; this way the server can use this information to turn on / off features based on the client capabilities. The task is only to create the preliminary versioning infrastructure; more sophistication can be added later on.

sachouche:
@paul-rogers with regard to the design aspects that you brought up:
** Corrections about the proposed Design **
Your analysis somehow assumes the Vector is the one driving the loading operation. This is not the case, a) it is the reader which invokes the bulk save API b) the Vector save method runs a loop requesting the rest of the data. The reader can any time stop the loading phase; currently the reader uses the batch size to make this decision. The plan is to enhance this logic by having a feedback from the Vector save method about memory usage (e.g., save current row set with the condition the memory usage doesn't get X bytes). I believe I have raised this design decision early on and the agreement was to implement the memory batch restrictions in a next Drill release.

** Agile Development **
Through the years I came to appreciate the value of agile development. One needs to prioritize design activities; it is completely valid to start with a design (which is not guaranteed to be perfect) and then refine it overtime as long as the underlying implementation doesn't spill to other modules (encapsulation). This has the merit of showcasing incremental improvements and allowing the developer to get new insight as they have a better understanding.

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -51,6 +57,17 @@
 public final class ${className} extends BaseDataValueVector implements <#if type.major == "VarLen">VariableWidth<#else>FixedWidth</#if>Vector, NullableVector {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(${className}.class);
 
+  /**
+   * Optimization to set contiguous values nullable state in a bulk manner; cannot define this array
+   * within the Mutator class as Java doesn't allow static initialization within a non static inner class.
+   */
+  private static final int DEFINED_VALUES_ARRAY_LEN = 1 << 10;
```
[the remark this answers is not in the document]
sachouche: 1 << 10 is 1024; not sure how did you arrive at 10,000,000,000..

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -51,6 +57,17 @@
...
+  private static final byte[] DEFINED_VALUES_ARRAY  = new byte[DEFINED_VALUES_ARRAY_LEN];
```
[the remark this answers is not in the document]
sachouche: I rather have it within the sub-class which defines the associated functionality (that is nullable DT); the wasted memory is also negligible.

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -68,96 +85,441 @@
...
+  /** Optimization for cases where all values are identical */
+  private boolean duplicateValuesOnly;
+  /** logical number of values */
+  private int logicalNumValues;
+  /** logical value capacity */
+  private int logicalValueCapacity;
+  /** Mutator instance for duplicate values vector */
+
+  /** true if this vector holds the same value albeit repeated */
+  public boolean isDuplicateValsOnly() {
```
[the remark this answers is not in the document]
sachouche: Tried that (that was my first attempt) but the main issue is that the Drill factory creates vectors based on the column metadata alone. This design has the advantage of enabling / rolling-back optimizations transparently from the consumer. I also made sure there is no performance penalty (or at least minimal).

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -68,96 +85,441 @@
...
+    if (!fromAccessor.isNull(fromIndex)) {
+      fromIndex = !from.isDuplicateValsOnly() ? fromIndex : 0;
+      mutator.fillEmpties(thisIndex);
+      bits.copyFrom(fromIndex, thisIndex, from.bits);
+      values.copyFrom(fromIndex, thisIndex, from.values);
+    }
+  }
+
+  public void copyFromSafe(int fromIndex, int thisIndex, ${minor.class}Vector from) {
+    // We currently don't have a use-case where a target dup-vector is provisioned by a non-dup source vector
```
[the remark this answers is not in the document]
sachouche: No error will occur as I added runtime rollback. I am also planning to support multiple different values using dictionary encoding.

exec/memory/base/src/main/java/org/apache/drill/exec/util/MemoryUtils.java
```diff
@@ -0,0 +1,186 @@
...
+//----------------------------------------------------------------------------
+// APIs
+//----------------------------------------------------------------------------
+
+  /**
+   * @param data source byte array
+   * @param index index within the byte array
+   * @return short value starting at data+index
+   */
+  public static short getShort(byte[] data, int index) {
```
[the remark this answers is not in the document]
vrozov: Unless there is a significant performance benefit in introducing two ways to access direct buffers, Drill should have a unified way to work with the off-heap memory preferably using DrillBuf. With bounds checking disabled by default, I don't see why MemoryUtils will outperform DrillBuf.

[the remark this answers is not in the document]
sachouche: Vlad,

- The reader has loaded data from direct memory into CPU level-1 cache through DrillBuf
- The data is stored (4k at a time) in a Java byte[]
- The only reason for using MemoryUtil is to perform 64bit processing (I have previously indicated that many other processing intensive applications such as Snappy and Hadoop Comparison use the same pattern)
- I solely added this logic after noticing the Hotspot was generating 8bit instructions 

Again, the intent is not to access data from DrillBuf (the data has been already loaded into CPU cache); this is mainly an optimization step.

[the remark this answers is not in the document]
vrozov: @sachouche It is not clear why using MemoryUtils will make difference for hotspot in using cache or 8bit instruction set. It will be used for both MemoryUtils and DrillBuf.

exec/memory/base/src/main/java/io/netty/buffer/DrillBuf.java
```diff
@@ -703,7 +703,18 @@ protected void _setLong(int index, long value) {
 
   @Override
   public ByteBuf getBytes(int index, ByteBuf dst, int dstIndex, int length) {
-    udle.getBytes(index + offset, dst, dstIndex, length);
+    final int BULK_COPY_THR = 1024;
```
vrozov: The optimization is not obvious (and is counterintuitive) without more details.

exec/memory/base/src/main/java/org/apache/drill/exec/util/MemoryUtils.java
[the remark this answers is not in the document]
sachouche: Vlad,

Can you please clarify your point? Are you suggesting that every memory access has to be via DrillBuf even if the data is loaded in Java heap and the developer is performing an optimization. If this is your goal, then I would think the danger is that the DrillBuf class will be cluttered and contracts hard to understand:
- Notice that DrillBuf performs expensive sanity checks (boundary and reference count checks); this is due to Drill memory optimizations for minimizing the overall memory usage.
- When the code is processing already load data, these checks are not applicable anymore as this data is private.
- I could have easily moved the new APIs there but I strongly felt this was inappropriate..

exec/memory/base/src/main/java/io/netty/buffer/DrillBuf.java
sachouche: You are right, I'll put more information on this optimization:
- During code profiling, I have noticed that getBytes() doesn't perform well when called with small length (lower than 1k)
- Its throughput improves as the length increases

Analysis :
- Java exposes two intrinsics for writing to direct memory: putByte and copyMemory
- The JVM is able to inline memory access (no function call) for putByte 
- copyMemory is a bulk API and this internally invokes libc memcpy (requires function call)
- The rational is that we are willing to incur a function call if the associated processing is larger than the overhead; this is almost never the case for small memory accesses.

exec/memory/base/src/main/java/org/apache/drill/exec/util/MemoryUtils.java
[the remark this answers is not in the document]
vrozov: The sanity check that can be turned on and off is required, otherwise, there is no difference between Netty `PlatformDependend` and `MemoryUtils` and it will be better to have a unified way independent from Java assertions to turn bounds checking on or off.

My assumption is that Drill either reads from DrillBuf or writes to DrillBuf as data is not supposed to be in the heap, so DrillBuf looks like a natural choice for the functionality.

exec/memory/base/src/main/java/io/netty/buffer/DrillBuf.java
vrozov: My testing shows that `PlatformDependent.copyMemory` outperforms multiple calls to `PlatformDependent.getByte()`/`PlatformDependent.putByte()` starting with 16 bytes. Copying 1024 bytes buffer using `getByte()`/`putByte()` is 10x slower compared to `copyMemory()`.

sachouche: Vlad,

- I had a chat with @bitblender and he explains that Java was invoking a stub (not a function call) to perform copyMemory; he agreed copyMemory will be slower for small buffers and the task was to determine the cutoff point
- My tests (I will send you my test) indicate that a length of 1024bytes is the length were copyMemory starts performing exactly as getByte()

NOTE - I am using JRE 1.8; static buffers initialized once; payload 1MB (1048576bytes) and loop-count of 102400; MacOS High Sierra; 1 thread, 4GB MX, MS

parthchandra:
I feel putting this PR in without finalizing DRILL-6301 is putting the cart before the horse. (BTW, it would help the discussion if the benchmarks were published !). My observation based on profiling I did sometime back is that the performance gains seen here are roughly in line with removing bounds checks. Paul has seen similar gains in the batch sizing project.
Which takes us back to the question, raised by Paul in his first comment, of how we want to reconcile batch sizing and vectorizing of scans; a question we have deferred. If removing bounds checks gets us the same performance gains, then why not would put our efforts in implementing batch sizing with the accompanying elimination in bounds checking. 
I'm mostly not in favor of having MemoryUtils unless you make a compelling argument that it is the only way to save the planet (i.e get the performance you want). I feel operators should not establish the pattern of accessing memory directly. So far, I'm -0 on this as my arguments are mostly high level (and somewhat philosophical). 
Minor nitpick - The prefix VL is not as informative as say, VarLen or VariableLength.

sachouche:
Parth,

- I have attached, within the DRILL-5846, two profiles with latest Apache code and this PR request (bounds checks are off):
  o Used one thread in each run
  o I observe ~3x performance difference when the new logic is turned on
  o The difference is 4x if I include the implicit column optimization (which is not part of this PR)
  o The impact of the new optimizations can be felt when there are many variable length columns

- The rational of trying to approve this PR
   o The optimizations that I have included are local to the Flat Parquet Reader (incapsulated)
   o The logic is backward compatible and turned off by default
   o I have added the new Batch Sizing functionality on top of this PR (columnar processing pattern)
   o The result of DRILL-6301 would only result in a local refactoring step
   o Not being able to add the new code results in a substantial maintenance overhead

exec/java-exec/pom.xml
```diff
@@ -836,6 +836,14 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>2.3.2</version>
+        <configuration>
+            <compilerArgument>-XDignore.symbol.file</compilerArgument>
```
vrozov: Consider using `io.netty.util.internal.PlatformDependent` instead of `sun.misc.Unsafe`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/MemoryUtils.java
```diff
@@ -0,0 +1,199 @@
+/**
```
vrozov: Please do not use javadoc comment.

exec/java-exec/pom.xml
sachouche: Will do!

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/MemoryUtils.java
sachouche: Vlad, can you please explain what you meant? are you only referring for specific members or is it a general comment?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ParquetRecordReader.java
```diff
@@ -161,30 +162,25 @@ public ParquetRecordReader(
...
       List<SchemaPath> columns,
       ParquetReaderUtility.DateCorruptionStatus dateCorruptionStatus) throws ExecutionSetupException {
-    this.name = path;
-    this.hadoopPath = new Path(path);
-    this.fileSystem = fs;
-    this.codecFactory = codecFactory;
-    this.rowGroupIndex = rowGroupIndex;
-    this.batchSize = batchSize;
-    this.footer = footer;
+
```
vrozov: Please avoid changes that affect code style only.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VLAbstractEntryReader.java
```diff
@@ -0,0 +1,214 @@
+/*******************************************************************************
```
vrozov: Please use '/*' and '*/`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/MemoryUtils.java
vrozov: I refer to `/**`, please use `/*`

sachouche:
@parthchandra , @vrozov 

I have done the following modifications:
- Renamed newly added files with the prefix "VL" with "VarLen" as suggested by @parthchandra 
- After talking offline with @vrozov about his objections to the MemoryUtils and also from previous feedbacks from @parthchandra , I have a) deleted this utility, b) exposed the needed functionality under the class DrillBuf (using Netty API's), and c) used the same configuration to control the checks.

@parthchandra , @vrozov, please review this [document](https://docs.google.com/document/d/1BSNem_ItP-Vxlr6auSP_iwwOLM9rwWZYxGwCsXi-IE8/edit?usp=sharing) and the associated JMH code [here](https://github.com/sachouche/drill-jmh). This should provide you with performance benchmarks regarding the bulk approach that I have used in this PR request (please focus on test-3 which is Parquet specific); I will be adding more fine-grained tests with regard to the Memory Access tests (requested by @vrozov).

sachouche:
Thanks both @parthchandra and @vrozov for the detailed review!
- I am using the Drill coding style for Eclipse
- I'll grep the code for other VL prefix that I might have missed; I'll also need to resolve the new conflicts

sachouche:
- Addressed conflicts introduced with the Decimal Support changes
- Refactored few classes that still had the prefix "VL"

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -289,6 +289,10 @@ private ExecConstants() {
 
   public static final OptionValidator COMPILE_SCALAR_REPLACEMENT = new BooleanValidator("exec.compile.scalar_replacement");
 
+  // Controls whether to enable bulk parquet reader processing
+  public static final String PARQUET_FLAT_READER_BULK                    = "store.parquet.flat.reader.bulk";
```
arina-ielchiieva: Could you please fix shift between variable name and equals? Looks like it correspond to the general code style in Drill project. Please consider the same in other changed classes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenAbstractEntryReader.java
```diff
@@ -0,0 +1,266 @@
+/*******************************************************************************
```
arina-ielchiieva: Please license headers, Travis fails. Consider other classes as well.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenAbstractEntryReader.java
```diff
@@ -0,0 +1,266 @@
...
+  protected final VarLenColumnBulkEntry entry;
+
+  /**
+   * CTOR.
+   * @param _buffer byte buffer for data buffering (within CPU cache)
+   * @param _pageInfo page being processed information
+   * @param _columnPrecInfo column precision information
+   * @param _entry reusable bulk entry object
+   */
+  VarLenAbstractEntryReader(ByteBuffer _buffer,
```
arina-ielchiieva: Not sure naming parameters with underscore is according to Java standard. Please consider renaming here and in other classes.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
sachouche: A mentor of mine insisted on code formatting to improve the overall code readability; it is very hard for me to forgo  such a practice.. but I did. I went through the changes and removed all formatting around the equal sign.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenAbstractEntryReader.java
sachouche: I have fixed the headers and modified my maven script to always set "-Dlicense.skip" to true; shouldn't be set by default?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenAbstractEntryReader.java
sachouche: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenAbstractEntryReader.java
vrozov: It should be disabled by default, IMO. There is no need to check license headers all the time. Please either run license check explicitly or rely on Travis.

exec/memory/pom.xml
```diff
@@ -34,4 +34,5 @@
   <modules>
     <module>base</module>
   </modules>
+
```
vrozov: Avoid formatting changes

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
vrozov: While it marginally improves code readability, maintaining the code style that aligns on the longest line is not sustainable as it requires modifications to existing code when a new line is introduced or otherwise code becomes unreadable. My vote is to avoid newly introduced code formatting as well and keep existing Drill style.

exec/memory/pom.xml
sachouche: Removed the new empty line (introduced by accident).

arina-ielchiieva:
@sachouche all unit test passed but I see numerous failures in Functional / Advanced tests, will send you run output in email.

sachouche:
@arina-ielchiieva,
- I have fixed the failing tests; I noticed I missed porting few fixes from the mapr private branch because it contained other features (Parquet batch size batching and implicit columns)
- There were three failing tests in the Advanced tests but all of them failed because of timeouts; the functional tests and unit tests passed

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -520,4 +520,4 @@ private boolean verifyImplcitColumns(int numReaders, List<Map<String, String>> i
 
     return true;
   }
-}
+}
```
vrozov: Please avoid format only changes.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -449,7 +451,7 @@ public void close() throws Exception {
     // been created. Gracefully handle that case.
 
     if (options != null) {
-      options.close();
-    }
+    options.close();
   }
 }
+}
```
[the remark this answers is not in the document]
vrozov: Please address @paul-rogers comment. The new formatting of the close() method is off.

sachouche:
Addressed formatting requests.

PR #1062, opened 2017-12-05
DRILL-6007: Use default refresh timeout (10 seconds) if graceful shut…

arina-ielchiieva:
…down timeout is set to 0.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/index.ftl
+++ b/exec/java-exec/src/main/resources/rest/index.ftl
@@ -185,11 +185,10 @@
       </div>
   </div>
    <script charset="utf-8">
-      var refreshTime = 2000;
+      var refreshTime = 10000;
       var refresh = getRefreshTime();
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -185,11 +185,10 @@
       </div>
   </div>
    <script charset="utf-8">
-      var refreshTime = 2000;
+      var refreshTime = 10000;
       var refresh = getRefreshTime();
```
parthchandra: This is confusing isn't it? The function does not return a value and yet the return value (<undefined>) is stored in a variable that is not used. 
Luckily it's not used anywhere.

arina-ielchiieva:
Agree, I have created Jira for js part improvement where I have indicated some other things that should be improved (https://issues.apache.org/jira/browse/DRILL-6022).

PR #1063, opened 2017-12-07
DRILL-5702: Jdbc Driver Class not found

vdiravka:
- Setting "package.namespace.prefix" to "oadd." by default for all profiles. It can be overridden if necessary within any profile.
- Removing duplicated and redundant excluding of commons-logging packages.


vdiravka:
The issue is enabling "apache-release" profile disabled "default" profile automatically and then `"package.namespace.prefix"` had invalid value `${package.namespace.prefix}`.

parthchandra:
+1 LGTM

PR #1066, opened 2017-12-08
DRILL-3993: Changes to support Calcite 1.15

KulykRoman:
Works with Drill-specific Calcite 1.15 from a branch: https://github.com/KulykRoman/incubator-calcite/commits/DrillCalcite1.15.0_rc0.


exec/java-exec/src/test/resources/record/test_recorditerator.json
```diff
@@ -60,7 +60,7 @@
         @id:2,
         child:1,
         pop:"project",
-        exprs:[ { ref : "`*`", expr : "`*`"} ]
+        exprs:[ { ref : "`**`", expr : "`**`"} ]
```
chunhui-shi: Not sure I understand this '**' thing, can you explain more about this change?

exec/jdbc-all/pom.xml
```diff
@@ -572,7 +572,7 @@
                           This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                         </message>
-                        <maxsize>29000000</maxsize>
+                        <maxsize>31000000</maxsize>
```
chunhui-shi: I played with this branch and I have to change the size of 31000000 to 32000000, it might due to my build environment but we may want to increase it to 32000000.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -351,4 +351,23 @@ SqlNode SqlDropFunction() :
    {
        return new SqlDropFunction(pos, jar);
    }
-}
\ No newline at end of file
+}
+
+<#if !parser.includeCompoundIdentifier >
```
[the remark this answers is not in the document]
vvysotskyi: Actually, it does not cause a new behaviour or functionality, it just helps to preserve old one after changes in Calcite `Parser.jj`. Therefore existing unit tests cover this change.

chunhui-shi:
Can you file a pull request to mapr/incubator-calcite with your changes in https://github.com/KulykRoman/incubator-calcite/commits/DrillCalcite1.15.0_rc0?
So we can review and get these changes into official branch to publish.

vvysotskyi:
@chunhui-shi, this branch is not ready completely. There is opened a pull request on Apache Calcite [CALCITE-2018](https://issues.apache.org/jira/browse/CALCITE-2018) which should be cherry-picked into this branch. When it is merged into Apache master, we will update both branches and create a pull request on mapr/incubator-calcite.

vvysotskyi:
@chunhui-shi, I have made additional fixes in new commits (commits after DRILL-3993: Changes after code review. 3120762). Could you please take a look? 
Also, I have created pull request on incubator-calcite: https://github.com/mapr/incubator-calcite/pull/16

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/AggPruleBase.java
```diff
@@ -82,4 +83,19 @@ protected boolean create2PhasePlan(RelOptRuleCall call, DrillAggregateRel aggreg
...
     return true;
   }
+
+  /**
+   * Returns group-by keys with the remapped arguments for specified aggregate.
+   *
+   * @param groupSet ImmutableBitSet of aggregate rel node, whose group-by keys should be remapped.
+   * @return {@link ImmutableBitSet} instance with remapped keys.
+   */
+  public static ImmutableBitSet remapGroupSet(ImmutableBitSet groupSet) {
```
chunhui-shi: what is the reason we are going this remap with new calcite? 
And if the result is only depended on size of groupSet, we don't really need to iterate through the groupSet.

vvysotskyi: After the changes in CALCITE-1930 in the class `AggregateExpandDistinctAggregatesRule`, this rule started applying more correctly, since in the older version there were checks like this:
```
aggCall.getAggregation() instanceof SqlCountAggFunction
```
but they were replaced by checks like this:
```
final SqlKind aggCallKind = aggCall.getAggregation().getKind();
switch (aggCallKind)
```
So for the cases when instead of Calcite s `SqlCountAggFunction` were used `DrillCalciteSqlAggFunctionWrapper`s this rule changed its behaviour and instead of returning joins of distinct and non-distinct aggregate rel nodes, it returns distinct aggregate which has an input with non-distinct aggregates grouped by column, which is distinct in outer aggregate.

These Drill rules were able to work correctly only for the cases when aggregate rel node does not contain ` aggCalls` and contains only the single field in `rowType` (in this case `groupSet` is always `{0}`, and it still correct for outer aggregates which are created in `*AggPrule`s)

With the new version of `AggregateExpandDistinctAggregatesRule` these Drill rules received aggregate rel nodes with the non-empty lists of ` aggCalls`, therefore its `rowType` contains more than one field. But before my changes, the same `groupSet` was passed to the constructor of outer aggregate and row type of aggregate differed from the row type of its input. So it was incorrect `groupSet`. 
Aggregate rel nodes always specify the group by columns in the first positions of the list, so correct `groupSet` for outer aggregate should be `ImmutableBitSet` with the same size as the `groupSet` of nested aggregate, but the ordinals of columns should start from 0. 

As for the point of iterating through the group set, `ImmutableBitSet.size()`, `ImmutableBitSet.length()` and `ImmutableBitSet.cardinality()` does not return desired "size" of the `groupSet`. `AggregateExpandDistinctAggregatesRule` also contains similar code which iterating through the `groupSet` for similar purposes.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -1303,6 +1305,8 @@ private void checkGroupAndAggrValues(int incomingRowIdx) {
       long memDiff = allocator.getAllocatedMemory() - allocatedBeforeHTput;
       if ( memDiff > 0 ) { logger.warn("Leak: HashTable put() OOM left behind {} bytes allocated",memDiff); }
 
+      checkForSpillPossibility(currentPartition);
```
chunhui-shi: Not sure this check 'chooseAPartitionToFlush'  is needed. If an exception is desired, I would think modifying doSpill() is better way e.g. modifying this line: "  if ( victimPartition < 0 ) { return; } " Otherwise in this process chooseAPartitionToFlush will be called twice.

      int victimPartition = chooseAPartitionToFlush(currentPartition, forceSpill);

      // In case no partition has more than one batch -- try and "push the limits"; maybe next
      // time the spill could work.
      if ( victimPartition < 0 ) { return; }

vvysotskyi: These checks were needed to avoid infinite loop when there is not enough memory for the spill. 
I moved these checks into `spillIfNeeded()` method, so when called `doSpill()`, `forceSpill` in `spillIfNeeded()` is true and check should be done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/DrillRelBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+  protected DrillRelBuilder(Context context, RelOptCluster cluster, RelOptSchema relOptSchema) {
+    super(context, cluster, relOptSchema);
+    this.filterFactory =
+        Util.first(context.unwrap(RelFactories.FilterFactory.class),
+            RelFactories.DEFAULT_FILTER_FACTORY);
+  }
+
+  /**
+   * Original method {@link RelBuilder#empty} returns empty values rel.
+   * In the order to preserve dara row types, filter with false predicate is created.
```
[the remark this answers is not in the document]
vvysotskyi: Thanks, fixed

exec/java-exec/src/main/java/org/apache/drill/exec/planner/DrillRelBuilder.java
```diff
@@ -0,0 +1,69 @@
...
+   */
+  @Override
+  public RelBuilder empty() {
```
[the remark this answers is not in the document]
vvysotskyi: In some cases, RowType of the input may not be known at the moment when this method is called (some of the columns or all columns were not dynamically discovered yet), therefore I had to make changes directly in Calcite to allow using of custom RelBuilder and make changes in Drill to pass custom RelBuilder into them. For more details please see [1] and [2].

[1] https://lists.apache.org/list.html?dev@calcite.apache.org:lte=1y:Make%20RelBuilder.filter%28%29%20configurable
[2] https://issues.apache.org/jira/browse/CALCITE-2043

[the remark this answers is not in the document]
amansinha100: Ok, thanks for clarifying.  The schema-on-read does make it difficult to avoid overriding the empty() behavior.

amansinha100:
@vvysotskyi could you pls separate out the HashAggregate OOM related change from this PR and file a separate JIRA for it since it is unrelated to the calcite rebase per-se.  Lumping them together causes confusion.   Thanks.

amansinha100:
Other than one comment above, rest LGTM.  +1

vvysotskyi:
@amansinha100, regarding HashAggregate OOM related change, it was done in the scope of this pull request since with new Calcite a physical plan for the query was changed to the correct one but it caused an infinite loop in HashAgg operator. Therefore I made these changes in order to prevent this infinite loop for the queries that previously worked. 
I still think that it is better to keep this change in this PR.

vvysotskyi:
FYI: changes in pom files:
We observed issue with Avatica similar to the issue described in https://issues.apache.org/jira/browse/CALCITE-1694. Therefore we had to make changes in our pom files similar to the changes, made in CALCITE-1694 to fix NoClassDefFoundError.

PR #1068, opened 2017-12-12
DRILL-6020 Fix NullPointerException when querying JSON untyped path w…

MitchelLabonte:
…ith Union setting on

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/vector/complex/FieldIdUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/vector/complex/FieldIdUtil.java
@@ -31,21 +31,29 @@ public class FieldIdUtil {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FieldIdUtil.class);
 
   public static TypedFieldId getFieldIdIfMatchesUnion(UnionVector unionVector, TypedFieldId.Builder builder, boolean addToBreadCrumb, PathSegment seg) {
-    if (seg.isNamed()) {
-      ValueVector v = unionVector.getMap();
-      if (v != null) {
-        return getFieldIdIfMatches(v, builder, addToBreadCrumb, seg);
-      } else {
-        return null;
+    if (seg != null) {
+      if (seg.isNamed()) {
+        ValueVector v = unionVector.getMap();
+        if (v != null) {
+          return getFieldIdIfMatches(v, builder, addToBreadCrumb, seg);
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/vector/complex/FieldIdUtil.java
```diff
@@ -131,7 +131,7 @@ public static TypedFieldId getFieldIdIfMatches(ValueVector vector, TypedFieldId.
     } else if(v instanceof ListVector) {
       ListVector list = (ListVector) v;
       return getFieldIdIfMatches(list, builder, addToBreadCrumb, seg.getChild());
-    } else if (v instanceof  UnionVector) {
+    } else if (v instanceof  UnionVector && !seg.isLastPath()) {
```
vvysotskyi: Good catch! 
I think we should add check for nulls into method `getFieldIdIfMatchesUnion()` as it was done for `getFieldIdIfMatches()`. Also please add a unit test for this change. You may use [testFieldWithDots()](https://github.com/apache/drill/blob/acc5ed927e1fa4011ac1c3724d15197484b9f45b/exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestJsonReader.java#L705) as an example.

But one more point: with this change error is not thrown, but only nulls are returned. I think we should also fix this issue on the borders of this pull request.

MitchelLabonte:
> But one more point: with this change error is not thrown, but only nulls are returned. I think we should also fix this issue is the scope of this pull request.

@vvysotskyi I am not sure what to do to fix this. Any suggestion?

vvysotskyi:
@MitchelLabonte, I think this NPE is just a consequence of the bug that should be fixed. Please investigate why Drill is trying to use child `PathSegment` when a value has VarChar type.

MitchelLabonte:
@vvysotskyi
This is happening because the type is cached as a json object from the previous row. The fix is similar to the getFieldIdIfMatches() method so it looks like this is intended behaviour.
As you can see in the unit test, the results are what is expected after the fix. I am not sure what else could be done.

vvysotskyi:
Sorry, I read your removed comment and made a suggestion that this test fails. 
Did you run all unit tests to see that this change does not break anything?

MitchelLabonte:
@vvysotskyi No problem, yes all tests pass now.

vvysotskyi:
@MitchelLabonte, thanks for the pull request, +1

PR #1069, opened 2017-12-12
DRILL-5994 Added webserver maxThreads configuration option to enable …

MitchelLabonte:
…launching on a machine with more than 200 cores

### DIFF
--- a/exec/java-exec/src/main/resources/drill-module.conf
+++ b/exec/java-exec/src/main/resources/drill-module.conf
@@ -377,7 +377,9 @@ drill.exec: {
   # refresh time.
   grace_period_ms : 0,
   //port hunting for drillbits. Enabled only for testing purposes.
-  port_hunt : false
+  port_hunt : false,
+  // Max threads of embedded Jetty
+  webserver.threadpool_max: 200
 
 }
 
--- a/exec/java-exec/src/main/resources/drill-module.conf
+++ b/exec/java-exec/src/main/resources/drill-module.conf
@@ -379,7 +379,7 @@ drill.exec: {
   //port hunting for drillbits. Enabled only for testing purposes.
   port_hunt : false,
   // Max threads of embedded Jetty
-  webserver.threadpool_max: 200
+  web_server.thread_pool_max: 200
 
 }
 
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -377,7 +377,9 @@ drill.exec: {
   # refresh time.
   grace_period_ms : 0,
   //port hunting for drillbits. Enabled only for testing purposes.
-  port_hunt : false
+  port_hunt : false,
+  // Max threads of embedded Jetty
+  webserver.threadpool_max: 200
```
arina-ielchiieva: Please add underscores: `web_server.thread_pool_max`

arina-ielchiieva:
+1, LGTM.

vrozov:
@MitchelLabonte @arina-ielchiieva I don't think that Drill needs that many threads/acceptors to handle HTTP(s) requests as it is not a real web (REST API) server. For proper resource utilization, it will be better to limit the number of acceptors to a small value (let's say 2 or 4 by default) instead of the current default that uses a number of available processors and can be huge on machines with lots of cores (32 or more). @paul-rogers  What is your take on this?

MitchelLabonte:
@vrozov the Web server definitely doesn't need this much, but Drill won't launch on a machine with more than 200 cores and throw an exception:
Exception in thread "main" org.apache.drill.exec.exception.DrillStartupException: Failure
during initial startup of Drillbit:
> at org.apache.drill.exec.server.Drillbit.start(Drillbit.java:313)
> at org.apache.drill.exec.server.Drillbit.start(Drillbit.java:289)
> at org.apache.drill.exec.server.Drillbit.start(Drillbit.java:285)
> Caused by: java.lang.IllegalStateException: Insufficient max threads in ThreadPool: max=200
< needed=206
> at org.eclipse.jetty.server.Server.doStart(Server.java:321)
> at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
> at org.eclipse.drill.exec.server.rest.WebServer.start(WebServer.java:197)
> at org.eclipse.drill.exec.server.Drillbit.run(Drillbit.java:140)
> at org.eclipse.drill.exec.server.Drillbit.start(Drillbit.java:309)
> ... 2 more

The configuration is there to have an option to not hit this exception on startup on a machine with more than 200 cores.

vrozov:
@MitchelLabonte My proposal is to limit the number of acceptors instead of allowing to increase the max number of threads in the connection pool. If the number of acceptors is limited, the number of "needed" threads will be limited too and the exception will not be raised.

PR #1072, opened 2017-12-13
DRILL-5879: Improved SQL Pattern Contains Performance

sachouche:
**BACKGROUND**
- JIRA [DRILL-5879](https://issues.apache.org/jira/browse/DRILL-5879) goal is to improve the "Contains" pattern performance
- [DRILL-5899](https://issues.apache.org/jira/browse/DRILL-5899) (sub-task) was created subsequently to avoid the ASCII / Unicode pre-processing overhead.
- This pull-request addresses the algorithmic part of this functionality

**ANALYSIS**
- Contains has O(n*m) complexity
- There are two ways to optimize the associated runtime
1) Minimize the number of instructions, pipelining, and CPU stalls
2) Use smarter algorithms (compared to the naive one); for example, the Boyer-Moore algorithm (which is implemented by several popular Open Source tools such as grep)

**IMPLEMENTATION**
Our approach contains both suggestions (listed in the analysis)
- Created five matchers that are chosen based on the pattern length
- The first three, are based on pattern 1) and target patterns with length [1..4[ 
- The forth one, has a similar runtime then the current implementation and targets patterns with length [4..10[
- We use the the Boyer-Moore algorithm for patterns with a length larger than 9 bytes
NOTE - the JDK doesn't use this algorithm because of two main  reasons
- Two extra arrays are necessary with a size relative to the supported character-set and the pattern length (this would be particularly costly for unicode as this would require 64k entries)
- Each Contains (or indexOf) invocation would require memory and initialization overhead
- Drill doesn't have this issue as a) initialization overhead is amortized since the pattern will be matched against many input values and b) our Contains logic is centered around bytes so the memory overhead is around 256 integers per fragment

**PERFORMANCE IMPROVEMENTS**
- We observe at least 25% improvement per Contains operation for matchers with pattern length lower than 4; 100% for negative cases (as the code never accesses a secondary for-loop)
- It was noticed the Boyer-Moor algorithm performs poorly for small patterns as lookup accesses erase the associated optimizations; this algorithm performs extremely well when the pattern length increases as unlike the naive implementation it is able to use the lookup table to jump beyond the next character (since the matching phase has already gained so insight).
- One popular introduction to this algorithm is the following use-case
- Input: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
- Pattern: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaax

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
@@ -29,12 +29,14 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
     // Pattern matching is 1) a CPU intensive operation and 2) pattern and input dependent. The conclusion is
     // that there is no single implementation that can do it all well. So, we use multiple implementations
     // chosen based on the pattern length.
-    if (patternLength == 1) {
-      matcherFcn = new Matcher1();
+    if (patternLength == 0) {
+      matcherFcn = new MatcherZero();
+    } else if (patternLength == 1) {
+      matcherFcn = new MatcherOne();
     } else if (patternLength == 2) {
-      matcherFcn = new Matcher2();
+      matcherFcn = new MatcherTwo();
     } else if (patternLength == 3) {
-      matcherFcn = new Matcher3();
+      matcherFcn = new MatcherThree();
     } else if (patternLength < 10) {
       matcherFcn = new MatcherN();
     } else {
@@ -68,9 +70,22 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
   }
 
   /** Handles patterns with length one */
-  private final class Matcher1 extends MatcherFcn {
+  private final class MatcherZero extends MatcherFcn {
 
-    private Matcher1() {
+    private MatcherZero() {
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+      return 1;
+    }
+  }
+
+  /** Handles patterns with length one */
+  private final class MatcherOne extends MatcherFcn {
+
+    private MatcherOne() {
       super();
     }
 
@@ -95,10 +110,9 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
   }
 
   /** Handles patterns with length two */
-  private final class Matcher2 extends MatcherFcn {
+  private final class MatcherTwo extends MatcherFcn {
 
-    private Matcher2() {
-      super();
+    private MatcherTwo() {
     }
 
     /** {@inheritDoc} */
@@ -128,10 +142,9 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
   }
 
   /** Handles patterns with length three */
-  private final class Matcher3 extends MatcherFcn {
+  private final class MatcherThree extends MatcherFcn {
 
-    private Matcher3() {
-      super();
+    private MatcherThree() {
     }
 
     /** {@inheritDoc} */
@@ -166,17 +179,11 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
   private final class MatcherN extends MatcherFcn {
 
     private MatcherN() {
-      super();
     }
 
     /** {@inheritDoc} */
     @Override
     protected final int match(int start, int end, DrillBuf drillBuf) {
-
-      if (patternLength == 0) {
-        return 1;
-      }
-
       final int lengthToProcess = end - start - patternLength + 1;
       int patternIndex          = 0;
       final byte firstPattByte  = patternArray[0];
@@ -197,11 +204,11 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
           }
 
           if (patternIndex == patternLength) {
-            break;
+            return 1;
           }
         }
       }
-      return patternIndex == patternLength ? 1 : 0;
+      return 0;
     }
   }
 
@@ -223,10 +230,6 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
     /** {@inheritDoc} */
     @Override
     protected int match(int start, int end, DrillBuf drillBuf)  {
-      if (patternLength == 0) {
-        return 1;
-      }
-
       final int inputLength = end - start;
 
       for (int idx1 = patternLength - 1, idx2; idx1 < inputLength;) {
@@ -235,7 +238,7 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
             return 1;
           }
         }
-        // i += pattern.length - j; // For naive method
+        // idx1 += pattern.length - idx2; // For naive method
         idx1 += Math.max(offsetTable[patternLength - 1 - idx2], characterTable[drillBuf.getByte(start + idx1) & 0xFF]);
       }
       return 0;
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
@@ -69,7 +69,7 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
     protected abstract int match(int start, int end, DrillBuf drillBuf);
   }
 
-  /** Handles patterns with length one */
+  /** Handles patterns with length zero */
   private final class MatcherZero extends MatcherFcn {
 
     private MatcherZero() {
@@ -84,23 +84,23 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
 
   /** Handles patterns with length one */
   private final class MatcherOne extends MatcherFcn {
+    final byte firstPatternByte;
 
     private MatcherOne() {
-      super();
+      firstPatternByte  = patternArray[0];
     }
 
     /** {@inheritDoc} */
     @Override
     protected final int match(int start, int end, DrillBuf drillBuf) {
       final int lengthToProcess = end - start;
-      final byte firstPattByte  = patternArray[0];
 
       // simplePattern string has meta characters i.e % and _ and escape characters removed.
       // so, we can just directly compare.
       for (int idx = 0; idx < lengthToProcess; idx++) {
         byte inputByte = drillBuf.getByte(start + idx);
 
-        if (firstPattByte != inputByte) {
+        if (firstPatternByte != inputByte) {
           continue;
         }
         return 1;
@@ -111,28 +111,30 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
 
   /** Handles patterns with length two */
   private final class MatcherTwo extends MatcherFcn {
+    final byte firstPatternByte;
+    final byte secondPatternByte;
 
     private MatcherTwo() {
+      firstPatternByte  = patternArray[0];
+      secondPatternByte = patternArray[1];
     }
 
     /** {@inheritDoc} */
     @Override
     protected final int match(int start, int end, DrillBuf drillBuf) {
       final int lengthToProcess = end - start - 1;
-      final byte firstPattByte  = patternArray[0];
-      final byte secondPattByte = patternArray[1];
 
       // simplePattern string has meta characters i.e % and _ and escape characters removed.
       // so, we can just directly compare.
       for (int idx = 0; idx < lengthToProcess; idx++) {
         final byte firstInByte = drillBuf.getByte(start + idx);
 
-        if (firstPattByte != firstInByte) {
+        if (firstPatternByte != firstInByte) {
           continue;
         } else {
-          final byte secondInByte = drillBuf.getByte(start + idx +1);
+          final byte secondInByte = drillBuf.getByte(start + idx + 1);
 
-          if (secondInByte == secondPattByte) {
+          if (secondInByte == secondPatternByte) {
             return 1;
           }
         }
@@ -143,30 +145,33 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
 
   /** Handles patterns with length three */
   private final class MatcherThree extends MatcherFcn {
+    final byte firstPatternByte;
+    final byte secondPatternByte;
+    final byte thirdPatternByte;
 
     private MatcherThree() {
+      firstPatternByte   = patternArray[0];
+      secondPatternByte  = patternArray[1];
+      thirdPatternByte   = patternArray[2];
     }
 
     /** {@inheritDoc} */
     @Override
     protected final int match(int start, int end, DrillBuf drillBuf) {
-      final int lengthToProcess  = end - start -2;
-      final byte firstPattByte   = patternArray[0];
-      final byte secondPattByte  = patternArray[1];
-      final byte thirdPattByte   = patternArray[2];
+      final int lengthToProcess = end - start - 2;
 
       // simplePattern string has meta characters i.e % and _ and escape characters removed.
       // so, we can just directly compare.
       for (int idx = 0; idx < lengthToProcess; idx++) {
         final byte inputByte = drillBuf.getByte(start + idx);
 
-        if (firstPattByte != inputByte) {
+        if (firstPatternByte != inputByte) {
           continue;
         } else {
-          final byte secondInByte = drillBuf.getByte(start + idx +1);
-          final byte thirdInByte  = drillBuf.getByte(start + idx +2);
+          final byte secondInByte = drillBuf.getByte(start + idx + 1);
+          final byte thirdInByte  = drillBuf.getByte(start + idx + 2);
 
-          if (secondInByte == secondPattByte && thirdInByte == thirdPattByte) {
+          if (secondInByte == secondPatternByte && thirdInByte == thirdPatternByte) {
             return 1;
           }
         }
@@ -177,8 +182,10 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
 
   /** Handles patterns with arbitrary length */
   private final class MatcherN extends MatcherFcn {
+    final byte firstPatternByte;
 
     private MatcherN() {
+      firstPatternByte = patternArray[0];
     }
 
     /** {@inheritDoc} */
@@ -186,14 +193,13 @@ public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
     protected final int match(int start, int end, DrillBuf drillBuf) {
       final int lengthToProcess = end - start - patternLength + 1;
       int patternIndex          = 0;
-      final byte firstPattByte  = patternArray[0];
 
       // simplePattern string has meta characters i.e % and _ and escape characters removed.
       // so, we can just directly compare.
       for (int idx = 0; idx < lengthToProcess; idx++) {
         final byte inputByte = drillBuf.getByte(start + idx);
 
-        if (firstPattByte == inputByte) {
+        if (firstPatternByte == inputByte) {
           for (patternIndex = 1; patternIndex < patternLength; ++patternIndex) {
             final byte currInByte   = drillBuf.getByte(start + idx + patternIndex);
             final byte currPattByte = patternArray[patternIndex];
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
@@ -19,44 +19,283 @@ package org.apache.drill.exec.expr.fn.impl;
 
 import io.netty.buffer.DrillBuf;
 
-public class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
+/** SQL Pattern Contains implementation */
+public final class SqlPatternContainsMatcher extends AbstractSqlPatternMatcher {
+  private final MatcherFcn matcherFcn;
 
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
-      for (int patternIndex = 0; patternIndex < patternLength; patternIndex++) {
-        if (patternByteBuffer.get(patternIndex) != drillBuf.getByte(start + txtIndex + patternIndex)) {
-          continue outer;
+      for (int idx = 0; idx < lengthToProcess; idx++) {
+        final byte firstInByte = drillBuf.getByte(start + idx);
+
+        if (firstPattByte != firstInByte) {
+          continue;
+        } else {
+          final byte secondInByte = drillBuf.getByte(start + idx +1);
```
ppadma: same comment as above. No need to save in local variables

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
-      return 1;
+    private Matcher3() {
+      super();
     }
 
-    return  0;
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+      final int lengthToProcess  = end - start -2;
```
ppadma: same problem as above. This can become negative.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+
+      if (patternLength == 0) {
+        return 1;
+      }
+
+      final int lengthToProcess = end - start - patternLength + 1;
```
ppadma: same problem. can become negative.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+  //--------------------------------------------------------------------------
+  // Inner Data Structure
+  // --------------------------------------------------------------------------
+
+  /** Abstract matcher class to allow us pick the most efficient implementation */
+  private abstract class MatcherFcn {
+    protected final byte[] patternArray;
+
+    protected MatcherFcn() {
+      assert patternByteBuffer.hasArray();
```
ppadma: is this true for null pattern ?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+      final byte firstPattByte  = patternArray[0];
+
+      // simplePattern string has meta characters i.e % and _ and escape characters removed.
+      // so, we can just directly compare.
+      for (int idx = 0; idx < lengthToProcess; idx++) {
+        final byte inputByte = drillBuf.getByte(start + idx);
+
+        if (firstPattByte == inputByte) {
+          for (patternIndex = 1; patternIndex < patternLength; ++patternIndex) {
+            final byte currInByte   = drillBuf.getByte(start + idx + patternIndex);
```
ppadma: you might end up accessing drillBuf out of bound if idx+patternIndex > length of drillBuf.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+      super();
+
+      this.offsetTable    = makeOffsetTable();
+      this.characterTable = makeCharTable();
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    protected int match(int start, int end, DrillBuf drillBuf)  {
+      if (patternLength == 0) {
```
ppadma: do we even come here if patternLength is 0 ?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+            final byte currInByte   = drillBuf.getByte(start + idx + patternIndex);
+            final byte currPattByte = patternArray[patternIndex];
+
+            if (currInByte != currPattByte) {
+              break;
+            }
+          }
+
+          if (patternIndex == patternLength) {
+            break;
```
ppadma: can we just return from here ?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
     super(patternString);
+
+    // Pattern matching is 1) a CPU intensive operation and 2) pattern and input dependent. The conclusion is
+    // that there is no single implementation that can do it all well. So, we use multiple implementations
+    // chosen based on the pattern length.
+    if (patternLength == 1) {
+      matcherFcn = new Matcher1();
+    } else if (patternLength == 2) {
+      matcherFcn = new Matcher2();
+    } else if (patternLength == 3) {
```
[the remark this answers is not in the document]
sachouche: will do.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: HeapByteBuffer will always create an internal buffer. Empty string will have a byte array with zero bytes.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+      final byte firstPattByte  = patternArray[0];
 
-    // no match if input string length is less than pattern length
-    if (txtLength < patternLength) {
+      // simplePattern string has meta characters i.e % and _ and escape characters removed.
+      // so, we can just directly compare.
+      for (int idx = 0; idx < lengthToProcess; idx++) {
+        byte inputByte = drillBuf.getByte(start + idx);
+
+        if (firstPattByte != inputByte) {
```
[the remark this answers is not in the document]
sachouche: These optimizations are useless with modern compilers. This would have been useful if the inputByte value was the same.. then we would save few instructions.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+    private Matcher2() {
+      super();
+    }
 
-    outer:
-    for (int txtIndex = 0; txtIndex <= outerEnd; txtIndex++) {
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+      final int lengthToProcess = end - start - 1;
```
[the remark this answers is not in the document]
sachouche: This is not a problem as the for-loop condition (index = 0; index < lengthToProcess; ..) will prevent any processing to take place unless the correct condition is met. The key point here is that the input length is always provided by the "end - start" formula.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: same answer.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: This will not happen because of the outer-loop condition; idx < lengthToProcess which guarantees that we always have enough bytes to fully match the pattern; example:
- input-len = 6; pattern = 4 this means lengthToProcess = 3 (6-4+1)
- index 0, 1, and 2 are allowed since we have at least 4 bytes we can match (1 within the outer-loop and 3 in the inner one); index 3 will be rejected since we have only three bytes left (b3, b4, and b5)
- If no match, then

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: will do.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: Yes, it would have come here; but for simplicity I guess I can handle that use-case within the switch case.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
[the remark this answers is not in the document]
paul-rogers: But switch does not work for the last case: < 10

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+      // simplePattern string has meta characters i.e % and _ and escape characters removed.
+      // so, we can just directly compare.
+      for (int idx = 0; idx < lengthToProcess; idx++) {
+        final byte inputByte = drillBuf.getByte(start + idx);
+
+        if (firstPattByte != inputByte) {
+          continue;
+        } else {
+          final byte secondInByte = drillBuf.getByte(start + idx +1);
+          final byte thirdInByte  = drillBuf.getByte(start + idx +2);
```
paul-rogers: Is it faster to say `secondInByte = thirdInByte`; shuffle values between variables in each iteration rather than diving into direct memory multiple types for the same byte?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+
+    private MatcherN() {
+      super();
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+
+      if (patternLength == 0) {
```
paul-rogers: Should there be a special case for an empty pattern? Just return 1 always. Else, seems odd for this to be the length = 0 || length > n case.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
paul-rogers: In this code, the `lengthToProcess` is used only in the `for` loop, so a negative value causes the loop to never execute. The termination condition works OK. But, since the termination index is to compare the pattern length and index, and both will be 0 in the empty-pattern case, there is no reason for the special length == 0 check above.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+  /**
+   * Boyer-Moore matcher algorithm; excellent for large patterns and for prefix patterns which appear
+   * frequently in the input.
+   */
+  private final class BoyerMooreMatcher extends MatcherFcn {
+    private final int[] offsetTable;
+    private final int[] characterTable;
+
+    private BoyerMooreMatcher() {
+      super();
```
paul-rogers: Not needed. Call to no-argument `super()` is implicit.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+
+      this.offsetTable    = makeOffsetTable();
+      this.characterTable = makeCharTable();
```
paul-rogers: `this.` not really needed; there is no name conflict.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+    /** {@inheritDoc} */
+    @Override
+    protected int match(int start, int end, DrillBuf drillBuf)  {
+      if (patternLength == 0) {
+        return 1;
+      }
+
+      final int inputLength = end - start;
+
+      for (int idx1 = patternLength - 1, idx2; idx1 < inputLength;) {
```
paul-rogers: Where is `idx2` initialized?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,283 @@
...
+    private int[] makeCharTable() {
+      final int TABLE_SIZE = 256; // This implementation is based on byte comparison
+      int[] resultTable    = new int[TABLE_SIZE];
+
+      for (int idx = 0; idx < resultTable.length; ++idx) {
+        resultTable[idx] = patternLength;
+      }
+
+      for (int idx = 0; idx < patternLength - 1; ++idx) {
+        final int patternValue    = ((int) patternArray[idx]) & 0xFF;
```
paul-rogers: The cast may be unnecessary. `0xFF` is an `int` which will case the `byte` `patternArray[idx]` to be promoted to `int` implicitly.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestSqlPatterns.java
```diff
@@ -446,6 +446,61 @@ public void testSqlPatternComplex() {
     assertEquals(1, sqlPatternComplex.match(0, byteBuffer.limit(), drillBuf)); // should match
   }
 
+  @Test
+  public void testSqlPatternContainsMUltipleMatchers() {
```
paul-rogers: These tests are incomplete. We do not test:

* Pattern length = input length, match
* Pattern length = input length, no match
* Pattern length < input length
* Pattern length = 0
* Non-ASCII paterns and/or data
* Data longer than the internal table sizes
* Patterns of each of the special lengths for each of the above cases: match, not match, shorter, equals, longer

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,286 @@
...
-    final int outerEnd = txtLength - patternLength;
+    private MatcherThree() {
+    }
 
-    outer:
-    for (int txtIndex = 0; txtIndex <= outerEnd; txtIndex++) {
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+      final int lengthToProcess  = end - start -2;
```
ppadma: space between - and 2

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,286 @@
...
-        if (patternByteBuffer.get(patternIndex) != drillBuf.getByte(start + txtIndex + patternIndex)) {
-          continue outer;
+      for (int idx = 0; idx < lengthToProcess; idx++) {
+        final byte inputByte = drillBuf.getByte(start + idx);
+
+        if (firstPattByte != inputByte) {
+          continue;
+        } else {
+          final byte secondInByte = drillBuf.getByte(start + idx +1);
+          final byte thirdInByte  = drillBuf.getByte(start + idx +2);
```
ppadma: space between + and 1

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,286 @@
...
+
+    private MatcherTwo() {
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+      final int lengthToProcess = end - start - 1;
+      final byte firstPattByte  = patternArray[0];
+      final byte secondPattByte = patternArray[1];
```
ppadma: can you initialize them in the constructor for matcher function ? that way you can initialize once and reuse for each match.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,286 @@
...
+      // simplePattern string has meta characters i.e % and _ and escape characters removed.
+      // so, we can just directly compare.
+      for (int idx = 0; idx < lengthToProcess; idx++) {
+        final byte firstInByte = drillBuf.getByte(start + idx);
 
-    final int txtLength = end - start;
+        if (firstPattByte != firstInByte) {
+          continue;
+        } else {
+          final byte secondInByte = drillBuf.getByte(start + idx +1);
```
ppadma: space between + and 1

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,286 @@
...
+      patternArray = patternByteBuffer.array();
+    }
+
+    /**
+     * @return 1 if the pattern was matched; 0 otherwise
+     */
+    protected abstract int match(int start, int end, DrillBuf drillBuf);
+  }
+
+  /** Handles patterns with length one */
```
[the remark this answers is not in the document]
sachouche: cut-and-paste issue :-)

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,286 @@
...
+    protected final int match(int start, int end, DrillBuf drillBuf) {
       return 1;
     }
+  }
+
+  /** Handles patterns with length one */
+  private final class MatcherOne extends MatcherFcn {
+
+    private MatcherOne() {
+      super();
```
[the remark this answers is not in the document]
sachouche: removed.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
```diff
@@ -19,44 +19,286 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    protected final int match(int start, int end, DrillBuf drillBuf) {
+      final int lengthToProcess = end - start;
+      final byte firstPattByte  = patternArray[0];
```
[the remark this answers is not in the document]
sachouche: renamed all such variables to XXXPatternBytes

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: Good idea; done this for all the matchers.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: fixed

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SqlPatternContainsMatcher.java
sachouche: fixed

priteshm:
@paul-rogers Is this ready for commit?

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestSqlPatterns.java
paul-rogers: If tests exist, and you have verified that each of the lengths is, in fact, tested by those tests, then I'm fine with adding a comment identifying the cases and where to find the existing tests. The point is to make sure all variations are fully tested -- one way or another.

PR #1073, opened 2017-12-14
DRILL-5967: Fixed memory leak in OrderedPartitionSender

ilooner:
The OrderedPartitionSender was leaking memory every time it was created because it created a wrapper RecordBatch which allocated memory but was never closed.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
@@ -43,6 +43,7 @@ import org.apache.drill.exec.planner.physical.PlannerSettings;
 import org.apache.drill.exec.proto.ExecProtos.FragmentHandle;
 import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.BatchSchema.SelectionVectorMode;
+import org.apache.drill.exec.record.CloseableRecordBatch;
 import org.apache.drill.exec.record.FragmentWritableBatch;
 import org.apache.drill.exec.record.RecordBatch;
 import org.apache.drill.exec.record.RecordBatch.IterOutcome;
@@ -72,6 +73,7 @@ public class PartitionSenderRootExec extends BaseRootExec {
   private final AtomicInteger remaingReceiverCount;
   private volatile boolean done = false;
   private boolean first = true;
+  private boolean closeIncoming;
 
   long minReceiverRecordCount = Long.MAX_VALUE;
   long maxReceiverRecordCount = Long.MIN_VALUE;
@@ -99,9 +101,17 @@ public class PartitionSenderRootExec extends BaseRootExec {
   public PartitionSenderRootExec(FragmentContext context,
                                  RecordBatch incoming,
                                  HashPartitionSender operator) throws OutOfMemoryException {
+    this(context, incoming, operator, false);
+  }
+
+  public PartitionSenderRootExec(FragmentContext context,
+                                 RecordBatch incoming,
+                                 HashPartitionSender operator,
+                                 boolean closeIncoming) throws OutOfMemoryException {
     super(context, context.newOperatorContext(operator, null), operator);
     this.incoming = incoming;
     this.operator = operator;
+    this.closeIncoming = closeIncoming;
     this.context = context;
     outGoingBatchCount = operator.getDestinations().size();
     popConfig = operator;
@@ -340,6 +350,10 @@ public class PartitionSenderRootExec extends BaseRootExec {
     if (partitioner != null) {
       updateAggregateStats();
       partitioner.clear();
+    }
+
+    if (closeIncoming) {
+      ((CloseableRecordBatch) incoming).close();
     }
   }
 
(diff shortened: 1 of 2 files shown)


ilooner:
@paul-rogers

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
```diff
@@ -98,10 +100,12 @@ public int metricId() {
 
   public PartitionSenderRootExec(FragmentContext context,
                                  RecordBatch incoming,
-                                 HashPartitionSender operator) throws OutOfMemoryException {
+                                 HashPartitionSender operator,
+                                 boolean closeIncoming) throws OutOfMemoryException {
```
vrozov: Introduce back constructor without `closeIncoming` argument and call this constructor from it with `closeIncoming` set to `false`.

ilooner:
Applied review comments

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/partitionsender/TestPartitionSender.java
```diff
@@ -350,7 +350,7 @@ public void testAlgorithm() throws Exception {
     public MockPartitionSenderRootExec(FragmentContext context,
         RecordBatch incoming, HashPartitionSender operator)
         throws OutOfMemoryException {
-      super(context, incoming, operator);
+      super(context, incoming, operator, false);
```
vrozov: Is the change still necessary?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderCreator.java
```diff
@@ -35,7 +35,5 @@ public RootExec getRoot(FragmentContext context,
 
     assert children != null && children.size() == 1;
     return new PartitionSenderRootExec(context, children.iterator().next(), config);
-
```
vrozov: Will be good to exclude format only change from the PR.

ilooner:
Applied review comments

vrozov:
LGTM

ilooner:
Removed unnecessary creation of list in OrderedPartitionSenderCreator as discussed. @paul-rogers  please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
```diff
@@ -341,6 +351,10 @@ public void close() throws Exception {
       updateAggregateStats();
       partitioner.clear();
     }
+
+    if (closeIncoming) {
+      ((CloseableRecordBatch) incoming).close();
+    }
```
paul-rogers: Under what scenario would we *not* close the incoming? This fix introduces two paths: one that closes, one that does not. Is the non-closing path needed?

More to the point, do we have unit tests that verify both paths?

Can we simplify this and just define that this operator will close the incoming?

Actually, let's take a step back. This fix is based on a false premise (but one a mistake that is easy to make.) This fix assumes that a parent batch (operator) is responsible for closing its child. But, that is not how the Drill protocol works.

Instead, the fragment executor maintains the operator tree. Upon exit, it walks the tree and closes all its children in order from top down.

So, the true source of this bug is that either a) the fragment exec is failing to do the close, b) that the incoming is not releasing memory on close, or c) that some close operation reallocates memory that a previous close tried to release.

Let's suppose the issue is a: that the fragment exec does not manage the incoming. Let's fix that. Similar with the other cases.

Can you maybe do a bit more research to understand the full scope of the issue?

ilooner: Hi Paul, I understand the FragmentExecutor closes all the operators. However, the FragmentExecutor is only aware of the operators in the tree that it has a reference too. In the case of the OrderedPartitionSenderCreator, an upstream record batch is wrapped in an OrderedPartitionRecordBatch. Since the OrderedPartitionRecordBatch is instantiated in the creator the FragmentExecutor does not have a reference to it. So when the FragmentExecutor closes the operators it closes the original operator, but not not the wrapping OrderedPartitionRecordBatch. This is an issue since the OrderedPartitionRecordBatch allocates VectorContainers which are consequentially never released. This is clearly the source of the memory leak and the fix was verified by QA.

There are two possible Fixes. 

A) We change the Creators in some way to communicate to the FragmentExecutor that they have wrapped an operator, so the FragmentExecutor can close the wrapped operator instead of the original operator. My impression is that this will evolve into a fairly large change.

B) Or we take a less invasive approach and simply tell the PartitionSenderRootExec whether to close the wrapped operator (This is what I did).

I agree approach B is not that great and that approach A is cleaner. But taking approach A is a more intensive project, and I was not keen on having another simple change explode into another 100 file PR. It would be easier to make that change after we've made more improvements to Drill unit testing.

Let me know what you think. I am not against taking approach A, but then we'll have to talk with Pritesh about fitting that project in.

paul-rogers: OK, let's limit scope to stay on schedule. Please include a comment that explains the above, perhaps in the `OrderedPartitionSenderCreator` class.

As you described, the proper solution would be to simply add both operators to the stack of operators that the fragment executor should close. But, since the code assumes that each builder builds a single operator, fixing this design flaw does, as you noted, require changes that are out of scope.

ilooner:
@paul-rogers Added the comment. Please take a look and let me know if you'd like any other changes.

PR #1075, opened 2017-12-18
DRILL-6030: Managed sort should minimize number of batches in a k-way merge

vrozov:
@paul-rogers Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
@@ -42,6 +42,7 @@ public class SortConfig {
   public static final int MIN_SPILL_BATCH_SIZE = 256 * 1024;
   public static final int MIN_MERGE_BATCH_SIZE = 256 * 1024;
 
+  public static final int DEFAULT_MERGE_LIMIT = 128;
   public static final int MIN_MERGE_LIMIT = 2;
 
   private final long maxMemory;
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortMemoryManager.java
@@ -687,11 +687,11 @@ public class SortMemoryManager {
                                 spillBatchSize.maxBufferSize);
     memMergeLimit = Math.max(0, memMergeLimit);
 
-    // If batches are in memory, and we need more memory to merge
-    // them all than is actually available, then spill some in-memory
-    // batches.
+    // If batches are in memory, and final merge count will exceed
+    // merge limit or we need more memory to merge them all than is
+    // actually available, then spill some in-memory batches.
 
-    if (inMemCount > 0  &&  memMergeLimit < spilledRunsCount) {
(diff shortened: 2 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
```diff
@@ -84,7 +85,7 @@ public SortConfig(DrillConfig config) {
     if (limit > 0) {
       mergeLimit = Math.max(limit, MIN_MERGE_LIMIT);
     } else {
-      mergeLimit = Integer.MAX_VALUE;
+      mergeLimit = DEFAULT_MERGE_LIMIT;
```
paul-rogers: The merge limit is already a config option. (I'd forgotten about that.) The comment on the config option says "Limit on the number of spilled batches that can be merged in a single pass." So, let's just set that default (in `drill-override-conf`) to your new value of 128 and leave the code here unchanged.

paul-rogers:
One additional thought. This bug was found when sorting 18 GB of data in 8 GB of memory. That is, a case in which the sort must spill.

What happens in the case in which the 18 GB of data is sorted in, say, 20 GB of memory (an in-memory sort)? We don't want the merge limit to force a spill in this case; kind of defeats the purpose of an in-memory sort.

So:

1. Does the limit affect in memory sort? If so, we need to revise the solution.
2. Does the in-memory sort suffer from a similar performance issue? If so, we need to revise the in memory sort.

One possible solution is to:

1. Defer sorting of individual batches until necessary.
2. Sort batches just before spilling.
3. If all batches fit in memory, do a single, combined sort (using an SV4).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
vrozov: IMO, it is better to change the default to avoid upgrade problems. In an upgrade scenario,  users may simply overwrite `drill-override.conf` from their prior installations and forget to set the merge limit. Is there a reason not to change the default merge limit?

vrozov:
The scenario when all batches can be merged in memory is covered by 'if (canUseMemoryMerge())` check in `SortImpl.java:399`. The affected code path applies only to cases where merge between spilled and in-memory batches is necessary. Note that this is a short term fix to improve managed sort performance, in a long run, it is necessary to have an ability to merge all batches in memory (using SV4) without spilling and be able to merge it with the spilled data.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
paul-rogers: There may be a misunderstanding of how config options work. We define the defaults in Drill's own source code: `drill-module.conf` in each module. (Here it is in `java-exec`.)

To change the default option, we change the value in `drill-module.conf`. In the highly unlikely case that a user has overridden this value in `drill-override.conf`, their value will be used. But, the option is not documented in `drill-override-example.conf` so it is very, very unlikely that anyone created an override. (The property is meant to be internal, for use in tests.)

So, rather than introducing yet another variable, we might as well use the existing config property. This has the added advantage that, if experience suggests that we need a smaller or larger limit for some scenarios, we can make the adjustment in the field via the config system.

vrozov:
@paul-rogers Please review

PR #1076, opened 2017-12-20
DRILL-6036: Create sys.connections table

kkhatua:
Introduced a distributed (i.e. each Drillbit executes a fragment) system table to list all the active client connections to all the Drillbits in a cluster. This is achieved by having the UserServer maintain a static (class-level) map of these connections, which is updated as and when connections are established and closed.

The following details are provided by the table:
    user
    targetUser
    client
    drillbit
    established
    duration
    queries
    isAuthenticated
    isEncrypted
    usingSSL
    session

A sample run shows the following:
```
[root@kk190 ~]# /opt/mapr/drill/apache-drill-1.13.0/bin/sqlline -u "jdbc:drill:drillbit=`hostname -i`"
...
apache drill 1.13.0-SNAPSHOT 
"just drill it"
0: jdbc:drill:drillbit=10.10.106.190> select * from sys.connections;
+------------+-------------+----------------+---------------+--------------------------+------------+----------+------------------+--------------+-----------+---------------------------------------+
|    user    | targetUser  |     client     |   drillbit    |       established        |  duration  | queries  | isAuthenticated  | isEncrypted  | usingSSL  |                session                |
+------------+-------------+----------------+---------------+--------------------------+------------+----------+------------------+--------------+-----------+---------------------------------------+
| anonymous  | anonymous   | 10.10.106.190  | kk190.qa.lab  | 2017-12-20 10:27:30.377  | 5.189 sec  | 1        | false            | false        | false     | b95e76ab-be51-4dde-9554-4d31a36e47a5  |
+------------+-------------+----------------+---------------+--------------------------+------------+----------+------------------+--------------+-----------+---------------------------------------+
1 row selected (0.163 seconds)
```

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
@@ -230,6 +230,10 @@ public class FragmentContext extends BaseFragmentContext implements AutoCloseabl
     return context;
   }
 
+  public QueryContext getQueryContext() {
+    return queryContext;
+  }
+
   /**
    * This method is only used to construt InfoSchemaReader, it is for the reader to get full schema, so here we
    * are going to return a fully initialized schema tree.
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
@@ -230,8 +230,8 @@ public class FragmentContext extends BaseFragmentContext implements AutoCloseabl
     return context;
   }
 
-  public QueryContext getQueryContext() {
-    return queryContext;
+  public boolean isUserAuthenticationEnabled() {
+    return queryContext.isUserAuthenticationEnabled();
   }
 
   /**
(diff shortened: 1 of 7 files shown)


parthchandra:
+1. Nice work.

kkhatua:
@parthchandra I've added the same security check being done for DRILL-5068. But I'm not sure if the match for connections should be in the original user or the impersonated user, or both. Suggestions?

sohami:
Filtering should be done based on impersonated user which will be stored in UserSession as well. So your check [here](https://github.com/kkhatua/drill/blob/66ab15df7e0df6804b1938adc048728c7436450b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java#L81) is fine.

But we shouldn't display the [target User](https://github.com/kkhatua/drill/blob/66ab15df7e0df6804b1938adc048728c7436450b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java#L123) separately in the table. Just displaying the session user is fine which will be same as querying user. Because when we do `session.getTargetUserName()`  it get's the target name which was passed in Connection URL during connection time. But during query time the actual targeted user or the session user can change.

kkhatua:
@sohami updated the commit. Please review.

sohami:
+1 LGTM. Thanks for the changes.

parthchandra:
+1 (binding)

PR #1078, opened 2017-12-29
DRILL-6054: don't try to split the filter when it is not AND

chunhui-shi:


exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/FindPartitionConditions.java
```diff
@@ -228,13 +236,16 @@ private boolean isHolisticExpression(RexCall call) {
     return false;
   }
 
+  protected boolean inputRefToPush(RexInputRef inputRef) {
```
arina-ielchiieva: Can be made private. Should be placed under public methods.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/FindPartitionConditions.java
```diff
@@ -195,8 +195,16 @@ private void popOpStackAndBuildFilter() {
          * For all other operators we clear the children if one of the
          * children is a no push.
          */
-        assert currentOp.getOp().getKind() == SqlKind.AND;
-        newFilter = currentOp.getChildren().get(0);
+        if (currentOp.getOp().getKind() == SqlKind.AND) {
+          newFilter = currentOp.getChildren().get(0);
+          for(OpState opState : opStack) {
```
arina-ielchiieva: Please add space: `for (`

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/FindPartitionConditions.java
chunhui-shi: This is intentionally made to be 'protected' for future extension.
Right now, FindPartitionCondition use position based inputRef(using BitSet dirs) to mark which inputRef should be pushed. But in future, we may use name based policy to decide which one to push.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/FindPartitionConditions.java
chunhui-shi: done.

arina-ielchiieva:
+1, LGTM.

PR #1079, opened 2018-01-03
DRILL-6063: Set correct ThreadContext ClassLoader before using Hadoop…

sohami:
… Configuration class in DrillClient

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserClient.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserClient.java
@@ -344,7 +344,18 @@ public class UserClient
       mechanismName = factory.getSimpleName();
       logger.trace("Will try to authenticate to server using {} mechanism with encryption context {}",
           mechanismName, connection.getEncryptionCtxtString());
+
+      // Update the thread context class loader to current class loader
+      // See DRILL-6063 for detailed description
+      final ClassLoader oldThreadCtxtCL = Thread.currentThread().getContextClassLoader();
+      final ClassLoader newThreadCtxtCL = this.getClass().getClassLoader();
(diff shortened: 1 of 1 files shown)


sohami:
@parthchandra - Please help to review.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserClient.java
```diff
@@ -344,7 +344,18 @@ private boolean clientNeedsAuthExceptPlain(DrillProperties props) {
       mechanismName = factory.getSimpleName();
       logger.trace("Will try to authenticate to server using {} mechanism with encryption context {}",
           mechanismName, connection.getEncryptionCtxtString());
+
+      // Update the thread context class loader to current class loader
+      // See DRILL-6063 for detailed description
+      final ClassLoader oldThreadCtxtCL = Thread.currentThread().getContextClassLoader();
+      final ClassLoader newThreadCtxtCL = this.getClass().getClassLoader();
+      Thread.currentThread().setContextClassLoader(newThreadCtxtCL);
+
```
parthchandra: Is this the only place we need to update? For eaxmple, with SASL the authentication mechanism plugin would have to be in the class path when the authentication method is dtermined. But you have already restored the thread context. (Or does the oldThreadContextCL have the SASL jars in the classpath?)

sohami: This is the only place since this is when `Hadoop Configuration `object is created inside `createAndLoginUser`. SASL authentication mechanism factories are scanned before this code itself and moreover they are scanned in context of current class ClassLoader not in ThreadContext class loader.

parthchandra: makes sense

sohami:
Thanks for the review!

PR #1080, opened 2018-01-03
Add acknowledgement sequence number and flags fields.

tdunning:
This pull request relates to DRILL-6067 but is not ready to merge


tdunning:
The Travis failure appears unrelated to this pull request.

cgivre:
Hi Ted, 
Thanks for doing this.  This looks really great!  The PCAP files came from here: https://github.com/chrissanders/packets.  The author said that they are free to use, but asks for retribution.  

When I started poking at this, and my original thought was to add a boolean column for each TCP flag which would facilitate analysis, as well as a field which contains all the flags. My original thought was that would enable you to quickly detect things like SYN scans and the like.  I've been going through `Practical Packet Analysis` by Chris Sanders and trying to do some of the same things he does in Wireshark with Drill.  The next thing I was going to try to do was figure out a way of getting Drill to follow sequences.

tdunning:
To follow sequences, group by session id, order by time or sequence number.

This assumes ports are not reused to the same host very often. Will break
occasionally under odd conditions such as super high connection rate behind
a broken proxy with no keep alive.

PR #1081, opened 2018-01-03
DRILL-6068: Support user/distrib-specific config checks during startup

kkhatua:
1. Allows for distrib/user specific checks to be done
2. Place-holder files for distribution and user specific checks
3. Moved JVM Version Check to head of script

PR #1082, opened 2018-01-03
DRILL-5741: Automatically manage memory allocations during startup

kkhatua:
**Note:** This commit is rebased on (hence, requires) DRILL-6068 commit

This commit provides a way for distributions and users to automatically define the Drillbit JVM's Heap, Direct and CodeCache allocations. To do this, it leverages the DRILL-6068 commit, which provides for configuration setup.

The only new variable introduced is an environment variable - `DRILLBIT_MAX_PROC_MEM` that defines a Drillbit's max memory parameters. The logic defined by `auto-setup.sh` will check that if any (or all) of the JVM memory parameters have been declared; they, cumulatively, don't exceed the value specified by `DRILLBIT_MAX_PROC_MEM`.

**NOTE:** To disable this, simply unset (or don't define) the environment variable `DRILLBIT_MAX_PROC_MEM`

The variable can be defined in KB, MB, or  GB; similar in syntax to how the JVM MaxHeap is specified. For e.g. 
```
DRILLBIT_MAX_PROC_MEM=13G
DRILLBIT_MAX_PROC_MEM=8192m
DRILLBIT_MAX_PROC_MEM=4194304K
```
In addition, you can specify it as a percent of the total system memory prior to the Drillbit starting up:
`DRILLBIT_MAX_PROC_MEM=40%`

For a system with with 48GB free memory, when set to (say) 25% (with settings defined in drill-env.sh), and heap (8GB) and direct (10GB) are defined; the Drillbit fails startup with the following message:
```
2018-01-03 14:27:57  [WARN] 25% of System Memory (47 GB) translates to 12 GB
2018-01-03 14:27:57  [ERROR]    Unable to start Drillbit due to memory constraint violations
  Total Memory Requested : 19 GB
  Check the following settings to possibly modify (or increase the Max Memory Permitted):
        DRILLBIT_MAX_PROC_MEM=25%
        DRILL_HEAP=8G
        DRILL_MAX_DIRECT_MEMORY=10G
        DRILLBIT_CODE_CACHE_SIZE=1024m 
        *NOTE: It is recommended not to specify DRILLBIT_CODE_CACHE_SIZE as this will be auto-computed based on the HeapSize and would not exceed 1GB
```

For all other combinations, the undefined parameters are adjusted to ensure that the total memory allocated is within the value specified by `DRILLBIT_MAX_PROC_MEM`,

For a system with with 48GB free memory, when set to (say) 50% (with settings defined in drill-env.sh), and heap (8GB) and direct (10GB) are defined; the Drillbit startup with the following warning:
```
2018-01-03 14:31:06  [WARN] 50% of System Memory (47 GB) translates to 24 GB
2018-01-03 14:31:06  [WARN] You have an allocation of 4 GB that is currently unused from a total of 24 GB. You can increase your existing memory configuration to use this extra memory
        DRILLBIT_MAX_PROC_MEM=50%
        DRILL_HEAP=8G
        DRILL_MAX_DIRECT_MEMORY=10G
        DRILLBIT_CODE_CACHE_SIZE=1024m 
        *NOTE: It is recommended not to specify DRILLBIT_CODE_CACHE_SIZE as this will be auto-computed based on the HeapSize and would not exceed 1GB
```

In addition, if the available free memory is less than the allocation, an additional warning is provided under the assumption that the OS will reclaim more free memory when required:
```
2018-01-03 14:31:06  [WARN] Total Memory Allocation for Drillbit (19GB) exceeds available free memory (11GB)
2018-01-03 14:31:06  [WARN] Drillbit will start up, but can potentially crash due to oversubscribing of system memory.
```

For more details, refer the attachments in https://issues.apache.org/jira/browse/DRILL-5741


distribution/src/resources/distrib-auto.sh
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env bash
```
paul-rogers: In general, "distrib" files are meant to be added by Drill distributions, not by Drill itself. If we want to provide the memory checking in Drill, then it should go into some other file other than "distrib."

distribution/src/assemble/bin.xml
```diff
@@ -345,6 +345,16 @@
       <fileMode>0755</fileMode>
       <outputDirectory>conf</outputDirectory>
     </file>
+    <file>
+      <source>src/resources/drill-auto.sh</source>
```
paul-rogers: "-auto"? Is this for the self-driving car feature of Drill? :-)

Is this meant to be for startup-time environment checks? For other add-on processing? Perhaps we can pick a name that makes that a bit clearer.

distribution/src/resources/drill-config.sh
```diff
@@ -180,18 +251,46 @@ else
...
 
-# Default memory settings if none provided by the environment or
+# Checking if being executed in context of Drillbit and not SQLLine
+if [ "$DRILLBIT_CONTEXT" == "1" ]; then 
+  # *-auto.sh allows for distrib/user specific checks to be done
+  distribAuto="$DRILL_CONF_DIR/distrib-auto.sh"
+  if [ ! -r "$distribAuto" ]; then distribAuto="$DRILL_HOME/conf/distrib-auto.sh"; fi
+  if [ ! -r "$distribAuto" ]; then distribAuto=""; fi
+  drillAuto="$DRILL_CONF_DIR/drill-auto.sh"
+  if [ ! -r "$drillAuto" ]; then drillAuto="$DRILL_HOME/conf/drill-auto.sh"; fi
```
paul-rogers: This part is just a bit tricky. Drill- and distribution- specific files do, in fact, reside in `$DRILL_HOME/conf`. But, user files reside in a number of places. So, please use:

- `$DRILL_HOME/conf` for `distrib-auto.sh`
- `$DRILL_CONF_DIR` for user-supplied files (such as `drill-auto.sh`)

distribution/src/resources/drill-config.sh
```diff
@@ -180,18 +251,46 @@ else
...
+  if [ ! -r "$drillAuto" ]; then drillAuto=""; fi
+
+  # Enforcing checks in order (distrib-auto.sh , drill-auto.sh)
+  # (NOTE: A script is executed only if it has relevant executable lines)
+  if [ -n "$distribAuto" ] && [ $(executableLineCount $distribAuto) -gt 0 ]; then
+    . "$distribAuto"
+    if [ $? -gt 0 ]; then fatal_error "Aborting Drill Startup due failed checks from $distribAuto"; fi
+  fi
+  if [ -n "$drillAuto" ] && [ $(executableLineCount $drillAuto) -gt 0 ]; then
```
paul-rogers: To make the code a bit simpler, touch each file once:

* Check if it exists.
* Invoke it if so.
* Handle any errors that are reported.

distribution/src/assemble/bin.xml
kkhatua: AutoConfig was the intent, but looked a bit verbose. How about `drill-autocheck.sh` ? :)

distribution/src/resources/distrib-auto.sh
kkhatua: Wasn't sure which would be a a suitable place. I reasoned the "Apache" distribution is what we are defining it for, hence the choice

distribution/src/resources/drill-config.sh
kkhatua: Agreed. The drill-config file follows this for `drill-env.sh`, but checks for both locations in case of `distrib-env.sh`.

distribution/src/assemble/bin.xml
kkhatua: Modifying the base commit (#1081) to reflect the name change from `[distrib/drill]-auto.sh` to `[distrib/drill]-setup.sh`

distribution/src/resources/drill-config.sh
kkhatua: Passed the checks for the file to the renamed function: `checkExecutableLineCount`

kkhatua:
Done the following changes:
1. The file names have been changed to the following:
* `./bin/auto-setup.sh`
* `./conf/distrib-setup.sh`
* `./conf/drill-setup.sh`
`auto-setup.sh` and `distrib-setup.sh` files are mutually exclusive in execution.
	i. `distrib-setup.sh` -> if it has executable content. Else `auto-setup.sh`
		The assumption here is that the distribution might have a different mechanism for doing things (along lines of specifying a preferred setup service).
	ii. `drill-setup.sh` --> This is any additional setup activities beyond what the distribution offers. I don't think this should be mutually exclusive to the previous step, but should be an add on.

A distribution of Drill should reimplement the setup task done by `auto-setup.sh` , if it wants to do the same separately.

2. For managing memory automatically, the `auto-setup.sh` script is looking for the `DRILLBIT_MAX_PROC_MEM` variable.
If this is not set (which is the case by default), the existing procedure for inferring memory parameters will be followed.

3. If the `DRILLBIT_MAX_PROC_MEM` variable is set, the logic defined within the `auto-setup.sh` script is executed to assign and validate that the total memory allocation of Heap, Direct and CodeCache is within that value.

4. The JIRA and the PR has been updated with the details of the feature.

distribution/src/resources/drill-config.sh
```diff
@@ -180,18 +251,61 @@ else
...
+    drillSetup="$DRILL_HOME/conf/drill-setup.sh" ; #Install-based drill-setup.sh
+    if [ $(checkExecutableLineCount $drillSetup) -eq 0 ]; then drillSetup=""; fi
+  fi
+
+  # Enforcing checks in order (distrib-setup.sh , drill-setup.sh)
+  # (NOTE: A script is executed only if it has relevant executable lines)
+  # Both distribSetup & drillSetup are executed because the user might have introduced additional checks
+  if [ -n "$distribSetup" ]; then
+    . "$distribSetup"
+    if [ $? -gt 0 ]; then fatal_error "Aborting Drill Startup due failed setup by $distribSetup"; fi
```
paul-rogers: In the interests of simplicity, I would just have the script itself report errors and exit. See the existing code in `drill-config.sh`. This way, the script can give a reasonable error message. And, it is not very hard to write `exit 1` to exit the scripts.

distribution/src/assemble/bin.xml
```diff
@@ -345,6 +345,21 @@
...
+      <fileMode>0755</fileMode>
+      <outputDirectory>bin</outputDirectory>
+    </file>
+    <file>
+      <source>src/resources/drill-setup.sh</source>
+      <fileMode>0755</fileMode>
+      <outputDirectory>conf</outputDirectory>
+    </file>
+    <file>
+      <source>src/resources/distrib-setup.sh</source>
```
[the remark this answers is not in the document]
kkhatua: The `distrib-setup.sh` file is empty, but provided the placeholder to indicate where distributions should make the change.
This is identical to the intent of having `distrib-env.sh` in the Apache distribution, which is also empty but serves the same purpose.
https://github.com/apache/drill/blob/master/distribution/src/resources/distrib-env.sh
Just following the same convention.

distribution/src/resources/auto-setup.sh
```diff
@@ -0,0 +1,222 @@
...
+    echo `date +%Y-%m-%d" "%H:%M:%S`"  [WARN] Drillbit not will start up"
+    exit 127
+  elif [ $totalDBitMem_inMB -gt $freeRAM_inMB ]; then
+    echo `date +%Y-%m-%d" "%H:%M:%S`"  [WARN] Total Memory Allocation for Drillbit ("$(valueInGB $totalDBitMem_inMB'm')"GB) exceeds available free memory ("$(valueInGB $freeRAM_inMB'm')"GB)"
+    echo `date +%Y-%m-%d" "%H:%M:%S`"  [WARN] Drillbit will start up, but can potentially crash due to oversubscribing of system memory."
+  fi
+fi
+
+#Implicit that checks have passed
+AutoMemConfigStatus="PASSED"
```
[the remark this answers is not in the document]
kkhatua: I tried with the `debug` argument and it not only shows the environment variables, but if there are any warnings, they also would appear. So, I'd say we are self-sufficient in this regard.
I think we can indicate that we are auto-configuring the memory parameters and allow users to trace back the source of the settings.

distribution/src/resources/drill-config.sh
```diff
@@ -180,18 +251,61 @@ else
...
+# distrib-setup.sh is optional; it is created by some distribution installers
+# that need additional distribution-specific setup to be done.
+# Because installers will have site-specific steps, the file
+# should be moved into the site directory, if the user employs one.
+
+# Checking if being executed in context of Drillbit and not SQLLine
+if [ "$DRILLBIT_CONTEXT" == "1" ]; then
+  # Check whether to run exclusively distrib-setup.sh OR auto-setup.sh
+  distribSetup="$DRILL_CONF_DIR/distrib-setup.sh" ; #Site-based distrib-setup.sh
+  if [ $(checkExecutableLineCount $distribSetup) -eq 0 ]; then
```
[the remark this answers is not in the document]
kkhatua: I'd have liked the KISS principle, but I thought there was a need for placeholder `distrib-setup.sh` file.
Based on that, I need to figure out whether there is a distribtion-specific setup, or should we revert to executing the `auto-setup.sh`. Unlike sourcing environment files, where an unset variable can be set, for auto-setup, the choice of execution has to be mutually exclusive.
This block looks complicated (and verbose with the comments), but is only identifying *what* setup script needs to execute. Hence, all we do here is an assignment of the variables.

distribution/src/resources/auto-setup.sh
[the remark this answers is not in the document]
kkhatua: Added the additional messaging of when auto-configuration is being done. This should ensure no surprises emerge when a partial set of parameters is provided by the user.

kkhatua:
Good point. Let me send out a post on the user-list since this affects a broader audience beyond just Dev. Thanks for reviewing!

PR #1086, opened 2018-01-09
DRILL-6076: Reduce the default memory from a total of 13GB to 5GB

kkhatua:
Reduce minimum memory requirements from 13GB to <= 5GB
For this, the default need to be changed as follows:

Allocation | Current | New 
------------|--------|------
 Heap | 4GB | 1GB 
 Direct | 8GB | 3GB 
 CodeCache | 1GB | 512MB 
 MaxPermSize | 512MB | 512MB (Ignored if JDK8+) 
 **Total** | 13.5GB | 5GB 
 **Total (JDK8+)** | 13GB | 4.5GB

### DIFF
--- a/distribution/src/resources/drill-config.sh
+++ b/distribution/src/resources/drill-config.sh
@@ -295,9 +295,9 @@ fi
 # DEFAULT memory settings if none provided by the environment or
 # above config files.
 # The Drillbit needs a large code cache.
-export DRILL_MAX_DIRECT_MEMORY=${DRILL_MAX_DIRECT_MEMORY:-"8G"}
-export DRILL_HEAP=${DRILL_HEAP:-"4G"}
-export DRILLBIT_CODE_CACHE_SIZE=${DRILLBIT_CODE_CACHE_SIZE:-"1G"}
+export DRILL_MAX_DIRECT_MEMORY=${DRILL_MAX_DIRECT_MEMORY:-"3G"}
+export DRILL_HEAP=${DRILL_HEAP:-"1G"}
+export DRILLBIT_CODE_CACHE_SIZE=${DRILLBIT_CODE_CACHE_SIZE:-"512M"}
 
 export DRILLBIT_OPTS="-Xms$DRILL_HEAP -Xmx$DRILL_HEAP -XX:MaxDirectMemorySize=$DRILL_MAX_DIRECT_MEMORY"
 export DRILLBIT_OPTS="$DRILLBIT_OPTS -XX:ReservedCodeCacheSize=$DRILLBIT_CODE_CACHE_SIZE -Ddrill.exec.enable-epoll=false"
--- a/distribution/src/resources/drill-env.sh
+++ b/distribution/src/resources/drill-env.sh
@@ -50,12 +50,12 @@
 #export DRILLBIT_MAX_PROC_MEM=${DRILLBIT_MAX_PROC_MEM:-"13G"}
 
 # Amount of heap memory for the Drillbit process. Values are those supported by
-# the Java -Xms option. The default is 4G.
+# the Java -Xms option. The recommended minimum is 4G.
 
 #export DRILL_HEAP=${DRILL_HEAP:-"4G"}
 
 # Maximum amount of direct memory to allocate to the Drillbit in the format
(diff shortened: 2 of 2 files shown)


Agirish:
+1. I think this is a welcome change for newbies trying out Drill on personal systems. For production use, the current defaults were anyway not always sufficient and had to be updated.

priteshm:
@paul-rogers, @parthchandra  can you review/ comment on this change?

parthchandra:
The settings for running the unit tests are (from the pom file :
              -Xms512m -Xmx4096m 
              -XX:MaxPermSize=512M -XX:MaxDirectMemorySize=4096M
              -XX:+CMSClassUnloadingEnabled -ea

At the very least, the default settings for Drill and the unit tests should match.

kkhatua:
I think folks running unit tests can continue with the existing limits of 4GB heap+4GB Direct. The idea is to get Drill up and running for minimal use cases, and you are expected to bump up memory to higher limits if queries get OOM errors. Unit tests, if carrying any tests that require much more memory, would qualify under this.
I've launched a [fork](https://github.com/kkhatua/drill/commits/LowerMemUnitTest) of this branch with reduced memory settings for the unit tests as well and letting TravisCI test it out:
https://travis-ci.org/kkhatua/drill/builds/328250155
If it passes, we can either reduce those settings in the POM files as well, or leave it intact to ensure developers doing unit tests are able to run the tests in a reasonable amount of time.

kkhatua:
Ok. Looks like [DRILL-5926](https://issues.apache.org/jira/browse/DRILL-5926) (committed into Apache master) explicitly has a need for bumping up the memory to 4GB for heap alone. I guess we'll need to figure out something else to manage this.

kkhatua:
The primary objective of the JIRA (and this PR) is to allow Drill to truly start up in a bare-min memory footprint. Most people trying Drill would start experimenting with a LocalFS, before trying on a small distributedFS and then eventually scaling out the number of nodes for the DFS.

The challenge has been to figure out what is a reasonably small size for Drill to function without failing. A setting of *5GB* (`1G Heap`+`3G Direct`+`512M CodeCache`) was sufficient for a single node Drill instance to startup and run the TPCH benchmark (single user, via SQLLine) for *scaleFactor=1*; against +CSV-Text+ and uncompressed +Parquet+ formats. Queries that failed were because of HashAgg operator not having sufficient memory. To work around this, as the error message suggested, the HashAgg fallback option was enabled. The heap memory peaked at `800MB`, while the Drillbit process itself maxed out at `2.2GB`.
Functional tests from the drill-test framework also ran to completion without issue.

However, the unit-tests hang because of insufficient memory. The tests typically need about `3GB Heap` and `4GB Direct` (~8GB total), which is defined separately in the `pom.xml` .

Based on this, we have the following options, if we wish to move towards reducing the default memory footprint:

1. Reduce default memory to 5GB, but leave the unit test’s memory requirements (8GB at last check) intact.
  We don’t expect anyone to run unit tests, so if they did run unit tests with the 8GB setting, it would run to completion if the user has enough memory.
  If a user does provide an excesssively large input to process to Drill, it is expected to correctly report insufficient memory availability and there is good documentation for explaining how to increase memory.

2. Reduce default memory to match the unit-test memory requirements of 8GB
  This reduces the memory requirement substantially from the 8GB, but is still high for someone trying out Drill in a memory-constrained VM (or sandboxed) environment. This also means that we need to track the minimum memory requirements of the unit-tests and keep in sync with it. Even if the user is not intending to run unit-tests in the limited-memory environment.

3. Change nothing, but introduce sample files on lines of `$DRILL_HOME/conf/drill-override.conf.example`
  The proposal here would be to have a minimum viable memory footprint config file (e.g. `drill-env.sh.minMem` ), which a first time user of Drill can swap with `./conf/drill-env.sh` when bringing up Drill. The flip side of this is that a Drill trial user would need to have the knowledge of swapping the config file for low-memory usage before starting up Drill.

There seems to be consensus that users don’t feel intimidated to install Drill and start working with small-to-medium workloads on something as low end as a laptop. Based on that, I'd recommend option # 1, which is this PR. I'll leave this PR open to discussion.

jnturton:
I propose that we return to this after Drill 2.0 and the probably ValueVector overhaul.

PR #1087, opened 2018-01-09
DRILL-6079: Attempt to fix memory leak in Parquet

sachouche:
** Problem Description **
This is an extremely rare leak which I was able to emulate by putting a sleep in the AsyncPageReader right after reading the page and before enqueue in the result queue. This is how this issue could manifest itself in real life scenario:
- AsyncPageReader reads a page into a buffer but didn't enqueue yet the result (thread got preempted)
- Parquet Scan thread blocked waiting on the task (Future object dequeued)
- Cancel received and Scan thread interrupted 
- Future.get() returns (Future object is lost)
- Scan thread executes release logic
- Scan thread is not able to interrupt the AsyncPageReader thread since the future object is lost
-  AsyncPageReader thread resumes and enqueues the DrillBuf in the result queue
- This results in a leak since this buffer is not properly released

** Fix Description **
- The fix is straightforward as we peek the Future object during the blocking get() method
- This way, an exception (such as an interrupt) will leave the Future object in the task queue
- The cleanup logic will be able to guarantee the DrillBuf object is either GCed by the AsyncPageReader or ParquetScan thread

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
@@ -17,39 +17,40 @@
  */
 package org.apache.drill.exec.store.parquet.columnreaders;
 
-import com.google.common.base.Stopwatch;
-import io.netty.buffer.DrillBuf;
-import org.apache.drill.common.exceptions.DrillRuntimeException;
-import org.apache.drill.common.exceptions.ExecutionSetupException;
-import org.apache.drill.common.exceptions.UserException;
-import org.apache.drill.exec.ExecConstants;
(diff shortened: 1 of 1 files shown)


sachouche:
@parthchandra can you please review this fix?

arina-ielchiieva:
@sachouche do you have Jira for this fix (if not please create it)? Could you please include Jira number in PR and description and PR commit message (it's a Drill standard) [1, 2, 3].

[2] https://github.com/apache/drill/commits/master
[3] https://github.com/apache/drill/pull/1065

sachouche:
Thank you Arina for catching this; I created the commit for QA before vacation so that they could verify the fix. At that time, I didn't have an Apache JIRA. I have now updated the comment to reflect the JIRA id DRILL-6079 which is also the name of this remote branch.

parthchandra:
+1
Thanks Salim.

PR #1092, opened 2018-01-15
DRILL-6088: MainLoginPageModel errors out when http.auth.mechanisms i…

sohami:
…s not configured


sohami:
@arina-ielchiieva - Can you please help to review this PR ?

sohami:
@arina-ielchiieva - I have rebased the commit's on latest master and made the change as suggested. I have also made the MainLoginPageModel class public since it was causing the issue without that at runtime. Looks like during testing I kept it as public but later while opening the PR I might have changed it to package-private and was seeing issues today again during test.

PR #1094, opened 2018-01-17
DRILL-6090: While connecting to drill-bits using JDBC Driver through …

milindt:
…Zookeeper, a lot of "Curator-Framework-0" threads are created if connection to drill-bit is not successful(no drill-bits are up/reachable)

I am using Drill JDBC driver 1.12.0 to connect to MapR-DB. I am finding the available drill-bits using Zookeepers. When drill-bits are not up or not reachable, the connection is failed with exception: "Failure in connecting to Drill: oadd.org.apache.drill.exec.rpc.RpcException: Failure setting up ZK for client", which is expected, but number of threads created by ZKClusterCoordinator just keeps on increasing.

Steps to reproduce the issue

Setup a connection with a drill-bit using Apache Drill JDBC driver 1.12.0 through Zookeeper hosts(port 5181)
Now stop the drill-bit services or block the drill-bit IPs using iptable rules
Truncate catalina logs
Try to connect to the drill-bit/hit a code path that requires connection to drill-bits.
Take thread dump using kill -QUIT <java process id>
grep -c "Curator-Framework-0" catalina.out
Observe that the curator framework thread just keep on accumulating

RCA:

ZKClusterCoordinator creates curator threads in the constructor
ZKClusterCoordinator is instantiated by DrillClient.connect
DrillClient.connect is called in DrillConnectionImpl constructor

Fix:

Call DrillConnectionImpl .cleanup() from all the catch blocks in the DrillConnectionImpl  constructor.

[JIRA link](https://issues.apache.org/jira/browse/DRILL-6090)

If this fix is accepted, will there be a release from this branch(1.12.0)? A release with minor version may be.

### DIFF
--- a/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillConnectionImpl.java
+++ b/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillConnectionImpl.java
@@ -132,10 +132,12 @@ class DrillConnectionImpl extends AvaticaConnection
             bit = new Drillbit(dConfig, serviceSet);
             bit.run();
           } catch (final UserException e) {
+        	cleanup();
             throw new SQLException(
                 "Failure in starting embedded Drillbit: " + e.getMessage(),
                 e);
           } catch (Exception e) {
+        	cleanup();
             // (Include cause exception's text in wrapping exception's text so
             // it's more likely to get to user (e.g., via SQLLine), and use
             // toString() since getMessage() text doesn't always mention error:)
@@ -167,15 +169,25 @@ class DrillConnectionImpl extends AvaticaConnection
       this.client.setClientName("Apache Drill JDBC Driver");
       this.client.connect(connect, info);
     } catch (OutOfMemoryException e) {
+      cleanup();
       throw new SQLNonTransientConnectionException("Failure creating root allocator", e);
     } catch (InvalidConnectionInfoException e) {
+      cleanup();
       throw new SQLNonTransientConnectionException("Invalid parameter in connection string: " + e.getMessage(), e);
     } catch (RpcException e) {
+      cleanup();
       // (Include cause exception's text in wrapping exception's text so
       // it's more likely to get to user (e.g., via SQLLine), and use
       // toString() since getMessage() text doesn't always mention error:)
       throw new SQLNonTransientConnectionException("Failure in connecting to Drill: " + e, e);
-    }
+    } catch(SQLException e) {
+    	cleanup();
+    	throw e;
+    } catch (Exception e) {
+    	cleanup();
+    	throw new SQLException("Failure in creating DrillConnectionImpl: " + e, e);
+	}
+    
   }
 
 
--- a/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillConnectionImpl.java
+++ b/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillConnectionImpl.java
@@ -108,84 +108,81 @@ class DrillConnectionImpl extends AvaticaConnection
     super.setReadOnly(false);
 
     this.config = new DrillConnectionConfig(info);
+    try{
+      try {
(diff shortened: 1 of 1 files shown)


milindt:
Details in [DRILL-6090](https://issues.apache.org/jira/browse/DRILL-6090)

milindt:
@paul-rogers can you please review this fix?

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillConnectionImpl.java
```diff
@@ -132,10 +132,12 @@ protected DrillConnectionImpl(DriverImpl driver, AvaticaFactory factory,
             bit = new Drillbit(dConfig, serviceSet);
             bit.run();
           } catch (final UserException e) {
+            cleanup();
             throw new SQLException(
                 "Failure in starting embedded Drillbit: " + e.getMessage(),
                 e);
           } catch (Exception e) {
+            cleanup();
```
arina-ielchiieva: I suppose you can use finally block to perform clean up instead of repeating `cleanup` method call after each cached exception.

arina-ielchiieva:
@milindt release 1.12.0 is out. Regarding this fix, once you address code review comments it will be added in 1.13.0.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillConnectionImpl.java
milindt: We don't want clean up to run on successful connection though, we just want to cover all the exceptions. On a successful connection, the "Curator" threads will be created but the number of threads will be proportional to number of connections, which can be be controlled by using a connection pool, hence it won't cause this issue. 

On the other hand,  calling cleanup on a successful connection may terminate that connection.  Thoughts?

arina-ielchiieva: What if you use flag `doCleanup` which will be true by default and after `this.client.connect(connect, info);` succeedes, it would be set to false. In finally block you'll do clean up only if flag is true?

paul-rogers: One trick is this: used nested exceptions. The inner set "parses" the exception types, the outer does cleanup:

```
try {
  try {
    // do something
  } catch (ExceptionA e) {
    // Do something
  } catch (ExceptionB e) {
    // Do something else
  ... }
 } catch (Throwable t) {
    cleanup();
    throw t;
 }
```

arina-ielchiieva:
@milindt could you please address code review comments so changes can be pushed into master?

milindt:
@arina-ielchiieva, @paul-rogers can you please review?

arina-ielchiieva:
+1, LGTM.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillConnectionImpl.java
```diff
@@ -108,73 +108,81 @@ protected DrillConnectionImpl(DriverImpl driver, AvaticaFactory factory,
...
-            throw new SQLException("Failure in starting embedded Drillbit: " + e, e);
+            Class.forName("org.eclipse.jetty.server.Handler");
+          } catch (final ClassNotFoundException e) {
+            throw new SQLNonTransientConnectionException(
+                "Running Drill in embedded mode using Drill's jdbc-all JDBC"
+                + " driver Jar file alone is not supported.",  e);
+          }
+
+          final DrillConfig dConfig = DrillConfig.create(info);
+          this.allocator = RootAllocatorFactory.newRoot(dConfig);
```
paul-rogers: Not your bug: error in original code. Drill JDBC allows creation of multiple connections. Each will have a separate root allocator. This might oversubscribe memory in extreme case. Probably need to file a JIRA for this case.

PR #1096, opened 2018-01-19
DRILL-6099 : Push limit past flatten(project) without pushdown into scan

gparai:
@amansinha100 @chunhui-shi can you please review this PR? Thanks!

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
@@ -88,7 +88,7 @@ import java.util.List;
 public enum PlannerPhase {
   //private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DrillRuleSets.class);
 
-  LOGICAL_PRUNE_AND_JOIN("Loigcal Planning (with join and partition pruning)") {
+  LOGICAL_PRUNE_AND_JOIN("Logical Planning (with join and partition pruning)") {
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(
           getDrillBasicRules(context),
@@ -274,6 +274,7 @@ public enum PlannerPhase {
        */
       DrillPushProjectPastFilterRule.INSTANCE,
       DrillPushProjectPastJoinRule.INSTANCE,
+
       // Due to infinite loop in planning (DRILL-3257), temporarily disable this rule
       //DrillProjectSetOpTransposeRule.INSTANCE,
       RuleInstance.PROJECT_WINDOW_TRANSPOSE_RULE,
@@ -342,8 +343,9 @@ public enum PlannerPhase {
             PruneScanRule.getDirFilterOnScan(optimizerRulesContext),
             ParquetPruneScanRule.getFilterOnProjectParquet(optimizerRulesContext),
             ParquetPruneScanRule.getFilterOnScanParquet(optimizerRulesContext),
-            DrillPushLimitToScanRule.LIMIT_ON_SCAN,
-            DrillPushLimitToScanRule.LIMIT_ON_PROJECT
+            // Include LIMIT_ON_PROJECT since LIMIT_ON_SCAN may not work without it
+            DrillPushLimitToScanRule.LIMIT_ON_PROJECT,
+            DrillPushLimitToScanRule.LIMIT_ON_SCAN
         )
         .build();
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
@@ -30,9 +30,11 @@ import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
+import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexVisitor;
 import org.apache.calcite.rex.RexVisitorImpl;
+import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.util.Pair;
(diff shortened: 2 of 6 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
```diff
@@ -341,7 +346,7 @@ static RuleSet getPruneScanRules(OptimizerRulesContext optimizerRulesContext) {
             ParquetPruneScanRule.getFilterOnProjectParquet(optimizerRulesContext),
             ParquetPruneScanRule.getFilterOnScanParquet(optimizerRulesContext),
             DrillPushLimitToScanRule.LIMIT_ON_SCAN,
-            DrillPushLimitToScanRule.LIMIT_ON_PROJECT
+            DrillPushLimitToScanRule.LIMIT_ON_PROJECT_SCAN
```
chunhui-shi: Not sure if we still need "limit_on_project_scan". In theory, limit_on_project and limit_on_scan should already cover all the cases. Have you tested with "limit_on_project_scan" disabled?

chunhui-shi:
+1

chunhui-shi:
Once all tests are done, I think it is fine to add 'ready-to-commit' label to the JIRA.

gparai:
@chunhui-shi can you please review the new changes (in commit e6dcf14)? Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushLimitToScanRule.java
```diff
@@ -121,4 +132,50 @@ protected void doOnMatch(RelOptRuleCall call, DrillLimitRel limitRel, DrillScanR
     }
 
   }
+
+  private static boolean isProjectFlatten(RelNode project) {
```
chunhui-shi: I think it might be more general to name the functions to schemaUnknown(for conert_fromJson), rowCountUnknown(for flatten), so if in future we have some other functions fall in these two categories, we could easily add these functions to the categories. What do you think?

gparai: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushLimitToScanRule.java
```diff
@@ -55,18 +62,21 @@ public void onMatch(RelOptRuleCall call) {
...
+      RelOptHelper.some(DrillLimitRel.class, RelOptHelper.any(DrillProjectRel.class)), "DrillPushLimitToScanRule_LimitOnProject") {
     @Override
     public boolean matches(RelOptRuleCall call) {
       DrillLimitRel limitRel = call.rel(0);
-      DrillScanRel scanRel = call.rel(2);
-      // For now only applies to Parquet. And pushdown only apply limit but not offset,
+      DrillProjectRel projectRel = call.rel(1);
+      // pushdown only apply limit but not offset,
       // so if getFetch() return null no need to run this rule.
-      if (scanRel.getGroupScan().supportsLimitPushdown() && (limitRel.getFetch() != null)) {
```
amansinha100: I can understand that this check was removed because this matches() method no longer is checking for DrillScanRel, but does that mean that no one is checking the GroupScan for supportsLimitPushdown() ?

priteshm:
@gparai did you get a chance to address @amansinha100 's comment for this PR?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushLimitToScanRule.java
gparai: We still have the LIMIT_ON_SCAN rule which does that check. This rule is changed from LIMIT_PROJECT_SCAN to LIMIT_PROJECT. The LIMIT_SCAN along with the LIMIT_PROJECT would work as the LIMIT_PROJECT_SCAN.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushLimitToScanRule.java
```diff
@@ -55,18 +62,21 @@ public void onMatch(RelOptRuleCall call) {
     }
   };
 
-  public static DrillPushLimitToScanRule LIMIT_ON_PROJECT =
-      new DrillPushLimitToScanRule(
-          RelOptHelper.some(DrillLimitRel.class, RelOptHelper.some(
-              DrillProjectRel.class, RelOptHelper.any(DrillScanRel.class))),
-          "DrillPushLimitToScanRule_LimitOnProject") {
+  public static DrillPushLimitToScanRule LIMIT_ON_PROJECT = new DrillPushLimitToScanRule(
```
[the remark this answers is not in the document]
gparai: There are many instances where we would have a PROJECT on top of the SCAN. The way the rule is refactored now the LIMIT_SCAN rule would not work unless we do LIMIT_PROJECT. Hence, these rules should go together for LIMIT_SCAN to work effectively. That is the reason I kept the rule here rather than creating a new rule.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
```diff
@@ -224,4 +226,64 @@ public Void visitInputRef(RexInputRef inputRef) {
...
+        case DECIMAL:
+          if (((long) l.getValue2()) == 0) {
+            return true;
+          }
+      }
+    }
+    return false;
+  }
+
+  public static boolean isProjectOutputRowcountUnknown(RelNode project) {
```
amansinha100: Could you add javadoc for this utility function.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
```diff
@@ -224,4 +226,64 @@ public Void visitInputRef(RexInputRef inputRef) {
...
+        rex.accept(visitor);
+      }
+    } catch (Util.FoundOne e) {
+      Util.swallow(e, null);
+      return true;
+    }
+    return false;
+  }
+
+  public static boolean isProjectOutputSchemaUnknown(RelNode project) {
```
amansinha100: Javadoc

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
```diff
@@ -224,4 +226,64 @@ public Void visitInputRef(RexInputRef inputRef) {
...
+          }
+      }
+    }
+    return false;
+  }
+
+  public static boolean isProjectOutputRowcountUnknown(RelNode project) {
+    assert project instanceof Project : "Rel is NOT an instance of project!";
+    try {
+      RexVisitor<Void> visitor =
```
amansinha100: Would FLATTEN ever occur within other expressions ?  I believe it always occurs as an independent expression.  If that's the case, it seems to me that having a visitor is overkill.. what do you think ?  Even the original rewrite from project to flatten just iterates over the project exprs here [1].  

[1]  https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/RewriteProjectToFlatten.java#L77

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushLimitToScanRule.java
amansinha100: One implication of this is suppose the underlying Scan does not support Limit pushdown, you could end up with a plan `Scan->Limit->Project->Limit`  where the Limit above the Scan is redundant (assume that there is no FLATTEN in this query).  Can this be avoided ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
gparai: Yes, you are correct. If the rewrite does not consider it as embedded within other expressions then it is fine for the utility function to do the same.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
gparai: Done

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
gparai: Done

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushLimitToScanRule.java
gparai: Without a FLATTEN, the LIMIT would be fully pushed past the PROJECT i.e. we would not have a LIMIT on top of the project.

gparai:
@amansinha100 I have addressed your review comments. Please take a look. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushLimitToScanRule.java
amansinha100: Ok, yeah in that case we are not generating a redundant limit.

amansinha100:
Updated version lgtm.  +1

PR #1099, opened 2018-01-24
DRILL-6106: Use valueOf method instead of constructor since valueOf has  …

reudismam:
… a higher performance by caching frequently requested values.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ssl/SSLConfigClient.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ssl/SSLConfigClient.java
@@ -93,8 +93,11 @@ public class SSLConfigClient extends SSLConfig {
 
   private int getIntProperty(String name, int defaultValue) {
     int value = defaultValue;
-    if ( (properties != null) && (properties.containsKey(name))) {
-      value = new Integer(properties.getProperty(name)).intValue();
+    if (properties != null) {
+      String property = properties.getProperty(name);
+      if (property != null && property.length() > 0) {
+        value = Integer.decode(property);
+      }
     }
     return value;
   }
--- a/contrib/storage-hbase/src/test/java/org/apache/drill/hbase/TestOrderedBytesConvertFunctions.java
+++ b/contrib/storage-hbase/src/test/java/org/apache/drill/hbase/TestOrderedBytesConvertFunctions.java
@@ -54,7 +54,7 @@ public class TestOrderedBytesConvertFunctions extends BaseTestQuery {
 
   @Test
   public void testOrderedBytesDoubleConvertFrom() throws Throwable {
-    verifyPhysicalPlan("convert_from(binary_string('\\x31\\x80\\x00\\x00\\x00\\x00\\x00\\x00\\x01'), 'DOUBLE_OB')", new Double(4.9e-324));
+    verifyPhysicalPlan("convert_from(binary_string('\\x31\\x80\\x00\\x00\\x00\\x00\\x00\\x00\\x01'), 'DOUBLE_OB')", Double.valueOf(4.9e-324));
   }
 
   protected <T> void verifyPhysicalPlan(String expression, T expectedResults) throws Throwable {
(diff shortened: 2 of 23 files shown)


kkhatua:
Hi @reudismam , 
Could you file a JIRA for this PR? Also, you'll need to modify the PR's subject/title to be prefixed with that JIRA number. You can see other PRs for an example. 
This allows us to track and approve PRs in the right process.

@arina-ielchiieva , could you take a look at this?

reudismam:
I have edited the PR to include the Jira prefix.

arina-ielchiieva:
@reudismam the idea of the fix seems reasonable but there are many other places in code where constructor is used instead of `valueOf`. I suggest you revisit all classes rather then doing partial fix.

arina-ielchiieva:
@reudismam classes in protocol package are generated using .proto files, you can not change them.

reudismam:
Reverted changes to classes in the protocol package.

exec/java-exec/src/main/java/org/apache/drill/exec/ssl/SSLConfigClient.java
```diff
@@ -94,7 +94,7 @@ private String getStringProperty(String name, String defaultValue) {
   private int getIntProperty(String name, int defaultValue) {
     int value = defaultValue;
     if ( (properties != null) && (properties.containsKey(name))) {
-      value = new Integer(properties.getProperty(name)).intValue();
+      value = Integer.valueOf(properties.getProperty(name)).intValue();
```
vrozov:
```
    if (properties != null) {
      String property = properties.getProperty(name);
      if (property != null && property.length() > 0) {
        value = Integer.decode(property);
      }
```

reudismam: According to Java documentation, there are benefits in using valueOf for Double and Float.

https://docs.oracle.com/javase/7/docs/api/java/lang/Double.html#valueOf(double)
https://docs.oracle.com/javase/7/docs/api/java/lang/Float.html#valueOf(float)

vrozov: Do not always trust Java doc :). Here is the actual implementation:

```
    public static Double valueOf(double d) {
        return new Double(d);
    }
```

arina-ielchiieva:
@reudismam could you please address code review comments so changes can be pushed into master? Also please resolve merge conflict.

reudismam:
Conflicts resolved. Do you want to revert the edits to Double and Float? I think it is ok for consistency.

arina-ielchiieva:
@reudismam you can leave `valueOf` for Double and Float for consistency. 
There are two other comments @vrozov has left. Could you please address hem as well? They are not directly related to your changes but will make Drill code better. Also please squash commits into one.

vrozov:
What was the reason to revert the change in SSLConfigClient.java?

reudismam:
I reverted the edit because the code does not pass in Travis CI.

vrozov:
Travis CI failed due to a known issue `TestSortImpl.testLargeBatch:502->runJumboBatchTest:475->runLargeSortTest:454 Value of 1:0 expected:<0> but was:<1>` unrelated to the SSLConfigClient.java changes.

arina-ielchiieva:
@reudismam I guess you can apply back your change for `SSLConfigClient.java` class.

arina-ielchiieva:
LGTM, @vrozov?

reudismam:
Since I have to resolve some conflicts after applying the edit, squashing could be tricky because the commits are non-consecutive. I allowed the merging using the squashing option in the forked project setting. Thus, when someone merges the pull request, he or she can decide to squash the commits.

image

vrozov:
@reudismam For drill, merging/squashing on github is not an option (due to the way how Apache Drill git is setup). Please rebase and squash.

reudismam:
Maybe it has not worked as expected. It squashed the commit (first commit), but as the commit mix commits from other persons, they come together. Maybe it will be the case of creating a patch file for the desired commit and apply this patch to a new pull request.

arina-ielchiieva:
Well, you can always use force push to override your previous changes or even replace your remote branch with new local.

reudismam:
I have squashed the commits, but I’m getting an error in Travis CI similar to the previous one when I reverted some changes.
Column a-offsets of type UInt4Vector: Offset (0) must be 0 but was 1

reudismam:
Only pass Travis CI by removing the edits to SSLConfigClient.java

vrozov:
@reudismam Travis fails in other PRs as well. See #1105.

PR #1101, opened 2018-01-27
DRILL-6032: Made the batch sizing for HashAgg more accurate.

ilooner:
## RecordBatchSizer changes

 - The RecordBatchSizer previously computed fixed width column sizes by measuring the total size of a vector and dividing by the number of elements. Because of this the RecordBatchSizer would return a zero size for FixedWidth vectors that had no data. So I added a method to FixedWidth vectors to get the size of a record and use that method to compute the column width in the RecordBatchSizer.  
 - In some cases it was possible for the RecordBatchSizer to return a column width of 0, when it is not possible to have vectors with a width of less than 1 in practice. So I made the minimum column width returned by the RecordBatchSizer 1.
 
## HashAgg changes

 - Removed commented out code and unused variables.
 - Removed if statements for printing debug statements and instead used logger.debug
 - Removed the extraNonNullColumns and extraRowBytes tweak parameters for computing the sizes of batches
 - The RecordBatchSizer is used to compute the width of each column instead of adhoc custom logic. 
 - Using the real width of each column to estimate column sizes instead of taking the max width of all columns and assuming each column has the max width
 - Removed the assumption that varchars will not exceed 50 characters in length
 - Removed unnecessary condition checks in delayedSetup and updateEstMaxBatchSize


ilooner:
@Ben-Zvi please review

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -215,6 +206,7 @@ public BatchHolder() {
           MaterializedField outputField = materializedValueFields[i];
           // Create a type-specific ValueVector for this value
           vector = TypeHelper.getNewVector(outputField, allocator);
+          int columnSize = new RecordBatchSizer.ColumnSize(vector).estSize;
```
ppadma: I wonder what is wrong if estSize is 0 when there is no data. If there is no data for a column, why would we want to add it's value width to outgoing row width ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -397,11 +384,9 @@ private void delayedSetup() {
     }
     numPartitions = BaseAllocator.nextPowerOfTwo(numPartitions); // in case not a power of 2
 
-    if ( schema == null ) { estValuesBatchSize = estOutgoingAllocSize = estMaxBatchSize = 0; } // incoming was an empty batch
```
Ben-Zvi: Why was the ( schema == null ) check removed ?  Without it, calling updateEstMaxBatchSize() would return an NPE when accessing the batch's schema (when calling RecordBatchSizer(incoming) ).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -516,43 +501,48 @@ private void initializeSetup(RecordBatch newIncoming) throws SchemaChangeExcepti
...
     // (When there are no aggr functions, use '1' as later code relies on this size being non-zero)
+    // Note: estValuesBatchSize cannot be 0 because a zero value for estValuesBatchSize will cause reserveValueBatchMemory to have a value of 0. And the meaning
+    // of a reserveValueBatchMemory value of 0 has multiple meanings in different contexts. So estValuesBatchSize has an enforced minimum value of 1, without this
+    // estValuesBatchsize could have a value of 0 in the case were there are no value columns and all the columns are key columns.
     estValuesBatchSize = Math.max(estValuesRowWidth, 1) * MAX_BATCH_SIZE;
     estOutgoingAllocSize = estValuesBatchSize; // initially assume same size
 
-    logger.trace("{} phase. Estimated internal row width: {} Values row width: {} batch size: {}  memory limit: {}  max column width: {}",
-        isTwoPhase?(is2ndPhase?"2nd":"1st"):"Single",estRowWidth,estValuesRowWidth,estMaxBatchSize,allocator.getLimit(),maxColumnWidth);
+    if (logger.isTraceEnabled()) {
```
Ben-Zvi: Why is this "if()" needed ? If trace is not enabled, there is no logging ....

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -516,43 +501,48 @@ private void initializeSetup(RecordBatch newIncoming) throws SchemaChangeExcepti
...
+      keySizes.put(columnName, columnSize);
+      estOutputRowWidth += columnSize;
     }
+
+    long estValuesRowWidth = 0;
+
+    for (int columnIndex = numGroupByOutFields; columnIndex < outContainer.getNumberOfColumns(); columnIndex++) {
+      VectorWrapper vectorWrapper = outContainer.getValueVector(columnIndex);
+      RecordBatchSizer.ColumnSize columnSize = new RecordBatchSizer.ColumnSize(vectorWrapper.getValueVector());
+      estOutputRowWidth += columnSize.estSize;
```
Ben-Zvi: What is missing (maybe in the original code as well) is those extras: The Values batch sometimes uses those 8 byte vectors to represent nullable, and for some aggregations (e.g. STDDEV), internally there are two columns ("count" and "sum of squares", or something like that), but the outContainer has only one - with the result (this is why the Values vectors in the outContainer are always allocated, instead of passing the existing vectors - see DRILL-5588 )

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -84,13 +85,6 @@
 public abstract class HashAggTemplate implements HashAggregator {
   protected static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(HashAggregator.class);
 
-  private static final int VARIABLE_MAX_WIDTH_VALUE_SIZE = 50;
-  private static final int VARIABLE_MIN_WIDTH_VALUE_SIZE = 8;
-
-  private static final boolean EXTRA_DEBUG_1 = false;
```
Ben-Zvi: This is Aman's implementation of an #ifdef .....
Basically he wanted to preserve these debugging messages, but not seeing them (there are too many) while doing some other debugging work.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -956,21 +925,8 @@ private void spillAPartition(int part) {
       this.htables[part].outputKeys(currOutBatchIndex, this.outContainer, outStartIdxHolder.value, outNumRecordsHolder.value, numPendingOutput);
 
       // set the value count for outgoing batch value vectors
-      /* int i = 0; */
       for (VectorWrapper<?> v : outgoing) {
         v.getValueVector().getMutator().setValueCount(numOutputRecords);
-        /*
```
Ben-Zvi: We don't have any facility to take a peek at actual data for debugging. While this commented code looks ugly, it allows for easy enabling to look at data, if needed (or an example of how to do so, for elsewhere).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Logging frameworks were invented to solve this problem. We should just use the logging framework.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Yes nothing gets printed, but the following argument still gets evaluated even when trace is disabled.

```
isTwoPhase ? (is2ndPhase ? "2nd" : "1st") : "Single"
```

I wanted to avoid any performance overhead incurred by executing that statement every time.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: If this logic is really critical for debugging we should have utility methods to help with it so that others can benefit from the code. Maybe we can brainstorm about what methods would be helpful for debugging and start building a utility library. In the meantime if you need this code to debug you can easily look into the history and copy and paste it from there.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -805,7 +803,12 @@ private IntVector allocMetadataVector(int size, int initialValue) {
   }
 
   @Override
-  public void setMaxVarcharSize(int size) { maxVarcharSize = size; }
+  public void setKeySizes(Map<String, Integer> keySizes) {
+    Preconditions.checkNotNull(keySizes);
+
+    this.keySizes = CaseInsensitiveMap.newHashMap();
```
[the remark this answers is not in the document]
ilooner: It helps to avoid bugs. It is assumed that the keySizes map will never change once it is set, copying the map helps enforce that constraint. If we don't copy the map and a user calls this method and passes a keySizes map and then later updates the keySizes map or reuses it, errors would occur. Some languages like Scala have immutable flavors of data structures for this reason.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -129,11 +143,16 @@ public ColumnSize(ValueVector v, String prefix) {
         // No standard size for Union type
         dataSize = v.getPayloadByteCount(valueCount);
         break;
+      case GENERIC_OBJECT:
+        // We cannot provide a size for Generic Objects
```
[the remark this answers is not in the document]
ilooner: Execution gets here in some of the HashAgg functional tests. Probably in the case where varchars are aggregated, since as you explained to me varchars are stored in object vectors on heap.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -232,9 +251,8 @@ else if (width > 0) {
     }
   }
 
-  public static final int MAX_VECTOR_SIZE = ValueVector.MAX_BUFFER_SIZE; // 16 MiB
-
   private List<ColumnSize> columnSizes = new ArrayList<>();
+  private Map<String, ColumnSize> columnSizeMap = CaseInsensitiveMap.newHashMap();
```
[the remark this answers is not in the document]
ilooner: I will follow up on this on the mailing list. I was under the impression that Drill is case-**in**sensitive and did not support case sensitivity of columns in an external store like HBase or MapRDB. If Drill is case-sensitive then we have a bigger issue because some of the functional tests assume that it is case-**in**sensitive.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/agg/TestHashAggrSpill.java
```diff
@@ -53,59 +51,54 @@
...
   public final BaseDirTestWatcher dirTestWatcher = new BaseDirTestWatcher();
 
-    /**
-     *  A template for Hash Aggr spilling tests
-     *
-     * @throws Exception
-     */
-    private void testSpill(long maxMem, long numPartitions, long minBatches, int maxParallel, boolean fallback ,boolean predict,
-                           String sql, long expectedRows, int cycle, int fromPart, int toPart) throws Exception {
-        LogFixture.LogFixtureBuilder logBuilder = LogFixture.builder()
```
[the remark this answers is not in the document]
ilooner: I can add this back. In general I don't like leaving unused code in the codebase though.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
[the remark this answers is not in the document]
paul-rogers: Drill is case insensitive internally. The case insensitive map is correct. Thanks for catching this @ilooner!

Unfortunately, record batches have no name space: they are just a collection of vectors. So, we could end up with columns called both "c" and "C". This situation will case the column size map to end up with one entry for both columns, with the last writer winning.

The best solution would be to enforce name space uniqueness when creating vectors. The new "result set loader" does this, but I suspect other readers might not -- depending on the particular way that they create their vectors. Still, creating names that differ only in case is a bug and any code doing that should be fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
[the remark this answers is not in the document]
paul-rogers: The `GENERIC_OBJECT` type is used any time we do a system table query: system tables are represented as Java objects.

There is an open question about certain aggregate functions. (See a note sent a week or so ago.) These aggregate functions use an `ObjectHolder` as their `\@Workspace`. @Ben-Zvi and I discussed whether such aggregates are spillable. This may be an unresolved issue.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
paul-rogers: Agree that the "sizer" should return the size of a fixed-width column even if there are 0 rows. Thanks for catching and fixing this. Though, I will point out that if you get a 0-length batch, just throw it away and read another; no useful purpose is served by doing anything with an empty batch (except to pass it along for schema purposes if this is the first batch: the so-called "fast schema" that is supposed to work, but seems to not be fully implemented...)

I've found that Drill behaves poorly and erratically when trying to allocate zero-sized vectors. Best not to do that. If your batch size is 0, don't allocate vectors would be my general advice to avoid lots of grief.

ilooner: @ppadma I thought estSize represented the estimated column width. For FixedWidth vectors we know the exact column width, so why can't we use the exact value? Also why are there two different things for measuring column sizes, when do you use RecordBatchSizer and when do you use TypeHelper?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: All the unit and functional tests passed without an NPE. The null check was redundant because the code in **doWork** that calls **delayedSetup** sets the schema if it is null.

```
      // This would be called only once - first time actual data arrives on incoming
      if ( schema == null && incoming.getRecordCount() > 0 ) {
        this.schema = incoming.getSchema();
        currentBatchRecordCount = incoming.getRecordCount(); // initialize for first non empty batch
        // Calculate the number of partitions based on actual incoming data
        delayedSetup();
      }
```

So schema will never be null when delayed setup is called

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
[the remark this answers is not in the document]
ilooner: Thanks for the explanation here and on the dev list @paul-rogers.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ppadma: @ilooner That is the point. If we know the exact value, why do we need RecordBatchSizer ? we should use RecordBatchSizer when we need to get sizing information for a batch (in most cases, incoming batch). In this case, you are allocating memory for value vectors for the batch you are building. For fixed width columns, you can get the column width size for each type you are allocating memory for using TypeHelper.getSize. For variable width columns, TypeHelper.getSize assumes it is 50 bytes.  If you want to adjust memory you are allocating for variable width columns for outgoing batch based on incoming batch, that's when you use RecordBatchSizer on actual incoming batch to figure out the average size of that column.  You can also use RecordBatchSizer on incoming batch if you want to figure out how many values you want to allocate memory for in the outgoing batch. Note that, with your change, for just created value vectors, variable width columns will return estSize of 1, which is not what you want.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -255,7 +254,6 @@ private HashAggregator createAggregatorInternal() throws SchemaChangeException,
       groupByOutFieldIds[i] = container.add(vv);
     }
 
-    int extraNonNullColumns = 0; // each of SUM, MAX and MIN gets an extra bigint column
```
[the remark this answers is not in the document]
ilooner: Thanks for catching this. Then we should fix the underlying problem instead of passing around additional parameters to work around the issue. I will work on fixing the codegen for the BatchHolder as part of this PR.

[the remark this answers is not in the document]
Ben-Zvi: Maybe do this work as a separate PR (for DRILL-5728) ?  Else it would delay this PR, and overload it ...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Oh but there is! slf4j and logback have a feature called markers, which allows you to associate a tag with a statement. When you print logs you can specify to filter by level and by marker. There is a working example here https://examples.javacodegeeks.com/enterprise-java/slf4j/slf4j-markers-example/ . I will update the log statements to use markers in this PR.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
paul-rogers: I can think of three reasons to use the sizer:

* Type logic is complex: we have multiple sets of rules depending on the data type. Best to encapsulate the logic in a single place. So, either 1) use the "sizer", or 2) move the logic from the "sizer" to a common utility.
* Column size is tricky as it depends on `DataMode`. The size or a `Required INT` is 4. The (total memory) size of an `Optional INT` is 5. For a `Repeated INT`? You need to know the average array cardinality, which the "sizer" provides (by analyzing an input batch.)
* As discussed, variable-width columns (`VARCHAR`, `VARBINARY` for HBase) have no known size. We really have to completely forget about that awful "50" estimate. We can only estimate size from input, which is, again, what the "sizer" does.

Of course, all the above only works I you actually sample the input.

A current limitation (and good enhancement) is that the Sizer is aware of just one batch. The sort (the first user of the "sizer") needed only aggregate row size, so it just kept track of the widest row ever seen. If you need detailed column information, you may want another layer: one that aggregates information across batches. (For arrays and variable-width columns, you can take the weighted average or the maximum depending on your needs.)

Remember, if the purpose of this number is to estimate memory use, then you have to add a 33% (average) allowance for internal fragmentation. (Each vector is, on average, 75% full.)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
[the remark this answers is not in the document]
ilooner: Ok. I am going to move all the code into one place though. Previously half the code was in HashAggBatch and the other half was in HashAggTemplate. I'll also add a comment that this extra code should be removed after DRILL-5728, and I will also add a comment to DRILL-5728 to reminder the implementer to remove this extra code.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Thanks for the info @paul-rogers . Since we want to have the RecordBatchSizer be a one stop shop for size related information I propose to do the following to address @ppadma 's concerns:

 - Remove changes to estSize in the RecordBatchSizer that I introduced.
 - Add a field called knownSize to the RecordBatchSizer. This field will contain the known column width for FixedWidthVectors and will be -1 for things like varchars and repeated columns. Then we can add a method getKnownSize() for a column size. This method will return the known size if available and will throw an exception if the column is not FixedWidth. So this method should only be used by the caller when they know they are dealing with FixedWidth columns as is the case with HashAgg for now (things will probably change when strings will be stored in varchars instead of objects). The rest of RecordBatchSizer will remain as is.
 - Add a constant called FRAGMENTATION_FACTOR (1.33) that people can use to multiply their custom crafted batch sizes by when they are estimating batch sizes from scratch. In HashAgg I would multiply the custom batch size I compute by this factor.

Thoughts?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: We have to account for the unknown columns in the value batch estimate, but the columns in the output batch are well defined and do not contain these "extra" columns. So why do we have to augment the estimate for the output batch? Why is looking at the columns in the outputContainer insufficient?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
paul-rogers: See the `RowSet` classes. Chunhui and others have successfully used these to print bathes for debugging. Roughly:

```
RowSet rs = DirectRowSet.fromContainer(container);
rs.print();
```

This can be modified to take a row range (to avoid, say, printing 64K rows.) There are versions to do the printing using an SV2 or SV4.

In the code I've not yet posted is a utility to print the contents of an offset vector. Very handy when your offsets get messed up.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
paul-rogers: The columns in the output batch for the hash agg are usually required numerics (fixed width). But, the can be nullable in some cases (talk to Boaz.) If you are doing a max or min on a string, then the output will be a VarChar, so you have to account for the offset and data vectors.

If you want to know how much memory a vector takes after creating it, just measure memory before and after the vector allocation. If you want to predict the size to do an allocation, then for fixed width you just need a row count. For VarChar, you need to predict the width also.

You only need the "sizer" when trying to predict a vector size before you create it, and the vector has variable-width or complex parts.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Thanks!

ilooner:
@Ben-Zvi I have responded to comments and implemented requested changes. Please see latest commits for changes. I have made some additional changes after noticing that some of the batch sizing calculations were incorrect, I have also made various documentation and naming changes to improve the readability of the code and to document what needs to be improved in the future.:

- The size of varchar columns were not properly accounted for in the outgoing batch in the case where varchars are aggregated. I have added logic to the updateEst... method to account for this.
- I have added docs to the updateEst method explaining what the various batch sizing estimates do.
- I have changed the variable names for the batch size estimates to more accurately reflect what they do.
- Previously if a batch size estimate was smaller than the actual amount of memory allocated, the estimate was updated to be the larger size. I think this was actually the wrong behavior since it causes the HashAgg operator's memory estimates to be extremely aggressive and in the worst case can cause the operator to run out of memory prematurely. Ideally a good estimate will over estimate 50% of the time and under estimate 50% of the time.
- I have changed the HashAgg defaults. Although tests passed with the previous defaults, I felt they were too aggressive. I have changed the default number of partitions to 16 and the minimum number of batches to 1.

Let me know if you have anymore comments. Thanks.

ilooner:
An additional comment is that after the defaults were reduced the number of random failures we see in our functional tests decreased to just 3. Typically there are more than that.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -733,28 +780,32 @@ private void restoreReservedMemory() {
...
+
+      if (columnSizer.hasKnownSize()) {
+        // For fixed width vectors we know the size of each record
+        columnSize = columnSizer.getKnownSize();
+      } else {
+        // For var chars we need to use the input estimate
+        columnSize = varcharValueSizes.get(columnIndex);
+      }
+
+      AllocationHelper.allocatePrecomputedChildCount(vv, records, columnSize, 0);
```
ppadma: I think we should also get elementCount from sizer and use that instead of passing 0.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -65,6 +70,14 @@
 
     public int stdSize;
 
+    /**
+     * If the we can determine the exact width of the row of a vector upfront,
+     * the row widths is saved here. If we cannot determine the exact width
+     * (for example for VarChar or Repeated vectors), then
+     */
+
+    private int knownSize = -1;
```
ppadma: Like I mentioned in other comment, seems like we can just use stdSize.

exec/vector/src/main/codegen/templates/FixedValueVectors.java
```diff
@@ -298,6 +298,11 @@ public int getPayloadByteCount(int valueCount) {
     return valueCount * ${type.width};
   }
 
+  @Override
+  public int getValueWidth() {
```
ppadma: If we are using stdSize and get the value from TypeHelper, we don't need all value vectors to have this new function.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -140,6 +131,9 @@
   private OperatorContext oContext;
   private BufferAllocator allocator;
 
+  private Map<String, Integer> keySizes;
+  // The size estimates for varchar value columns. The keys are the index of the varchar value columns.
+  private Map<Integer, Integer> varcharValueSizes;
```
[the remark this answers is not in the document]
ilooner: As far as I know we don't support aggregations on repeated types. Varchar is the only non FixedWidth type we can aggregate.

exec/vector/src/main/codegen/templates/FixedValueVectors.java
ilooner: TypeHelper doesn't return the correct size for Nullable columns. For example NullableIntVector has a width of 5 instead of 4.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
ilooner: Thanks for catching this. I'll use stdSize instead.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -226,7 +221,7 @@ public BatchHolder() {
             ((FixedWidthVector) vector).allocateNew(HashTable.BATCH_SIZE);
           } else if (vector instanceof VariableWidthVector) {
             // This case is never used .... a varchar falls under ObjectVector which is allocated on the heap !
-            ((VariableWidthVector) vector).allocateNew(maxColumnWidth, HashTable.BATCH_SIZE);
+            ((VariableWidthVector) vector).allocateNew(columnSize, HashTable.BATCH_SIZE);
```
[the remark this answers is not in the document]
ilooner: Thanks for catching this. It should use stdSize here.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Hmm. The element count from the sizer would tell us the number of rows in the incoming batch. But that will not give us an accurate prediction for the number of keys we have aggregations for. Maybe we should use the number of keys in the hashtable instead.

ilooner:
@ppadma Responded to your comments. Please look at the last three commits for changes

ilooner:
Travis failure is unrelated

ilooner:
@Ben-Zvi I have addressed review comments. All unit tests and functional tests (except for typical random failures are passing). I have included a commit for DRILL-6144 in this PR in order to fix other issues that were breaking our builds and validate that these changes pass all tests. After these changes are +1 I will remove the commit for DRILL-6144. I will open a separate PR to merge the changes in DRILL-6144.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/agg/TestHashAggrSpill.java
```diff
@@ -93,7 +93,7 @@ private void testSpill(long maxMem, long numPartitions, long minBatches, int max
   @Test
   public void testSimpleHashAggrSpill() throws Exception {
     testSpill(68_000_000, 16, 2, 2, false,
-      true, null, 1_200_000, 1, 1, 1, 3);
+      true, null, 1_200_000, 1, 2, 1, 3);
```
Ben-Zvi: This is supposed to be a "simple spill" (see comment above); not a "recursive" one (with cycle > 1).
Maybe just reducing the number of rows would do the trick ....

ilooner: Fixed. I increased the memory slightly.

ilooner:
@Ben-Zvi I applied your review comment, please let me know if you have any more comments.

priteshm:
@Ben-Zvi can you please do a final review?

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -427,8 +427,8 @@ drill.exec.options: {
     exec.enable_union_type: false,
     exec.errors.verbose: false,
     exec.hashagg.mem_limit: 0,
-    exec.hashagg.min_batches_per_partition: 2,
-    exec.hashagg.num_partitions: 32,
+    exec.hashagg.min_batches_per_partition: 1,
```
Ben-Zvi: This option was meant to create a "slack". **1** is the lowest value - requiring only 1 batch per each partition, i.e., no slack; so that requires the memory computations to be more precise now !!

ilooner: @Ben-Zvi This setting controls the minimum number of batches kept in memory per partition. Making this larger will cause us to consume more memory. Making it smaller makes us consume less memory. Also in general the purpose of this PR was to make the memory calculations more precise and deterministic and it passes all regression tests.

priteshm:
Spoke with Chun, he will run the tests and update the PR with the test results.

cchang738:
My test fail with OOM. @ilooner has test log.

kkhatua:
@ilooner  do you has the fixes for the failure reported by @cchang738 ? It also looks like the PR needs rebasing.

ilooner:
While making this change I learned that it is extremely difficult to make changes to the HashAgg operator, specifically the memory calculations. After a month of work I was not able to bring even a simple change like this to a state where it would not OOM. Others experience similar frustrations when they make changes that either impact the HashAgg operator directly, or indirectly. Considering the fact that a comprehensive and stable memory calculator was created for HashJoin in less than a month, I feel that there is an easier way forward.

As a result of this experience I was motivated to create https://issues.apache.org/jira/browse/DRILL-6253 which focuses on refactoring and unit testing the memory calculations for HashAgg . DRILL-6253 will fulfill the original intention of this PR, which was to make the memory calculations more accurate, and also improve the general stability and testability of HashAgg. As a result of this, I am going to abandon this PR and all future work will be done on the PR for DRILL-6253 once it is opened.

PR #1103, opened 2018-01-31
DRILL-6124: Fixed possible NPE when no injection site is defined for …

ilooner:
…"partitioner-sender-latch" in PartitionerDecorator.

```
injector.getLatch(context.getExecutionControls(), "partitioner-sender-latch");
```

Can return null if the user does not have the appropriate injection defined in their config.


ilooner:
@sachouche @arina-ielchiieva

arina-ielchiieva:
@ilooner it looks like if latch is not found, execution control will return dummy latch [1]? If I am missing something, please explain.

[1] https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/testing/ExecutionControls.java#L206

ilooner:
You are right @arina-ielchiieva . Thanks for catching this, I will close the PR and mark the jira as invalid.

PR #1104, opened 2018-01-31
DRILL-6118: Handle item star columns during project / filter push dow…

arina-ielchiieva:
…n and directory pruning

1. Added DrillFilterItemStarReWriterRule to re-write item star fields to regular field references.
2. Refactored DrillPushProjectIntoScanRule to handle item star fields, factored out helper classes and methods from PreUitl.class.
3. Fixed issue with dynamic star usage (after Calcite upgrade old usage of star was still present, replaced WILDCARD -> DYNAMIC_STAR  for clarity).
4. Added unit tests to check project / filter push down and directory pruning with item star.

Details in [DRILL-6118](https://issues.apache.org/jira/browse/DRILL-6118).


arina-ielchiieva:
@chunhui-shi please review.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillFilterItemStarReWriterRule.java
```diff
@@ -0,0 +1,232 @@
...
+    super(operand, id);
+  }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    Filter filterRel = call.rel(0);
+    Project projectRel = call.rel(1);
+    TableScan scanRel = call.rel(2);
+
+    ItemStarFieldsVisitor itemStarFieldsVisitor = new ItemStarFieldsVisitor(filterRel.getRowType().getFieldNames());
```
chunhui-shi: To make sure this works for complex expressions, could you add tests similar to testPushProjectIntoScan to test some expressions that include more than two fields in expressions in project(e.g. "select a+b") and in filter(e.g. "where a+b < 5")?

chunhui-shi: Other test cases should be covered are: 
nested field names, 
refer to two different fields under the same parent, eg. a.b and a.c.
and array type referred in filters and projects.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -596,10 +596,10 @@ private void classifyExpr(final NamedExpression ex, final RecordBatch incoming,
     final NameSegment ref = ex.getRef().getRootSegment();
     final boolean exprHasPrefix = expr.getPath().contains(StarColumnHelper.PREFIX_DELIMITER);
     final boolean refHasPrefix = ref.getPath().contains(StarColumnHelper.PREFIX_DELIMITER);
-    final boolean exprIsStar = expr.getPath().equals(SchemaPath.WILDCARD);
-    final boolean refContainsStar = ref.getPath().contains(SchemaPath.WILDCARD);
-    final boolean exprContainsStar = expr.getPath().contains(SchemaPath.WILDCARD);
-    final boolean refEndsWithStar = ref.getPath().endsWith(SchemaPath.WILDCARD);
+    final boolean exprIsStar = expr.getPath().equals(SchemaPath.DYNAMIC_STAR);
```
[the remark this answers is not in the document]
vvysotskyi: This change became required after Calcite update. With the changes in CALCITE-1150, `*` is replaced by `**` after a query is parsed and `**` is added to the RowType. Therefore WILDCARD can't come from the plan and its usage should be replaced by `**`.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillFilterItemStarReWriterRule.java
arina-ielchiieva: @chunhui-shi added more unit tests. Please review.

amansinha100:
+1

amansinha100:
@arina-ielchiieva pls rebase on latest master.

arina-ielchiieva:
@amansinha100 rebased and ran pre-commit tests. All good expept of 16 expected failures.

arina-ielchiieva:
Merged into Apache master with commit id 9073aed67d89e8b2188870d6c812706085c9c41b

paul-rogers:
I noticed this change when merging with the latest master. What does `DYNAMIC_STAR` mean? The original `WILDCARD` name was more theoretical. How is `DYNAMIC_STAR` different from `STAR` (other than being `**`)?

Can we call it `WILDCARD` and `isWildcard()` that the string is `**`, presumably for internal reasons, seems less worth calling out in the name.

vvysotskyi:
@paul-rogers, as you know, there were some differences between the commits in the older Calcite version (1.4.0-r23) which was used by Drill and commits merged into Apache master. One of such commits was `[CALCITE-1150] Add dynamic record type and dynamic star for schema-on-read table`. In our older Calcite fork `DYNAMIC_STAR` was `*`, but after rebase onto Calcite 1.15, it became `**`.

As I understand, `DYNAMIC_STAR` itself means "default column" which will be added to table row type for the case when we have `select *` query and the schema of the underlying table is not known. 

As for the name of constant, I think current `DYNAMIC_STAR` more suitable, since Calcite also uses a similar name for the same string: `DynamicRecordType.DYNAMIC_STAR_PREFIX`. So current name at least helps to avoid divergence in naming for Drill and Calcite.

PR #1105, opened 2018-01-31
DRILL-6125: Fix possible memory leak when query is finished early.

ilooner:
A detailed description of the problem and solution can be found here: 

https://issues.apache.org/jira/browse/DRILL-6125

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
@@ -65,14 +65,13 @@ public class PartitionSenderRootExec extends BaseRootExec {
   private PartitionerDecorator partitioner;
 
   private ExchangeFragmentContext context;
-  private boolean ok = true;
   private final int outGoingBatchCount;
   private final HashPartitionSender popConfig;
   private final double cost;
 
   private final AtomicIntegerArray remainingReceivers;
   private final AtomicInteger remaingReceiverCount;
-  private volatile boolean done = false;
+  private boolean done = false;
   private boolean first = true;
   private boolean closeIncoming;
 
@@ -146,11 +145,8 @@ public class PartitionSenderRootExec extends BaseRootExec {
 
   @Override
   public boolean innerNext() {
-    if (!ok) {
-      return false;
-    }
-
     IterOutcome out;
+
     if (!done) {
       out = next(incoming);
     } else {
@@ -252,13 +248,11 @@ public class PartitionSenderRootExec extends BaseRootExec {
             startIndex, endIndex);
       }
 
-      synchronized (this) {
-        partitioner = new PartitionerDecorator(subPartitioners, stats, context);
-        for (int index = 0; index < terminations.size(); index++) {
-          partitioner.getOutgoingBatches(terminations.buffer[index]).terminate();
-        }
-        terminations.clear();
+      partitioner = new PartitionerDecorator(subPartitioners, stats, context);
+      for (int index = 0; index < terminations.size(); index++) {
+        partitioner.getOutgoingBatches(terminations.buffer[index]).terminate();
       }
+      terminations.clear();
 
       success = true;
     } finally {
@@ -328,12 +322,10 @@ public class PartitionSenderRootExec extends BaseRootExec {
   public void receivingFragmentFinished(FragmentHandle handle) {
     final int id = handle.getMinorFragmentId();
     if (remainingReceivers.compareAndSet(id, 0, 1)) {
-      synchronized (this) {
-        if (partitioner == null) {
-          terminations.add(id);
-        } else {
-          partitioner.getOutgoingBatches(id).terminate();
-        }
+      if (partitioner == null) {
+        terminations.add(id);
+      } else {
+        partitioner.getOutgoingBatches(id).terminate();
       }
 
       int remaining = remaingReceiverCount.decrementAndGet();
@@ -347,7 +339,7 @@ public class PartitionSenderRootExec extends BaseRootExec {
   public void close() throws Exception {
     logger.debug("Partition sender stopping.");
     super.close();
-    ok = false;
+
     if (partitioner != null) {
       updateAggregateStats();
       partitioner.clear();
@@ -358,7 +350,7 @@ public class PartitionSenderRootExec extends BaseRootExec {
     }
   }
 
-  public void sendEmptyBatch(boolean isLast) {
+  private void sendEmptyBatch(boolean isLast) {
     BatchSchema schema = incoming.getSchema();
     if (schema == null) {
       // If the incoming batch has no schema (possible when there are no input records),
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
@@ -19,7 +19,9 @@ package org.apache.drill.exec.work.fragment;
 
 import java.io.IOException;
 import java.security.PrivilegedExceptionAction;
+import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -31,7 +33,6 @@ import org.apache.drill.exec.coord.ClusterCoordinator;
 import org.apache.drill.exec.exception.OutOfMemoryException;
 import org.apache.drill.exec.ops.ExecutorFragmentContext;
 import org.apache.drill.exec.ops.FragmentContext;
-import org.apache.drill.exec.ops.FragmentContextImpl;
 import org.apache.drill.exec.physical.base.FragmentRoot;
 import org.apache.drill.exec.physical.impl.ImplCreator;
 import org.apache.drill.exec.physical.impl.RootExec;
@@ -49,14 +50,58 @@ import org.apache.drill.exec.work.foreman.DrillbitStatusListener;
 import org.apache.hadoop.security.UserGroupInformation;
 
 /**
- * Responsible for running a single fragment on a single Drillbit. Listens/responds to status request
- * and cancellation messages.
+ * <h2>Overview</h2>
+ * <p>
+ *   Responsible for running a single fragment on a single Drillbit. Listens/responds to status request and cancellation messages.
+ * </p>
+ * <h2>Theory of Operation</h2>
+ * <p>
+ *  The {@link FragmentExecutor} runs a fragment's {@link RootExec} in the {@link FragmentExecutor#run()} method in a single thread. While a fragment is running
+ *  it may be subject to termination requests. The {@link FragmentExecutor} is reponsible for gracefully handling termination requests for the {@link RootExec}. There
+ *  are two types of termination messages:
+ *  <ol>
+ *    <li><b>Cancellation Request:</b> This signals that the fragment and therefore the {@link RootExec} need to terminate immediately.</li>
+ *    <li><b>Receiver Finished:</b> This signals that a downstream receiver no longer needs anymore data. A fragment may recieve multiple receiver finished requests
+ *    (one for each downstream receiver). The {@link RootExec} will only terminate once it has recieved {@link FragmentExecutor.EventType#RECEIVER_FINISHED} messages
+ *    for all downstream receivers.</li>
+ *  </ol>
+ * </p>
+ * <p>
+ *   The {@link FragmentExecutor} processes termination requests appropriately for the {@link RootExec}. A <b>Cancellation Request</b> is signalled when
+ *   {@link FragmentExecutor#cancel()} is called. A <b>Receiver Finished</b> event is signalled when {@link FragmentExecutor#receivingFragmentFinished(FragmentHandle)} is
+ *   called. The way in which these signals are handled is the following:
+ * </p>
+ * <h3>Cancellation Request</h3>
+ * <p>
+ *   There are two ways in which a cancellation request can be handled when {@link FragmentExecutor#cancel()} is called.
+ *   <ol>
+ *     <li>The Cancellation Request is recieved before the {@link RootExec} for the fragment is even started. In this case we can cleanup resources allocated for the fragment
+ *     and never start a {@link RootExec}</li>
+ *     <li>The Cancellation Request is recieve after the {@link RootExec} for the fragment is started. In this the cancellation request is sent to the
+ *     {@link FragmentEventProcessor}. If this is not the first cancellation request it is ignored. If this is the first cancellation request the {@link RootExec} for this
+ *     fragment is terminated by interrupting it. Then the {@link FragmentExecutor#run()} thread proceeds to cleanup resources normally</li>
+ *   </ol>
+ * </p>
+ * <h3>Receiver Finished</h3>
+ * <p>
+ *  When {@link FragmentExecutor#receivingFragmentFinished(FragmentHandle)} is called, the message is passed to the {@link FragmentEventProcessor} if we
+ *  did not already recieve a Cancellation request. Then the finished message is queued in {@link FragmentExecutor#receiverFinishedQueue}. The {@link FragmentExecutor#run()} polls
+ *  {@link FragmentExecutor#receiverFinishedQueue} and singlas the {@link RootExec} with {@link RootExec#receivingFragmentFinished(FragmentHandle)} appropriately.
+ * </p>
+ * <h2>Possible Design Flaws / Poorly Defined Behavoir</h2>
+ * <p>
+ *   There are still a few aspects of the {@link FragmentExecutor} design that are not clear.
+ *   <ol>
+ *     <li>If we get a <b>Receiver Finished</b> message for one downstream receiver, will we eventually get one from every downstream receiver?</li>
+ *     <li>What happens when we process a <b>Receiver Finished</b> message for some (but not all) downstream receivers and then we cancel the fragment?</li>
+ *     <li>What happens when we process a <b>Receiver Finished</b> message for some (but not all) downstream receivers and then we run out of data from the upstream?</li>
+ *   </ol>
+ * </p>
  */
 public class FragmentExecutor implements Runnable {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FragmentExecutor.class);
   private static final ControlsInjector injector = ControlsInjectorFactory.getInjector(FragmentExecutor.class);
 
-  private final AtomicBoolean hasCloseoutThread = new AtomicBoolean(false);
   private final String fragmentName;
   private final ExecutorFragmentContext fragmentContext;
   private final FragmentStatusReporter statusReporter;
@@ -66,6 +111,11 @@ public class FragmentExecutor implements Runnable {
 
   private volatile RootExec root;
   private final AtomicReference<FragmentState> fragmentState = new AtomicReference<>(FragmentState.AWAITING_ALLOCATION);
+  /**
+   * Holds all of the messages sent by downstream receivers that have finished. The {@link FragmentExecutor#run()} thread reads from this queue and passes the
+   * finished messages to the fragment's {@link RootExec} via the {@link RootExec#receivingFragmentFinished(FragmentHandle)} method.
+   */
+  private final Queue<FragmentHandle> receiverFinishedQueue = new ConcurrentLinkedQueue<>();
   private final FragmentEventProcessor eventProcessor = new FragmentEventProcessor();
 
   // Thread that is currently executing the Fragment. Value is null if the fragment hasn't started running or finished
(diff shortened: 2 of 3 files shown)


ilooner:
@sachouche @arina-ielchiieva

ilooner:
@sachouche I traced through the code. The updateAggregateStats method is called, which then calls the getOutgoingBatches method of the code generated Partitioners. That method is just a simple getter. So no one is acquiring the same lock. But even if someone else was, the code in the close method is single threaded, and synchronize blocks are reentrant.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
```diff
@@ -348,9 +348,12 @@ public void close() throws Exception {
     logger.debug("Partition sender stopping.");
     super.close();
     ok = false;
-    if (partitioner != null) {
-      updateAggregateStats();
-      partitioner.clear();
+
+    synchronized (this) {
```
arina-ielchiieva: 1. Should partitioner be volatile?
2. Should we check if partitioner is not null before synchronization as well (DCL)?

ilooner:
@priteshm @arina-ielchiieva I should have updated this PR earlier this week, here is my update. After reflecting on Arina's comments and reading some more docs about how java implements volatile and synchronization, I think this solution might not fix the original race condition. I need to to more reading to get a better understanding. Additionally I realized there is another race condition where two threads are simultaneously calling close and innerNext which could cause a memory leak. Haven't had a chance to dig further this week, so I will try to wrap this up next week.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
```diff
@@ -62,7 +62,7 @@
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PartitionSenderRootExec.class);
   private RecordBatch incoming;
   private HashPartitionSender operator;
-  private PartitionerDecorator partitioner;
+  private volatile PartitionerDecorator partitioner;
```
vrozov: Consider using `AtomicReference` instead of `volatile`

ilooner:
@arina-ielchiieva @vrozov I believe I have a solution. There were several issues with the original code.

1. It made incorrect assumptions about how cache invalidation works with java **synchronized**.
2. It assumed **innerNext** and **close** would be called sequentially.

I believe this PR fixes these issues now and I have gone into more detail about the problems below.

# 1. Incorrect Cache Invalidation Assumptions

The original code was trying to be smart by trying to reduce synchronization overhead on **innerNext**. So the code in **innerNext** did not synchronize before changing the partitioner object since this would be called often. The code in **close()** and ** receivingFragmentFinished()** synchronized before accessing the partitioner with the intention that this would trigger an update of the partitioner variable state across all threads. Unfortunately, this assumption is invalid (see https://stackoverflow.com/questions/22706739/does-synchronized-guarantee-a-thread-will-see-the-latest-value-of-a-non-volatile). Every thread that accesses a variable must synchronize before accessing a variable in order to properly invalidate cached data on a core. 

For example if **Thread A** modifies **Variable 1** then **Thread B** synchronizes before accessing **Variable 1** then there is no guarantee **Thread B** will see the most updated value for **Variable 1** since it might .

## Solution

In summary the right thing to do is the simple thing. Make the methods synchronized. Unfortunately there is no way to outsmart the system and reduce synchronization overhead without causing race conditions.

# 2. Concurrent InnerNext and Close Calls

The original code did not consider the case that innerNext was in the middle of execution when close was called. It did try to handle the case where **innerNext** could be called after **close** by setting the **ok** variable. But it didn't even do that right because there was no synchronization around the **ok** variable.

## Solution

The right thing to do is the simple thing. Make sure the methods are synchronized so close has to wait until innerNext is done before executing. Also when a query is cancelled the executing thread should be interrupted the thread running innerNext incase it is on a blocking call.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
```diff
@@ -231,7 +231,7 @@ public boolean innerNext() {
   }
 
   @VisibleForTesting
-  protected void createPartitioner() throws SchemaChangeException {
+  protected synchronized void createPartitioner() throws SchemaChangeException {
```
arina-ielchiieva: Why you need to synchronized`createParitioner`? It's being called from `innerNext` which is already synchronized.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -245,6 +244,8 @@ public Void run() throws Exception {
         // we have a heap out of memory error. The JVM in unstable, exit.
         CatastrophicFailure.exit(e, "Unable to handle out of memory condition in FragmentExecutor.", -2);
       }
+    } catch (InterruptedException e) {
+      // Swallow interrupted exceptions since we intentionally interrupt the root when cancelling a query
```
arina-ielchiieva: Maybe add logger.trace?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
```diff
@@ -358,7 +357,7 @@ public void close() throws Exception {
     }
   }
 
-  public void sendEmptyBatch(boolean isLast) {
+  public synchronized void sendEmptyBatch(boolean isLast) {
```
arina-ielchiieva: Method can be made private. Plus maybe not be synchronized as well since is being called from `innerNext`.

vrozov:
@ilooner What state is protected by `syncrhonized`? Why is it not sufficient to use `volatile` and `AtomicReference`?

ilooner:
@vrozov The main issue is that InnerNext and Close should not execute concurrently. Even when we are using AtomicReferences and volatile the following sequence of events could happen which could cause a memory leak:

  1. Let's say there are two Threads. The **Close thread** which starts at the beginning of the close method, and the **Next thread** which starts at the beginning of the innerNext method.
  1. Now let's say the **Next Thread** runs and checks **ok**. Since close has not been called yet **ok** is true.
  1. Now the **Next Thread** is after the ok check.
  1. The **Close thread** now starta executing. And the close thread clears the partitioner.
  1. Now after the partitioner is cleared the **Next Thread** can resume executing. If the next thread receives an OK_SCHEMA he will allocate a new partitioner. Since the OK_SCHEMA message may include records the partitioner may partition some data as well.
  1. Now the **Close thread** is done, but there is a partitioner that has not been closed, and we will leak memory.

In order to property solve this problem we need to make sure that the innerNext and close methods are mutually exclusive so the above scenario can never happen. The easiest way to do that is to use the synchronized key word. If we use the synchronized keyword then we don't have to use volatile or atomic reference.

Also as a side note using synchronized will probably be more efficient since a cache flush would only be triggered at the start of the innerNext and close method. Alternatively if we used volatile and AtomicReference a cache flush would be triggered every time we accessed the ok and partitioner variables.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
ilooner: Removed

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionSenderRootExec.java
ilooner: Done

ilooner:
@arina-ielchiieva Applied review comments.

arina-ielchiieva:
Looks good, @vrozov are you ok with the changes?

ilooner:
@vrozov Thanks for catching this, I believe you are right. hasClouseoutThread guarantees innerNext and close won't be called concurrently. However, I still believe innerNext and receivingFragmentFinished could be called concurrently, since the ControlMessageHandler thread executes recievingFragmentFinished. Additionally in rare cases where a limit query is cancelled recievingFragmentFinished and close could be called concurrently as well. While reflecting on your comments I also saw another issue where the root could be blocked on a next call but a Finished event would not cause root to terminate.

In light of all of this I actually think the **synchronized** is not sufficient. We will have to have some way to interrupt the execution of the root when we received a finish signal and only close out the resources after receivingFragmentFinished has been called. Similarly if we receive a finish signal we should ignore any cancellation requests instead of trying to cancel and finish simultaneously and vice versa.

I will rework the solution to address these issues and let you know when I have an update.

ilooner:
@arina-ielchiieva @vrozov In light of Vlad's comments I have reworked the synchronization model yet again. This change now removes all synchronization from PartitionSenderRootExec and enforces the guarantee that all the lifecycle methods of the PartitionSenderRootExec will only be called by a single Run thread in the FragmentExecutor. Also while making this change I discovered a few other bugs with how cancellations and receiver finishes are handled, so I have addressed those bugs as well. I will go into more detail about what I changed below.

# Motivation

As Vlad pointed out **close** and **innerNext** are never called concurrently. After closer inspection of the code I also released that currently (in apache master) innerNext and close will always be called by the **FragmentExecutor#run** thread. The only method of PartitionSenderRootExec that is not called by the **FragmentExecutor#run** thread is **receivingFragmentFinished**. In order to simplify the implementation of PartitionSenderRootExec and also simplify the design of the FragmentExecutor I changed the code so that only the **FragmentExecutor#run** thread calls **receivingFragmentFinished** as well. In this way we can remove all the synchronization from PartitionSenderRootExec. This was done by by:
 1. Making the event processor save the FragmentHandle in the event that a receiver finish request was sent. 
 2. After the **root.next()** loop terminates in **FragmentExecutor#run** the eventProcessor is checked to see if a finish request was received. If so **receivingFragmentFinished** is called on root by the **FragmentExecutor#run** method.

# Other Bugs

## Processing of multiple termination requests

The event processor would process all cancellation and finish requests, even if there is more than one. This doesn't really make sense, since we can only cancel or finish once. So I have changed the event processor to execute only the first termination request and ignore all the others.

## Simultaneous Cancellation and Finishing

Since the event processor would process multiple termination requests concurrently it was possible for a cancel and a finish message to be received and processed simultaneously. The results of this were not well defined since **close** and **receivingFragmentFinished** could be called concurrently.

# Other Improvements

Vlad also pointed out that we did not need the hasCloseoutThread atomic reference, since we were already using the myThreadRef atomic reference. That cleaned up the code a bit.

ilooner:
Nope never mind I think I spoke too soon. I just realized we may get multiple receivingFragmentFinished requests, one for each downstream receiver. Back to the drawing board.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -67,6 +88,11 @@
 
   private volatile RootExec root;
   private final AtomicReference<FragmentState> fragmentState = new AtomicReference<>(FragmentState.AWAITING_ALLOCATION);
+  /**
+   * Holds all of the messages sent by downstream recievers that have finished. The {@link FragmentExecutor#run()} thread reads from this queue and passes the
+   * finished messages to the fragment's {@link RootExec} via the {@link RootExec#receivingFragmentFinished(FragmentHandle)} method.
+   */
+  private final Queue<FragmentHandle> recieverFinishedQueue = new ConcurrentLinkedQueue<>();
```
vrozov: please fix typo (receiver)

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -231,9 +261,9 @@ public Void run() throws Exception {
           while (shouldContinue()) {
             // Fragment is not cancelled
 
-            if (eventProcessor.hasFinishedRequests()) {
+            if (!recieverFinishedQueue.isEmpty()) {
```
vrozov: I don't think `isEmpty()` is necessary, it is the same as `poll() != null`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -488,47 +527,66 @@ void receiverFinished(FragmentHandle handle) {
...
+     * Tell the {@link FragmentEventProcessor} not to process anymore events. This keeps stray cancellation requests
+     * from being processed after the root has finished running and interrupts in the root thread have been cleared.
+     */
+    public synchronized void terminate() {
+      terminate.set(true);
+    }
+
     @Override
     protected void processEvent(FragmentEvent event) {
+      if (terminate.get()) {
```
vrozov: use atomic compare and set to avoid race condition.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -475,6 +483,8 @@ public void drillbitUnregistered(final Set<CoordinationProtos.DrillbitEndpoint>
    * This is especially important as fragments can take longer to start
    */
   private class FragmentEventProcessor extends EventProcessor<FragmentEvent> {
+    private boolean terminate = false;
+    private List<FragmentHandle> finishedHandles = Lists.newArrayList();
```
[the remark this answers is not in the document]
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -475,6 +483,8 @@ public void drillbitUnregistered(final Set<CoordinationProtos.DrillbitEndpoint>
    * This is especially important as fragments can take longer to start
    */
   private class FragmentEventProcessor extends EventProcessor<FragmentEvent> {
+    private boolean terminate = false;
```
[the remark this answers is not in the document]
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: Thanks for catching. Fixed all the misspellings of receiver.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: Thanks for catching. Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: Thanks for catching removed.

ilooner:
@vrozov @arina-ielchiieva Applied review comments, please let me know if there are anymore comments.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -245,19 +306,17 @@ public Void run() throws Exception {
...
-      // interruption after we have moved beyond this block.
-      synchronized (myThreadRef) {
-        myThreadRef.set(null);
-        Thread.interrupted();
-      }
-
-      // Make sure the event processor is started at least once
-      eventProcessor.start();
+      // Don't process any more termination requests, we are done.
+      eventProcessor.terminate();
```
vrozov: Is this terminate() required?

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -227,11 +280,19 @@ public void run() {
...
-          /*
-           * Run the query until root.next returns false OR we no longer need to continue.
-           */
-          while (shouldContinue() && root.next()) {
-            // loop
+
+          while (shouldContinue()) {
+            // Fragment is not cancelled
+
+            for (FragmentHandle fragmentHandle; (fragmentHandle = receiverFinishedQueue.poll()) != null;) {
```
vrozov: How this processing affects `root.next()`? Should it be also/only done after fragment is canceled or done (after while loop)?

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -488,47 +548,74 @@ void receiverFinished(FragmentHandle handle) {
...
-          synchronized (myThreadRef) {
-            final Thread myThread = myThreadRef.get();
-            if (myThread != null) {
-              logger.debug("Interrupting fragment thread {}", myThread.getName());
-              myThread.interrupt();
-            }
-          }
+          // The root was started so we have to interrupt it in case it is performing a blocking operation.
+          killThread();
+          terminate.set(true);
```
vrozov: it is not necessary (terminate should be already true).

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
```diff
@@ -488,47 +548,74 @@ void receiverFinished(FragmentHandle handle) {
...
           break;
-
         case CANCEL_AND_FINISH:
+          // In this case the root was never started so we do not have to interrupt the thread.
           updateState(FragmentState.CANCELLATION_REQUESTED);
+          // The FragmentExecutor#run() loop will not execute in this case so we have to cleanup resources here
           cleanup(FragmentState.FINISHED);
+          terminate.set(true);
```
vrozov: it is not necessary (terminate should be already true).

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: Doh! Thanks for catching

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: Thanks for catching

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: The original code allowed PartitionSenderRootExec#receivingFragmentFinished() to be called before, after or concurrently with PartitionSenderRootExec#innerNext(). The idea was that PartitionSenderRootExec#receivingFragmentFinished would terminate a partition. The terminate method of OutgoingRecordBatch would cause all outgoing records to a finished receiver to be dropped. So the finished downstream receivers would stop receiving data from the PartitionSenderRootExec, but the downstream receivers that did not finish would continue to receive records. There is a gray area here because I'm not sure why only some of the downstream receivers would send a RECEIVER_FINISHED message and others wouldn't, but the original design seems to make an assumption that this is a very common thing and is optimized for it. So I assume the original authors know something that we don't with respect to that.

Also calling receivingFragmentFinished after we finished processing all the data would defeat the purpose, since the intention was to allow our fragment to terminate early if the downstream receivers decided they don't need anymore data (ex. A select limit query which only asks the first 100 rows of a result with 10 million rows). If we called it only after the next() loop was done we would always process all upstream records even when we didn't have to and we would see significant degradation in the performance of limit queries.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
ilooner: There is a corner case. If we didn't include eventProcessor.terminate() we could theoretically receive a cancellation request for the first time after the interrupts were cleared for the FragmentExecutor#run thread. The cancellation would then interrupt the Thread again, and our FragmentExecutor would finish and leave the thread it used in an interrupted state. This could cause problems for the next FragmentExecutor that uses the same thread.

exec/java-exec/src/main/java/org/apache/drill/exec/work/fragment/FragmentExecutor.java
vrozov: OK, I see.

ilooner:
Squashed commits. @arina-ielchiieva please let me know if you have any comments.

vrozov:
@arina-ielchiieva Please review.

arina-ielchiieva:
+1

PR #1107, opened 2018-02-01
DRILL-6123: Limit batch size for Merge Join based on memory

ppadma:
Merge join limits output batch size to 32K rows irrespective of row size. This can create large batches (in terms of memory), depending upon average row width. Changed the logic to figure out output row count based on memory specified with the new outputBatchSize option and average outgoing row width. Average outgoing row width will be sum of  left and right batch row widths. Output row count will be minimum of 1 and max of 64k. 
Added AbstractRecordBatchMemoryManager class to be used across all operators. Restructured the code a little bit for that.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -77,7 +77,9 @@ public final class ExecConstants {
   public static final String SPILL_DIRS = "drill.exec.spill.directories";
 
   public static final String OUTPUT_BATCH_SIZE = "drill.exec.memory.operator.output_batch_size";
-  public static final LongValidator OUTPUT_BATCH_SIZE_VALIDATOR = new RangeLongValidator(OUTPUT_BATCH_SIZE, 1024, 512 * 1024 * 1024);
+  // Output Batch Size in Bytes. We have a small lower bound so we can test with unit tests without the
+  // need to produce very large batches that take up lot of memory.
+  public static final LongValidator OUTPUT_BATCH_SIZE_VALIDATOR = new RangeLongValidator(OUTPUT_BATCH_SIZE, 128, 512 * 1024 * 1024);
 
   // External Sort Boot configuration
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinStatus.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinStatus.java
@@ -29,8 +29,6 @@ import org.apache.calcite.rel.core.JoinRelType;
 public final class JoinStatus {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(JoinStatus.class);
 
-  private static final int OUTPUT_BATCH_SIZE = 32*1024;
-
   public final RecordIterator left;
   public final RecordIterator right;
   private boolean iteratorInitialized;
@@ -43,6 +41,8 @@ public final class JoinStatus {
 
   public boolean ok = true;
   public boolean hasMoreData = false;
+
+  private int targetOutputRowCount;
 
   public JoinStatus(RecordIterator left, RecordIterator right, MergeJoinBatch output) {
     this.left = left;
@@ -101,8 +101,12 @@ public final class JoinStatus {
   }
 
   public final boolean isOutgoingBatchFull() {
-    Preconditions.checkArgument(outputPosition <= OUTPUT_BATCH_SIZE);
-    return outputPosition == OUTPUT_BATCH_SIZE;
+    Preconditions.checkArgument(outputPosition <= targetOutputRowCount);
+    return outputPosition >= targetOutputRowCount;
+  }
+
+  public final void setTargetOutputRowCount(int outputRowCount) {
+    this.targetOutputRowCount = outputRowCount;
   }
 
   public final void incOutputPos() {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
@@ -33,6 +33,7 @@ import org.apache.drill.common.types.TypeProtos.DataMode;
 import org.apache.drill.common.types.TypeProtos.MajorType;
 import org.apache.drill.common.types.TypeProtos.MinorType;
 import org.apache.drill.common.types.Types;
+import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.compile.sig.MappingSet;
 import org.apache.drill.exec.exception.ClassTransformationException;
 import org.apache.drill.exec.exception.OutOfMemoryException;
@@ -45,6 +46,7 @@ import org.apache.drill.exec.expr.fn.FunctionGenerationHelper;
 import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.physical.config.MergeJoinPOP;
 import org.apache.drill.exec.physical.impl.common.Comparator;
+import org.apache.drill.exec.record.RecordBatchSizer;
 import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.RecordBatch;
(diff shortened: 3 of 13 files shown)


ppadma:
@sachouche @ilooner @paul-rogers Can one of you review this PR for me ?

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -77,7 +77,7 @@ private ExecConstants() {
   public static final String SPILL_DIRS = "drill.exec.spill.directories";
 
   public static final String OUTPUT_BATCH_SIZE = "drill.exec.memory.operator.output_batch_size";
-  public static final LongValidator OUTPUT_BATCH_SIZE_VALIDATOR = new RangeLongValidator(OUTPUT_BATCH_SIZE, 1024, 512 * 1024 * 1024);
+  public static final LongValidator OUTPUT_BATCH_SIZE_VALIDATOR = new RangeLongValidator(OUTPUT_BATCH_SIZE, 1, 512 * 1024 * 1024);
```
paul-rogers: Maybe add a comment to explain the units here. Bytes? MB? A minimum batch size of 1 byte seems small, but a max size of 512 GB seems large, so not sure of the limits...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinStatus.java
```diff
@@ -101,8 +101,12 @@ public final void resetOutputPos() {
   }
 
   public final boolean isOutgoingBatchFull() {
-    Preconditions.checkArgument(outputPosition <= OUTPUT_BATCH_SIZE);
-    return outputPosition == OUTPUT_BATCH_SIZE;
+    Preconditions.checkArgument(outputPosition <= outputRowCount);
+    return outputPosition == outputRowCount;
```
paul-rogers: Maybe be just a bit more paranoid? `outputPosition >= outputRowCount`?

And, while we're at it, maybe `outputRowCount` -> `targetOutputRowCount`? To make clear that the value is our target, not the actual, current row count.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
```diff
@@ -102,20 +105,78 @@
...
+
+      // If outgoing row width is 0, just return. This is possible for empty batches or
+      // when first set of batches come with OK_NEW_SCHEMA and no data.
+      if (newOutgoingRowWidth == 0) {
+        return;
+      }
+
+      // update the value to be used for next batch(es)
+      setOutputRowCount(Math.min(ValueVector.MAX_ROW_COUNT,
+        Math.max(RecordBatchSizer.safeDivide(outputBatchSize/WORST_CASE_FRAGMENTATION_FACTOR, newOutgoingRowWidth), MIN_NUM_ROWS)));
```
paul-rogers: Maybe wrap this in a method since it is used multiple times.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
```diff
@@ -102,20 +105,78 @@
...
+      // calculate memory used so far based on previous outgoing row width and how many rows we already processed.
+      final long memoryUsed = status.getOutPosition() * getOutgoingRowWidth();
+      // This is the remaining memory.
+      final long remainingMemory = Math.max(outputBatchSize/WORST_CASE_FRAGMENTATION_FACTOR - memoryUsed, 0);
+      // These are number of rows we can fit in remaining memory based on new outgoing row width.
+      final int numOutputRowsRemaining = RecordBatchSizer.safeDivide(remainingMemory, newOutgoingRowWidth);
+
+      final int adjustedOutputRowCount = Math.min(MAX_NUM_ROWS, Math.max(status.getOutPosition() + numOutputRowsRemaining, MIN_NUM_ROWS));
+      status.setOutputRowCount(adjustedOutputRowCount);
+      setOutgoingRowWidth(newOutgoingRowWidth);
```
paul-rogers: This number is valid only for this one batch. The next batch doesn't have the "legacy" rows. Do we recompute the number at the start of the next output batch?

ppadma: Yes, this is how it works.
We read from left and right side using RecordIterator abstraction, which reads full record batches underneath and gives one record at a time with its next call.  I have a callback mechanism when we read a new batch in record iterator to adjust the row widths. When I get the callback, I adjust row count for the current batch we are working on based on remaining memory available for the current batch and also compute and save the row count I should use for next full batch. In the innerNext, when we start working on the next output batch, I am setting the target output row count based on this value.

Addressed all other code review comments. Please take a look when you get a chance.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
sachouche: Theoretically, 2Gb-1 should be the largest value as our value vectors use integers as offset values.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
sachouche: Our goal is to eventually use Paul's framework for controlling batch sizes. I feel the BatchRecordMemoryManager terminology is a bit ambitious (kind of misleading).

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractRecordBatchMemoryManager.java
```diff
@@ -0,0 +1,63 @@
...
+    return outputRowCount;
+  }
+
+  /**
+   * Given batchSize and rowWidth, this will set output rowCount taking into account
+   * the min and max that is allowed.
+   */
+  public void setOutputRowCount(long batchSize, int rowWidth) {
+    this.outputRowCount = Math.min(ValueVector.MAX_ROW_COUNT,
+      Math.max(RecordBatchSizer.safeDivide(batchSize/WORST_CASE_FRAGMENTATION_FACTOR, rowWidth), MIN_NUM_ROWS));
```
paul-rogers: I suspect the math may be off here. `batchSize` is the actual memory for the batch? If so, then it already includes actual internal fragmentation which, depending on the source, can be 25% (for "good" operators) to 50% (if from the network) to 95% (pathological Parquet cases.) Using the fragmentation factor on this value gets us even more lost.

A better way is to do:

```
rawBatchSize = rowWidth * rowCount;
actualBatchSize = rowBatchSize * FRAG_FACTOR;
```

Or to compute the output size:

```
targetOutputSize = // some value you set
targetDataSize = targetOutputSize / FRAG_FACTOR
targetRowCount = targetDataSize / rowWidth
```

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractRecordBatchMemoryManager.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * This will set rowCount taking into account the min and max that is allowed.
+   */
+  public int setOutputRowCount(int rowCount) {
```
paul-rogers: The `set` name is a misnomer. We are not setting anything. We are computing. And, since we don't reference fields, and don't change state, this can be `static`.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractRecordBatchMemoryManager.java
paul-rogers: Should this method use the one below to set the limits?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/spill/RecordBatchSizer.java
```diff
@@ -311,8 +311,8 @@ public static ColumnSize getColumn(ValueVector v, String prefix) {
 
   public RecordBatchSizer(RecordBatch batch) {
     this(batch,
-         (batch.getSchema().getSelectionVectorMode() == BatchSchema.SelectionVectorMode.TWO_BYTE) ?
-         batch.getSelectionVector2() : null);
+      (batch.getSchema() == null ? null : (batch.getSchema().getSelectionVectorMode() == BatchSchema.SelectionVectorMode.TWO_BYTE ?
```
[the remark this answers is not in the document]
ppadma: yes, we can get empty batches with empty schema  and I think it makes sense to add the check here instead of calling code. That way, it be transparently handled underneath.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractRecordBatchMemoryManager.java
ppadma: batchSize here is the configured outputBatchSize that we should conform to, not the incoming batch size.

ppadma:
@paul-rogers Thank you very much for the review. Updated the PR with review comments taken care of. Please take a look.

PR #1109, opened 2018-02-03
DRILL-6128: Wrong Result with Nested Loop Join

sohami:


sohami:
@amansinha100 - Please help to review.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/NestedLoopJoinTemplate.java
```diff
@@ -117,9 +117,12 @@ private int populateOutgoingBatch(JoinRelType joinType, int outputIndex) {
       for (; nextRightBatchToProcess < rightCounts.size(); nextRightBatchToProcess++) {
         int rightRecordCount = rightCounts.get(nextRightBatchToProcess);
         // for every record in right batch
+        final int currentRightBatchIndex = nextRightBatchToProcess << 16;
         for (; nextRightRecordToProcess < rightRecordCount; nextRightRecordToProcess++) {
 
-          if (doEval(nextLeftRecordToProcess, nextRightBatchToProcess, nextRightRecordToProcess)) {
+          // Since right container is a hyper container, in doEval generated code it expects the
```
amansinha100: Minor: this comment ideally could be moved up to the statement where you do the left shift since that's the main change.

amansinha100:
+1 .  Thanks for adding the unit tests.

sohami:
Updated the comment and rebased on top of latest master

PR #1111, opened 2018-02-05
DRILL-5978: Updating of Apache and MapR Hive libraries to 2.3.2 and 2.1.2-mapr-1710 versions respectively

vdiravka:
Updating hive properties for tests and resolving dependencies and API conflicts:

* Allowing of using Hive's own calcite-core and avatica versions by hive-exec.
Calcite version is removed from root POM Dependency Management
* Fix for "hive.metastore.schema.verification", MetaException(message:
Version information not found in metastore)
https://cwiki.apache.org/confluence/display/Hive/Hive+Schema+Tool
METASTORE_SCHEMA_VERIFICATION="false" property is added
* Fix JSONException class is not found (excluded banned org.json dependency)
* Added METASTORE_AUTO_CREATE_ALL="true", properties to tests, because some additional
tables are necessary in Hive metastore
* Disabling calcite CBO for (Hive's CalcitePlanner) for tests, because it is in conflict
with Drill's Calcite version for Drill unit tests. HIVE_CBO_ENABLED="false" property
* jackson and parquet libraries are relocated in hive-exec-shade module
* org.apache.parquet:parquet-column Drill version is added to "hive-exec" to
allow of using Parquet empty group on MessageType level (PARQUET-278)
* Removing of commons-codec exclusion from hive core. This dependency is
necessary for hive-exec and hive-metastore.
* Setting Hive's internal properties for transactional scan:
HiveConf.HIVE_TRANSACTIONAL_TABLE_SCAN and for schema evolution: HiveConf.HIVE_SCHEMA_EVOLUTION,
IOConstants.SCHEMA_EVOLUTION_COLUMNS, IOConstants.SCHEMA_EVOLUTION_COLUMNS_TYPES

### DIFF
--- a/contrib/storage-hive/core/pom.xml
+++ b/contrib/storage-hive/core/pom.xml
@@ -63,12 +63,6 @@
     <dependency>
       <groupId>org.apache.hive</groupId>
       <artifactId>hive-hbase-handler</artifactId>
-      <exclusions>
-        <exclusion>
-          <groupId>org.apache.hive</groupId>
-          <artifactId>hive-exec</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.hbase</groupId>
--- a/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveMetadataProvider.java
+++ b/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveMetadataProvider.java
@@ -264,6 +264,7 @@ public class HiveMetadataProvider {
           final List<LogicalInputSplit> splits = Lists.newArrayList();
           final JobConf job = new JobConf(hiveConf);
           HiveUtilities.addConfToJob(job, properties);
+          HiveUtilities.verifyAndAddTransactionalProperties(job, sd);
           job.setInputFormat(HiveUtilities.getInputFormatClass(job, sd, hiveReadEntry.getTable()));
           final Path path = new Path(sd.getLocation());
           final FileSystem fs = path.getFileSystem(job);
--- a/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
+++ b/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
@@ -17,8 +17,11 @@
  */
 package org.apache.drill.exec.store.hive;
 
+import com.google.common.base.Function;
+import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import com.google.common.collect.Lists;
 import io.netty.buffer.DrillBuf;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.exceptions.ExecutionSetupException;
@@ -51,10 +54,14 @@ import org.apache.drill.exec.vector.ValueVector;
 import org.apache.drill.exec.work.ExecErrorConstants;
 
 import org.apache.hadoop.hive.common.type.HiveDecimal;
+import org.apache.hadoop.hive.metastore.api.FieldSchema;
 import org.apache.hadoop.hive.metastore.MetaStoreUtils;
 import org.apache.hadoop.hive.metastore.api.Partition;
 import org.apache.hadoop.hive.metastore.api.StorageDescriptor;
 import org.apache.hadoop.hive.metastore.api.Table;
+import org.apache.hadoop.hive.ql.exec.Utilities;
+import org.apache.hadoop.hive.ql.io.AcidUtils;
+import org.apache.hadoop.hive.ql.io.IOConstants;
 import org.apache.hadoop.hive.ql.metadata.HiveStorageHandler;
 import org.apache.hadoop.hive.ql.metadata.HiveUtils;
 import org.apache.hadoop.hive.serde.serdeConstants;
@@ -70,6 +77,7 @@ import org.apache.hadoop.mapred.JobConf;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 
+import javax.annotation.Nullable;
 import java.math.BigDecimal;
 import java.sql.Date;
 import java.sql.Timestamp;
@@ -104,8 +112,7 @@ public class HiveUtilities {
           return Boolean.parseBoolean(value);
         case DECIMAL: {
           DecimalTypeInfo decimalTypeInfo = (DecimalTypeInfo) typeInfo;
-          return HiveDecimalUtils.enforcePrecisionScale(HiveDecimal.create(value),
-              decimalTypeInfo.precision(), decimalTypeInfo.scale());
+          return HiveDecimalUtils.enforcePrecisionScale(HiveDecimal.create(value), decimalTypeInfo);
         }
         case DOUBLE:
           return Double.parseDouble(value);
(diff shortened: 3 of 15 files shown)


priteshm:
@vrozov can you please review this change?

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -507,5 +509,51 @@ public static boolean hasHeaderOrFooter(HiveTableWithColumnCache table) {
...
+          isFirst = false;
+        } else {
+          colNamesBuilder.append(',');
+          colTypesBuilder.append(',');
+        }
+        colNamesBuilder.append(col.getName());
+        colTypesBuilder.append(col.getType());
+      }
+      colNames = colNamesBuilder.toString();
+      colTypes = colTypesBuilder.toString();
```
sohami: how about changing the loop as below:

```
final StringBuilder colNamesBuilder = new StringBuilder();
final StringBuilder colTypesBuilder = new StringBuilder();

     for(FieldSchema col: sd.getCols()) {
           colNamesBuilder.append(col.getName());
           colTypesBuilder.append(col.getType());
           colNamesBuilder.append(',');
           colTypesBuilder.append(',');
      }
      colNames = colNamesBuilder.substring(0, colNamesBuilder.length() - 1);
      colTypes = colTypesBuilder.substring(0, colTypesBuilder.length() - 1);
```

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveMetadataProvider.java
```diff
@@ -264,6 +265,10 @@ private HiveStats getStatsEstimateFromInputSplits(final List<LogicalInputSplit>
           final List<LogicalInputSplit> splits = Lists.newArrayList();
           final JobConf job = new JobConf(hiveConf);
           HiveUtilities.addConfToJob(job, properties);
+          if (AcidUtils.isTablePropertyTransactional(properties)) {
+            AcidUtils.setTransactionalTableScan(job, true);
+            HiveUtilities.setColumnTypes(job, properties, true, sd);
+          }
```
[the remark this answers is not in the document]
vdiravka: It makes sense. Moreover since schema_evolution is required for acid tables [HIVE-12799](https://issues.apache.org/jira/browse/HIVE-12799) I've combined it with setColumnTypes() helper method.

common/pom.xml
```diff
@@ -45,6 +45,7 @@
     <dependency>
       <groupId>org.apache.calcite</groupId>
       <artifactId>calcite-core</artifactId>
+      <version>${calcite.version}</version>
```
vrozov: Why is this change necessary? The version should come from `<dependencyManagement>` of the parent pom.

contrib/storage-hive/core/pom.xml
```diff
@@ -101,6 +105,7 @@
     <dependency>
       <groupId>org.apache.calcite</groupId>
       <artifactId>calcite-core</artifactId>
+      <version>${calcite.version}</version>
```
vrozov: The same question as for common/pom.xml.

contrib/storage-hive/core/pom.xml
```diff
@@ -58,6 +58,10 @@
           <groupId>commons-codec</groupId>
           <artifactId>commons-codec</artifactId>
         </exclusion>
+        <exclusion>
```
vrozov: Is the exclusion necessary due to a version conflict or the dependency is not required?

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -507,5 +510,52 @@ public static boolean hasHeaderOrFooter(HiveTableWithColumnCache table) {
...
+  /**
+   * This method checks whether the table is transactional and set necessary properties in {@link JobConf}.
+   * If schema evolution properties aren't set in job conf for the input format, method sets the column names
+   * and types from table/partition properties or storage descriptor.
+   *
+   * @param job the job to update
+   * @param properties table or partition properties
+   * @param sd storage descriptor
+   */
+  public static void verifyAndAddTransactionalProperties(JobConf job, Properties properties, StorageDescriptor sd) {
```
vrozov: Is it necessary to pass both `JobConf` and `Properties`? As far as I can see `job` is always populated using passed `properties`.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -507,5 +510,52 @@ public static boolean hasHeaderOrFooter(HiveTableWithColumnCache table) {
...
+      String colNames;
+      String colTypes;
+
+      // Try to get get column names and types from table or partition properties. If they are absent there, get columns
+      // data from storage descriptor of the table
+      if (properties.containsKey(serdeConstants.LIST_COLUMNS) && properties.containsKey(serdeConstants.LIST_COLUMN_TYPES)) {
+        colNames = job.get(serdeConstants.LIST_COLUMNS);
+        colTypes = job.get(serdeConstants.LIST_COLUMN_TYPES);
+      } else {
+        final StringBuilder colNamesBuilder = new StringBuilder();
```
vrozov: consider using `Joiner`.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -507,5 +510,52 @@ public static boolean hasHeaderOrFooter(HiveTableWithColumnCache table) {
...
+          job.get(IOConstants.SCHEMA_EVOLUTION_COLUMNS_TYPES) != null) {
+        return;
+      }
+
+      String colNames;
+      String colTypes;
+
+      // Try to get get column names and types from table or partition properties. If they are absent there, get columns
+      // data from storage descriptor of the table
+      if (properties.containsKey(serdeConstants.LIST_COLUMNS) && properties.containsKey(serdeConstants.LIST_COLUMN_TYPES)) {
```
vrozov: avoid double `get()` (`containsKey()`) if possible.

contrib/storage-hive/hive-exec-shade/pom.xml
```diff
@@ -39,23 +39,28 @@
           <groupId>log4j</groupId>
         </exclusion>
         <exclusion>
-          <groupId>commons-codec</groupId>
-          <artifactId>commons-codec</artifactId>
-        </exclusion>
-        <exclusion>
-          <artifactId>calcite-avatica</artifactId>
-          <groupId>org.apache.calcite</groupId>
+          <groupId>org.json</groupId>
```
vrozov: Has new version of hive introduced the dependency on org.json:json?

contrib/storage-hive/hive-exec-shade/pom.xml
```diff
@@ -39,23 +39,28 @@
...
+          <artifactId>json</artifactId>
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-column</artifactId>
+      <version>${parquet.version}</version>
```
vrozov: Any other parquet dependencies? If they are not needed, why the explicit dependency on org.apache.parquet:parquet-column is necessary?

contrib/storage-hive/hive-exec-shade/pom.xml
```diff
@@ -39,23 +39,28 @@
...
+    </dependency>
   </dependencies>
 
   <build>
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
-        <version>2.1</version>
+        <version>3.1.0</version>
```
vrozov: What is the reason for the version change and should it be applied to other modules where shading is used?

common/pom.xml
vrozov: Will it be better to explicitly specify the version of calcite-core in hive-exec?

contrib/storage-hive/core/pom.xml
vrozov: For a transitive dependency will it be better to use `DependencyManagement`? Please add comments with details.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -507,5 +512,49 @@ public static boolean hasHeaderOrFooter(HiveTableWithColumnCache table) {
...
+      colTypes = job.get(serdeConstants.LIST_COLUMN_TYPES);
+
+      if (colNames == null || colTypes == null) {
+        List<String> colNamesList = Lists.newArrayList();
+        List<String> colTypesList = Lists.newArrayList();
+        for (FieldSchema col: sd.getCols()) {
+          colNamesList.add(col.getName());
+          colTypesList.add(col.getType());
+        }
+        colNames = Joiner.on(",").join(colNamesList);
```
vrozov: Consider `Joiner.on(",").join(Iterables.transform(sd.getCols(), toName));` where `toName` is 
```
  private static Function<FieldSchema, String> toName = new Function<FieldSchema, String>()
  {
    @Nullable
    @Override
    public String apply(@Nullable FieldSchema input)
    {
      return input.getName();
    }
  };
```

contrib/storage-hive/hive-exec-shade/pom.xml
vrozov: Why is it safe to change one library? Will it be safer to upgrade all hive dependencies on parquet to the same version?

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/readers/HiveAbstractReader.java
```diff
@@ -86,7 +86,7 @@
   protected List<Object> selectedPartitionValues = Lists.newArrayList();
 
   // SerDe of the reading partition (or table if the table is non-partitioned)
-  protected SerDe partitionSerDe;
+  protected Deserializer partitionSerDe;
```
vrozov: rename to 'partitionDeserializer'

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/readers/HiveAbstractReader.java
```diff
@@ -143,7 +143,7 @@ private void init() throws ExecutionSetupException {
               HiveUtilities.getPartitionMetadata(partition, table);
       HiveUtilities.addConfToJob(job, partitionProperties);
 
-      final SerDe tableSerDe = createSerDe(job, table.getSd().getSerdeInfo().getSerializationLib(), tableProperties);
+      final Deserializer tableSerDe = createSerDe(job, table.getSd().getSerdeInfo().getSerializationLib(), tableProperties);
```
vrozov: rename both function and the variable.

contrib/storage-hive/hive-exec-shade/pom.xml
```diff
@@ -34,28 +34,47 @@
...
+          <artifactId>calcite-linq4j</artifactId>
         </exclusion>
         <exclusion>
-          <artifactId>calcite-avatica</artifactId>
           <groupId>org.apache.calcite</groupId>
+          <artifactId>calcite-druid</artifactId>
         </exclusion>
       </exclusions>
     </dependency>
+    <!--Once newer hive-exec version leverages parquet-column 1.9.0, this dependency can be deleted -->
```
vrozov: Can it be moved to the dependency management if this is still necessary?

contrib/storage-hive/hive-exec-shade/pom.xml
vrozov: Can it be unified in the plugin management of the Drill root pom?

vdiravka:
@vrozov Thank you for CR. 
Commits are squashed, Jira and PR are renamed to 2.3.2 version upgrade.

amansinha100:
+1.   merged the PR in commit 27aa236.

PR #1112, opened 2018-02-06
DRILL-6114: Metadata revisions

paul-rogers:
This PR is part of the ["batch handling updates"](https://github.com/paul-rogers/drill/wiki/Batch-Handling-Upgrades) project. It completes the new internal metadata system by including support for the remaining vector types: unions, lists and repeated.

The metadata code was refactored. Previously, it was small enough to fit into a single file (with nested classes). With the added complexity, the metadata classes were split out into separate classes, grouped into its own Java package.

A few fixes were made here and there to ensure the unit tests pass.

@ppadma or @bitblender, can one of you run the pre-commit tests and send me the details of any failures?


ppadma:
@paul-rogers I ran the pre commit tests. No issues. Everything passed. Will do one more time once code reviews are done.

paul-rogers:
Thanks @ppadma!

Next we'll need a reviewer. @parthchandra or @arina-ielchiieva, is this something you can review?

Once this one is good, I'll follow up with another that includes revisions to the `SchemaBuilder` that allows tests to build schemas using the additional data types added here.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/RepeatedListColumnMetadata.java
```diff
@@ -0,0 +1,94 @@
...
+  @Override
+  public ColumnMetadata childSchema() { return childSchema; }
+
+  @Override
+  public int dimensions() {
+
+    // If there is no child, then we don't know the
+    // dimensionality.
+
+    return childSchema == null ? -1
```
ppadma: can we use a static final instead of -1 ?

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/ColumnMetadata.java
```diff
@@ -15,36 +15,115 @@
...
+     * if we have a separate category for 2D lists, we should have
+     * a separate category for 1D lists. But, again, that is not how
+     * the code has evolved.
+     */
+
+    MULTI_ARRAY
   }
 
-  public static final int DEFAULT_ARRAY_SIZE = 10;
+  int DEFAULT_ARRAY_SIZE = 10;
```
ppadma: why is this changed from static final int to int ?

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/AbstractColumnMetadata.java
```diff
@@ -0,0 +1,206 @@
...
+  // or MajorType because then encode child information that we encode here
+  // as a child schema. Keeping the two in sync is nearly impossible.
+
+  protected final String name;
+  protected final MinorType type;
+  protected final DataMode mode;
+  protected final int precision;
+  protected final int scale;
+  protected boolean projected = true;
+
```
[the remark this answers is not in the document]
paul-rogers: AFAIK, we do not support the time-zone aware data types at present.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/AbstractColumnMetadata.java
```diff
@@ -0,0 +1,206 @@
...
+  public boolean isArray() { return mode() == DataMode.REPEATED; }
+
+  @Override
+  public int dimensions() { return isArray() ? 1 : 0; }
+
+  @Override
+  public boolean isMap() { return false; }
+
+  @Override
+  public boolean isVariant() { return false; }
```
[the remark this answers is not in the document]
paul-rogers: I think there is a write-up somewhere about this. "Variant" is an old term from Visual Basic for a tagged union (which is what our "Union" type really is.) It is used here because both a (non-repeated) List and a Union are both unions, which is confusing, so I used a separate term. Also, the "Union" term is confusing since there is also a UNION operator in SQL. `See VariantMetadata.java`.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/MetadataUtils.java
```diff
@@ -0,0 +1,165 @@
...
+      if (field.getType().getMode() != DataMode.OPTIONAL) {
+        throw new UnsupportedOperationException(type.name() + " type must be nullable");
+      }
+      return new VariantColumnMetadata(field);
+    case LIST:
+      switch (field.getType().getMode()) {
+      case OPTIONAL:
+
+        // List of unions (or a degenerate union of a single type.)
+        // Not supported in Drill.
```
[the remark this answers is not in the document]
paul-rogers: We do. This comment is in the wrong place, should be in default.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/RepeatedListColumnMetadata.java
paul-rogers: Fixed.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/ColumnMetadata.java
paul-rogers: This is an interface. The default (and only legal) form of fields is `static final`. Thanks to Tim for reminding me of this.

paul-rogers:
@arina-ielchiieva, @parthchandra can either of you perhaps give this one a committer review? Thanks!

PR #1113, opened 2018-02-06
DRILL-5902: Queries encounter random failure due to RPC connection timed out

vrozov:


vrozov:
@arina-ielchiieva Please review

vrozov:
@ilooner DRILL-6143 is not related to DRILL-5902. DRILL-6143 requires a separate RCA. See DRILL-5902 for details.

ilooner:
K thx @vrozov

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryStateProcessor.java
```diff
@@ -125,20 +125,17 @@ public void cancel() {
...
       case PLANNING:
       case ENQUEUED:
-        moveToState(QueryState.CANCELLATION_REQUESTED, null);
-        return;
-
       case STARTING:
       case RUNNING:
-        addToEventQueue(QueryState.CANCELLATION_REQUESTED, null);
-        return;
+        moveToState(QueryState.CANCELLATION_REQUESTED, null);
```
arina-ielchiieva: 1. Could you please explain why `addToEventQueue` was changed to `moveToState`. Per my understanding `addToEventQueue` is used to ensure that cancellation will be requested only when all fragments are sent out to avoid hanging in cancellation requested state. For preparing, planning and enqueued states we cancel immediately since these states are done locally.
2. Why this Jira title states it's a regression? Do we know what has caused the regression?

vrozov: 1. Per my understanding, only during a transition from `STARTING` to `RUNNING` it is necessary to delay the processing of `CANCELLATION_REQUESTED` till requests are submitted to remote nodes for execution. Once the state is transitioned to `RUNNING`, remote drillbits are ready to start processing cancellation request, so no delay is necessary for the `RUNNING` state. In case of `STARTING` there is already a call to `addToEventQueue()` inside `QueryStateProcessor.starting()` that ensures that cancellation will be processed after the transition.
1. I can't say why JIRA title states that it is a regression, as far as I can tell, any delay in saving query profile may cause the issue. One possibility is that the issue is amplified by [DRILL-6053](https://issues.apache.org/jira/browse/DRILL-6053) that is a regression caused by the fix for [DRILL-4963](https://issues.apache.org/jira/browse/DRILL-4963) that introduces even longer delay (due to synchronization).

arina-ielchiieva: 1. Your point makes sense. In this case could you please update java doc for the `cancel` method to be consistent with new changes?
2. Maybe we should remove word `regression` from the commit message to avoid confusion?

vrozov: @arina-ielchiieva Please update JIRA title.

arina-ielchiieva:
+1, LGTM.

PR #1114, opened 2018-02-06
DRILL-6104: Added Logfile Reader

cgivre:
I would like to submit a format plugin that will enable Drill to read log files.   Here is a link to the github repo which contains documentation: 
https://github.com/cgivre/drill-logfile-plugin

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.store.log;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.google.common.base.Objects;
+import org.apache.drill.common.logical.FormatPluginConfig;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@JsonTypeName("logRegex")
+public class LogFormatConfig implements FormatPluginConfig {
+
+  private String regex;
+  private String extension;
+  private int maxErrors = 10;
+  private List<LogFormatField> schema;
+
+  public String getRegex() {
+    return regex;
+  }
+
+  public String getExtension() {
+    return extension;
+  }
+
+  public int getMaxErrors() {
+    return maxErrors;
+  }
+
+  public List<LogFormatField> getSchema() {
+    return schema;
+  }
+
+  //Setters
+  public void setExtension(String ext) {
+    this.extension = ext;
+  }
+
+  public void setMaxErrors(int errors) {
+    this.maxErrors = errors;
+  }
+
+  public void setRegex(String regex) {
+    this.regex = regex;
+  }
+
+  public void setSchema() {
+    this.schema = new ArrayList<LogFormatField>();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
+    if (obj == null || getClass() != obj.getClass()) {
+      return false;
+    }
+    LogFormatConfig other = (LogFormatConfig) obj;
+    return Objects.equal(regex, other.regex) &&
+        Objects.equal(maxErrors, other.maxErrors) &&
+        Objects.equal(schema, other.schema) &&
+        Objects.equal(extension, other.extension);
+  }
+
+  @Override
+  public int hashCode() {
+    return Arrays.hashCode(new Object[]{regex, maxErrors, schema, extension});
+  }
+
+  @JsonIgnore
+  public List<String> getFieldNames() {
+    List<String> result = new ArrayList<String>();
+    if (this.schema == null) {
+      return result;
+    }
+
+    for (LogFormatField field : this.schema) {
+      result.add(field.getFieldName());
+    }
+    return result;
+  }
+
+  @JsonIgnore
+  public String getDataType(int fieldIndex) {
+    LogFormatField f = this.schema.get(fieldIndex);
+    return f.getFieldType().toUpperCase();
+  }
+
+  @JsonIgnore
+  public LogFormatField getField(int fieldIndex) {
+    return this.schema.get(fieldIndex);
+  }
+
+  @JsonIgnore
+  public String getDateFormat(int patternIndex) {
+    LogFormatField f = this.schema.get(patternIndex);
+    return f.getFormat();
+  }
+}
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.drill.exec.store.log;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+@JsonTypeName("regexReaderFieldDescription")
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class LogFormatField {
+
+  /*
+   * The three configuration options for a field are:
+   * 1.  The field name
+   * 2.  The data type (fieldType).  Field type defaults to VARCHAR if it is not specified
+   * 3.  The format string which is used for date/time fields.  This field is ignored if used with a non
+   * date/time field.
+   * */
+
+  private String fieldName = "";
+  private String fieldType = "VARCHAR";
+  private String format;
+
+  //These will be used in the future for field validation and masking
+  //public String validator;
+  //public double minValue;
+  //public double maxValue;
+
+
+  public LogFormatField() {
+  }
+
+  //These constructors are used for unit testing
+  public LogFormatField(String fieldName) {
+    this(fieldName, null, null);
+  }
+
+  public LogFormatField(String fieldName, String fieldType) {
+    this(fieldName, fieldType, null);
+  }
+
+  public LogFormatField(String fieldName, String fieldType, String format) {
+    this.fieldName = fieldName;
+    this.fieldType = fieldType;
+    this.format = format;
+  }
+
+  public String getFieldName() {
+    return fieldName;
+  }
+
+  public String getFieldType() {
+    return fieldType;
+  }
+
+  public String getFormat() {
+    return format;
+  }
+
+
+  /*
+  public String getValidator() { return validator; }
+
+  public double getMinValue() { return minValue; }
+
+  public double getMaxValue() {
+    return maxValue;
+  }
+  */
+}
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.drill.exec.store.log;
+
+import com.google.common.collect.Lists;
+import org.apache.drill.common.exceptions.ExecutionSetupException;
+import org.apache.drill.common.expression.SchemaPath;
+import org.apache.drill.common.logical.StoragePluginConfig;
+import org.apache.drill.exec.ops.FragmentContext;
+import org.apache.drill.exec.proto.UserBitShared;
+import org.apache.drill.exec.server.DrillbitContext;
+import org.apache.drill.exec.store.RecordReader;
+import org.apache.drill.exec.store.RecordWriter;
+import org.apache.drill.exec.store.dfs.DrillFileSystem;
+import org.apache.drill.exec.store.dfs.easy.EasyFormatPlugin;
+import org.apache.drill.exec.store.dfs.easy.EasyWriter;
+import org.apache.drill.exec.store.dfs.easy.FileWork;
+import org.apache.hadoop.conf.Configuration;
+
+import java.util.List;
+
+public class LogFormatPlugin extends EasyFormatPlugin<LogFormatConfig> {
+
+  public static final String DEFAULT_NAME = "logRegex";
+  private final LogFormatConfig formatConfig;
+
+  public LogFormatPlugin(String name, DrillbitContext context,
+                         Configuration fsConf, StoragePluginConfig storageConfig,
+                         LogFormatConfig formatConfig) {
+    super(name, context, fsConf, storageConfig, formatConfig,
+        true,  // readable
+        false, // writable
+        true, // blockSplittable
+        true,  // compressible
+        Lists.newArrayList(formatConfig.getExtension()),
+        DEFAULT_NAME);
+    this.formatConfig = formatConfig;
+  }
+
+  @Override
+  public RecordReader getRecordReader(FragmentContext context,
+                                      DrillFileSystem dfs, FileWork fileWork, List<SchemaPath> columns,
+                                      String userName) throws ExecutionSetupException {
+    return new LogRecordReader(context, dfs, fileWork,
+        columns, userName, formatConfig);
+  }
+
+  @Override
+  public boolean supportsPushDown() {
+    return true;
+  }
+
+  @Override
+  public RecordWriter getRecordWriter(FragmentContext context,
+                                      EasyWriter writer) throws UnsupportedOperationException {
+    throw new UnsupportedOperationException("unimplemented");
+  }
+
+  @Override
+  public int getReaderOperatorType() {
+    return UserBitShared.CoreOperatorType.REGEX_SUB_SCAN_VALUE;
+  }
+
+  @Override
+  public int getWriterOperatorType() {
+    throw new UnsupportedOperationException("unimplemented");
+  }
+}
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
@@ -1,0 +1,764 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.store.log;
+
+import com.google.common.base.Charsets;
+import org.apache.drill.common.exceptions.UserException;
+import org.apache.drill.common.expression.SchemaPath;
+import org.apache.drill.common.types.TypeProtos.MinorType;
+import org.apache.drill.common.types.Types;
+import org.apache.drill.exec.exception.SchemaChangeException;
+import org.apache.drill.exec.ops.FragmentContext;
+import org.apache.drill.exec.ops.OperatorContext;
+import org.apache.drill.exec.physical.impl.OutputMutator;
+import org.apache.drill.exec.record.MaterializedField;
+import org.apache.drill.exec.store.AbstractRecordReader;
+import org.apache.drill.exec.store.dfs.DrillFileSystem;
+import org.apache.drill.exec.store.dfs.easy.FileWork;
+import org.apache.drill.exec.vector.NullableIntVector;
+import org.apache.drill.exec.vector.NullableBigIntVector;
+import org.apache.drill.exec.vector.NullableSmallIntVector;
+import org.apache.drill.exec.vector.NullableFloat4Vector;
+import org.apache.drill.exec.vector.NullableFloat8Vector;
+import org.apache.drill.exec.vector.BaseValueVector;
+import org.apache.drill.exec.vector.NullableDateVector;
+import org.apache.drill.exec.vector.NullableVarCharVector;
+import org.apache.drill.exec.vector.NullableTimeStampVector;
+import org.apache.drill.exec.vector.NullableTimeVector;
+
+
+import org.apache.hadoop.fs.Path;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Collection;
+import java.util.Date;
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+public class LogRecordReader extends AbstractRecordReader {
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LogRecordReader.class);
+
+  private abstract static class ColumnDefn {
+    private final String name;
+    private final int index;
+    private final String format;
+
+    public ColumnDefn(String name, int index) {
+      this(name, index, null);
+    }
+
+    public ColumnDefn(String name, int index, String format) {
+      this.name = name;
+      this.index = index;
+      this.format = format;
+    }
+
+    public abstract void define(OutputMutator outputMutator) throws SchemaChangeException;
+
+    public abstract void load(int rowIndex, String value);
+
+    public String getName() { return this.name; }
+
+    public int getIndex() { return this.index; }
+
+    public String getFormat() { return this.format;}
+
+    @Override
+    //For testing
+    public String toString() {
+      return "Name: " + name + ", Index: " + index + ", Format: " + format;
+    }
+  }
+
+  private static class VarCharDefn extends ColumnDefn {
+
+    private NullableVarCharVector.Mutator mutator;
+
+    public VarCharDefn(String name, int index) {
+      super(name, index);
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.VARCHAR));
+      mutator = outputMutator.addField(field, NullableVarCharVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      mutator.set(rowIndex, value.getBytes());
+    }
+  }
+
+  private static class BigIntDefn extends ColumnDefn {
+
+    private NullableBigIntVector.Mutator mutator;
+
+    public BigIntDefn(String name, int index) {
+      super(name, index);
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.BIGINT));
+      mutator = outputMutator.addField(field, NullableBigIntVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        mutator.set(rowIndex, Long.parseLong(value));
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse an INT field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class SmallIntDefn extends ColumnDefn {
+
+    private NullableSmallIntVector.Mutator mutator;
+
+    public SmallIntDefn(String name, int index) {
+      super(name, index);
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.SMALLINT));
+      mutator = outputMutator.addField(field, NullableSmallIntVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        mutator.set(rowIndex, Short.parseShort(value));
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse an INT field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class IntDefn extends ColumnDefn {
+
+    private NullableIntVector.Mutator mutator;
+
+    public IntDefn(String name, int index) {
+      super(name, index);
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.INT));
+      mutator = outputMutator.addField(field, NullableIntVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        mutator.set(rowIndex, Integer.parseInt(value));
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse an INT field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class Float4Defn extends ColumnDefn {
+
+    private NullableFloat4Vector.Mutator mutator;
+
+    public Float4Defn(String name, int index) {
+      super(name, index);
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.FLOAT4));
+      mutator = outputMutator.addField(field, NullableFloat4Vector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        mutator.set(rowIndex, Float.parseFloat(value));
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse an FLOAT field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class DoubleDefn extends ColumnDefn {
+
+    private NullableFloat8Vector.Mutator mutator;
+
+    public DoubleDefn(String name, int index) {
+      super(name, index);
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.FLOAT8));
+      mutator = outputMutator.addField(field, NullableFloat8Vector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        mutator.set(rowIndex, Double.parseDouble(value));
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse an FLOAT field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class DateDefn extends ColumnDefn {
+
+    private NullableDateVector.Mutator mutator;
+    private SimpleDateFormat df;
+
+    public DateDefn(String name, int index, String dateFormat) {
+      super(name, index, dateFormat);
+      df = getValidDateObject(dateFormat);
+    }
+
+    private SimpleDateFormat getValidDateObject(String d) {
+      SimpleDateFormat tempDateFormat;
+      if (d != null && !d.isEmpty()) {
+        tempDateFormat = new SimpleDateFormat(d);
+      } else {
+        throw UserException.parseError()
+            .message("Invalid date format.  The date formatting string was empty.")
+            .build(logger);
+      }
+      return tempDateFormat;
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.DATE));
+      mutator = outputMutator.addField(field, NullableDateVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        Date d = df.parse(value);
+        long milliseconds = d.getTime();
+        mutator.set(rowIndex, milliseconds);
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse an DATE field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      } catch (ParseException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Date Format String does not match field value.")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Format String", getFormat())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class TimeDefn extends ColumnDefn {
+
+    private NullableTimeVector.Mutator mutator;
+    private SimpleDateFormat df;
+
+    public TimeDefn(String name, int index, String dateFormat) {
+      super(name, index, dateFormat);
+      df = getValidDateObject(dateFormat);
+    }
+
+    private SimpleDateFormat getValidDateObject(String d) {
+      SimpleDateFormat tempDateFormat;
+      if (d != null && !d.isEmpty()) {
+        tempDateFormat = new SimpleDateFormat(d);
+      } else {
+        throw UserException.parseError()
+            .message("Invalid date format.  The date formatting string was empty.")
+            .build(logger);
+      }
+      return tempDateFormat;
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.TIME));
+      mutator = outputMutator.addField(field, NullableTimeVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        Date d = df.parse(value);
+        int milliseconds = (int) d.getTime();
+        mutator.set(rowIndex, milliseconds);
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse an Time field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      } catch (ParseException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Date Format String does not match field value.")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Format String", getFormat())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class TimeStampDefn extends ColumnDefn {
+
+    private NullableTimeStampVector.Mutator mutator;
+    private SimpleDateFormat df;
+
+    public TimeStampDefn(String name, int index, String dateFormat) {
+      super(name, index, dateFormat);
+      df = getValidDateObject(dateFormat);
+    }
+
+    private SimpleDateFormat getValidDateObject(String d) {
+      SimpleDateFormat tempDateFormat;
+      if (d != null && !d.isEmpty()) {
+        tempDateFormat = new SimpleDateFormat(d);
+      } else {
+        throw UserException.parseError()
+            .message("Invalid date format.  The date formatting string was empty.")
+            .build(logger);
+      }
+      return tempDateFormat;
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(getName(),
+          Types.optional(MinorType.TIMESTAMP));
+      mutator = outputMutator.addField(field, NullableTimeStampVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        Date d = df.parse(value);
+        long milliseconds = d.getTime();
+        mutator.set(rowIndex, milliseconds);
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse a TIMESTAMP field")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Value", value)
+            .build(logger);
+      } catch (ParseException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Date Format String does not match field value.")
+            .addContext("Column", getName())
+            .addContext("Position", getIndex())
+            .addContext("Format String", getFormat())
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static final int BATCH_SIZE = BaseValueVector.INITIAL_VALUE_ALLOCATION;
+
+  private final DrillFileSystem dfs;
+  private final FileWork fileWork;
+  private final String userName;
+  private final LogFormatConfig formatConfig;
+  private ColumnDefn columns[];
+  private Pattern pattern;
+  private BufferedReader reader;
+  private int rowIndex;
+  private int capturingGroups;
+  private OutputMutator outputMutator;
+  private int unmatchedColumnIndex;
+  private int unmatchedRowIndex;
+  private boolean unmatchedRows;
+  private int maxErrors;
+
+
+  private int errorCount;
+
+
+  public LogRecordReader(FragmentContext context, DrillFileSystem dfs,
+                         FileWork fileWork, List<SchemaPath> columns, String userName,
+                         LogFormatConfig formatConfig) {
+    this.dfs = dfs;
+    this.fileWork = fileWork;
+    this.userName = userName;
+    this.formatConfig = formatConfig;
+    this.unmatchedColumnIndex = -1;
+    this.unmatchedRowIndex = 0;
+    this.unmatchedRows = false;
+    this.maxErrors = formatConfig.getMaxErrors();
+
+    // Ask the superclass to parse the projection list.
+    setColumns(columns);
+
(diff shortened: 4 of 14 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+     */
+    if (regex.isEmpty()) {
+      throw UserException.parseError().message("Log parser requires a valid, non-empty regex in the plugin configuration").build(logger);
+    } else {
+      //TODO Check for invalid regex
+      r = Pattern.compile(regex);
+      Matcher m = r.matcher("test");
+      if (m.groupCount() == 0) {
+        throw UserException.parseError().message("Invalid Regular Expression: No Capturing Groups", 0).build(logger);
+      } else if (m.groupCount() != (fieldNames.size())) {
```
paul-rogers: Suggestion: `!=` -> `>=`. This allows debugging. Define all fields. Then, work out the pattern for the first. Then do the second. Very hard to get all fields right the first time. Nuisance to remove fields temporarily.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+        throw UserException.parseError().message("Invalid Regular Expression: Field names do not match capturing groups.  There are " + m.groupCount() + " captured groups in the data and " + fieldNames.size() + " specified in the configuration.", 0).build(logger);
```
paul-rogers: Generally we try to keep lines to a reasonable size. Also, use the form that will do formatting for you. Suggest:

```
       throw UserException.parseError()
            .message("Invalid Regular Expression: Field names do not match capturing groups.  " +
                             "There are %d captured groups in the data and "
                             " %d specified in the configuration.",
                             m.groupCount(), fieldNames.size())
            .build(logger);
```

Also, maybe shorten the message: `"Found %d groups but %d fields"`

Further, since both might have been in the query (via table properties), might be worth while including the regex and the fields.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+
+      } else if ((dataTypes == null) || m.groupCount() != dataTypes.size()) {
+        //If the number of data types is not correct, create a list of varchar
```
paul-rogers: Oh my! If I'm short one type, I get no warning, just the wrong data types?

Maybe:

* If null, silently fill in VarChar types.
* If fewer types than fields, fill in missing types with VarChar, logging that we are doing this.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+        dataTypes = new ArrayList<String>();
+        for (int i = -0; i < m.groupCount(); i++) {
```
paul-rogers: `i = 0` should work, `-0` isn't necessary.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+          dataTypes.add("VARCHAR");
```
paul-rogers: `MinorType.VARCHAR.name()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+      }
+    }
+
+    //Check and set up date formats
+    if (dataTypes.contains("DATE") || dataTypes.contains("TIMESTAMP")) {
```
paul-rogers: As above.

But, Drill uses SQL rules: names are case insensitive. So, need to use a different solution. (Does Guava have something?)

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+      if (dateFormat != null && !dateFormat.isEmpty()) {
+        df = new java.text.SimpleDateFormat(dateFormat);
```
paul-rogers: I believe this will throw an exception for a bad format. Should catch it.

Also, OK to use `SimpleDateFormat` and import the class. This code does not have the same draconian restrictions as UDFs.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+      } else {
+        throw UserException.parseError().message("Invalid date format.  The date formatting string was empty.  Please specify a valid date format string in the configuration for this data source.", 0).build(logger);
```
paul-rogers: No need to include the `0` argument. It is for filling in patterns as shown above. Here and below.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+      }
+    }
+
+    if (dataTypes.contains("TIME")) {
```
paul-rogers: As above.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  @Override
+  public int getWriterOperatorType() {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public boolean supportsPushDown() {
+    return true;
```
paul-rogers: This claims that this plugin handles (projection) push-down, but no such code exists in the implementation.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+
+    try {
+      BaseWriter.MapWriter map = this.writer.rootAsMap();
+      String line = null;
+
+      while (recordCount < MAX_RECORDS_PER_BATCH && (line = this.reader.readLine()) != null) {
+        lineCount++;
+
+        // Skip empty lines
+        if (line.trim().length() == 0) {
```
paul-rogers: Better:

```
line = line.trim();
if (line.isEmpty) { ...
```

That way, the matcher below gets the benefit of the trimmed whitespace.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+          continue;
+        }
+
+        this.writer.setPosition(recordCount);
+        map.start();
```
paul-rogers: Do this below after matching the line. Else, we'll do this for non-matching lines.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+
+        Matcher m = r.matcher(line);
+        if (m.find()) {
+          for (int i = 1; i <= m.groupCount(); i++) {
```
paul-rogers: Again, the computer will handle this just fine if we move the row-processing code into another method. But, it will reduce the nesting levels us poor reviewers have to track.

paul-rogers: Why a `1`-based loop with the funny `i - 1` indexing? Just do `i + 1` as needed to fetch the group.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+            String type = dataTypes.get(i - 1);
+            String fieldValue;
+
+            fieldValue = m.group(i);
+
+            if (fieldValue == null) {
+              fieldValue = "";
+            }
+
+            if (type.toUpperCase().equals("INT") || type.toUpperCase().equals("INTEGER")) {
```
paul-rogers: `type.equalsIgnoreCase(MinorType.INT.name())` and so on.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+              map.integer(fieldName).writeInt(Integer.parseInt(fieldValue));
+            } else if (type.toUpperCase().equals("DOUBLE") || type.toUpperCase().equals("FLOAT8")) {
```
paul-rogers: Even better. This is an inner loop. Presumably we want it to go fast (which is why it is deeply nested.)

So, rather than doing (slow) string compares (you could have used a `switch` statement instead), in the `setup()` method, map names to ordinals: `MinorType.ordinal()`. Then do a `switch` on the resulting numbers using, say, `MinorType.VARCHAR.ordinal()`. Much faster.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+              map.float8(fieldName).writeFloat8(Double.parseDouble(fieldValue));
```
paul-rogers: The above is going to be slow as well. For speed, also cache the writers.

Or, you can use the new `ResultSetLoader` which provides index access to its writers.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+
+    } catch (final Exception e) {
+      throw UserException.dataReadError(e).build(logger);
+    }
+  }
+
+  public void close() throws Exception {
+    this.reader.close();
+  }
+}
```
paul-rogers: No unit tests?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
cgivre: Hi @paul-rogers 
Is there an example of the `ResultSetLoader` that you could point me to?

paul-rogers:
See [this example](https://github.com/paul-rogers/drill/tree/regex-plugin/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/regex), and [this test](https://github.com/paul-rogers/drill/blob/regex-plugin/exec/java-exec/src/test/java/org/apache/drill/exec/store/easy/regex/TestRegexReader.java) for examples of one way to address some of the comments made in the code review. That example handles projection, which was mentioned in one of the review comments.

Since the regex is file-specific, the regex format plugin is most useful if it can be configured per-file using table functions. See [DRILL-6167](https://issues.apache.org/jira/browse/DRILL-6167), [DRILL-6168](https://issues.apache.org/jira/browse/DRILL-6168) and [DRILL-6169](https://issues.apache.org/jira/browse/DRILL-6169) for problems that will be encountered. See the example and tests above for how to work around the bugs.

Although I suggested looking at the `ResultSetLoader`, it is a bit premature to do so. That mechanism relies on additional mechanisms that have not yet been committed to master. So, we need to work with the mechanisms we have now. See the example reader above for how to cache the per-column mutator needed to write to vectors without the switch statements in the code in this PR.

To generalize, the example has an object per column that holds per-column info. The example uses only `VarChar` columns. To add additional types, create a base column state class with subclasses for each type. Then, simply call a `save(String value)` method to write a column. That method can handle nulls (for projected non-existent columns) and type conversions (where needed).

Finally, feel free to borrow liberally from the example. (The example was created for our Drill book, so is fair game to reuse.)

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,151 @@
+package org.apache.drill.exec.store.log;
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+  public static class LogFormatConfig implements FormatPluginConfig {
+    public List<String> extensions;
+    public List<String> fieldNames;
+    public List<String> dataTypes;
+    public String dateFormat = "";
+    public String timeFormat = "HH:mm:ss";
+    public String pattern;
+    public Boolean errorOnMismatch = false;
+
+    private static final List<String> DEFAULT_EXTS = ImmutableList.of("log");
```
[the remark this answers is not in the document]
cgivre: I added a default regex of `(.*)` which should capture the entire line as well as a field name of `full_line` by default.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+      return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (this == obj) {
+        return true;
+      } else if (obj == null) {
+        return false;
+      } else if (getClass() == obj.getClass()) {
```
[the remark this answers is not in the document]
cgivre: I'm honestly not really sure what this code does, but I fixed it per your recommendation.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public Boolean getErrorOnMismatch() {
+      return errorOnMismatch;
+    }
+
+    @Override
+    public int hashCode() {
+      int result = pattern != null ? pattern.hashCode() : 0;
+      result = 31 * result + (dateFormat != null ? dateFormat.hashCode() : 0) + (timeFormat != null ? timeFormat.hashCode() : 0);
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+      Path hdfsPath = new Path(inputPath);
+      Configuration conf = new Configuration();
+      FSDataInputStream fsStream = fileSystem.open(hdfsPath);
+      CompressionCodecFactory factory = new CompressionCodecFactory(conf);
+      CompressionCodec codec = factory.getCodec(hdfsPath);
+      if (codec == null) {
+        reader = new BufferedReader(new InputStreamReader(fsStream.getWrappedStream(), "UTF-8"));
+      } else {
+        CompressionInputStream comInputStream = codec.createInputStream(fsStream.getWrappedStream());
+        reader = new BufferedReader(new InputStreamReader(comInputStream));
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,261 @@
...
+    1.  Empty regex
+    2.  Invalid Regex
+    3.  Empty date string if the date format is used
+    4.  No capturing groups in the regex
+    5.  Incorrect number of data types
+    6.  Invalid data types
+     */
+    if (regex.isEmpty()) {
+      throw UserException.parseError().message("Log parser requires a valid, non-empty regex in the plugin configuration").build(logger);
+    } else {
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
cgivre: @paul-rogers Are there any examples of how to implement this?  I would like this to be as fast as possible.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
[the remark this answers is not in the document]
paul-rogers: The comment says fixed. Did you forget to commit the changes to your branch? Github still shows the original code...

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+        patternIndex = Integer.parseInt(m.group(1));
+      } else {
+        for (int i = 0; i < fields.size(); i++) {
+          if (fields.get(i).equalsIgnoreCase(name)) {
+            patternIndex = i;
+            break;
+          } else if( name.equals("_raw")){
+            patternIndex = -1;
+            break;
+          } else if( name.equals("_unmatched_rows")){
```
paul-rogers: `_unmatched_rows` sounds like there is more than one. But, we operate only at the row level. So, maybe `_unmatched` is more consistent with `_raw`.

But, we have `_raw`, so no need to repeat the whole row for `_unmatched`.

Maybe we need a config option about whether to ignore unmatched rows. If so, they'll never show up in the output. If we include them, then we just need a Boolean to say if the value is matched or not. So, instead of `_unmatched`, we might include a `_match`. Then, to get all the information:

```
SELECT a, b, _raw, _match FROM ...
```

I always get the `_raw` row. The `_match` tells me if the `a`, `b` values are filled in.

Or, maybe to be frugal, we keep the `_unmatched` which is like `_raw`, but is filled in only for unmatched rows.

Still, seems weird to have a single query return both matched and unmatched rows. I'd think I'd want to scan for unmatched rows so I can fix them, then have all other queries worry only about matched rows, else consumer of the data has to handle the match/unmatch issue. How, for example, would a BI tool chart values that contain unmatched rows?

So, is there a way to select either matched or unmatched rows? Taking a step back, maybe explain these use cases somewhere.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+      }
+      columns[colIndex++] = makeColumn(name, patternIndex);
+    }
+  }
+
+  private ColumnDefn makeColumn(String name, int patternIndex) {
+    String typeName = null;
+    if (patternIndex <= -1) {
+      // Use VARCHAR for missing columns
+      // (instead of Drill standard of nullable int)
```
paul-rogers: In my original example, -1 indicated a missing column, as described in the comment. In this version, negative values are also used for the whole row. So, perhaps update the comment accordingly.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+    } else if (patternIndex < formatConfig.dataTypes.size()) {
+      typeName = formatConfig.dataTypes.get(patternIndex);
+    }
+    if (typeName == null) {
+      // No type name. VARCHAR is a safe guess
+      typeName = MinorType.VARCHAR.name();
+    }
+    if( name.equals("_raw")){
+      return new VarCharDefn(name, patternIndex);
+    }
```
paul-rogers: The above is redundant. You've used negative indexes for `_raw` and other special columns. The negative index is handled in the if-statements above. No need to special case it here, is there?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+
+    MinorType type = MinorType.valueOf(typeName);
+    switch (type) {
+      case VARCHAR:
+        return new VarCharDefn(name, patternIndex);
+      case INT:
+      case SMALLINT:
+      case BIGINT:
```
paul-rogers: Isn't this a bit of a lie? The user expresses a data type, then we second-guess it here. If I ask for `BigInt`, I will be somewhat surprised that it only fits `Int` values. Similarly, below, if I ask for a `Decimal` value, I'll be surprised that I get a float, which can lose precision.

Should we be more strict about the types supported?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+            .build(logger);
+    }
+  }
+
+  private void defineVectors() {
+    for (int i = 0; i < columns.length; i++) {
+      try {
+        columns[i].define(outputMutator);
+      } catch (SchemaChangeException e) {
+        throw UserException
```
paul-rogers: Was not in my example, but perhaps throw in a comment here saying why the above call might fail. There is only one reason: if this reader reads multiple files, and the schema of a prior file differed from the one for this file.

The way this plugin is defined (schema is part of the format plugin definition) that case can never occur, so we can add a comment saying that.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+      logger.warn("Unmatached line: {}", line);
+    } else if( errorCount > formatConfig.getMaxErrors() ){
+      throw UserException.parseError()
+          .message("Too many errors.  Max error threshold exceeded.")
+          .addContext("Line", line)
+          .addContext("Line number", rowIndex)
+          .build(logger);
+    }
+    //If the user asked for the unmatched columns display them
+    if(unmatchedColumnIndex != -1 ) {
```
paul-rogers: `if (unmatchedColumnIndex != -1) {`

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+
+  @Override
+  public int next() {
+    rowIndex = 0;
+    while (nextLine()) {
+    }
+    return rowIndex;
+  }
+
+  private boolean nextLine() {
```
paul-rogers: The code above mentions a `_raw` column, but I don't see it being filled in anywhere here.

Should the `_raw` column be set only for matched rows? For both matched and unmatched rows? See discussion above.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+  }
+
+  private void loadVectors(Matcher m) {
+    for (int i = 0; i < columns.length; i++) {
+      if (columns[i].index >= 0) {
+        String value = m.group(columns[i].index + 1);
+        if (value != null) {
+          columns[i].load(rowIndex, value);
+        }
+      } else if( columns[i].index == -1 ){
```
paul-rogers: Oh, my. In my example, the value of `-1` indicated a column that was projected, but did not exist in the data source and so should always be null (the default value).

Here, the value of `-1` is being used to fill in the entire row. So, if I select `bogus` from a file, the value won't be null, it will be the input row.

Suggestion: this behavior is needed only for the `_raw` column. Simply set a field to the index of the raw column if it exists, as was done earlier for the `_unmatched` column.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+        String value = m.group(0);
+        if (value != null) {
+          columns[i].load(rowIndex, value);
+        } else {
+          rowIndex++;
```
paul-rogers: Ah oh... The row index is incremented twice. Is that what was wanted? Returning two rows, one with values filled in, one with all nulls?

The above issues raise an important question: do unit tests exist for all these cases? If so, the tests should have found the problems, not your code reviewer.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
```diff
@@ -0,0 +1,177 @@
...
+import org.apache.drill.test.rowSet.RowSet;
+import org.apache.drill.test.rowSet.RowSetComparison;
+import org.apache.drill.test.rowSet.schema.SchemaBuilder;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class TestLogReader extends ClusterTest {
```
paul-rogers: The tests here are pretty much identical to the regex log plugin example that I created. The hope was that the tests would be extended to cover all functionality that is claimed in the plugin:

* Test date, time and timestamp formats -- both valid and invalid.
* Test with valid and invalid values where parsed. (Ints, dates, floats, etc.)
* Test all supporte data types.
* Test the two special columns.

In general, when submitting code, it will save lots of time if thorough unit tests are also provided. As it is, I had to find these bugs via code insertion which is not automated.

exec/java-exec/src/test/resources/regex/simple.log2
```diff
@@ -0,0 +1,5 @@
+2017-12-17 10:52:41,820 [main] INFO  o.a.d.e.e.f.FunctionImplementationRegistry - Function registry loaded.  459 functions loaded in 1396 ms.
```
paul-rogers: As far as I can tell, these two files are identical. I think you got these from my example, so that is probably my fault; I guess I didn't realize that they were the same...

We really only need one of them.

src/main/resources/checkstyle-config.xml
```diff
@@ -30,7 +30,7 @@
     <module name="UnusedImports"/>
     <module name="IllegalImport">
       <!-- For "org.apache.commons.lang.*" classes use the corresponding class from "org.apache.commons.lang3.*" -->
-      <property name="illegalPkgs" value="com.google.hive12,com.beust.jcommander.internal,jersey.repackaged,org.apache.commons.lang"/>
+      <property name="illegalPkgs" value="com.google.hive12,jersey.repackaged,org.apache.commons.lang"/>
```
paul-rogers: Why was this changed? Rebase error? Maybe revert to `master` version (for this file only)?

cgivre: The unit tests for this use `com.beust.jcommander.internals` which was listed as an illegal import.   Not sure why that's the case, but I had to do that in order to make it work.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+@JsonTypeName("log")
+public class LogFormatConfig implements FormatPluginConfig {
+
+  public String regex;
+  public List<String> fields = new ArrayList<>();
+  public String extension;
+  public List<String> dataTypes = new ArrayList<>();
```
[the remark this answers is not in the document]
cgivre: That's my ultimate goal yes.  But I know there's an impending deadline for Drill 1.14 and I'd like to get this in there, then I'll work on feature enhancements.  Actually, what I was really thinking was to be able to include additional validators.

IE:
```
{
  "name":"port_num".
  "type":"int",
  "min": 0,
  "max": 1024
}
```
Or include some other validators for things like IPv4, IPv6, MAC addresses, etc.   This way the user could write a relatively simple regex and rely on the validators to make sure that the data is clean.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
cgivre: So the way it works, is that if you have a data file with 5 rows, 4 of which match, and you do a SELECT *, you get 4 rows back.   If you do a `SELECT _raw` you also get 4 rows back.  
Right now, if you do a `SELECT _unmatched_rows` you get four null rows and 1 row with the unmatched row. 

I'm not sure what the best behavior is honestly, but I was trying to balance usability in that we don't want Drill to break every time it encounters bad data, but at the same time, we also want the user to have the ability to identify rows that don't match the regex.  I really like the idea of the `max_errors` parameter, in that you can tune how sensitive you want Drill to be.  If you want Drill to blow up every time it encounters an error you can set that to zero.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
```diff
@@ -0,0 +1,89 @@
...
+        Objects.equal(dataTypes, other.dataTypes) &&
+        Objects.equal(dateFormat, other.dateFormat) &&
+        Objects.equal(timeFormat, other.timeFormat) &&
+        Objects.equal(maxErrors, other.maxErrors) &&
+        Objects.equal(extension, other.extension);
+  }
+
+  @Override
+  public int hashCode() {
+    return Arrays.hashCode(new Object[]{regex, fields, extension});
```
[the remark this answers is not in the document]
cgivre: I don't know either.   They seem to be present in all the format plugins.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
cgivre: To see all the rows that match, you could simply do a:
```
SELECT _raw FROM dfs.`log`;
```
IF you want to see the unmatched rows with no nulls:
```
SELECT `_unmatched_rows` 
FROM <data>
WHERE _unmatched_rows IS NOT NULL
```

Depending on the `max_errors` setting, the format-plugin pretty much ignores unmatched rows.

exec/java-exec/src/test/resources/regex/simple.log2
cgivre: Removed log2.

exec/java-exec/src/test/java/org/apache/drill/exec/store/dfs/TestFormatPluginOptionExtractor.java
```diff
@@ -72,9 +72,15 @@ public void test() {
               "(type: String, fileSystemMetadata: boolean, descriptive: boolean, timeZone: String)", d.presentParams()
           );
           break;
+        case "logRegex":
+          assertEquals(LogFormatConfig.class, d.pluginConfigClass);
+          assertEquals(
+              "(type: String, regex: String, extension: String, maxErrors: int, schema: List)", d.presentParams()
```
paul-rogers: This test is now tightly coupled to the `toString()` implementation, which is awkward since, in general, folks might change that at any time to make debugging easier.

Better:
```
assertEquals("logRegex", d.type);
assertNull(d.extension);
...
assertEquals(0, d.maxErrors);
```

Which raises an interesting question: what are good default values?

Similar test is needed for fields. Here, what are good defaults? What should be done if certain values are bogus or missing?

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
```diff
@@ -0,0 +1,368 @@
...
+        .addRow("070823 21:00:56       1 Query       select * from category", null )
+        .addRow("070917 16:29:01      21 Query       select * from location", null )
+        .addRow("070917 16:29:12      21 Query       select * from location where id = 1 LIMIT 1", null )
+        .addRow( null, "dfadkfjaldkjafsdfjlksdjflksjdlkfjsldkfjslkjl")
+        .build();
+
+    new RowSetComparison(expected).verifyAndClearAll(results);
+  }
+
+}
```
paul-rogers: Missing a test for a bogus type name, repeated field names and other similar error conditions. What else should be checked? Null or empty file extension? Negative minimum errors? min/max errors out of sync? What else?

The way to think about this is: we only need to test those cases that we don't want the Drill team to have to fix in response to an escalation in a production system. If it is OK to wait to fix the issue until an escalation, we don't need to test it. (Now, I'm sure the Drill team will suggest that there are no such cases that are OK to fix in an escalation...)

Here's how I think about this. For every field that that the user might set, think of all possible values, write code to test for, and fail, for bad values, then write tests to provide that he bad values are rejected.

pom.xml
```diff
@@ -216,6 +216,10 @@
         <configuration>
           <excludeSubprojects>false</excludeSubprojects>
           <excludes>
+            <exclude>**/*.log1</exclude>
```
paul-rogers: Suggestion: to help future folks, include a brief comment:

```
<!-- Types log1, log2, sqllog an sqllog2 are used to test he logRegex format plugin. -->
```

src/main/resources/checkstyle-config.xml
paul-rogers: Which aspect was being used? There is no comment in the config file to say why that library was excluded. Without that info, I'd kind of assume that there was a good reason, and find some other way to get the tests to work.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+public class LogRecordReader extends AbstractRecordReader {
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LogRecordReader.class);
+
+  private abstract static class ColumnDefn {
+    public final String name;
```
paul-rogers: Now that you've added a class to hold per-field properties, does it make sense to change this internal structure to reference that config structure? That way, all the info is in one place: both config and runtime stuff.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+    public final int index;
+    public final String format;
+
+    public ColumnDefn(String name, int index) {
+      this.name = name;
```
paul-rogers:
```
this(name, index, "");
```

Should a missing format be null or empty?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+    }
+
+    public abstract void define(OutputMutator outputMutator) throws SchemaChangeException;
+
+    public abstract void load(int rowIndex, String value);
+
+    @Override
+    //For testing
+    public String toString() {
+      return "Name: " + name + ", Index: " + index;
```
paul-rogers: Format?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+    private NullableSmallIntVector.Mutator mutator;
+
+    public SmallIntDefn(String name, int index) {
+      super(name, index);
+    }
+
+    @Override
+    public void define(OutputMutator outputMutator) throws SchemaChangeException {
+      MaterializedField field = MaterializedField.create(name,
+          Types.optional(MinorType.BIGINT));
```
paul-rogers: Hmm... Kind of confused. The class name is `SmallIntDefn`, the mutter is for `NullableSmallIntVector`, but the minor type being used is `BIGINT`.

In fact, this tells us that we never tested this field type. Should we go back and make sure that we have a consistent story across accepted type names, supported types internally, and unit tests?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
```diff
@@ -0,0 +1,97 @@
...
+    return f.fieldType.toUpperCase();
+  }
+
+  @JsonIgnore
+  public String getDateFormat( int patternIndex ){
+    LogFormatField f = this.schema.get(patternIndex);
+    return f.format;
+  }
+
+}
```
paul-rogers: General comment: this reader has gotten robust enough that it probably needs a bit of explanation. Can we include a `README.md` file (or `package-info.java`, if you prefer JavaDoc) to explain how a user should configure and use this reader?

Maybe document the supported column types, the syntax of the formats (point to the proper Java classes), and the existence of your special columns.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+      mutator = outputMutator.addField(field, NullableSmallIntVector.class).getMutator();
+    }
+
+    @Override
+    public void load(int rowIndex, String value) {
+      try {
+        mutator.set(rowIndex, Integer.parseInt(value));
```
paul-rogers: This is for `SmallInt` which accepts a reduced set of values, but the parser above is for `Integer` which accepts the full int range. Should the above code include a range check to ensure that the parsed int is within the allowed range?

Is there a unit test that verifies the behavior of when a field is outside the valid range (both on the + and - sides)?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+
+    public DateDefn(String name, int index, String dateFormat) {
+      super(name, index, dateFormat);
+      df = getValidDateObject(dateFormat);
+    }
+
+    private SimpleDateFormat getValidDateObject(String d) {
+      SimpleDateFormat tempDateFormat;
+      if (d != null && !d.isEmpty()) {
+        tempDateFormat = new SimpleDateFormat(d);
```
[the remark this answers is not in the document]
paul-rogers: Any resolution for the above?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+            .addContext("Failed to parse an FLOAT field")
+            .addContext("Column", name)
+            .addContext("Position", index)
+            .addContext("Value", value)
+            .build(logger);
+      }
+    }
+  }
+
+  private static class DoubleDefn extends ColumnDefn {
```
paul-rogers: Do we have detailed tests for each data type?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,623 @@
...
+
+    public TimeDefn(String name, int index, String dateFormat) {
+      super(name, index, dateFormat);
+      df = getValidDateObject(dateFormat);
+    }
+
+    private SimpleDateFormat getValidDateObject(String d) {
+      SimpleDateFormat tempDateFormat;
+      if (d != null && !d.isEmpty()) {
+        tempDateFormat = new SimpleDateFormat(d);
```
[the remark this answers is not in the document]
paul-rogers: Does the combination of date and time actually work to handle a timestamp? The timestamp is in which timezone? Do we need code to convert the timestamp from its ISO+TZ format to UTC (which drill kinda-sorta-not-really uses internally)? Or, to be more accurate, should we convert the ISO UTC time into the local time zone which Drill uses in place of UTC in its dates? (Yes, this is VERY broken and confusing.)

At the very least, should we use a timestamp type in place of separate date and time types so that there is at least some chance that the combined timestamp can be converted correctly across timezones (and converted correctly by JDBC)?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+
+    public TimeStampDefn(String name, int index, String dateFormat) {
+      super(name, index, dateFormat);
+      df = getValidDateObject(dateFormat);
+    }
+
+    private SimpleDateFormat getValidDateObject(String d) {
+      SimpleDateFormat tempDateFormat;
+      if (d != null && !d.isEmpty()) {
+        tempDateFormat = new SimpleDateFormat(d);
```
paul-rogers: Skip `tempDateFormat`, just do:

```
return new SimpleDateFormat(d);
```

Java will figure out that the `else` clause throws an exception.

In fact, no need for an `else`. Since the `then` will return, the default action can be to throw the exception.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+      } catch (NumberFormatException e) {
+        throw UserException
+            .dataReadError(e)
+            .addContext("Failed to parse a TIMESTAMP field")
+            .addContext("Column", name)
+            .addContext("Position", index)
+            .addContext("Value", value)
+            .build(logger);
+      } catch (ParseException e) {
+        throw UserException
```
paul-rogers: Thoughts?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+        return new Float4Defn(name, patternIndex);
+      case FLOAT8:
+        return new DoubleDefn(name, patternIndex);
+      case DATE:
+        return new DateDefn(name, patternIndex, formatConfig.getDateFormat(patternIndex));
+      case TIMESTAMP:
+        return new TimeStampDefn(name, patternIndex, formatConfig.getDateFormat(patternIndex));
+      case TIME:
+        return new TimeDefn(name, patternIndex, formatConfig.getDateFormat(patternIndex));
+      default:
```
paul-rogers: Again, unit tests for each of these formats? And, for disallowed type names?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+      case FLOAT4:
+        return new Float4Defn(name, patternIndex);
+      case FLOAT8:
+        return new DoubleDefn(name, patternIndex);
+      case DATE:
+        return new DateDefn(name, patternIndex, formatConfig.getDateFormat(patternIndex));
+      case TIMESTAMP:
+        return new TimeStampDefn(name, patternIndex, formatConfig.getDateFormat(patternIndex));
+      case TIME:
+        return new TimeDefn(name, patternIndex, formatConfig.getDateFormat(patternIndex));
```
paul-rogers: Per earlier comment, now that you have the field format config, just get that object, and pass it into each constructor:

```
return new TimeDefn(patternIndex, fieldConfig);
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+
+  private void loadVectors(Matcher m) {
+    String value = null;
+    /*if( unmatchedRows && columns.length == 1 ){
+      return;
+    }*/
+
+    for (int i = 0; i < columns.length; i++) {
+      //Skip the unmatched rows column
+      if( columns[i].name.equals("_unmatched_rows")){
```
paul-rogers: This is pretty expensive: doing a string compare for each column for each row. We also do this below.

Maybe define another column type for a "Special" column. Then, have every type implement `isSpecial()`. Now, you can simply skip `isSpecial()` columns rather than doing string comparisons.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
```diff
@@ -0,0 +1,750 @@
...
+
+      if (columns[i].index >= 0) {
+        //Get the value of the regex group
+        value = m.group(columns[i].index + 1);
+
+        //If the value is not null, assign it to the column
+        if (value != null) {
+          columns[i].load(rowIndex, value);
+        }
+      } else if (columns[i].name.equals("_raw")) {
```
paul-rogers: Seems awkward.

During setup, there should be a choice: either pick out all columns, or just pick out the raw row. This means that we either get the raw row, or we get the columns.

Otherwise, handle it by updating the column class methods. Perhaps those columns that implement `isSpecial()` are given the entire row. Just making something up:

```
  if (m.isSpecial()) {
    ((RowColumn) m).loadRow(rowIndex, m.group(0), hasMatch);
  }
```

That is, how can we use method dispatch to a subclass (which is relatively fast) to substitute for if-statements with string comparisons (which is slow)?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
```diff
@@ -0,0 +1,97 @@
...
+
+  public String getExtension() {
+    return extension;
+  }
+
+  public int getMaxErrors() {
+    return maxErrors;
+  }
+
+  public List<LogFormatField> getSchema() { return schema; }
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @JsonIgnore
+  public String getDataType(int fieldIndex) {
+    LogFormatField f = this.schema.get(fieldIndex);
+    return f.fieldType.toUpperCase();
+  }
+
+  @JsonIgnore
+  public String getDateFormat( int patternIndex ){
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    for( LogFormatField field: this.schema ){
+      result.add( field.getFieldName() );
+    }
+    return result;
+  }
+
+  @JsonIgnore
+  public String getDataType(int fieldIndex) {
```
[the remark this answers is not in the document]
cgivre: Well...   This was the result of me coding at 2AM, but this function is used once:
```
else if (patternIndex < formatConfig.schema.size()) {
      typeName = formatConfig.getDataType(patternIndex);
    }
```
in `LogRecordReader.java.`.  
Not sure which is easier, but I added an additional function as requested.  Resultant code in LogRecordReader is:
```
LogFormatField tempField = formatConfig.getField(patternIndex);
      typeName = tempField.getFieldType();
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
```diff
@@ -0,0 +1,72 @@
...
+
+@JsonTypeName("fieldDescription")
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class LogFormatField {
+  public String fieldName = "";
+  public String fieldType = "VARCHAR";
+  public String validator;
+  public double minValue;
+  public double maxValue;
+  public String format;
```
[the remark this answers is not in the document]
cgivre: These fields are experimental at this point.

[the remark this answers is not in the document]
cgivre: My plan for the future is to implement various ways of validating the data, but first I wanted to get this initial release working.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
```diff
@@ -0,0 +1,72 @@
...
+
+  public double getMinValue() { return minValue; }
+
+  public double getMaxValue() {
+    return maxValue;
+  }
+
+  public String getFormat() {
+    return format;
+  }
```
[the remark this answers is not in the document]
cgivre: Fixed BUT... FYSA, the code template for IntelliJ did this automatically.  I'm not sure why it picked some getters to be one liners and others not.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,84 @@
...
+
+  public LogFormatPlugin(String name, DrillbitContext context,
+                         Configuration fsConf, StoragePluginConfig storageConfig,
+                         LogFormatConfig formatConfig) {
+    super(name, context, fsConf, storageConfig, formatConfig,
+        true,  // readable
+        false, // writable
+        true, // blockSplittable
+        true,  // compressible
+        Lists.newArrayList(formatConfig.extension),
```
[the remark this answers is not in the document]
cgivre: I looked and I don't think other format plugins validate the extension.  I would suspect that if it is blank, the format plugin won't ever get called as it won't find any files.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Override
+  public RecordWriter getRecordWriter(FragmentContext context,
+                                      EasyWriter writer) throws IOException {
+    return null;
+  }
+
+  @Override
+  public int getReaderOperatorType() {
+    return 0;
```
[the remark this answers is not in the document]
arina-ielchiieva: Yes, we definitely need to generate proto name.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public int getWriterOperatorType() {
+    return 0;
```
arina-ielchiieva: Please throw `UnsupportedOperationException` here.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
arina-ielchiieva: Plus consider using not public access variable.

src/main/resources/checkstyle-config.xml
arina-ielchiieva: That's definitely needs to be reverted.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
cgivre: Hi @paul-rogers 
Not necessarily.   There are cases where you might want the `_raw` row AND the fields.  Debugging the regex for example.  You might want to compare the raw data with what Drill is extracting.  

I could think of other situations where you have a complex log, but are only extracting a few select fields by default.

src/main/resources/checkstyle-config.xml
cgivre: Reverted

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.drill.exec.store.log;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+@JsonTypeName("fieldDescription")
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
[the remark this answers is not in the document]
cgivre: I commented out these variables with an explanation in the source.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  public LogFormatField( String fieldName, String fieldType ){
+    this.fieldName = fieldName;
+    this.fieldType = fieldType;
+  }
+
+  public LogFormatField( String fieldName, String fieldType, String format ){
+    this.fieldName = fieldName;
+    this.fieldType = fieldType;
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogRecordReader.java
cgivre: Part of me wonders whether there is any utility in offering data types other than Date/Time, int, Float. String.

contrib/native/client/patches/zookeeper-3.4.6-x64.patch
```diff
@@ -1,163 +0,0 @@
...
-+    struct adaptor_threads *adaptor = zh->adaptor_priv;
-+    if(adaptor)
-+        pthread_mutex_lock(&adaptor->zh_lock);
-+}
-+
- void enter_critical(zhandle_t* zh)
- {
-     struct adaptor_threads *adaptor = zh->adaptor_priv;
--- 
-1.9.2.msysgit.0
```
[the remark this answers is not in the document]
cgivre: Gone...

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatConfig.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.drill.common.logical.FormatPluginConfig;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@JsonTypeName("logRegex")
+public class LogFormatConfig implements FormatPluginConfig {
+
+  public String regex;
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
```diff
@@ -0,0 +1,72 @@
...
+
+package org.apache.drill.exec.store.log;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+@JsonTypeName("fieldDescription")
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class LogFormatField {
+  public String fieldName = "";
```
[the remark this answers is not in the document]
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
cgivre: Thrown

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
paul-rogers: Just for future reference: we did provide an example and the latest code reflects that example.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
paul-rogers: For example: invalid date, invalid number, etc.

arina-ielchiieva:
@cgivre looks like PR is ready to commit. Please squash the commits and leave one with proper commit message. Maybe add extra unit tests as Paul asked if you have time.

cgivre:
Commits squashed and proper commit message added.  

Regarding the unit tests, is there an example of a unit test for when we are expecting Drill to throw an exception?

PR #1115, opened 2018-02-06
DRILL-6138: Move RecordBatchSizer to org.apache.drill.exec.record pac…

ppadma:
…kage
Also,  changed columnSizes in RecordBatchSizer from list to map so we can lookup using field names.

@Ben-Zvi can you please review ?


exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -260,7 +254,7 @@ public static ColumnSize getColumn(ValueVector v, String prefix) {
 
   public static final int MAX_VECTOR_SIZE = ValueVector.MAX_BUFFER_SIZE; // 16 MiB
 
-  private List<ColumnSize> columnSizes = new ArrayList<>();
+  private Map<String, ColumnSize> columnSizes = CaseInsensitiveMap.newHashMap();
```
Ben-Zvi: Tim (@ilooner) made the same change for DRILL-6032 ( #1101 )

ilooner: I'll rebase after this change goes in.

PR #1117, opened 2018-02-08
DRILL-6089 Removed ordering trait from HashJoin in planner

ilooner:
HashJoin typically does not preserve ordering in most databases. Drill's HashJoin operator technically preserved ordering up to this point, but after spilling is implemented it will no longer preserve ordering. This change makes sure the planner knows that HashJoin does not preserve ordering.

All unit and functional tests are passing @amansinha100 @Ben-Zvi please review.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/JoinPruleBase.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/JoinPruleBase.java
@@ -116,7 +116,6 @@ public abstract class JoinPruleBase extends Prule {
     }
   }
 
-
   // Create join plan with both left and right children hash distributed. If the physical join type
   // is MergeJoin, a collation must be provided for both left and right child and the plan will contain
   // sort converter if necessary to provide the collation.
@@ -126,8 +125,6 @@ public abstract class JoinPruleBase extends Prule {
       RelCollation collationLeft, RelCollation collationRight,
       DrillDistributionTrait hashLeftPartition, DrillDistributionTrait hashRightPartition) throws InvalidRelException {
 
-    //DrillDistributionTrait hashLeftPartition = new DrillDistributionTrait(DrillDistributionTrait.DistributionType.HASH_DISTRIBUTED, ImmutableList.copyOf(getDistributionField(join.getLeftKeys())));
-    //DrillDistributionTrait hashRightPartition = new DrillDistributionTrait(DrillDistributionTrait.DistributionType.HASH_DISTRIBUTED, ImmutableList.copyOf(getDistributionField(join.getRightKeys())));
     RelTraitSet traitsLeft = null;
     RelTraitSet traitsRight = null;
 
@@ -146,7 +143,8 @@ public abstract class JoinPruleBase extends Prule {
     DrillJoinRelBase newJoin = null;
 
     if (physicalJoinType == PhysicalJoinType.HASH_JOIN) {
-      newJoin = new HashJoinPrel(join.getCluster(), traitsLeft,
+      final RelTraitSet traitSet = PrelUtil.removeCollation(traitsLeft, call);
+      newJoin = new HashJoinPrel(join.getCluster(), traitSet,
                                  convertedLeft, convertedRight, join.getCondition(),
                                  join.getJoinType());
 
@@ -236,7 +234,8 @@ public abstract class JoinPruleBase extends Prule {
         call.transformTo(new MergeJoinPrel(join.getCluster(), convertedLeft.getTraitSet(), convertedLeft,
             convertedRight, joinCondition, join.getJoinType()));
       } else if (physicalJoinType == PhysicalJoinType.HASH_JOIN) {
-        call.transformTo(new HashJoinPrel(join.getCluster(), convertedLeft.getTraitSet(), convertedLeft,
+        final RelTraitSet traitSet = PrelUtil.removeCollation(convertedLeft.getTraitSet(), call);
+        call.transformTo(new HashJoinPrel(join.getCluster(), traitSet, convertedLeft,
             convertedRight, joinCondition, join.getJoinType()));
       } else if (physicalJoinType == PhysicalJoinType.NESTEDLOOP_JOIN) {
         call.transformTo(new NestedLoopJoinPrel(join.getCluster(), convertedLeft.getTraitSet(), convertedLeft,
@@ -245,5 +244,4 @@ public abstract class JoinPruleBase extends Prule {
     }
 
   }
-
 }
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinAdvanced.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinAdvanced.java
@@ -18,7 +18,6 @@
 
 package org.apache.drill.exec.physical.impl.join;
 
-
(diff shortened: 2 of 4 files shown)


HanumathRao:
@ilooner Changes look fine to me. However, as discussed offline I couldn't reproduce the plan which was sorting one of the inputs for a HASHJOIN. This might be because in the hashjoinprule we are not asking for those traits in the first place. However, I do agree that this fix can be a safegaurd in case if that plan happens to show up in some scenarios.


Thanks for the fix.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/TestHashJoinOrdering.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.drill.exec.planner.physical.PlannerSettings;
+import org.apache.drill.test.QueryTestUtil;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category({SlowTest.class, HiveStorageTest.class})
+public class TestHashJoinOrdering extends HiveTestBase {
```
amansinha100: I am wondering why a separate HashJoinOrdering test is needed for Hive. The Drill join planner is not doing something specific for Hive tables.  If we have a Hive test, why not for other storage plugins such as HBase etc but since the proposed fix removes Collation property for any input stream of a HashJoin, it should not matter where that input originated from.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/JoinPruleBase.java
```diff
@@ -246,4 +246,17 @@ public RelNode convertChild(final DrillJoinRel join,  final RelNode rel) throws
 
   }
 
+  // DRILL-6089 make sure no collations are added to HashJoin
+  public static RelTraitSet removeCollation(RelTraitSet traitSet, RelOptRuleCall call)
```
amansinha100: Since this is static method, better to add it to PrelUtil class for general purpose use.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/JoinPruleBase.java
```diff
@@ -246,4 +246,17 @@ public RelNode convertChild(final DrillJoinRel join,  final RelNode rel) throws
...
+  {
+    RelTraitSet newTraitSet = call.getPlanner().emptyTraitSet();
+
+    for (RelTrait trait: traitSet) {
+      if (!trait.getTraitDef().getTraitClass().equals(RelCollation.class)) {
```
amansinha100: I suppose you already considered Calcite's RelTraitSet.replace() method and found this to be simpler ?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/FragmentsRunner.java
```diff
@@ -61,7 +61,7 @@
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FragmentsRunner.class);
   private static final ControlsInjector injector = ControlsInjectorFactory.getInjector(FragmentsRunner.class);
 
-  private static final long RPC_WAIT_IN_MSECS_PER_FRAGMENT = 5000;
+  private static final long RPC_WAIT_IN_MSECS_PER_FRAGMENT = 30000;
```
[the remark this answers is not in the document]
ilooner: These changes were added as a separate commit to fix issues with our functional tests and build so that I could test this change. I've opened separate PRs for these changes so I'll remove them from this PR.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/TestHashJoinOrdering.java
ilooner: removed

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/JoinPruleBase.java
ilooner: moved

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/JoinPruleBase.java
ilooner: I don't see a way to remove a trait with RelTraitSet.replace(). So I made this removeCollation method instead.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinAdvanced.java
```diff
@@ -197,4 +198,24 @@ public void emptyPartTest() throws Exception {
...
     }
   }
+
+  @Test // DRILL-6089
+  public void testJoinOrdering() throws Exception {
+    final String query = "select * from dfs.`sample-data/nation.parquet` nation left outer join " +
+      "(select * from dfs.`sample-data/region.parquet`) " +
+      "as region on region.r_regionkey = nation.n_nationkey order by region.r_name desc";
+    final String plan = getPlanInString("EXPLAIN PLAN for " + QueryTestUtil.normalizeQuery(query), OPTIQ_FORMAT);
+    lastSortAfterJoin(plan);
```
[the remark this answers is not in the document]
ilooner: The method `testPlanMatchingPatterns(String query, String[] expectedPatterns, String[] excludedPatterns)` is not sufficient out of the box since I have to add the `Pattern.DOTALL` flag when compiling the pattern in order to have the regex match across new lines. So I've added `testPlanMatchingPatterns(String query, Pattern[] expectedPatterns, Pattern[] excludedPatterns)`

ilooner:
@amansinha100 I have addressed review comments and all tests are passing. Please let me know if you have anymore comments.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinAdvanced.java
```diff
@@ -197,4 +199,14 @@ public void emptyPartTest() throws Exception {
       BaseTestQuery.resetSessionOption(ExecConstants.SLICE_TARGET);
     }
   }
+
+  @Test // DRILL-6089
+  public void testJoinOrdering() throws Exception {
+    final String query = "select * from dfs.`sample-data/nation.parquet` nation left outer join " +
```
amansinha100: I missed this in the prior review but I think the ORDER BY should be on the column coming from the left input of the join (assuming that nation table is the left input).  The reason is the original code was using 'convertedLeft' to propagate the collation trait from the Probe (left) side, so the test should match that scenario. Another thing is the you don't necessarily need the subquery.. you could  do  `select * from  nation.parquet nation  left outer join region.parquet region on ... order by nation.n_name desc`.

amansinha100:
@ilooner thanks for making the changes.  I have a minor comment on the test query.  
Otherwise LGTM.  +1.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinAdvanced.java
ilooner: Thanks for catching this, updated the test.

PR #1118, opened 2018-02-08
DRILL-6144: Make directMemory amount configurable in tests

ilooner:
Travis builds have been hanging as well as unit tests on our Jenkins server. It is not clear what the underlying cause is but increasing direct memory resolves the issue. 

One possible root cause is that some unit tests leak direct memory by creating allocators which allocate memory and are then never closed. It will take a while to find a true fix for the issue and resolve it, so to unblock our work for now I would like to increase direct memory available for the build.

**Update:** For now we will not change the default amount of direct memory allocated until we find the root cause. If you need to you can pass the -DdirectMemoryMb property to override the amount of memory in your build.


ilooner:
@Ben-Zvi Please review

PR #1119, opened 2018-02-09
DRILL-6143: Made FragmentsRunner's rpc timeout a SystemOption

ilooner:
Made FragmentsRunner's rpc timeout a SystemOption, and made the default larger to minimize random failures.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
@@ -212,7 +212,8 @@ public class SystemOptionManager extends BaseOptionManager implements AutoClosea
       new OptionDefinition(ExecConstants.CPU_LOAD_AVERAGE),
       new OptionDefinition(ExecConstants.ENABLE_VECTOR_VALIDATOR),
       new OptionDefinition(ExecConstants.ENABLE_ITERATOR_VALIDATOR),
-      new OptionDefinition(ExecConstants.OUTPUT_BATCH_SIZE_VALIDATOR, new OptionMetaData(OptionValue.AccessibleScopes.SYSTEM, true, false))
+      new OptionDefinition(ExecConstants.OUTPUT_BATCH_SIZE_VALIDATOR, new OptionMetaData(OptionValue.AccessibleScopes.SYSTEM, true, false)),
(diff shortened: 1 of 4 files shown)


ilooner:
@arina-ielchiieva

priteshm:
@arina-ielchiieva is on vacation. @vrozov, @Ben-Zvi  can you take a look?

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -413,6 +413,7 @@ drill.exec.options: {
     # to start at least 2 partitions then HashAgg fallbacks to this case. It can be
     # enabled by setting this flag to true. By default it's set to false such that
     # query will fail if there is not enough memory
+    drill.exec.rpc.fragrunner.timeout: 30000,
```
vrozov: The value looks a little bit high and it needs to be moved either below `drill.exec.hashagg.fallback.enabled` or above the preceding comment, otherwise LGTM.

ilooner: Thanks for catching the ordering. I reduced the default to 10000.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -212,7 +212,8 @@
       new OptionDefinition(ExecConstants.CPU_LOAD_AVERAGE),
       new OptionDefinition(ExecConstants.ENABLE_VECTOR_VALIDATOR),
       new OptionDefinition(ExecConstants.ENABLE_ITERATOR_VALIDATOR),
-      new OptionDefinition(ExecConstants.OUTPUT_BATCH_SIZE_VALIDATOR, new OptionMetaData(OptionValue.AccessibleScopes.SYSTEM, true, false))
+      new OptionDefinition(ExecConstants.OUTPUT_BATCH_SIZE_VALIDATOR, new OptionMetaData(OptionValue.AccessibleScopes.SYSTEM, true, false)),
+      new OptionDefinition(ExecConstants.FRAG_RUNNER_RPC_TIMEOUT_VALIDATOR, new OptionMetaData(OptionValue.AccessibleScopes.SYSTEM, false, true)),
```
Ben-Zvi: Question: Should the last two parameters be instead:
   adminOnly = true
   internal = false

ilooner: internal should be true since we want this to show up in the internal options table and not the standard system options table. Changing to adminOnly = true seems reasonable.

vrozov:
LGTM

PR #1120, opened 2018-02-09
DRILL-6148: TestSortSpillWithException is sometimes failing.

HanumathRao:
I have changed the test case to use the EXTERNAL_SORT_MAX_MEMORY to configure the sort operator to use less memory instead of relying on MAX_QUERY_MEMORY_PER_NODE_KEY.

This way the test case passes all the time (it is failing intermittently in my branch).

@paul-rogers  can you please review this change.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSortSpillWithException.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSortSpillWithException.java
@@ -59,6 +59,11 @@ public class TestSortSpillWithException extends ClusterTest {
     ClusterFixtureBuilder builder = ClusterFixture.builder(dirTestWatcher)
         .configProperty(ExecConstants.EXTERNAL_SORT_SPILL_THRESHOLD, 1) // Unmanaged
         .configProperty(ExecConstants.EXTERNAL_SORT_SPILL_GROUP_SIZE, 1) // Unmanaged
+        // Using EXTERNAL_SORT_MAX_MEMORY to set low values of memory for SORT instead
+        // of lowering MAX_QUERY_MEMORY_PER_NODE_KEY because computation of operator memory
+        // cannot go lower than MIN_MEMORY_PER_BUFFERED_OP (the default value of this parameter
+        // is 40MB). The 40MB memory is sufficient for this testcase to run sort without spilling.
+        .configProperty(ExecConstants.EXTERNAL_SORT_MAX_MEMORY, 10 * 1024 * 1024)
         .sessionOption(ExecConstants.MAX_QUERY_MEMORY_PER_NODE_KEY, 60 * 1024 * 1024) // Spill early
         // Prevent the percent-based memory rule from second-guessing the above.
         .sessionOption(ExecConstants.PERCENT_MEMORY_PER_QUERY_KEY, 0.0)


HanumathRao:
@Ben-Zvi  Can you please review this change.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/xsort/TestSortSpillWithException.java
```diff
@@ -59,6 +59,7 @@ public static void setup() throws Exception {
     ClusterFixtureBuilder builder = ClusterFixture.builder(dirTestWatcher)
         .configProperty(ExecConstants.EXTERNAL_SORT_SPILL_THRESHOLD, 1) // Unmanaged
         .configProperty(ExecConstants.EXTERNAL_SORT_SPILL_GROUP_SIZE, 1) // Unmanaged
+        .configProperty(ExecConstants.EXTERNAL_SORT_MAX_MEMORY, 10 * 1024 * 1024) //use less memory for sorting.
```
paul-rogers: The need to set this property suggest something is wrong with the next one. We set max query memory per node to 60 MB. Then, the `MemoryAllocationUtilitites` is supposed to divide that up. If you have to override that, then something is amiss. We should track down that problem.

A good place to start is to turn on logging for this test (using the log fixture.) The sort will dump its configured memory. If it is still 2 GB, then debug the memory allocation utilities.

Perhaps it is as simple as using the wrong option below...

HanumathRao: @paul-rogers  Thank you for the review. In this case, the memory configured per node is 60MB and the maxParallelization is set to 1. The min buffered op memory is 40MB by default.
   * Minimum memory allocated to each buffered operator instance.
   * <p/>
   * DEFAULT: 40 MB

In my branch (which has got other changes also) 40MB is being correctly set for the sort to use. However, this memory is sufficient for the concerned query to not to spill the data to disk and it is executing fine without triggering the exception mechanism.

I tried to set max_per_node memory to 10MB but even after setting the memorycomputation utilities will only set sort memory to 40MB. Hence it is of no use in this case.

Therefore, I have used this option so that this test will consistently pass. Please let me know if I am missing anything here.

paul-rogers: There is something very wrong here. This means that the memory calcs are not working and users will try to sort in unusually small amounts of memory. The query probably has one sort. It should have just a few threads. (We might want to control this if we don't now.) Take the 60 MB of memory, divide by the number of sorts (1) and you get 60 MB for the sort in this case. If we are getting 40, then something is wrong. Let's find out where things went off the rails.

The minimum buffer op memory is supposed to set a floor for those cases in which a query has, say, 5 sorts on a machine with 30 cores and we try to divide the 2 GB by 150 and end up trying to sort in, say, 13 MB of memory.

HanumathRao: @paul-rogers  I am sorry for not being clear in my previous reply. This query has only one sort and also one minor fragment. If we set 60MB as MAX_QUERY_MEMORY_PER_NODE_KEY, then the result of the computeOperatorMemory() is 60MB.

However, as I said this memory is sufficient for the query to run successfully without spilling.

Then I tried out setting MAX_QUERY_MEMORY_PER_NODE_KEY to as low as 10MB so that this query should spill and fail with an exception (for this testcase). But even after setting so low the output of the computeOperatorMemory is 40MB. This is because of the default value of MIN_MEMORY_PER_BUFFERED_OP is 40MB. Hence, 40MB is used by the sort to sort the rows. This memory is enough for the sort to successfully run without an exception.

Thats when I used the setting EXTERNAL_SORT_MAX_MEMORY to set low values so that Sort for sure can spill and throw the exception which is being tested by this testcase.

paul-rogers: Please insert a comment to this effect in the code. This is a regression introduced when we added the memory floor feature. Explain that we need to turn off the floor (or lower it) for this test case. Thanks.

HanumathRao: @paul-rogers  Thank you. I have added the comment for the testcase. Please let me know if any other changes are required.

PR #1121, opened 2018-02-13
DRILL-6153: Operator framework

paul-rogers:
Includes the core files for the operator framework revision. See [this writeup](https://github.com/paul-rogers/drill/wiki/BH-Operator-Framework) for details.

In this commit, nothing depends on this new code. It is, instead, the foundation for the revised scan operator to be added after the revised result set loader code is committed. Doing this commit now allows small PRs to be done in parallel.


ilooner:
Hi @paul-rogers, nice to see this code being encapsulated, standardized, and unit tested. There seems to be a Travis test failure with your changes though.

```
Failed tests: 
  TestOperatorRecordBatch.testNormalLifeCycle:156 expected null, but was:<org.apache.drill.test.OperatorFixture$MockFragmentContext@45f1feca>
```

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorDriver.java
```diff
@@ -0,0 +1,183 @@
...
+      OperatorRecordBatch.logger.error("Exception thrown from cancel() for " + operatorLabel(), t);
+    }
+  }
+
+  private String operatorLabel() {
+    return operatorExec.getClass().getCanonicalName();
+  }
+
+  public void close() {
+    if (state == State.CLOSED) {
```
ppadma: should we check for  FAILED and END state also here ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorExec.java
```diff
@@ -0,0 +1,127 @@
...
+   *
+   * @return true if another batch of data is available, false if
+   * EOF was reached and no more data is available
+   */
+
+  boolean next();
+
+  /**
+   * Alerts the operator that the query was cancelled. Generally
+   * optional, but allows the operator to realize that a cancellation
```
ppadma: why is this optional ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorRecordBatch.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+
+  @Override
+  public FragmentContext getContext() {
+
+    // Backward compatibility with the full server context. Awkward for testing
+
+    FragmentContext fragmentContext = fragmentContext();
+    if (fragmentContext instanceof FragmentContext) {
```
ppadma: not clear what we are doing here. why can't we just return fragmentContext ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorDriver.java
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * State machine that drives the operator executable. Converts
+ * between the iterator protocol and the operator executable protocol.
+ * Implemented as a separate class in anticipation of eventually
+ * changing the record batch (iterator) protocol.
+ */
+
+public class OperatorDriver {
+  public enum State { START, SCHEMA, RUN, END, FAILED, CLOSED }
```
ppadma: Do we need SCHEMA state as you are moving from START to RUN or END or FAILED. may be combine START and SCHEMA states to say GET_SCHEMA or something like that ? Also, would it be good to have two states FAILED and CANCELLED to differentiate whether the query was cancelled or failed due to error.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorDriver.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+ /**
+   * Start the operator executor. Bind it to the various contexts.
+   * Then start the executor and fetch the first schema.
+   * @return result of the first batch, which should contain
+   * only a schema, or EOF
+   */
+
+  private IterOutcome start() {
```
ppadma: would be good if we can capture what exceptions each of these functions might throw.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorDriver.java
```diff
@@ -0,0 +1,183 @@
...
+   * Operator context. The driver "owns" the context and is responsible
+   * for closing it.
+   */
+
+  private final OperatorContext opContext;
+  private final OperatorExec operatorExec;
+  private final BatchAccessor batchAccessor;
+  private int schemaVersion;
+
+  public OperatorDriver(OperatorContext opServicees, OperatorExec opExec) {
```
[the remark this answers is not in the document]
paul-rogers: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorDriver.java
paul-rogers: Somewhere I explained the policy assumed here:

* Implementations of the `OperatorExec` interface are responsible for error handling.
* Implementations should catch exceptions then translate them to a `UserException` with user-oriented information.
* If implementations do the above, this layer simply passes along the `UserException`.
* Otherwise, this class acts as a translation layer to convert generic unchecked exceptions into `UserException`s. This layer can't provide much context; but it can do a slightly better job than the fragment exec, which is the last line of defense.

As a result, this layer always throws `UserException`s.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorDriver.java
paul-rogers: No. The goal here is simply to silently catch duplicate calls to `close()`. Whether the operator succeeded or failed, the implementation still must release its resources. But, by catching repeated `close()` calls, the implementation can assume that its own `close()` is called exactly once.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/OperatorExec.java
paul-rogers: An operator will work if it ignores `cancel()`. It's `close()` method should "do the right thing":

* If the operator reached EOF
* If the operator did not reach EOF (still rows left)
* If the operator failed

The `cancel()` call is just a hint that "I won't be reading any more rows; feel free to release resources now if you like."

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/protocol/VectorContainerAccessor.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Override
+  public WritableBatch getWritableBatch() {
+    return WritableBatch.get(container);
+  }
+
+  @Override
+  public SelectionVector2 getSelectionVector2() {
+    // Throws an exception by default
```
ppadma: should we make that explicit by indicating what exceptions it might throw.

paul-rogers:
Rebased on latest master.

paul-rogers:
@arina-ielchiieva, can you do a committer review of this one?

PR #1122, opened 2018-02-16
DRILL-6164: Heap memory leak during parquet scan and OOM

vrozov:
@parthchandra Please take a look


parthchandra:
+1. Very nice catch!

vrozov:
One of the functional tests fails. I am looking into it, please don't merge yet.

vrozov:
The functional test failure is fixed. Please merge.

PR #1125, opened 2018-02-21
 DRILL-6126: Allocate memory for value vectors upfront in flatten operator

ppadma:
Made changes to allocate memory upfront for flatten operator based on sizing calculations.
Need to do allocation of single column (can be nested) for a particular record count
and allocation hints. Refactored the code a bit for that.
Instead of assuming worst case fragmentation factor of 2, changed the logic to round down the number of rows calculated to nearest power of two. This will allow us to pack value vectors more densely and will help with memory utilization.
RepeatedMapvector and RepeatedListVector are extended from RepeatedFixedWidthVectorLike. This is wrong and causing problems in Allocation logic (allocatePrecomputedChildCount in AllocationHelper more specifically). Fixed that.
This PR has 2 commits. One for all of the above and second one for
DRILL-6162: Enhance record batch sizer to retain nesting information.


ppadma:
@paul-rogers Paul, will you be able to review this ?

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -418,11 +438,13 @@ private void measureColumn(ValueVector v, String prefix) {
...
         // above change 8 to 4 after DRILL-5446 is fixed
+
+    return colSize;
   }
 
-  private void expandMap(AbstractMapVector mapVector, String prefix) {
+  private void expandMap(ColumnSize colSize, AbstractMapVector mapVector, String prefix) {
     for (ValueVector vector : mapVector) {
-      measureColumn(vector, prefix);
+      colSize.childColumnSizes.put(prefix + vector.getField().getName(), measureColumn(vector, prefix));
```
paul-rogers: This is subject to aliasing. Suppose I have two maps:

```
aa(b)
a(ab)
```
When I add the child vectors, both will produce a combined name of `aab`.

We can't use dots n names for the same reason:

```
a.b(c)
a(b.c)
```

Both will produce `a.b.c`.

In the new "result set loader" code, all places that handle trees of columns use actual trees of maps.

A crude-but-effecive solution is to use a non-legal name character. The only valid one is the back-tick since we use that in SQL to quote names. If we do that, we now have

```
aa`b
a`ab
a.b`c
a`b.c
```

And the names are now un-aliased.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -108,6 +108,12 @@
     public final float estElementCountPerArray;
     public final boolean isVariableWidth;
 
+    public Map<String, ColumnSize> childColumnSizes = CaseInsensitiveMap.newHashMap();
```
paul-rogers: Nit: `children` might be shorter and simpler. It is clear that the children are sizes from the member type.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -245,16 +251,30 @@ private void buildVectorInitializer(VectorInitializer initializer) {
...
 
+  public ColumnSize getColumn(String name) {
+    return allColumnSizes.get(name);
+  }
+
   public static final int MAX_VECTOR_SIZE = ValueVector.MAX_BUFFER_SIZE; // 16 MiB
 
-  private Map<String, ColumnSize> columnSizes = CaseInsensitiveMap.newHashMap();
+  // This keeps information for all columns i.e. all top columns and nested columns underneath
+  private Map<String, ColumnSize> allColumnSizes = CaseInsensitiveMap.newHashMap();
```
paul-rogers: I'm a bit confused. I saw above that a `ColumnSize` has a list of children. Why are the children repeated here, introducing the naming issues described below?

Given the column alias issues, and how column index is used elsewhere, can this just be an index of top-level columns? Then, to size map vectors (the only one that has the nesting issue), use the code for recursion that already exists in the `VectorInitializer`?

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorInitializer.java
```diff
@@ -97,7 +97,7 @@ public void allocateBatch(VectorAccessible va, int recordCount) {
     }
   }
 
-  private void allocateVector(ValueVector vector, String prefix, int recordCount) {
+  public void allocateVector(ValueVector vector, String prefix, int recordCount) {
```
paul-rogers: Note the error in the member variable in this class:

```
   private Map<String, AllocationHint> hints = new HashMap<>();
```

Should be a case insensitive map.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -245,16 +251,30 @@ private void buildVectorInitializer(VectorInitializer initializer) {
       else if (width > 0) {
         initializer.variableWidth(name, width);
       }
+
```
paul-rogers: There may be a but in the original code above this line.

```
       String name = prefix + metadata.getName();
...
           initializer.variableWidthArray(name, ...
```

First problem: this method does not recurse into the child vectors of maps. Second, if it did, since each initializer is supposed to hold a name relative to its parent tuple (row or map), the prefix is not needed.

The vector initializer code was originally created for sort and none of the tests for sort include maps, so I probably did not catch the fact that maps don't work correctly with the original code.

Suggestion: add unit tests for this stuff. This code is getting very complex and tests would be very helpful to catch these subtle bugs.

Disclaimer: this is all supposed to be replaced with the ResultSetLoader code. The vector allocation code there is much cleaner, does handle maps, and has been unit tested.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
```diff
@@ -266,12 +270,18 @@ public void addComplexWriter(ComplexWriter writer) {
...
-        return false;
-      }
+      // build vector initializer for the column.
+      // This will iteratively include all nested columns underneath.
+      RecordBatchSizer.ColumnSize colSize = flattenMemoryManager.getColumnSize(v.getField().getName());
+      VectorInitializer initializer = new VectorInitializer();
+      colSize.buildVectorInitializer(initializer);
+      // Allocate memory for the vector. If it is map, it will allocate memory
+      // for all nested child columns as well.
+      initializer.allocateVector(v, "", recordCount);
```
paul-rogers: While this code can be made to work, it does introduce a more complex path than was intended. The idea is that a `VectorInirializer` is a recursive structure. The top-level one holds hints for the row. Nested instances can hold data for maps.

Because this class was meant to be temporary, it holds "hints": the information is used if available, defaults used if the hints are not available.

So, a better approach would be to assemble a `VectorInitializer` for the output row in one step. Then, apply it to the entire row in another step.

Further if we do that, we don't have to create initializer objects for every vector allocation; we can reuse the same set if the output row sizes don't change.

Further, the code will be easier to reason about since we won't have two distinct paths.

ppadma:
@paul-rogers Paul, Thanks a lot for your review comments and bringing up some good issues. Just want to let you know. I am working on refactoring the batch sizer code, writing bunch of unit tests to test sizing and vector allocation for all different vector types. Found some bugs in the process and fixed them. I will be posting new changes soon and need your review once they are ready.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
ppadma: I made the change to maintain trees of columns for trees of maps. Any one who wants to access the lower level columns, they get top level column and have to walk through the tree.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
ppadma: yes, I got rid of allColumnSizes. We will have only top level columns.

ppadma:
@paul-rogers Updated the PR with latest changes.  I have decided not to use vector initializer for allocation as it is subject to alias issues like you mentioned. Instead, I added allocate vector method to columnSize in batch sizer, which will use internal sizing information it has to allocate memory (including all it's children) for a particular record count. 
Refactored the batch sizer code, added unit tests for verifying sizing and allocation for different vector types. 
Please take a look when you get a chance and let me know what you think.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
```diff
@@ -492,8 +504,12 @@ private void allocateBatch(boolean newSchema) {
     } else {
       container.zeroVectors();
     }
+
+    // Allocate memory for the vectors.
+    // This will iteratively allocate memory for all nested columns underneath.
     for (VectorWrapper w : container) {
-      AllocationHelper.allocateNew(w.getValueVector(), Character.MAX_VALUE);
+      RecordBatchSizer.ColumnSize colSize = mergeJoinMemoryManager.getColumnSize(w.getField().getName());
+      colSize.allocateVector(w.getValueVector(), mergeJoinMemoryManager.getOutputRowCount());
```
paul-rogers: Nice! Much simpler. Just one nit: might as well get the row count once and store it in a local variable to avoid getting the same number repeatedly.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -36,13 +39,17 @@
...
 import com.google.common.collect.Sets;
+import org.apache.drill.exec.vector.complex.RepeatedVariableWidthVectorLike;
 
 /**
  * Given a record batch or vector container, determines the actual memory
  * consumed by each column, the average row, and the entire record batch.
  */
 
 public class RecordBatchSizer {
+  private static final int OFFSET_VECTOR_WIDTH = 4;
```
paul-rogers: Perhaps use `UInt4Vector.VALUE_WIDTH` to initialize this constant. Use `UInt1Vector.VALUE_WIDTH` for the bit vector width.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -76,110 +82,327 @@
...
     /**
-     * Size of the top level value vector. For map and repeated list,
-     * this is just size of offset vector.
+     * The estimated, average number of elements per parent value.
+     * Always 1 for a non-repeated type. For a repeated type,
+     * this is the average entries per array (per repeated element).
      */
-    public int dataSize;
+
+    private float estElementCountPerArray;
```
paul-rogers: Perhaps `estCardinality`? I've been using the term "cardinality" in new code to refer to array sizes.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -76,110 +82,327 @@
...
+    private boolean isVariableWidth;
 
     /**
-     * The estimated, average number of elements per parent value.
-     * Always 1 for a non-repeated type. For a repeated type,
-     * this is the average entries per array (per repeated element).
+     * Indicates if cardinality is repeated(top level only).
+     */
+
+    private boolean isRepeated;
```
paul-rogers: Might be fun to check out the new metadata classes added for the result set loader. They parse the `MajorType` to pull out this kind of information. You could embed an instance of the `ColumnMetadata` class here to provide this detailed information.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -76,110 +82,327 @@
...
       return children;
     }
 
+    public boolean isComplex() {
+      if (metadata.getType().getMinorType() == MinorType.MAP ||
+        metadata.getType().getMinorType() == MinorType.UNION ||
+        metadata.getType().getMinorType() == MinorType.LIST) {
+        return true;
+      }
+      return false;
```
paul-rogers: Nit, but
```
return metadata.getType().getMinorType() == MinorType.MAP ||
       metadata.getType().getMinorType() == MinorType.UNION ||
       metadata.getType().getMinorType() == MinorType.LIST;
```

And below.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -76,110 +82,327 @@
...
-      default:
-        dataSize = v.getPayloadByteCount(valueCount);
-        try {
-          stdSize = TypeHelper.getSize(metadata.getType()) * elementCount;
-        } catch (Exception e) {
-          // For unsupported types, just set stdSize to 0.
-          stdSize = 0;
-        }
+
+      switch(v.getField().getDataMode()) {
```
paul-rogers: Once code gets this complex, it may be time to create subclasses for the three modes. Plus subclasses for the categories of types (variable-width, map, etc.)

The column size would be a composition of the type size class plus a cardinality size class.

Maybe overkill for the current round of changes, but something to keep in mind if this code continues to grow in complexity.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -525,4 +763,11 @@ public VectorInitializer buildVectorInitializer() {
     }
     return initializer;
   }
+
+  public void allocateVectors(VectorContainer container, int recordCount) {
+    for (VectorWrapper w : container) {
+      ColumnSize colSize = columnSizes.get(w.getField().getName());
+      colSize.allocateVector(w.getValueVector(), recordCount);
+    }
+  }
```
paul-rogers: Very nice clean-up and enhancements. Now I don't feel so bad about this stuff being quick & dirty...

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/RepeatedMapVector.java
```diff
@@ -608,4 +608,22 @@ public void collectLedgers(Set<BufferLedger> ledgers) {
...
+  @Override
+  public int getPayloadByteCount(int valueCount) {
+    if (valueCount == 0) {
+      return 0;
+    }
+
+    int count = 0;
+
+    int entryCount = offsets.getAccessor().get(valueCount);
+    count += offsets.getPayloadByteCount(valueCount);
```
paul-rogers:
```
int count = offsets.get...
```

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -199,12 +422,18 @@ public String toString() {
...
+          .append(", std net size per entry: ")
+          .append(getStdNetSizePerEntry())
+          .append(", data size per entry: ")
+          .append(getDataSizePerEntry())
+          .append(", net size per entry: ")
+          .append(getNetSizePerEntry())
+          .append(", totalDataSize: ")
+          .append(getTotalDataSize())
+          .append(", totalNetSize: ")
+          .append(getTotalNetSize())
```
paul-rogers: Extra info is always good. This info is printed by the sort operator when debug logging is enabled. (It was invaluable when investigating issues found by QA tests.) Suggestion: find a way to compact the labels. Maybe:
```
Per entry: std size: xx, std net: xxx; Totals: data size: xxx, net size: xxx
```

Also, `totalDataSize` and `totalNetSize` should have spaces: they are labels, not variable names here.

ppadma:
@paul-rogers Thank you Paul. I will be incorporating some of your suggestions in future PRs. This change will allow us to build more things on top.

PR #1127, opened 2018-02-22
DRILL-6021:Show shutdown button when authentication is not enabled

dvjyothsna:
Display the shutdown button when authentication is not enabled since the user is by default the admin when the authentication is not enabled.


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -355,14 +356,16 @@ public ClusterInfo(Collection<DrillbitInfo> drillbits,
                      boolean userEncryption,
                      boolean bitEncryption,
                      boolean shouldShowAdminInfo,
-                     QueueInfo queueInfo) {
+                     QueueInfo queueInfo,
+                     boolean authEnabled ) {
```
arina-ielchiieva: Please remove space after `authEnabled` and before `)`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/DrillRoot.java
```diff
@@ -411,6 +415,8 @@ public String getCurrentVersion() {
   public boolean shouldShowAdminInfo() { return shouldShowAdminInfo; }
 
   public QueueInfo queueInfo() { return queueInfo; }
+
+  public boolean authEnabled() { return  authEnabled; }
```
arina-ielchiieva: Please rename `authEnabled` -> `isAuthEnabled`.

arina-ielchiieva:
+1

PR #1129, opened 2018-02-23
DRILL-6180: Use System Option "output_batch_size" for External Sort

ppadma:
External Sort has boot time configuration for output batch size "drill.exec.sort.external.spill.merge_batch_size" which is defaulted to 16M.
To make batch sizing configuration uniform across all operators, change this to use new system option that is added "drill.exec.memory.operator.output_batch_size". 
This option has default value of 32M. Changed it to 16M.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
@@ -19,6 +19,7 @@ package org.apache.drill.exec.physical.impl.xsort.managed;
 
 import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.exec.ExecConstants;
+import org.apache.drill.exec.server.options.OptionManager;
 
 public class SortConfig {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ExternalSortBatch.class);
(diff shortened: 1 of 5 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/SortConfig.java
```diff
@@ -71,8 +72,8 @@
 
   private final int mSortBatchSize;
 
-  public SortConfig(DrillConfig config) {
-
+  public SortConfig(FragmentContext context) {
+    DrillConfig config = context.getConfig();
```
paul-rogers: Suggestion: pass in the original `DrillConfig` plus an option manager rather than the fragment context. The suggestion minimizes undesired dependencies.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -421,7 +416,7 @@ drill.exec.options: {
     drill.exec.storage.implicit.fqn.column.label: "fqn",
     drill.exec.storage.implicit.suffix.column.label: "suffix",
     drill.exec.testing.controls: "{}",
-    drill.exec.memory.operator.output_batch_size : 33554432, # 32 MB
+    drill.exec.memory.operator.output_batch_size : 16777216, # 16 MB
```
paul-rogers: Thanks for making this adjustment.

ppadma:
@paul-rogers Made the change you suggested. Please take a look when you get a chance.

PR #1135, opened 2018-02-27
DRILL-6040: Added usage for graceful_stop in drillbit.sh

dvjyothsna:


arina-ielchiieva:
+1, LGTM.

distribution/src/resources/drillbit.sh
```diff
@@ -45,7 +45,7 @@
 # configuration file. The option takes precedence over the
 # DRILL_CONF_DIR environment variable.
 #
-# The command is one of: start|stop|status|restart|run
+# The command is one of: start|stop|status|restart|run|graceful_stop
```
paul-rogers: This command will be typed by hand sometimes. Can we find a shorter command? Retire? Remove? Or, should stop be changed to do this, with a new kill for the "ungraceful" stop?

priteshm: not sure if this is critical, but other options to consider are "finish" or "drain".

priteshm: How about we move forward with "graceful_stop" and add a different keyword once we have finalized an option? We can have the same behavior for graceful_stop and the new command to be the same.

arina-ielchiieva: Sounds good to me. If case when we would like to rename existing command and we leave the old one as well for backward compatibility.

PR #1136, opened 2018-02-27
DRILL-6192: Drill is vulnerable to CVE-2017-12197

vladimirtkach:
Changed libpam4j version from 1.8-rev1 to 1.9-mapr


arina-ielchiieva:
Lib version should be renamed to `1.8-rev2` and sources should be published as well.

PR #1137, opened 2018-02-27
DRILL-6185: Fixed error while displaying system profiles via the Web-UI

kkhatua:
The bug lies in the absence of a text plan for profiles related to system queries:
e.g.
```
use dfs.tmp;
show tables;
alter session reset all;
```
This addresses that by ensuring an empty string is substituted and the tokenization is done correctly.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
@@ -331,7 +331,7 @@ public class ProfileWrapper {
 
   //Generates operator names inferred from physical plan
   private void generateOpMap(String plan) {
-    this.physicalOperatorMap = new HashMap<String,String>();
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -330,10 +332,16 @@ public boolean isOnlyImpersonationEnabled() {
   //Generates operator names inferred from physical plan
   private void generateOpMap(String plan) {
     this.physicalOperatorMap = new HashMap<String,String>();
```
arina-ielchiieva: Please replace with diamonds -> `new HashMap<>();`

arina-ielchiieva:
+1, LGTM.

arina-ielchiieva:
I meant that we parse text plan, obviously it was generated from the some object. For the future we may consider to create special plan object from initial one which will have structure suitable for Web UI and we won't need to parse plan string...

PR #1138, opened 2018-02-28
DRILL-4120: Allow implicit columns for Avro storage format

vvysotskyi:
Existing implementation of `AvroDrillTabl` does not allow dynamic columns discovering. `AvroDrillTable.getRowType()` method returns `RelDataTypeImlp` instance with the list of all table columns. It forces validator to check columns from select list in `RowType` list. It makes impossible to use implicit columns.

This fix replaces the usage of `AvroDrillTable` by `DynamicDrillTable` for Avro format and also allows usage of non-existent columns in Avro tables to be consistent with other storage formats.


arina-ielchiieva:
You are basically reverting changes done in DRILL-3810 to support schema validation in Avro. 
Avro format is strict and has schema. Should Drill treat it the same way or do loosen parsing?

We should evaluate the option of leaving schema for avro but adding implicit columns. Maybe the change won't be as easy as changing `AvroDrillTable` to `DynamicDrillTable` but it might be more correct.

You can also start mailing thread on dev / user list, asking about treating avro as dynamic format (listing pros and cons) and get feedback from the users. 

[1] https://issues.apache.org/jira/browse/DRILL-3810

paul-rogers:
General comment: if we could move to the new scan framework; it handles implicit columns for all file-based readers. It also handles projection, missing columns, etc...

exec/java-exec/src/test/java/org/apache/drill/exec/store/avro/AvroFormatTest.java
```diff
@@ -170,25 +169,35 @@ public void testSimplePrimitiveSchema_SelectColumnSubset() throws Exception {
...
+      .baselineColumns("h_boolean", "e_dummy2")
+      .baselineValues(true, null)
+      .go();
+  }
+
+  @Test
+  public void testImplicitColumnFilename() throws Exception {
+    final String file = generateSimplePrimitiveSchema_NoNullValues(1).getFileName();
+    testBuilder()
+      .sqlQuery("select filename from dfs.`%s`", file)
```
arina-ielchiieva: Please test all implicit columns and at least one partition column.

vvysotskyi: Thanks for pointing this, modified existing test to check except the `filename` also `suffix`, `fqn` and `filepath` implicit columns. Added separate test for partition column.

exec/java-exec/src/main/java/org/apache/drill/exec/store/avro/AvroRecordReader.java
```diff
@@ -154,6 +156,12 @@ public int next() {
 
       writer.setValueCount(recordCount);
 
+      // adds fields which don't exist in the table but should be present in the schema
+      if (recordCount > 0) {
+        JsonReaderUtils.ensureAtLeastOneField(writer, getColumns(), false,
```
paul-rogers: In general, this is a bad idea, though existing code does this. If we find an empty file in one scanner, but a real file in another, we create an unnecessary schema change by making up a column.

Jinfeng's changes last year are supposed to handle the "fast none" case of a reader with no rows. There should be no reason to add a dummy column. Old code that adds such a column should be fixed. IMHO, code that does not add dummy columns should not begin to do so.

exec/java-exec/src/main/java/org/apache/drill/exec/store/avro/AvroRecordReader.java
```diff
@@ -295,7 +301,8 @@ private void processPrimitive(final Object value, final Schema.Type type, final
         writer.binary(fieldName).writeVarBinary(0, length, buffer);
         break;
       case NULL:
-        // Nothing to do for null type
+        // The default Drill behaviour is to create int column
+        writer.integer(fieldName);
```
paul-rogers: This maps a NULL type to integer. Probably OK if we do this consistently.

paul-rogers:
As @arina-ielchiieva points out, this change backs out plan-time knowledge of schema. This may not affect run-time accuracy. However, it does mean that queries can be planned, based on not knowing types, that fail at runtime when types are learned. This seems more like a bug that a feature. In general, we should use all information available. It is not helpful to ignore information if doing so results in poorer user experience.

paul-rogers:
Another thought. The removed code is at plan time. Did the original code have to open each file to retrieve schema? If so, does removing the code remove that load? If so, then this change could be a huge performance improvement if avoids the need to open every file in the Foreman.

Then, the the next question is: do we actually do anything with the plan-time type information? Few files have that information. Given that, does the planner actually use the information? Is this something we get for free from Calcite? If we are not using the type information at plan time, then clearly there is no harm in removing the code that retrieves the type information.

vvysotskyi:
@paul-rogers, schema is taken from the first file in the `FormatSelection`. Therefore for the case, when we have a table with several files with a different scheme, Drill query will fail.

As for the plan-time type information, besides the validation at the stage when a query is converted into rel nodes, field list may be used in project rel nodes instead of the dynamic star for `DynamicDrillTable`.

vvysotskyi:
@arina-ielchiieva, @paul-rogers, I have reworked this pull request to use `AvroDrillTable`, as it was before my changes. Also, with `AvroDrillTable` there is no need to make changes in `AvroRecordReader`, so I reverted them.
Could you please take a look again?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -277,6 +279,25 @@ public String deriveAlias(
       return SqlValidatorUtil.getAlias(node, ordinal);
     }
 
+    /**
+     * Checks that specified expression is not implicit column and
+     * adds is to a select list, ensuring that its alias does not
+     * clash with any existing expressions on the list.
+     */
+    @Override
+    protected void addToSelectList(
```
arina-ielchiieva: Could you please add comment explaining when this method is used and how know when to add columns to Avro table and not for Dynamic?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/types/ExtendableRelDataTypeHolder.java
```diff
@@ -0,0 +1,82 @@
...
+
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rel.type.RelDataTypeFieldImpl;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+import java.util.List;
+
+/**
+ * Holder for list of RelDataTypeField which may be expanded by partition or implicit columns.
```
arina-ielchiieva: When this holder is extended with partition columns? I can see only implicit.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
vvysotskyi: Added comment into Javadoc, which describes when and why this method is used.

As for Avro and Dynamic tables, the key point is the result of `RelDataType.isDynamicStruct()` method. 
For `AvroDrillTable` we use `ExtendableRelDataType` whose `isDynamicStruct()` method returns `false`, but for `DynamicDrillTable` we use `RelDataTypeDrillImpl`  whose `isDynamicStruct()` method returns `true`. 
In such way, Calcites `SqlValidatorImpl` determines whether columns should be added.

paul-rogers:
@vvysotskyi, thanks for addressing the schema issues!

arina-ielchiieva:
+1. LGTM.

PR #1139, opened 2018-02-28
DRILL-6189: Security: passwords logging and file permisions

vladimirtkach:
1. Overrided serialization methods for instances with passwords
2. Changed file permissions for configuration files


contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcStorageConfig.java
```diff
@@ -17,13 +17,15 @@
...
 
+import com.fasterxml.jackson.annotation.JsonFilter;
 import org.apache.drill.common.logical.StoragePluginConfig;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 
 @JsonTypeName(JdbcStorageConfig.NAME)
+@JsonFilter("passwordFilter")
```
arina-ielchiieva: Please explain how this works?

logical/src/main/java/org/apache/drill/common/config/LogicalPlanPersistence.java
```diff
@@ -52,6 +53,7 @@ public LogicalPlanPersistence(DrillConfig conf, ScanResult scanResult) {
     mapper.configure(Feature.ALLOW_UNQUOTED_FIELD_NAMES, true);
     mapper.configure(JsonGenerator.Feature.QUOTE_FIELD_NAMES, true);
     mapper.configure(Feature.ALLOW_COMMENTS, true);
+    mapper.setFilterProvider(new SimpleFilterProvider().setFailOnUnknownId(false));
```
arina-ielchiieva: Will filtering passwords work when profiles are sent between nodes (i.e. when we have several major fragments)?

protocol/src/main/java/org/apache/drill/exec/proto/UserProtos.java
```diff
@@ -5798,6 +5798,34 @@ public static UserToBitHandshake getDefaultInstance() {
     public UserToBitHandshake getDefaultInstanceForType() {
       return defaultInstance;
     }
+    public String safeLogString() {
```
arina-ielchiieva: You cannot add custom methods to proto buffers. Also consider using tabs instead of multiple spaces.
Please add to Jira example how log files looked before your changes and after.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -91,6 +91,34 @@
   static {
     userConnectionMap = new ConcurrentHashMap<>();
   }
+  public static  String safeLogString(UserToBitHandshake inbound) {
```
arina-ielchiieva: 1. Please remove one space -> `static  String`/
2. Can this method be just private? Not public static? If yes, please move it to the end of the class.
3. Please add javadoc to the method.
4. Please consider method renaming to depict actual work it does.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -91,6 +91,34 @@
...
+    StringBuilder sb = new StringBuilder();
+    sb.append("rpc_version: ");
+    sb.append(inbound.getRpcVersion());
+    sb.append("\ncredentials:\n\t");
+    sb.append(inbound.getCredentials());
+    sb.append("properties:");
+    java.util.List<Property> props = inbound.getProperties().getPropertiesList();
```
arina-ielchiieva: Why do you need full import?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -91,6 +91,34 @@
...
+    for (Property p: props){
+      if(!p.getKey().equalsIgnoreCase("password")) {
```
arina-ielchiieva: Please add spaces missing spaces...

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -320,7 +348,7 @@ protected void consumeHandshake(ChannelHandlerContext ctx, UserToBitHandshake in
 
       @Override
       public BitToUserHandshake getHandshakeResponse(UserToBitHandshake inbound) throws Exception {
-        logger.trace("Handling handshake from user to bit. {}", inbound);
+        logger.trace("Handling handshake from user to bit. {}", safeLogString(inbound));
```
arina-ielchiieva: Should we add `if (logger.isTraceEnabled()) {`? so `safeLogString` will be called only when we do need it for trace?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -158,7 +162,9 @@ protected void logAndSetTextPlan(final String description, final Prel prel, fina
 
   protected void log(final String name, final PhysicalPlan plan, final Logger logger) throws JsonProcessingException {
     if (logger.isDebugEnabled()) {
-      String planText = plan.unparse(context.getLpPersistence().getMapper().writer());
+      PropertyFilter theFilter = new SimpleBeanPropertyFilter.SerializeExceptFilter(Sets.newHashSet("password"));
```
arina-ielchiieva: Please rename to `filter`.

contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcStorageConfig.java
vladimirtkach: To apply filter:
1) Mark the entity with you want to filter out fields from.
2) Create filter provider and register property filter with reference to your entity
3) When creating ObjectWriter pass your filter provider

vladimirtkach:
@arina-ielchiieva made changes, please take a look

logical/src/main/java/org/apache/drill/common/config/LogicalPlanPersistence.java
vladimirtkach: submitted physical plan directly to node, it was successfully deserialized

arina-ielchiieva:
+1, LGTM.

PR #1141, opened 2018-02-28
DRILL-6197: Skip duplicate entry for OperatorStats

kkhatua:
`org.apache.drill.exec.ops.FragmentStats` should skip injecting the `org.apache.drill.exec.ops.OperatorStats` instance for these operators:
```
org.apache.drill.exec.proto.beans.CoreOperatorType.SCREEN
org.apache.drill.exec.proto.beans.CoreOperatorType.SINGLE_SENDER
org.apache.drill.exec.proto.beans.CoreOperatorType.BROADCAST_SENDER
org.apache.drill.exec.proto.beans.CoreOperatorType.HASH_PARTITION_SENDER
```
They all use the `org.apache.drill.exec.physical.impl.BaseRootExec` to inject the correct statistics.
**NOTE:**
To avoid going out of sync for new operators' `...Exec` , the BaseRootExec does the swapping of the matching `OperatorStats` instead of having `FragmentStats` perform skipping.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentStats.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentStats.java
@@ -17,13 +17,14 @@
  */
 package org.apache.drill.exec.ops;
 
-import java.util.List;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Map.Entry;
 
+import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.proto.CoordinationProtos.DrillbitEndpoint;
 import org.apache.drill.exec.proto.UserBitShared.MinorFragmentProfile;
-
-import com.google.common.collect.Lists;
 
 /**
  * Holds statistics of a particular (minor) fragment.
@@ -31,7 +32,7 @@ import com.google.common.collect.Lists;
 public class FragmentStats {
 //  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FragmentStats.class);
 
-  private List<OperatorStats> operators = Lists.newArrayList();
+  private Map<ImmutablePair<Integer, Integer>, OperatorStats> operators = new LinkedHashMap<>();
   private final long startTime;
   private final DrillbitEndpoint endpoint;
   private final BufferAllocator allocator;
@@ -47,8 +48,8 @@ public class FragmentStats {
     prfB.setMaxMemoryUsed(allocator.getPeakMemoryAllocation());
     prfB.setEndTime(System.currentTimeMillis());
     prfB.setEndpoint(endpoint);
-    for(OperatorStats o : operators){
-      prfB.addOperatorProfile(o.getProfile());
+    for(Entry<ImmutablePair<Integer, Integer>, OperatorStats> o : operators.entrySet()){
+      prfB.addOperatorProfile(o.getValue().getProfile());
     }
   }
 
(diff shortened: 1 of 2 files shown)


kkhatua:
@amansinha100  please review

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentStats.java
```diff
@@ -31,6 +32,13 @@
 public class FragmentStats {
 //  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FragmentStats.class);
 
+  //Skip operators that already have stats reported by org.apache.drill.exec.physical.impl.BaseRootExec
+  private static final List<Integer> operatorStatsInitToSkip = Lists.newArrayList(
```
amansinha100: This could get out of sync with the types of senders that extend BaseRootExec.

kkhatua: The add-on commit refactors by having the BaseRootExec constructor handle the substition without risking going out of sync with other senders extending BaseRootExec.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentStats.java
```diff
@@ -79,4 +71,21 @@ public void addOperatorStats(OperatorStats stats) {
     operators.add(stats);
   }
 
+  //DRILL-6197
+  public OperatorStats addOrReplaceOperatorStats(OperatorStats stats) {
+    //Remove existing stat
+    OperatorStats replacedStat = null;
+    int index = 0;
+    for (OperatorStats opStat : operators) {
```
amansinha100: I am worried about the small overheads of doing this linear search adding up for each operator, especially for queries with complex query plans.  The stats collection should ideally impose minimal overhead.  Does the operator stats have to be a list or can just use a Set ?

amansinha100: Some TPC-DS queries have fairly long list of operators within a fragment and in general it would be preferable to not do this search.  Can you point to where this Json serialization happens ?  my guess is it just needs to preserve the insertion order.  In that case we could use a LinkedHashSet which would provide both the duplicate removal and keep insertion order.

kkhatua: I see your point. Also, digging into the code shows I can substitute with a LinkedHashMap, since the list is only referenced here for consumption of its contents:
https://github.com/kkhatua/drill/blob/65efe3ea0c5777490488d3d56cbdb0cb011b9f33/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentStats.java#L45

I can't use a Set, because I need the Stats object hashed on the operator ID & Type, and not the rest of the contents. I'll refactor and try to confirm nothing else breaks.

kkhatua: I added a new commit, but I haven't tested it for performance. Can you take a look, @amansinha100 ?

amansinha100: LGTM.  Hopefully it did not break existing stuff..so will wait for your confirmation.

kkhatua: Everything worked fine. Tried doing a join for TPCH tables - `lineitem` and `orders`, and confirmed no more duplicates for SCREEN, SINGLE_SENDER and HASH_PARTITION_SENDER. For a smaller substitute of `orders` with `supplier` ; confirmed that the BROADCAST_SENDER was also not having duplicates..

amansinha100:
+1.

PR #1142, opened 2018-03-01
DRILL-6198: OpenTSDB unit tests fail when Lilith client is run

vvysotskyi:
Added method which checks that the default port 10_000 is free, otherwise this method increases port number and checks until free port is found.

### DIFF
--- a/contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
+++ b/contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
@@ -21,13 +21,18 @@ import com.github.tomakehurst.wiremock.junit.WireMockRule;
 import org.apache.drill.PlanTestBase;
 import org.apache.drill.common.exceptions.UserRemoteException;
 import org.apache.drill.exec.store.StoragePluginRegistry;
-import org.apache.drill.exec.store.openTSDB.OpenTSDBStoragePlugin;
 import org.apache.drill.exec.store.openTSDB.OpenTSDBStoragePluginConfig;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.net.BindException;
+import java.net.ServerSocket;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.github.tomakehurst.wiremock.client.WireMock.equalToJson;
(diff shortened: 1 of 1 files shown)


contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
```diff
@@ -51,17 +54,17 @@
 
 public class TestOpenTSDBPlugin extends PlanTestBase {
 
-  protected static OpenTSDBStoragePlugin storagePlugin;
-  protected static OpenTSDBStoragePluginConfig storagePluginConfig;
+  private static int portNum = 10_000;
```
arina-ielchiieva: Why do you set value right away? It looks you will always re-write in `setup`.

contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
```diff
@@ -51,17 +54,17 @@
...
 
   @Rule
-  public WireMockRule wireMockRule = new WireMockRule(10000);
+  public WireMockRule wireMockRule = new WireMockRule(portNum);
 
   @BeforeClass
   public static void setup() throws Exception {
+    portNum = getFreePortNum(portNum, 1000);
```
arina-ielchiieva: May be we can decrease number of attempt to 200?

contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
```diff
@@ -185,4 +188,26 @@ public void testDescribe() throws Exception {
...
+   * Otherwise, increases port number and checks until free port is found
+   * or the number of attempts is reached specified numAttempts
+   *
+   * @param portNum     initial port number
+   * @param numAttempts max number of attempts to find port with greater number
+   * @return free port number
+   * @throws BindException if free port was not found and all attempts were used.
+   */
+  private static int getFreePortNum(int portNum, int numAttempts) throws IOException {
+    while (numAttempts > 0) {
```
arina-ielchiieva: 1. Please re-write using for loop.
2. Please add more details to the exception, include initial port number, which ports were occupied. Suggest to check which ports are free etc.

contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
vvysotskyi: Thanks, removed.

contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
vvysotskyi: Done.

contrib/storage-opentsdb/src/test/java/org/apache/drill/store/openTSDB/TestOpenTSDBPlugin.java
vvysotskyi: 1. Thanks, it looks better with for loop.
2. Added more details to the error message.

arina-ielchiieva:

PR #1144, opened 2018-03-01
DRILL-6202: Deprecate usage of IndexOutOfBoundsException to re-alloc vectors

vrozov:


vrozov:
@parthchandra Please take a look.

parthchandra:
AFAIK, IndexOutOfBoundsException is thrown by Netty when a write to a DrillBuf goes out of bounds (see https://github.com/netty/netty/blob/netty-4.0.48.Final/buffer/src/main/java/io/netty/buffer/AbstractByteBuf.java#L1123  and https://github.com/netty/netty/blob/netty-4.0.48.Final/buffer/src/main/java/io/netty/buffer/UnsafeByteBufUtil.java#L283 ).
This is independent of the bounds check in Drill which may be disabled.

vrozov:
Is there a reason to delegate `get/setBytes` to `AbstractByteBuf`? If not, this PR will be a preparation step to use `PlatformDependent` directly bypassing Netty bounds checking.

parthchandra:
To be honest, I don't know. There are too many paths available to write to direct memory, some which were developed over time and so not all vector code may be using them.
Ideally, every bit of code should go thru DrillBuf which wraps a UDLE (UnsafeDirectLittleEndian). UDLE has one level of bounds check (which I feel is necessary) and uses PlatformDependent directly.
Vector set safe methods provides vector level bounds checking and the vector set methods bypass bounds checking. ResultSetLoader methods provide guarantees that may make the set safe methods redundant.
Our goal should be to not depend on the set safe methods except for debugging.
What exactly are you thinking of when proposing using PlatformDependent directly.

vrozov:
It is not clear why get/set Byte/Char/Short/Int/Long/Float/Double do not delegate to UDLE, while get/set Bytes delegates to UDLE and relies on netty 'AbstractByteBuf` for bounds checking. IMO, it will be good to have the behavior consistent for all methods.

In many cases including `VariableLengthVectors`, there is no need to rely on UDLE boundary checking as a caller already provides or can provide a guarantee that an index is within a buffer boundaries. In those cases, boundary check becomes an extra cost. IMO, it will be good to have a consistent behavior with ability to enable bounds checking for debugging.

sachouche:
I have added a comment within this PR associated JIRA: [DRILL-6202](https://issues.apache.org/jira/browse/DRILL-6202)

vrozov:
IMO, this PR or DRILL-6202 is not the best place to discuss boundary checking as the PR and JIRA deals with `IndexOutOfBoundsException` but does not change how DrillBuf, Vector or Operators ensure that reads and writes are done within proper boundaries. I'll bring that to dev@drill. Please repeat your arguments on the mailing list.

paul-rogers:
My two cents... DrillBuf is the only memory-level abstraction that (low level) Drill code should reference. The UDLE and other bits should be fully encapsulated. This guideline lets us evolve the representation if we ever need to do so.

The original design appeared to be that value vectors would be the primary interface to memory. But, a great many issues made that difficult, not least of which is that vector access methods are heavily typed, resulting in far too much casting. Also, the mutator methods try to do the full operation, leading to inefficiency (especially around VarChars).

A more general rule is that application code should work with vectors until they can migrate to working with the result set loader or reader. (We should probably call these the row set emitter and collector to be more Hadoop-like...) The higher-level abstractions handle the grunt work currently spread throughout operators.

(And, to answer a prior question: we want to use the row set abstractions so we have a uniform way to write to vectors, to control batch size, to handle schema issues and so on on write. And, to have a standard way to handle indirection vectors and vector navigation on read.)

Ideally only, the vector mutator or row set loader implementation works with DrillBuf to do actual data reads and writes. In an early version, the row set loader code used `PlatformDependent` to avoid bounds checks. But, with @vrozov's improvements, doing so became unnecessary -- a nice improvement.

Still, bounds checks should be done during tests: it is handy to work with a safety net.

Since bounds checks are optional (turned off in production), then the changes here make good sense: no code should count on bounds checks from the "unchecked" methods for the simple reason that the checks are normally off.

That said, if there is a reason to have "checked" access, we could provide such methods. Those methods would throw the `IndexOutOfBoundsException`.  That is, the checked methods would recreate the original "get/set" methods prior to @vrozov's improvements. I can't think of a reason to do that off the top of my head, but someone might present a valid use case.

exec/memory/base/src/main/java/io/netty/buffer/DrillBuf.java
```diff
@@ -777,23 +778,20 @@ public int getActualMemoryConsumed() {
...
+    Preconditions.checkArgument(start >= 0);
+    final StringBuilder sb = new StringBuilder("buffer byte dump");
+    final int end = Math.min(length, this.length - start);
+    for (int i = 0; i < end; i++) {
+      if (i % LOG_BYTES_PER_ROW == 0) {
+        sb.append(String.format("%n [%05d-%05d]", i + start, Math.min(i + LOG_BYTES_PER_ROW - 1, end - 1) + start));
       }
-      sb.append('\n');
+      byte b = _getByte(i + start);
+      sb.append(" 0x").append(HEX_CHAR[b >> 4]).append(HEX_CHAR[b & 0x0F]);
```
paul-rogers: Is this complexity needed relative to the original `format(" 0x02x", b)` implementation? Is this code performance sensitive enough to justify the extra complexity?

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -534,15 +534,11 @@ public void setSafe(int index, byte[] bytes) {
...
       final int currentOffset = offsetVector.getAccessor().get(index);
-      offsetVector.getMutator().setSafe(index + 1, currentOffset + bytes.length);
-      try {
-        data.setBytes(currentOffset, bytes, 0, bytes.length);
-      } catch (IndexOutOfBoundsException e) {
-        while (data.capacity() < currentOffset + bytes.length) {
-          reAlloc();
-        }
-        data.setBytes(currentOffset, bytes, 0, bytes.length);
+      while (data.capacity() < currentOffset + bytes.length) {
```
paul-rogers: If we compare this implementation to the one from, say, a year or 18 months ago, I think we're back where we started.

This would actually be a good use case for a "checkedSetBytes" method that does the bounds checks. Still, whether done here or in the underlying method, the `if` statement is needed, so no savings in using a "checked" method,

paul-rogers: One trick used in the row set reader code is to avoid redundant re-allocs by computing the new size and rounding to a power of two. This is handy at the start with the vector doubles from 256 to 512 to 1K and the data size is, say, 600 bytes.

However, it is not clear that such an optimization is as necessary as it once was: the team's been doing a good job at adding up-front vector allocation where it was missing, reducing the gratuitous reallocs.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/BaseScalarWriter.java
```diff
@@ -211,7 +211,7 @@ protected boolean canExpand(int delta) {
 
   protected void overflowed() {
     if (listener == null) {
-      throw new IndexOutOfBoundsException("Overflow not supported");
+      throw new UnsupportedOperationException("Overflow not supported");
```
paul-rogers: Maybe change to `OversizedAllocationException`. See `FixedValueVectors.setInitialCapacity()`.

src/main/resources/checkstyle-config.xml
```diff
@@ -30,10 +30,15 @@
 
     <module name="NeedBraces"/>
 
+    <module name="IllegalCatch">
```
paul-rogers: Do we want to do this? Why are all these forbidden? `Throwable` was needed by @Ben-Zvi in his very first Drill bug fix. `RuntimeException` is the root of the unchecked exceptions beloved by Drill code and is used as a general "something went wrong" placeholder.

These are all commented out, but still...

Finally, `IndexOutOfBoundsException` is a perfectly fine exception. Adding this line is probably handy to catch all reference as part of this fix, but probably not helpful as a general policy.

src/main/resources/checkstyle-suppressions.xml
```diff
@@ -16,4 +16,13 @@
 <!--  Checkstyle Suppressions for Apache Drill -->
 <suppressions>
   <suppress files="[\\/]generated-sources[\\/]" checks="AvoidStarImport,NeedBraces"/>
+  <suppress checks="IllegalCatch"
```
paul-rogers: In ML this is called over-fitting: I think we're adding a bit too much complexity for too little value. I can just see myself, in a couple of months, tearing my hair out wondering why the build fails on some perfectly benign construct...

Maybe explain this rule in a comment?

exec/memory/base/src/main/java/io/netty/buffer/DrillBuf.java
vrozov: The goal is to avoid new String allocation for every output byte inside DrillBuf.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/BaseScalarWriter.java
vrozov: I also considered `OversizedAllocationException` but IMO `UnsupportedOperationException` is a better fit.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
parthchandra: I'm on the fence on this one. The change from 18 months ago resulted in a performance improvement of a few percent for variable length columns. This change essentially undoes that and takes us back to where we were. It seems that with the recent changes to batch sizing, we are on the path to eliminating the realloc behaviour that is part of the performance bottleneck for the setSafe methods. So, perhaps, it might be ok to let this one stay.

src/main/resources/checkstyle-config.xml
parthchandra: I think IOBE should be caught and converted to a UserException so that the end user does not get an incomprehensible error message.

vrozov: My point is that there should be no special handling for IOBE compared to NPE or CCE. If it is in process call to a Java library all run-time exceptions including NPE, IOBE and CCE should be propagated to the caller. If it is a remote call, runtime and checked exceptions need to be handled by propagating them to a remote caller as well.

What an end user tool displays to an end user depends on a tool. IMO, the best option is to have a configuration parameter that determines verbosity level. With default verbosity level, the tool may show a generic error message (possibly explaining how to increase verbosity level to get more info, see `maven -X`), and there should be an option to change the verbosity level and print a stack trace that initially triggered the error. Such stack trace is required when an end user logs a bug or asks for help on the dev list.

parthchandra:
Looks good. +1

PR #1145, opened 2018-03-02
DRILL-6187: Exception in RPC communication between DataClient/Control…

sohami:
…Client and respective servers when bit-to-bit security is on


sohami:
@vrozov - Please help to review this PR.
It address the concurrency issue during authentication of control/data client to server side. Rather than adding the connection into connection holder right after TCP connection is available, the listener for connection success is called after successful authentication (if needed).

exec/rpc/src/main/java/org/apache/drill/exec/rpc/BasicClient.java
```diff
@@ -69,6 +74,11 @@
   private final IdlePingHandler pingHandler;
   private ConnectionMultiListener.SSLHandshakeListener sslHandshakeListener = null;
 
+  // Authentication related parameters
+  protected volatile List<String> serverAuthMechanisms = null;
```
vrozov: Is `volatile` necessary? Consider making all variables private and providing `setAuthRequired(List<String> authMechanisms)`.

exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
```diff
@@ -312,6 +312,11 @@ public synchronized void connect(String connect, Properties props) throws RpcExc
     if (connected) {
       return;
     }
+
+    if (props == null) {
```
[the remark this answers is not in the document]
vrozov: My recommendation is to change other 2 overloaded methods to pass `new Properties()` instead of `null` and making it explicit that `null` is not allowed (avoid passing `null` and checking for `null` at the same time).

exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
```diff
@@ -371,17 +376,20 @@ protected void afterExecute(final Runnable r, final Throwable t) {
 
     while (triedEndpointIndex < connectTriesVal) {
       endpoint = endpoints.get(triedEndpointIndex);
+
+      // Set in both props and properties since props is passed to UserClient
+      if (!properties.containsKey(DrillProperties.SERVICE_HOST)) {
```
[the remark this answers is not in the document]
vrozov: - Use `put` and add `TODO` comment.

 - What is a reason not to use API available in the minimum supported version assuming that support for JDK 7 is dropped as part of https://issues.apache.org/jira/browse/DRILL-1491?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/control/ControlClient.java
```diff
@@ -103,6 +96,23 @@ protected void handle(ControlConnection connection, int rpcType, ByteBuf pBody,
     connection.getCurrentHandler().handle(connection, rpcType, pBody, dBody, sender);
   }
 
+  @Override
+  protected void prepareSaslHandshake(final RpcConnectionHandler<ControlConnection> connectionListener)
```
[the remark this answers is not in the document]
vrozov: It is defined in `protocol`. Consider adding the dependency on the `protocol`jar.

sohami:
@vrozov - Thanks for the feedback. Update the PR with latest changes. Please help to review.

exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillClient.java
```diff
@@ -96,6 +96,8 @@
 
 import io.netty.channel.EventLoopGroup;
 
+import javax.validation.constraints.NotNull;
```
vrozov: Please use java.annotation.Nonnull or org.jetbrains.annotations.NotNull (or do not annotate, it is inferred).

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/BitRpcUtility.java
```diff
@@ -0,0 +1,110 @@
...
+    } catch (final IOException e) {
+      logger.error("Failed while doing setup for starting sasl handshake for connection", connection.getName());
+      final Exception ex = new RpcException(String.format("Failed to initiate authentication to %s",
+        endpoint.getAddress()), e);
+      connectionHandler.connectionFailed(RpcConnectionHandler.FailureType.AUTHENTICATION, ex);
+    }
+  }
+
+  // Suppress default constructor
+  private BitRpcUtility() throws Exception {
```
vrozov: Exception is not required, just make the constructor private.

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/BitRpcUtility.java
```diff
@@ -0,0 +1,110 @@
...
+    throw new UnsupportedOperationException("BitRpcUtility is a utility class and should not be instantiated.");
+  }
+}
```
vrozov: add LF.

sohami:
Made all the changes and squashed changes into 2 commits

parthchandra:
+1

PR #1148, opened 2018-03-02
DRILL-5994: enable configuring number of Jetty acceptors and selectors (default to 1 acceptor and 2 selectors)

vrozov:
@arina-ielchiieva @ilooner @MitchelLabonte Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
@@ -111,8 +111,6 @@ public class WebServer implements AutoCloseable {
 
   private final Drillbit drillbit;
 
-  private int port;
-
   /**
    * Create Jetty based web server.
    *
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -154,35 +152,31 @@ public void start() throws Exception {
...
       try {
         embeddedJetty.start();
+        return;
       } catch (BindException e) {
         if (portHunt) {
-          int nextPort = port + 1;
-          logger.info("Failed to start on port {}, trying port {}", port, nextPort);
-          port = nextPort;
-          embeddedJetty.stop();
+          logger.info("Failed to start on port {}, trying port {}", port, ++port, e);
```
ilooner: Originally I stopped the jetty server in the event that a BindException was throw since it looks like jetty starts it's beans before it starts its connectors (see org.eclipse.jetty.server.Server#doStart). Since the thread pool is treated like a bean by Jetty this means the ThreadPool is started before the connectors are started. So if a connector fails with a BindException then we will have a leaked QueuedThreadPool. My understanding is if we want to release resources created during a failed start we would have to stop the jetty server after a failure.

vrozov: The QueuedThreadPool is now provided and will be reused. Please see line 159.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -424,10 +418,11 @@ private ServerConnector createHttpsConnector(int port) throws Exception {
    * @return Initialized {@link ServerConnector} instance for HTTP connections.
    * @throws Exception
    */
-  private ServerConnector createHttpConnector(int port) throws Exception {
+  private ServerConnector createHttpConnector(int port, int acceptors, int selectors) throws Exception {
     logger.info("Setting up HTTP connector for web server");
     final HttpConfiguration httpConfig = new HttpConfiguration();
-    final ServerConnector httpConnector = new ServerConnector(embeddedJetty, new HttpConnectionFactory(httpConfig));
+    final ServerConnector httpConnector =
+        new ServerConnector(embeddedJetty, null, null, null, acceptors, selectors, new HttpConnectionFactory(httpConfig));
```
ilooner: Why not
 
```
new ServerConnector(embeddedJetty, acceptors, selectors, new HttpConnectionFactory(httpConfig));
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -411,6 +404,7 @@ private ServerConnector createHttpsConnector(int port) throws Exception {
 
     // SSL Connector
     final ServerConnector sslConnector = new ServerConnector(embeddedJetty,
+        null, null, null, -1, -1,
```
ilooner: Why not use the configured values for HTTPS?

vrozov: good catch 👍

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
ilooner: I see thx

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
vrozov: I don't think there is such constructor (only one that takes `SslContextFactory` as the last argument), please double check.

PR #1151, opened 2018-03-06
DRILL-6151: Fragment executors may terminate without sending final batch to a downstream causing query to hang

vrozov:


arina-ielchiieva:
The reason of the change is that some other errors can be thrown. For example, `StackOverflowError`.
+1, LGTM.

PR #1153, opened 2018-03-07
DRILL-6044: Fixed shutdown button in Web UI when ssl,auth are enabled

dvjyothsna:
Fixed shutdown button to avoid cross domain requests since they are causing some issues when ssl and auth are enabled

### DIFF
--- a/exec/java-exec/src/main/resources/rest/index.ftl
+++ b/exec/java-exec/src/main/resources/rest/index.ftl
@@ -66,7 +66,7 @@
               <tr id="row-${i}">
                 <td>${i}</td>
                 <td id="address" >${drillbit.getAddress()}<#if drillbit.isCurrent()>
-                    <span class="label label-info" id = "current">Current</span>
+                    <span class="label label-info" id="current">Current</span>
                   </#if>
                 </td>
                 <td id="port" >${drillbit.getUserPort()}</td>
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -193,7 +192,11 @@
       var port = getPortNum();
       var timeout;
       var size = $("#size").html();
+      var host;
 
+      window.onload = function () {
+          host = location.host;
+      };
```
sohami: Please remove this `function()` and `var host`.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -269,9 +277,12 @@
...
       }
-      <#if model.shouldShowAdminInfo() || !model.isAuthEnabled() >
-          function shutdown(address,button) {
-              url = "http://"+address+":"+portNum+"/gracefulShutdown";
+       <#if model.shouldShowAdminInfo() || !model.isAuthEnabled()>
+          function shutdown(button) {
+              var protocol = location.protocol;
+              var host = location.host;
+              var requestPath = "/gracefulShutdown";
+              var url = protocol+host+requestPath;
```
sohami: How about refactoring above lines into separate function `getRequestUrl(requestPath)` and call that from both the methods (`shutdown` & `fillQueryCount`)

```
function getRequestUrl(requestPath) {
    var protocol = location.protocol;
    var host = location.host;
    var url = protocol + host + requestPath;
    return url;
}
```

dvjyothsna:
@sohami Please review

sohami:
+1 LGTM. Thanks for making the changes.

parthchandra:
+1

PR #1155, opened 2018-03-07
DRILL-6213: During restart drillbit should be killed forcefully if ex…

arina-ielchiieva:
…ceeds allowed timeout


sohami:
+1 LGTM.

PR #1157, opened 2018-03-07
DRILL-6216: Metadata mismatch when connecting to a Drill 1.12.0 with …

sohami:
…a Drill-1.13.0-SNAPSHOT driver


amansinha100:
+1

PR #1158, opened 2018-03-09
DRILL-6145: Implement Hive MapR-DB JSON handler

vdiravka:
@sohami, @vrozov Please review

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -324,6 +324,14 @@
           <groupId>org.apache.hbase</groupId>
           <artifactId>hbase-protocol</artifactId>
         </dependency>
+        <dependency>
+          <groupId>org.apache.hive</groupId>
+          <artifactId>hive-maprdb-json-handler</artifactId>
+        </dependency>
+        <dependency>
+          <groupId>com.mapr.db</groupId>
+          <artifactId>maprdb-mapreduce</artifactId>
+        </dependency>
       </dependencies>
       <build>
       </build>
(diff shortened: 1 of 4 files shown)


distribution/pom.xml
```diff
@@ -324,6 +324,14 @@
           <groupId>org.apache.hbase</groupId>
           <artifactId>hbase-protocol</artifactId>
         </dependency>
+        <dependency>
```
vrozov: Should the dependency be part of hive module (even if it is not part of compile and is a run-time dependency)?

vdiravka: I did it at first time, but jars are downloaded usually only for dependencies from common part of pom file.
Looks like to add these dependencies into mar profile of drill-hive module, it is necessary to add maven-assembly-plugin to mapr profile of storage-hive module or to edit distribution pom file somehow to leverage dependencies from profiles section. I will investigate it.

vrozov: maven dependency handling is indeed somewhat inconsistent in regards to profile. Check maven 3.5 or above. AFAIK it (maven reactor) handles profile dependencies better compared to prior versions.

vdiravka: I have tried maven 3.5.3 version (I used 3.5.0 by default for couple of month), but nothing is changed.
I have also tried the newest `maven-assembly-plugin`, but no luck too. 

So I have added dependencies for `storage-hive/core` module for `mapr` profile without removing them from `distribution` module. I believe that it is better than adding a new module to storage-hive or adding extra assembly plugins.
But to create a JIra to investigate it in future can be a good point here.

priteshm:
@vdiravka please do link the JIRA that you will create.
@vrozov can you take a final look?

vdiravka:
@priteshm @priteshm I have created a Jira for above mentioned issue: [DRILL-6258](https://issues.apache.org/jira/browse/DRILL-6258)

arina-ielchiieva:
+1

PR #1159, opened 2018-03-10
DRILL-6215: Changed Statement to PreparedStatement in JdbcRecordReader

khurram-faraaz:
@bitblender @kkhatua  Can you please review ?
JDBC unit tests ran clean for me.

### DIFF
--- a/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcRecordReader.java
+++ b/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcRecordReader.java
@@ -25,6 +25,7 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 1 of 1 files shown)


kkhatua:
@kfaraaz Are there unit tests specific to the JDBC Storage Plugin? They're not the same as the JDBC unit tests, which test the Drill JDBC driver.

While the change seems straightforward, I'm not sure if all JDBC drivers support PreparedStatement. For e.g., within Drill, the PreparedStatement is executed as a standard Statement object, which is why this _works_ functionally.

What happens if a JDBC driver backing the JDBC storage plugin does not support PreparedStatement (i.e. it is a No-Op)?

kfaraaz:
I don't know about the other file, I didn't add it. Let me check.

Thanks,
Khurram

arina-ielchiieva:
@kfaraaz did you have a chance to address issue with second file?

vvysotskyi:
Fixed in https://github.com/apache/drill/commit/4f55e71dc971d42054a031acd000ddf8337e90d9

PR #1160, opened 2018-03-11
DRILL-6224: Publish current memory usage for Drillbits in webpages

kkhatua:
The `metrics.ftl` page had gauges incorrectly set to near zero values. The commit for metrics.ftl fixes that, and also provides an estimate of the current direct memory actively in use (based on the `drill.allocator.root.used` value reported by the Drillbit)
The `index.ftl` page expands this by additionally providing the active heap and direct memory usage for all Drillbits. This is done by the JavaScript actively pinging all the Drillbits on their current memory usage in 10 second intervals.
The percentages in both cases indicate the % usage in terms of the max (heap) or the peak (direct) observed.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
+++ b/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
@@ -43,6 +43,11 @@
         <div id="totalUsage" class="progress-bar" role="progressbar" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100" style="width: 50%;">
         </div>
       </div>
+      Direct (Estimate)
+      <div class="progress">
+        <div id="estDirectUsage" class="progress-bar" role="progressbar" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100" style="width: 50%;">
+        </div>
+      </div>
     </div>
 
     <div id="mainDiv" class="col-md-9" role="main">
@@ -109,18 +114,23 @@
     };
 
     function updateBars(gauges) {
-      $.each(["heap","non-heap","total"], function(i, key) {
+      $.each(["heap","non-heap","total","drill.allocator.root"], function(i, key) {
         var used    = gauges[key + ".used"].value;
-        var max     = gauges[key + ".max"].value;
-        var usage   = round((used / 1073741824), 2) + "GB";
-        var percent = round((used / max), 2);
+        var max;
+        if (key.startsWith("drill.allocator")) {
+          max       = gauges[key + ".peak"].value;
+        } else {
+          max       = gauges[key + ".max"].value;
+        }
+        var percent = round((100 * used / max), 2);
+        var usage   = round((used / 1073741824), 2) + "GB (" + Math.max(0, percent) + "%)";
 
-        var styleVal = "width: " + percent + "%;"
-        $("#" + key + "Usage").attr({
+        var styleVal = "width: " + percent + "%;color: #202020;white-space: nowrap"
+        $("#" + (key.startsWith("drill.allocator") ? "estDirect" : key) + "Usage").attr({
           "aria-valuenow" : percent,
           "style" : styleVal
         });
-        $("#" + key + "Usage").html(usage);
+        $("#" + (key.startsWith("drill.allocator") ? "estDirect" : key) + "Usage").html(usage);
       });
     };
 
@@ -159,7 +169,7 @@
     };
 
     update();
-    setInterval(update, 2000);
+    setInterval(update, 3000);
   </script>
 </#macro>
 
--- a/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
+++ b/exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
@@ -43,7 +43,7 @@
         <div id="totalUsage" class="progress-bar" role="progressbar" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100" style="width: 50%;">
         </div>
       </div>
-      Direct (Estimate)
+      Actively Used Direct (Estimate)
       <div class="progress">
         <div id="estDirectUsage" class="progress-bar" role="progressbar" aria-valuenow="50" aria-valuemin="0" aria-valuemax="100" style="width: 50%;">
         </div>
@@ -82,6 +82,15 @@
     var round = function(val, n) {
       return Math.round(val * Math.pow(10, n)) / Math.pow(10, n);
     };
+    var isAllocator = function(metricName) {
+      return (metricName.startsWith("drill.allocator"));
+    };
+    function getGBUsageText(val, perc) {
+      if (isNaN(perc)) {
+        perc = 0;
+      }
+      return round((val / 1073741824), 2) + "GB (" + Math.max(0, perc) + "%)";
+    }
 
     function updateGauges(gauges) {
       $("#gaugesTable").html(function() {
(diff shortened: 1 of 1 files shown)


kkhatua:
@parthchandra / @arina-ielchiieva  would you guys want to review this? 
Here is a preview of how the details appear on the default page:
image

parthchandra:
Looks wrong dunnit? Direct memory is (still?) showing as zero?

kkhatua:
I think I need to put up some tooltips.

We're getting the metrics from the `drill.allocator.root.[peak|used]` values. I haven't seen a metric that shows the actual Direct allocation from the system. What Drill seems to report is the current **actively** used Direct memory. So, when running queries on the cluster, these numbers move up. The percentage for Direct is the current usage as a percent of the peak memory that was reported by the root allocator.

Check out any active Drillbit's `#metrics` page. The `drill.allocator.*` entries are what I was referring to. JVM itself doesn't appear to expose the Direct memory information readily.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -53,11 +53,16 @@
...
               <th>#</th>
               <th>Address</th>
+              <th>Heap Usage</th>
+              <th>Direct Usage</th>
               <th>User Port</th>
               <th>Control Port</th>
               <th>Data Port</th>
               <th>Version</th>
               <th>Status</th>
+              <#if model.shouldShowAdminInfo() || !model.isAuthEnabled() >
```
arina-ielchiieva: On master we have different line `<#if (model.shouldShowAdminInfo() || !model.isAuthEnabled()) && drillbit.isCurrent() >`

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -291,6 +300,46 @@
...
+              var address = $("#row-"+i).find("#address").contents().get(0).nodeValue.trim();
+              updateMemory(address, i);
+          }
+      }
+
+      //Update memory
+      function updateMemory(drillbit,idx) {
+        var result = $.ajax({
+          type: 'GET',
+          url: "http://"+drillbit+":8047/status/metrics",
```
arina-ielchiieva: You can not hard-code port, it can be changed using property. See how graceful shutdown is checking drillbits statuses.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -291,6 +300,46 @@
...
+          dataType: "json",
+          complete: function(data) {
+            var metrics = data.responseJSON['gauges'];
+            var usedHeap = metrics['heap.used'].value;
+            var maxHeap  = metrics['heap.max'].value;
+            var usedDirect = metrics['drill.allocator.root.used'].value;
+            var peakDirect = metrics['drill.allocator.root.peak'].value;
+            var heapUsage    = computeMemUsage(usedHeap,maxHeap);
```
arina-ielchiieva: Please remove unnecessary spaces here and in other file as well.

exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
```diff
@@ -109,18 +114,23 @@
...
+      $.each(["heap","non-heap","total","drill.allocator.root"], function(i, key) {
         var used    = gauges[key + ".used"].value;
-        var max     = gauges[key + ".max"].value;
-        var usage   = round((used / 1073741824), 2) + "GB";
-        var percent = round((used / max), 2);
-
-        var styleVal = "width: " + percent + "%;"
-        $("#" + key + "Usage").attr({
+        var max;
+        if (key.startsWith("drill.allocator")) {
```
arina-ielchiieva: Using `key.startsWith("drill.allocator")` several times in this method is not very elegant. Besides it he fact it is repeated action, I do not like such checks in general. I guess that can be re-written.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -291,6 +300,46 @@
...
+            }
+          });
+      }
+
+      //Compute Usage
+      function computeMemUsage(used, max) {
+        var percent = 0;
+        if ( max > 0)
+          var percent = Math.round((100 * used / max), 2);
+        var usage   = Math.round((used / 1073741824), 2) + "GB (" + Math.max(0, percent) + "%)";
```
arina-ielchiieva: This logic repeats in several files, so can create js function in separate file and import it.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -192,6 +199,8 @@
       var port = getPortNum();
       var timeout;
       var size = $("#size").html();
+      reloadMemory();
+      setInterval(reloadMemory, refreshTime);
```
arina-ielchiieva: In graceful shutdown `setTimeout` is used. Which one is better? Anayway it's better to be consistent in scheduler usage. Plus we do why need two schedulers? Can one work for both tasks? Plus if one drillbit is unavailable, there is no need to check memory.

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: I need to rebase. DRILL-6044 fix changed that entry.

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: Good point. I'll follow the same port extraction logic.

kkhatua: Actually, this is a challenge. We don't seem to be exposing any information on the http-port.  The only information I am getting back into DrillbitInfo are the other 3 ports: control, data and user. If I need to had HTTP to this (which I think it should), we need to modify the protobuf (UserBitShared). :(
Do you have any suggestions?

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: That's because the scheduler for graceful shutdown can change, while in this case we need a periodic refresh that shouldn't change.
As for a Drillbit going down, it can always come back online. As long as the page continues to list the Drillbit and it does get updated to reflect the status as "ONLINE" again, we should be pinging for the metrics too.

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: So one way to find out a Drillbit's HTTP port is through this call:
`http://localhost:8047/portNum `
to get 
```
{
  "port" : 8047
}
```
Notice how I need to use the port number in the request's URL to get that very number! 🤣

kkhatua:
@arina-ielchiieva 
Restricting this PR to address the bug in `metrics.ftl`
Will create a new PR for [DRILL-6289](https://issues.apache.org/jira/browse/DRILL-6289)

exec/java-exec/src/main/resources/rest/metrics/metrics.ftl
arina-ielchiieva: Please address the below comment.

arina-ielchiieva:
+1

kkhatua:
During a freshly setup Drillbit:
image

While running 2 large (600M rows) join query on a 4 node setup, and one completes:
image

The Heap Usage percentage is based on {{heapUsed/heapMax}}
The Actively Used Direct percentage is based on {{currentDirectInUse/PeakDirectUsed}} 
This is because the actual total direct memory used by Netty is not exposed. However, this does tell how much the Drillbit is using at the moment.

arina-ielchiieva:

PR #1164, opened 2018-03-14
DRILL-6234: Improved documentation for VariableWidthVector mutators

ilooner:
I had some confusion about how setValueCount should behave for variable width vectors. I added some documentation and unit tests which explain its behavior so that others don't waste time in the future.

### DIFF
--- a/exec/vector/src/main/codegen/templates/VariableLengthVectors.java
+++ b/exec/vector/src/main/codegen/templates/VariableLengthVectors.java
@@ -512,6 +512,8 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements V
   }
 
   /**
+   * <h4>Overview</h4>
+   * <p>
    * Mutable${minor.class} implements a vector of variable width values.  Elements in the vector
    * are accessed by position from the logical start of the vector.  A fixed width offsetVector
    * is used to convert an element's position to it's offset from the start of the (0-based)
@@ -520,6 +522,46 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements V
    *   The equivalent Java primitive is '${minor.javaType!type.javaType}'
    *
    * NB: this class is automatically generated from ValueVectorTypes.tdd using FreeMarker.
+   * </p>
+   * <h4>Contract</h4>
+   * <p>
+   *   <ol>
+   *     <li>
+   *       <b>Supported Writes:</b> {@link VariableWidthVector}s do not support random writes. In contrast {@link org.apache.drill.exec.vector.FixedWidthVector}s do
+   *       allow random writes but special care is needed.
+   *       </li>
+   *     <li>
+   *       <b>Writing Values:</b> All set methods must be called with a consecutive sequence of indices. With a few exceptions:
+   *       <ol>
+   *         <li>You can update the last index you just set.</li>
+   *         <li>You can reset a previous index (call it Idx), but you must assume all the data after Idx is corrupt. Also
+   *         note that the memory consumed by data that came after Idx is not released.</li>
+   *       </ol>
+   *     </li>
+   *     <li>
+   *       <b>Setting Value Count:</b> Vectors aren't explicitly aware of how many values they contain. So you must keep track of the
+   *       number of values you've written to the vector and once you are done writing to the vector you must call {@link Mutator#setValueCount(int)}.
+   *       It is possible to trim the vector by setting the value count to be less than the number of values currently contained in the vector. Note the extra memory consumed in
+   *       the data buffer is not freed when this is done.
+   *     </li>
+   *     <li>
+   *       <b>Memory Allocation:</b> When setting a value at an index you must do one of the following to ensure you do not get an {@link IndexOutOfBoundsException}.
+   *       <ol>
+   *         <li>
+   *           Allocate the exact amount of memory you need when using the {@link Mutator#set(int, byte[])} methods. If you do not
+   *           manually allocate sufficient memory an {@link IndexOutOfBoundsException} can be thrown when the data buffer runs out of space.
+   *         </li>
+   *         <li>
+   *           Or you can use the {@link Mutator#setSafe(int, byte[])} methods, which will automatically grow your data buffer to
+   *           fit your data.
+   *         </li>
+   *       </ol>
+   *     </li>
+   *     <li>
+   *       <b>Immutability:</b> Once a vector has been populated with data and {@link #setValueCount(int)} has been called, it should be considered immutable.
+   *     </li>
+   *   </ol>
+   * </p>
    */
   public final class Mutator extends BaseValueVector.BaseMutator implements VariableWidthVector.VariableWidthMutator {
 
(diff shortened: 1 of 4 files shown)


ilooner:
@paul-rogers

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -506,6 +506,8 @@ public boolean isNull(int index){
   }
 
   /**
+   * <h2>Overview</h2>
```
paul-rogers: Nit, but I think that h4 is the usual heading level used in Java doc comments. The higher levels are used in the surrounding generated HTML.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -514,6 +516,22 @@ public boolean isNull(int index){
...
    * NB: this class is automatically generated from ValueVectorTypes.tdd using FreeMarker.
+   * </p>
+   * <h2>Contract</h2>
+   * <p>
+   *   Variable length vectors do not support random writes. All set methods must be called for with a monotonically increasing consecutive sequence of indexes.
+   *   It is possible to trim the vector by setting the value count to be less than the number of values currently container in the vector with {@link #setValueCount(int)}, then
+   *   the process of setting values starts with the index after the last index.
+   * </p>
+   * <p>
+   *   It is also possible to back track and set the value at an index earlier than the current index, however, the caller must assume that all data container after the last
```
paul-rogers: "all data container after" --> "all data after the updated index" ?

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -514,6 +516,22 @@ public boolean isNull(int index){
...
+   *   set index is corrupted.
```
paul-rogers: Maybe add "changing the index does not release memory after the index."

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -514,6 +516,22 @@ public boolean isNull(int index){
    *   The equivalent Java primitive is '${minor.javaType!type.javaType}'
    *
    * NB: this class is automatically generated from ValueVectorTypes.tdd using FreeMarker.
+   * </p>
+   * <h2>Contract</h2>
+   * <p>
+   *   Variable length vectors do not support random writes. All set methods must be called for with a monotonically increasing consecutive sequence of indexes.
```
paul-rogers: This is very important to know. This is why spill-to-disk for hash agg will eventually cause a serious customer failure. Aggregate UDFs write to vectors to store intermediate group values. A "max" string can't. Instead, it writes to a Java object. That object will be lost on spill and reread. Will result in loosing prior max values and the aggregate starting over.

So, this little note is not just a nuisance, it is the fatal flaw in how we handle the (albeit obscure) string aggregate values.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -514,6 +516,22 @@ public boolean isNull(int index){
...
+   *   the process of setting values starts with the index after the last index.
+   * </p>
+   * <p>
+   *   It is also possible to back track and set the value at an index earlier than the current index, however, the caller must assume that all data container after the last
+   *   set index is corrupted.
+   * </p>
+   * <h2>Notes</h2>
+   * <p>
+   *   There is no gaurantee the data buffer for the {@link VariableWidthVector} will have enough space to contain the data you set unless you use setSafe. If you
+   *   use set you may get array index out of bounds exceptions.
```
paul-rogers: Said another way, either 1) be careful to manage your own memory, or 2) call `setSafe()`. That is, in fact, why `setSafe()` exists.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -514,6 +516,22 @@ public boolean isNull(int index){
    *   The equivalent Java primitive is '${minor.javaType!type.javaType}'
    *
    * NB: this class is automatically generated from ValueVectorTypes.tdd using FreeMarker.
+   * </p>
+   * <h2>Contract</h2>
```
paul-rogers: Might be worth summarizing how to use this:

1) Write to values sequentially. Fixed-width vectors allow random access, but special care is needed.
2) Keep track in client code of the total time count. Call `setValueCount()` once the vector is full to set the final value. (The vector does not know its count while a write is in progress.)
3) Either take responsibility for allocating enough memory, or call the `setSafe()` methods to automatically extend the vector.
4) Once vectors are written, they are immutable; no additional writes of any kind are allowed to that vector.

exec/vector/src/test/java/org/apache/drill/exec/vector/VariableLengthVectorTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+        setSafeIndexStrings("", 0, size, mutator);
+        setSafeIndexStrings("first cut ", size, fluffSize, mutator);
+
+        mutator.setValueCount(fluffSize);
+        Assert.assertEquals(fluffSize, accessor.getValueCount());
+
+        mutator.setValueCount(size);
+        Assert.assertEquals(size, accessor.getValueCount());
+        setSafeIndexStrings("redone cut ", size, fluffSize, mutator);
```
paul-rogers: While this works, we are actually violating the vector contract which is "once the value count is set, the vector becomes immutable." If the client is not done writing to the vector, the client should maintain the value count until it is finally done.

exec/vector/src/test/java/org/apache/drill/exec/vector/VariableLengthVectorTest.java
```diff
@@ -0,0 +1,152 @@
...
+        checkIndexStrings("", 0, size, accessor);
+
+      } finally {
+        vector.clear();
+      }
+    }
+  }
+
+  /**
+   * Set 10000 values. Then go back and set new values starting at the 1001 the record.
```
paul-rogers: Just FYI: the vector writers handle all this stuff for you. They allow overwriting the most recent value. They keep track of vector counts and data offsets. And so on. This is why I can offer such detailed comments: I learned how all this works when creating those classes. Would be very wonderful to start reusing that work rather than creating multiple implementations of the same thing... (Here, however, I recognize that the random access pattern wanted in your project does not match the sequential approach taken by the writers. But, as you are seeing, we can't really implement random access anyway...)

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
ilooner: Thanks for bringing this up. I'm sharing a design doc on the dev list tomorrow or the day after about how I plan to refactor HashAgg. It will cover how to facilitate unit tests and how to change the memory handling to use a deterministic calculator like the SortMemoryManager and soon to be introduced HashJoinMemoryCalculator (instead of catch OOMs). Perhaps you could comment on the doc about how to set ourselves up to fix this case.

exec/vector/src/test/java/org/apache/drill/exec/vector/VariableLengthVectorTest.java
ilooner: I agree the vector writers should be used. The reason why I was looking into this is that I saw strange behavior in the legacy HashTable where setValueCount was being called with a larger valueCount than there was data in a VarCharVector. I did an ugly (and now I think incorrect work around) for the issue and set about to make setValueCount support setting a valueCount larger than the number elements in the VarCharVector. Now I am realizing the issue is with the HashTableTemplate and I need to look into why it is behaving incorrectly.

Your review has been very helpful, I have a much better understanding of how value vectors should be used now. Thanks!

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
ilooner: Fixed

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
ilooner: Updated

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
ilooner: Added

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
ilooner: Liked this refactored phrasing

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
ilooner: This is a good way to layout the information. I switched the javadoc to follow this outline.

exec/vector/src/test/java/org/apache/drill/exec/vector/VariableLengthVectorTest.java
ilooner: Yikes! I didn't know this. Thanks for catching.

ilooner:
@paul-rogers Applied review comments.

PR #1167, opened 2018-03-14
DRILL-6241: Saffron properties config has the excessive permissions

vladimirtkach:
changed saffron.properties permission to 640


arina-ielchiieva:
+1

PR #1168, opened 2018-03-15
DRILL-6246: Reduced the size of the jdbc-all jar file

sachouche:
- The jdbc-all client jar has been growing in size from version to the next (~20MB in version 1.10, ~27MB in version 1.12, and 34MB in version 1.13
- Note the exact size of the size depends on the maven profile used during compilation (e.g., "mapr" which have more excludes)
- Originally, I tried to exclude the following calcite/avatica packages (responsible for the recent size increase): metrics, proto, org/apache/, com/fasterxml, com/google but some tests failed as the calcite code was referencing some of them
- In this pull request, I have excluded the following packages: calcite/avatica/org and calcite/avatica/metrics

### DIFF
--- a/exec/jdbc-all/pom.xml
+++ b/exec/jdbc-all/pom.xml
@@ -485,6 +485,8 @@
                <exclude>org/yaml/**</exclude>
                <exclude>hello/**</exclude>
                <exclude>webapps/**</exclude>
+               <exclude>**/org/apache/calcite/avatica/metrics/**</exclude>
+               <exclude>**/org/apache/calcite/avatica/org/**</exclude>
              </excludes>
            </filter>
          </filters>
@@ -511,7 +513,7 @@
                   This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                   </message>
-                  <maxsize>39000000</maxsize>
+                  <maxsize>38000000</maxsize>
                   <minsize>15000000</minsize>
                   <files>
                    <file>${project.build.directory}/drill-jdbc-all-${project.version}.jar</file>
@@ -571,7 +573,7 @@
                           This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                         </message>
-                        <maxsize>37000000</maxsize>
+                        <maxsize>35000000</maxsize>
                         <minsize>15000000</minsize>
                         <files>
                           <file>${project.build.directory}/drill-jdbc-all-${project.version}.jar</file>
@@ -788,6 +790,8 @@
                       <exclude>org/yaml/**</exclude>
                       <exclude>hello/**</exclude>
                       <exclude>webapps/**</exclude>
+                      <exclude>**/org/apache/calcite/avatica/metrics/**</exclude>
+                      <exclude>**/org/apache/calcite/avatica/org/**</exclude>
                     </excludes>
                   </filter>
                 </filters>


sachouche:
@sohami can you please review this pull request?

Thanks!

exec/jdbc-all/pom.xml
```diff
@@ -473,6 +473,8 @@
                <exclude>org/yaml/**</exclude>
                <exclude>hello/**</exclude>
                <exclude>webapps/**</exclude>
+               <exclude>**/org/apache/calcite/avatica/metrics/**</exclude>
+               <exclude>**/org/apache/calcite/avatica/org/**</exclude>
```
sohami: After removing these exclusion for apache profile what is the size of jdbc-all jar ? based on final size we should reduce the allowed maxSize from ~35MB to that value.

sachouche: The Apache jar size has shrunk by ~1MB from over 34MB to 33393541 bytes; I guess we can decrease the limit from 35MB to 34MB.

sohami: Would be great to make that change. Other than that, changes looks good to me

sachouche: done!

sohami:
+1 LGTM. Thanks for making the changes.

vvysotskyi:
Classes from `avatica.metrics` are used in `JsonHandler`, `ProtobufHandler` and `LocalService`. If Drill does not use these classes than I agree that we can exclude it from `jdbc-all` jar. 
Regarding excluding `avatica/org/**`, looks like the problem is in the Avatica pom files since there are no dependencies to `org.apache.commons` and `org.apache.http`, but they are shaded to the jar. Created Jira CALCITE-2215 to fix this issue, but for now, I think it's ok to exclude them.

sachouche:
I successfully ran the test-suite and some advanced tests which loaded the jdbc artifacts. Is this enough to validate the fix (and future exclusions to keep the jar size in-check)? if not, what other tests(s) do we need to run?

Thanks!

parthchandra:
I would recommend trying to setup a connection using Spotfire or Squirrel and running a couple of metadata queries and a couple of queries on complex data. (These have traditionally been areas that were affected when the jar was incomplete). 
Also try on a secure cluster. You might need help from other folks on some of this.

kkhatua:
@sachouche have you had a chance to test the generated JDBC driver with Spotfire/SQuirreL ?

sachouche:
No I didn't.

Regards,
Salim

Regards,
Salim

-----Original Message-----

exec/jdbc-all/pom.xml
sachouche: The problem with this PR is that running the test-suite is not enough; as @parthchandra indicated We need to setup few tools and run metadata / DDL & DML queries (involving complex data) to ensure there are no regressions. I am afraid this kind of optimizations will not be easy to include unless we automate the JDBC testing process.

krystaln:
cat git.properties
#Generated by Git-Commit-Id-Plugin
#Fri Sep 28 10:10:32 PDT 2018
git.commit.id.abbrev=527ce72
git.commit.user.email=sachouche2@gmail.com
git.commit.message.full=DRILL-6246\: Reduced the size of the jdbc-all jar file\n
git.commit.id=527ce7261b6fc6bf374a3eb0d6c40da6dc34a6dd
git.commit.message.short=DRILL-6246\: Reduced the size of the jdbc-all jar file
git.commit.user.name=Salim Achouche

Tested the drill-jdbc-all-1.14.0-SNAPSHOT.jar from the above branch with SQuirrel.  Tested the following items:
- text files
-parquet files
-json files
-hive
- complex data
-ctas
-create views
-information schema

I did not see any issues.

vdiravka:
@krystaln Could you please proceed and check the same cases for secure cluster as Parth mentioned?

Also it would be good to automate the JDBC testing process as Salim mentioned. Earlier there were issues with these dependencies in the process of validating the Drill rc. Possibly to create a ticket for that issue can be the step forward.

rhou1:
Automating the JDBC testing process is ongoing.  Like Drill dev, Drill QA has a list of enhancements that we would like to make.  Simba JDBC testing, Simba ODBC testing, Web UI testing, REST API testing, client application testing are all on the list, and we are making some progress.

vdiravka:
@rhou1 Is there public ticket for this task? Having this, we can rely on it in other similar cases.

@krystaln @sachouche The only one case is necessary to be verified to merge this PR to master branch.
It may also be helpful not to get blockers before Drill 1.15 release.

sachouche:
Closing this pull request since there is a new effort to separate the client and server code which should address the jdbc size issue.

PR #1171, opened 2018-03-16
DRILL-6231: Fix memory allocation for repeated list vector

ppadma:


ppadma:
@paul-rogers please review.

ppadma:
@paul-rogers I added the null check. Also, added a test case for 3D array i.e. repeated repeated list.

PR #1174, opened 2018-03-17
DRILL-6250: Sqlline start command with password appears in the sqllin…

vladimirtkach:
…e.log

### DIFF
--- a/common/src/main/java/org/apache/drill/common/config/DrillConfig.java
+++ b/common/src/main/java/org/apache/drill/common/config/DrillConfig.java
@@ -52,8 +52,10 @@ public class DrillConfig extends NestedConfig {
   public DrillConfig(Config config) {
     super(config);
     logger.debug("Setting up DrillConfig object.");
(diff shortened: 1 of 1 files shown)


common/src/main/java/org/apache/drill/common/config/DrillConfig.java
```diff
@@ -52,8 +52,8 @@
   public DrillConfig(Config config) {
     super(config);
     logger.debug("Setting up DrillConfig object.");
-    logger.trace("Given Config object is:\n{}",
-                 config.root().render(ConfigRenderOptions.defaults()));
+    logger.trace("Given Config object is:\n{}", config.withoutPath("password").withoutPath("sun.java.command")
```
arina-ielchiieva: Please add comment why we exclude `sun.jaba.command`.

vladimirtkach:
@arina-ielchiieva addressed code review  comments

arina-ielchiieva:
+1

PR #1175, opened 2018-03-19
DRILL-6262: IndexOutOfBoundException in RecordBatchSize for empty var…

sohami:
…iableWidthVector

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
@@ -321,10 +321,8 @@ public class RecordBatchSizer {
 
           // Calculate pure data size.
           if (isVariableWidth) {
-            UInt4Vector offsetVector = ((RepeatedValueVector) v).getOffsetVector();
-            int innerValueCount = offsetVector.getAccessor().get(valueCount);
             VariableWidthVector dataVector = ((VariableWidthVector) ((RepeatedValueVector) v).getDataVector());
-            totalDataSize = dataVector.getOffsetVector().getAccessor().get(innerValueCount);
+            totalDataSize = dataVector.getCurrentSizeInBytes();
           } else {
             ValueVector dataVector = ((RepeatedValueVector) v).getDataVector();
             totalDataSize = dataVector.getPayloadByteCount(elementCount);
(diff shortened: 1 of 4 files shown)


sohami:
@ppadma - Please review

sohami:
@bitblender - Updated the test. Please review.

ppadma:
LGTM. +1.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -321,10 +321,8 @@ public ColumnSize(ValueVector v, String prefix) {
 
           // Calculate pure data size.
           if (isVariableWidth) {
-            UInt4Vector offsetVector = ((RepeatedValueVector) v).getOffsetVector();
-            int innerValueCount = offsetVector.getAccessor().get(valueCount);
             VariableWidthVector dataVector = ((VariableWidthVector) ((RepeatedValueVector) v).getDataVector());
-            totalDataSize = dataVector.getOffsetVector().getAccessor().get(innerValueCount);
+            totalDataSize = dataVector.getCurrentSizeInBytes();
```
paul-rogers: Good improvement. The original code exposes far too much of the implementation.

After all these changes, does the "dataSize" include both the offset vector and bytes? It should, else calls will be wrong. There are supposed to be three sizes:

* Payload size: actual data bytes.
* Data size: data + offsets + bits
* Overall size: full length of all vectors.

Payload size is what the user sees. Data size is how we calculate row width (since the rows must contain the overhead bytes). Vector length, here, only helps compute density, but is generated elsewhere. The point is, keep all three in mind, but keep the code separate. Otherwise, it is *very* easy to get confused and have the calculations blow up...

sohami: @paul-rogers - I don't think `totalDataSize` includes both offset vector side and bytes size. It was meant to only include **pure data size only** for all entries in that column and that's what comment also suggests.

Instead `totalNetSize` includes the size for data and offset vector which is used for computing the rowWidth.

paul-rogers:
Thanks. See how confusing it is? I wrote the darn thing originally and even I can't keep the names straight... :-)

PR #1176, opened 2018-03-19
DRILL-6275: Fixed direct memory reporting in sys.memory.

ilooner:
@kkhatua Thanks for pinpointing the root cause! Please review.


ilooner:
Tested fix manually on my laptop.

kkhatua:
LGTM. Verified on a 4-node setup with running queries.
+1

PR #1178, opened 2018-03-21
DRILL-6278: Removed temp codegen directory in testing framework.

ilooner:
@vvysotskyi Please review

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TopN/TopNBatchTest.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TopN/TopNBatchTest.java
@@ -63,8 +63,6 @@ public class TopNBatchTest extends PopUnitTestBase {
   @Test
   public void priorityQueueOrderingTest() throws Exception {
     Properties properties = new Properties();
-    properties.setProperty(ClassBuilder.CODE_DIR_OPTION, dirTestWatcher.getDir().getAbsolutePath());
-
     DrillConfig drillConfig = DrillConfig.create(properties);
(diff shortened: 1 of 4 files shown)


vvysotskyi:
@ilooner, I have noticed that `ClassBuilder.CODE_DIR_OPTION` is used in `TopNBatchTest` class. Shoul we also remove it?

ilooner-mapr:
@vvysotskyi thanks for catching this. I've removed it from TopNBatchTest .

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TopN/TopNBatchTest.java
```diff
@@ -159,7 +157,6 @@ public void priorityQueueOrderingTest() throws Exception {
   @Test
   public void sortOneKeyAscending() throws Throwable {
     ClusterFixtureBuilder builder = ClusterFixture.builder(dirTestWatcher)
-      .configProperty(ClassBuilder.CODE_DIR_OPTION, dirTestWatcher.getDir().getAbsolutePath())
       .configProperty(CodeCompiler.ENABLE_SAVE_CODE_FOR_DEBUG_TOPN, true);
```
vvysotskyi: I think this line should be also removed since `ENABLE_SAVE_CODE_FOR_DEBUG_TOPN` should be disabled for tests.

ilooner: Thanks for catching. Removed it.

PR #1179, opened 2018-03-21
DRILL-6254: IllegalArgumentException: the requested size must be non-…

ppadma:
…negative

We should limit memory allocation to number of records that are going to be in the next batch, not the total number of records remaining. For very large remaining record count, when multiplied with high cardinality, integer overflows and is becoming negative.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
@@ -237,7 +237,10 @@ public class FlattenRecordBatch extends AbstractSingleRecordBatch<FlattenPOP> {
 
   private void handleRemainder() {
     int remainingRecordCount = flattener.getFlattenField().getAccessor().getInnerValueCount() - remainderIndex;
-    if (!doAlloc(remainingRecordCount)) {
+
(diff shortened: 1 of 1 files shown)


priteshm:
@bitblender can you please review this?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
```diff
@@ -237,7 +237,10 @@ protected IterOutcome doWork() {
 
   private void handleRemainder() {
     int remainingRecordCount = flattener.getFlattenField().getAccessor().getInnerValueCount() - remainderIndex;
-    if (!doAlloc(remainingRecordCount)) {
+
+    // remainingRecordCount can be much higher than number of rows we will have in outgoing batch.
+    // Do memory allocation only for number of rows we are going to have in the batch.
+    if (!doAlloc(Math.min(remainingRecordCount, flattenMemoryManager.getOutputRowCount()))) {
```
paul-rogers: Not related to this fix at all... If we run out of memory, we should throw an `OutOfMemoryException` rather than trying to set flags and handle the case. In particular, after the batch sizing fixes, if we can't allocate memory now, something is very wrong and we'll never be able to. This code may be a vestige of the old system where operators tried to negotiate via `OUT_OF_MEMORY` statuses...

PR #1143, opened 2018-03-01
DRILL-1491: Support for JDK 8

vladimirtkach:
Changed jdk version from 7 to 8 in pom.xml, drill-config.sh and others

PR #1182, opened 2018-03-22
DRILL-6287: apache-release profile should be disabled by default

vrozov:
@parthchandra @dvjyothsna Please review

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@
     <excludedGroups />
     <memoryMb>4096</memoryMb>
     <directMemoryMb>4096</directMemoryMb>
+    <additionalparam>-Xdoclint:none</additionalparam>
   </properties>
 
   <scm>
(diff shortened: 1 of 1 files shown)


vrozov:
@parthchandra Please review

vrozov:
There are two issues with enabling `apache-release` by default:
-  it triggers creating source `apache-drill-...-src.tar.gz` and `apache-drill-...-src.zip` archives.
- maven build for any sub-module fails.

The change disables activation of the `apache-release` profile based on JDK version and requires explicit activation during the Apache release process.

JDK 1.7 is not supported. See DRILL-1491 and #1143.

parthchandra:
Sorry, Maven not being a strong point, I didn't understand initially what I was looking at.

+1

pom.xml
```diff
@@ -66,6 +66,7 @@
     <excludedGroups />
     <memoryMb>4096</memoryMb>
     <directMemoryMb>4096</directMemoryMb>
+    <additionalparam>-Xdoclint:none</additionalparam>
```
vdiravka: Do we need a task (new Jira) for refactoring the Drill java docs and moving onto Java 8 doclint?

vrozov: @vdiravka Please see DRILL-4547.

vdiravka: Thanks

PR #1183, opened 2018-03-22
DRILL-6248: Added limit push down support for system tables

arina-ielchiieva:
1. PojoRecordReader started returning data in batches instead of returing all in one batch. Default batch size is 4000.
2. SystemTableScan supports limit push down while extrating data in record reader based on given max records to read.
3. Profiles and profiles_json tables apply limit push down while extracting data from store accessing profiles by range.

Details in [DRILL-6248](https://issues.apache.org/jira/browse/DRILL-6248).


arina-ielchiieva:
@kkhatua please review.

PR #1185, opened 2018-03-23
DRILL-6288: Upgrade org.javassist:javassist and org.reflections:reflections

vrozov:
@vdiravka Please review


exec/jdbc-all/pom.xml
```diff
@@ -559,7 +559,7 @@
                           This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                         </message>
-                        <maxsize>32000000</maxsize>
+                        <maxsize>33000000</maxsize>
```
vdiravka: This has increased due to the new size of javassist library?

vrozov: Yes.

PR #1187, opened 2018-03-24
DRILL-6286: Updated copyright range in NOTICE

dvjyothsna:


dvjyothsna:
@vrozov Please review

PR #1189, opened 2018-03-24
DRILL-6282: Update Drill's Metrics dependencies

vdiravka:
Hive upgrade brings unnecessary `io.dropwizard.metrics` dependencies  for Drill. They are used only for Hive test purposes. Moreover they are in conflict with Drill's `com.codahale.metrics` dependencies.

### DIFF
--- a/logical/pom.xml
+++ b/logical/pom.xml
@@ -85,17 +85,6 @@
     </dependency>
     
     <dependency>
-      <groupId>com.codahale.metrics</groupId>
-      <artifactId>metrics-core</artifactId>
-      <version>3.0.1</version>
-    </dependency>
-    <dependency>
-      <groupId>com.codahale.metrics</groupId>
-      <artifactId>metrics-servlets</artifactId>
-      <version>3.0.1</version>
-    </dependency>
-        
-    <dependency>
       <groupId>org.antlr</groupId>
       <artifactId>antlr-runtime</artifactId>
       <version>3.4</version>
--- a/pom.xml
+++ b/pom.xml
@@ -72,6 +72,7 @@
     <javassist.version>3.22.0-GA</javassist.version>
     <msgpack.version>0.6.6</msgpack.version>
     <reflections.version>0.9.10</reflections.version>
+    <metrics.version>4.0.2</metrics.version>
     <excludedGroups/>
     <memoryMb>4096</memoryMb>
     <directMemoryMb>4096</directMemoryMb>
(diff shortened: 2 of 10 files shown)


vdiravka:
@vrozov Could you please review this, since you were the reviewer for DRILL-5978 (Hive upgrade)

vrozov:
Why not make Drill to use `io.dropwizard.metrics` instead of `com.codahale.metrics`? As far as I can see it is the same library (this is why it causes the conflict).

vdiravka:
@vrozov I understand what you mean: `com.codahale` isn't updated for a long time
https://mvnrepository.com/artifact/com.codahale.metrics/metrics-core, 
but `io.dropwizard.metrics` is updated recently https://mvnrepository.com/artifact/io.dropwizard.metrics/metrics-core. 

So we can switch into the last one. I will update my PR and let you know. Thank you.

vrozov:
@vdiravka right, check the first link, the new `groupId` for `com.codahale.metrics` is `io.dropwizard.metrics`, so all new versions will be deployed using the new `groupId`.

vdiravka:
@vrozov I have replaced `com.codahale.metrics` with last `io.dropwizard.metrics`. 
Please review.

pom.xml
```diff
@@ -1164,7 +1164,27 @@
       <dependency>
         <groupId>io.dropwizard.metrics</groupId>
         <artifactId>metrics-core</artifactId>
-        <version>4.0.2</version>
+        <version>4.1.0-rc0</version>
```
vrozov: define version as a property and avoid using `rc` version unless it provides a significant benefit.

pom.xml
```diff
@@ -1333,6 +1353,12 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
```
vrozov: Where is the dependency used? Can it be replaced with the `io.dropwizard.metrics`?

logical/pom.xml
```diff
@@ -85,14 +85,12 @@
     </dependency>
     
     <dependency>
-      <groupId>com.codahale.metrics</groupId>
+      <groupId>io.dropwizard.metrics</groupId>
```
vrozov: Is it used in logical?

pom.xml
vrozov: I am not sure why is it necessary to have `com.yammer.metrics` in dependencyManagement?

vrozov:
Please update JIRA, PR and commit titles and squash commits.

pom.xml
vdiravka: I meant that it can be useful if some new dependencies will have older transitive `org.yummer.metrics` in future. 
But I do not mind, this isn't so important. Let's leave it without dependencyManagement control.

logical/pom.xml
vdiravka: No, it is. Thanks for catching this. 
Also I have noticed that `java-exec` used `io.dropwizard.metrics` as transitive from `drill-common`.
For consistency I have removed  `io.dropwizard.metrics` from `drill-memory-base` (it can leverage metrics from `drill-common` too).

vrozov: It is not a best practice to rely on a transitive dependency for a compile scope dependency. I'd recommend specifying the dependency on `io.dropwizard.metrics` explicitly in case `java-exec` or `drill-memory-base` uses it.

vrozov:
LGTM

PR #1191, opened 2018-03-27
DRILL-6103: lsb_release: command not found

kkhatua:
Thanks to Sanel Zukan for providing a small patch that checks for /etc/fedora-release path. This is more common, than lsb_release command on Linux distros.


kkhatua:
@chunhui-shi please review

amansinha100:
+1

PR #1192, opened 2018-03-28
DRILL-6299: Fixed a filter pushed down issue when a column doesn't ha…

sachouche:
This bug happens when the isNull predicate is applied on a column without statistics. @arina-ielchiieva  can you please review this pull request?

Thanks!


arina-ielchiieva:
+1

PR #1198, opened 2018-03-30
DRILL-6294: Changes to support Calcite 1.16.0

vvysotskyi:


vvysotskyi:
@amansinha100, @chunhui-shi could you please take a look?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillReduceAggregatesRule.java
```diff
@@ -218,7 +218,8 @@ private void reduceAggs(
         RelOptUtil.createProject(
             newAggRel,
             projList,
-            oldAggRel.getRowType().getFieldNames());
+            oldAggRel.getRowType().getFieldNames(),
+            DrillRelFactories.LOGICAL_BUILDER);
```
chunhui-shi: Could you explain why we are using DrillRelFactories.LOGICAL_BUILDER but not relBuilderFactory that was used in line 211? And could you point me to this 4 param createProject method with Factory as the last param?

vvysotskyi: In the second commit it was fixed and used relBuilderFactory to create builder and project.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillResultSetImpl.java
```diff
@@ -1142,16 +1142,8 @@ public void moveToCurrentRow() throws SQLException {
...
-      throwIfClosed();
-    } catch (AlreadyClosedSqlException e) {
-      // Can't throw any SQLException because AvaticaConnection's
-      // getStatement() is missing "throws SQLException".
-      throw new RuntimeException(e.getMessage(), e);
-    } catch (SQLException e) {
-      throw new RuntimeException(e.getMessage(), e);
-    }
+  public AvaticaStatement getStatement() throws SQLException {
+    throwIfClosed();
```
chunhui-shi: Since you are touching this file. You might want to remove not needed Exceptions for throwIfClosed() method that are derives of SqlException.

vvysotskyi: Thanks, it looks clearer now.

vvysotskyi:
@chunhui-shi, I have addressed all CR comments, could you please take a look again?

chunhui-shi:
+1. Thank you for making the fix.

PR #1197, opened 2018-03-30
DRILL-6279: UI indicates operators that spilled in-memory data to disk

kkhatua:
1. Detect the presence of an operator that has spilled to disk, in the Operators Overview section of a query's profile page.
2. Introduced API to inject html attributes into a cell. This is used to inject details about the spill in the avg peak and max peak memory usage
3. Javascript is leveraged to detect spilled operators' HTML elements, and inject a symbolic font for indicating the operator as having spilt.
4. Mouse-over effect indicates the average and max number of spill cycles the operator went through. 

For example, at the bottom right of this screenshot, you can see how the spilling operator is indicated, along with the mouse-over information
spilltodisksnapshot

NOTE: Bootstrap library comes with Glyphicons Halfling Fonts and meets the Apache license requirements.
http://glyphicons.com/license#halflingsbootstrap

PR #1199, opened 2018-04-01
DRILL-6303: Provide a button to copy the Drillbit's JStack shown in /threads

kkhatua:
Provides a button in the web UI to copy the thread stack to the user's clipboard.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/threads/threads.ftl
+++ b/exec/java-exec/src/main/resources/rest/threads/threads.ftl
@@ -17,17 +17,21 @@
   <a href="/queries">back</a><br/>
   <div class="page-header">
   </div>
-  <p>Auto refreshes every 2 seconds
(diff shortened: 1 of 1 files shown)


kkhatua:
@arina-ielchiieva 

I've added 2 snapshots and tested in Chrome.

**Mouse Over**, showing tooltip
mouseover

**On clicking**, the copy of the text to the clipboard, followed by the alert.
mouseonclick

arina-ielchiieva:
@kkhatua thanks for making the changes, it believe it's very useful but I suggest we do consider some design changes:
1. Information and auto-refresh display as badege and do not hard-code refresh limit.
2. Use glyphicon instead of regular button and move to right side. 
3. Simplify tooltip to `Copy to clipboard`.
4. Remove onlick logic.

kkhatua:
@arina-ielchiieva I've made updates based on the review. 
image

I've also skipped adding re-adding the Glyphicons package because we already have that as part of DRILL-6279 (PR #1197 ).

arina-ielchiieva:
+1

PR #1202, opened 2018-04-05
DRILL-6311: No logging information in drillbit.log / drillbit.out

sohami:


sohami:
@vrozov - Please help to review.

exec/vector/pom.xml
```diff
@@ -69,6 +69,7 @@
       <groupId>org.apache.drill</groupId>
       <artifactId>drill-common</artifactId>
       <version>${project.version}</version>
+      <scope>test</scope>
```
vrozov: @ilooner why was the dependency introduced?

ilooner: @vrozov JUnit supports including or excluding tests based on categories. Categories are just marker interfaces that are passed to the `@Category` annotation. For example a test class annotated with `@Category(SlowTest.Class)` will be excluded from a Travis run since Travis excludes the SlowTest category. All the test categories needed to be included in a common artifact since all the drill submodules require the test categories. So the test categories were added to drill-common. 

If drill-common is not included mvn would fail with an error when running tests marked with `@Category(SlowTest.class)` since JUnit would not be able to find the SlowTest class.

vrozov: How it worked before? In the commit that introduced the test dependency, there is no test categorization.

ilooner: The `<classifier>tests</classifier>` tag includes the test classes in drill-common. The test category classes are included in common/src/test/java/org/apache/drill/categories. So test category classes will be accessible to maven. However, looks like I forgot the test scope so the test classes and resources from drill-common were being included in Drill's standard (non-test) classpath.

vrozov: I mean that there are no unit tests that are annotated with `@Category(SlowTest.class)` or any other `@Category`, so it is not clear why the dependency is necessary.

ilooner: @vrozov As we discussed offline drill-common-tests is not a dependency of the vector project itself, it is a dependency required by a maven invocation when a category is specified. For example in `mvn test -DexcludedGroups="org.apache.drill.categories.SlowTest,org.apache.drill.categories.UnlikelyTest,org.apache.drill.categories.SecurityTest"` the category classes are required by maven. However when `mvn test` is used the category classes are not required by maven.

vrozov: OK, I see. @sohami I'd suggest moving the dependency to the parent exec module pom.

sohami:
@vrozov - Update the PR based on your suggestion. The command `mvn test -DexcludedGroups="org.apache.drill.categories.SlowTest,org.apache.drill.categories.UnlikelyTest,org.apache.drill.categories.SecurityTest"` was not working on latest master as well. Spoke with @ilooner and made sure atleast the command used in Travis is working with this change `mvn install -DexcludedGroups="org.apache.drill.categories.SlowTest,org.apache.drill.categories.UnlikelyTest,org.apache.drill.categories.SecurityTest"`

vrozov:
LGTM +1.

PR #1203, opened 2018-04-06
DRILL-6289: Cluster view should show more relevant information

kkhatua:
The commits make use of AJAX calls by the browser to all the Drillbits for status metrics, which are then displayed here. This works with unsecured Drill sessions, while secure setups will not be able to show the metrics because of certificate authentication.
To achieve this, the HTTP port needed to be discoverable, for which changes to the protobuf were done to provide the HTTP port number. CORS was allowed by the webserver as well.
The Drillbit homepage has been extended to show the following:
* Heap Memory in use
* Direct Memory (actively) in use - Since we're not able to get the total memory held by Netty at the moment, but only what is currently allocated to running queries. The base memory shown is the total peak allocation by Drill for Direct, because the actual Direct memory held is not available. In an active stable system in production, this should be around 90%.
* Average (System) Load Factor reported by the OS for that node.
* Pill indicating the current Drillbit whose WebUI is being viewed (existing UX feature), with an option to load (in a new window) the UI from other Drillbits.

Information such as the User, Data and Control port numbers don't help much during general cluster health, so it might be worth removing this information if more real-estate is needed. For now, we are retaining this.

### DIFF
--- a/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
@@ -1,0 +1,62 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.metrics;
+
+import java.lang.management.OperatingSystemMXBean;
+import java.util.HashMap;
+import java.util.Map;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricSet;
+
+/**
+ * Creates a Cpu GaugeSet
+ */
+class CpuGaugeSet implements MetricSet {
+
+  private OperatingSystemMXBean osMXBean;
+
+  CpuGaugeSet() {};
+  CpuGaugeSet(OperatingSystemMXBean osBean) {
+    this.osMXBean = osBean;
+  }
+
+  @Override
+  public Map<String, Metric> getMetrics() {
+    final Map<String, Metric> metric = new HashMap<>(1);
+    metric.put("os.load.avg", new OperatingSystemLoad(osMXBean));
+    return metric;
+  }
+}
+
+/**
+ * Creating an AverageSystemLoad Gauge
+ */
+final class OperatingSystemLoad implements Gauge<Double> {
+  private OperatingSystemMXBean osMXBean;
+  public OperatingSystemLoad(OperatingSystemMXBean osBean) {
+    this.osMXBean = osBean;
+  }
+
+  @Override
+  public Double getValue() {
+    return osMXBean.getSystemLoadAverage();
+  }
+
+}
--- a/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -17,30 +17,36 @@
  */
 package org.apache.drill.exec.metrics;
 
-import java.lang.management.OperatingSystemMXBean;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
 import java.util.HashMap;
 import java.util.Map;
 
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Metric;
 import com.codahale.metrics.MetricSet;
+import com.sun.management.OperatingSystemMXBean;
 
 /**
  * Creates a Cpu GaugeSet
  */
-class CpuGaugeSet implements MetricSet {
+@SuppressWarnings("restriction")
+public class CpuGaugeSet implements MetricSet {
 
   private OperatingSystemMXBean osMXBean;
+  private RuntimeMXBean rtMXBean;
 
-  CpuGaugeSet() {};
-  CpuGaugeSet(OperatingSystemMXBean osBean) {
-    this.osMXBean = osBean;
+  public CpuGaugeSet() {
+    this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    this.rtMXBean = ManagementFactory.getRuntimeMXBean();
   }
 
   @Override
   public Map<String, Metric> getMetrics() {
     final Map<String, Metric> metric = new HashMap<>(1);
     metric.put("os.load.avg", new OperatingSystemLoad(osMXBean));
+    metric.put("drillbit.load.avg", new DrillbitProcessLoad(osMXBean));
+    metric.put("drillbit.uptime", new DrillbitUptime(rtMXBean));
     return metric;
   }
 }
@@ -48,6 +54,7 @@ class CpuGaugeSet implements MetricSet {
 /**
  * Creating an AverageSystemLoad Gauge
  */
+@SuppressWarnings("restriction")
 final class OperatingSystemLoad implements Gauge<Double> {
   private OperatingSystemMXBean osMXBean;
   public OperatingSystemLoad(OperatingSystemMXBean osBean) {
@@ -57,6 +64,38 @@ final class OperatingSystemLoad implements Gauge<Double> {
   @Override
   public Double getValue() {
     return osMXBean.getSystemLoadAverage();
+  }
+
+}
+
+/**
+ * Creating an AverageDrillbitLoad Gauge
+ */
+@SuppressWarnings("restriction")
+final class DrillbitProcessLoad implements Gauge<Double> {
+  private OperatingSystemMXBean osMXBean;
+  public DrillbitProcessLoad(OperatingSystemMXBean osBean) {
+    this.osMXBean = osBean;
+  }
+
+  @Override
+  public Double getValue() {
+    return osMXBean.getProcessCpuLoad();
+  }
+
+}
+/**
+ * Creating an DrillbitUptime Gauge
+ */
+final class DrillbitUptime implements Gauge<Long> {
+  private RuntimeMXBean rtMXBean;
+  public DrillbitUptime(RuntimeMXBean osBean) {
+    this.rtMXBean = osBean;
+  }
+
+  @Override
+  public Long getValue() {
+    return rtMXBean.getUptime();
   }
 
 }
--- a/common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
@@ -61,7 +61,7 @@ public final class DrillMetrics {
       REGISTRY.registerAll(new BufferPoolMetricSet(ManagementFactory.getPlatformMBeanServer()));
       REGISTRY.registerAll(new MemoryUsageGaugeSet());
       REGISTRY.registerAll(new ThreadStatesGaugeSet());
-      REGISTRY.registerAll(new CpuGaugeSet(ManagementFactory.getOperatingSystemMXBean()));
+      REGISTRY.registerAll(new CpuGaugeSet());
       register("fd.usage", new FileDescriptorRatioGauge());
     }
 
--- a/common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
@@ -18,7 +18,6 @@
 package org.apache.drill.exec.metrics;
 
 import java.lang.management.ManagementFactory;
-import java.lang.management.OperatingSystemMXBean;
 import java.util.concurrent.TimeUnit;
 
 import com.codahale.metrics.JmxReporter;
--- a/common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
@@ -18,6 +18,7 @@
 package org.apache.drill.exec.metrics;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
 import java.util.concurrent.TimeUnit;
 
 import com.codahale.metrics.JmxReporter;
@@ -60,6 +61,7 @@ public final class DrillMetrics {
       REGISTRY.registerAll(new BufferPoolMetricSet(ManagementFactory.getPlatformMBeanServer()));
       REGISTRY.registerAll(new MemoryUsageGaugeSet());
       REGISTRY.registerAll(new ThreadStatesGaugeSet());
+      REGISTRY.registerAll(new CpuGaugeSet(ManagementFactory.getOperatingSystemMXBean()));
       register("fd.usage", new FileDescriptorRatioGauge());
     }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -191,11 +191,14 @@ public class Drillbit implements AutoCloseable {
     manager.getContext().getRemoteFunctionRegistry().init(context.getConfig(), storeProvider, coord);
     webServer.start();
     //Discovering HTTP port (in case of port hunting)
-    int httpPort = webServer.getPort();
-    // Registering Drillbit with cluster coordinator
-    registrationHandle = coord.register(md.toBuilder().setHttpPort(httpPort).build());
+    if (webServer.isRunning()) {
+      int httpPort = getWebServerPort();
+      registrationHandle = coord.register(md.toBuilder().setHttpPort(httpPort).build());
+    } else {
+      //WebServer is not running
+      registrationHandle = coord.register(md);
+    }
     // Must start the RM after the above since it needs to read system options.
-
     drillbitContext.startRM();
 
     Runtime.getRuntime().addShutdownHook(new ShutdownThread(this, new StackTrace()));
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -157,7 +157,7 @@ public class Drillbit implements AutoCloseable {
       profileStoreProvider = storeProvider;
     }
 
-    engine = new ServiceEngine(manager, context, allowPortHunting, isDistributedMode);
+    engine = new ServiceEngine(manager, context, webServer, allowPortHunting, isDistributedMode);
 
     stateManager = new DrillbitStateManager(DrillbitState.STARTUP);
     logger.info("Construction completed ({} ms).", w.elapsed(TimeUnit.MILLISECONDS));
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -182,7 +182,8 @@ public class Drillbit implements AutoCloseable {
     }
     DrillbitEndpoint md = engine.start();
     manager.start(md, engine.getController(), engine.getDataConnectionCreator(), coord, storeProvider, profileStoreProvider);
-    DrillbitContext drillbitContext = manager.getContext();
+    @SuppressWarnings("resource")
+    final DrillbitContext drillbitContext = manager.getContext();
     storageRegistry = drillbitContext.getStorage();
     storageRegistry.init();
     drillbitContext.getOptionManager().init();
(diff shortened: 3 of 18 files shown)


kkhatua:
**Snapshot:**
screenshot

kkhatua:
**PopOut to open in new window:**
popout

Tooltip describing the new columns
tooltipinfo

kkhatua:
@arina-ielchiieva could you please review this?

arina-ielchiieva:
Before the review I guess we need to clarify one thing. After DRILL-6044 Shutdown button was shown only for the current drillbit. As far as I understood, you cannot shutdown other drillbits from Web UI except of current. @dvjyothsna please confirm.

kkhatua:
@arina-ielchiieva You're correct about the inability to shutdown other Drillbits from the WebUI. But, from what @dvjyothsna explained, it seems that this was the result of a technical hurdle with HTTP**S** protocol. My understanding was that issuing shutdown to secure remote nodes won't work because of lack of certificates for those nodes, making it difficult for the remote Drillbit to authenticate the shutdown command. Keeping that in mind, the SHUTDOWN buttons would appear only for unsecured  Drillbits, i.e. when Authentication is **_not_** enabled.
https://github.com/kkhatua/drill/blob/dd507ea95a7d0ef4904092ef4fc43ee3b44ef058/exec/java-exec/src/main/resources/rest/index.ftl#L100
For setups requiring Authentication, the shutdown button would not appear.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -289,14 +314,94 @@
...
+              var httpPort = $("#row-"+i).find("#httpPort").contents().get(0).nodeValue.trim();
+              updateMetricsHtml(address, httpPort, i);
+          }
+      }
+
+      //Update memory
+      //TODO: HTTPS?
+      function updateMetricsHtml(drillbit,webport,idx) {
+        var result = $.ajax({
+          type: 'GET',
```
dvjyothsna: Will the ajax call works in case of HTTPS? or will it throw invalid certificate error since the certificate exception is not yet added for that host?

kkhatua: Good catch. I believe we'll get invalid certificate errors since the certificate exceptions most likely are not added. 
So the question I have then is whether we should disable making **any** remote HTTPS calls at all, or take a chance with the hope that the invalid certificate +might+ have been added?

dvjyothsna: I think its safe to not make any remote ajax requests until we figure out a way to perform them with BOTH ssl and auth enabled. The user might not notice the ajax failures and assume that we are displaying wrong info (The default values i.e., zero).

exec/java-exec/src/main/java/org/apache/drill/exec/service/ServiceEngine.java
```diff
@@ -102,6 +105,7 @@ public DrillbitEndpoint start() throws DrillbitStartupException, UnknownHostExce
     DrillbitEndpoint partialEndpoint = DrillbitEndpoint.newBuilder()
         .setAddress(hostName)
         .setUserPort(userPort)
+        .setHttpPort(httpPort)
```
dvjyothsna: Recently port-hunting capability is added to WebServer. So I think we should get port number from the WebServer.

kkhatua: That's tricky. The webserver doesn't come up until after the Endpoint is created and ready. One possible solution would be to perform the port hunt before the endpoint, and assign it to a newly instantiated webserver after the Endpoint is created.

kkhatua: Thanks to @dvjyothsna , referred to an outdated PR #1042 for the endpoint registration after the web-server starts up.

kkhatua:
@arina-ielchiieva / @dvjyothsna 
Updated the PR with changes based on the comments. Could you please review and let me know what else would need to be fixed.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -0,0 +1,62 @@
+/**
```
arina-ielchiieva: Javadoc should be changed to comment.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Map;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricSet;
+
+/**
+ * Creates a Cpu GaugeSet
+ */
+class CpuGaugeSet implements MetricSet {
```
arina-ielchiieva: Why not public...

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -0,0 +1,62 @@
...
+
+  private OperatingSystemMXBean osMXBean;
+
+  CpuGaugeSet() {};
```
arina-ielchiieva: Why do you need default constructor?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -296,12 +302,19 @@ private ConstraintSecurityHandler createSecurityHandler() {
...
 
   public int getPort() {
-    if (embeddedJetty == null || embeddedJetty.getConnectors().length != 1) {
+    if (!isRunning()) {
       throw new UnsupportedOperationException("Http is not enabled");
     }
     return ((ServerConnector)embeddedJetty.getConnectors()[0]).getPort();
   }
 
+  public boolean isRunning() {
```
arina-ielchiieva: Method code can be simplified to one line.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -52,12 +52,18 @@
           <thead>
             <tr>
               <th>#</th>
-              <th>Address</th>
+              <th title="Drillbits in Cluster" style="cursor: help;">Address <span class="glyphicon glyphicon-info-sign" style="font-size:100%"/></th>
+              <th title="Heap Memory in use (as percent of Total Heap)" style="cursor: help;">Heap Memory Usage <span class="glyphicon glyphicon-info-sign" style="font-size:100%"/></th>
+              <th title="Direct Memory ACTIVELY in use (as percent of Peak Usage so far).&#013;The percentage is an estimate of the total because the Netty library directly negotiates with the OS for memory" style="cursor: help;">Direct Memory Usage <span class="glyphicon glyphicon-info-sign" style="font-size:100%"/></th>
```
arina-ielchiieva: 1. `Direct Memory ACTIVELY in use (as percent of Peak Usage so far)` -> maybe we can remove `so far`?
2. `The percentage is an estimate of the total because the Netty library directly negotiates with the OS for memory` -> not sure that information about netty is useful for the user...

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -79,9 +91,14 @@
                   </span>
                 </td>
                 <td id="status" >${drillbit.getState()}</td>
+                <!-- #if (model.shouldShowAdminInfo() || !model.isAuthEnabled()) -->
```
arina-ielchiieva: Please remove the comment.

arina-ielchiieva:
Some more points:
1. Please add updated screenshots.
2. Since we have only one shutdown button, we should consider redesigning, it's position, since it looks really strange:

arina-ielchiieva:
@sohami could you please also review this PR (it contains some changes with port hunting, you might be more familiar with)?

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
kkhatua: Never noticed that the license header is actually a javadoc comment. Will fix this.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
kkhatua: I was originally thinking of making this private, so that the instance can only be created with the OS Bean. I might have left it as is, thinking the default behaviour could be without the need to explicitly pass the bean. I'll make it private.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: Agreed

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: 1. The problem is that the Peak Usage can go up over time. Adding 'so far' indicates that this isn't a final value.
2. We could drop it, but this was meant to indicate less confusion for users wondering why the MaxDirectMemory value is not reflected here. It does. however, make it rather verbose.

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: Missed catching that without an IDEA. Thanks

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
kkhatua: It belongs to the same package where it is referenced, so i thought it didn't need public visibility.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
kkhatua: It belongs to the same package where it is referenced, so i thought it didn't need public visibility.

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: Reduced it to
`Estimated Direct Memory ACTIVELY in use (as percent of Peak Usage)`

kkhatua:
Will add some new screenshots. 
In the meanwhile, I was thinking of reduce the 'SHUTDOWN' button to just the power Symbol, because we already have the 'STATUS' field. The single button does look odd in this case, but would be fine when the option appears for all Drillbits in an non-secured Drill cluster setup. Would that work, @arina-ielchiieva ?

kkhatua:
Latest screenshot with the power button,
image

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -283,20 +308,114 @@
...
+                    contentType : 'text/plain',
+                    complete: function(data) {
+                        alert(data.responseJSON["response"]);
+                        button.prop('disabled',true).css('opacity',0.5);
+                    }
+              });
+          }
+          </#if>
+
+      function popOutRemoteDbitUI(dbitHost, dbitPort) {
```
sohami: this and functions below should be under `<#if !model.isAuthEnabled()> <#/if>` block

arina-ielchiieva:
Regarding shutdown button (icon) placement, there are two more options:
1. have separate column for shutdown, when shutdown for the particular drillbits are unavailable, show disabled button / icon. Maybe better then just have empty rows as we do now.
2. remove shutdown column and add shutdown icon in address column near those drillbits which we can shutdown.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
kkhatua: I thought we might want to expand the GaugeSet to carry additional metrics like `ProcessCpuLoad` and `ProcessCpuTime` . 
Since we can shrink the {{SHUTDOWN}} button to a symbol, we do have some real-estate to provide the ProcessCPULoad information as well. Would it help to have that? 
https://docs.oracle.com/javase/7/docs/jre/api/management/extension/com/sun/management/OperatingSystemMXBean.html#getProcessCpuLoad()

sohami: I am not sure how useful `getProcessCpuLoad` alone will be, but in addition with total system load it can give the idea about how Drill and other services on node are using the CPU.  If you plan to include another metric then in that case `CpuGaugeSet` class is fine. Thanks for clarification.

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: We can't put this and the remaining functions within the IF block since the localhost operations are still valid. i.e. an authenticated client on a specific Drillbit can ping that Drillbit for refresh of the metrics. Only shutdown function needs to be managed with the IF block

kkhatua:
Forgot to mention, this also resolves DRILL-5261 because of the protobuf carrying the HTTP port info.
@arina-ielchiieva / @sohami  / @dvjyothsna  , please have a review of the changes.

arina-ielchiieva:
@kkhatua I definitely like new design more, thanks for the changes! I see some unanswered comments from Saurabh in index.ftl, could you please take a look?

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -180,7 +180,7 @@ public void run() throws Exception {
     if (profileStoreProvider != storeProvider) {
       profileStoreProvider.start();
     }
-    final DrillbitEndpoint md = engine.start();
+    DrillbitEndpoint md = engine.start();
```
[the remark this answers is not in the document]
kkhatua: Removed since the scope is anyway within this method only.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -260,7 +285,7 @@
       function fillQueryCount(row_id) {
           var requestPath = "/queriesCount";
           var url = getRequestUrl(requestPath);
-	   var result = $.ajax({
+       var result = $.ajax({
```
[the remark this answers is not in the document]
kkhatua: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -158,7 +158,13 @@ public void start() throws Exception {
     final int selectors = config.getInt(ExecConstants.HTTP_JETTY_SERVER_SELECTORS);
     final QueuedThreadPool threadPool = new QueuedThreadPool(2, 2, 60000);
     embeddedJetty = new Server(threadPool);
-    embeddedJetty.setHandler(createServletContextHandler(authEnabled));
+    ServletContextHandler webServerContext = createServletContextHandler(authEnabled);
+    //Allow for Other Drillbits to make REST calls
+    FilterHolder filterHolder = new FilterHolder(CrossOriginFilter.class);
+    filterHolder.setInitParameter("allowedOrigins", "*");
```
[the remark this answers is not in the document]
kkhatua: Yes. CORS is basically one of the means to prevent DoS attacks. I've added additional filter that allows access only for `/status/metrics` path.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -67,9 +73,15 @@
                 <td>${i}</td>
                 <td id="address" >${drillbit.getAddress()}<#if drillbit.isCurrent()>
                     <span class="label label-info" id="current">Current</span>
+                    <#else>
+                    <a onclick="popOutRemoteDbitUI('${drillbit.getAddress()}','${drillbit.getHttpPort()}');" style="cursor:pointer;color:blue" title="Open in new window"><span class="glyphicon glyphicon-new-window"/></a>
```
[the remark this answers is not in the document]
kkhatua: The `shutdown()` method should be in the freemarker's IF block that checks for authentication. Any other method is benign and does not risk changing the state of the Drillbit.
Also, this method simply will pop out a new window from which that Drillbit's UI can be accessed. This should help if someone wishes to navigate to that node to, say, review logs or even submit queries with that node as the foreman, without having to type the address (which might have a different HTTP port).

common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
```diff
@@ -60,6 +61,7 @@ private static void registerSystemMetrics() {
       REGISTRY.registerAll(new BufferPoolMetricSet(ManagementFactory.getPlatformMBeanServer()));
       REGISTRY.registerAll(new MemoryUsageGaugeSet());
       REGISTRY.registerAll(new ThreadStatesGaugeSet());
+      REGISTRY.registerAll(new CpuGaugeSet(ManagementFactory.getOperatingSystemMXBean()));
```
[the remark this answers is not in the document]
kkhatua: Retained the GaugeSet to carry ProcessCPU and Uptime information as well. So, I've not replaced `registerAll`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -189,11 +189,16 @@ public void run() throws Exception {
...
     webServer.start();
-
+    //Discovering HTTP port (in case of port hunting)
+    if (webServer.isRunning()) {
+      int httpPort = getWebServerPort();
+      registrationHandle = coord.register(md.toBuilder().setHttpPort(httpPort).build());
+    } else {
+      //WebServer is not running
+      registrationHandle = coord.register(md);
+    }
```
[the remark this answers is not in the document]
kkhatua: Done.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -79,9 +91,14 @@
...
                 </td>
                 <td id="status" >${drillbit.getState()}</td>
+                <!-- #if (model.shouldShowAdminInfo() || !model.isAuthEnabled()) -->
                 <#if (model.shouldShowAdminInfo() || !model.isAuthEnabled()) && drillbit.isCurrent() >
                   <td>
-                      <button type="button" id="shutdown" onClick="shutdown($(this));"> SHUTDOWN </button>
+                      <button type="button" id="shutdown" onClick="shutdown($(this));"><span class="glyphicon glyphicon-off"></span> SHUTDOWN </button>
+                  </td>
+                <#elseif (model.shouldShowAdminInfo() || !model.isAuthEnabled()) && !drillbit.isCurrent() >
+                  <td>
```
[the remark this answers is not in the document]
kkhatua: Done the change for checking for HTTPS. However, this commit just marginally fixes the UI, which (otherwise) should have a full revamp. I did attempt some of the details proposed in DRILL-6244, but the effort is not trivial.
For an unsecured setup, there is little reason to restrict the ability to shutdown multiple nodes, because you can always go directly and shut down those nodes without authentication. This simply makes it convenient to have a single UI window to achieve the same outcome.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -283,20 +308,114 @@
...
-			           alert(data["response"]);
+                                   alert(data["response"]);
                                    button.prop('disabled',true).css('opacity',0.5);
                                }
                   });
               }
           }
-      </#if>
+
+          function remoteShutdown(button,host) {
```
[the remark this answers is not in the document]
kkhatua: Done. Combined into one.

kkhatua:
@sohami , I've answered your questions in the review.

protocol/src/main/protobuf/Coordination.proto
```diff
@@ -18,6 +18,7 @@ message DrillbitEndpoint{
       OFFLINE = 3;
   }
   optional State state = 7;
+  optional int32 http_port = 8;
```
sohami: Please compile protobuf for c++ client as well (Instructions are [here](https://github.com/apache/drill/blob/master/contrib/native/client/readme.linux)). It would be really helpful if you can put a note in [readme.txt of this package](https://github.com/apache/drill/blob/master/protocol/readme.txt) for future dev to compile c++ client protobufs whenever a change is made in any of .proto files.

exec/java-exec/src/main/resources/rest/index.ftl
```diff
@@ -79,11 +94,15 @@
                   </span>
                 </td>
                 <td id="status" >${drillbit.getState()}</td>
-                <#if (model.shouldShowAdminInfo() || !model.isAuthEnabled()) && drillbit.isCurrent() >
+                <td class="uptime" >Not Available</td>
                   <td>
-                      <button type="button" id="shutdown" onClick="shutdown($(this));"> SHUTDOWN </button>
-                  </td>
+                <#if (model.shouldShowAdminInfo() || !model.isAuthEnabled()) && (drillbit.isCurrent() || !model.isUserEncryptionEnabled()) >
```
sohami: Please re-visit this condition. `IsUserEncryptionEnabled()` only checks for encryption between JDBC/ODBC client to Drillbit path not for Https. You have to check it using protocol.
You can put the check something like below:
```
if(model.shouldShowAdminInfo() && (drillbit.isCurrent() || (!model.isAuthEnabled() && location.protocol != https))) {
    showShutdownButton();
}
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -180,20 +180,22 @@ public void run() throws Exception {
     if (profileStoreProvider != storeProvider) {
       profileStoreProvider.start();
     }
-    final DrillbitEndpoint md = engine.start();
+    DrillbitEndpoint md = engine.start();
     manager.start(md, engine.getController(), engine.getDataConnectionCreator(), coord, storeProvider, profileStoreProvider);
-    @SuppressWarnings("resource")
-    final DrillbitContext drillbitContext = manager.getContext();
+    DrillbitContext drillbitContext = manager.getContext();
```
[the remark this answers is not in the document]
kkhatua: Done.

common/src/main/java/org/apache/drill/exec/metrics/DrillMetrics.java
```diff
@@ -18,6 +18,7 @@
 package org.apache.drill.exec.metrics;
 
 import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
```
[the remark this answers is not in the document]
kkhatua: Right. Fixed.

exec/java-exec/src/main/resources/rest/index.ftl
kkhatua: Fixed

sohami:
Thanks for all the changes Kunal. LGTM +1

PR #1206, opened 2018-04-11
DRILL_6314: Add complex types to result set loader

paul-rogers:
This PR is a bit of a large one as it adds Union, (non repeated) List and Repeated List type support to the column accessors, row set abstraction, result set loader abstraction, and associated mechanisms. The good new is that, after this PR, all the row set and result set loader work will be complete; we'll then move onto the scan operator and readers.

Both Union and (non-repeated) List have very odd semantics that required some creative gyrations in the existing code. A (non-repeated) List can old a single type (List of VarChar) say, in which the list entries can be null to model a JSON list:
```
{a: ["foo", "bar"]} {a: null}
```

List entries can also be unions (which can include null values.) A List starts as a simple list (one type), then gets "promoted" to a Union. Much complexity was needed to hide this process behind the simple row set abstractions.

There is similarity between List and Union, between List, Repeated List and "normal" Repeated (array) types. Refactoring reflects these commonalities.

Due to the complexity of the added types, this PR revises the mechanisms that build a row set from an existing schema. or a schema from a container.

This PR also includes a somewhat orthogonal projection mechanism that implements projection at the row set mechanism for simple columns, array values and elements within maps. This code is closely intertwined with schema creation, and it was not worth the effort to tease the two apart into separate PRs. 

Extensive unit tests show the results in action. These are probably the best place to start to understand the client view of the new mechanisms.

The work is divided up in a number of commits to help sort out work to each layer.

The row set mechanism is fully described [here](https://github.com/paul-rogers/drill/wiki/Batch-Handling-Upgrades).

Rather than write a long description here, please take a look at the code and the Wiki post. Then, post questions (specific or general) and I'll address those particular topics which need additional clarification.


paul-rogers:
@ppadma can you look at this one? It's a big one. Suggestion: work through small groups of files, perhaps one commit at a time as the layers are separated into separate commits. Maybe do a series of small reviews to keep things tractable.

paul-rogers:
Going to close this one and try to split this into smaller chunks.

paul-rogers:
Replacing this PR with a set of smaller ones.

PR #1207, opened 2018-04-11
DRILL-6320: Fixed License Headers

ilooner:
# Overview

There were several issues with license headers before this change:

1. We used the apache-rat-plugin to check licenses which was not thread safe. So when doing concurrent builds of submodules with `mvn -T 2C clean install` it could fail.
2. The apache-rat-plugin did not check all files for licenses and did not provide a warning for the files it skipped.
3. It allows inconsistent formatting of license headers.
4. It allowed license headers to be in java doc comments. This is problematic since we don't want the license to be included in every documentation page when we decide to publish java docs. Further more rat provided no way to configure the comment style allowed, so we were forced to live with this.

All of these issues were fixed by moving to the license-maven-plugin. I have reformatted all the license headers and manually verified the headers on each file to make sure they are correct. Now every file in the project is checked for a license except for some shell scripts that do not have file extensions in their names. Also none of the license headers are in java doc comments now!!!!

# Important Note

This change bumped up the required version of maven from 3.0.1 to 3.3.1 . This is because maven introduce a very useful feature in 3.3.1 which is the `${maven.multiModuleProjectDirectory}` property. This property holds the absolute path to the base directory of the project's root module regardless of which submodule the maven command is executed from. This property is used to define the path of the license header file and allows us to check license headers even when our current working directory is a sub module of the project.

Before this change is merged we need to update Drill's jenkins CI to maven 3.3.1 or later.


ilooner:
Abhishek has already updated jenkins to the latest maven. So we should be clear to merge after review is done.

arina-ielchiieva:
@ilooner since mix required maven version has changed, please update `install.md` as well.

vrozov:
-1 on removing apache-rat maven plugin. It is an Apache plugin to invoke "Release Audit Tool". More plugins can be added to check license headers if they do a better job in addition to apache-rat.

Both apache-rat and any other license verification plugins can be disabled by default (this will allow to run `mvn -T 2C clean install` if necessary) and only enabled in Travis CI.

ilooner:
@arina-ielchiieva Updated INSTALL.md

@vrozov Added the rat plugin back. Disabled license checking for both rat and the license-maven-plugin by default and enabled it for travis

pom.xml
```diff
@@ -198,6 +200,78 @@
 
   <build>
     <plugins>
+      <plugin>
+        <groupId>org.apache.rat</groupId>
+        <artifactId>apache-rat-plugin</artifactId>
+        <version>0.12</version>
```
vrozov: Do not specify the version explicitly, the one from the parent Apache pom should be used.

pom.xml
```diff
@@ -198,6 +200,78 @@
...
+        <executions>
+          <execution>
```
vrozov: Is this necessary? Should not `apache-rat:check` be already bound to the `validate` phase?

pom.xml
```diff
@@ -198,6 +200,78 @@
...
+            <id>rat-checks</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <skip>${rat.skip}</skip>  <!-- DRILL-6320 License checks are intentionally skipped by default for development -->
```
vrozov: Is this necessary? Should not `<skip>` use `rat.skip` property by default?

pom.xml
```diff
@@ -375,12 +449,12 @@
           </gitDescribe>
         </configuration>
       </plugin>
-
       <plugin>
         <groupId>com.mycila</groupId>
         <artifactId>license-maven-plugin</artifactId>
         <version>3.0</version>
         <configuration>
+          <skip>${license.skip}</skip> <!-- DRILL-6320 License checks are intentionally skipped by default for development -->
```
vrozov: Is the `skip` necessary?

vrozov:
@arina-ielchiieva In Apex we use special git user Apex Dev `dev@apex.apache.org` to make format only changes. I'd recommend doing the same for Drill.

pom.xml
ilooner: I originally tried `<skip>true</skip>` . However, this does not allow the skip to be overridden when the `-Drat.skip=false` argument is provided. It seems the only way to have an overridable default is to explicitly define a property with the default value.

pom.xml
ilooner: see comment above.

pom.xml
ilooner: This was here originally. I tried removing it. When it is removed the rat plugin is not executed when doing `mvn clean install`. So it seems to be required.

ilooner:
@arina-ielchiieva @vrozov I'll split the change into two commits. One with a change to the pom files and install.md . Another attributed to `dev@drill.apache.org` with all the formatting changes.

pom.xml
vrozov: try with `<skip>${rat.skip}<skip>` removed. Keep `<rat.skip>true</rat.skip>` in properties.

pom.xml
vrozov: I refer to `<phase>validate</phase>`.

ilooner:
@arina-ielchiieva fixed

pom.xml
ilooner: You are right, the docs say it is bound to the validate phase by default http://creadur.apache.org/rat/apache-rat-plugin/check-mojo.html . I also removed `<phase>validate</phase>` and tested that it work correctly.

pom.xml
ilooner: Your suggestion works. Made the change.

pom.xml
ilooner: updated with suggestion above

ilooner:
@vrozov @arina-ielchiieva Addressed comments.

vrozov:
LGTM, @ilooner please double check that Apache source distribution can be built with '-Drat.skip=false -Dlicense.skip=false': 
- build with `-P apache.release -Dgpg.skip=true`
- extract created source .tar.gz or .zip to a temp directory
- build with apache-rat and license check enabled (you may need to add more files to the license check exclusion).

ilooner:
@vrozov followed your instructions. Building the source distribution with license checks enabled works.

arina-ielchiieva:
+1, thanks for making the changes!

priteshm:
@arina-ielchiieva or @Ben-Zvi would it make sense to merge this in earlier before the batch commits later in the week? the number of files changes are large so don't want @ilooner to have to rebase each time.

PR #1208, opened 2018-04-12
DRILL-6295: PartitionerDecorator may close partitioners while CustomRunnable are active during query cancellation

vrozov:
@ilooner Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerDecorator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerDecorator.java
@@ -18,10 +18,15 @@
 package org.apache.drill.exec.physical.impl.partitionsender;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.LockSupport;
 
 import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.ops.OperatorStats;
@@ -31,7 +36,8 @@ import org.apache.drill.exec.testing.ControlsInjectorFactory;
 import org.apache.drill.exec.testing.CountDownLatchInjection;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
+import com.google.common.base.Preconditions;
+import com.google.common.util.concurrent.MoreExecutors;
 
 /**
  * Decorator class to hide multiple Partitioner existence from the caller
@@ -41,34 +47,37 @@ import com.google.common.collect.Lists;
  * The algorithm to figure out processing versus wait time is based on following formula:
  * totalWaitTime = totalAllPartitionersProcessingTime - max(sum(processingTime) by partitioner)
  */
-public class PartitionerDecorator {
+public final class PartitionerDecorator {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PartitionerDecorator.class);
   private static final ControlsInjector injector = ControlsInjectorFactory.getInjector(PartitionerDecorator.class);
 
   private List<Partitioner> partitioners;
   private final OperatorStats stats;
-  private final String tName;
-  private final String childThreadPrefix;
   private final ExecutorService executor;
   private final FragmentContext context;
+  private final Thread thread;
+  private final boolean enableParallelTaskExecution;
 
+  PartitionerDecorator(List<Partitioner> partitioners, OperatorStats stats, FragmentContext context) {
+    this(partitioners, stats, context, partitioners.size() > 1);
+  }
 
-  public PartitionerDecorator(List<Partitioner> partitioners, OperatorStats stats, FragmentContext context) {
+  PartitionerDecorator(List<Partitioner> partitioners, OperatorStats stats, FragmentContext context, boolean enableParallelTaskExecution) {
     this.partitioners = partitioners;
     this.stats = stats;
     this.context = context;
-    this.executor = context.getExecutor();
-    this.tName = Thread.currentThread().getName();
-    this.childThreadPrefix = "Partitioner-" + tName + "-";
+    this.enableParallelTaskExecution = enableParallelTaskExecution;
+    executor =  enableParallelTaskExecution ?  context.getExecutor() : MoreExecutors.newDirectExecutorService();
+    thread = Thread.currentThread();
   }
 
   /**
    * partitionBatch - decorator method to call real Partitioner(s) to process incoming batch
    * uses either threading or not threading approach based on number Partitioners
    * @param incoming
-   * @throws IOException
+   * @throws ExecutionException
    */
-  public void partitionBatch(final RecordBatch incoming) throws IOException {
+  public void partitionBatch(final RecordBatch incoming) throws ExecutionException {
     executeMethodLogic(new PartitionBatchHandlingClass(incoming));
   }
 
@@ -76,9 +85,9 @@ public class PartitionerDecorator {
    * flushOutgoingBatches - decorator to call real Partitioner(s) flushOutgoingBatches
    * @param isLastBatch
    * @param schemaChanged
-   * @throws IOException
+   * @throws ExecutionException
    */
-  public void flushOutgoingBatches(final boolean isLastBatch, final boolean schemaChanged) throws IOException {
+  public void flushOutgoingBatches(final boolean isLastBatch, final boolean schemaChanged) throws ExecutionException {
     executeMethodLogic(new FlushBatchesHandlingClass(isLastBatch, schemaChanged));
   }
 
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerDecorator.java
```diff
@@ -262,68 +279,120 @@ public FlushBatchesHandlingClass(boolean isLastBatch, boolean schemaChanged) {
...
-      }
-
-      final Thread currThread = Thread.currentThread();
-      final String currThreadName = currThread.getName();
-      final OperatorStats localStats = part.getStats();
-      try {
-        final String newThreadName = parentThreadName + currThread.getId();
-        currThread.setName(newThreadName);
+      final Thread thread = Thread.currentThread();
+      if (state.get() == STATE.NEW && runner.compareAndSet(null, thread)) {
```
ilooner: Why is this check necessary?

ilooner: Never mind I think I see why

ilooner: I think there is a race condition here. Consider the following case:

1. A PartitionTask starts executing, let's call it **Task A**
2. The PartitionTask executes the state check `state.get() == STATE.NEW` and then execution stops temporarily.
3. The main PartitionDecorator thread executes await(count, partitionerTasks)
4. `context.getExecutorState().shouldContinue()` is false so the PartitionTasks are cancelled.
5. The cancel method is called for **Task A**
6. In the cancel method ` (state.compareAndSet(STATE.NEW, mayInterruptIfRunning ? STATE.INTERRUPTING : STATE.CANCELLED)` will return true
7. `Thread thread = runner.get();` is executed but it is null since **Task A** has not set the runner yet.
8. The else statement in the cancel method is executed and `((ThreadPoolExecutor)partitionerDecorator.executor).remove(this);` is called.
9. The remove method does not cancel **Task A** since it has already started executing, and the interrupt is not set so it continue running.
10. `count.decrementAndGet();` is executed so the count will be zero but **Task A** is still running.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerDecorator.java
```diff
@@ -262,68 +280,122 @@ public FlushBatchesHandlingClass(boolean isLastBatch, boolean schemaChanged) {
...
+            state.lazySet(STATE.EXCEPTIONAL);
+          }
+        }
+        if (count.decrementAndGet() == 0) {
+          LockSupport.unpark(partitionerDecorator.thread);
+        }
+        thread.setName(name);
+      }
+      runner.set(null);
+      while (state.get() == STATE.INTERRUPTING) {
```
ilooner: I think we can still leave a dirty interrupt flag. Consider the following case.

 1. We create a PartitionerTask **Task A**.
 2. **Task A** executes and reaches `thread.setName(name)` in the run method.
 3. Now a cancel is called.
 4. `thread.interrupt()` and `state.lazySet(STATE.INTERRUPTED)` is executed.
 5. Now **Task A** can continue executing
 6. **Task A** checks `state.get() == STATE.INTERRUPTING` which is false. So Thread.sleep is never called and the interrupt flag is never cleared.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerDecorator.java
```diff
@@ -262,68 +280,124 @@ public FlushBatchesHandlingClass(boolean isLastBatch, boolean schemaChanged) {
...
+        Thread.yield();
+      }
+      // Clear interrupt flag
       try {
-        testCountDownLatch.await();
-      } catch (final InterruptedException e) {
-        logger.debug("Test only: partitioner thread is interrupted in test countdown latch await()", e);
+        Thread.sleep(0);
```
ilooner: Could we use Thread.interrupted() instead? Javadoc suggests it's a good alternative to use for clearing the interrupt flag. Also it avoids an unnecessary yield on some JVM implementations.

```
    /**
     * Tests whether the current thread has been interrupted.  The
     * <i>interrupted status</i> of the thread is cleared by this method.  In
     * other words, if this method were to be called twice in succession, the
     * second call would return false (unless the current thread were
     * interrupted again, after the first call had cleared its interrupted
     * status and before the second call had examined it).
     *
     * <p>A thread interruption ignored because a thread was not alive
     * at the time of the interrupt will be reflected by this method
     * returning false.
     *
     * @return  <code>true</code> if the current thread has been interrupted;
     *          <code>false</code> otherwise.
     * @see #isInterrupted()
     * @revised 6.0
     */
```

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerDecorator.java
```diff
@@ -118,105 +127,114 @@ public PartitionOutgoingBatch getOutgoingBatches(int index) {
...
-      // no need for threads
-      final OperatorStats localStatsSingle = partitioners.get(0).getStats();
-      localStatsSingle.clear();
-      localStatsSingle.startProcessing();
+  @VisibleForTesting
+  void executeMethodLogic(final GeneralExecuteIface iface) throws ExecutionException {
+    // To simulate interruption of main fragment thread and interrupting the partition threads, create a
+    // CountDownInject latch. Partitioner threads await on the latch and main fragment thread counts down or
+    // interrupts waiting threads. This makes sure that we are actually interrupting the blocked partitioner threads.
+    try (CountDownLatchInjection testCountDownLatch = injector.getLatch(context.getExecutionControls(), "partitioner-sender-latch")) {
```
ilooner: I'm not sure that we should be using the injector to create a count down latch here. My understanding is that we have to define a `partitioner-sender-latch` injection site on the `"drill.exec.testing.controls"` property and it is intended only for testing. See ControlsInjectionUtil.createLatch(). The default value for `drill.exec.testing.controls` is empty so the getLatch method would return a Noop latch since `partitioner-sender-latch` is undefined. Since we always want to create a count down latch here (not just for testing) shouldn't we directly create one?

vrozov: The `testCountDownLatch` is used only for testing and initialized to 1. The wait is on `count`.

ilooner: I see thx.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java
```diff
@@ -161,8 +161,11 @@ public OperatorStats getStats() {
    * @param schemaChanged  true if the schema has changed
    */
   @Override
-  public void flushOutgoingBatches(boolean isLastBatch, boolean schemaChanged) throws IOException {
+  public void flushOutgoingBatches(boolean isLastBatch, boolean schemaChanged) throws IOException, InterruptedException {
     for (OutgoingRecordBatch batch : outgoingBatches) {
+      if (Thread.interrupted()) {
+        throw new InterruptedException();
```
ilooner: Since we are checking for interrupts here already could we remove `Thread.currentThread().isInterrupted()` in the flush(boolean schemaChanged) method?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerDecorator.java
```diff
@@ -262,68 +280,122 @@ public FlushBatchesHandlingClass(boolean isLastBatch, boolean schemaChanged) {
...
+    }
+
+    void cancel(boolean mayInterruptIfRunning) {
+      Preconditions.checkState(Thread.currentThread() == partitionerDecorator.thread,
+          String.format("PartitionerTask can be cancelled only from the main %s thread", partitionerDecorator.thread.getName()));
+      if (runner.compareAndSet(null, partitionerDecorator.thread)) {
+        if (partitionerDecorator.executor instanceof ThreadPoolExecutor) {
+          ((ThreadPoolExecutor)partitionerDecorator.executor).remove(this);
+        }
+        count.decrementAndGet();
```
ilooner: The main PartitionDecorator thread may not get unparked.

1. Two PartitionerTasks are created.
2. Before any of the PartitionerTasks set their runner, the main Partitioner thread enters await() and cancels each PartitionerTask.
3. The cancel method executes ```
        if (partitionerDecorator.executor instanceof ThreadPoolExecutor) {
          ((ThreadPoolExecutor)partitionerDecorator.executor).remove(this);
        }
        count.decrementAndGet();```
    For each PartitionerTask.
 4. The count goes to zero, but the main PartitionerDecorator thread still calls Locksupport.park() in await().
 5. Unpark wasn't called when the count reached zero so the main PartitionDecorator thread can remain parked.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java
vrozov: I'll revert back throwing 'InterruptedException' to avoid mishandling of the last batch.

ilooner:
+1 LGTM I do not see anymore race conditions

vrozov:
@parthchandra Please review

parthchandra:
+1

PR #1210, opened 2018-04-16
DRILL-6270: Add debug startup option flag for drill in embedded and s…

agozhiy:
…erver mode

Works with the drillbit.sh, sqlline and sqlline.bat:
 Usage: --debug:[parameter1=value,parameter2=value]
 Optional parameters:
 port=[port_number] - debug port number
 suspend=[y/n] - pause until the IDE connects


distribution/src/resources/runbit
```diff
@@ -65,6 +65,47 @@ drill_rotate_log ()
     fi
 }
 
+args=( $@ )
+RBARGS=()
```
arina-ielchiieva: Does `RBARGS` have some meaning? Maybe it better to give clearer naming?
Also please add comment describing that you remove debug string from original args but leave all other args.

distribution/src/resources/runbit
```diff
@@ -65,6 +65,47 @@ drill_rotate_log ()
...
+
+  if [ -z $DEBUG_PORT ]; then
+    DEBUG_PORT=50000
+  fi
+  if [ -z $DEBUG_SUSPEND ]; then
+    DEBUG_SUSPEND='n'
+  fi
+
+  JAVA_DEBUG="-Xdebug -Xnoagent -Xrunjdwp:transport=dt_socket,address=$DEBUG_PORT,server=y,suspend=$DEBUG_SUSPEND"
+fi
```
paul-rogers: This is overly complex and puts Drill in the business of tracking the arguments that the JVM wants to enable debugging. Those flags changed in the past and may change again.

Would recommend instead:

```
--jvm -anyArgumentYouWant
```

Multiple `--jvm` arguments can appear. Simply append this to the the `DRILL_JAVA_OPTS` variable. Very simple.

arina-ielchiieva: Sounds reasonable to me.

agozhiy: Done.

agozhiy:
@paul-rogers / @arina-ielchiieva , I addressed your comments, could you please review?

arina-ielchiieva:
LGTM.

PR #1213, opened 2018-04-17
DRILL-6334: Minor code cleanup

paul-rogers:
Pulled the remaining code cleanup items out of the Result Set work into this simple PR.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestPromotableWriter.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestPromotableWriter.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -33,6 +33,7 @@ public class TestPromotableWriter {
   public void list() throws Exception {
     BufferAllocator allocator = RootAllocatorFactory.newRoot(DrillConfig.create());
     TestOutputMutator output = new TestOutputMutator(allocator);
+    @SuppressWarnings("resource")
     ComplexWriter rootWriter = new VectorContainerWriter(output, true);
     MapWriter writer = rootWriter.rootAsMap();
 
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestPromotableWriter.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestPromotableWriter.java
@@ -1,4 +1,4 @@
-/*
+/**
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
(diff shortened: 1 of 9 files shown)


exec/vector/src/main/java/org/apache/drill/exec/exception/OversizedAllocationException.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
vrozov: Should be already handled by Tim's PR #1207.

arina-ielchiieva:
Should we have Jira for this? Otherwise, how it can be tracked for weekly batch commits?

exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestPromotableWriter.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
vrozov: All license headers should be already fixed as part of @ilooner PR #1207 or #1215.

exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestPromotableWriter.java
```diff
@@ -33,6 +33,7 @@
   public void list() throws Exception {
     BufferAllocator allocator = RootAllocatorFactory.newRoot(DrillConfig.create());
     TestOutputMutator output = new TestOutputMutator(allocator);
+    @SuppressWarnings("resource")
```
vrozov: Why is it necessary to suppress resource not being closed warning here and OK to remove warning suppression in other places?

exec/vector/src/main/java/org/apache/drill/exec/exception/OversizedAllocationException.java
```diff
@@ -27,6 +27,7 @@
  * {@link RecordBatch#next() iteration}.</p>
  *
  */
+@SuppressWarnings("serial")
```
vrozov: Will it be better to provide `serialVersionUUID` instead of warning suppression?

exec/vector/src/main/java/org/apache/drill/exec/util/CallBack.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
vrozov: The same as the other license header

exec/vector/src/test/java/org/apache/drill/exec/vector/VariableLengthVectorTest.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
vrozov: The same as the other license header

exec/vector/src/test/java/org/apache/drill/exec/vector/VariableLengthVectorTest.java
```diff
@@ -37,6 +37,7 @@ public void testSettingSameValueCount()
   {
     try (RootAllocator allocator = new RootAllocator(10_000_000)) {
       final MaterializedField field = MaterializedField.create("stringCol", Types.required(TypeProtos.MinorType.VARCHAR));
+      @SuppressWarnings("resource")
```
vrozov: @paul-rogers Thank you for making the changes. Can you please double check whether it is necessary to suppress the warning here. It is confusing why the same warning suppression is removed in some places and added in others.

paul-rogers: These warnings may be unique to Eclipse. In this case, the vector is created, but not used in try-with-catch block which is the key purpose of anything that derives from `AutoCloseable`. Drill uses that class simply as a "close() without throwing exceptions" interface.

Since the warning really does not help in Drill, (and may not appear in IntelliJ), I can just turn the warning off in Eclipse so I'm not tempted to silence the warnings by inserting these annotations.

vrozov:
LGTM

PR #1215, opened 2018-04-17
DRILL-6338: Do not skip license maven plugin when formatting generate…

ilooner:
…d protobuf files.

And fixed malformed .travis.yml

### DIFF
--- a/protocol/pom.xml
+++ b/protocol/pom.xml
@@ -55,6 +55,9 @@
       <activation>
         <activeByDefault>false</activeByDefault>
       </activation>
+      <properties>
+        <license.skip>false</license.skip>
+      </properties>
       <build>
         <plugins>
           <plugin>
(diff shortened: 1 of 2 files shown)


ilooner:
@vrozov @arina-ielchiieva Please review.

protocol/pom.xml
```diff
@@ -149,8 +152,12 @@
           <plugin>
             <groupId>com.mycila</groupId>
             <artifactId>license-maven-plugin</artifactId>
+            <configuration>
```
vrozov: Do no force `<skip>`. It should be already `false`.

ilooner: Thanks for catching. Removed, verified that `mvn process-sources -P proto-compile` and `mvn clean install -P proto-compile` generated code with the headers and that `mvn process-sources -P proto-compile -Dlicense.skip=true` generates sources without license headers.

arina-ielchiieva:
LGTM.

PR #1217, opened 2018-04-17
DRILL-6302: Fixed NPE in Drillbit close method

dvjyothsna:
@sohami Please review


exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -225,14 +225,18 @@ public synchronized void close() {
     }
     final Stopwatch w = Stopwatch.createStarted();
     logger.debug("Shutdown begun.");
-    registrationHandle = coord.update(registrationHandle, State.QUIESCENT);
+    if (registrationHandle != null) {
+      registrationHandle = coord.update(registrationHandle, State.QUIESCENT);
+    }
```
sohami: may be create a private method like `UpdateState(State newState)` and encapsulate null check in that method

sohami:
+1 LGTM.

PR #1219, opened 2018-04-18
DRILL-6339: Add a new option to disable TopN (for testing)

Ben-Zvi:
To allow simple testing of the External Sort operator (e.g., test spilling), we would like to use a query of the form: ".... ORDER BY ... LIMIT ... "; however the current planner unconditionally selects the TopN operator for such a query.
   Solution: Introduce an internal option that can negate the planner's choice in this case. The implementation overrides the matches() method (default was returning true) in PushLimitToTopN , returning the new boolean option (default true) instead.
  (The last 'true' in SystemOptionManager makes this option 'internal'). This new option was tested manually.


arina-ielchiieva:
LGTM.

PR #1223, opened 2018-04-18
DRILL-6324: Unnest initial implementation

parthchandra:
Implementation of the unnest operator that works in sync with the Lateral Join operator. The code is based on the Flatten implementation except that it does not do the cross join that flatten does. As a result, the operator does not need any code generation.
Commit # 2 - adds specific handling for kill that may be as a result of a limit being reached by a downstream operator. 

@sohami, @amansinha100 please review.


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/Unnest.java
```diff
@@ -0,0 +1,49 @@
...
+ * @see UnnestImpl
+ */
+public interface Unnest {
+  //TemplateClassDefinition<Unnest> TEMPLATE_DEFINITION = new TemplateClassDefinition<Unnest>(Unnest.class, UnnestImpl
+  // .class);
+
+  void setup(FragmentContext context, RecordBatch incoming, RecordBatch outgoing, List<TransferPair> transfers,
+      LateralContract lateral) throws SchemaChangeException;
+
+  int unnestRecords(int recordCount);
```
amansinha100: Would be good to add a one line javadoc for these apis

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
```diff
@@ -0,0 +1,137 @@
...
+      default:
+        throw new UnsupportedOperationException();
+    }
+  }
+
+  @Override public final void setup(FragmentContext context, RecordBatch incoming, RecordBatch outgoing,
+      List<TransferPair> transfers, LateralContract lateral) throws SchemaChangeException {
+
+    this.svMode = incoming.getSchema().getSelectionVectorMode();
+    switch (svMode) {
```
amansinha100: Could be simplified to if(svMode != SelectionVectorMode.NONE)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  @Override
+  public void setOutputCount(int outputCount) {
+    outputLimit = outputCount;
+  }
+
+  @Override
+  public final int unnestRecords(final int recordCount) {
+    switch (svMode) {
```
amansinha100: Is this additional check needed since setup() would have already checked svMode ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestRecordBatch.java
```diff
@@ -0,0 +1,451 @@
...
+      logger.error("setUnnestVector cast failed and recordcount is 0, create empty vector anyway.");
+      vector = new RepeatedMapVector(field, oContext.getAllocator(), null);
+    } else {
+      vector = RepeatedValueVector.class.cast(inVV);
+    }
+    unnest.setUnnestField(vector);
+  }
+
+  @Override
+  protected IterOutcome doWork() {
```
amansinha100: Flatten's doWork() could potentially throw an OOM.  Can a similar situation occur for Unnest ?  (although much less likely due to the row-by-row nature of unnest when running in the context of Lateral join).

PR #1224, opened 2018-04-19
DRILL-6321: Customize Drill's conformance. Allow support to APPLY key…

chunhui-shi:
…words

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillConformance.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillConformance.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.planner.sql;
+
+import org.apache.calcite.sql.validate.SqlConformanceEnum;
+import org.apache.calcite.sql.validate.SqlDelegatingConformance;
+
+/**
+ * Drill's SQL conformance is SqlConformanceEnum.DEFAULT except for method isApplyAllowed().
+ * Since Drill is going to allow OUTER APPLY and CROSS APPLY to allow each row from left child of Join
+ * to join with output of right side (sub-query or table function that will be invoked for each row).
+ * Refer to DRILL-5999 for more information.
+ */
+public class DrillConformance extends SqlDelegatingConformance {
+
+  public DrillConformance() {
+    super(SqlConformanceEnum.DEFAULT);
+  }
+
+  public DrillConformance(SqlConformanceEnum flavor) {
+    super(flavor);
+  }
+
+  @Override
+  public boolean isApplyAllowed() {
+    return true;
+  }
+}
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillConformance.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.calcite.sql.validate.SqlConformanceEnum;
+import org.apache.calcite.sql.validate.SqlDelegatingConformance;
+
+/**
+ * Drill's SQL conformance is SqlConformanceEnum.DEFAULT except for method isApplyAllowed().
+ * Since Drill is going to allow OUTER APPLY and CROSS APPLY to allow each row from left child of Join
+ * to join with output of right side (sub-query or table function that will be invoked for each row).
+ * Refer to DRILL-5999 for more information.
+ */
+public class DrillConformance extends SqlDelegatingConformance {
```
vrozov: Is it necessary to have `DrillConformance` as a top level public class? Consider using an anonymous class directly in `DrillParserConfig`.

arina-ielchiieva:
@chunhui-shi could you please address @vrozov comment?
@vvysotskyi could you please alos take a look at PR?

vvysotskyi:
As I understand from DRILL-1921, cross join was prevented due to the `CannotPlanException` exception at the planning stage. 
Can we get the same problem using `APPLY`? If yes, should be discussed the possibility of adding some limitations for `APPLY`, for example, deny usage for the case when a filter is absent in the query etc.

parthchandra:
@vvysotskyi  The lateral join implementation committed in DRILL-6323 supports cross join for all conditions. The main difficulty in cross join is the management of the memory needed to produce a cartesian product (M x N). In cross apply though, the cross join is only computed on one row in the left side at a time (1 x N), since the left and right sides are correlated. This allows cross apply to operate with a very small amount of memory.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillConformance.java
chunhui-shi: I think DrillConformance should be an independent class since we rely on this class to define the conformance used in Drill and we might add something more in the future to enable some other syntax. Let me know if this answer your suggestion.

vrozov: If changes are introduced to `DrillConformance`, it can be refactored later to be a top-level class. For now, I suggest avoid optimizing for future requirements that may never materialize.

parthchandra: Given the number of times we have to update Calcite, the need to update this will occur more often than not, I think.

vrozov: Why not to introduce top-level class when needed. To override the behavior of the single method an anonymous class is more than sufficient.

vvysotskyi:
@parthchandra, thanks for the explanation. 
@chunhui-shi, thanks for making changes, +1

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillConformance.java
vvysotskyi: @vrozov, I don't think this problem is critical in overall.
But if using top-level public class will help to avoid extra refactoring changes in further PRs for implementing Lateral join and Unnest support, I think it should be left it as it is.

@vrozov, please let me know if we can merge this PR as it is.

vrozov: Personally, I don't see a need for the upper-level class in the future, so I implemented a different approach in https://github.com/apache/drill/compare/master...vrozov:DRILL-6321. A committer should decide what approach to follow, it is not that I block the PR with -1.

vrozov: Premature optimization for the future is the evil of the software development. It is possible to override any behavior using the anonymous class.

parthchandra: Vlad, you make a good point. From what I understand of Chunhui's explanation, SqlConformance is a class that logically stands on its own. That also mirrors what Calcite does. So I would recommend we let this stand.

vrozov: Parth, I don't see what a standalone class provides that an anonymous class does not. I would expect a standalone class in case it is used in multiple places and/or adds methods to an existing interface/class. `SqlConformance` is created/used in a single place and it is not necessary to jump to a different place/file to see the definition of Drill SQL conformance and how it is different from the Calcite default.

amansinha100:
Jumping in late here..I see the PR has 2 +1s and possibly a -0 from Vlad :-)   In this case it seems the anonymous class would have been sufficient for the current requirements but since the debate is more about the design pattern rather than functionality, I am inclined to commit this PR.

PR #1226, opened 2018-04-19
DRILL-3855: Enable FilterSetOpTransposeRule, DrillProjectSetOpTranspo…

vdiravka:
…seRule


arina-ielchiieva:
@amansinha100 / @vvysotskyi could you please review?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -238,26 +238,27 @@ protected DrillRel convertToRawDrel(final RelNode relNode) throws SqlUnsupported
       // HEP Directory pruning .
       final RelNode pruned = transform(PlannerType.HEP_BOTTOM_UP, PlannerPhase.DIRECTORY_PRUNING, relNode);
       final RelTraitSet logicalTraits = pruned.getTraitSet().plus(DrillRel.DRILL_LOGICAL);
+      final RelNode intermediateNode = transform(PlannerType.HEP, PlannerPhase.PRE_LOGICAL_PLANNING, pruned);
```
amansinha100: Did you consider applying this even before the DIRECTORY_PRUNING ?  If a filter is on dirN column and gets pushed on both sides of the union-all, then directory pruning can be done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
```diff
@@ -178,6 +178,12 @@ public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin>
           PlannerPhase.getPhysicalRules(context),
           getStorageRules(context, plugins, this));
     }
+  },
+
+  PRE_LOGICAL_PLANNING("Planning with Hep planner only for rules, which are failed for Volcano planner") {
```
amansinha100: I thought about whether a separate group of rules is needed instead of adding it to existing group of HEP rules.  But I suppose it is ok to create the separate group with the expectation that once Calcite-1271 is fixed for Volcano planner, we have the option to remove this group.   Thinking more about this..filter pushdown past a union-all is cost-safe since union-all output rows is equal or greater than input rows.  (it is not cost-safe for pushdown past union-distinct).

exec/java-exec/src/test/java/org/apache/drill/TestUnionAll.java
```diff
@@ -634,10 +634,14 @@ public void testFilterPushDownOverUnionAll() throws Exception {
         + "order by n_regionkey";
 
     // Validate the plan
-    final String[] expectedPlan = {".*Filter.*\n" +
-            ".*UnionAll.*\n" +
-            ".*Scan.*columns=\\[`n_regionkey`\\].*\n" +
-            ".*Scan.*columns=\\[`r_regionkey`\\].*"};
+    final String[] expectedPlan = {"Sort.*\n" +
```
amansinha100: Apart from these tests, do we have any unit tests that produce empty results once the filter is pushed on either side of the union-all ?  I expect that there could be some issues uncovered due to empty batch handling.  However, those issues could be treated as separate JIRA.

vdiravka: Looks like there are no issues related to it. I have added additional test case to verify it:
`select n_nationkey from (select n_nationkey, n_name, n_comment from cp."tpch/nation.parquet" union all select r_regionkey, r_name, r_comment  from cp."tpch/region.parquet") where n_nationkey > 4;`
When filter is pushed down the result from right side of UNION will be empty.

Generally speaking using UNION operator with empty data was resolved in [DRILL-4185](https://issues.apache.org/jira/browse/DRILL-4185) and we have test cases with empty data batches.

amansinha100:
Thanks for making the changes.  +1

PR #1227, opened 2018-04-19
DRILL-6236: batch sizing for hash join

ppadma:


exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchMemoryManager.java
```diff
@@ -147,6 +149,12 @@ public int update(int inputIndex, int outputPosition) {
     return getOutputRowCount();
   }
 
+  public int update(int inputIndex, int outputPosition, boolean useAggregate) {
+    // by default just return the outputRowCount
+    return update(inputIndex, outputPosition, false);
```
Ben-Zvi: Is this an infinite recursive call ??

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -300,13 +322,14 @@ public void setupHashTable() throws IOException, SchemaChangeException, ClassTra
...
     //Setup the underlying hash table
-
     // skip first batch if count is zero, as it may be an empty schema batch
     if (isFurtherProcessingRequired(rightUpstream) && right.getRecordCount() == 0) {
       for (final VectorWrapper<?> w : right) {
         w.clear();
       }
       rightUpstream = next(right);
+      // For build side, use aggregate i.e. average row width across batches
+      batchMemoryManager.update(RIGHT_INDEX, 0,true);
```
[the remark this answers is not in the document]
ppadma: There is a call to "next" right above the update.

ppadma:
@Ben-Zvi Thanks a lot for the review. updated PR with review comments taken care of. Please take a look.

Regarding spill files, here are my thoughts. 
For build side, I am using aggregate statistics i.e. average of all batches.  On probe side, I am using stats for each batch coming in and adjusting the output row count. So,  we can skip applying sizing for batches spilled from build side and continue to do what I am doing on the probe side. Once your code is merged in, I will refactor the code as needed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -147,7 +150,19 @@
...
+    LEFT_AVG_INPUT_ROW_BYTES,
+    LEFT_INPUT_RECORD_COUNT,
+    RIGHT_INPUT_BATCH_COUNT,
+    RIGHT_AVG_INPUT_BATCH_BYTES,
+    RIGHT_AVG_INPUT_ROW_BYTES,
+    RIGHT_INPUT_RECORD_COUNT,
+    OUTPUT_BATCH_COUNT,
+    AVG_OUTPUT_BATCH_BYTES,
+    AVG_OUTPUT_ROW_BYTES,
+    OUTPUT_RECORD_COUNT;
```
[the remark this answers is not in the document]
ppadma: It is relevant in the sense that they provide high level picture of amount of data being processed, memory usage etc. by each operator. This is also helpful when debugging trying to figure out what is going on.

ppadma:
@sohami thanks for the review. updated with review comments addressed. please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -560,6 +554,40 @@ public void close() {
     super.close();
   }
 
+  @Override
+  protected void updateBatchMemoryManagerStats() {
+    stats.setLongStat(Metric.LEFT_INPUT_BATCH_COUNT, batchMemoryManager.getNumIncomingBatches(LEFT_INDEX));
```
sohami: @ppadma - The main motive of moving the metrics inside JoinBatchMemoryManager was to avoid duplicate definition in all the BinaryRecordBatches. I think we should improve on OperatorMetricsRegistry rather than just overriding this method.
Or if you want you can create a new JIRA to track metrics related improvement.

Ben-Zvi:
Need to update the subject line of this PR: the '-' is missing between DRILL and 6236 (should be DRILL**-**6236) ; because of this missing '-' the PR is not listed in the Jira ....

ppadma:
@Ben-Zvi  my bad. I updated the title. but, it has not updated the JIRA. trying to figure this out.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
sohami: Should be fine.

ppadma:
@Ben-Zvi I manually added the PR link to the JIRA. all code review comments are addressed. can you look at the latest changes ?

ilooner:
@ppadma Please fix travis failure

ppadma:
@Ben-Zvi I merged with spill to disk changes and fixed all issues. please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinMemoryCalculatorImpl.java
```diff
@@ -234,6 +235,7 @@ public HashJoinState getState() {
     private int recordsPerPartitionBatchBuild;
     private int recordsPerPartitionBatchProbe;
     private int outputBatchNumRecords;
+    private int outputBatchSize;
     private Map<String, Long> buildValueSizes;
     private Map<String, Long> probeValueSizes;
     private Map<String, Long> keySizes;
```
ilooner: @ppadma I think outputBatchNumRecords, buildValueSizes, probeValueSizes, and keySizes are unused now since we are directly passing in the outputBatchSize. This is great since the calculator has been simplified. Could you also delete these unused parameters?

ppadma:
@Ben-Zvi I rebased and updated the PR. Please review the latest diffs.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinMemoryCalculatorImpl.java
```diff
@@ -448,8 +442,7 @@ private void calculateMemoryUsage()
         safetyFactor,
         reserveHash);
 
-      maxOutputBatchSize = computeMaxOutputBatchSize(buildValueSizes, probeValueSizes, keySizes,
-        outputBatchNumRecords, safetyFactor, fragmentationFactor);
+      maxOutputBatchSize = (long) (outputBatchSize * fragmentationFactor * safetyFactor);
```
Ben-Zvi: Maybe the "outputBatchSize" needs to be casted to (double) to ensure that the whole multiplication is performed as a double-multiplication.

ppadma: done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinProbeTemplate.java
```diff
@@ -262,6 +272,7 @@ private void executeProbePhase() throws SchemaChangeException {
                 probeBatch.getSchema());
             }
           case OK:
+            setTargetOutputCount(outgoingJoinBatch.getBatchMemoryManager().update(probeBatch, LEFT_INDEX,outputRecords));
```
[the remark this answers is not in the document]
ppadma: It will not make it bigger. It will look at remaining memory and adjust the row count based on that. 

    final long remainingMemory = Math.max(configOutputBatchSize - memoryUsed, 0);
    // These are number of rows we can fit in remaining memory based on new outgoing row width.
    final int numOutputRowsRemaining = RecordBatchSizer.safeDivide(remainingMemory, newOutgoingRowWidth);

exec/java-exec/src/main/java/org/apache/drill/exec/record/JoinBatchMemoryManager.java
```diff
@@ -85,13 +71,50 @@ public int update(int inputIndex, int outputPosition) {
   }
 
   @Override
-  public RecordBatchSizer.ColumnSize getColumnSize(String name) {
```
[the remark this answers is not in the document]
ppadma: Because it is redundant. super class is doing the same thing.

exec/java-exec/src/main/java/org/apache/drill/exec/record/JoinBatchMemoryManager.java
```diff
@@ -85,13 +71,50 @@ public int update(int inputIndex, int outputPosition) {
...
-    RecordBatchSizer leftSizer = getRecordBatchSizer(LEFT_INDEX);
-    RecordBatchSizer rightSizer = getRecordBatchSizer(RIGHT_INDEX);
+  public int update(int inputIndex, int outputPosition, boolean useAggregate) {
+    switch (inputIndex) {
+      case LEFT_INDEX:
```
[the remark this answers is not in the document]
ppadma: I rearranged the code. Got rid of left and right. Instead, using array called rowWidth which can be indexed by input index.  It is better now. Unfortunately, each operator calls update with different parameters. So, we have different versions of the same function.

Right is not always "use aggregate". 
It is based on the operator. For example, for merge join, we do not use aggregate. It is batch by batch.

PR #1228, opened 2018-04-19
DRILL-6307: Handle empty batches in record batch sizer correctly

ppadma:
When we get empty batch, record batch sizer calculates row width as zero. In that case, we do not do accounting and memory allocation correctly for outgoing batches. 

For ex., for outer left join, if right side batch is empty, we still have to include the right side columns as null in outgoing batch. Say first batch is empty. Then, for outgoing, we allocate empty vectors with zero capacity.  When we read the next batch with data, we will end up going through realloc loop as we write values. Also, if we use right side row width as 0 in outgoing row width calculation, number of rows (to include in the outgoing batch) we will calculate will be higher and later when we get a non empty batch, we might exceed the memory limits. 

This PR tries to address these problems by allocating memory based on std size for empty input batch. Uses allocation width as width of the batch in number of rows calculation for binary operators. For unary operators, this is not a problem since we drop empty batches without doing any processing.


exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -50,7 +50,7 @@
 public class RecordBatchSizer {
   private static final int OFFSET_VECTOR_WIDTH = UInt4Vector.VALUE_WIDTH;
   private static final int BIT_VECTOR_WIDTH = UInt1Vector.VALUE_WIDTH;
-  private static final int STD_REPETITION_FACTOR = 10;
+  public static final int STD_REPETITION_FACTOR = 10;
```
paul-rogers: This is another of those silly fudge factors that really have no meaning. The value of 10 came from the vector allocation code in `AllocationHelper` (or I thought it did, the magic number there is 5.)

Maybe move this to `AllocationHelper` and set it to 5, then use it here and in `AllocationHelper` so we use a consistent guess everywhere.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -277,18 +286,29 @@ public boolean isRepeatedList() {
...
+        width = getAllocSizePerEntry() - OFFSET_VECTOR_WIDTH;
 
         // Subtract out the bits (is-set) vector width
-        if (metadata.getDataMode() == DataMode.OPTIONAL) {
+        if (isOptional) {
           width -= BIT_VECTOR_WIDTH;
         }
+
+        if (isRepeated && getValueCount() == 0) {
+          return (safeDivide(width, STD_REPETITION_FACTOR));
```
paul-rogers: If the value count is zero, but the row count is non-zero, then a very low repetition rate is more realistic than 10. In earlier drafts, I found the repetition rate had to be a float since, in some data, the rate is something like 0.7 or 1.4. Rounding to an integer caused quite an error when multiplying by, say, 50K rows.

Any reason we can't use the actual computed amount here? If we really have 1 or 2 rows, then a guess of 10 is fine. But, if we have 60K rows, with an actual estimate of 0, then guessing 10 will allocate 600K values when we probably needed close to 0. (Unless I'm missing something.)

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -277,18 +286,29 @@ public boolean isRepeatedList() {
...
 
-      return (safeDivide(width, cardinality));
+      return (safeDivide(width, getEntryCardinalityForAlloc()));
+    }
+
+    /**
+     * This is the average per entry cardinality, used for vector allocation.
+     */
+    private float getEntryCardinalityForAlloc() {
+      return getCardinality() == 0 ? (isRepeated ? STD_REPETITION_FACTOR : 1) :getCardinality();
```
paul-rogers: I'm a bit curious: under what scenario do we want to allocate vectors given no input rows? Can we just wait to see data before we do our allocations? It is very hard to make reasonable estimates of future size based on no data...

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -536,6 +556,11 @@ public ColumnSize getColumn(String name) {
    */
   private int netRowWidth;
   private int netRowWidthCap50;
+
+  /**
+   * actual row size if input is not empty. Otherwise, standard size.
+   */
+  private int rowAllocSize;
```
paul-rogers: I wonder if this all would be clearer if we handed it at size estimation time. If the row count is 0, set up everything using the standard sizes. (Note: the whole reason this class exists is that the standard sizes turned out to be *very* poor estimators of actual size.)

So, if we have no data, guess the same size as `AllocationHelper`, else use real sizes.

And, again the question: under what situation do we want to use the sizer if we don't actually have any data? For the first batch, if no data, just throw away the empty batch and don't size it. Turn around and get another until we receive a non-empty batch.

If we've already received at least one non-empty batch, then we receive an empty batch, we should just retain the estimates from the non-empty batch since they will be much better than just making up numbers.

exec/java-exec/src/test/java/org/apache/drill/exec/record/TestRecordBatchSizer.java
```diff
@@ -45,7 +45,7 @@
   private void verifyColumnValues(ColumnSize column, int stdDataSizePerEntry, int stdNetSizePerEntry,
                                   int dataSizePerEntry, int netSizePerEntry, int totalDataSize,
                                   int totalNetSize, int valueCount, int elementCount,
-                                  int estElementCountPerArray, boolean isVariableWidth) {
+                                  int cardinality, boolean isVariableWidth) {
```
paul-rogers: Maybe explain the cardinality here since I've seen it (or have used it) in three different ways:

* Type cardinality: (Required: 1, Optional: (0, 1), Repeated: *)
* Batch cardinality: the number of rows in a batch. (Or, equivalently, the number of top-level values in a vector.)
* Array cardinality: the number of values in an array (a single top-level value which is repeated)

"Cardinality" is a good word, may just need disambiguation.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
ppadma: done. using 5 in both places now.

ppadma:
@paul-rogers Paul, I addressed code review comments. Can you take a look when you get a chance ?

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
ppadma: You are right. row count can be non-zero and valueCount can still be zero.  Since this is intended for empty batch case, changed this to check for row count instead.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
paul-rogers: Makes sense, but why would a batch be empty unless that path hit EOF? Otherwise, the batch might be due to an empty input file. We'd just skip it and move to the next batch until we find one with data. Any reason the "get next batch" code can just loop to be "get next non-empty batch" instead? Otherwise, we can't really do any effective batch sizing as we have no data to go on...

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
paul-rogers: Unless I'm missing something, we can't move forward on a join if one side is empty: we won't know if we have the rows we need. Consider a merge join (simplest). The left gets some data, but the right is empty. We can't proceed unless the right hit EOF. Otherwise, we don't know if we have a match or not for the first left row.

We need to read another right batch and keep going until we either hit EOF (no matching rows) or get some data.

Once we have some data, we can go row-by-row to see if we have a left-only, right-only, or matching set of rows. If we get to EOF on either side, we know that their are no matches for the other side.

What we do in the no-match case depends on whether we are doing LEFT OUTER, RIGHT OUTER or an INNER join.

The point is, we can't make progress until we get that non-empty right batch (in this example). So, no reason to allocate space based on an empty batch (unless the entire input is empty) because we'll need to find a non-empty (or EOF) batch anyway.

ppadma: actually, this is a problem only for lateral join. In lateral join, right side will work on one row at a time from left side. If right side produces zero rows because of empty array or some other reason, for left outer join, we still have to finish cross join for that row from left side by having nulls for right side columns. Then, we go to next row on left side, continuing to work on filling the output batch till it is full.  What that means is we have to allocate vectors based on that first batch on right side, which can be empty.

paul-rogers: I see. In this case, however, arrays (repeated values) will be empty. If we have 10 such rows, there is no reason to have 50 "inner" values. Also, for VarChar, no values will be stored; all columns will be null. (If we are handling non-null columns, then the non-null VarChar will be an empty string.)

So, we probably need a bit of a special case: prepare data for a run of null rows (with arrays and VarChar of length 0) vs. take our best guess with no knowledge at all about lengths (which may be non-empty.)

Probably not a huge issue if you only need to handle a single row. But, creating a batch with only one row will cause all kinds of performance issues downstream. (I found that out the hard way when a but in sort produced a series of one-row batches...)

ppadma: @paul-rogers yes, this PR is trying to do the second thing i.e. make best guess with no knowledge at all about lengths. Also, right side can produce no rows for reasons other than empty array i.e. for ex. we might be filtering everything out after unnest.  
Please let me know what you think about the approach.

paul-rogers: @ppadma, not much more to add. If the code requires you do estimates based on no information, we won't get very good estimates. But, if we know we have 0 rows, then that itself is a good estimate of the size we'll need.

If there is a way to improve the estimate, I'm guessing you'll find it as work proceeds and you seem more examples and test cases.

PR #1231, opened 2018-04-20
DRILL-6342: Fix schema path unIndexed method to return correct value

arina-ielchiieva:
Details in [DRILL-6342](https://issues.apache.org/jira/browse/DRILL-6342).


arina-ielchiieva:
@parthchandra since you reviewed original PR about filter push down for complex parquet fields, could you please also review the follow up fix?

PR #1234, opened 2018-04-20
DRILL-5927: Fixed memory leak in TestBsonRecordReader, and sped up th…

ilooner:
…e test.


ilooner:
@parthchandra Please review

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -272,6 +276,9 @@ public static void cleanUp() {
     } catch (Exception e) {
 
     }
+
+    buffer.close();
```
vrozov: buffer.release()?

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -49,17 +50,20 @@
...
   private static VectorContainerWriter writer;
   private static TestOutputMutator mutator;
+  private static DrillBuf buffer;
   private static BsonRecordReader bsonReader;
 
   @BeforeClass
   public static void setUp() {
-    BufferAllocator bufferAllocator = getDrillbitContext().getAllocator();
-    mutator = new TestOutputMutator(bufferAllocator);
+    allocator = new RootAllocator(100_000_000);
```
vrozov: Is it necessary to reserve 100MB for allocator? Does it allocate more than 1K?

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -49,17 +50,20 @@
 import org.junit.BeforeClass;
 import org.junit.Test;
 
-public class TestBsonRecordReader extends BaseTestQuery {
+public class TestBsonRecordReader {
+  private static BufferAllocator allocator;
```
vrozov: What is a reason to define `BufferAllocator` and other fields as static? What if tests are executed in parallel?

ilooner: JUnit requires tests annotated with @BeforeClass and @AfterClass to be static so these variables have to be static. Currently the Drill unit tests do not support concurrent execution of methods within a test class within the same process. Drill has surefire launch multiple test processes and the the test classes to execute are divided among the test processes. Within each forked surefire process tests are executed sequentially, so this will not be an issue.

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
ilooner: Currently close() just calls release(). I will update to use release(), but in general what is the best practice regarding when to call close() and when to call release()?

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
ilooner: After testing just now, the lowest I could push this was 9,000,000 bytes. Going to 8,000,000 bytes causes an OOM. It looks like the BsonRecordReader allocates a bunch of value vectors. I will adjust this to 9,000,000.

```
org.apache.drill.exec.exception.OutOfMemoryException: Unable to allocate buffer of size 8192 due to memory limit (800000). Current allocation: 795648

	at org.apache.drill.exec.memory.BaseAllocator.buffer(BaseAllocator.java:236)
	at org.apache.drill.exec.memory.BaseAllocator.buffer(BaseAllocator.java:211)
	at org.apache.drill.exec.vector.UInt1Vector.reallocRaw(UInt1Vector.java:262)
	at org.apache.drill.exec.vector.UInt1Vector.reAlloc(UInt1Vector.java:251)
	at org.apache.drill.exec.vector.UInt1Vector$Mutator.setSafe(UInt1Vector.java:480)
	at org.apache.drill.exec.vector.NullableVarCharVector$Mutator.setSafe(NullableVarCharVector.java:608)
	at org.apache.drill.exec.vector.complex.impl.NullableVarCharWriterImpl.write(NullableVarCharWriterImpl.java:110)
	at org.apache.drill.exec.store.bson.BsonRecordReader.writeString(BsonRecordReader.java:276)
	at org.apache.drill.exec.store.bson.BsonRecordReader.writeBinary(BsonRecordReader.java:205)
	at org.apache.drill.exec.store.bson.BsonRecordReader.writeToListOrMap(BsonRecordReader.java:117)
	at org.apache.drill.exec.store.bson.BsonRecordReader.write(BsonRecordReader.java:75)
	at org.apache.drill.exec.store.bson.TestBsonRecordReader.testBinaryTypes(TestBsonRecordReader.java:244)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)

```

ilooner:
@vrozov addressed comments

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
vrozov: What is a reason `allocator` and other variables are initialized in `@BeforeClass`? Can they be initialized in `@Before` instead?

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
vrozov: IMO `close()` should be used within try-with-resources only (no explicit calls to `close()`).

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -49,17 +50,20 @@
...
   private static VectorContainerWriter writer;
   private static TestOutputMutator mutator;
+  private static DrillBuf buffer;
   private static BsonRecordReader bsonReader;
 
   @BeforeClass
   public static void setUp() {
-    BufferAllocator bufferAllocator = getDrillbitContext().getAllocator();
-    mutator = new TestOutputMutator(bufferAllocator);
+    allocator = new RootAllocator(9_000_00);
```
vrozov: It does not sound right as there is not much data used in the test.  I'd suggest filing a JIRA to look into why it takes 900 KB or almost 1 MB to hold data and using unlimited allocator (the same as before).

ilooner:
@vrozov addressed comments

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -45,21 +46,24 @@
...
+  private DrillBuf buffer;
+  private BsonRecordReader bsonReader;
 
-  @BeforeClass
-  public static void setUp() {
-    BufferAllocator bufferAllocator = getDrillbitContext().getAllocator();
-    mutator = new TestOutputMutator(bufferAllocator);
+  @Before
+  public void setUp() {
+    allocator = new RootAllocator(400_000);
```
vrozov: Two minor suggestions:
- use `RootAllocator` with an unlimited amount of memory as limiting it to 400K does not test `BsonRecordReader` and depends more on how Vectors are allocated.
- instead of working with `DrillBuf` directly, use `BufferManager`.

ilooner: @vrozov Made the changes

vrozov:
LGTM

ilooner:
@parthchandra Please let me know if you have any comments.

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -45,21 +47,24 @@
...
+  private BufferManager bufferManager;
+  private BsonRecordReader bsonReader;
 
-  @BeforeClass
-  public static void setUp() {
-    BufferAllocator bufferAllocator = getDrillbitContext().getAllocator();
-    mutator = new TestOutputMutator(bufferAllocator);
+  @Before
+  public void setUp() {
+    allocator = new RootAllocator(Long.MAX_VALUE);
```
parthchandra: Any reason why you had to change to RootAllocator? Should not really be necessary.

PR #1235, opened 2018-04-21
DRILL-6336: Inconsistent method name.

BruceKuiLiu:
Change the method name "append" to "print" since its body code is an method invocation "writer.print( s )". The method named "print" should be more clear.
The method name "append" is labeled as "@Deprecated" in case of breaking backward compatibility.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
@@ -835,18 +835,18 @@ public class ClassGenerator<T>{
     public String toString() {
       DebugStringBuilder buf = new DebugStringBuilder(this);
       if (isConstant()) {
-        buf.append("const ");
+        buf.print("const ");
       }
-      buf.append(holder.type().fullName())
-        .append(" ")
-        .append(holder.name())
-        .append(", ")
-        .append(type.getMode().name())
-        .append(" ")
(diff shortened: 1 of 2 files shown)


arina-ielchiieva:
@BruceKuiLiu it is not public interface, no need to deprecate the method. You can simply replace.

vrozov:
There is no need to expose implementation details as part of the class API. Whether `DebugStringBuilder` uses `PrintWriter.print()` or something else to implement `append()` must be hidden from `DebugStringBuilder` consumers. Method name should never depend on details of implementation as the implementation may change, but API should not.

arina-ielchiieva:
@vrozov so you suggest to leave as is, correct?
@paul-rogers since you have originally added `DebugStringBuilder`, do you agree?

vrozov:
My take is that "append" is more common for classes with the similar functionality, see for example `ToStringBuilder`. As there is no added benefit of using "print" vs "append", my recommendation is to keep "append" as is and see if `DebugStringBuilder` can be replaced with the `ToStringBuilder`.

arina-ielchiieva:
@BruceKuiLiu could you please check if @vrozov suggestion is feasible?

BruceKuiLiu:
The target of this class is  to build a debug string for an object.
Change the class name may break this target.
I agree with @vrozov  and @paul-rogers. Considering the class is a _builder_, `append()` should be better than `print()`.

PR #1236, opened 2018-04-21
DRILL-6347: Inconsistent method name "field".

BruceKuiLiu:
The method is names as "field", but the method is mainly doing appending. So that, rename the method as "append" should be better.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/PrelVisualizerVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/PrelVisualizerVisitor.java
@@ -87,11 +87,11 @@ public class PrelVisualizerVisitor
 
     }
 
-    public void field(String label, boolean value) {
-      field(label, Boolean.toString(value));
+    public void visitField(String label, boolean value) {
+      visitField(label, Boolean.toString(value));
     }
 
-    private void field(String label, String value) {
+    private void visitField(String label, String value) {
       indent();
       out.append(label)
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/PrelVisualizerVisitor.java
```diff
@@ -88,10 +88,10 @@ public void endFields() {
     }
 
     public void field(String label, boolean value) {
-      field(label, Boolean.toString(value));
+      append(label, Boolean.toString(value));
     }
 
-    private void field(String label, String value) {
+    private void append(String label, String value) {
```
arina-ielchiieva: `appendField`?

HanumathRao: Thanks Bruce for the changes. As Arina pointed out, appendField would be more meaningful in this context.

vrozov:
IMO, neither `append` or `appendField` is a good choice (otherwise it is necessary to change `startNode/endNode` to `appendStart/EndNode`). It is either `field` or `visitField` and should follow visitor design pattern. It may be also good to change `listField` to the same name used for `Boolean` and `String`.

arina-ielchiieva:
@BruceKuiLiu could you please address @vrozov comments?

BruceKuiLiu:
@vrozov Thanks.

vrozov:
LGTM. Please squash commits.

PR #1237, opened 2018-04-23
DRILL-6348: Fixed code so that Unordered Receiver reports its memory …

sachouche:
Problem Description -
- The Unordered Receiver doesn't report any memory usage because the RPC infrastructure associated the allocated memory ownership to the minor fragment container

Proposed Fix -
- Modified the code to change the received DrillBuf memory ownership from the parent fragment to the operator (as discussed with @parthchandra)
- Made sure that the buffer accounting logic is preserved

@parthchandra, can you please review this pull request?

Thanks!

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/MergingReceiverCreator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/MergingReceiverCreator.java
@@ -42,7 +42,7 @@ public class MergingReceiverCreator implements BatchCreator<MergingReceiverPOP> 
     IncomingBuffers bufHolder = context.getBuffers();
 
     assert bufHolder != null : "IncomingBuffers must be defined for any place a receiver is declared.";
-    RawBatchBuffer[] buffers = bufHolder.getBuffers(receiver.getOppositeMajorFragmentId());
+    RawBatchBuffer[] buffers = bufHolder.getCollector(receiver.getOppositeMajorFragmentId()).getBuffers();
 
     return new MergingRecordBatch(context, receiver, buffers);
   }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
@@ -86,6 +86,9 @@ public class UnorderedReceiverBatch implements CloseableRecordBatch {
     this.stats = oContext.getStats();
     this.stats.setLongStat(Metric.NUM_SENDERS, config.getNumSenders());
     this.config = config;
+
+    // Register this operator's buffer allocator so that incoming buffers are owned by this allocator
+    context.getBuffers().getCollector(config.getOppositeMajorFragmentId()).setAllocator(oContext.getAllocator());
   }
 
   @Override
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverCreator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverCreator.java
@@ -37,7 +37,7 @@ public class UnorderedReceiverCreator implements BatchCreator<UnorderedReceiver>
     IncomingBuffers bufHolder = context.getBuffers();
     assert bufHolder != null : "IncomingBuffers must be defined for any place a receiver is declared.";
 
-    RawBatchBuffer[] buffers = bufHolder.getBuffers(receiver.getOppositeMajorFragmentId());
+    RawBatchBuffer[] buffers = bufHolder.getCollector(receiver.getOppositeMajorFragmentId()).getBuffers();
     assert buffers.length == 1;
     RawBatchBuffer buffer = buffers[0];
     return new UnorderedReceiverBatch(context, buffer, receiver);
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/AbstractDataCollector.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/AbstractDataCollector.java
@@ -22,6 +22,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerArray;
 
 import org.apache.drill.common.AutoCloseables;
+import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.proto.BitControl.Collector;
 import org.apache.drill.exec.record.RawFragmentBatch;
@@ -37,6 +38,8 @@ public abstract class AbstractDataCollector implements DataCollector {
   private final int incomingStreams;
   protected final RawBatchBuffer[] buffers;
   protected final ArrayWrappedIntIntMap fragmentMap;
+  /** Allocator which owns incoming batches */
+  protected BufferAllocator ownerAllocator;
 
   /**
    * @param parentAccounter
@@ -53,6 +56,7 @@ public abstract class AbstractDataCollector implements DataCollector {
     this.parentAccounter = parentAccounter;
     this.remainders = new AtomicIntegerArray(incomingStreams);
     this.oppositeMajorFragmentId = collector.getOppositeMajorFragmentId();
+    this.ownerAllocator = context.getAllocator();
     // Create fragmentId to index that is within the range [0, incoming.size()-1]
     // We use this mapping to find objects belonging to the fragment in buffers and remainders arrays.
     fragmentMap = new ArrayWrappedIntIntMap();
@@ -114,6 +118,19 @@ public abstract class AbstractDataCollector implements DataCollector {
   @Override
   public void close() throws Exception {
     AutoCloseables.close(buffers);
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public BufferAllocator getAllocator() {
+    return this.ownerAllocator;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public void setAllocator(BufferAllocator allocator) {
+    Preconditions.checkArgument(allocator != null, "buffer allocator cannot be null");
+    this.ownerAllocator = allocator;
   }
 
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/DataCollector.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/DataCollector.java
@@ -19,13 +19,25 @@ package org.apache.drill.exec.work.batch;
 
 import java.io.IOException;
 
+import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.record.RawFragmentBatch;
 
-interface DataCollector extends AutoCloseable {
+public interface DataCollector extends AutoCloseable {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DataCollector.class);
   public boolean batchArrived(int minorFragmentId, RawFragmentBatch batch) throws IOException ;
   public int getOppositeMajorFragmentId();
   public RawBatchBuffer[] getBuffers();
   public int getTotalIncomingFragments();
   public void close() throws Exception;
+  /**
+   * Enables caller (e.g., receiver) to attach its buffer allocator to this Data Collector in order
+   * to claim ownership of incoming batches; by default, the fragment allocator owns these batches.
+   *
+   * @param allocator operator buffer allocator
+   */
+  void setAllocator(BufferAllocator allocator);
+  /**
+   * @return allocator
+   */
+  BufferAllocator getAllocator();
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/IncomingBuffers.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/IncomingBuffers.java
@@ -27,6 +27,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.apache.drill.common.AutoCloseables;
 import org.apache.drill.common.concurrent.AutoCloseableLock;
 import org.apache.drill.exec.exception.FragmentSetupException;
+import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.proto.BitControl.Collector;
 import org.apache.drill.exec.proto.BitControl.PlanFragment;
@@ -103,8 +104,11 @@ public class IncomingBuffers implements AutoCloseable {
             Arrays.toString(collectorMap.values().toArray())));
       }
 
+      // Use the Data Collector's buffer allocator if set, otherwise the fragment's one
+      BufferAllocator ownerAllocator = collector.getAllocator();
+
       synchronized (collector) {
-        final RawFragmentBatch newRawFragmentBatch = incomingBatch.newRawFragmentBatch(context.getAllocator());
+        final RawFragmentBatch newRawFragmentBatch = incomingBatch.newRawFragmentBatch(ownerAllocator);
         boolean decrementedToZero = collector
             .batchArrived(incomingBatch.getHeader().getSendingMinorFragmentId(), newRawFragmentBatch);
         newRawFragmentBatch.release();
@@ -125,8 +129,8 @@ public class IncomingBuffers implements AutoCloseable {
     return rem;
   }
 
-  public RawBatchBuffer[] getBuffers(int senderMajorFragmentId) {
-    return collectorMap.get(senderMajorFragmentId).getBuffers();
+  public DataCollector getCollector(int senderMajorFragmentId) {
+    return collectorMap.get(senderMajorFragmentId);
   }
 
   public boolean isDone() {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContextImpl.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContextImpl.java
@@ -428,13 +428,16 @@ public class FragmentContextImpl extends BaseFragmentContext implements Executor
   public void close() {
     waitForSendComplete();
 
+    // Close the buffers before closing the operators; this is needed as buffer ownership
+    // is attached to the receive operators.
+    suppressingClose(buffers);
+
     // close operator context
     for (OperatorContextImpl opContext : contexts) {
       suppressingClose(opContext);
     }
 
     suppressingClose(bufferManager);
-    suppressingClose(buffers);
     suppressingClose(allocator);
   }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
@@ -136,6 +136,9 @@ public class MergingRecordBatch extends AbstractRecordBatch<MergingReceiverPOP> 
     this.config = config;
     this.inputCounts = new long[config.getNumSenders()];
     this.outputCounts = new long[config.getNumSenders()];
+
+    // Register this operator's buffer allocator so that incoming buffers are owned by this allocator
+    context.getBuffers().getCollector(config.getOppositeMajorFragmentId()).setAllocator(oContext.getAllocator());
   }
 
   @SuppressWarnings("resource")


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -182,13 +184,18 @@ public IterOutcome next() {
         return IterOutcome.OUT_OF_MEMORY;
       }
 
+      // Transfer the ownership of this raw-batch to this operator for proper memory statistics reporting
+      batch = batch.transferBodyOwnership(oContext.getAllocator());
```
HanumathRao: @sachouche  Thank you for the changes. I was just contemplating if the below change
batch = batch.transferBodyOwnership(oContext.getAllocator()); 
needs to be moved before the check at line number 183.

sachouche: @HanumathRao,

- This will not affect the outcome since the unordered-receiver is a child of the context allocator (the minor fragment)
- Doing the check earlier seemed cleaner as there is no point of changing ownership when there is already an out-of-memory condition
- I think the code was written this way since implicitly, the network handlers are receiving the batch and then performing a change of ownership (from RPC to context allocator); this step could lead to an out-of-memory condition at the fragment level
- If your point is about reporting, that is attributing the OOM condition to the child operator, then I guess I don't have a problem with that.
@parthchandra what do you think?

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
```diff
@@ -77,4 +83,46 @@ public long getByteCount() {
...
+    int writerIndex               = body.writerIndex();
+    TransferResult transferResult = body.transferOwnership(targetAllocator);
+
+    // Set the index and increment reference count
+    transferResult.buffer.writerIndex(writerIndex);
+
+    // Clear the current Drillbuffer since caller will perform release() on the new one
+    body.release();
+
+    return new RawFragmentBatch(getHeader(), transferResult.buffer, getSender(), false);
```
vrozov: Why is it necessary to return new `RawFragmentBatch` instead of setting `body` to `transferResult.buffer`?

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
```diff
@@ -77,4 +83,46 @@ public long getByteCount() {
...
+    }
+
+    if (!body.getLedger().isOwningLedger()
+     || body.getLedger().isOwner(targetAllocator)) {
+
+      return this;
+    }
+
+    int writerIndex               = body.writerIndex();
+    TransferResult transferResult = body.transferOwnership(targetAllocator);
```
vrozov: There is no check that the transfer is within new allocator limits, is it guaranteed to succeed?

exec/memory/base/src/main/java/org/apache/drill/exec/memory/AllocationManager.java
```diff
@@ -253,10 +261,12 @@ public boolean transferBalance(final BufferLedger target) {
           target.historicalLog.recordEvent("incoming(from %s)", owningLedger.allocator.name);
         }
 
-        boolean overlimit = target.allocator.forceAllocate(size);
+        // Release first to handle the case where the current and target allocators were part of the same
+        // parent / child tree.
         allocator.releaseBytes(size);
+        boolean allocationFit = target.allocator.forceAllocate(size);
```
vrozov: move `owningLedger = target;` before `return target.allocator.forceAllocate(size)` as there is no check for the result of the call anyway.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -153,8 +153,10 @@ private RawFragmentBatch getNextBatch() throws IOException {
   public IterOutcome next() {
     batchLoader.resetRecordCount();
     stats.startProcessing();
+
+    RawFragmentBatch batch = null;
     try{
-      RawFragmentBatch batch;
+
```
vrozov: Consider moving try block to a separate function.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -182,13 +184,18 @@ public IterOutcome next() {
...
+      batch = batch.transferBodyOwnership(oContext.getAllocator());
+
       final RecordBatchDef rbd = batch.getHeader().getDef();
       final boolean schemaChanged = batchLoader.load(rbd, batch.getBody());
       // TODO:  Clean:  DRILL-2933:  That load(...) no longer throws
       // SchemaChangeException, so check/clean catch clause below.
       stats.addLongStat(Metric.BYTES_RECEIVED, batch.getByteCount());
 
       batch.release();
+      batch = null;
```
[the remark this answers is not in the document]
vrozov: Why can't it be done in finally?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -153,8 +153,10 @@ private RawFragmentBatch getNextBatch() throws IOException {
   public IterOutcome next() {
     batchLoader.resetRecordCount();
     stats.startProcessing();
+
+    RawFragmentBatch batch = null;
```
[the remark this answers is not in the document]
vrozov: It is not obvious that `body` is `null` when the `if` condition is met. Additionally, what if a release logic changes in future? It is a bad practice to rely on the details of implementation.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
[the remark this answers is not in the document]
sachouche: @vrozov 

My bad, the highlight was on the "batch = null" code; I guess, you meant why can't we move the whole release logic within the finally block. I agree with your proposal as delaying a bit the release phase doesn't hurt u in this case. I will make the change.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
vrozov: I don't see where RawFragmentBatch is cached. Is not it removed from a queue using poll()?

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
vrozov: But what if it is an over limit?

exec/memory/base/src/main/java/org/apache/drill/exec/memory/AllocationManager.java
vrozov: If this happens, is not there a problem that the old allocator already released the memory? In any case, won't runtime exception cancel the query anyway and all allocators will be closed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
vrozov: create function `getNextNotEmptyBatch()` that calls `getNextBatch()` and either returns not empty batch or `null`.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
sachouche: Look at the IncomingBuffers.batchArrived:
- It holds a reference to the RawFragmentBatch
- Holds a lock on the parent collector for insertion
- Though, the collector enqueue method itself is not synchronized on the "this" object
- Instead, it relies on the synchronization provided as part of the RawBatchBuffer queue
- This means a getNext() call will be able to dequeue a RawFragmentBatch instance
- Concurrently, the IncomingBuffers.batchArrived will invoke release on the cached reference

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
sachouche: - After a call to getNext(), the operator code checks whether the transfer has succeeded
- So that an OOM condition is reported within the fragment
- When the change of ownership happens from parent to child, there should not be any OOM condition since this is internal account

For code clarity, I can improve the javadoc to ask the caller to check for an OOM condition (similarly to the BaseAlocator behavior).

exec/memory/base/src/main/java/org/apache/drill/exec/memory/AllocationManager.java
sachouche: What about debugging:
- An operation fails
- And yet there is a change of ownership
 
@vrozov, I rather not make this change as I strongly believe the change of ownership should happen only on success.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
sachouche: will do. Thanks for the suggestion.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -149,25 +149,32 @@ private RawFragmentBatch getNextBatch() throws IOException {
     }
   }
 
+  private RawFragmentBatch getNextNotEmptyBatch() throws IOException {
```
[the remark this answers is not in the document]
sachouche: why is that? this method is used once and doing so will make us duplicate exception handling code. I would appreciate if you provide reason(s) when you make suggestions to avoid the back and forth.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -149,25 +149,32 @@ private RawFragmentBatch getNextBatch() throws IOException {
...
+    RawFragmentBatch batch;
+    try {
+      stats.startWait();
```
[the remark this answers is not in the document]
sachouche: Sure, but does it really matter or is it your personal preference?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -149,25 +149,32 @@ private RawFragmentBatch getNextBatch() throws IOException {
...
+      batch = getNextBatch();
+
+      // skip over empty batches. we do this since these are basically control messages.
+      while (batch != null && batch.getHeader().getDef().getRecordCount() == 0
```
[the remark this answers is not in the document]
sachouche: Sure. 

FYI - We cannot release the batch within this method.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -201,6 +208,11 @@ public IterOutcome next() {
       context.getExecutorState().fail(ex);
       return IterOutcome.STOP;
     } finally {
+
+      if (batch != null) {
+        batch.release();
+        batch = null;
```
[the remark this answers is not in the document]
sachouche: This pattern ensures that after release the batch cannot be accessed anymore (e.g., additions to the current code).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -149,25 +149,32 @@ private RawFragmentBatch getNextBatch() throws IOException {
...
-      } finally {
-        stats.stopWait();
-      }
+    RawFragmentBatch batch = null;
+    try {
 
+      batch = getNextNotEmptyBatch();
       first = false;
 
       if (batch == null) {
```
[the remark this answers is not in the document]
sachouche: Sure.

sachouche:
@vrozov, your observation is valid, we need more JIRAs to fix the reporting problem
**Current Fix**
- At this time, the UnorderedReceiver didn't account for any consumed memory
- This fix, taxes the operator only when it consumes buffers

**Potential Enhancements**
Solution I
- Create a new fragment child Allocator which will own the received (not yet consumed) batches
- Improve the UI to report this allocator size
- This solution is simple and complementary to the current work

Solution II
- Have the receiver operator drain the batch queue
- Essentially, the receiver will have a private queue from where to consume batches
- I personally don't like this solution as it prevents us from improving the Drill network protocol
- Draining the batches for the sake of reporting will make it harder for the network layer to prefetch batches in an optimal manner; the queue size should be an indicator on how many pending batches there are.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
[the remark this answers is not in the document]
vrozov: Yes, it does matter. If something goes wrong during `startWait()` there should be no call to `stopWait()`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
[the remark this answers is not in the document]
vrozov: Please explain why it can't be released. Note that the release is done for the batch that will *not* be returned from the method.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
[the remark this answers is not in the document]
vrozov: No, there is no need to handle `IOException` twice. There are no other methods that throw `IOException`. `SchemaChangeException` is actually never thrown. Please see *TODO* comment and DRILL-2933.

vrozov:
IMO, it is better not to report memory usage at all compared to reporting a wrong number. In case incoming batches are accumulated in a queue, they should be reported as owned by a receiver. Taking ownership just before passing a batch to the next operator does not sound right to me.

I don't think it is necessary to create new fragment child allocator. Receiver allocator should be used instead of fragment allocator when an incoming batch is placed into a queue.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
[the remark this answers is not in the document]
vrozov: OK, but note that additions to the current code can also delete `batch = null` assignment and that `batch` can be used after `release()` call. For example, with the current implementation, `batch.getHeader()` is perfectly valid after the batch was released.

exec/memory/base/src/main/java/org/apache/drill/exec/memory/AllocationManager.java
vrozov: In this case, changing the order of `forceAllocate()` and `releaseBytes()` is incorrect as ownership is not changed, but the old owner does not account for that memory anymore.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
vrozov: This actually brings a question why `newRawFragmentBatch` is released in `IncomingBuffers.batchArrived()` instead of releasing `transferredBuffer` after `RawFragmentBatch` is constructed in `newRawFragmentBatch`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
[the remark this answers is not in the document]
sachouche: The point of this pattern is that if you would like to continue using this object then be prepared to know what can and what cannot be used.

exec/memory/base/src/main/java/org/apache/drill/exec/memory/AllocationManager.java
sachouche: - The change of order is an optimization for a parent / child relationship as if we don't release first, then we could unnecessarily go over the memory budget (double counting).
- The force-alloc() / free() failures should never happen on normal conditions; when they do, the best thing to do is to exit. I still prefer not to promote the target allocator till it is 100% successful.

sachouche:
@vrozov,

**What are we trying to solve / improve**
- Drill is currently not properly reporting memory held in Fragment's receive queues
- This makes it hard to analyze OOM conditions
This is what I want to see
- Every operator reporting on the resources it is currently using (needed)
- Fragment held resources (other than the ones already reported by the child operators)
- Drilbit level (metadata caches, web-server, ..)
- I am ok to incrementally reach this goal

**Data Exchange Logistic**
- Ideally, the data exchange fabric should be decoupled from the Drill Receive / Send operators
- The fabric should be handling all the aspects of pre-fetch / pressuring and so forth
- It will tune to the speed of producers / consumers when writing / reading data from it
- This infrastructure should have its own resource management and reporting capabilities

**Operator based Reporting**
- Receive and Send operators shall not worry about batches they didn't consume yet
- Doing so is counter productive as the Data Exchange fabric will interpret a "drain" operation as the operator "needing" more data. 
- For example, the merge-receiver should not be managing the receive queues; it should only advertise the pattern of data consumption and let the exchange fabric figure out the rest. 

The main difference in the two approaches, is that essentially, you are preaching for Receive and Send operators to control all aspects of communication whereas I am preaching for decoupling such aspects.

vrozov:
@sachouche I'd suggest moving the discussion to dev list as the topic of the batch ownership is beyond PR review (code changes).

sachouche:
That was not my intention as my current change aimed at describing the system the way it is. 

@parthchandra, any feedback?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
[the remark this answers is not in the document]
vrozov: it may throw `AssertException` now and other exceptions may be added in the future.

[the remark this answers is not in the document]
sachouche: I see; I will then fix any such occurrences when opportunity presents itself as I have seen both patterns in the Drill code base.

vrozov:
IMO, it will be good to understand what other operators do as well. For example what Project or Filter operators do. Do they take ownership of incoming batches? And if they do, when is the ownership taken?

I do not suggest that we change how Sender and Receiver control **all** aspects of communication, at least not as part of this JIRA/PR. The difference in my and your approach is whether or not UnorderedReceiver and other receivers are pass-through operators. My view is that receivers are not pass-through operators and they are buffering operators as they receive batches from the network and buffer them before downstream operators are ready to consume those batches. In your view, receivers are pass-through operators that get batches from fragment queue or some other queue and pass them to downstream. As there is no wait and no processing between getting a batch from fragment queue and passing it to the next operator, I don't see why a receiver needs to take the ownership.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RawFragmentBatch.java
```diff
@@ -77,4 +83,47 @@ public long getByteCount() {
...
+   *     calls should be routed to the previous DrillBuf
+   * <li>Caller is responsible for checking an OOM condition
+   * </ul>
+   *
+   * @param targetAllocator target allocator
+   * @return a new {@link RawFragmentBatch} object instance on success (where the buffer ownership has
+   *         been switched to the target allocator); otherwise this operation is a NOOP (current instance
+   *         returned)
+   */
+  public RawFragmentBatch transferBodyOwnership(BufferAllocator targetAllocator) {
```
parthchandra: DrillBuf has a transferOwnership method that could be used instead? See the transferTo methods in any of the vector classes -  [FixedValueVectors.transferTo(....)](https://github.com/apache/drill/blob/master/exec/vector/src/main/codegen/templates/FixedValueVectors.java#L282) 
Saves you the trouble of rolling your own and also avoids the need to make changes to any of the existing classes.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -182,13 +189,15 @@ public IterOutcome next() {
         return IterOutcome.OUT_OF_MEMORY;
       }
 
+      // Transfer the ownership of this raw-batch to this operator for proper memory statistics reporting
+      batch = batch.transferBodyOwnership(oContext.getAllocator());
```
parthchandra: This should probably be done inside the [ RecordBatchLoader.load() ](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchLoader.java#L78) method.
Note that `MergingRecordBatch` has similar code and so probably suffers from the same memory accounting issue. All other uses of `RecordBatchLoader.load()` appear to be in the client code or test code so we are unlikely to break anything by making the change in `RecordBatchLoader`.

sachouche:
Met with @parthchandra  and  @vrozov to discuss a more comprehensive fix:

**Agreement**
It was agreed that received batches should be owned by the associated receiver (not the fragment).
This association is done at the framework level (Data Collector) so that the receiver doesn't have to perform any extra processing (such as explicit draining); this is to ensure that no side effect will occur (e.g., acknowledgment logic since it is sensitive to operator record consumption)

**Fix**
- Modified the Unordered & Merge receivers to register their buffer allocators with the associated Data Collector
- The IncomingBuffers class now uses the operator's buffer allocator instead of the fragment allocator

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/MergingReceiverCreator.java
```diff
@@ -44,6 +44,11 @@ public MergingRecordBatch getBatch(ExecutorFragmentContext context,
     assert bufHolder != null : "IncomingBuffers must be defined for any place a receiver is declared.";
     RawBatchBuffer[] buffers = bufHolder.getBuffers(receiver.getOppositeMajorFragmentId());
 
-    return new MergingRecordBatch(context, receiver, buffers);
+    MergingRecordBatch mergeReceiver = new MergingRecordBatch(context, receiver, buffers);
+
+    // Register this operator's buffer allocator so that incoming buffers are owned by this allocator
+    bufHolder.setOprAllocator(receiver.getOppositeMajorFragmentId(), mergeReceiver.getOprAllocator());
```
vrozov: Consider moving registration of the buffer allocator inside `MerginRecordBatch` constructor (change constructor to accept `ExchangeFragmentContext` and `MergingReceiverPOP` only).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverCreator.java
```diff
@@ -40,6 +40,11 @@ public UnorderedReceiverBatch getBatch(ExecutorFragmentContext context, Unordere
     RawBatchBuffer[] buffers = bufHolder.getBuffers(receiver.getOppositeMajorFragmentId());
     assert buffers.length == 1;
     RawBatchBuffer buffer = buffers[0];
-    return new UnorderedReceiverBatch(context, buffer, receiver);
+    UnorderedReceiverBatch receiverBatch = new UnorderedReceiverBatch(context, buffer, receiver);
```
vrozov: The same as for `MerginRecordBatch`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/IncomingBuffers.java
```diff
@@ -129,6 +134,10 @@ public int getRemainingRequired() {
     return collectorMap.get(senderMajorFragmentId).getBuffers();
   }
 
+  public void setOprAllocator(int senderMajorFragmentId, BufferAllocator oprAllocator) {
```
vrozov: Consider introducing `getCollector(int senderMajorFragmentId)` instead of `setOprAllocator` and `getBuffers`;

sachouche: I feel the Collector is an internal class (also used for synchronization). Why exposed it if this is not necessary?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/MergingReceiverCreator.java
sachouche: sure.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverCreator.java
sachouche: sure.

sachouche:
Implemented the changes suggested by @vrozov and fixed a memory leak warning message as the receive operators were closed before the buffers. Now closing the buffers before closing the operators.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/IncomingBuffers.java
vrozov: `DataCollector` is a top-level interface.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/DataCollector.java
```diff
@@ -28,4 +29,16 @@
...
   public int getTotalIncomingFragments();
   public void close() throws Exception;
+  /**
+   * Allows an operator (e.g., receiver) to attach its buffer allocator to this Data Collector so that
+   * incoming batches are owned by the operator instead of the parent (fragment). Note
+   * that this operation is optional; the fragment allocator is used in case no operator invokes this method.
+   *
+   * @param oprAllocator operator buffer allocator
+   */
+  void setOprAllocator(BufferAllocator oprAllocator);
```
vrozov: I am not sure that `Opr` qualification is necessary. There is no assumption that it is an operator. It can be any `BufferAllocator`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/IncomingBuffers.java
```diff
@@ -103,8 +104,12 @@ public boolean batchArrived(final IncomingDataBatch incomingBatch) throws Fragme
             Arrays.toString(collectorMap.values().toArray())));
       }
 
+      // Use the Data Collector's buffer allocator if set, otherwise the fragment's one
+      BufferAllocator ownerAllocator =
+        collector.getOprAllocator() != null ? collector.getOprAllocator() : context.getAllocator();
```
vrozov: Consider moving this logic to `AbstractDataCollector.getAllocator()` or `AbstractDataCollector.setAllocator()`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/DataCollector.java
sachouche: Agreed. What about get/setOwningAllocator?

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/IncomingBuffers.java
sachouche: I don't mind doing so if you think Collector is a top-level interface. I wish there is a better way to indicate which class is an internal framework class vs a public one safe for all operators to access. If you look closely at the code, the DataCollector interface was created to allow multiple implementations; the Drill code exposed the RawFragmentBatchProvider as the main IF for operators to consume incoming data.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/DataCollector.java
vrozov: IMO, get/setAllocator is good enough.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/IncomingBuffers.java
vrozov: It can be done using annotations (I do not suggest to introduce such annotations now).

I don't see a need to hide DataCollector as long as HashMap is not exposed directly or concrete type of a class that implements DataCollector is exposed.

sachouche:
@vrozov, made the changes but note that I had to change the DataCollector IF visibility to public; as previously indicated, this IF was used only within the exec.work.batch package.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/AbstractDataCollector.java
```diff
@@ -116,4 +120,16 @@ public void close() throws Exception {
...
+  /** {@inheritDoc} */
+  @Override
+  public BufferAllocator getAllocator() {
+    return this.ownerAllocator;
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public void setAllocator(BufferAllocator allocator) {
+    this.ownerAllocator = allocator;
```
vrozov: Can passed allocator be `null`? If yes, `ownerAllocator` should not probably be overwritten. If no, please add an `assert` or better `Preconditions` check.

sachouche: I wondered about the same; the only reasons for the caller to pass a null, is either a bug or allocator unregistration. I guess for now, I'll add an assert since we do not really need an unregistration step (this will also help us catch erroneous setting of the allocator).

vrozov: I'd recommend `Preconditions` over `assert`.

sachouche: sure.

arina-ielchiieva:
@sachouche travis build fails with check style errors. Please fix.

sachouche:
@arina-ielchiieva 
I have fixed the Travis issue.

PR #1238, opened 2018-04-23
 DRILL-6281: Refactor TimedRunnable

vrozov:


vrozov:
@parthchandra Please review.

Note to a committer: please do *not* squash commits.

exec/java-exec/src/main/java/org/apache/drill/exec/store/TimedCallable.java
```diff
@@ -0,0 +1,258 @@
...
+        earliestStart = Math.min(earliestStart, threadStart);
+        latestStart = Math.max(latestStart, threadStart);
+        totalStart += threadStart;
+        long executionTime = task.getExecutionTime(TimeUnit.NANOSECONDS);
+        if (executionTime != -1) {
+          doneCount++;
+          totalExecution += executionTime;
+          maxExecution = Math.max(maxExecution, executionTime);
+        } else {
+          logger.info("Task {} started at {} did not finish", task, threadStart);
```
arina-ielchiieva: I suggest we move it to debug level here and below? (I used to do these changes in my PR-1214 as part of `Reduced excessive logging when parquet files metadata is read.`).

arina-ielchiieva:
@vrozov could you please post some line about the changes being done during refactoring, reasons and benefits?

exec/java-exec/src/main/java/org/apache/drill/exec/store/TimedCallable.java
vrozov: The "info" level here is intentional. It provides info about an incomplete or not started `callable`. The regular execution path assumes that all tasks should be complete.

vrozov:
DRILL-6281 is a subtask (preparation step) for DRILL-5908 Regression: Query intermittently may fail with error "Waited for 15000ms, but tasks for 'Get block maps' are not complete." Refactoring is necessary to do RCA.

exec/java-exec/src/main/java/org/apache/drill/exec/store/TimedCallable.java
```diff
@@ -0,0 +1,258 @@
...
+        }
+      } else {
+        logger.info("Task {} never commenced execution", task);
+      }
+    }
+
+    void info(final String activity, final Logger logger, final List<TimedCallable<V>> tasks, int parallelism) {
+      tasks.forEach(this);
```
HanumathRao: Is it better to maintain state that tasks.forEach is collected and print the information in the logger if it is collected instead of tasks.forEach(this) for each invocation of info. I know that at this point of time this is only called once, but in future if a call is introduced this can be avoided.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/FooterGatherer.java
```diff
@@ -66,8 +69,8 @@ private static void checkMagicBytes(FileStatus status, byte[] data, int offset)
   }
 
   public static List<Footer> getFooters(final Configuration conf, List<FileStatus> statuses, int parallelism) throws IOException {
-    final List<TimedRunnable<Footer>> readers = Lists.newArrayList();
-    List<Footer> foundFooters = Lists.newArrayList();
+    final List<TimedCallable<Footer>> readers = new ArrayList<>();
+    final List<Footer> foundFooters = new ArrayList<>();
```
HanumathRao: Any specific reason to not use Lists.newArrayList?

vrozov: Please see Guava Java doc [Lists.newArrayList()](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Lists.html#newArrayList--):

> **Note for Java 7 and later:** this method is now unnecessary and should be treated as deprecated. Instead, use the ArrayList [constructor](https://docs.oracle.com/javase/9/docs/api/java/util/ArrayList.html?is-external=true#ArrayList--) directly, taking advantage of the new ["diamond" syntax](http://goo.gl/iz2Wi).

exec/java-exec/src/main/java/org/apache/drill/exec/store/TimedCallable.java
```diff
@@ -0,0 +1,265 @@
...
+   * @param parallelism  The number of threads that should be run to complete this task.
+   * @return The list of outcome objects.
+   * @throws IOException All exceptions are coerced to IOException since this was build for storage system tasks initially.
+   */
+  public static <V> List<V> run(final String activity, final Logger logger, final List<TimedCallable<V>> tasks, int parallelism) throws IOException {
+    Preconditions.checkArgument(!Preconditions.checkNotNull(tasks).isEmpty(), "list of tasks is empty");
+    parallelism = Math.min(parallelism, tasks.size());
+    final ExecutorService threadPool = parallelism == 1 ? MoreExecutors.newDirectExecutorService()
+        : Executors.newFixedThreadPool(parallelism, new ThreadFactoryBuilder().setNameFormat(activity + "-%d").build());
+    final long timeout = (long)Math.ceil((TIMEOUT_PER_RUNNABLE_IN_MSECS * tasks.size())/parallelism);
```
HanumathRao: IMO, this code part of the original logic needs to be changed to TIMEOUT_PER_RUNNABLE_IN_MSECS * (long)Math.ceil(tasks.size()/parallelism); instead of (long)Math.ceil((TIMEOUT_PER_RUNNABLE_IN_MSECS * tasks.size())/parallelism).

HanumathRao:
@vrozov  Thanks for making the changes. Code changes looks good to me.

parthchandra:
I had the same question as Arina about the need for this change and I got info from DRILL-5908 about the problem cause and how this change will help fix it. Can you elucidate?

vrozov:
The step is necessary to do RCA for DRILL-5908. There are way too many issues with the current implementation to list them in JIRA or PR and the major issue is the usage of homegrown solutions where Java (or other 3rd party libraries) already provides a required functionality out of the box. There is no need to use `Runnable` instead of `Callable` and provide custom `Callable` functionality. It is not necessary to wait on a `CountDownLatch` when `ExecutionService` provides the ability to invoke all tasks and return results when all tasks complete or a timeout expires.

arina-ielchiieva:
@vrozov there is a todo about fork-join, did you consider it? My main point, if yes and it's not applicable, please remove todo. If no, do nothing.

vrozov:
I did not change how tasks (`Runnable` or `Callable`) behave and did not look into converting `Callable/Runnable` to a `ForkJoinTask`. Whether existing tasks can be scheduled recursively or not depends on the nature of those tasks and is not the scope of this PR. I'd suggest filing a JIRA if one does not exist already to look into `Fork/Join` (this is what I would expect from the developer who put "TODO").

parthchandra:
Fair enough. But that *still* does not give me a clue about the problem(s) you were trying to fix, or how the refactoring helps. Is the cause of the problem in TimedRunnable? 
Not an unreasonable request is it?

vrozov:
There is not enough info available to debug and/or troubleshoot DRILL-5908 and I prefer instead of trying to find bugs in homegrown solution replace it with Java out of the box functionality and at the same time provide an ability to log enough information to do RCA for DRILL-5908.

IMO, there are no unreasonable requests on PR/JIRA 😄.

exec/java-exec/src/main/java/org/apache/drill/exec/store/TimedCallable.java
arina-ielchiieva: I still think it should be debug level, such information does not carry any useful information to the user. If you'll query many parquet files, log will be overloaded with these messages.

vrozov: If all tasks complete within the timeout limit as expected there will be no log entry.

arina-ielchiieva: warning instead of info?

exec/java-exec/src/main/java/org/apache/drill/exec/store/TimedCallable.java
```diff
@@ -0,0 +1,266 @@
...
+            activity, startedCount, count, parallelism,
+            TimeUnit.NANOSECONDS.toMillis(earliestStart),
+            TimeUnit.NANOSECONDS.toMillis(totalStart) / startedCount,
+            TimeUnit.NANOSECONDS.toMillis(latestStart));
+      } else {
+        logger.info("{}: started {} out of {} using {} threads.", activity, startedCount, count, parallelism);
+      }
+
+      if (doneCount > 0) {
+        logger.info("{}: completed {} out of {} using {} threads (execution time: total {} ms, avg {} ms, max {} ms).",
```
arina-ielchiieva: What about this info?

vrozov: This is original level from TimedRunnable. If currently there is an excessive logging (I doubt, production system should use WARN or ERROR), please file a JIRA to change it.

vrozov: Sure, but in the future, I'd recommend
- avoid mixing unrelated changes in a single JIRA/PR
- review usage of INFO logging in a separate JIRA/PR
- review production log level also in a separate JIRA/PR (it is uncommon to have INFO level in production currently used by Drill). INFO level is usually used in dev deployments.

Please also see my comment in #1214.

common/src/main/java/org/apache/drill/common/collections/Collectors.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+import com.google.common.base.Preconditions;
+
+public class Collectors
+{
```
arina-ielchiieva: Please move to the previous line.

PR #1239, opened 2018-04-23
DRILL-143: CGroup Support for Drill-on-YARN

kkhatua:
Original commit works for stand-alone Drill. During testing with Drill-on-YARN, it was discovered that while the environment is sourced, the DrillApplicationManager uses a different resource for spinning up the Drillbits. This commit applies the CGroups application logic for YARN specifically.

### DIFF
--- a/distribution/src/resources/yarn-drillbit.sh
+++ b/distribution/src/resources/yarn-drillbit.sh
@@ -58,6 +58,10 @@
 #     The amount of Java heap memory set in the
 #     drill.yarn.drillbit.heap setting. Same override rules as
 #     DRILL_MAX_DIRECT_MEMORY.
+# DRILLBIT_CGROUP
+#     Optional name of CGroup for managing a Drillbit's resources
+# SYS_CGROUP_DIR
+#     Option to specify the cgroup location if it is different from the default
 # DRILL_JAVA_OPTS
 #     The standard JVM options needed to launch Drill. Must be set in
 #     drill-env.sh.
@@ -109,6 +113,36 @@
 # ENABLE_GC_LOG
 #     Enables Java GC logging. Passed from the drill.yarn.drillbit.log-gc
 #     garbage collection option.
+
+### Function to enforce CGroup (Refer local drillbit.sh)
+check_and_enforce_cgroup(){
+    dbitPid=$1;
+    kill -0 $dbitPid
+    if [ $? -gt 0 ]; then 
+      echo "ERROR: Failed to add Drillbit to CGroup ( $DRILLBIT_CGROUP ) for 'cpu'. Ensure that the Drillbit ( pid=$dbitPid ) started up." >&2
+      exit 1
+    fi
+    SYS_CGROUP_DIR=${SYS_CGROUP_DIR:-"/sys/fs/cgroup"}
+    if [ -f $SYS_CGROUP_DIR/cpu/$DRILLBIT_CGROUP/cgroup.procs ]; then
+      echo $dbitPid > $SYS_CGROUP_DIR/cpu/$DRILLBIT_CGROUP/cgroup.procs
+      # Verify Enforcement
+      cgroupStatus=`grep -w $pid $SYS_CGROUP_DIR/cpu/${DRILLBIT_CGROUP}/cgroup.procs`
+      if [ -z "$cgroupStatus" ]; then
+        #Ref: https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt
+        cpu_quota=`cat ${SYS_CGROUP_DIR}/cpu/${DRILLBIT_CGROUP}/cpu.cfs_quota_us`
+        cpu_period=`cat ${SYS_CGROUP_DIR}/cpu/${DRILLBIT_CGROUP}/cpu.cfs_period_us`
+        if [ $cpu_period -gt 0 ] && [ $cpu_quota -gt 0 ]; then
+          coresAllowed=`echo $(( 100 * $cpu_quota / $cpu_period )) | sed 's/..$/.&/'`
+          echo "INFO: CGroup (drillcpu) will limit Drill to $coresAllowed cpu(s)"
+        fi
+      else
+        echo "ERROR: Failed to add Drillbit to CGroup ( $DRILLBIT_CGROUP ) for 'cpu'. Ensure that the cgroup manages 'cpu'" >&2
+      fi
+    else
+      echo "ERROR: CGroup $DRILLBIT_CGROUP not found. Ensure that daemon is running, SYS_CGROUP_DIR is correctly set (currently, $SYS_CGROUP_DIR ), and that the CGroup exists" >&2
+    fi
+}
+
 
 if [ -n "$DRILL_DEBUG" ]; then
   echo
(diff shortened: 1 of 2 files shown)


kkhatua:
@Ben-Zvi  please review. QA verified that Drill-on-YARN works with this patch.

distribution/src/resources/yarn-drillbit.sh
```diff
@@ -110,6 +114,36 @@
...
+    if [ $? -gt 0 ]; then 
+      echo "ERROR: Failed to add Drillbit to CGroup ( $DRILLBIT_CGROUP ) for 'cpu'. Ensure that the Drillbit ( pid=$dbitPid ) started up." >&2
+      exit 1
+    fi
+    SYS_CGROUP_DIR=${SYS_CGROUP_DIR:-"/sys/fs/cgroup"}
+    if [ -f $SYS_CGROUP_DIR/cpu/$DRILLBIT_CGROUP/cgroup.procs ]; then
+      echo $dbitPid > $SYS_CGROUP_DIR/cpu/$DRILLBIT_CGROUP/cgroup.procs
+      # Verify Enforcement
+      cgroupStatus=`grep -w $pid $SYS_CGROUP_DIR/cpu/${DRILLBIT_CGROUP}/cgroup.procs`
+      if [ -z "$cgroupStatus" ]; then
```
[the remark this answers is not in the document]
kkhatua: You're right. I seem to have missed negating the check. Since this check only affects publication of a message and not the actual application of the CGroup, we didn't catch it during testing. I'll fix this port and the original patch as well.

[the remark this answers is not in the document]
kkhatua: Fixed the changes.

paul-rogers:
@kkhatua, it turns out that upstream YARN has long had effective cgroup support per container. ( have the pleasure of sitting near the guy who maintains that work.)There has long been a discussion about whether the MapR version of YARN picked up those changes, we believe that MapR does *not* support this upstream work.

As a result, under Apache YARN, the YARN NM itself will impose cgroup controls and Drill need not do it itself. For MapR YARN (only) Drill (and all other YARN apps) must do their own cgroup control.

Please make sure that this feature is off by default to allow YARN to do the work. Only enable it for versions of YARN (such as MapR) which do not provide cgroup control in YARN itself.

distribution/src/resources/yarn-drillbit.sh
```diff
@@ -175,4 +209,11 @@ fi
 echo "`date` Starting drillbit on `hostname` under YARN, logging to $DRILLBIT_LOG_PATH"
 echo "`ulimit -a`" >> "$DRILLBIT_LOG_PATH" 2>&1
 
-"$DRILL_HOME/bin/runbit" exec
+# Run in background
+"$DRILL_HOME/bin/runbit" exec &
```
paul-rogers: This can't be for YARN. Under YARN, Drill must be run in the foreground. The original Drill-on-YARN work ensured that al this works.

kkhatua:
Thanks for that pointer, @paul-rogers ! I'll make the relevant changes and add to this commit.

paul-rogers:
There may be some misunderstanding of how DoY works. The only info that users can pass to DoY is that which is in the DoY config file. We should add arguments to that file which will be passed through the DoY client to the DoY AM, and from there, as an env var, to the Drillbit containers. We already do this for memory so that both Drill and YARN agree on memory. We should do the same for CPU so that Drill, YARN and cgroups agree on the number of CPUs that Drill can use.

Since this feature is MapR-only, it might be OK to require that users alter their `drill-env.sh` to set an environment variable the forces Drill to police itself under YARN.

distribution/src/resources/yarn-drillbit.sh
kkhatua: The process is momentarily in the background to capture the PID. We eventually wait for it. Are you saying that a process will not continue to run because it is in the background??

kkhatua:
@paul-rogers DoY is no more a MapR-only feature, and if it helps to have Drill self-enforce, this works. If YARN is able to enforce for Drill, the user need not specify the settings in their `drill-env.sh`. 

I see https://issues.apache.org/jira/browse/YARN-810 as being an open issue. (Source: [Issue List](https://issues.apache.org/jira/issues/?jql=project%20%3D%20YARN%20AND%20status%20in%20(Open%2C%20"Patch%20Available")%20AND%20text%20~%20"cgroup"%20ORDER%20BY%20key%20%2C%20status%20ASC) ). So, I probably don't need to do any explicit checks for hadoop distros and documentation should be sufficient to address this, IMO.

distribution/src/resources/yarn-drillbit.sh
paul-rogers: Under YARN, it is YARN that maintains the pid, not Drill. YARN expects its child processes to run in the foreground and will handle capturing the pid.

This is a case in which "native" Apache YARN works differently than "MapR YARN." Since Apache YARN handles cgroups, it is the one that needs the pid. Under MapR's limited YARN, then Drill is second-guessing YARN and needs the pid. It may be that MapR's YARN can handle a background process; I don't recall the details.

Is there way way to run Drill in the background, get the pid, then return to the foreground so that the script does not exit until Drill itself exits?

In fact, if I remember correctly, the scripts have two layers; in one layer, the script replaces itself with the Drill process. Something to check.

paul-rogers:
@kkhatua, putting on my Apache hat... Apache Drill is an Apache project that must work with other Apache projects such as Apache YARN. The Apache Drill DoY support is designed to work well with Apache YARN (and has a few special additions for MapR YARN's unique limitations.) It is important that the Apache DoY work well with the generic YARN. No harm in adding tweaks (such as this one) to work with vendor-specific limitations. But, the overriding concern is that the DoY feature be useful in Apache.

paul-rogers:
One other thing to highlight from an earlier comment. CPU is something that the user specifies in the DoY config file. That information is passed to YARN in container requests. This feature asks the user to modify the drill-env.sh file to enable cgroups to limit CPU. As a result, the user must specify the CPU limit in two places.

DoY went to extreme lengths to unify memory configuration so it is set in one place. The assumption was that, since Apache YARN handles cgroups, we've also got unified CPU specification. But, in this "side-car" approach we don't. So, would be good to capture the CPU amount from the YARN config as explained earlier and use that to set the Drill cgroups env vars -- but only if some "self-enforcing cgroup" flag is enabled in the config (which can be done by default for the limited MapR YARN.)

kkhatua:
@paul-rogers I went through with support on this and found that the issue is not specific to MapR. However, you make a strong argument in favor of letting YARN handle the CGroup management rather than Drill over-reaching. 
I've reverted the change to `yarn-drillbit.sh` in the latest commit.

kkhatua:
Yes, I agree. Support folks gave me similar feedback, so I'll commit this change in the mapr distro _IFF_ there is a request for that. YARN is already a complex beast with numerous settings. Introducing caveats like this will only add to the confusion.

vvysotskyi:
@paul-rogers, @kkhatua, is this PR ready to be merged, or some additional work should be done?

paul-rogers:
Please check with @kkhatua about his intention. The change is OK as far as it goes; the discussion was about whether it went far enough.

kkhatua:
@vvysotskyi  Yes. This commit should carry only changes to one file, instead of the original 2 files. You can go ahead and commit it.

vvysotskyi:
@paul-rogers, @kkhatua, thanks for the clarification, will merge it soon.

PR #1240, opened 2018-04-25
DRILL-6327: Update unary operators to handle IterOutcome.EMIT

sohami:
Note: Handles for Non-Blocking Unary operators (like Filter/Project/etc) with EMIT Iter.Outcome


parthchandra:
+1. Very nicely done.

PR #1242, opened 2018-04-29
DRILL-6361: Added typeOf() function variations

paul-rogers:
Drill provides the `typeof()` function to return the type of a column. However, this function has two key limitations:

1. It returns NULL if any column value is NULL. But, Drill has no NULL type, so this masks the underlying type. This is especially annoying for columns which are all NULL, such as "missing" columns.
2. It does not return the cardinality (AKA "mode") of the column.

This PR introduces two new functions that solve these issues.

### New Functions

`sqlTypeOf()` returns the data type (using the SQL names) whether the column is NULL or not. The SQL name is the one that can be used in a CAST statement. Thus,

```
sqlTypeOf( CAST(x AS <type> ))
```

returns &lt;type> as the type name.

If the type is `DECIMAL`, then the type also includes precision and scale. Example: `DECIMAL(6, 3)`.

`modeOf()` returns the cardinality (mode) of the column as "NOT NULL", "NULLABLE" or "ARRAY". (Suggestions for better terms are welcome.) The Drill terms are not used because they are more Parquet-like than SQL-like.

Finally, the `drillTypeOf()` function that works just like `sqlTypeOf()`, but returns the internal Drill names.

### Example

Here is an example usage that highlights our old friend, "nullable int" for a missing column:

```
SELECT sqlTypeOf(a) AS a_type, modeOf(a) AS a_mode FROM `json/all-null.json`;

+----------+-----------+
|  a_type  |  a_mode   |
+----------+-----------+
| INTEGER  | NULLABLE  |
+----------+-----------+
```

For arrays (repeated) types:

```
SELECT sqlTypeOf(columns) as col_type, modeOf(columns) as col_mode
FROM `csv/cust.csv`;

+--------------------+-----------+
|      col_type      | col_mode  |
+--------------------+-----------+
| CHARACTER VARYING  | ARRAY     |
+--------------------+-----------+
```

For non-null types:

```
SELECT sqlTypeOf(`name`) AS name_type, 
modeOf(`name`) AS name_mode FROM `csvh/cust.csvh`;

+--------------------+------------+
|     name_type      | name_mode  |
+--------------------+------------+
| CHARACTER VARYING  | NOT NULL   |
+--------------------+------------+
```

The result is that the internal Drill type is made very plain to the user of `sqlline`.


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/UnionFunctions.java
```diff
@@ -146,16 +139,131 @@ private static int getTypeValue(MinorType type) {
...
+
+      // Note: extendType is a static function because the byte code magic
+      // for UDFS can't handle switch statements.
+
+      String typeName =
+          org.apache.drill.exec.expr.fn.impl.UnionFunctions.extendType(
+              type,
+              org.apache.drill.common.types.Types.getBaseSqlTypeName(type));
+
+      org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.varCharOutput(
```
vvysotskyi: I think passing a holder into the method may prevent scalar replacement for this holder.

paul-rogers: True. However, since this function is not performance critical, the convenience of a single call vs. multiple lines of code, is a win.

Further, I believe experiments have been done (I did some myself) to show that scalar replacement adds no value: The Java 8 JIT compiler does it automatically so there is no need for Drill to do it itself.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestTypeFns.java
```diff
@@ -0,0 +1,196 @@
...
+    doTypeOfTest("FLOAT", "FLOAT4");
+    doTypeOfTest("DOUBLE", "FLOAT8");
+    doTypeOfTestSpecial("a", "true", "BIT");
+    doTypeOfTestSpecial("a", "CURRENT_DATE", "DATE");
+    doTypeOfTestSpecial("a", "CURRENT_TIME", "TIME");
+    doTypeOfTestSpecial("a", "CURRENT_TIMESTAMP", "TIMESTAMP");
+    doTypeOfTestSpecial("a", "AGE(CURRENT_TIMESTAMP)", "INTERVAL");
+    doTypeOfTestSpecial("BINARY_STRING(a)", "'\\xde\\xad\\xbe\\xef'", "VARBINARY");
+    try {
+      client.alterSession("planner.enable_decimal_data_type", true);
```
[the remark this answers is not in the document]
paul-rogers: Done.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestTypeFns.java
```diff
@@ -0,0 +1,197 @@
...
+    doTypeOfTestSpecial("a", "CURRENT_TIME", "TIME");
+    doTypeOfTestSpecial("a", "CURRENT_TIMESTAMP", "TIMESTAMP");
+    doTypeOfTestSpecial("a", "AGE(CURRENT_TIMESTAMP)", "INTERVAL");
+    doTypeOfTestSpecial("BINARY_STRING(a)", "'\\xde\\xad\\xbe\\xef'", "VARBINARY");
+    try {
+      client.alterSession(PlannerSettings.ENABLE_DECIMAL_DATA_TYPE_KEY, true);
+      doTypeOfTestSpecial("CAST(a AS DECIMAL)", "1", "DECIMAL38SPARSE");
+      doTypeOfTestSpecial("CAST(a AS DECIMAL(6, 3))", "1", "DECIMAL9");
+    } finally {
+      client.resetSession("planner.enable_decimal_data_type");
```
vvysotskyi: Please also replace with `PlannerSettings.ENABLE_DECIMAL_DATA_TYPE_KEY`

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestTypeFns.java
```diff
@@ -0,0 +1,197 @@
...
+    try {
+      client.alterSession(PlannerSettings.ENABLE_DECIMAL_DATA_TYPE_KEY, true);
+
+      // These should include precision and scale: DECIMAL(p, s)
+      // But, see DRILL-6378
+
+      doSqlTypeOfTestSpecial("CAST(a AS DECIMAL)", "1", "DECIMAL");
+      doSqlTypeOfTestSpecial("CAST(a AS DECIMAL(6, 3))", "1", "DECIMAL");
+    } finally {
+      client.resetSession("planner.enable_decimal_data_type");
```
vvysotskyi: And there

paul-rogers:
Addressed code review comments.

Added support for the new `VARBINARY` type. The `sqlTypeOf()` function now returns precision and scale. Example: `DECIMAL(6, 3)`.

vvysotskyi:
@paul-rogers, could you please rebase onto the master?

paul-rogers:
@vvysotskyi, done.

PR #1243, opened 2018-04-29
Solve unable to get jquery in the intranet

mayyamus:
Running on the intranet, access is slow due to the inability to get jquery. Modified to directly access local jQuery resources.


mayyamus:
@arina-ielchiieva Actually using CDN has many benefits，But I think Drill are deployed in the internal network in most case, and webUI does not have a large number of concurrent requests and does not stress the server.In contrast, the speed of each visit is more important.

PR #1244, opened 2018-05-01
DRILL-6373: Refactor Result Set Loader for Union, List support

paul-rogers:
This PR builds on the previous refactoring of the column accessors to prepare for Union, (non-repeated) List and Repeated List support. The PR includes four closely related changes divided across four commits:

### Correct the Type of the Data Vector in a Nullable Vector

The nullable vectors contain a "bits" vector and a "data" vector. The data vector has historically been created using the same `MaterializedField` as the nullable vector, meaning that the data vector is labeled as "nullable" even though it has no bits vector.

This PR creates a clone `MaterializedField` with the same name as the outer nullable vector, but with a Required type.

This change ensures that the overflow logic works correctly as it uses the vector metadata (in the `MaterializedField`) to know what kind of vector to create for the "lookahead" vector.

### Result Set Loader Refactor

The second commit pretty much just rearranges the deck chairs in a way that we an slot in the new types in the next PR. The need for the changes can be seen in the full code set (the union and list support was pulled out for this PR.)

A union is a container, like a map, so the tuple state was refactored to create a common parent container state.

List and unions are very complex to build, so the code to build the internal workings of each vector was pulled out into a separate builder class.

### Projection Handling and the Vector Cache

Previous versions of the result set loader handled projection and a cache for vectors reused across readers in the same Scan operator. Once we introduce nested maps, projection within maps, unions and lists, projection gets much more complex, as does vector caching.

This PR adds logic to support projection and vector caching to any arbitrary level of maps. It turns out that handling projection of an entire map, and projection of fields within maps, is far more complex than you'd think, requiring quite a bit of internal state to keep everything straight. The result is that we can now handle a map `m` with three fields `{a, b, c}` and project just one of them, `m.a`, say.

Further, Drill allows projection of non-existent columns. So, we might ask for field `m.d` which does not exist in the above map. The projection mechanism handles this case as well, creating the right kind of null column.

### Unit Tests

New tests are added to exercise the projection and cache mechanisms. Existing tests were updated for the changes made in the refactoring.

### Reference Design

All of this work is done in support of the overall "batch sizing" project explained [here](https://github.com/paul-rogers/drill/wiki/Batch-Handling-Upgrades).


paul-rogers:
@ppadma, here is the next step in our long saga to check in the batch sizing work. Please give it a review at your convenience. Thanks!

ppadma:
@paul-rogers  LGTM. +1.

paul-rogers:
Thanks @ppadma!

paul-rogers:
Rebased onto latest master. Fixed check style issues due to new rules.

This one is ready to commit; can someone merge it and run the tests?

Ben-Zvi:
There was a failure (after rebasing on the latest master), in the functional test emptyHbaseLeftJoin.sql :
```
select t.data.partkey, t1.data.ps_partkey 
  from schema_change_empty_batch_empty as t 
      LEFT JOIN schema_change_empty_batch_partsupp as t1 
          ON t.data.partkey = t1.data.ps_partkey

Exception:

java.sql.SQLException: SYSTEM ERROR: ConcurrentModificationException

Fragment 2:0

[Error Id: bfb0598f-e598-453b-93f2-699bd87a0061 on drill194:31010]

  (java.util.ConcurrentModificationException) null
    java.util.LinkedHashMap$LinkedHashIterator.nextNode():719
    java.util.LinkedHashMap$LinkedKeyIterator.next():742
    org.apache.drill.exec.record.MaterializedField.withPathAndType():144
    org.apache.drill.exec.record.MaterializedField.clone():125
    org.apache.drill.exec.record.MaterializedField.withPathAndType():145
    org.apache.drill.exec.record.MaterializedField.clone():125
    org.apache.drill.exec.vector.complex.AbstractMapVector.<init>():49
    org.apache.drill.exec.vector.complex.MapVector.<init>():65
    org.apache.drill.exec.expr.BasicTypeHelper.getNewVector():1275
    org.apache.drill.exec.expr.BasicTypeHelper.getNewVector():1261
    org.apache.drill.exec.test.generated.PartitionerGen84661$OutgoingRecordBatch.initializeBatch():381
    org.apache.drill.exec.test.generated.PartitionerGen84661.flushOutgoingBatches():173
    org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$FlushBatchesHandlingClass.execute():285
    org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$PartitionerTask.run():340
    java.util.concurrent.ThreadPoolExecutor.runWorker():1149
    java.util.concurrent.ThreadPoolExecutor$Worker.run():624
    java.lang.Thread.run():748

	at org.apache.drill.jdbc.impl.DrillCursor.nextRowInternally(DrillCursor.java:528)
	at org.apache.drill.jdbc.impl.DrillCursor.loadInitialSchema(DrillCursor.java:600)
	at org.apache.drill.jdbc.impl.DrillResultSetImpl.execute(DrillResultSetImpl.java:1904)
	at org.apache.drill.jdbc.impl.DrillResultSetImpl.execute(DrillResultSetImpl.java:64)
	at oadd.org.apache.calcite.avatica.AvaticaConnection$1.execute(AvaticaConnection.java:630)
	at org.apache.drill.jdbc.impl.DrillMetaImpl.prepareAndExecute(DrillMetaImpl.java:1109)
	at org.apache.drill.jdbc.impl.DrillMetaImpl.prepareAndExecute(DrillMetaImpl.java:1120)
	at oadd.org.apache.calcite.avatica.AvaticaConnection.prepareAndExecuteInternal(AvaticaConnection.java:638)
	at org.apache.drill.jdbc.impl.DrillConnectionImpl.prepareAndExecuteInternal(DrillConnectionImpl.java:200)
	at oadd.org.apache.calcite.avatica.AvaticaStatement.executeInternal(AvaticaStatement.java:149)
	at oadd.org.apache.calcite.avatica.AvaticaStatement.executeQuery(AvaticaStatement.java:218)
	at org.apache.drill.jdbc.impl.DrillStatementImpl.executeQuery(DrillStatementImpl.java:110)
	at org.apache.drill.test.framework.DrillTestJdbc.executeQuery(DrillTestJdbc.java:210)
	at org.apache.drill.test.framework.DrillTestJdbc.run(DrillTestJdbc.java:115)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: oadd.org.apache.drill.common.exceptions.UserRemoteException: SYSTEM ERROR: ConcurrentModificationException

Fragment 2:0

[Error Id: bfb0598f-e598-453b-93f2-699bd87a0061 on drill194:31010]

  (java.util.ConcurrentModificationException) null
    java.util.LinkedHashMap$LinkedHashIterator.nextNode():719
    java.util.LinkedHashMap$LinkedKeyIterator.next():742
    org.apache.drill.exec.record.MaterializedField.withPathAndType():144
    org.apache.drill.exec.record.MaterializedField.clone():125
    org.apache.drill.exec.record.MaterializedField.withPathAndType():145
    org.apache.drill.exec.record.MaterializedField.clone():125
    org.apache.drill.exec.vector.complex.AbstractMapVector.<init>():49
    org.apache.drill.exec.vector.complex.MapVector.<init>():65
    org.apache.drill.exec.expr.BasicTypeHelper.getNewVector():1275
    org.apache.drill.exec.expr.BasicTypeHelper.getNewVector():1261
    org.apache.drill.exec.test.generated.PartitionerGen84661$OutgoingRecordBatch.initializeBatch():381
    org.apache.drill.exec.test.generated.PartitionerGen84661.flushOutgoingBatches():173
    org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$FlushBatchesHandlingClass.execute():285
    org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$PartitionerTask.run():340
    java.util.concurrent.ThreadPoolExecutor.runWorker():1149
    java.util.concurrent.ThreadPoolExecutor$Worker.run():624
    java.lang.Thread.run():748

	at oadd.org.apache.drill.exec.rpc.user.QueryResultHandler.resultArrived(QueryResultHandler.java:123)
	at oadd.org.apache.drill.exec.rpc.user.UserClient.handle(UserClient.java:422)
	at oadd.org.apache.drill.exec.rpc.user.UserClient.handle(UserClient.java:96)
	at oadd.org.apache.drill.exec.rpc.RpcBus$InboundHandler.decode(RpcBus.java:274)
	at oadd.org.apache.drill.exec.rpc.RpcBus$InboundHandler.decode(RpcBus.java:244)
	at oadd.io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:88)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:356)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:342)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:335)
	at oadd.io.netty.handler.timeout.IdleStateHandler.channelRead(IdleStateHandler.java:287)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:356)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:342)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:335)
	at oadd.io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:356)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:342)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:335)
	at oadd.io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:312)
	at oadd.io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:286)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:356)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:342)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:335)
	at oadd.io.netty.channel.ChannelInboundHandlerAdapter.channelRead(ChannelInboundHandlerAdapter.java:86)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:356)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:342)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:335)
	at oadd.io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1294)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:356)
	at oadd.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:342)
	at oadd.io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:911)
	at oadd.io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:131)
	at oadd.io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:645)
	at oadd.io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:580)
	at oadd.io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:497)
	at oadd.io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:459)
	at oadd.io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:131)
	... 1 more
```

paul-rogers:
@Ben-Zvi, the stack trace indicates an error in `MaterializedField`, but that file is not changed by this commit. Nor are any of the other classes in the call stack.

The test is in an HBase test, which is a bit hard to set up. As you know, it is not possible to run functional tests outside of MapR -- which is why only MapR folks can do the actual commits.

The failure occurs inside the partitioned sender, generated code, when attempting to clone a `MaterializedField` for a Map.

Looking at the code, the only way the `MaterializedField.clone()` method can fail is if another thread is adding/removing members to the map while the thread that failed is attempting to clone the map schema. I'm not familiar enough with the partioned sender to track down the issue.

Looking at history, it appears we modified the partition sender back in April. I wonder if that triggered this failure? Can someone who knows more about that class check if there is some way that the same `MaterializedField` is being used across threads, and mutated in one of them?

Otherwise, this may be a race condition; try running the test again and the problem may go away. That may show that the problem is not in this particular PR, but rather a pre-existing problem that requires a separate research effort.

ppadma:
@paul-rogers @Ben-Zvi I ran the tests and got the same failure. Seems like a day one issue. I am not sure why it happens with your PR consistently. I will try to chase it down.

paul-rogers:
Thanks @ppadma! If you can explain how to reproduce the issue without using the full functional tests, I can take a look.

In case it leads to anything... Earlier changes around `MaterializedField` did try to keep the various views of data in sync. The map's `MaterializedField` is supposed to contain, as children, all the map members. This means that changes to any map field requires a matching change to the children. This didn't always happen, leading to bugs that earlier PRs attempted to fix. Not sure if that has an impact here.

The other question is whether the partition sender is the only code that calls the `clone()` method. Do we have unit tests for that method?

Is there some interaction in cloning that reaches back into the original child list? A quick scan of the code didn't suggest any, which is why I wonder if the change occurs in another thread; perhaps triggered as a result of recent fixes in the partitioned sender.

Please let us know what you find, then I'd be happy to fix it.

vrozov:
Changes in partition sender inroduced back in April are not related to the `ConcurrentModificationException` . The regression is introduced in `NullableValueVectors.java:110` as it tries to modify `children` collection while the same collection is iterated over in a different thread. Please take a look at the stack traces for the competing threads:

```
        at org.apache.drill.exec.record.MaterializedField.addChild(MaterializedField.java:132) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.vector.NullableVarBinaryVector.<init>(NullableVarBinaryVector.java:132) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.expr.BasicTypeHelper.getNewVector(BasicTypeHelper.java:1490) [vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.vector.complex.AbstractMapVector.<init>(AbstractMapVector.java:56) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.vector.complex.MapVector.<init>(MapVector.java:65) [vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.expr.BasicTypeHelper.getNewVector(BasicTypeHelper.java:1275) [vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.expr.BasicTypeHelper.getNewVector(BasicTypeHelper.java:1261) [vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.test.generated.PartitionerGen33655$OutgoingRecordBatch.initializeBatch(PartitionerTemplate.java:381) [na:na]
        at org.apache.drill.exec.test.generated.PartitionerGen33655.flushOutgoingBatches(PartitionerTemplate.java:173) [na:na]
        at org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$FlushBatchesHandlingClass.execute(PartitionerDecorator.java:285) [drill-java-exec-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$PartitionerTask.run(PartitionerDecorator.java:340) [drill-java-exec-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_161]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_161]
        at java.lang.Thread.run(Thread.java:748) [na:1.8.0_161]
```
and 
```
        at java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:719) ~[na:1.8.0_161]
        at java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:742) ~[na:1.8.0_161]
        at org.apache.drill.exec.record.MaterializedField.withPathAndType(MaterializedField.java:206) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.record.MaterializedField.clone(MaterializedField.java:185) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.record.MaterializedField.withPathAndType(MaterializedField.java:207) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.record.MaterializedField.clone(MaterializedField.java:185) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.vector.complex.AbstractMapVector.<init>(AbstractMapVector.java:49) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.vector.complex.MapVector.<init>(MapVector.java:65) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.expr.BasicTypeHelper.getNewVector(BasicTypeHelper.java:1275) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.expr.BasicTypeHelper.getNewVector(BasicTypeHelper.java:1261) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.test.generated.PartitionerGen33655$OutgoingRecordBatch.initializeBatch(PartitionerTemplate.java:381) ~[na:na]
        at org.apache.drill.exec.test.generated.PartitionerGen33655.flushOutgoingBatches(PartitionerTemplate.java:173) ~[na:na]
        at org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$FlushBatchesHandlingClass.execute(PartitionerDecorator.java:285) ~[drill-java-exec-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.physical.impl.partitionsender.PartitionerDecorator$PartitionerTask.run(PartitionerDecorator.java:340) ~[drill-java-exec-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        ... 3 common frames omitted
```

vrozov:
Another issue that I noticed and that is likely related to the same changes to `NullableValueVectors.java` is multiple attempts to add the same element to `children` collection on different threads. Due to `Set` nature of `children` collection only one attempt is supposed to succeed, but it is not guaranteed as `children` is not a concurrent collection and access is not synchornized accross threads.

```
2018-06-10 10:57:30,024 [Partitioner-24e29976-0190-774c-c213-3e59bd47533b:frag:2:1-3635] ERROR o.a.d.exec.record.MaterializedField - Added 16923214 to 21b92297. children fee2fd9, size 2,  mod
 count 2
2018-06-10 10:57:30,024 [Partitioner-24e29976-0190-774c-c213-3e59bd47533b:frag:2:1-4243] ERROR o.a.d.exec.record.MaterializedField - Added 16923214 to 21b92297. children fee2fd9, size 2,  mod
 count 2
2018-06-10 10:57:30,024 [Partitioner-24e29976-0190-774c-c213-3e59bd47533b:frag:2:1-656] ERROR o.a.d.exec.record.MaterializedField - Added 16923214 to 21b92297. children fee2fd9, size 3,  mod
count 3
```

paul-rogers:
Thanks much @vrozov for the analysis. I must say I'm a bit stumped. Value vectors are clearly not designed for concurrent modification. That is not simply a code bug, it is a fundamental design decision. Somewhere in code or documentation I recall a statement that says that value vectors are meant to be created once (by a single thread), then be immutable thereafter.

It should be perfectly fine for any number of readers, in separate threads, to access the vector once it has entered its immutable phase. But, nothing about vectors allows concurrent access while mutable.

What is going on in this use case to cause concurrent modification. Is that a "bug" or a "feature"? In the stack trace you provided, both threads are creating a new vector, which should not cause a conflict. If, however, they are modifying the same record batch, then we are violating a design assumption that, like vectors, batches are immutable once created, and that each batch is mutated by a single thread.

The one other possibility is that a bit of code has a bug that is modifying the immutable schema when it should be modifying the mutable one (if working with two vectors), but I'm not sure how that could happen since code that adds fields is not aware of other vectors. Also, AFAIK, while I did change some code to keep metadata in sync (the design of `MaterializedField` really works only for simple vectors; it is a muddle for complex vectors such as maps), the changes only apply to the mutable stage of a vector's lifecycle.

Thoughts?

vrozov:
@paul-rogers Please track the stack trace for the modification part. There is an attempt to add a child to the children of a vector that is part of the `incoming` batch. Please see `PartitionerTemplate.java:381`:
```
        ValueVector outgoingVector = TypeHelper.getNewVector(v.getField(), allocator);
```
The existing assumption is that `v.getField()` is read-only (immutable), but it is passed to `AbstractMapVector` that iterates over passed `field` children instead of using cloned version. Then, 
`child` is passed to `NullableVarCharVector` that again modifies passed parameter.

Please consider changing constructor of vectors to clone passed in field and using cloned version to add child:
```
BaseValueVector.java (add clone()):
  protected BaseValueVector(MaterializedField field, BufferAllocator allocator) {
    this.field = Preconditions.checkNotNull(field, "field cannot be null").*clone*();
    this.allocator = Preconditions.checkNotNull(allocator, "allocator cannot be null");
  }

NullableVarCharVector.java (use cloned version):
  public NullableVarCharVector(MaterializedField field, BufferAllocator allocator) {
    super(field, allocator);
    // replace field with it's clone
    *field = getField();*
    
    // The values vector has the same type and attributes
    // as the nullable vector, but with a mode of required. This ensures that
    // things like scale and precision are preserved in the values vector.
    // For backward compatibility, the values vector must have the same
    // name as the enclosing vector.
    // Setting the child to REQUIRED is a change, prior it was OPTIONAL. But
    // if we then use the field to create a vector, we get the wrong type
    // (we get an optional child vector when we want required.)
    
    values = new VarCharVector(
        MaterializedField.create(field.getName(),
            field.getType().toBuilder()
              .setMode(DataMode.REQUIRED)
              .build()),
        allocator);
    
    field.addChild(bits.getField());
    field.addChild(values.getField());
    accessor = new Accessor();
  }

AbstractMapVector.java (use cloned version):
  protected AbstractMapVector(MaterializedField field, BufferAllocator allocator, CallBack callBack) {
    super(field.clone(), allocator, callBack);
    // replace field with it's clone
    *field = getField();*
    // create the hierarchy of the child vectors based on the materialized field
    for (MaterializedField child : field.getChildren()) {
      if (child.getName().equals(BaseRepeatedValueVector.OFFSETS_FIELD.getName())) {
        continue;
      }
      final String fieldName = child.getName();
      final ValueVector v = BasicTypeHelper.getNewVector(child, allocator, callBack);
      putVector(fieldName, v);
    }
  }
```

paul-rogers:
@vrozov, this is what I meant by the code only working well for a flat vector. Maps have a `MaterializedField` for their own metadata. That metadata contains a list of the child metadata. So:

```
m:Map, {a:Int, b:Varchar}
. a:INT
. b:Varchar
```

This means, as we add `a`, then, `b` to map `m`, we must mutate the `MaterializedField` for `m` to add the child fields. This completely breaks the idea that a `MaterializedField` is always immutable. This is a place where a design goal (immutable `MateraializedField`) collides with implementation (readers add to maps as they find new fields).

So. We could make a clone on each modification and problem solved, right? As in many places in Drill, the solution is not so simple. The above is the easy case. What about a nested map:

```
m1:Map, {m2:Map {a:Int, b:Varchar}}
. m2:Map, {a:Int, b:Varchar}
. . a:INT
. . b:Varchar
```

Now when we add `a`, we have to update both the `m1` and `m2` `MaterializedField`s. If we clone the MaterializedField for `m2`, then the old version in `m1` will get out of sync. The result will be:

```
m1:Map, {m2:Map {}}
. m2:Map, {a:Int, b:Varchar}
. . a:INT
. . b:Varchar
```

Code that depends on accurate schema information then breaks.

So, we're left with a choice: clone and have a corrupt schema, or make `MaterializedField` mutable.

paul-rogers:
@vrozov, your proposed solution works for the single-level map case. It does *not* work for the two-level map case as the `MaterializedField` used inside the inner map is not the one passed from the outer map. Hacky code would be needed to pass in the `MaterializedField`, then fetch it to add it to the outer map's `MaterializedField`.

And, even the above hack does not work for three-level maps.

So, what else can we do? We can note one additional problem with maps. If we simply take a map field `m` from batch B1 and use it to create a map in batch B2, we have a dilemma. The `m` `MaterializedField` already contains "child" entries for `a` and `b`, say.

A similar issue exists for any vector with structure: VarChar (offsets and data), Nullable (bits and data), etc.

So, what to do? The new code, of which this PR is a part, faced similar difficulties with the additional complexity of the `ResultSetLoader` which must clone a vector to produce the "overflow" vector. (This is the reason for fixing the Nullable vectors so that the data vector has its type changed to Required, so that the vector clone works correctly.)

(If all this seems overly complex, I agree that it is. I spent many hours working out all these twists and turns. The original design is broken and was never cleanly revised once maps were added.)

So, we get our solution. Clone the `MaterializedField` in the `PartitionSender` *before* calling `BasicTypeHelper.getNewVector`.

We might thing that `getNewVector()` can do the clone. But, it is used by `ResultSetLoader` assuming that the new vector will use the `MaterializedField` provided. And, in the vast number of cases, we don't need a clone.

It is only when using one vector to create another that we need the clone, so it must be done in the caller. In this case, that is `PartitionSender`.

Does this make sense? Maybe the easiest thing is to just make the change and do a test run.

paul-rogers:
@vrozov, by the way thanks for taking the time to work on this -- a huge help since I can't run the functional tests myself.

If all this seems overly complex, I agree that it is. I spent many hours working out all these twists and turns. The original design is broken and was never cleanly revised once maps were added.

vrozov:
@paul-rogers I did not check the proposed solution, there may be bugs in it, but the main point is that when `PartitionSender` creates new vectors it should not mutate `MaterializedField` of the `incoming` batch vectors. It should be possible to construct a new vector based on an existing vector including two, three or more levels deep maps or other not flat vectors.

paul-rogers:
The specific fix proposal must reflect that the code doing the copying is generated:

```
        at org.apache.drill.exec.expr.BasicTypeHelper.getNewVector(BasicTypeHelper.java:1261) ~[vector-1.14.0-SNAPSHOT.jar:1.14.0-SNAPSHOT]
        at org.apache.drill.exec.test.generated.PartitionerGen33655$OutgoingRecordBatch.initializeBatch(PartitionerTemplate.java:381) ~[na:na]
```

So, one choice is to create a new method, `duplicateVector` that 1) clones the `MaterializedField` and 2) calls `getNewVector`. Then we change the generated code in the partitioner to call the new method.

The other choice is to create a `newVector()` method that does what `getNewVector()` does now, and to revise `getNewVector()` to do the clone. We could then review all uses of `getNewVector()` to determine if we should use the non-cloning `newVector()` instead. This forces us to think about which is needed in each case, but forces generated code to use the "most safe" path.

Thoughts @vrozov?

vrozov:
I would prefer the second approach but instead of introducing `newVector()` method, will it be possible to use a different signature for cloning and not cloning `getNewVector()`? I don't care much whether the method name that does not clone is `newVector()` or `getNewVector()`, but I would prefer that method that does not clone does not accept `MaterializedField` as a parameter at all. It should construct it instead similar to what `getNewVector(String name, ...)` already does.

paul-rogers:
@vrozov, I'll make the required changes as a separate PR once my schedule allows.

paul-rogers:
Rebased on latest paster. Cherry-picked the fix from DRILL-6585 to this branch.

@vrozov or @ppadma - can you kick off a pre-commit build to see if the DRILL-6586 fix resolves the failure we saw earlier? Thanks!

bitblender:
@ppadma  @vrozov  @paul-rogers  I meant I am starting pre-commit tests on this branch with the cherry-pick from 6585

paul-rogers:
@bitblender, thanks for running the pre-commit tests. As it turned out, I changed the wrong file originally: I changed a "Partition something" that turns out to not be used, rather than the `PartitionSenderTemplate` which I should have changed. Sigh...

Anyway, cherry-picked the revised change. This should (fingers crossed) fix the partition sender issue. Please run the test again so we can see if any other code creates a new vector from a "recycled" materialized field.

Thanks!

paul-rogers:
Revised the vector-clone fix. Found a simpler solution that requires just one line of code change. @vrozov or @bitblender, please run a pre-commit test on this PR. If test pass, then PR #1367 can perhaps be included in the next batch commit, this PR in the following one (since it has already obtained an approval). Then, in two weeks, I can provide the final PR for the result set loader work.

bitblender:
@vrozov @paul-rogers I have started a test run. Will update the PR once it is done.

sohami:
@paul-rogers - I have posted couple of questions for this change on other PR (https://github.com/apache/drill/pull/1367)

paul-rogers:
@sohami, the cost of my proposed fix has exceeded its benefit -- we're just not converging. I've closed that PR and will look for another solution. Rather than use the `MaterializedField` to get the type, I'll ad code that does a switch statement on the vector type to learn the "real" type, leaving the `MaterializedField` to hold the pretend type for the `values` field of a `Nullable` vector.

This should not be hard: we already have generated code that parses the vector type. I can use this to manufacture a `MajorType` that matches the vector class; bypassing the need for correct `MaterializedField` data, and thus eliminating the need to change the vector code.

Revision to be done soon.

paul-rogers:
Revised with an alternative fix to the "lying nullable vector values" metadata problem.

Rebased on latest master.

Should be ready for another attempt at merging to master; another pre-commit test run.

paul-rogers:
@vrozov or @sohami: Can you take a look at the latest commit with an alternative solution for inconsistent metadata? If this fix is OK, then I'll mark this as ready to commit, since it previously received a +1. Thanks!

vrozov:
@paul-rogers Did you try to pull changes from PR #1383?

paul-rogers:
@vrozov, not needed. The latest commit provides a different solution. The ResultSetLoader unit tests pass. Next step is for someone at MapR to try the pre-commit tests.

paul-rogers:
@vrozov, perhaps a bit more explanation would help. The bug that you described in the above comments was due (it seems) to a change I made in nullable vector to make the `value` vector carry the correct metadata. This caused problems when we "cloned" that vector. My first fix attempted to fix that secondary issue (the cloning issue). After strong push-back from @sohami, I decided we can't afford to fix the metadata inconsistency (at least not by someone such as myself that is getting rusty on the code.)

So, instead, I backed out the metadata change, which should fix the vector clone issue. Instead, in the result set loader code, I "parse" the vector type and handle the metadata inconsistency.

As a result, the tests for this PR pass. I do need someone, however, to rerun the pre-commit tests to determine if the clone issue has gone away; we have no unit tests that test that particular case.

If we can run those tests, and they pass, then perhaps we can finally commit this PR so I can add the final bit of code change in this series that we've been trying to get in for six months now...

vrozov:
@paul-rogers I disagree that the bug was introduced in the prior changes to this PR. The bug is already part of `AbstractMapVector` that clones MaterializedField but modifies children of the original `MaterializedField` and not the clone. In combination with changes to `NullableVarCharVector` it triggered functional test failures.

paul-rogers:
@vrozov, if the bug was not introduced by this PR, then why have I been trying for months to fix it without access to the test environment? Why did the pre-commit tests pass before this PR, but fail with this PR? Just double-checked. This PR did not touch `AbstractMapVector` and so could not have introduced new issues in that vector.

This PR did change the nullable vector. That change has been backed out and an alternative fix provided.

Could the team Kindly fix the pre-existing bug so that the tests run clean before we attempt to retest this PR? 

Or, can the team re-run the pre-commit tests so that we know if the work around solves the issue, or if there is some deeper issue to fix?

ilooner:
@paul-rogers My apologies, I don't recall discussing or disagreeing to any fix to the value vectors. I am in favor of whatever it takes to get this change in asap. I will try rebasing this PR ontop of the latest master. If the tests pass I will merge it. If the tests don't pass I will try to help debug it or provide a unit test that reproduces the issue. Will keep you posted.

paul-rogers:
@ilooner, the PR in question was a different one; one that tried a more general fix for the concurrent vector update issue. Thanks for rerunning this one so we can see if the workaround works.

I did just rebase the code, so hopefully you will find no issues when you do it again.

Thanks much for your help on this. Once this goes in, I can add the final vector types, then move on to the text and JSON readers.

ilooner:
@paul-rogers All tests passed. Your ordeal is over, I've merged the PR.

paul-rogers:
@ilooner, you are the man! Thanks much! The next PR will be available next week sometime.

PR #1248, opened 2018-05-02
DRIL-6027: Implement Spilling for the Hash-Join

Ben-Zvi:
This PR covers the work to enable the Hash-Join operator (**HJ**) to spill - when its limited memory becomes too small to hold the incoming data. 
 @ilooner is a co-contributor of this work.

Below is a high level description of the main changes, to help the reviewers. More design detail is available in the design document (https://docs.google.com/document/d/1-c_oGQY4E5d58qJYv_zc7ka834hSaB3wDQwqKcMoSAI/)
Some of this work follows a prior similar work done for the Hash-Aggregate (**HAG**) operator; some similarity to the HAG is mentioned to help reviewers familiar with those changes.

## Partitions:
Just like the HAG spilling, the main idea to enable spilling is to split the incoming rows into separate **Partitions**, such that the HJ can gradually adopt to a memory pressure situation by picking an in-memory partition and spilling it as the need arises, thus freeing some memory.
Unlike the HAG, the HJ has two incomings - the build/inner/right and the probe/outer/left. The HJ partitions its Build side first, and if needed, may spill some of these partitions as data is read. Later the Probe side is read and partitioned the same way, where outer partitions matching spilled inner partitions are spilled as well - unconditionally.

###### `HashPartition` class:
A new class `HashPartition` was created to encapsulate the work of each partition; this class handles the pair - the build-side partition and its matching probe-side partition. Most of its code was extracted from prior code in `HashJoinBatch`.

#### Hash Values:
The hash-values are computed at first time, then saved into a special column (named "Hash_Values"), which may be spilled, etc. This avoids re-computation (unlike the HAG, which recomputes). After reading a batch from a spill file, this Hash-values vector is separated (into `read_HV_vector`) and used instead of computing the hash values.

#### Build Hash Table:
Unlike the HAG - the hash-table (and "helper") are built (per each inner partition) only **after** that whole partition was read into memory. (This avoids wasted work, in case the partition needs to spill). Another improvement: As the number of entries is known at that final time (ignoring duplicates), then the hash table can be initially sized right, avoiding the need for later costly re-sizings (see `hashTable.updateInitialCapacity()`). 

#### Same as the HAG:
* Same metrics (NUM_PARTITIONS,	SPILLED_PARTITIONS, SPILL_MB, SPILL_CYCLE) 
* Using the `SpillSet` class.
* Recursive spilling. (Nearly the same code - see `innerNext()` in `HashJoinBatch.java`). Except that the HJ may have duplicate entries - so when the spill cycle has consumed more than 20 bits of the hash value, then err.
* Option controlling the number of partitions (and when that number is 1 --> spilling is disabled).

###### Avoid copying:
Copying the incoming build data into the partitions' batches is a new extra step, adding some overhead. To match performance with prior Drill, in case of a single partition (no spilling, no memory checks) -- the incoming vectors are used as is, without copying. Future work may extend this for the general case (involving memory checks, etc.)

#### Initial memory allocation:
The HJ was made a "buffered" operator (see `isBufferedOperator()`, just like the HAG and the External Sort), hence gets assigned an equal memory share (out of the "memory per query per node"; see `setupBufferedOpsMemoryAllocations()`). Except when the number of partitions is forced to be 1, when it "falls back" to the "old uncontrolled" behavior (similar to what was done for the HAG).

#### Memory Calculator:
The memory calculator is knowledgeable of the current and future memory needs (including current memory usage of all the partitions, an outgoing batch, an incoming outer batch, and the hash tables and "helpers"). The calculator is used first to find an optimal number of partitions (starting from a number controlled by `hashjoin_num_partitions`, default 32, and lowering if that number requires too much memory). The second use of the calculator is to determine if a spill is needed, prior to allocating more memory (see `shouldSpill()`). This check is performed at two places: When reading the build side and about to allocate a new batch (see `appendInnerRow()`). And when hash tables (and helpers) are allocated for the in-memory partitions (in `executeBuildPhase()`).

###### Implementation:
The `HashJoinMemoryCalculator` is an interface, implemented by `HashJoinMemoryCalculatorImpl` for regular work. For testing, we can limit the number of batches `max_batches_in_memory` - and then anther implementation takes over - `HashJoinMechanicalMemoryCalculator`, which uses the number of batches as the spilling trigger. When memory checks are disabled (e.g., when using a single partition), then a special no-op calculator is used instead - `NoopBuildSidePartitioningImpl`.

The memory estimates rely on statistics derived from the actual data. Thus the HJ now, during schema discovery, also tries to "sniff forward" and look at batches with real-data (see `sniffNonEmptyBatch`). 

## Generated Code:
The prior HJ code that generated code (used to move data from Probe and Build incomings into the Outgoing batch) was eliminated, replaced by various `appendRow` methods (see `VectorContainer.java`) that are based on the new `copyEntry()` method supported by all vector types.

## HashTable using VectorContainer:
The HashTable was modified to use `VectorContainer` instead of `RecordBatch`. Thus a `getContainer()` method was added to the `AbstractRecordBatch` and all its subclasses.

## `HashJoinProbe`:
The `HashJoinProbe[template]` were eliminated (partly because of the removal of the generated code) - the remaining code was mostly merged into the end of `HashJoinBatch.java`.

## Other:
* The batches used by the partitions are internal to the HJ. Their size is controlled by an option `num_rows_in_batch`, default 1024.
* Not using the `HyperContainer` anymore, just a list of containers.


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+    }
+  }
+
+  /**
+   *  Append the incoming batch (actually only the vectors of that batch) into the tmp list
+   */
+  public void appendBatch(VectorAccessible batch) {
+    assert numPartitions == 1;
+    int recordCount = batch.getRecordCount();
+    currHVVector = new IntVector(MaterializedField.create("Hash_Values", HVtype), allocator);
```
amansinha100: Use already defined constant for Hash_Values.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -465,82 +785,460 @@ public HashJoinProbe setupHashJoinProbe() throws ClassTransformationException, I
...
+  }
+
+  // ==============================================================
+  //
+  //    Methods used for the probe
+  //
+  // ============================================================
+  private BatchSchema probeSchema;
+
+  enum ProbeState {
```
amansinha100: I am not sure what the moving of all the probe code from HashJoinProbeTemplate to here is buying us.  I am not seeing any advantage and in fact it is odd to see member variables declared at the end of the class.  Further, I  think we would want to keep the option of having generated code for filter evaluation open for hash joins, so having the encapsulation of the HashJoinProbeTemplate is a good thing.

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
```diff
@@ -210,6 +214,109 @@ public void addCollection(Iterable<ValueVector> vectors) {
...
+   *
+   * This works with non-hyper {@link VectorContainer}s which have no selection vectors.
+   * Appends a row taken from two source {@link VectorContainer}s to this {@link VectorContainer}.
+   * @param buildSrcContainer The {@link VectorContainer} to copy the first columns of a row from.
+   * @param buildSrcIndex The index of the row to copy from the build side source {@link VectorContainer}.
+   * @param probeSrcContainer The {@link VectorContainer} to copy the last columns of a row from.
+   * @param probeSrcIndex The index of the row to copy from the probe side source {@link VectorContainer}.
+   * @return Number of records in the container after appending
+   */
+  public int appendRowXXX(VectorContainer buildSrcContainer, int buildSrcIndex, VectorContainer probeSrcContainer, int probeSrcIndex) {
```
ilooner: Since this method is specific to HashJoin it should not be on the VectorContainer which is used throughout Drill. I suggest moving this into a static helper method inside HashJoinBatch.

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
```diff
@@ -210,6 +214,109 @@ public void addCollection(Iterable<ValueVector> vectors) {
...
+        ValueVector srcVector = probeSrcContainer.wrappers.get(vectorIndex).getValueVector();
+        destVector.copyEntry(recordCount, srcVector, probeSrcIndex);
+      }
+    }
+    recordCount++;
+    initialized = true;
+    return recordCount;
+  }
+
+  private int appendBuild(VectorContainer buildSrcContainer, int buildSrcIndex) {
```
ilooner: Specific to HashJoin, move out of VectorContainer.

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
```diff
@@ -210,6 +214,109 @@ public void addCollection(Iterable<ValueVector> vectors) {
...
+    // "- 1" to skip the last "hash values" added column
+    int lastIndex = buildSrcContainer.wrappers.size() - 1 ;
+    for (int vectorIndex = 0; vectorIndex < lastIndex; vectorIndex++) {
+      ValueVector destVector = wrappers.get(vectorIndex).getValueVector();
+      ValueVector srcVector = buildSrcContainer.wrappers.get(vectorIndex).getValueVector();
+      destVector.copyEntry(recordCount, srcVector, buildSrcIndex);
+    }
+    return lastIndex;
+  }
+  private void appendProbe(VectorContainer probeSrcContainer, int probeSrcIndex, int baseIndex) {
```
ilooner: Specific to HashJoin, move out of VectorContainer.

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
```diff
@@ -210,6 +214,109 @@ public void addCollection(Iterable<ValueVector> vectors) {
...
+  /**
+   * A customised version of the special appendRow for HashJoin - used for Left
+   * Outer Join when there is no build side match - hence need a base index in
+   * this container's wrappers from where to start appending
+   * @param probeSrcContainer
+   * @param probeSrcIndex
+   * @param baseInd - index of this container's wrapper to start at
+   * @return
+   */
+  public int appendOuterRow(VectorContainer probeSrcContainer, int probeSrcIndex, int baseInd) {
```
ilooner: Specific to HashJoin, move out of VectorContainer.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -103,6 +103,28 @@ private ExecConstants() {
 
   public static final BooleanValidator EXTERNAL_SORT_DISABLE_MANAGED_OPTION = new BooleanValidator("exec.sort.disable_managed");
 
+  // Hash Join Options
+  public static String HASHJOIN_HASHTABLE_CALC_TYPE_KEY = "exec.hashjoin.hash_table_calc_type";
```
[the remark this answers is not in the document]
Ben-Zvi: Done, all are final ...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+ *  a hash table and a helper are created, and later this data would be probed.
+ *    If all this partition's build/inner data was spilled, then it begins to work as an outer
+ *  partition (see the flag "processingOuter") -- reusing some of the fields (e.g., currentBatch,
+ *  currHVVector, writer, spillFile, partitionBatchesCount) for the outer.
+ *  </p>
+ */
+public class HashPartition implements HashJoinMemoryCalculator.PartitionStat {
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(HashPartition.class);
+
+  public static final String HASH_VALUE_COLUMN_NAME = "Hash_Values";
```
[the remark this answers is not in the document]
Ben-Zvi: Done; though currently we don't search columns by name for this internal batch.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+
+  public static final MajorType HVtype = MajorType.newBuilder()
+    .setMinorType(MinorType.INT /* dataType */ )
+    .setMode(DataMode.REQUIRED /* mode */ )
+    .build();
+
+  // The vector containers storing all the inner rows
+  // * Records are retrieved from these containers when there is a matching record
+  // * on the probe side
+  private ArrayList<VectorContainer> containers;
```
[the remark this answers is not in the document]
Ben-Zvi: Yes; though not clear what are the benefits here of the single schema and the SV4 ...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+
+  // Underlying hashtable used by the hash join
+  private HashTable hashTable;
+
+  private VectorSerializer.Writer writer; // a vector writer for each spilled partition
+  private int partitionBatchesCount; // count number of batches spilled
+  private String spillFile;
+
+  private BufferAllocator allocator;
+  private int RECORDS_PER_BATCH;
```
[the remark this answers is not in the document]
Ben-Zvi: The original was a constant ..... Changed now.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+  ChainedHashTable baseHashTable;
```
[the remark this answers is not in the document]
Ben-Zvi: Very observant !! Now removed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+    } catch (IOException e) {
+      throw UserException.resourceError(e)
+        .message("IO Error while creating a hash table.")
+        .build(logger);
+    } catch (SchemaChangeException sce) {
+      throw new IllegalStateException("Unexpected Schema Change while creating a hash table",sce);
+    }
+    this.hjHelper = new HashJoinHelper(context, allocator);
+    tmpBatchesList = new ArrayList<>();
+    if ( numPartitions > 1 ) { allocateNewCurrentBatchAndHV(); }
```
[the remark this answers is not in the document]
Ben-Zvi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
Ben-Zvi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+        vc.clear();
+      }
+    }
+    while ( tmpBatchesList.size() > 0 ) {
+      VectorContainer vc = tmpBatchesList.remove(0);
+      vc.clear();
+    }
+    closeWriterInternal(deleteFile);
+    clearHashTableAndHelper();
+    if ( containers != null ) { containers.clear(); }
```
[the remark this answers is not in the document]
Ben-Zvi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
ilooner: The getValueVector(int index) method is public on VectorContainer so the VectorWrappers are accessible. Since the method is not used by any operator, please remove it.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -235,86 +338,304 @@ public IterOutcome innerNext() {
...
+              spilledPartitionsList.add(currSp); // so cleanup() would delete the curr spill files
+              this.cleanup();
+              throw UserException
+                .unsupportedError()
+                .message("Hash-Join can not partition inner data any further (too many join-key duplicates? - try merge-join)\n"
+                + "On cycle num %d mem available %d num partitions %d", cycleNum, allocator.getLimit(), numPartitions)
+                .build(logger);
+            }
+          }
+          logger.debug("Start reading spilled partition {} (prev {}) from cycle {} (with {}-{} batches). More {} spilled partitions left.", currSp.origPartn, currSp.prevOrigPartn, currSp.cycleNum, currSp.outerSpilledBatches, currSp.innerSpilledBatches, spilledPartitionsList.size());
```
amansinha100: line seems too long.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
+/*
```
ilooner: This class should be in the **org.apache.physical.impl.join** package since it is specific to hashjoin.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashTableSizeCalculatorConservativeImpl.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.common.collect.Maps;
+import org.apache.drill.exec.record.RecordBatchSizer;
+import org.apache.drill.exec.vector.UInt4Vector;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Map;
+
+public class TestHashTableSizeCalculatorConservativeImpl {
```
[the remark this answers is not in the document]
ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+      }
+    }
+    return newVC;
+  }
+
+  /**
+   *  Allocate a new current Vector Container and current HV vector
+   */
+  public void allocateNewCurrentBatchAndHV() {
+    if ( outerBatchNotNeeded ) { return; } // skip when the inner is whole in memory
```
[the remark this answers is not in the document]
Ben-Zvi: When the inner partition is whole in memory, the probe (and output) is done one row at a time, hence no need for a batch to hold all the rows. Anyway the new name (adding "Alloc") sounds good; change done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -235,86 +338,304 @@ public IterOutcome innerNext() {
...
+          HJSpilledPartition currSp = spilledPartitionsList.remove(0);
+
+          // Create a BUILD-side "incoming" out of the inner spill file of that partition
+          buildBatch = new SpilledRecordbatch(currSp.innerSpillFile, currSp.innerSpilledBatches, context, rightSchema, oContext, spillSet);
+          // The above ctor call also got the first batch; need to update the outcome
+          rightUpstream = ((SpilledRecordbatch) buildBatch).getInitialOutcome();
+
+          if ( currSp.outerSpilledBatches > 0 ) {
+            // Create a PROBE-side "incoming" out of the outer spill file of that partition
+            probeBatch = new SpilledRecordbatch(currSp.outerSpillFile, currSp.outerSpilledBatches, context, probeSchema, oContext, spillSet);
```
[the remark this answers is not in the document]
Ben-Zvi: Each instance corresponds to two spill files (inner/outer); so many instances imply many files. Even if these many instances trigger a GC, it should take much shorter time (I think) than all the IOs for the files.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -235,86 +338,304 @@ public IterOutcome innerNext() {
...
+            if (cycleNum == 2) { logger.info("SECONDARY SPILLING "); }
+            if (cycleNum == 3) { logger.warn("TERTIARY SPILLING ");  }
+            if (cycleNum == 4) { logger.warn("QUATERNARY SPILLING "); }
+            if (cycleNum == 5) { logger.warn("QUINARY SPILLING "); }
+            if ( cycleNum * bitsInMask > 20 ) {
+              spilledPartitionsList.add(currSp); // so cleanup() would delete the curr spill files
+              this.cleanup();
+              throw UserException
+                .unsupportedError()
+                .message("Hash-Join can not partition inner data any further (too many join-key duplicates? - try merge-join)\n"
```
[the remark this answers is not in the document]
Ben-Zvi: OK, removed "merge-join". Another option is that "hash loop" mentioned in the design document (i.e., read a *part* of the inner partition, then scan the *whole* outer, then read the next part of the inner, and scan the whole outer again, etc.)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -235,86 +338,304 @@ public IterOutcome innerNext() {
...
       context.getExecutorState().fail(e);
       killIncoming(false);
       return IterOutcome.STOP;
     }
   }
 
-  public void setupHashTable() throws IOException, SchemaChangeException, ClassTransformationException {
+  private void setupHashTable() throws SchemaChangeException {
+    final List<Comparator> comparators = Lists.newArrayListWithExpectedSize(conditions.size());
+    // When DRILL supports Java 8, use the following instead of the for() loop
```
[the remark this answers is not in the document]
Ben-Zvi: Not just the comment - but the code was changed as well.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+
+      // Holder contains the global index where the key is hashed into using the hash table
+      final IndexPointer htIndex = new IndexPointer();
+
+      assert nextBatch != null;
+      assert probeBatch != null;
+
+      hashTable.updateIncoming(nextBatch, probeBatch );
+
+      // IntVector HV_vector = (IntVector) nextBatch.getValueVector(rightHVColPosition).getValueVector();
```
ilooner: Remove commented out code pls

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
```diff
@@ -0,0 +1,579 @@
...
+      IntVector HV_vector = (IntVector) nextBatch.getLast();
+
+      for (int recInd = 0; recInd < currentRecordCount; recInd++) {
+        int hashCode = HV_vector.getAccessor().get(recInd);
+        try {
+          hashTable.put(recInd, htIndex, hashCode);
+        } catch (RetryAfterSpillException RE) {
+          throw new OutOfMemoryException("HT put");
+        } // Hash Join can not retry yet
```
ilooner: Please remove comment "// Hash Join can not retry yet". HashJoin uses a proactive memory approach so we don't need to do retrys.

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
ilooner: @Ben-Zvi In our discussion offline you were concerned about the overhead of moving these methods out of VectorContainer. Specifically you were concerned about the overhead of calling a get Method. People have benchmarked this overhead and analyzed the assembly generated by the JIT. It turns out calling a getter method generates the same assembly code as a local field access. Please see this post https://stackoverflow.com/questions/23931546/java-getter-and-setter-faster-than-direct-access . So there is no performance difference.

Since there is no performance benefit to adding HashJoin code to VectorContainer, please move these methods out. If all the operators start adding code to VectorContainer the class will become unwieldy.

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
Ben-Zvi: Done ...

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
Ben-Zvi: Done ....

exec/java-exec/src/main/java/org/apache/drill/exec/record/VectorContainer.java
Ben-Zvi: Done ...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
Ben-Zvi: We did raise the option of adopting it for the Hash-Aggregate as well .....

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
Ben-Zvi: Done ....

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
Ben-Zvi: Changed to "Hash Join does not retry"

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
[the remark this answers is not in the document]
amansinha100: Yes, that's the alternative I was describing above, although 'hash-loop' sounds like the right technical term.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
amansinha100: Have you given this any further thought ?  Most of the complex operators in Drill are organized by way of 'Batch' and 'Template' and this structure has been well understood by most developers historically and may even be documented somewhere.   I am in favor of keeping this structure for hash join even though after your current changes the template code is not doing code gen.  I am open to getting feedback from other drill developers on what they think.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashPartition.java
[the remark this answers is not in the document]
amansinha100: Agree..in this case there does not seem to be a compelling reason to use EHC. I am good with this.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -498,6 +497,13 @@ public void setup(HashTableConfig htConfig, BufferAllocator allocator, RecordBat
     currentIdxHolder = new IndexPointer();
   }
 
+  @Override
+  public void updateInitialCapacity(int initialCapacity) {
+    htConfig = htConfig.withInitialCapacity(initialCapacity);
+    allocationTracker = new HashTableAllocationTracker(htConfig, BATCH_SIZE);
+    enlargeEmptyHashTableIfNeeded(initialCapacity);
```
amansinha100: @Ben-Zvi , @ilooner for typical foreign-key-primary-key joins, there is likely to be lots of duplicates of the foreign keys.  For example, in TPC-H, the o_custkey column in Orders table may have as many as 32 duplicates per value(see below).   If Orders is in the build side of a customer-orders hash join, the initial capacity of the hash table would be quite overestimated I think.  

+------------+-------------+
| o_custkey  | cust_count  |
+------------+-------------+
| 898        | 32          |
| 712        | 32          |
| 79         | 32          |
| 1282       | 32          |
| 643        | 32          |
| 4          | 31          |
| 1213       | 30          |
| 334        | 30          |
| 1078       | 30          |
| 73         | 30          |
+------------+-------------+

Ben-Zvi: Maybe in the future we could use stats to lower the "initial capacity" .  Now we could create a new special session option to reduce (e.g., divide by that option) this figure, but this has to be used manually, and would yield a small improvement, so better not.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
Ben-Zvi: Broke this line into multiple lines ...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
amansinha100: Let's create a JIRA to handle this (either through session option or some other way until proper stats are available).  Since there will be performance analysis needed to check the memory footprint, it can be done as a follow-up later.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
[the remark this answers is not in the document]
amansinha100: Pls create an enhancement JIRA for the hash-loop as well.

amansinha100:
@Ben-Zvi thanks for addressing the review comments.  Other than the HashJoinProbeTemplate changes, I am good with the rest of this PR, so let's see if we can come to a resolution on this item.  Let me know your thoughts.

Ben-Zvi:
I am trying now to "undo" to HashJoinProbe elimination and see if this works well ...

kkhatua:
@Ben-Zvi  when ready with the changes, please change the JIRA's state to 'Reviewable'. We're hoping to get this committed by next week and unblock a couple of other PRs that might need rebasing

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinProbeTemplate.java
```diff
@@ -80,40 +68,157 @@
...
+   * @param buildSideIsEmpty
+   * @param numPartitions
+   * @param rightHVColPosition
+   */
   @Override
-  public void setupHashJoinProbe(FragmentContext context, VectorContainer buildBatch, RecordBatch probeBatch,
-                                 int probeRecordCount, HashJoinBatch outgoing, HashTable hashTable,
-                                 HashJoinHelper hjHelper, JoinRelType joinRelType, IterOutcome leftStartState) {
-
+  public void setupHashJoinProbe(RecordBatch probeBatch, HashJoinBatch outgoing, JoinRelType joinRelType, IterOutcome leftStartState, HashPartition[] partitions, int cycleNum, VectorContainer container, HashJoinBatch.HJSpilledPartition[] spilledInners, boolean buildSideIsEmpty, int numPartitions, int rightHVColPosition) {
```
amansinha100: Same comment as above for line too long.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinProbe.java
```diff
@@ -43,11 +39,7 @@
...
   }
 
-  void setupHashJoinProbe(FragmentContext context, VectorContainer buildBatch, RecordBatch probeBatch,
-                          int probeRecordCount, HashJoinBatch outgoing, HashTable hashTable, HashJoinHelper hjHelper,
-                          JoinRelType joinRelType, RecordBatch.IterOutcome leftStartState);
-  void doSetup(FragmentContext context, VectorContainer buildBatch, RecordBatch probeBatch, RecordBatch outgoing);
-  int  probeAndProject() throws SchemaChangeException, ClassTransformationException, IOException;
-  void projectBuildRecord(int buildIndex, int outIndex);
-  void projectProbeRecord(int probeIndex, int outIndex);
+  void setupHashJoinProbe(RecordBatch probeBatch, HashJoinBatch outgoing, JoinRelType joinRelType, RecordBatch.IterOutcome leftStartState, HashPartition[] partitions, int cycleNum, VectorContainer container, HashJoinBatch.HJSpilledPartition[] spilledInners, boolean buildSideIsEmpty, int numPartitions, int rightHVColPosition);
```
amansinha100: Line too long.. can you split this ?  Do we have a check style for max line length ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -169,50 +211,126 @@ protected void buildSchema() throws SchemaChangeException {
...
       if (state == BatchState.FIRST) {
         // Build the hash table, using the build side record batches.
         executeBuildPhase();
-        hashJoinProbe.setupHashJoinProbe(context, hyperContainer, left, left.getRecordCount(), this, hashTable,
-            hjHelper, joinType, leftUpstream);
         // Update the hash table related stats for the operator
-        updateStats(this.hashTable);
+        updateStats();
+        // Initialize various settings for the probe side
+        hashJoinProbe.setupHashJoinProbe(probeBatch, this, joinType, leftUpstream, partitions, cycleNum, container, spilledInners, buildSideIsEmpty, numPartitions, rightHVColPosition);
```
amansinha100: see other comments about line too long.

amansinha100: LGTM.  +1   @Ben-Zvi thanks for adding the HashJoinProbeTemplate.  Pls see couple of cosmetic comments.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableConfig.java
```diff
@@ -29,27 +29,58 @@
...
+    this.initialSizeIsFinal = false;
     this.loadFactor = loadFactor;
     this.keyExprsBuild = keyExprsBuild;
     this.keyExprsProbe = keyExprsProbe;
     this.comparators = comparators;
   }
 
+  @JsonCreator
+  public HashTableConfig(@JsonProperty("initialCapacity") int initialCapacity,
+                         @JsonProperty("initialCapacity") boolean initialSizeIsFinal,
```
vrozov: Is this constructor used with Json? If yes, how it works with `initialSizeIsFinal`bound to the same json `initialCapacity` property?

vrozov: Only one constructor should assign variables, the second constructor should use `this()` with parameters.

Ben-Zvi: Yes, that one looks like an error. And yes, we should reuse the "more comprehensive" constructor. Does this work call for a special Jira, or could it be lumped together with other expected changes ?

vrozov: IMO, it is OK to include code changes into another HashJoin PR.

PR #1249, opened 2018-05-02
DRILL-6380: Fix sporadic mongo db hangs.

ilooner:
Details about problem and solution in Jira https://issues.apache.org/jira/browse/DRILL-6380

### DIFF
--- a/contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
+++ b/contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
@@ -20,9 +20,10 @@ package org.apache.drill.exec.store.mongo;
 import java.io.IOException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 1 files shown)


ilooner:
@arina-ielchiieva

contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestSuit.java
```diff
@@ -94,7 +94,8 @@ private static void setup() throws Exception {
       configServers.add(crateConfigServerConfig(CONFIG_SERVER_3_PORT));
 
       // creating replicaSets
-      Map<String, List<IMongodConfig>> replicaSets = new HashMap<>();
+      // A TreeMap ensures that the config servers are started first.
+      Map<String, List<IMongodConfig>> replicaSets = new TreeMap<>();
```
vrozov: Does it require `TreeMap` or `LinkedHashMap`? What order needs to be preserved?

ilooner: The config servers need to be launched first. Using a tree map would guarantee that the CONFIG_REPLICA_SET key would be iterated over first, since it is lexicographically first compared to the other replica set names. So when the flapadoodle library iterates over the replica sets it will see CONFIG_REPLICA_SET first and launch the config servers.

I agree a LinkedHashMap is more intuitive though, so I have changed the code to use that instead.

vrozov:
LGTM

PR #1250, opened 2018-05-04

kkhatua:
_**Note:** Closed the old PR #755 and opening this._

When Drill is displaying profiles stored on the file system (Local or Distributed), it does so by loading the entire list of `.sys.drill` files in the profile directory, sorting and deserializing. This can get expensive, since only a single CPU thread does this.
As an example, a directory of 120K profiles, the time to just fetch the list of files alone is over 6 seconds. After that, based on the number of profiles being rendered, the time varies. An average of 30ms is needed to deserialize a standard profile, which translates to an additional 3sec for therendering of default 100 profiles.

A user reported issue confirms just that:
DRILL-5028 Opening profiles page from web ui gets very slow when a lot of history files have been stored in HDFS or Local FS

Additional JIRAs filed ask for managing these profiles
DRILL-2362 Drill should manage Query Profiling archiving
DRILL-2861 enhance drill profile file management

This PR brings the following enhancements to achieve that:
1. Mimick the In-memory persistence of profiles (DRILL-5481), by keeping only a predefined `max-capacity` number of profiles in the directory and moving the oldest to an 'archived' sub-directory.
2. Improve loading times by pinning the deserialized list in memory (TreeSet; for maintaining a memory-efficient sortedness of the profiles). That way, if we do not detect any new profiles in the profileStore (i.e. profile directory) since the last time a web-request for rendering the profiles was made, we can re-serve the same listing and skip making a trip to the filesystem to re-fetch all the profiles.

Reload & reconstruction of the profiles in the Tree is done in the event of any of the following states changing:
  i.   Modification Time of profile dir
  ii.  Number of profiles in the profile dir
  iii. Number of profiles requested exceeds existing the currently available list

3. When 2 or more web-requests for rendering arrive, the WebServer code already processes the requests sequentially. As a result, the earliest request will trigger the reconstruction of the in-memory profile-set, and the last-modified timestamp of the profileStore is tracked. This way, the remaining blocked requests can re-use the freshly-reconstructed profile-set for rendering if the underlying profileStore has not been modified. There is an assumption made here that the rate of profiles being added to the profileStore is not too high to trigger a reconstruction for every queued up request. 
4. To prevent frequent archiving, there is a threshold (max-capacity) defined for triggering the archive. However, the number of profiles archived is selected to ensure that the profiles not archived is 90% of the threshold.
5. To prevent the archiving process from taking too long, an archival rate (`drill.exec.profiles.store.archive.rate`) is defined so that upto that many number of profiles are archived in one go, before resumption of re-rendering takes place.
6. On a Distributed FileSystem (e.g. HDFS), multiple Drillbits might attempt to archive. To mitigate that, if a Drillbit detects that it is unable to archive a profile, it will assume that another Drillbit is also archiving, and stop archiving any more.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
@@ -77,8 +77,7 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
   private int version = -1;
   private Function<String, Entry<String, V>> transformer;
 
-  private TreeSet<String> profilesSet;
-  private int profilesSetSize;
+  private ProfileSet profilesSet;
   private PathFilter sysFileSuffixFilter;
 //  private String mostRecentProfile;
   private long basePathLastModified;
@@ -89,8 +88,7 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
 
   private boolean enableArchiving;
   private Path archivePath;
-  private TreeSet<String> pendingArchivalSet;
-  private int pendingArchivalSetSize;
+  private ProfileSet pendingArchivalSet;
   private int archivalThreshold;
   private int archivalRate;
   private Iterable<Entry<String, V>> iterableProfileSet;
@@ -102,24 +100,20 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
     this.config = config;
     this.fs = fs;
     //MRU Profile Cache
-    this.profilesSet = new TreeSet<String>();
-    this.profilesSetSize= 0;
+    this.profilesSet = new ProfileSet(drillConfig.getInt(ExecConstants.HTTP_MAX_PROFILES));
     this.basePathLastModified = 0L;
     this.lastKnownFileCount = 0L;
     this.sysFileSuffixFilter = new DrillSysFilePathFilter();
     //this.mostRecentProfile = null;
 
     //Initializing for archiving
-    if (drillConfig != null ) {
-      this.enableArchiving = drillConfig.getBoolean(ExecConstants.PROFILES_STORE_ARCHIVE_ENABLED); //(maxStoreCapacity > 0); //Implicitly infer
+    this.enableArchiving = drillConfig.getBoolean(ExecConstants.PROFILES_STORE_ARCHIVE_ENABLED); //(maxStoreCapacity > 0); //Implicitly infer
+    if (enableArchiving == true ) {
       this.archivalThreshold = drillConfig.getInt(ExecConstants.PROFILES_STORE_CAPACITY);
       this.archivalRate = drillConfig.getInt(ExecConstants.PROFILES_STORE_ARCHIVE_RATE);
-    } else {
-      this.enableArchiving = false;
+      this.pendingArchivalSet = new ProfileSet(archivalRate);
+      this.archivePath = new Path(basePath, ARCHIVE_LOCATION);
     }
-    this.pendingArchivalSet = new TreeSet<String>();
-    this.pendingArchivalSetSize = 0;
-    this.archivePath = new Path(basePath, ARCHIVE_LOCATION);
 
     // Timing
     this.listAndBuildWatch = Stopwatch.createUnstarted();
@@ -137,14 +131,14 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
 
     //Base Dir
     try {
-      if (!fs.mkdirs(basePath)) {
+      if (!mkdirs(basePath)) {
         version++;
       }
       //Creating Archive if required
       if (enableArchiving) {
         try {
           if (!fs.exists(archivePath)) {
-            fs.mkdirs(archivePath);
+            mkdirs(archivePath);
           }
         } catch (IOException e) {
           logger.warn("Disabling profile archiving due to failure in creating profile archive {} : {}", archivePath, e);
@@ -165,8 +159,8 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
     return PersistentStoreMode.PERSISTENT;
   }
 
-  private void mkdirs(Path path) throws IOException {
-    fs.mkdirs(path);
+  private boolean mkdirs(Path path) throws IOException {
+    return fs.mkdirs(path);
   }
 
   public static Path getLogDir() {
@@ -204,7 +198,6 @@ public class LocalPersistentStore<V> extends BasePersistentStore<V> {
       throw new RuntimeException(e);
     }
 
-    logger.info("Requesting thread: {}-{}" , Thread.currentThread().getName(), Thread.currentThread().getId());
     //No need to acquire lock since incoming requests are synchronized
     try {
       long expectedFileCount = fs.getFileStatus(basePath).getLen();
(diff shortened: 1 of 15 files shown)


kkhatua:
**[Current Apache Master]** User latency when 8 web-clients (wget) request for `/profiles`  against a profile store of 123K profiles (max scale range= 2min)
image

kkhatua:
[DRILL-5270] User latency when 8 web-clients (wget) request for `/profiles` against a profile store of 123K profiles (max scale range= 2min).
Note: Only caching is enabled and no new profiles have been written to the store during the 2 min window.
_Notice how all the subsequent responses go fast the moment the first response is complete, because of the profile cache._
image

kkhatua:
[DRILL-5270] User latency when 8 web-clients (wget) request for `/profiles` against a profile store of 123K profiles (max scale range= 2min). The requests are done in 2 waves
Note: Both caching **and** archiving is enabled and no new profiles have been written to the store during the 2 min window.
Notice how all the subsequent responses go fast the moment the third response is complete. The first 3 clients triggered archiving of profiles from 123K down to about 92K, each time trying to build the cache. By the time the fourth request comes, there is no more archiving, so the requests are served from cache (and, hence, they are barely 2-3 seconds apart). The second wave of requests from the 8 clients is now completely served by the cache.

image

kkhatua:
@arina-ielchiieva / @parthchandra could you review this?

ilooner:
@kkhatua Why not use the Guava Cache? http://www.baeldung.com/guava-cache . I think it would simplify the implementation.

kkhatua:
I did consider using the Gauva Cache initially, but I could not figure out how to specify the eviction policy based on the profile name. Guava provides a mechanism to limit the cache size and evict the oldest entry, but I wanted to override the mechanism that defines 'oldest'. Lastly, the TreeSet allows us to access the elements in a sorted order, which seemed missing in Guava.

ilooner:
@kkhatua I'm still not sure why you want to override the definition of oldest? Why is the default LRU eviction policy not sufficient? 

If you need an ordered list of keys for the cache you can accomplish this with the Guava cache by adding a key to a TreeSet when the Loader is called, and removing a key from a TreeSet when the Removal Listener is called.

My main concern is that implementing our own cache creates complexity and opens up the possibility for bugs. Whereas a pre-existing cache is already debugged and tested for us.

kkhatua:
The way the cache is constructed is by first listing all the profile files and sorting them (the profile ID is generated in a monotonically decreasing value to ensure sortedness in stores like HBase), This customized TreeSet is used to inject profiles (since the FileSystem is not guaranteed to return the list in order), so the TreeSet provides the ordering. We retain only the first N (which are, implicitly, the latest profiles). If we were to add more profiles  than the max capacity, the TreeSet is pruned at the rightmost end.
With Guava, the eviction policy provides the option of limiting the size, but the basis on which it would evict a profile would not work with the least-recently used/accessed profile.
Also, this is currently not a true cache, because the moment we detect changes in the underlying store, we reconstruct this 'cache'. Ideally, we'd want to identify the newest profiles returned from the FileSystem (using filename filters), but the Hadoop API performance is the same (irrespective of the filter).
We, primarily, save the time in fetching file list from the FS and in deserializing.
I can move the implementation of the TreeSet to a separate class to clean up the code. That would make debugging simpler too. With Guava, I don't see the value add beyond a lower risk of bugs, which should be minimal with the TreeSet too.

ilooner:
@kkhatua 

I think I understand the difference in our two perspectives. You wanted a cache that will always only contain the **N** most recently created profiles. If you happen to access the **N + 1**th youngest profile, the cache will not contain it and will never contain it, the cache will only hold the **N** most recently created profiles.

I still prefer the approach with the Guava cache because you can still effectively achieve the same result. As new profiles are created they can be added to the cache. If you access a very old profile, one more recently created profile will be evicted from the cache and the old profile will be added to the cache since a user just requested it. I would argue this behavior is not only easier to implement since we are leveraging a library, but actually more desirable since it caches a profile based on when it is used, not when it was created.

If you still disagree with using the Guava cache. I agree with your proposal of moving your cache into a separate class. I think you should also add some unit tests for the cache to verify that it works as expected. The unit tests will also make maintaining and enhancing the class easier for future developers.

kkhatua:
I actually like the Guava cache approach for its elegance and capabilities, but it expands the scope significantly without a huge benefit from what we currently have. The concept of the cache that you are envisioning is with the complete profile. This is only for listing of the profiles. When an individual profile is accessed, Drill ends up fetching a new copy from the PStore to serialize the contents to visualize it. 
I'll move the class and add some unit tests as well.

ilooner:
@kkhatua Sounds good. Thanks for the explanations and thanks for improving the performance so much :) !

kkhatua:
Done all the changes. Found an unused import in an unrelated file, so I fixed that to make sure the code builds after rebasing to latest master. 
@arina-ielchiieva / @parthchandra  / @ilooner 
Can any (or all) of you do a review?

exec/java-exec/src/test/java/org/apache/drill/exec/store/sys/TestProfileSet.java
```diff
@@ -0,0 +1,130 @@
...
+    ProfileSet testSet = new ProfileSet(finalCapacity);
+    List<String> srcList = new LinkedList<String>(masterList);
+
+    //Loading randomly
+    for (int i = 0; i < initCapacity; i++) {
+      String poppedProfile = testSet.add(srcList.remove(rand.nextInt(storeCount - i)));
+      assertEquals(null, poppedProfile);
+    }
+
+    assert(testSet.size() == initCapacity);
```
arina-ielchiieva: Please use junit assertions in tests.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/ProfileSet.java
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Add a profile name to the set, while removing the oldest or youngest, based on flag
+   * @param profile
+   * @param retainOldest indicate retaining policy as oldest
+   * @return youngest/oldest profile
+   */
+  public String add(String profile, boolean retainOldest) {
+    store.add(profile);
+    if ( size.incrementAndGet() > maxCapacity ) {
```
arina-ielchiieva: Please remove spaces.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -1,220 +1,377 @@
...
+              break;
+            }
+            archivedCount++;
+          }
+          logger.info("Archived {} profiles to {} in {} ms", archivedCount, archivePath, archiveWatch.stop().elapsed(TimeUnit.MILLISECONDS));
+        } else {
+          logger.error("Unable to archive {} profiles to {}", pendingArchivalSet.size(), archivePath.toString());
+        }
+      } catch (IOException e) {
+        e.printStackTrace();
```
arina-ielchiieva: Please replace with proper exception...

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -1,220 +1,377 @@
...
+  private static final int DRILL_SYS_FILE_EXT_SIZE = DRILL_SYS_FILE_SUFFIX.length();
+  private final Path basePath;
+  private final PersistentStoreConfig<V> config;
+  private final DrillFileSystem fs;
+  private int version = -1;
+  private Function<String, Entry<String, V>> transformer;
+
+  private ProfileSet profilesSet;
+  private PathFilter sysFileSuffixFilter;
+//  private String mostRecentProfile;
```
arina-ielchiieva: Please remove.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -1,220 +1,377 @@
...
+
+  private boolean enableArchiving;
+  private Path archivePath;
+  private ProfileSet pendingArchivalSet;
+  private int archivalThreshold;
+  private int archivalRate;
+  private Iterable<Entry<String, V>> iterableProfileSet;
+  private Stopwatch archiveWatch;
+
+  public LocalPersistentStore(DrillFileSystem fs, Path base, PersistentStoreConfig<V> config, DrillConfig drillConfig) {
```
arina-ielchiieva: I don't think that `LocalPersistentStore` should be responsible for archiving. Instead we can consider implementing some "Archivers" (dummy and not dummy, for not) and init them based on `PROFILES_STORE_ARCHIVE_ENABLED` property.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -1,220 +1,377 @@
...
+    //Marking currently seen modification time
+    long currBasePathModified = 0L;
+    try {
+      currBasePathModified = fs.getFileStatus(basePath).getModificationTime();
+    } catch (IOException e) {
+      logger.error("Failed to get FileStatus for {}", basePath, e);
+      throw new RuntimeException(e);
+    }
+
+    //No need to acquire lock since incoming requests are synchronized
```
arina-ielchiieva: Could you please point out to the code which proves this statement?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -1,220 +1,377 @@
...
+      //We'll attempt to reduce to 90% of threshold, but in batches of archivalRate
+      int excessCount = profilesInStoreCount - (int) Math.round(0.9*archivalThreshold);
+      int numToArchive = Math.min(excessCount, archivalRate);
+      logger.info("Found {} excess profiles. For now, will attempt archiving {} profiles to {}", excessCount
+          , numToArchive, archivePath);
+      try {
+        if (fs.isDirectory(archivePath)) {
+          int archivedCount = 0;
+          archiveWatch.reset().start(); //Clocking
+          //while (archivedCount < archivalRate) {
```
arina-ielchiieva: Please remove...

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
```diff
@@ -1,220 +1,377 @@
...
+  private boolean enableArchiving;
+  private Path archivePath;
+  private ProfileSet pendingArchivalSet;
+  private int archivalThreshold;
+  private int archivalRate;
+  private Iterable<Entry<String, V>> iterableProfileSet;
+  private Stopwatch archiveWatch;
+
+  public LocalPersistentStore(DrillFileSystem fs, Path base, PersistentStoreConfig<V> config, DrillConfig drillConfig) {
+    super();
```
arina-ielchiieva: Please remove...

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/store/LocalPersistentStore.java
kkhatua: The method services HTTP requests, which is reflected in the tests (and graphs above). Hence, the need for a lock seemed unnecessary.

kkhatua:
@arina-ielchiieva I've made the following changes: 
1. Refactored to introduce an Archiver
2. Allow for cache to only apply to WebServer 
3. For non-webserver request, like SysTables, support for recursive listing. This is because, while archiving speeds up performance for WebServers, SysTables would need access to archived profiles for analytics.
4. Added tests for the ProfileSet cache

priteshm:
@arina-ielchiieva @kkhatua would this make it into the 1.14 release?

kkhatua:
@arina-ielchiieva I've done some additional changes, please review.

kkhatua:
@priteshm this required some more rework, which I'm hoping I've addressed. We can review and try to get this in as part of 1.15.0
I've rebased this on top of latest master, accounting for conflicts due to  DRILL-6053 (locking of PStore), DRILL-6422 (shaded Guava imports) and DRILL-6492 (schema/workspace insensitivity).
@arina-ielchiieva / @parthchandra  / @ilooner .... any one up for reviewing this?

kkhatua:
Closing this in favor of PR #1654

PR #1252, opened 2018-05-07
DRILL-6386: Disallowed unused imports and removed them.

ilooner:


arina-ielchiieva:
1. Will this cover imports order?
2. Should we have two commits: one by drill dev that fixes unused imports, another by Tim that adds unused commits check?
3. Will this information be added in PR-1251 or here to specific .md document?

vrozov:
In addition to disabling unused imports, it will be good to disable wildcard imports as well.

ilooner:
@arina-ielchiieva Addressing your questions:

1. This doesn't cover import order. From what I vaguely remember with Apache Apex, import order was a highly controversial topic. I want to avoid that discussion for now and get everybody comfortable with adding more checkstyle checks bit by bit. Once people are more comfortable with stricter checkstyle checks I am hoping a discussion about adding import order checks can reach a consensus quicker.
2. I will refactor into two commits as you suggested.
3. I wasn't planning on adding any documentation about this now since the checkstyle errors are pretty descriptive. However, I was planning on writing a Checkstyle.md which will explain how to import a checkstyle.xml into your IDE so that you can see checkstyle warnings within your IDE as well as autoformat your code according to Drill's checkstyle config. Chandni figured out all these things for Apache Apex, but I don't think it's documented anywhere. So I'll sync up with her and get it written down as part of a separate Jira.

ilooner:
@vrozov Disallowed start imports
@arina-ielchiieva Address comments questions

Please let me know if you have anymore comments.

src/main/resources/checkstyle-config.xml
```diff
@@ -27,7 +27,11 @@
 <module name="Checker">
 
   <module name="TreeWalker">
-
+    <module name="UnusedImports"/>
+    <module name="AvoidStarImport">
+      <property name="allowClassImports" value="false"/>
```
vrozov: There is already 'AvoidStarImport' rule. Please change it to use defaults.

ilooner: Thanks for catching. Fixed.

PR #1257, opened 2018-05-11
DRILL-6410: Fixed memory leak in flat Parquet reader

sachouche:
**Problem Description**
- Occasionally, a memory leak is observed within the Parquet reader (flat) when query cancellation is invoked
- I tried to fix this leak in a previous attempt but it seems it is still happening
- Thus far, only QA have been able to observe this issue (and only occasionally)

**Analysis**
- There was a recent breakthrough which gives me hope for addressing this issue 
- The leak logged two piece of information: leak size and state of the child allocator
- The state of the child allocator indicated no leak (all allocated bytes were released)
- After code examination, it occurred to me this was happening because the Asynchronous Page Reader task was releasing the Drill buffer while the scan thread was closing the allocator
- The code attempts to cancel asynchronous tasks and then release allocated buffers, though there is one big caveat: the Java FutureTask.cancel(true) doesn't block during the cancellation process; this method **merely interrupts the asynchronous task** and proceeds
- This means if the asynchronous thread was context switched or doing computation (not blocked waiting), then the fragment cleanup logic can close the allocator before all resources have been released

**Fix**
- The Java ThreadPoolExecutor and FutureTask have few extension points to enhance the task termination process
- Created a new utility class which can create an ExecutorService with the ability to block during future cancellation
- Blocking will happen only when the cancel method is allowed to interrupt the asynchronous task
- Note that there shouldn't be any performance degradation as synchronization code was added only to cover the cancel path
- Also added a new test-suite to test the correctness of this new utility

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/BootStrapContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/BootStrapContext.java
@@ -17,17 +17,15 @@
  */
 package org.apache.drill.exec.server;
 
-import com.codahale.metrics.MetricRegistry;
-import io.netty.channel.EventLoopGroup;
-
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
-import java.util.concurrent.Executors;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.SynchronousQueue;
+
 import org.apache.drill.common.AutoCloseables;
 import org.apache.drill.common.KerberosUtil;
 import org.apache.drill.common.config.DrillConfig;
@@ -43,9 +41,14 @@ import org.apache.drill.exec.rpc.TransportCheck;
 import org.apache.drill.exec.rpc.security.AuthenticatorProvider;
 import org.apache.drill.exec.rpc.security.AuthenticatorProviderImpl;
 import org.apache.drill.exec.server.options.OptionDefinition;
+import org.apache.drill.exec.util.ExecutorServiceUtil;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.CommonConfigurationKeys;
 import org.apache.hadoop.security.UserGroupInformation;
+
+import com.codahale.metrics.MetricRegistry;
+
+import io.netty.channel.EventLoopGroup;
 
 public class BootStrapContext implements AutoCloseable {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(BootStrapContext.class);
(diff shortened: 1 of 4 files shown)


sachouche:
@parthchandra, can you please review this fix?

Thanks in advance!

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -297,11 +297,11 @@ private void waitForExecutionResult() throws InterruptedException, ExecutionExce
     while (asyncPageRead != null && !asyncPageRead.isEmpty()) {
       try {
         Future<Void> f = asyncPageRead.poll();
-        if(!f.isDone() && !f.isCancelled()){
+        if(!f.isDone() && !f.isCancelled()) {
           f.cancel(true);
-        } else {
-          f.get(1, TimeUnit.MILLISECONDS);
         }
+        // The framework guarantees a blocking version of FutureTask cancellation. At this time we are
```
vrozov: Please see `PartitionerTask` in `PartitionerDecorator`. The design goal for the `PartionerTask` is exactly the same, make sure that the main thread will not proceed till all tasks are either done or canceled. Currently, `PartitionerTask` is a private class, but it should be possible to generalize it.

exec/java-exec/src/main/java/org/apache/drill/exec/server/BootStrapContext.java
```diff
@@ -43,10 +41,15 @@
 import org.apache.drill.exec.rpc.security.AuthenticatorProvider;
 import org.apache.drill.exec.rpc.security.AuthenticatorProviderImpl;
 import org.apache.drill.exec.server.options.OptionDefinition;
+import org.apache.drill.exec.util.ExecutorServiceUtil;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.CommonConfigurationKeys;
 import org.apache.hadoop.security.UserGroupInformation;
 
+import com.codahale.metrics.MetricRegistry;
```
vrozov: Please avoid import order change.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
sachouche: @vrozov 

I have looked at that fix but decided against it as I felt the same logic could be achieved with a simpler implementation especially that the performance penalty is only incurred during the cancellation process.

exec/java-exec/src/main/java/org/apache/drill/exec/server/BootStrapContext.java
sachouche: Sure.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
parthchandra: My $0.02  is to tend towards  what @vrozov is suggesting. It would be preferable to have one reusable implementation that works for both use cases. 
I don't know if @sachouche needs the localStats in PartitionTask, nor, as he mentioned, do you want to pay a price for non cancellation paths. 
Is it possible to come up with a unified approach here?

PR #1258, opened 2018-05-14
DRILL-6321:  Lateral Join and Unnest - initial implementation for parser and planning

chunhui-shi:


exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/PlannerSettings.java
```diff
@@ -112,6 +112,12 @@
   public static final EnumeratedStringValidator QUOTING_IDENTIFIERS = new EnumeratedStringValidator(
       QUOTING_IDENTIFIERS_KEY, Quoting.BACK_TICK.string, Quoting.DOUBLE_QUOTE.string, Quoting.BRACKET.string);
 
+  /*
+    "planner.enable_unnest_lateral" is to allow users to choose enable unnest+lateraljoin feature.
+   */
+  public static final String ENABLE_UNNEST_LATERAL_KEY = "planner.enable_unnest_lateral";
```
amansinha100: Do we need this option ? Or is this temporary ?

chunhui-shi: This is temporary or for test purpose

amansinha100:
@chunhui-shi  I have a few comments/questions mostly around costing.  However, these are not blockers for the current PR.  You can comment on them and address them in a follow-up if needed.  Code changes overall LGTM.  +1

PR #367, opened 2016-02-06
DRILL-4364: Image Metadata Format Plugin

nagix:
Initial commit of the Image Metadata Format Plugin

PR #1259, opened 2018-05-14
DRILL-6353: Upgrade Parquet MR dependencies

vrozov:
@parthchandra Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/PageReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/PageReader.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.store.parquet.columnreaders;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import io.netty.buffer.ByteBufUtil;
 import org.apache.drill.exec.util.filereader.BufferedDirectBufInputStream;
@@ -30,6 +31,7 @@ import org.apache.drill.exec.util.filereader.DirectBufInputStream;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.apache.parquet.bytes.ByteBufferInputStream;
 import org.apache.parquet.bytes.BytesInput;
 import org.apache.parquet.column.Dictionary;
 import org.apache.parquet.column.Encoding;
@@ -250,7 +252,7 @@ class PageReader {
   }
 
   public static BytesInput asBytesInput(DrillBuf buf, int offset, int length) throws IOException {
-    return BytesInput.from(buf.nioBuffer(offset, length), 0, length);
+    return BytesInput.from(buf.nioBuffer(offset, length));
   }
 
 
@@ -319,41 +321,44 @@ class PageReader {
 
     byteLength = pageHeader.uncompressed_page_size;
 
-    final ByteBuffer pageDataBuffer = pageData.nioBuffer(0, pageData.capacity());
+    final ByteBufferInputStream in = ByteBufferInputStream.wrap(pageData.nioBuffer(0, pageData.capacity()));
 
     readPosInBytes = 0;
     if (parentColumnReader.getColumnDescriptor().getMaxRepetitionLevel() > 0) {
       repetitionLevels = rlEncoding.getValuesReader(parentColumnReader.columnDescriptor, ValuesType.REPETITION_LEVEL);
-      repetitionLevels.initFromPage(currentPageCount, pageDataBuffer, (int) readPosInBytes);
+      repetitionLevels.initFromPage(currentPageCount, in);
       // we know that the first value will be a 0, at the end of each list of repeated values we will hit another 0 indicating
       // a new record, although we don't know the length until we hit it (and this is a one way stream of integers) so we
       // read the first zero here to simplify the reading processes, and start reading the first value the same as all
       // of the rest. Effectively we are 'reading' the non-existent value in front of the first allowing direct access to
       // the first list of repetition levels
-      readPosInBytes = repetitionLevels.getNextOffset();
+      readPosInBytes = in.position();
       repetitionLevels.readInteger();
     }
-    if (parentColumnReader.columnDescriptor.getMaxDefinitionLevel() != 0){
+    if (parentColumnReader.columnDescriptor.getMaxDefinitionLevel() != 0) {
       parentColumnReader.currDefLevel = -1;
       definitionLevels = dlEncoding.getValuesReader(parentColumnReader.columnDescriptor, ValuesType.DEFINITION_LEVEL);
-      definitionLevels.initFromPage(currentPageCount, pageDataBuffer, (int) readPosInBytes);
-      readPosInBytes = definitionLevels.getNextOffset();
+      definitionLevels.initFromPage(currentPageCount, in);
+      readPosInBytes = in.position();
       if (!valueEncoding.usesDictionary()) {
         valueReader = valueEncoding.getValuesReader(parentColumnReader.columnDescriptor, ValuesType.VALUES);
-        valueReader.initFromPage(currentPageCount, pageDataBuffer, (int) readPosInBytes);
+        valueReader.initFromPage(currentPageCount, in);
       }
     }
-    if (parentColumnReader.columnDescriptor.getType() == PrimitiveType.PrimitiveTypeName.BOOLEAN) {
+    if (valueReader == null && parentColumnReader.columnDescriptor.getType() == PrimitiveType.PrimitiveTypeName.BOOLEAN) {
       valueReader = valueEncoding.getValuesReader(parentColumnReader.columnDescriptor, ValuesType.VALUES);
-      valueReader.initFromPage(currentPageCount, pageDataBuffer, (int) readPosInBytes);
+      valueReader.initFromPage(currentPageCount, in);
     }
     if (valueEncoding.usesDictionary()) {
       // initialize two of the dictionary readers, one is for determining the lengths of each value, the second is for
       // actually copying the values out into the vectors
+      Preconditions.checkState(readPosInBytes < pageData.capacity());
+      int index = (int)readPosInBytes;
+      ByteBuffer byteBuffer = pageData.nioBuffer(index, pageData.capacity() - index);
       dictionaryLengthDeterminingReader = new DictionaryValuesReader(dictionary);
-      dictionaryLengthDeterminingReader.initFromPage(currentPageCount, pageDataBuffer, (int) readPosInBytes);
+      dictionaryLengthDeterminingReader.initFromPage(currentPageCount, ByteBufferInputStream.wrap(byteBuffer));
       dictionaryValueReader = new DictionaryValuesReader(dictionary);
-      dictionaryValueReader.initFromPage(currentPageCount, pageDataBuffer, (int) readPosInBytes);
+      dictionaryValueReader.initFromPage(currentPageCount, ByteBufferInputStream.wrap(byteBuffer));
       parentColumnReader.usingDictionary = true;
     } else {
       parentColumnReader.usingDictionary = false;
@@ -445,25 +450,29 @@ class PageReader {
    * @throws IOException An IO related condition
    */
   void resetDefinitionLevelReader(int skipCount) throws IOException {
-    if (parentColumnReader.columnDescriptor.getMaxDefinitionLevel() != 0) {
-      throw new UnsupportedOperationException("Unsupoorted Operation");
+    Preconditions.checkState(parentColumnReader.columnDescriptor.getMaxDefinitionLevel() == 1);
+    Preconditions.checkState(currentPageCount > 0);
+
+    final Encoding rlEncoding = METADATA_CONVERTER.getEncoding(pageHeader.data_page_header.repetition_level_encoding);
+    final Encoding dlEncoding = METADATA_CONVERTER.getEncoding(pageHeader.data_page_header.definition_level_encoding);
+
+    final ByteBufferInputStream in = ByteBufferInputStream.wrap(pageData.nioBuffer(0, pageData.capacity()));
+
+    if (parentColumnReader.getColumnDescriptor().getMaxRepetitionLevel() > 0) {
+      repetitionLevels = rlEncoding.getValuesReader(parentColumnReader.columnDescriptor, ValuesType.REPETITION_LEVEL);
+      repetitionLevels.initFromPage(currentPageCount, in);
+      repetitionLevels.readInteger();
     }
 
-    final Encoding dlEncoding = METADATA_CONVERTER.getEncoding(pageHeader.data_page_header.definition_level_encoding);
-    final ByteBuffer pageDataBuffer = pageData.nioBuffer(0, pageData.capacity());
-    final int defStartPos = repetitionLevels != null ? repetitionLevels.getNextOffset() : 0;
     definitionLevels = dlEncoding.getValuesReader(parentColumnReader.columnDescriptor, ValuesType.DEFINITION_LEVEL);
     parentColumnReader.currDefLevel = -1;
 
     // Now reinitialize the underlying decoder
-    assert currentPageCount > 0 : "Page count should be strictly upper than zero";
-    definitionLevels.initFromPage(currentPageCount, pageDataBuffer, defStartPos);
+    definitionLevels.initFromPage(currentPageCount, in);
 
     // Skip values if requested by caller
     for (int idx = 0; idx < skipCount; ++idx) {
       definitionLevels.skip();
     }
   }
-
-
 }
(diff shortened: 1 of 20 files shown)


pom.xml
```diff
@@ -44,7 +44,7 @@
     <dep.slf4j.version>1.7.6</dep.slf4j.version>
     <dep.guava.version>18.0</dep.guava.version>
     <forkCount>2</forkCount>
-    <parquet.version>1.8.1-drill-r0</parquet.version>
+    <parquet.version>1.10.0</parquet.version>
```
parthchandra: 1.10? Is it safer to upgrade to 1.8.3 and then test out 1.9/1.10 before upgrading to it?
Also, with this change the Hive parquet version is 1.8.2. I wonder what impact that might have on compatibility?

vrozov: `1.8.3` as well as `1.8.1-drill-r0` are supposed to be a patch release on top of `1.8.0`. Unfortunately `parquet-mr` does not properly follow semantic versioning and have functional and API level changes in the patch releases. On top of that, both `1.9.0` and `1.10.0` are not backward compatible with `1.8.0` and/or `1.8.3`, so upgrade to `1.8.3` will not help us with the upgrade to `1.10.0` later either or make it safer. I'd suggest to pay the price once. Additionally, the latest parquet version may have a functionality to be used in filter pushdown. @arina-ielchiieva what is your take?

Parquet libraries used by hive are shaded within `drill-hive-exec-shaded`, so hive is guarded from the parquet-mr library upgrade.

ilooner:
@vrozov Please fix Travis failures

```
Tests in error: 
  TestCTASPartitionFilter.withoutDistribution Â» UserRemote SYSTEM ERROR: Illegal...
  TestCTASPartitionFilter>BaseTestQuery.closeClient:286 Â» Runtime Exception whil...
```

pom.xml
arina-ielchiieva: Well, I need this upgrade to implement varchar filter push down, since it has been fixed in 1.10.0 but not in 1.8.3. I think if you unit tests pass, along with Functional & Advanced we are safe to go.

vrozov:
I am still working on fixing the unit and functional tests. The PR is open to initiate a discussion on `parquet-mr` version upgrade.

pom.xml
parthchandra: Fair enough.

arina-ielchiieva:
@vrozov sounds like, we are all in consensus on upgrade to the latest parquet version. So when all tests pass, please ping us and we'll finish code review.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -737,6 +738,7 @@ public void testBooleanPartitionPruning() throws Exception {
     }
   }
 
+  @Ignore
```
vvysotskyi: Could you please explain the reason why these tests should be ignored?

arina-ielchiieva: @vvysotskyi no worries, work is still in progress.

arina-ielchiieva: In this case, I don't think it's a good idea to disable unit tests. You can consider asking for help to resolve unit tests failures but not disable them.

vrozov: The test needs to be fixed as part of a separate JIRA/PR (another option is to remove the check for the filter, but IMO it is even less desirable).

parthchandra: At what point do these tests fail? If they happen (as I suspect) during planning or building the metadata cache, then we are looking at a bigger issue that may need to be addressed before we switch to the upgraded version.
Essentially, pruning checks stats to determine whether the file should be read or not. If the stats are corrupted and we use them anyway, the query will return wrong results. If the stats are not corrupted and the Parquet library refuses to read them because it cannot be sure of the state of the stats, then the pruning will fail and the query will take too long to run. 
If the tests are failing to execute because the Parquet library fails to read the stats, then we will have a situation where queries that used to run correctly will stop running after we upgrade to the new Parquet library, which is likely to be *very* unpopular.

vrozov: The tests fail during plan validation as the new version of the parquet library ignores wrong statistics for the data types used by queries. Even if statistics is wrong for a small portion of parquet files and for the parquet files used by the tests it is correct, Drill can't rely on wrong statistics as it leads to the wrong query results. Basically, there is a bug in the version of the parquet library that Drill currently uses that may cause query result to be wrong and this bug is fixed in the new version causing 2 unit tests failures. IMO, it is better to upgrade to the new library sooner than later even if it will cause slowdown for some queries.

parthchandra: Hmm, we need to take a look at this. For a period of two years, files written by tools using Parquet libraries were writing incorrect statistics, but because Drill had its own build where we had fixed the issue (we found the issue in the first place), files written by Drill were correct. A very large number of Drill users use the Parquet files produced by Drill and it was decided that we cannot penalize them. We provided a migration tool to users to tag files produced by Drill. The tool added information in the extra metadata in Parquet files to indicate the file was written by Drill and stats from these files should be allowed. 
AFAIK, this should be in the current build of Drill Parquet as well as in the Parquet library v 1.8.2 and above. 
Do you know if the stats corruption that affects these tests is something that was fixed in a version after 1.8.2?

vrozov: It will be good if you can point to JIRA with the fix that Drill uses to correct statistics. Without JIRA it is not clear what particular fix is used by Drill to workaround bugs in how parquet library handles statistics and for what data types.

arina-ielchiieva: I have investigated why these tests fail. For example, let's take `testIntervalDayPartitionPruning`.
First test creates partitioned table using Drill. Since table is created at runtime, new parquet lib is used. Created table contains 4 files, one of them contains all nulls.  For this file with nulls, statistics for all types except of binary is `num_nulls: 3, min/max not defined`. For binary type it is `no stats for this column`. In previous parquet version, statistics was written correctly. Maybe this is bug in parquet, maybe in Drill writer. 

Another problem is with metadata file. We do write metadata for binary columns into it successfully. Example:
```
  "columnTypeInfo" : {
    "`col_intrvl_day`" : {
      "name" : [ "col_intrvl_day" ],
      "primitiveType" : "FIXED_LEN_BYTE_ARRAY",
      "originalType" : "INTERVAL",
      "precision" : 0,
      "scale" : 0,
      "repetitionLevel" : 0,
      "definitionLevel" : 1
    },
        "name" : [ "col_intrvl_day" ],
        "minValue" : "AAAAABoAAACQ4KEB",
        "maxValue" : "AAAAABoAAACQ4KEB",
        "nulls" : 0
```
But when reading it back from file, we read empty strings. Looks like this one is Drill bug.

@vrozov  I also have noticed that `ParquetFileReader.readFooter(conf, path, NO_FILTER);`. If you'll have a chance, please replace it.

vrozov: Where the `columnTypeInfo` comes from? Is it an output of parquet-tools? If yes, please provide details of the command line parameters used. If not, please check what parquet-tools report.

arina-ielchiieva: This is snippet from metadata file. Drill generates it using info from the footer.

https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java

https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V3.java

vrozov: Please check what parquet-tools report when it reads that file back.

parthchandra:
The fix for the stats was part of a big commit to add support for
ByteBuffers in Parquet (PARQUET-77
<https://issues.apache.org/jira/browse/PARQUET-77>; commit 6b605a4
<https://github.com/apache/parquet-mr/commit/6b605a4ea05b66e1a6bf843353abcb4834a4ced8>).
See the included commit 7bc2a4d
<https://github.com/apache/parquet-mr/commit/7bc2a4d0887a6d324bf5a12ec01701541dee30f8>
which
was to fix the overwriting of stats.

vrozov:
The fix for PARQUET-77 is included into 1.10.0, 1.9.0 and 1.8.3 as it is not specific for Apache Drill.

vrozov:
@parthchandra @arina-ielchiieva The PR is ready for the final review.

vrozov:
@parthchandra @arina-ielchiieva Please review

vrozov:
@arina-ielchiieva Please see [my comment](https://github.com/apache/drill/pull/1259#discussion_r190642661)

parthchandra:
Based on Arina's analysis, I don't think it is ok to ignore this test failure.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
vrozov: I attached the output of `org.apache.parquet.format.converter.ParquetMetadataConverter` in the debug mode to [DRILL-6353](https://issues.apache.org/jira/browse/DRILL-6353). As you can see there is no statistics available for the `col_intrvl_day`:
```
{
  "encodingStats" : null,
  "dictionaryPageOffset" : 0,
  "valueCount" : 6,
  "totalSize" : 92,
  "totalUncompressedSize" : 91,
  "statistics" : {
    "max" : null,
    "min" : null,
    "maxBytes" : null,
    "minBytes" : null,
    "empty" : true,
    "numNulls" : -1,
    "numNullsSet" : false
  },
  "firstDataPageOffset" : 532,
  "type" : "FIXED_LEN_BYTE_ARRAY",
  "path" : [ "col_intrvl_day" ],
  "primitiveType" : {
    "name" : "col_intrvl_day",
    "repetition" : "OPTIONAL",
    "originalType" : "INTERVAL",
    "id" : null,
    "primitive" : true,
    "primitiveTypeName" : "FIXED_LEN_BYTE_ARRAY",
    "decimalMetadata" : null,
    "typeLength" : 12
  },
  "codec" : "SNAPPY",
  "encodings" : [ "RLE", "BIT_PACKED", "PLAIN" ],
  "startingPos" : 532
}
```
This is result of parquet fix for "Deprecate type-defined sort ordering for INTERVAL type"([PARQUET-1064](https://issues.apache.org/jira/browse/PARQUET-1064))

vrozov: The same applies to `testIntervalYearPartitionPruning`: statistics for `col_intrvl_yr` is also not available for the same reason:
```
{
  "encodingStats" : null,
  "dictionaryPageOffset" : 0,
  "valueCount" : 6,
  "totalSize" : 81,
  "totalUncompressedSize" : 91,
  "statistics" : {
    "max" : null,
    "min" : null,
    "maxBytes" : null,
    "minBytes" : null,
    "empty" : true,
    "numNulls" : -1,
    "numNullsSet" : false
  },
  "firstDataPageOffset" : 451,
  "type" : "FIXED_LEN_BYTE_ARRAY",
  "path" : [ "col_intrvl_yr" ],
  "primitiveType" : {
    "name" : "col_intrvl_yr",
    "repetition" : "OPTIONAL",
    "originalType" : "INTERVAL",
    "id" : null,
    "primitive" : true,
    "primitiveTypeName" : "FIXED_LEN_BYTE_ARRAY",
    "decimalMetadata" : null,
    "typeLength" : 12
  },
  "codec" : "SNAPPY",
  "encodings" : [ "RLE", "BIT_PACKED", "PLAIN" ],
  "startingPos" : 451
}
```

vrozov: And for `testDecimalPartitionPruning` statistics for `MANAGER_ID` is not available either:
```
{
  "encodingStats" : null,
  "dictionaryPageOffset" : 0,
  "valueCount" : 107,
  "totalSize" : 168,
  "totalUncompressedSize" : 363,
  "statistics" : {
    "max" : null,
    "min" : null,
    "maxBytes" : null,
    "minBytes" : null,
    "empty" : true,
    "numNulls" : -1,
    "numNullsSet" : false
  },
  "firstDataPageOffset" : 5550,
  "type" : "FIXED_LEN_BYTE_ARRAY",
  "path" : [ "MANAGER_ID" ],
  "codec" : "SNAPPY",
  "primitiveType" : {
    "name" : "MANAGER_ID",
    "repetition" : "OPTIONAL",
    "originalType" : "DECIMAL",
    "id" : null,
    "primitive" : true,
    "primitiveTypeName" : "FIXED_LEN_BYTE_ARRAY",
    "decimalMetadata" : {
      "precision" : 6,
      "scale" : 0
    },
    "typeLength" : 3
  },
  "encodings" : [ "PLAIN", "BIT_PACKED", "RLE" ],
  "startingPos" : 5550
}
```

vrozov: Parquet library behavior for DECIMAL statistics was changed in [PARQUET-686](https://issues.apache.org/jira/browse/PARQUET-686) (see [Parquet PR #367](https://github.com/apache/parquet-mr/pull/367)). I filed [PARQUET-1322](https://issues.apache.org/jira/browse/PARQUET-1322) to track statistics availability for DECIMAL types.

arina-ielchiieva: Vlad thanks for investigating the issue. Since it's Parquet problem, it can leave tests to be ignored just please add comment in each of them to indicated the root cause. @parthchandra are you ok with this approach?

parthchandra: I also had an offline chat with Vlad on this one. The problem is that Parquet has changed its behaviour and will not give us the stats for Decimal when we read footers. 
We have, therefore, no way of knowing whether Decimal stats are correct or not (even if they are correct) unless we try to hack something in Parquet. Hacking something in Parquet is not an option since that is exactly what this PR is trying to fix !
Also, we have never supported Decimal in Drill, so we do not have to consider backward compatibility. There are some users using Decimal (based on posts to the mailing list), but the old implementation never worked reliably so this will be an overall improvement for all parties.

+1. And thanks Vlad, Arina for pursuing this one to the end :)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/PageReader.java
```diff
@@ -445,25 +450,29 @@ public void clear(){
    * @throws IOException An IO related condition
    */
   void resetDefinitionLevelReader(int skipCount) throws IOException {
-    if (parentColumnReader.columnDescriptor.getMaxDefinitionLevel() != 0) {
-      throw new UnsupportedOperationException("Unsupoorted Operation");
-    }
+    Preconditions.checkState(parentColumnReader.columnDescriptor.getMaxDefinitionLevel() == 1);
```
vrozov: @sachouche Please review fix for DRILL-6447

sachouche: @vrozov,

I believe that max-definition can be either zero or one. Zero if all columns are null.

FYI - Did you also include the fix that I made in "VarLenBulkPageReader.java"

vrozov: In case `getMaxDefinitionLevel()` is zero, the `resetDefinitionLevelReader()` should not be called as it unconditionally reads `definitionLevels` that is only valid when `getMaxDefinitionLevel()` is not zero.

vrozov: Yes, I included the other fix as well. Please review changes to `VarLenBulkPageReader.java`.

sachouche: Thank you Vlad! 
LGTM

xiexingguang:
it means , once Upgrade Parquet MR dependencies , filter push down can support  type of varchar ?

arina-ielchiieva:
@xiexingguang no, support for varchar push down is not fully implemented yet.

xiexingguang:
@arina-ielchiieva     varchar push down  function  seems  import for us,  do you have a plan to supoort it fully ? if do have  plan, please  inform us .tks

vrozov:
@xiexingguang Please create new JIRA. PR is a place to discuss code modifications.

PR #1260, opened 2018-05-14
DRILL-6411: Make batch memory sizing logs uniform across all operators

ppadma:


ppadma:
@Ben-Zvi Boaz, please review.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
```diff
@@ -263,6 +260,8 @@ protected IterOutcome doWork() {
 
     flattenMemoryManager.updateOutgoingStats(outputRecords);
 
+    logger.debug("BATCH_STATS, outgoing:\n {}", new RecordBatchSizer(this));
```
Ben-Zvi: Every execution (even when logging does not happen; i.e., above "debug") does invoke the call "new RecordBatchSizer()", which adds time/memory overhead (and GC to remove these objects).
   Instead -- better surround this logger.debug() call with ``` if (logger.isDebugEnabled()) { ``` 
see similar examples elsewhere (e.g. Foreman.java)

ppadma: @Ben-Zvi Good point. Thank you. I fixed that. Please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/LateralJoinBatch.java
```diff
@@ -160,6 +160,23 @@ public IterOutcome innerNext() {
...
+    logger.debug("BATCH_STATS, incoming aggregate right: batch count : {}, avg bytes : {},  avg row bytes : {}, record count : {}",
+      batchMemoryManager.getNumIncomingBatches(JoinBatchMemoryManager.RIGHT_INDEX),
+      batchMemoryManager.getAvgInputBatchSize(JoinBatchMemoryManager.RIGHT_INDEX),
+      batchMemoryManager.getAvgInputRowWidth(JoinBatchMemoryManager.RIGHT_INDEX),
+      batchMemoryManager.getTotalInputRecords(JoinBatchMemoryManager.RIGHT_INDEX));
+
+    logger.debug("BATCH_STATS, outgoing aggregate: batch count : {}, avg bytes : {},  avg row bytes : {}, record count : {}",
+      batchMemoryManager.getNumOutgoingBatches(), batchMemoryManager.getAvgOutputBatchSize(),
+      batchMemoryManager.getAvgOutputRowWidth(), batchMemoryManager.getTotalOutputRecords());
+
```
Ben-Zvi: Why these 3 loggings were taken out of ```updateBatchMemoryManagerStats()``` ?
Aren't they the same for all the Binary operators ?
So later Hash Join would also need the same code ?

ppadma: The reason is they get logged with JoinBatchMemoryManager class logger and we will not know which operator it is.

PR #1262, opened 2018-05-15
DRILL-6374: Transitive Closure leads to TPCH Queries regressions and OOM when run concurrency test

vdiravka:
**The issue**: Using of DRILL_FILTER_ON_JOIN in early planning stage leads to impossibility of removing redundant Projects in main LOGICAL stage. 

**Solution**: The main idea to use TRANSITIVE_CLOSURE rules after main LOGICAL stage of rules. 

* New STRICT_EQUAL_IS_DISTINCT_FROM predicate for FilterJoinRules is added to pulled up redundant filters from Join condition to above filter (similar to DrillJoinRule). Redundant conditions in Joins can lead to further errors in planning.
* Change performing LogicalOptimizerRules from LOGICAL stage to PARTITION_PRUNING. 
LOGICAL stage shouldn't involve pruning:
https://github.com/vdiravka/drill/blob/87beac7e6f64ce5f6dedfa2b22cc5c9099caeaec/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java#L166
Hive logical optimizer rules involves pruning:
https://github.com/vdiravka/drill/blob/87beac7e6f64ce5f6dedfa2b22cc5c9099caeaec/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveStoragePlugin.java#L171

* Refactoring in DrillPushFilterPastProjectRule.java is made, which isn't related to the root cause of the issue.
* One unit test is enabled, since optimizations for aggregations are done before TC.


vdiravka:
@amansinha100 Please review

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillFilterJoinRules.java
```diff
@@ -59,20 +58,44 @@ public boolean apply(Join join, JoinRelType joinType, RexNode exp) {
...
       };
 
+  /** Predicate that always returns true for any filter in OUTER join, and only true
+   * for strict EQUAL or IS_DISTINCT_FROM conditions (without functions) over RexInputRef in INNER join.
+   * With this predicate, the filter expression that return true will be kept in the JOIN OP.
+   * Example:  INNER JOIN,   L.C1 = R.C2 will be kepted in JOIN.
+   *                         L.C3 + 100 = R.C4 + 100 and L.C5 < R.C6 will be pulled up into Filter above JOIN.
+   *           OUTER JOIN,   Keep any filter in JOIN.
+  */
+  public static final FilterJoinRule.Predicate STRICT_EQUAL_IS_DISTINCT_FROM =
```
amansinha100: It's not clear what the meaning of STRICT_EQUAL is here ?  Does the expression  a + 1 = b + 1 qualify ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -237,12 +237,8 @@ protected DrillRel convertToRawDrel(final RelNode relNode) throws SqlUnsupported
...
       final RelNode setOpTransposeNode = transform(PlannerType.HEP, PlannerPhase.PRE_LOGICAL_PLANNING, relNode);
 
-      // HEP Join Push Transitive Predicates
-      final RelNode transitiveClosureNode =
-          transform(PlannerType.HEP, PlannerPhase.TRANSITIVE_CLOSURE, setOpTransposeNode);
-
-      // HEP Directory pruning .
-      final RelNode pruned = transform(PlannerType.HEP_BOTTOM_UP, PlannerPhase.DIRECTORY_PRUNING, transitiveClosureNode);
+      // HEP Directory pruning.
+      final RelNode pruned = transform(PlannerType.HEP_BOTTOM_UP, PlannerPhase.DIRECTORY_PRUNING, setOpTransposeNode);
```
amansinha100: I recall that back when I had reviewed the original transitive closure PR, one of the things I had asked was why not do the TC before the DIRECTORY_PRUNING phase such that new predicates that are created as a result of TC would be used for pruning.   So, now it seems that doing the TC too early is causing a problem, is that right ?   Does that mean that going forward we won't support the scenario whether an newly inferred directory predicate (e.g dir0 = 2015) would be used for directory pruning  ?

vdiravka: You have asked it for `SetOpTranspose` rules, but it is topical for TC too.

Actually directory pruning is performed by two rules: `DirPruneScanFilterOnProjectRule` and `DirPruneScanFilterOnScanRule`.
They matches for EnumerableTableScan and performed in `DIRECTORY_PRUNING` stage and matches for DrillScanRel and performed in `PARTITION_PRUNING` stage. 

In current implementation the TC is used before `PARTITION_PRUNING` stage, so directory pruning performed correctly. 
I have added test case to verify it.

amansinha100: @vdiravka thx for the explanation and for adding the test for TC done before partition pruning.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillFilterJoinRules.java
amansinha100: Yeah, basically STRICT_EQUAL means only a = b types of predicates.

amansinha100:
LGTM.  +1

PR #1264, opened 2018-05-16
 DRILL-6422: Update guava to 23.0 and shade it

vvysotskyi:
Description of made changes:
- Fixed all compilation errors after updating guava version;
- Removed usage of some deprecated API;
- Shaded guava version and added dependencies to the shaded jar;
- To avoid bug with highlighting in IDE, unpacked shaded guava sources to the target source directory;
- Increased size for JDBC driver to 38000000 and 36000000 since both unshaded and shaded versions of guava should be included
- Replaced all guava imports with shaded ones and banned imports to the unshaded classes
- Replaced some guava methods for collections creating by constructors according to the recommendations in it's javadocs
- Replaced usage of some guava interfaces (Function, Predicate etc.) and methods (Sets.filter(), Iterables.tryFind, transform etc.) by java ones
- Added methods to convert shaded ImmutableList to unshaded one for the cases when another library has in its API ImmutableList.


vvysotskyi:
@vrozov, could you please take a look at this PR?

kkhatua:
@vvysotskyi  there are merge conflict warnings showing up. Some of these, appear to affect files that haven't been updated since July 2017 (e.g. exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java ). Can you check if you need to rebase?

vvysotskyi:
This PR was on the last master before the last batch commits. `DrillOperatorTable.java` class contains changes, merged in https://github.com/apache/drill/pull/1268 (it was merged in this last batch commit)  and perhaps this PR caused merged conflicts in this class. 
I'll resolve all merge conflicts once the active review stage comes.

priteshm:
@vrozov can you review this today?

vrozov:
@priteshm I'll start to review the PR if time permits today, but with the number of changes involved, I doubt that it can be done in a day or two.

priteshm:
@vrozov would you be able to wrap up the review this week? I am assuming most of the changes are related to shading the library.

contrib/guava-shaded/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+              <createDependencyReducedPom>false</createDependencyReducedPom>
+              <artifactSet>
+                <includes>
+                  <include>com.google.guava:*</include>
+                </includes>
+              </artifactSet>
+              <relocations>
+                <relocation>
+                  <pattern>com.google.common</pattern>
+                  <shadedPattern>org.apache.drill.shaded.com.google.common</shadedPattern>
```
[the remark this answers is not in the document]
vvysotskyi: Added `guava` to the shaded package name.

Only guava classes were packed into the jar, but imports for other classes which match `com.google` pattern will be modified. So I left the same pattern to avoid this compilation error.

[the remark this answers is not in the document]
vrozov: Sorry, I don't follow. Don't we want all classes included into the guava library to be relocated?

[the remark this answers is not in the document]
vvysotskyi: Yes, we want to relocate all classes, included to the jar, but if these classes use other classes whose package matches the pattern, but they belong to the other lib, their packages also will be relocated. And since these classes with relocated packages do not exist, a build will fail. Therefore I left pattern the same. 

Also, since guava has a single package which does not match the pattern: `com.google.thirdparty.publicsuffix` which was not used in Drill or any other project which Drill use, I don't think it may cause problems.

common/pom.xml
```diff
@@ -86,6 +86,11 @@
       <artifactId>jackson-databind</artifactId>
     </dependency>
 
+    <dependency>
```
vrozov: Was guava a transitive dependency before? Should guava be added as a dependency to the drill root pom?

distribution/pom.xml
```diff
@@ -108,6 +108,10 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
```
vrozov: Why is this necessary?

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
```
vrozov: use `drill-shaded` as a parent. When drill needs more libraries to be shaded, `drill-shaded` will be a common parent.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+    <groupId>org.apache</groupId>
+    <artifactId>apache</artifactId>
+    <version>18</version>
+  </parent>
+
+  <groupId>org.apache.drill</groupId>
+  <artifactId>guava-shaded</artifactId>
+  <version>${dep.guava.version}</version>
```
vrozov: version needs to be specified explicitly.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+    <assembly.attach>false</assembly.attach>
+    <dep.guava.version>23.0</dep.guava.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${dep.guava.version}</version>
+      <type>jar</type>
```
vrozov: jar is the default type.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+
+    <sourceDirectory>${project.build.directory}/shaded-sources</sourceDirectory>
+
+    <plugins>
+      <plugin>
```
vrozov: move to the parent pom. Bind `default-compile` and `default-testCompile` to `phase` `none`. The same for other plugins.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <createSourcesJar>true</createSourcesJar>
+              <shadeSourcesContent>true</shadeSourcesContent>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
+              <artifactSet>
+                <includes>
+                  <include>com.google.guava:*</include>
```
vrozov: limit include to "com.google.guava:guava"

drill-shaded/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache</groupId>
+    <artifactId>apache</artifactId>
+    <version>18</version>
```
vrozov: add `relativePath`.

vvysotskyi:
@vrozov, I made requested changes, could you please take a look again?

contrib/guava-shaded/pom.xml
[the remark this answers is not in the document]
vvysotskyi: I agree that in this case, it may cause problems. Also relocated `com.google.thirdparty` to avoid this issue.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.drill</groupId>
+    <artifactId>drill-shaded</artifactId>
+    <version>1.0</version>
+    <relativePath/>
```
vrozov: use `relativePath` only in pom where parent pom is not in the parent directory (for example when the parent is apache pom).

drill-shaded/guava-shaded/pom.xml
vrozov: I mean bind to `phase` `none` execution `id` that need to be disabled. For compiler the plugin it is `default-compile`.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,133 @@
...
+              <shadeSourcesContent>true</shadeSourcesContent>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
+              <artifactSet>
+                <includes>
+                  <include>com.google.guava:guava</include>
+                </includes>
+              </artifactSet>
+              <relocations>
+                <relocation>
+                  <pattern>com.google.common</pattern>
```
vrozov: I still don't see why it is necessary to handle `common` differently from `thirdparty`.

drill-shaded/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+  <modules>
+    <module>guava-shaded</module>
+  </modules>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
```
vrozov: Why is this necessary?

pom.xml
```diff
@@ -3334,5 +3351,6 @@
     <module>exec</module>
     <module>drill-yarn</module>
     <module>distribution</module>
+    <module>drill-shaded</module>
```
vrozov: Remove. All artifacts in drill-shaded should be built once and published to maven. There is no need to build them in each drill build.

drill-shaded/pom.xml
vvysotskyi: Guava classes were compiled with 8 JDK, so this version should be specified in `maven-compiler-plugin` to avoid errors when lib is shading.

drill-shaded/guava-shaded/pom.xml
vvysotskyi: Sorry, but either I don't understand your intention or you missed this change in the last commit https://github.com/apache/drill/pull/1264/commits/872afebe0fb0fba68f971aaf4243716814753c43. Could you please take a look at this commit or clarify, what should be done.

pom.xml
vvysotskyi: I don't think that it is the right approach to publish artifacts to maven.
When we will make some changes in pom files, for example, we should deploy it again. And for the most of the contributors, it is unknown, how to deploy it to the maven, and I suppose most of us don't have an access to do that. So I think it should be built with drill build.

vrozov: What type of changes do you expect in pom.xml? Why would you prefer to build the same artifact over and over again even though it does not change?

drill-shaded/guava-shaded/pom.xml
vrozov: It meaningless to bind `default-compile` or `default-testCompile` execution `id` for `maven-surefire-plugin` as it does not define such execution `id`. The intention of binding default execution `id` to phase `none` is to disable execution of those plugins when desired, instead of relying on other settings (like `skip` or `skipTests`).

drill-shaded/pom.xml
vrozov: AFAIK class version should not change when a class is relocated. If guava classes have version 52, the same version will be in the shaded library.

common/src/main/java/org/apache/drill/common/collections/MapWithOrdinal.java
```diff
@@ -129,22 +128,16 @@ public void clear() {
 
     @Override
     public Collection<V> values() {
-      return Lists.newArrayList(Iterables.transform(secondary.entries(), new Function<IntObjectMap.Entry<V>, V>() {
-        @Override
-        public V apply(IntObjectMap.Entry<V> entry) {
-          return Preconditions.checkNotNull(entry).value();
-        }
-      }));
+      return Lists.newArrayList(Iterables.transform(secondary.entries(),
```
vrozov: Why not `secondary.values()`?

common/src/main/java/org/apache/drill/common/collections/MapWithOrdinal.java
```diff
@@ -129,22 +128,16 @@ public void clear() {
...
 
     @Override
     public Set<Entry<K, V>> entrySet() {
-      return Sets.newHashSet(Iterables.transform(primary.entrySet(), new Function<Entry<K, Entry<Integer, V>>, Entry<K, V>>() {
-        @Override
-        public Entry<K, V> apply(Entry<K, Entry<Integer, V>> entry) {
-          return new AbstractMap.SimpleImmutableEntry<>(entry.getKey(), entry.getValue().getValue());
-        }
-      }));
+      return primary.entrySet().stream()
```
vrozov: What is the purpose of that code?

common/src/main/java/org/apache/drill/common/collections/MapWithOrdinal.java
```diff
@@ -54,7 +53,7 @@
 public class MapWithOrdinal<K, V> implements Map<K, V> {
   private final static Logger logger = LoggerFactory.getLogger(MapWithOrdinal.class);
 
-  private final Map<K, Entry<Integer, V>> primary = Maps.newLinkedHashMap();
+  private final Map<K, Entry<Integer, V>> primary = new LinkedHashMap<>();
   private final IntObjectHashMap<V> secondary = new IntObjectHashMap<>();
 
   private final Map<K, V> delegate = new Map<K, V>() {
```
vrozov: What is the purpose of the `delegate`? I'd recommend limiting this class changes to updated imports only and refactor it in a separate JIRA/PR.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/schema/HiveDatabaseSchema.java
```diff
@@ -69,7 +70,7 @@ public Table getTable(String tableName) {
         tables = Sets.newHashSet(mClient.getTableNames(this.name, schemaConfig.getIgnoreAuthErrors()));
```
vrozov: new `HashSet<>(...)`?

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/schema/HiveDatabaseSchema.java
```diff
@@ -69,7 +70,7 @@ public Table getTable(String tableName) {
...
       } catch (final TException e) {
         logger.warn("Failure while attempting to access HiveDatabase '{}'.", this.name, e.getCause());
-        tables = Sets.newHashSet(); // empty set.
+        tables = new HashSet<>(); // empty set.
```
vrozov: `Collections.emptySet()`?

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.drill</groupId>
+    <artifactId>drill-shaded</artifactId>
+    <version>1.0</version>
+  </parent>
+
+  <artifactId>guava-shaded</artifactId>
```
vrozov: consider `drill-shaded-guava`

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+  <version>23.0</version>
+  <name>drill-shaded/guava-shaded</name>
+
+  <properties>
+    <!-- skip source archives -->
+    <assembly.attach>false</assembly.attach>
```
vrozov: Why is the property necessary?

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <dep.guava.version>23.0</dep.guava.version>
```
vrozov: Consider `guava.version` as a more common/consistent way to define a dependency version.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>${dep.guava.version}</version>
+    </dependency>
+    <dependency>
```
vrozov: Why is the dependency necessary?

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+      <version>2.0.18</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
vrozov: Move common settings to `pluginManagement` in the parent pom.

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+        <version>3.1.0</version>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <createSourcesJar>true</createSourcesJar>
+              <createDependencyReducedPom>false</createDependencyReducedPom>
```
vrozov: Why `false`?

drill-shaded/guava-shaded/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>3.1.0</version>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <createSourcesJar>true</createSourcesJar>
```
vrozov: `shadeSourcesContent`?

pom.xml
```diff
@@ -42,7 +42,8 @@
     <proto.cas.path>${project.basedir}/src/main/protobuf/</proto.cas.path>
     <dep.junit.version>4.12</dep.junit.version>
     <dep.slf4j.version>1.7.6</dep.slf4j.version>
-    <dep.guava.version>18.0</dep.guava.version>
+    <dep.guava.version>23.0</dep.guava.version>
```
vrozov: `shaded.guava.version`?

pom.xml
```diff
@@ -42,7 +42,8 @@
...
+    <dep.guava.transitive.version>19.0</dep.guava.transitive.version>
```
vrozov: `guava.version`

pom.xml
```diff
@@ -984,6 +984,23 @@
...
+            <artifactId>guava</artifactId>
+            <groupId>com.google.guava</groupId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava</artifactId>
+        <version>${dep.guava.transitive.version}</version>
+        <scope>compile</scope>
```
vrozov: remove `scope`

pom.xml
```diff
@@ -984,6 +984,23 @@
   <!-- Managed Dependencies -->
   <dependencyManagement>
     <dependencies>
+      <dependency>
+        <groupId>org.apache.drill</groupId>
+        <artifactId>guava-shaded</artifactId>
+        <version>${dep.guava.version}</version>
+        <exclusions>
```
vrozov: See my comment for `createDependencyReducedPom`

common/src/main/java/org/apache/drill/common/collections/MapWithOrdinal.java
vvysotskyi: OK, I have removed all these changes and will open a separate PR with them. So now this PR contains only changes required for compiling the code.

drill-shaded/guava-shaded/pom.xml
vrozov: I don't understand and/or see the problem.

drill-shaded/drill-shaded-guava/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <configuration>
+          <createSourcesJar>true</createSourcesJar>
```
vrozov: Should it be part of common configuration in `PluginManagement`?

drill-shaded/drill-shaded-guava/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+          <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+          <dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
+          <artifactSet>
+            <includes>
+              <include>com.google.guava:guava</include>
+            </includes>
+          </artifactSet>
+          <relocations>
+            <relocation>
+              <pattern>com.google.common</pattern>
```
vrozov: I still don't see why is it necessary to explicitly spell out "common"

drill-shaded/drill-shaded-guava/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+      <version>${guava.version}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <configuration>
```
vrozov: Add "ApacheLicenseResource" and "ApacheNoticeResource" transformers (preferrably to a common configuration)

drill-shaded/drill-shaded-guava/pom.xml
vvysotskyi: Yes, it should. Thanks, moved the common part to ` PluginManagement` in the parent pom.

drill-shaded/drill-shaded-guava/pom.xml
vvysotskyi: Thanks, added both these transformers to the common configuration.

drill-shaded/guava-shaded/pom.xml
vvysotskyi: If we have both `createSourcesJar` and `shadeSourcesContent` set to true and handle `com.google` pattern in relocation, all imports that match this pattern will be also relocated for source files, even if those classes don't belong to `guava` lib.

For example, with such config, in the source for `Preconditions` will be available import `org.apache.drill.shaded.guava.com.google.errorprone.annotations.CanIgnoreReturnValue`, but class `CanIgnoreReturnValue` was in `error_prone_annotations` jar, not in `guava`.

drill-shaded/drill-shaded-guava/pom.xml
vvysotskyi: I have described the reason in this comment: https://github.com/apache/drill/pull/1264#discussion_r202641140

drill-shaded/guava-shaded/pom.xml
vrozov: I see, but it is actually not a problem as those sources are provided for debugging, not for compiling shaded jars. Note that maven default Shader simply does regular expression replacement from `pattern` to `shadedPattern` when `shadeSourcesContent` is `true`. Please add a comment explaining why two patterns are used or use a single common pattern.

vrozov:
I'd recommend splitting this PR in two. The first PR to shade newer version of guava and the second PR to change drill code to use the shaded jar. The first PR needs to be merged first, so shaded jar can be published to maven.

drill-shaded/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+              <phase>package</phase>
+              <goals>
+                <goal>shade</goal>
+              </goals>
+            </execution>
+          </executions>
+          <configuration>
+            <createSourcesJar>true</createSourcesJar>
+            <shadeSourcesContent>true</shadeSourcesContent>
+            <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
```
vrozov: Likely optional for guava as dependencies are used for compile time annotations.

drill-shaded/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+            <dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
+            <transformers>
+              <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheLicenseResourceTransformer"/>
+              <transformer implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer"/>
```
vrozov: `addHeader`? see https://maven.apache.org/plugins/maven-shade-plugin/examples/resource-transformers.html

vrozov:
Shaded jar needs to be published to maven no matter what. In case of a single PR, publishing to maven needs to happen **before** PR is merged. In the case of two PR, there is no such dependency. The first PR can be merged without publishing shaded jars to maven. Otherwise LGTM.

vvysotskyi:
@vrozov, I don't mind of creating a separate PR, but I still think that it is unnecessary, since we need to merge the first PR, publish shaded guava package to the maven manually and then merge the second PR.

Could you please explain how to manually publish shaded guava package to the maven repository?

vrozov:
@vvysotskyi Correct, but note that with a single PR it is necessary to complete almost the same actions by a single committer in a short period of time (due to merge conflicts) while two PRs allow avoiding such dependency and one person can merge the first PR, another person can publish artifacts to maven and a third person can wait for the artifacts to become available before resolving conflicts and merging the second PR.

To manually publish artifacts, see https://repository.apache.org or use `maven deploy`. I don't remember who (all committers or only PMCs) are allowed to publish to the Apache maven repository.

vrozov:
@vvysotskyi Please make sure to publish the artifact to the Apache repository (not a MapR repository that is not mirrored).

drill-shaded/pom.xml
vvysotskyi: Added it back, since without this change when for example opened Preconditions class, errorprone annotations are not recognized.

PR #1265, opened 2018-05-16
DRILL-5305: Query Profile must display Query ID

kkhatua:
Introduced change to the Protobuf to inject the text-equivalent of the QueryID into the profile. This way, the profile's file name can be changed, but restored back based on this new field.
The Profile UI also shows the Query ID, though this is not inferred from this new field, for sake of backward compatibility with older profiles.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
@@ -63,7 +63,7 @@ public class ProfileWrapper {
 
   public ProfileWrapper(final QueryProfile profile, DrillConfig drillConfig) {
     this.profile = profile;
-    this.id = QueryIdHelper.getQueryId(profile.getId());
+    this.id = profile.getId().hasText() ? profile.getId().getText() : QueryIdHelper.getQueryId(profile.getId());
     //Generating Operator Name map (DRILL-6140)
     String profileTextPlan = profile.hasPlan() ? profile.getPlan() : "" ;
     generateOpMap(profileTextPlan);
--- a/exec/java-exec/src/main/resources/rest/profile/profile.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/profile.ftl
@@ -165,7 +165,7 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
   <#assign queued = queueName != "" && queueName != "-" />
 
   <div class="page-header"></div>
-  <h3>Query Profile</h3>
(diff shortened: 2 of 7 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -63,7 +63,7 @@
 
   public ProfileWrapper(final QueryProfile profile, DrillConfig drillConfig) {
     this.profile = profile;
-    this.id = QueryIdHelper.getQueryId(profile.getId());
+    this.id = profile.getId().hasText() ? profile.getId().getText() : QueryIdHelper.getQueryId(profile.getId());
```
ppadma: both of them should return the same value. isn't it ?

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -165,7 +165,7 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
   <#assign queued = queueName != "" && queueName != "-" />
 
   <div class="page-header"></div>
-  <h3>Query Profile</h3>
+  <h3>Query Profile : <span style='font-size:85%'>${model.getQueryId()}</span></h3>
```
ppadma: what does this return ? text ?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
kkhatua: Yes. But if the profile is of an older format, ``hasText()` will ensure that the query ID text is decoded from the ID components: part1 & part2.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
kkhatua: Yes. The backing ProfileWrapper returns the queryID as a text.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
ppadma: why not decode always instead of doing the check ? It is not a big deal either way.
LGTM +1.

kkhatua: The JIRA's goal is to embed the query ID text within the profile. Since it is within the profile, it would be more efficient to use that value versus decoding repeatedly. Hence, it is an optional part of the QueryID (see proto file), so that we can fall back to the decoding approach.

PR #1266, opened 2018-05-16
DRILL-6423: Export query result as a CSV file

kkhatua:
Added option for user specified delimiter
Also, we show query Id on results page as alink for popping out a new window

### DIFF
--- a/exec/java-exec/src/main/resources/rest/query/result.ftl
+++ b/exec/java-exec/src/main/resources/rest/query/result.ftl
@@ -30,6 +30,22 @@
   <a href="/queries">back</a><br/>
   <div class="page-header">
   </div>
+  <div>
+  <table><tr>
+    <td align='left'>
+      <button type="button"  title="Open in new window" onclick="popOutProfile('${model.getQueryId()}');" class="btn btn-default btn-sm">
+      <b>Query Profile:</b> ${model.getQueryId()} <span class="glyphicon glyphicon-new-window"/></button>
+     </td><td align="right" width="100%">
+       <div class="input-group">
+         <span class="input-group-addon" style="font-size:95%">Delimiter </span>
+         <input id="delimitBy" type="text" class="form-control input-sm" name="delimitBy" title="Specify delimiter" placeholder="Required" maxlength="2" size="2" value=",">
+       </div></td><td>
+       <button type="button"  title="Export visible table as CSV" onclick="exportTableAsCsv('${model.getQueryId()}');" class="btn btn-default btn-sm">
+       <b>Export </b> <span class="glyphicon glyphicon-export"/></button>
+     </td>
+  </tr>
+  </table>
+  </div>
   <#if model.isEmpty()>
     <div class="jumbotron">
       <p class="lead">No result found.</p>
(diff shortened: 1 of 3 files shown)


kkhatua:
Tested with a query on a TPC-H customer table with 1000 row limit. 
Things to note:

- The table is filtered for showing only 10 rows of 'FURNITURE' market segment, and ordered by the customer with the lowest (i.e. most negative) balances.
- Exporting this prompts to save the result as a CSV (default delimiter is comma, as seen on the top right). under the profile name.
- The query's profile can be opened in a new window with the link on the top left.
- Since there are only 10 rows visible, only the 10 rows will be exported into the CSV.

**Screenshot:**
image

kfaraaz:
User should be able to write all the rows returned by the query, to a file in one go ?

kkhatua:
There is a drop-down option to show `ALL` the rows. This way, you can export only what you see, or everything.

kfaraaz:
cool, thanks.

kkhatua:
@vdiravka 
1. The maximum number of rows is dependent on the client machine resources. 
This is a client side Javascript implementation, since the result set at the point of rendering is already on the client. The DataTables library will hide rows based on pagination, table length and filter conditions. The backend Javascript simply traverses the visible rows and constructs an in-memory CSV dataset, before prompting for it to be saved.

2. I haven't tried with the case of repeated values. As long as the result can be rendered on the WebUI in a specific format, this will save it as a CSV (with a delimiter specified by the user).

priteshm:
Nice work, @kkhatua !

exec/java-exec/src/main/resources/rest/query/result.ftl
```diff
@@ -59,11 +75,67 @@
...
+
+    // Add the link to your DOM
+    document.body.appendChild(downloadElem);
+
+    // Launch the download prompt
+    downloadElem.click();
+}
+
+function exportTableAsCsv(queryId) {
+    var filename = queryId+'.csv';
```
arina-ielchiieva: space after `queryId`.

exec/java-exec/src/main/resources/rest/query/result.ftl
```diff
@@ -59,11 +75,67 @@
...
+
+    //Pop out profile (needed to avoid losing query results)
+    function popOutProfile(queryId) {
+      var profileUrl = location.protocol+'//'+ location.host+'/profiles/'+queryId;
+      var tgtWindow = '_blank';
+      window.open(profileUrl, tgtWindow);
+    }
+
+//Ref: https://jsfiddle.net/gengns/j1jm2tjx/
+function download_csv(csvRecords, filename) {
```
arina-ielchiieva: `downloadCsv`

exec/java-exec/src/main/resources/rest/query/result.ftl
```diff
@@ -30,6 +30,22 @@
...
+  <table><tr>
+    <td align='left'>
+      <button type="button"  title="Open in new window" onclick="popOutProfile('${model.getQueryId()}');" class="btn btn-default btn-sm">
+      <b>Query Profile:</b> ${model.getQueryId()} <span class="glyphicon glyphicon-new-window"/></button>
+     </td><td align="right" width="100%">
+       <div class="input-group">
+         <span class="input-group-addon" style="font-size:95%">Delimiter </span>
+         <input id="delimitBy" type="text" class="form-control input-sm" name="delimitBy" title="Specify delimiter" placeholder="Required" maxlength="2" size="2" value=",">
+       </div></td><td>
+       <button type="button"  title="Export visible table as CSV" onclick="exportTableAsCsv('${model.getQueryId()}');" class="btn btn-default btn-sm">
```
arina-ielchiieva: Maybe we can add tooltip that explains that we export only visible table? If user needs to export all results, then should chose all rows to display?

kkhatua:
@arina-ielchiieva I've done the changes.

arina-ielchiieva:
+1

PR #1268, opened 2018-05-17
DRILL-4525: Allow SqlBetweenOperator to accept LOWER_OPERAND and UPPER_OPERAND with different types

KazydubB:
- Made code cleanup
- Fixed failing test

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java
@@ -22,6 +22,7 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlFunction;
+import org.apache.calcite.sql.fun.SqlBetweenOperator;
 import org.apache.drill.common.expression.FunctionCallFactory;
 import org.apache.drill.exec.expr.fn.DrillFuncHolder;
 import org.apache.drill.exec.expr.fn.FunctionImplementationRegistry;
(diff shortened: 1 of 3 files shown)


arina-ielchiieva:
@vvysotskyi please review.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java
```diff
@@ -158,26 +159,39 @@ private void populateFromWithoutTypeInference(SqlIdentifier opName, SqlFunctionC
...
             getFunctionListWithInference(calciteOperator.getName()));
+      } else if (calciteOperator instanceof SqlBetweenOperator) {
+        // During the procedure of converting to RexNode,
+        // StandardConvertletTable.convertBetween expects the SqlOperator to be a subclass of SqlBetweenOperator
+        final SqlBetweenOperator sqlBetweenOperator = (SqlBetweenOperator) calciteOperator;
+        wrapper = new DrillCalciteSqlBetweenOperatorWrapper(sqlBetweenOperator);
       } else {
-        final String drillOpName = FunctionCallFactory.replaceOpWithFuncName(calciteOperator.getName());
+        final String drillOpName;
+        // For UNARY_MINUS (-) or UNARY_PLUS (+), we do not rename them as function_add or function_subtract.
```
vvysotskyi: Could you please explain a little bit more the reason for these changes if they are still valid and please add a unit test for this.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java
```diff
@@ -43,7 +44,7 @@
  * {@link #inner SqlStdOperatorTable}, and Drill User Defined Functions.
  */
 public class DrillOperatorTable extends SqlStdOperatorTable {
-//  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DrillOperatorTable.class);
+  // private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DrillOperatorTable.class);
```
vvysotskyi: Please remove this string.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java
KazydubB: If left unchanged, for UNARY_MINUS and UNARY_PLUS operators Calcite's behaviour will be applied. The wrapper, on the other hand, ensures that customized behaviour will be adopted.
(These two operators are treated differently because as described in the comment found in file their drillOpName collides with binary MINUS and PLUS respectively.)

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java
KazydubB: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillOperatorTable.java
vvysotskyi: Thanks for the explanation.

PR #1270, opened 2018-05-17
DRILL-6425: Upgrade org.ojai:ojai version

vrozov:


arina-ielchiieva:
+1

arina-ielchiieva:
@vrozov  looks like build fails with mapr profile, please check.

pom.xml
```diff
@@ -52,7 +52,7 @@
     <jackson.version>2.7.9</jackson.version>
     <jackson.databind.version>2.7.9.1</jackson.databind.version>
     <mapr.release.version>5.2.1-mapr</mapr.release.version>
-    <ojai.version>1.1</ojai.version>
+    <ojai.version>2.0.1-mapr-1804</ojai.version>
```
parthchandra: maven repository has a ojai:2.0 and a ojai:2.0-mapr-1710. I don't see the 2.0.1 version. That apart, what is the difference between these two and why should Drill prefer one over the other?

vrozov: Please see http://repository.mapr.com/nexus/content/groups/mapr-public/org/ojai/ojai/2.0.1-mapr-1804/. The latest version is required to avoid `2.0.1-SNAPSHOT` dependency used by maprdb 6.0.1 otherwise Drill can simply drop that dependency from the `DependencyManagement`

vrozov:
@arina-ielchiieva I guess [mapr maven repository](http://repository.mapr.com/maven/) was down when you tried to build. Can you try one more time.

vrozov:
Please don't commit yet. It may be necessary to align this upgrade with the upgrade of mapr dependencies to the latest version.

PR #1271, opened 2018-05-18
DRILL-6418: Handle Schema change in Unnest And Lateral for unnest fie…

sohami:
…ld / non-unnest field

            Note: Changed Lateral to handle non-empty right batch with OK_NEW_SCHEMA


sohami:
@parthchandra - please help to review

PR #1272, opened 2018-05-18
DRILL-5977: Filter Pushdown in Drill-Kafka plugin

aravi5:
# Implementation of Filter Pushdown in Drill-Kafka plugin

This PR contains changes for implementing `Filter Pushdown` feature for conditions on `Metadata fields` such as *Timestamps*, *Offsets* and *Partitions*. Below is a high level description, to help the reviewers.

[Design document is available here](https://docs.google.com/document/d/1SZ4wO4Ii4nAHwgWbY6JJynPOseCha0DBBZJ2ig1RO0M/edit#heading=h.uvq8jbet5xdh)

## Topics, Partitions, Offsets and Timestamps

Apache Kafka is a distributed `publisher-subscriber` system. Kafka consists of `Topics` which are something like channels. Producer produce to topics and Consumers subscribe to topics. In the Drill world, every topic is considered as a `Table`.

To achieve load balancing and parallelism, the concept of `Partitions` was introduced. Each topic can have multiple partitions and messages produced to topic are distributed across partitions (producers can also specify partition to produce). 

Kafka is essentially a distributed log, so every message within a `Topic-Partition` is identified by a unique `Offset`. This offset represents the distance of any message from the _beginning_ of the topic-partition.

Since Kafka can be used to store events / event-logs, one would look to Kafka to find out when an event occurred. So every messages is thus associated with a `Timestamp`. Kafka allows the producer (user application) to specify the timestamp value or it can add its (server / broker) timestamp before storing the message / log in topic-partition. Note that since user can specify the timestamp value, it is possible to that the timestamp values can appear out of order as we consume from beginning to end.

## Drill-Kafka Plugin

Drill-Kafka Plugin is very useful to explore the data and perform analytics. Currently, all of the data that resides in a topic has to be consumed by Drill before applying filters. However, it is possible to reduce the number of messages read from Kafka based on conditions provided on fields - `kafkaMsgTimestamp`, `kafkaMsgOffset` and `kafkaPartitionId`.

Scan specification for Drill-Kafka plugin is a collection of specification for individual partitions within each topic. This specification consists of `topicName`, `partitionId`, `startOffset` and `endOffset`. 

## Filter Pushdown Implementation

Filter pushdown implementation is governed by the idea that conditions on `metadata fields` can be translated to modifying the `startOffset` and `endOffset` for each partition. Thus reducing the amount of data read.

### Pushdown Timestamp
A common query when querying a event log system involves a query with conditions on timestamp field. For Drill-Kafka plugin, `kafkaMsgTimestamp` field maps to timestamp stored as a part of each message in Kafka. A pushdown for following compare functions is supported `equal`, `greater_than` and `greater_than_or_equal_to`.

Kafka exposes a `Consumer API` to obtain earliest offset for a given timestamp value.
```
public java.util.Map<TopicPartition,OffsetAndTimestamp> offsetsForTimes(java.util.Map<TopicPartition,java.lang.Long> timestampsToSearch)
```
This API is used to figure out the `startOffset` for each partition in a topic. Note that the timestamps may not appear in increasing order when reading from a Kafka topic because Kafka gives user flexibility to define timestamp for a message. However, the above API returns the first offset (from the beginning of a topic partition) where timestamp is *greater or equal* to timestamp requested. Thus, we cannot support pushdown on `less_than` or `less_than_or_equal_to` because a lesser timestamp may exist beyond `endOffset` computed.

### Pushdown Offset
Conditions on `kafkaMsgOffset` will be pushed down for `equal`, `greater_than`, `greater_than_or_equal_to`, `less_than` and `less_than_or_equal_to` functions.

### Pushdown Partition
Conditions on `kafkaPartitionId` will help limit number of partitions to be scanned. (Useful for data exploration).

Even after creating new scan specification, based on the filter conditions, we will have to re-apply the filters because the optimization only restricts the scan range. 

### Pushdown with OR
At any level of the expression tree, OR node will return a new scan specification only if ALL conditions under OR support pushdown.

## Handling corner-cases
Kafka supports `Time-To-Live` thus messages can expire as at some point in time. So it is possible for queries to include expired values / non-existent values for timestamps and offsets. The implementation handles these cases as well.

## Unit Test
Unit tests are added and cover queries (both with pushdown supported and not supported).
```
Tests run: 13, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 158.327 sec - in org.apache.drill.exec.store.kafka.KafkaFilterPushdownTest
```


aravi5:
@akumarb2010, @kameshb, @kkhatua please review the changes for filter pushdown implementation.

akumarb2010:
@aravi5  Sorry for the delay in review and thanks for implementing this nice feature.

Before starting the code review, I have few comments on push down design.

1. *KafkaMsgOffset* predicates has to be partition specific right? We should not be applying these predicates globally across the partitions. For example p1[startOffset=1000, endOffset=2000], p2[1500,5000], So in this case, always better to consider the offsets per partition. But in test cases, I am not partition specific predicates.

2. Good to see that you have considered multiple scenarios for *kafkaMsgTimestamp* predicates and the above point will also applicable for *kafkaMsgTimestamp* as well. We might need to consider per partition specific *kafkaMsgTimestamp* predicates.  But this might cause the issues, as * offsetsForTimes* method is a blocking and it can block indefinitely if user provides wrong partition.

Can you please clarify above two comments?

aravi5:
@akumarb2010 - user has the flexibility to provide conditions w.r.t any of the metadata fields (`kafkaMsgOffset`, `kafkaMsgTimestamp`, `kafkaPartitionId`). The conditions on these metadata fields is pushed down and translated to a **Scan Spec** which is a collection of partition specific scan spec.

Let us consider the following Scan spec before pushdown.

```
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=0, startOffset=0, endOffset=1500)
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=1, startOffset=2000, endOffset=3000)
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=2, startOffset=25, endOffset=3500)
```

If the predicates are only on `kafkaMsgOffset` (for example `SELECT * FROM kafka.LogEventStream WHERE kafkaMsgOffset >= 1000 AND kafkaMsgOffset < 2000)`, this will apply the pushdown to ALL partitions within a topic. If there is a partition where such offsets do not exist (either the offsets have expired or messages for those offsets are yet to be produced), then such partition will not be scanned.

The scan spec in this case would be
```
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=0, startOffset=1000, endOffset=1500)
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=1, startOffset=1200, endOffset=2000)
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=2, startOffset=1000, endOffset=2000)
```

I am not sure if we should introduce per partition semantics in the query and would prefer keeping it generic. The *scenario you mentioned in (1)* can still be addressed with following predicates

```
SELECT * FROM kafka.LogEventStream WHERE (kafkaPartitionId = 1 AND kafkaMsgOffset >= 1000 AND kafkaMsgOffset < 2000) OR (kafkaPartitionId = 2 AND kafkaMsgOffset >= 1500 AND kafkaMsgOffset < 5000)
```

The scan spec in this case would be as follows (notice that there is no `partitionId 0`).
```
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=1, startOffset=1200, endOffset=2000)
KafkaPartitionScanSpec(topicName=TopicTable, partitionId=2, startOffset=1000, endOffset=2000)
```

This applies to conditions on `kafkaMsgTimestamp` as well. Most common use case is - user wants to view messages that belong to a specific time window (not very specific about the partition). This can be done by having predicates on `kafkaMsgTimestamp` alone without having to specify `kafkaPartitionId`.
```
SELECT * FROM kafka.LogEventStream WHERE kafkaMsgTimestamp > 1527092007199 AND kafkaMsgOffset < 1527092031717
```

However user can also specify conditions specific to partition - *scenario mentioned in (2)*

```
SELECT * FROM kafka.LogEventStream WHERE (kafkaPartitionId = 1 AND kafkaMsgTimestamp > 1527092007199 AND kafkaMsgOffset < 1527092031717) OR (kafkaPartitionId = 2 AND kafkaMsgTimestamp > 1527092133501)
```

This would also avoid the problem of calling `offsetsForTimes` on non-existing partitions. Since such partitions are filtered out from the scan spec.

akumarb2010:
@aravi5  Thanks for providing all the details. Please find my comments as below.

>> If the predicates are only on kafkaMsgOffset (for example SELECT * FROM kafka.LogEventStream WHERE kafkaMsgOffset >= 1000 AND kafkaMsgOffset < 2000), this will apply the pushdown to ALL partitions within a topic. If there is a partition where such offsets do not exist (either the offsets have expired or messages for those offsets are yet to be produced), then such partition will not be scanned.

In Kafka, `offset` scope itself is per partition. I am unable to find any use case, we can take the range of offsets and apply on all partitions. In most of the scenario's they may not be valid offsets.

IMHO, we should only apply predicate pushdown where have exact scan specs. 

For example, in case below cases we can apply the predicates without any issues.

```
SELECT * FROM kafka.LogEventStream WHERE (kafkaPartitionId = 1 AND kafkaMsgOffset > 1000 AND kafkaMsgOffset < 2000) OR (kafkaPartitionId = 2 AND kafkaMsgOffset > 4000)

```

And this way, we can use this predicates feature for external checkpointing mechanism.

And coming to time stamps, my point in case of invalid partitionId, query might block indefinitely with this feature. Where as in other case, we will return empty results.

It will be great if you can add few test cases with invalid partitions like below (Assuming partition 100 doesn't exist)

```
SELECT * FROM kafka.LogEventStream WHERE (kafkaPartitionId = 100 AND kafkaMsgTimestamp > 1527092007199)

```

aravi5:
Thanks @akumarb2010 for your comments. I have a few questions regarding your comments.

>In Kafka, offset scope itself is per partition. I am unable to find any use case, where we can take the range of offsets and apply on all partitions. In most of the scenario's they may not be valid offsets.
>IMHO, we should only apply predicate pushdown where we have exact scan specs.

1. Take this query for example
```
SELECT * FROM kafka.LogEventStream WHERE kafkaMsgOffset >= 1000 AND kafkaMsgOffset < 2000)
```

Did you mean that we do not apply predicate pushdown for such conditions? If we do not pushdown, all partitions will be scanned from `startOffset` to `endOffset`. Instead, by supporting pushdown for such queries, we are limiting the scan range wherever possible. Are there any drawbacks of applying pushdown on such conditions?

Also, pushdown implementation handles cases where specified offsets are invalid.

2. Can you elaborate more on this?
> we can use this predicate pushdown feature for external checkpointing mechanism.

3. 
>And coming to time stamps, my point is, in case of invalid partitionId, query might block indefinitely with this feature. Where as, without this feature, we will return empty results.

Even with pushdown we will return empty results. Pushdown is applied to each of the predicates independently and merged. The implementation is such that it ensures `offsetsForTimes` is called only for valid partitions (i.e., partitions returned by `partitionsFor`). Hence we will not run into the situation where `offsetsForTimes` blocks infinitely.

I will add test case for this situation. (I have already added cases for predicates with invalid offsets and timestamps.)

akumarb2010:
>> Did you mean that we do not apply predicate pushdown for such conditions?
Yes, IMO the users who is applying these predicates on offsets, should be aware of offset scope per partition. So, in such cases where offset predicates without partitionId can be ignored and let it be full scan as these are in valid queries in kafka perspective. Please let me know your thoughts.

>> If we do not pushdown, all partitions will be scanned from startOffset to endOffset.
Yes, but actually this kind queries itself are in valid except for topics with one partition. So better to not consider them.

>> Are there any drawbacks of applying pushdown on such conditions?
These are not drawbacks, but my point is let's not handle this invalid queries.

>> we can use this predicate pushdown feature for external checkpointing mechanism.
Mostly timestamp based predicates will be useful for the users. Only use case which i can see , for offset based predicates will be for checkpointing and all the check pointing strategies used in various systems like Storm/Spark streaming/Camus etc will be checkpointing on offset per partition. 


>> Even with pushdown we will return empty results. Pushdown is applied to each of the predicates independently and merged. The implementation is such that it ensures offsetsForTimes is called only for valid partitions (i.e., partitions returned by partitionsFor). Hence we will not run into the situation where offsetsForTimes blocks infinitely.
Thanks for clarifying it. Good to know that you are already handling these cases.

>> I will add test case for this situation. (I have already added cases for predicates with invalid offsets and timestamps.)
Thanks  @aravi5

aravi5:
>Yes, IMO the users who is applying these predicates on offsets, should be aware of offset scope per partition. So, in such cases where offset predicates without partitionId can be ignored and let it be full scan as these are in valid queries in kafka perspective. Please let me know your thoughts.

Thanks @akumarb2010 for clarifying. I agree with you that it is uncommon for user to specify conditions on `offsets` alone but DRILL will support such queries and will return messages that satisfy the condition (across all partitions). 

My opinion on this is that pushdown for such queries is OKAY and change to NOT pushdown may not be required for the following reasons -

1. Changes to NOT pushdown would require a LOT more effort and would add a lot more complexity to the code while providing the same result to the user.

Current implementation of pushdown is such that, when parsing the condition tree, a scan spec is generated for each predicate independently and then merged based on how different predicates are joined. 
For example `kafkaMsgOffset > 1000 AND kafkaPartitionId > 3` would create a scan spec (collection of `KafkaPartitionScanSpec`for each partition) for each of the predicates independently and scan spec is later merged based on `AND` condition.
With this implementation, creation of scan spec for `kafkaMsgOffset` does not have to know whether `kafkaPartitionId` is specified and the same applies to `kafkaPartitionId`. 

>These are not drawbacks, but my point is let's not handle this invalid queries.

2. DRILL supports such queries (though it may not make complete sense in Kafka world) and applying pushdown for such queries does not have any drawback.

3. Such queries may be common during Data Exploration phase. For example, user would want to see the first few messages produced to a topic (user may not know how the messages were distributed across partitions during produce).

For the sake of code simplicity and no user impact in terms of functionality (just better performance) I am leaning towards applying pushdown uniformly across partitions if `kafkaPartitionId` is not specified.

Let me know your thoughts.

kkhatua:
@akumarb2010  I haven't gone through @aravi5 's PR in detail, but I did see an interactive demo where we tried out quite a few of the scenarios. 
Semantically, it might not make much sense define offsets without specifying partitions, but it doesn't hurt to have the capability. We were trying to do  exploration of stream data during the demo, and offsets might prove to be a good way to just explore data in a topic on the assumption that the offsets for all the partitions are not wildly different for a similar time range (that is not granular).
Only if there is a benefit of adding to the complexity of the code, such as performance (or even debugging issues), it  _might_ be worth pursuing the additional changes.

aravi5:
@akumarb2010 - I checked in a commit to fix an issue where condition on non-existing partitions could result in a scan batch with no readers. In such a scenario I add `partition 0` to scan spec and point `startOffset` and `endOffset` to end of partition so no messages are actually read. I have also added test cases similar to the ones discussed earlier in the thread.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaNodeProcessor.java
```diff
@@ -0,0 +1,186 @@
...
+         */
+        if(!functionName.equals("equal") && !functionName.equals("greater_than")
+               && !functionName.equals("greater_than_or_equal_to")) {
+          return false;
+        }
+
+        if(valueArg instanceof IntExpression) {
+          value = (long) ((IntExpression) valueArg).getInt();
+          return true;
+        }
```
akumarb2010: I think we no need to handle IntExpression, as timestamp cannot be Int.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaPartitionScanSpec.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        if(scanSpec.endOffset > endOffset) {
+          endOffset = scanSpec.endOffset;
+        }
+
+        break;
+      default:
+        assert(false);
+    }
```
akumarb2010: Here we no need to fail for default. In other plugin's default, it will create empty ScanSpec.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaPartitionScanSpecBuilder.java
```diff
@@ -0,0 +1,346 @@
...
+          //Ensure scan range is between startOffset and endOffset,
+          long val = Math.max(fullScanSpec.get(tp).getStartOffset(),
+              Math.min(fieldValue, fullScanSpec.get(tp).getEndOffset()));
+
+          scanSpec.add(
+              new KafkaPartitionScanSpec(tp.topic(), tp.partition(),
+                  fullScanSpec.get(tp).getStartOffset(), val));
+        }
+        break;
+    }
```
akumarb2010: Can you please refactor the above code?

akumarb2010: @aravi5  This is a minor comment for readability. All the above cases can be extracted into one method by taking filedValue or fieldValue+1 as arguments.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaNodeProcessor.java
aravi5: This was mostly added for unit tests. It is possible (although unlikely) for Kafka topic be configured for `CreateTime` and one can specify timestamp as Int.

I have, however, moved the check for `IntExpression` to the end since it is least probable.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaPartitionScanSpec.java
aravi5: Yes. Thanks.

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaPartitionScanSpecBuilder.java
aravi5: Got it! I have extracted that to a small utility method.

priteshm:
@akumarb2010 is this ready to be committed?

aravi5:
@akumarb2010 - Thank you for reviewing the design and implementation, and providing valuable inputs.

arina-ielchiieva:
@aravi5 please squash all commits into one.

aravi5:
@arina-ielchiieva - Done.

PR #1273, opened 2018-05-18
DRILL-6426: Refactor TestFunctionsWithTypeExpoQueries test to be independent on limit 0 optimization option

arina-ielchiieva:


arina-ielchiieva:
@vvysotskyi please review.

vvysotskyi:
It is fine to make these tests independent of the default value of limit 0 optimization option.
LGTM, +1

PR #1274, opened 2018-05-18
DRILL-6424: Updating FasterXML Jackson libraries

vdiravka:
* all jackson libraries are updated to the latest version
* using of old jackson library by hive-exec-shaded for mapr profile is removed
* redundant Sequence class is removed


arina-ielchiieva:
@vdiravka why you need to remove sequence class, how it affected lib version upgrade?

vdiravka:
@arina-ielchiieva It is necessary to resolve conflicts there with a new Jackson API. But this class is deprecated and nowhere used.
I have tried to submit old logical plan with `sequence` and it was failed, since the LogicalPlan has newer structure to be deserialized by Jackson. Therefore no need to keep this class anymore.

arina-ielchiieva:
Looks good, thanks for the upgrade.

PR #1275, opened 2018-05-18
DRILL-6321: Lateral Join and Unnest - star column prefix

chunhui-shi:
* Add prefix to unnest's star column
  * add tests for output column verification


exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestLateralPlans.java
```diff
@@ -90,8 +90,50 @@ public void testLateralSqlPlainCol() throws Exception {
...
   @Test
   public void testLateralSqlStar() throws Exception {
-    String Sql = "select * from cp.`lateraljoin/nested-customer.json` t, unnest(t.orders) t2 limit 1";
-    test(Sql);
+    String Sql = "select * from cp.`lateraljoin/nested-customer.json` t, unnest(t.orders) t2 limit 0";
+
+    testBuilder()
+            .unOrdered()
+            .sqlQuery(Sql)
+            .baselineColumns("c_name", "c_id", "c_phone", "orders", "c_address", "o_id", "o_shop", "o_amount", "items")
```
amansinha100: I think this set of columns is not right.  The UNNEST should produce only the top level column since that's what the  planner knows about.  De-constructing the underlying map in the data means that we will not be able to specify a column alias for the unnest output:  e.g   UNNEST(t.orders) as t2(o).

chunhui-shi: The set of columns returned from UNNEST is another issue. And can be changed later if really needed. This fix is mainly about star case.

vdiravka:
[DRILL-6321](https://issues.apache.org/jira/browse/DRILL-6321) is marked as resolved with 1.14 Drill version. But it involves several PRs.
@chunhui-shi What is the purpose of that? Do we need reopen it or to open a new one for this PR?

PR #536, opened 2016-06-28
DRILL-4699: Add description column to sys.options table

sudheeshkatkam:
- Add new constructor for each validator in TypeValidators with
  description field; deprecate old constructors
- Add descriptions for some validators

PR #1279, opened 2018-05-20
DRILL-5735: UI options grouping and filtering & Metrics hints

kkhatua:
(Note: DRILL-4699 is also resolved in this)
Additional details, like the description is provided as well in a JScript lookup map. This helps reduce the need for the server to constantly recreate the entire page with the description details, as the client browser can fill in these details.
Developers will be expected to update the description as old/new options are introduced or deprecated.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
@@ -20,7 +20,12 @@ package org.apache.drill.exec.server.rest;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.servlets.MetricsServlet;
 import com.codahale.metrics.servlets.ThreadDumpServlet;
+import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
+import org.apache.drill.shaded.guava.com.google.common.io.Files;
+
+import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.RandomStringUtils;
+import org.apache.commons.lang3.StringEscapeUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.common.exceptions.DrillException;
@@ -29,6 +34,10 @@ import org.apache.drill.exec.ssl.SSLConfig;
 import org.apache.drill.exec.exception.DrillbitStartupException;
 import org.apache.drill.exec.server.BootStrapContext;
 import org.apache.drill.exec.server.Drillbit;
+import org.apache.drill.exec.server.options.OptionList;
+import org.apache.drill.exec.server.options.OptionManager;
+import org.apache.drill.exec.server.options.OptionValidator.OptionDescription;
+import org.apache.drill.exec.server.options.OptionValue;
 import org.apache.drill.exec.server.rest.auth.DrillErrorHandler;
 import org.apache.drill.exec.server.rest.auth.DrillHttpSecurityHandlerProvider;
 import org.apache.drill.exec.ssl.SSLConfigBuilder;
@@ -68,7 +77,11 @@ import javax.servlet.DispatcherType;
 import javax.servlet.http.HttpSession;
 import javax.servlet.http.HttpSessionEvent;
 import javax.servlet.http.HttpSessionListener;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
+import java.io.InputStream;
 import java.math.BigInteger;
 import java.net.BindException;
 import java.security.KeyPair;
@@ -76,14 +89,18 @@ import java.security.KeyPairGenerator;
 import java.security.KeyStore;
 import java.security.SecureRandom;
 import java.security.cert.X509Certificate;
+import java.util.Collections;
 import java.util.Date;
 import java.util.EnumSet;
+import java.util.List;
 
 /**
  * Wrapper class around jetty based webserver.
  */
 public class WebServer implements AutoCloseable {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(WebServer.class);
+  private static final String OPTIONS_DESCRIBE_JS = "options.describe.js";
+  private static final String OPTIONS_DESCRIBE_TEMPLATE_JS = "options.describe.template.js";
 
   private static final int PORT_HUNT_TRIES = 100;
   private static final String BASE_STATIC_PATH = "/rest/static/";
@@ -95,6 +112,22 @@ public class WebServer implements AutoCloseable {
   private final Drillbit drillbit;
 
   private Server embeddedJetty;
+
+  private File tmpJavaScriptDir;
+
+  public File getTmpJavaScriptDir() {
+    if (tmpJavaScriptDir == null) {
+      tmpJavaScriptDir = Files.createTempDir();
+      tmpJavaScriptDir.deleteOnExit();
+      //Perform All auto generated files at this point
+      try {
+        generateOptionsDescriptionJSFile();
+      } catch (IOException e) {
+        logger.error("Unable to create temp dir for JavaScripts. {}", e);
+      }
+    }
+    return tmpJavaScriptDir;
+  }
 
   /**
    * Create Jetty based web server.
@@ -194,6 +227,16 @@ public class WebServer implements AutoCloseable {
     staticHolder.setInitParameter("dirAllowed", "false");
     staticHolder.setInitParameter("pathInfoOnly", "true");
     servletContextHandler.addServlet(staticHolder, "/static/*");
+
+    //Add Local path resource (This will allow access to dynamically created files like JavaScript)
+    final ServletHolder dynamicHolder = new ServletHolder("dynamic", DefaultServlet.class);
+    //Skip if unable to get a temp directory (e.g. during Unit tests)
+    if (getTmpJavaScriptDir() != null) {
+      dynamicHolder.setInitParameter("resourceBase", getTmpJavaScriptDir().getAbsolutePath());
+      dynamicHolder.setInitParameter("dirAllowed", "true");
+      dynamicHolder.setInitParameter("pathInfoOnly", "true");
+      servletContextHandler.addServlet(dynamicHolder, "/dynamic/*");
+    }
 
     if (authEnabled) {
       //DrillSecurityHandler is used to support SPNEGO and FORM authentication together
@@ -408,6 +451,48 @@ public class WebServer implements AutoCloseable {
   public void close() throws Exception {
     if (embeddedJetty != null) {
       embeddedJetty.stop();
+    }
+    //Deleting temp directory
+    FileUtils.deleteDirectory(getTmpJavaScriptDir());
+  }
+
+  private static final String FILE_CONTENT_FOOTER = "};";
+
+  //Generate Options Description JavaScript
+  private void generateOptionsDescriptionJSFile() throws IOException {
+    //Obtain list of Options & their descriptions
+    OptionManager optionManager = this.drillbit.getContext().getOptionManager();
+    OptionList publicOptions = optionManager.getPublicOptionList();
+    List<OptionValue> options = Lists.newArrayList(publicOptions);
+    Collections.sort(options);
+    int numLeftToWrite = options.size();
+
+    //Template source Javascript file
+    InputStream optionsDescripTemplateStream = Resource.newClassPathResource(OPTIONS_DESCRIBE_TEMPLATE_JS).getInputStream();
+    //Generated file
+    File optionsDescriptionFile = new File(getTmpJavaScriptDir(), OPTIONS_DESCRIBE_JS);
+    optionsDescriptionFile.deleteOnExit();
+    //Create a copy of a template and write with that!
+    java.nio.file.Files.copy(optionsDescripTemplateStream, optionsDescriptionFile.toPath());
+    logger.info("Will write {} descriptions to {}", numLeftToWrite, optionsDescriptionFile.getAbsolutePath());
+
+    try (BufferedWriter writer = new BufferedWriter(new FileWriter(optionsDescriptionFile, true))) {
+      //Iterate through options
+      for (OptionValue option : options) {
+        numLeftToWrite--;
+        String optionName = option.getName();
+        OptionDescription optionDescription = optionManager.getOptionDefinition(optionName).getValidator().getOptionDescription();
+        if (optionDescription != null) {
+          //Note: We don't need to worry about short descriptions for WebUI, since they will never be explicitly accessed from the map
+          writer.append("  \"").append(optionName).append("\" : \"")
+          .append(StringEscapeUtils.escapeEcmaScript(optionDescription.getDescription()))
+          .append( numLeftToWrite > 0 ? "\"," : "\"");
+          writer.newLine();
+        }
+      }
+      writer.append(FILE_CONTENT_FOOTER);
+      writer.newLine();
+      writer.flush();
     }
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
@@ -27,6 +27,8 @@ import java.util.Map;
 import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.server.options.OptionManager;
 import org.apache.drill.exec.server.options.OptionValue;
+import org.apache.drill.exec.server.options.OptionValidator.OptionDescription;
+import org.apache.drill.exec.server.options.OptionValue.AccessibleScopes;
 import org.apache.drill.exec.server.options.OptionValue.Kind;
 import org.apache.drill.exec.server.options.OptionValue.OptionScope;
 import org.apache.drill.exec.store.pojo.NonNullable;
@@ -53,14 +55,26 @@ import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
  *  only the value set at SESSION level.
  */
 public class ExtendedOptionIterator implements Iterator<Object> {
-//  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(OptionIterator.class);
+  //private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ExtendedOptionIterator.class);
 
   private final OptionManager fragmentOptions;
   private final Iterator<OptionValue> mergedOptions;
+  private Map<OptionValue.Kind, String> typeMapping;
+  private Map<OptionScope, Integer> preference;
+  private static final int SHORT_DESCRIP_MAX_SIZE = 110;
 
   public ExtendedOptionIterator(FragmentContext context, boolean internal) {
     fragmentOptions = context.getOptions();
-    final Iterator<OptionValue> optionList;
+    preference = new HashMap<OptionScope, Integer>();
+    preference.put(OptionScope.SESSION, 0);
+    preference.put(OptionScope.SYSTEM, 1);
+    preference.put(OptionScope.BOOT, 2);
+
+    typeMapping = new HashMap<Kind, String>();
+    typeMapping.put(Kind.STRING, "VARCHAR");
+    typeMapping.put(Kind.DOUBLE, "FLOAT");
+    typeMapping.put(Kind.LONG, "BIGINT");
+    typeMapping.put(Kind.BOOLEAN, "BIT");
 
     if (!internal) {
       mergedOptions = sortOptions(fragmentOptions.getPublicOptionList().iterator());
@@ -76,11 +90,6 @@ public class ExtendedOptionIterator implements Iterator<Object> {
   public Iterator<OptionValue> sortOptions(Iterator<OptionValue> options) {
     List<OptionValue> optionslist = Lists.newArrayList(options);
     HashMap<String, OptionValue> optionsmap = new HashMap<>();
-    final Map<OptionScope, Integer> preference = new HashMap<OptionScope, Integer>() {{
-      put(OptionScope.SESSION, 0);
-      put(OptionScope.SYSTEM, 1);
-      put(OptionScope.BOOT, 2);
-    }};
 
     for (OptionValue option : optionslist) {
       if (optionsmap.containsKey(option.getName())) {
@@ -116,14 +125,17 @@ public class ExtendedOptionIterator implements Iterator<Object> {
   @Override
   public ExtendedOptionValueWrapper next() {
     final OptionValue value = mergedOptions.next();
-    final HashMap<OptionValue.Kind,String> typeMapping = new HashMap() {{
-      put(Kind.STRING,"VARCHAR");
-      put(Kind.DOUBLE,"FLOAT");
-      put(Kind.LONG,"BIGINT");
-      put(Kind.BOOLEAN,"BIT");
 
-    }};
-    return new ExtendedOptionValueWrapper(value.name, typeMapping.get(value.kind), value.accessibleScopes,value.getValue().toString(), value.scope);
+    final Status status;
+    if (value.accessibleScopes == AccessibleScopes.BOOT) {
+      status = Status.BOOT;
+    } else {
+      final OptionValue def = fragmentOptions.getDefault(value.name);
+      status = (value.equalsIgnoreType(def) ? Status.DEFAULT : Status.CHANGED);
+    }
+
+    return new ExtendedOptionValueWrapper(value.name, typeMapping.get(value.kind), value.accessibleScopes,value.getValue().toString(), status, value.scope,
+        getShortDescription(value.name));
   }
 
   public enum Status {
@@ -134,7 +146,6 @@ public class ExtendedOptionIterator implements Iterator<Object> {
    * Wrapper class for Extended Option Value
    */
   public static class ExtendedOptionValueWrapper {
-
     @NonNullable
     public final String name;
     @NonNullable
@@ -142,17 +153,39 @@ public class ExtendedOptionIterator implements Iterator<Object> {
     @NonNullable
     public final OptionValue.AccessibleScopes accessibleScopes;
     public final String val;
+    public final Status status;
     @NonNullable
     public final OptionScope optionScope;
+    public final String description;
 
-
-    public ExtendedOptionValueWrapper(final String name, final String kind, final OptionValue.AccessibleScopes type, final String value, final OptionScope scope) {
+    public ExtendedOptionValueWrapper(final String name, final String kind, final OptionValue.AccessibleScopes type, final String value, final Status status, final OptionScope scope,
+        final String description) {
       this.name = name;
       this.kind = kind;
       this.accessibleScopes = type;
       this.val = value;
+      this.status = status;
       this.optionScope = scope;
+      this.description = description;
     }
+  }
+
+  //Extract a limited length from the original description if not available
+  private String getShortDescription(String name) {
+    OptionDescription optionDescription = fragmentOptions.getOptionDefinition(name).getValidator().getOptionDescription();
+    if (optionDescription == null) {
+      return "";
+    }
+    String description = null;
+    if (optionDescription.hasShortDescription()) {
+      description = optionDescription.getShortDescription();
+    } else {
+      description = optionDescription.getDescription();
+      if (description.length() > SHORT_DESCRIP_MAX_SIZE) {
+        return description.substring(0, SHORT_DESCRIP_MAX_SIZE-3).concat("..."); //Append ellipsis (trailing dots)
+      }
+    }
+    return description;
   }
 
   @Override
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
@@ -140,6 +140,4 @@ public enum SystemTable {
   public Class<?> getPojoClass() {
     return pojoClass;
   }
-
-
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -21,6 +21,7 @@ import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.exec.physical.impl.common.HashTable;
 import org.apache.drill.exec.rpc.user.InboundImpersonationManager;
 import org.apache.drill.exec.server.options.OptionValidator;
+import org.apache.drill.exec.server.options.OptionValidator.OptionDescription;
 import org.apache.drill.exec.server.options.TypeValidators.IntegerValidator;
 import org.apache.drill.exec.server.options.TypeValidators.BooleanValidator;
 import org.apache.drill.exec.server.options.TypeValidators.DoubleValidator;
@@ -63,7 +64,7 @@ public final class ExecConstants {
   public static final String BIT_SERVER_RPC_THREADS = "drill.exec.rpc.bit.server.threads";
   public static final String USER_SERVER_RPC_THREADS = "drill.exec.rpc.user.server.threads";
   public static final String FRAG_RUNNER_RPC_TIMEOUT = "drill.exec.rpc.fragrunner.timeout";
-  public static final PositiveLongValidator FRAG_RUNNER_RPC_TIMEOUT_VALIDATOR = new PositiveLongValidator(FRAG_RUNNER_RPC_TIMEOUT, Long.MAX_VALUE);
+  public static final PositiveLongValidator FRAG_RUNNER_RPC_TIMEOUT_VALIDATOR = new PositiveLongValidator(FRAG_RUNNER_RPC_TIMEOUT, Long.MAX_VALUE, null);
   public static final String TRACE_DUMP_DIRECTORY = "drill.exec.trace.directory";
   public static final String TRACE_DUMP_FILESYSTEM = "drill.exec.trace.filesystem";
   public static final String TEMP_DIRECTORIES = "drill.exec.tmp.directories";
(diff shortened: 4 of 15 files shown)


kkhatua:
Sample snapshot when filtering with one of the quick filters being clicked:
image

kkhatua:
@dvjyothsna  / @sohami / @parthchandra can one of you please review this?

dvjyothsna:
This is really handy. One more thing If I'm not wrong, I think John Omernick was asking for description for options but I think he wanted it as a another table or something. But I think this will help him.

kkhatua:
Oh.. good! I am running out of code review volunteers. 
@johnomernik can you review this PR?

kkhatua:
@JohnOmernik I agree. Single-source of truth for descriptions will help a long way. 
A similar ask for `sys.functions` is something that @arina-ielchiieva  had pointed to when we introduced syntax highlighting. That, of course, is more trickier due to dynamic UDFs. 
We can try to address that separately as a combined commit. For now, does this look sufficient within the context of the JIRA ?

JohnOmernik:
Yes, I like what I am seeing and in the context of the JIRA it looks good.

JohnOmernik:
+1

dvjyothsna:
Very good enhancement Kunal. LGTM +1

JohnOmernik:
So on the idea of UDFs, I don't think UDFs that are NOT shipped with the
product should be a concern of ours.  Any UDF that ships with Drill needs
to be in the sys.functions.

Perhaps it would be cool to have a way to register UDFs in sys.function so
that folks who have UDFs that want to use  sys.functions as a documentation
store can.  Basically, a way in the UDF to add "company only" UDFs to the
sys.functions?

John

arina-ielchiieva:
@kkhatua though the idea of the improvement is nice but having options description in js file is not a good, looks like a hack. :) We should find a way to add description in drill-module.conf file or other approach you can think of. Thus description will be loaded along with the options defaults at start up and thus it would be easily added to the options page, as well as in sys.options table as part of description column,

arina-ielchiieva:
@kkhatua, we don't have that many options to have significant load on web server. I understand you want to make options page user-friendly but maintaining options descriptions separately from options is not a good idea. Developers definitely won't remember to update js file when adding new options. So at some point we'll end up with stale descriptions. I am sorry but I would rather have less-user friendly options page then implemented this way. Until we don't have better solution, you can add link to the Drill documentation where options are described on options page.

JohnOmernik:
What if the options were loaded into the JS form on first connection to the
web UI and cached? These settings don't change much, so setting a longer
cache timeout and refreshing cache on change would not be a resource drain.
This could be a file (maybe even JS like it is now) but not part of the
code base, it's generated from the system of record.

parthchandra:
There is a PR for DRILL-4699 (adding option descriptions) which, sadly, is a bit out of date. An advantage of using that PR as the base for the UI would be that the same descriptions would also be available in sqlline.

kkhatua:
I would like to have the work for descriptions be **delinked** from this issue, which is a usability experience. 
The issue pointed out is related to maintaining of these (and other) descriptions in a coherent state across the UX, the backend code and and documentation. We already have JIRAs - DRILL-4699 & DRILL-3988 .

parthchandra:
IMHO, code maintainability and coherence are perfectly valid requirements from a reviewer. In particular, having help text being out of date or inconsistent is a fatal problem for user facing code. If you're inclined, we can work on updating DRILL-4699 together.

ilooner:
Since it looks like reviewers have requested additional work to be done on this change, I have removed the ready-to-commit label from the Jira.

arina-ielchiieva:
Looks like Kunal wants to separate out sorting & search from descriptions work. I think this is acceptable.

kkhatua:
Looks like only @arina-ielchiieva is in agreement with me on having this done incrementally :)

@parthchandra , if updating DRILL-4699 (#536) is not too complex and time-consuming, we can follow that route. I can create a dynamic one-time JScript to be generated during the web server startup, and serve the file from there.

parthchandra:
Sounds good to me.

kkhatua:
@parthchandra 
Rebased and updated with a resource file as the primary source of truth. The Webserver will now create a JavaScript on the fly and serve that to browsers which can use it to render the information.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -408,5 +448,47 @@ public void close() throws Exception {
...
+      "  }\n" +
+      "  var optsDescripMap = {";
+  final String FILE_CONTENT_FOOTER =
+      "};\n";
+
+  //Generate Options Description JavaScript
+  private void generateOptionsDescriptionJSFile() throws IOException {
+    //Obtain properties from Resource file
+    Properties optionsDescriptions = new Properties();
+    InputStream resrcInputStream = this.getClass().getClassLoader().getResourceAsStream(OPTIONS_DESCRIBE_PROPERTIES);
```
parthchandra: This is nice, but using a ResourceBundle would be even better since that will allow the descriptions to be localized without code change.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -96,6 +107,23 @@
...
+  public File getTmpJavaScriptDir() {
+    if (tmpJavaScriptDir == null) {
+      tmpJavaScriptDir = Files.createTempDir();
+      tmpJavaScriptDir.deleteOnExit();
+      //Perform All auto generated files at this point
+      try {
+        generateOptionsDescriptionJSFile();
+      } catch (IOException e) {
+        // TODO Auto-generated catch block
+        e.printStackTrace();
```
arina-ielchiieva: Please remove this and throw appropriate exception.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -408,5 +448,47 @@ public void close() throws Exception {
...
+    //Obtain properties from Resource file
+    Properties optionsDescriptions = new Properties();
+    InputStream resrcInputStream = this.getClass().getClassLoader().getResourceAsStream(OPTIONS_DESCRIBE_PROPERTIES);
+    optionsDescriptions.load(resrcInputStream);
+    int numLeftToWrite = optionsDescriptions.size();
+
+    File optionsDescriptionFile = new File(getTmpJavaScriptDir(), OPTIONS_DESCRIBE_JS);
+    optionsDescriptionFile.deleteOnExit();
+    logger.info("Will write {} descriptions to {}", numLeftToWrite, optionsDescriptionFile.getAbsolutePath());
+    BufferedWriter writer = new BufferedWriter(new FileWriter(optionsDescriptionFile));
```
arina-ielchiieva: Is try with resources possible here?

exec/java-exec/src/main/resources/options.describe.properties
```diff
@@ -0,0 +1,106 @@
...
+store.hive.optimize_scan_with_native_readers=Optimize reads of Parquet-backed external tables from Hive by using Drill native readers instead of the Hive Serde interface. (Drill 1.2 and later)
+store.json.all_text_mode=Drill reads all data from the JSON files as VARCHAR. Prevents schema change errors.
+store.json.extended_types=Turns on special JSON structures that Drill serializes for storing more type information than the four basic JSON types.
+store.json.read_numbers_as_double=Reads numbers with or without a decimal point as DOUBLE. Prevents schema change errors.
+store.mongo.all_text_mode=Similar to store.json.all_text_mode for MongoDB.
+store.mongo.read_numbers_as_double=Similar to store.json.read_numbers_as_double.
+store.parquet.block-size=Sets the size of a Parquet row group to the number of bytes less than or equal to the block size of MFS, HDFS, or the file system.
+store.parquet.compression=Compression type for storing Parquet output. Allowed values: snappy, gzip, none
+store.parquet.enable_dictionary_encoding=For internal use. Do not change.
+store.parquet.dictionary.page-size=
```
arina-ielchiieva: description?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -408,5 +448,47 @@ public void close() throws Exception {
     if (embeddedJetty != null) {
       embeddedJetty.stop();
     }
+    //Deleting temp directory
+    FileUtils.deleteDirectory(getTmpJavaScriptDir());
```
arina-ielchiieva: Will this fail if we unable to delete temp dir?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -96,6 +107,23 @@
 
   private Server embeddedJetty;
 
+  private static File tmpJavaScriptDir;
+
+  public File getTmpJavaScriptDir() {
+    if (tmpJavaScriptDir == null) {
```
arina-ielchiieva: How many times this would be called? I am just wondering why we have this null check.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -96,6 +107,23 @@
 
   private Server embeddedJetty;
 
+  private static File tmpJavaScriptDir;
```
arina-ielchiieva: Can we store it as local variable?

kkhatua: Being a single instance, I thought a static would be preferable.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: Ok.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: I wanted to bundle the creation as a part of the dir creation, so this check is for that.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: My guess is that deleting a non-empty directory without this would not work. This should recursively go and clean up.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: SHould be. I'll incorporate that change.

exec/java-exec/src/main/resources/options.describe.properties
kkhatua: Picked this up from the website, so I guess this slipped through. :)

kkhatua:
@arina-ielchiieva I should be able to make most of the changes you've asked for.
We'll need to inform devs about this as well. 
However, i have a second problem related to DRILL-4699, which can be easily built off this. The descriptions are very lengthy and break the console formatting when applying to the `sys.options` table. To make things more confusing, there is now a `sys.options_val` table as well ! I'm in favor of replacing the `sys.options` with the `sys.options_val` implementation. So, I'm going to slightly revise this PR in anticipation of an additional PR for DRILL-4699, where the same/similar resource can be used for displaying a shorter description. (limiting to <110 chars doesn't seem to break formatting)

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: Let's skip localization for now :)
I'll work towards setting up the PR for DRILL-4699's PR as well, which would be a minor patch over this.

kkhatua:
@arina-ielchiieva  / @parthchandra 

Modified the PR with additional work:
1. The latest commit now carries an implementation for description in `sys.options_val` table (DRILL-4699), making PR #536 obsolete
2. The descriptions-carrying resource file is loaded by the `ExtendedOptionIterator` and made accessible at a class level to the `WebServer` instance
4. The full description is to be rendered in the WebUI (i.e. http://drillbit:8047/options )
5. The short description (if available) is shown in the System table, and truncated with suffixing of ellipsis to indicate more content. (See bottom). This is required so as to ensure that the formatting within SqlLine is maintained.
```
0: jdbc:drill:drillbit=kk127.qa.lab> select * from sys.options_val where name like 'planner.p%' or name like 'planner.width.%' or name like '%cpu%' order by name;
+--------------------------------------------+----------+-------------------+-------+----------+--------------+-----------------------------------------------------------------------------------------------------------------+
|                    name                    |   kind   | accessibleScopes  |  val  |  status  | optionScope  |                                                   description                                                   |
+--------------------------------------------+----------+-------------------+-------+----------+--------------+-----------------------------------------------------------------------------------------------------------------+
| planner.cpu_load_average                   | FLOAT    | ALL               | 0.7   | DEFAULT  | BOOT         |                                                                                                                 |
| planner.parser.quoting_identifiers         | VARCHAR  | ALL               | `     | DEFAULT  | BOOT         |                                                                                                                 |
| planner.partitioner_sender_max_threads     | BIGINT   | ALL               | 8     | DEFAULT  | BOOT         | Upper limit of threads for outbound queuing.                                                                    |
| planner.partitioner_sender_set_threads     | BIGINT   | ALL               | -1    | DEFAULT  | BOOT         | Overwrites the number of threads used to send out batches of records. Set to -1 to disable. Typically not c...  |
| planner.partitioner_sender_threads_factor  | BIGINT   | ALL               | 2     | DEFAULT  | BOOT         | A heuristic param to use to influence final number of threads. The higher the value the fewer the number of...  |
| planner.producer_consumer_queue_size       | BIGINT   | ALL               | 10    | DEFAULT  | BOOT         | How much data to prefetch from disk in record batches out-of-band of query execution. The larger the queue ...  |
| planner.width.max_per_node                 | BIGINT   | ALL               | 0     | DEFAULT  | BOOT         | Max number of threads that can run in parallel for a query on a node.                                           |
| planner.width.max_per_query                | BIGINT   | ALL               | 1000  | DEFAULT  | BOOT         | Same as max per node but applies to the query as executed by the entire cluster. For example, this value mi...  |
+--------------------------------------------+----------+-------------------+-------+----------+--------------+-----------------------------------------------------------------------------------------------------------------+
8 rows selected (0.797 seconds)
```
6. Review comments have been addressed:
  a. JavaScript is written using try-resources
  b. tmpJavaScriptDir is a class instance member
  c. Missing description has been added. For testing purposes, a `planner.width.max_per_node.short` entry also exists (as shown above)

Going forward, **ALL** developers need to introduce the description of their introduced options.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -52,14 +55,44 @@
...
  */
 public class ExtendedOptionIterator implements Iterator<Object> {
-//  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(OptionIterator.class);
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ExtendedOptionIterator.class);
+  private static final String OPTIONS_DESCRIBE_PROPERTIES = "options.describe.properties";
+  private static final Properties descriptionMap = new Properties();
 
   private final OptionManager fragmentOptions;
   private final Iterator<OptionValue> mergedOptions;
+  private HashMap<OptionValue.Kind,String> typeMapping;
```
arina-ielchiieva: Can we use more generic `Map`, plus space after comma.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -52,14 +55,44 @@
...
+  private Map<OptionScope, Integer> preference;
+
+  //Load & Provide a descriptions map
+  public static Properties getDescriptionMap() {
+      if (descriptionMap.isEmpty()) { //Expect a one time load
+        synchronized (descriptionMap) {
```
arina-ielchiieva: Re-check for empty after sync (DCL).

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -52,14 +55,44 @@
...
+          try (InputStream optionDescInputStream = OptionIterator.class.getClassLoader().getResourceAsStream(OPTIONS_DESCRIBE_PROPERTIES)) {
+            descriptionMap.load(optionDescInputStream);
+          } catch (IOException e) {
+            logger.warn("Unable to load descriptions from {}", OPTIONS_DESCRIBE_PROPERTIES);
```
arina-ielchiieva: Add exception to logger.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -52,14 +55,44 @@
...
+        }
+      }
+      return descriptionMap;
+  }
 
+  @SuppressWarnings("serial")
   public ExtendedOptionIterator(FragmentContext context, boolean internal) {
     fragmentOptions = context.getOptions();
-    final Iterator<OptionValue> optionList;
+    preference = new HashMap<OptionScope, Integer>() {{
```
arina-ielchiieva: You are actually extending the HashMap, not sure this is preferable. Please use standard way of adding values.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -115,14 +143,20 @@ public boolean hasNext() {
...
-      put(Kind.BOOLEAN,"BIT");
 
-    }};
-    return new ExtendedOptionValueWrapper(value.name, typeMapping.get(value.kind), value.accessibleScopes,value.getValue().toString(), value.scope);
+    final Status status;
+    if (value.accessibleScopes == AccessibleScopes.BOOT) {
+      status = Status.BOOT;
+    } else {
+      final OptionValue def = fragmentOptions.getDefault(value.name);
+      if (value.equalsIgnoreType(def)) {
```
arina-ielchiieva: Use ternary operator.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -134,19 +168,36 @@ public ExtendedOptionValueWrapper next() {
...
       this.name = name;
       this.kind = kind;
       this.accessibleScopes = type;
       this.val = value;
+      this.status = status;
       this.optionScope = scope;
+      this.description = getShortDescription(name);
+    }
+
+    private String getShortDescription(String name) {
```
arina-ielchiieva: Please add comment why we might need short description.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
```diff
@@ -31,6 +31,7 @@
  * </p>
  */
 public enum SystemTable {
+  @Deprecated
```
arina-ielchiieva: Why we need to deprecate?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
arina-ielchiieva: I meant what if we fail to delete temp directory? May be use delete quietly.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -194,6 +220,16 @@ private ServletContextHandler createServletContextHandler(final boolean authEnab
     staticHolder.setInitParameter("pathInfoOnly", "true");
     servletContextHandler.addServlet(staticHolder, "/static/*");
 
+    //Add Local path resource (This will allow access to dynamically created files like JavaScript)
+    final ServletHolder dynamicHolder = new ServletHolder("dynamic", DefaultServlet.class);
+    //Skip if unable to get a temp directory (e.g. during Unit tests)
+    if (getTmpJavaScriptDir() != null) {
```
arina-ielchiieva: Server start won't be called several times. I guess just calling getTmpJavaScriptDir is sufficient and hen refer to tmpJavaScriptDir. And do null check during shutdown.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
arina-ielchiieva: Again defending `getTmpJavaScriptDir()` usage is incorrect. Consider the case when server start will fail before generating js temp dir, so in close method you'll call `getTmpJavaScriptDir()` and it will generate js options since it did not do it at start up since all we need is to delete the directory.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -408,5 +444,46 @@ public void close() throws Exception {
     if (embeddedJetty != null) {
       embeddedJetty.stop();
     }
+    //Deleting temp directory
+    FileUtils.deleteDirectory(getTmpJavaScriptDir());
+  }
+
+  /* Static wrapper text for JavaScript content */
+  private static final String FILE_CONTENT_HEADER =
```
arina-ielchiieva: Consider having some template where you'll insert only options and descriptions rather then concatenating js file from string pieces.

parthchandra:
Why not add the description in sys.options? It is the more commonly used table, and also the logical place to look for options and their descriptions. (And yes, I know that it provides less space for the description; but you just need to wrap around and live with the slightly ugly look). 
I think we also need to merge the sys.options and sys.options_val tables (there is no need to have both), but that is a topic for a different Jira/PR.

kkhatua:
@parthchandra 
I actually have my original PR with that. But I seemed to be breaking some backward compatibility tests. Not sure why a sys.options table needs backward compatibility, with the exception of options persisted in ZK, but that is the case.  That's why ( @arina-ielchiieva  ), for now, I marked the `sys.options` as deprecated.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: Ok, just moved the method out from where it originally was. Will make the change

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: My bad. I think I missed adding it in the refactor.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: Ok.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: I wanted to generate the JS lazily, hence this approach. I'll make the necessary change and generate it on startup.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: You mean a freemarker template? That would involve the webserver regenerating the file on every request

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: 👍 Ok. Was avoiding refactoring the original code, just I copied as is from OptionsIterator. :)

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
kkhatua: I want to actually replace the sys.options completely with sys.options_val , but I broke tests. Hoping to discourage the usage this way.

parthchandra: It should really be the other way around; i.e deprecate sys.options_val and retain sys.options. Options_val coalesces the xxx_value columns to show only one value column which is desirable. But it also shows only the value in effect for the user which IMO is less useful. 
@amansinha100 had a good suggestion which was to have one large sys.options table but provide appropriate views for different use cases.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: 👍

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
kkhatua: I don't think the challenge is so much in deprecating either. I actually have a commit that renames the two tables as `sys.options_old` and s`ys.options`. The concern seems to be around breaking backward compatibility tests.
Are you suggesting we show default values in addition to changed value??

parthchandra: I see inconsistent behaviour for sys.options. If I override an option at the session level, two rows show up in the table, one for the default and one for the overridden (sys.options_val always shows only one). If, however, I override an option at the system level only one row shows up in the table, the overridden one. 
Ideal would be to show both, but also coalesce the columns for the value (as in options_val).
I think we should log a separate Jira for that.

kkhatua: Interesting. I've never seen that whenever I've modified at a session level. Could you file a JIRA for that?

kkhatua:
@arina-ielchiieva I'm going to split this PR into 2 parts, with the UI part going in first. The changes to the `sys.options` table will follow, after I figure out a fix to Parth's reported bug.  Would that be sufficient for a +1 ?

Agirish:
This is really cool! Love the UI. Not sure what's with the new options_val table - is this going to replace sys.options? Didn't get the reasoning behind that.

kkhatua:
Ok. Just checked. The bug reported by Parth has been around even before my commit. As such, I don't see the value in splitting the commit then. 
@arina-ielchiieva , I'll try to address the comments in your review.

@Agirish the `options_val` table was introduced to allow a unified value column instead of different column values for for each possible type (e.g. bool_val , string_val , ...) most of which would be null.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: This is just refactored code that I moved from elsewhere.  Since the entire statement is creating a constant map, this encapsulates that. Should I still make the change?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
kkhatua: The bug is not from this commit, so will ignore this for now. 
Should I leave this deprecated or remove it?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: @arina-ielchiieva  I'm unable to reply to your comment on line 452, so doing it here. 
```
arina-ielchiieva on Jun 15  Contributor
I meant what if we fail to delete temp directory? May be use delete quietly.
```
The temp directory is already marked as `deleteOnExit()` , so the JVM should also do that during shutdown. However, it didn't seem to work always, so an explicit delete is fine.

```
arina-ielchiieva on Jun 15  Contributor
Again defending getTmpJavaScriptDir() usage is incorrect. Consider the case when server start will fail before generating js temp dir, so in close method you'll call getTmpJavaScriptDir() and it will generate js options since it did not do it at start up since all we need is to delete the directory.
```
The creation of the JS options is a negligible cost, as long as we are cleaning up correctly.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: I could use a standalone template file (say, `options.describe.js.template`) and then construct the the final file with a copy from this and inject the options and descriptions. I think that makes it cleaner and helps keep JavaScript code outside of the Java code.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
arina-ielchiieva: What is the decision on this part?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
arina-ielchiieva: Has comment been added?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
arina-ielchiieva: I would prefer if you would :)

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
```diff
@@ -116,14 +146,16 @@ public boolean hasNext() {
...
-      put(Kind.BOOLEAN,"BIT");
 
-    }};
-    return new ExtendedOptionValueWrapper(value.name, typeMapping.get(value.kind), value.accessibleScopes,value.getValue().toString(), value.scope);
+    final Status status;
+    if (value.accessibleScopes == AccessibleScopes.BOOT) {
+      status = Status.BOOT;
+    } else {
+      final OptionValue def = fragmentOptions.getDefault(value.name);
+      status = (value.equalsIgnoreType(def) ? Status.DEFAULT : Status.CHANGED );
```
arina-ielchiieva: Please remove space before `)`.

exec/java-exec/src/main/resources/options.describe.properties
```diff
@@ -0,0 +1,107 @@
+#
```
arina-ielchiieva: Since we have now options.describe.properties file, how did we agree to enforce developers to add descriptions?
Will project build fail if description was not added? Or maybe instead of using properties file, we can make `OptionValidator` to accept description along with option name, thus all options will be required to have description? And we won't have to duplicate options names in options.describe.properties?

kkhatua: For future additions, we could do this, but I didn't want to right away change the existing OptionValidator usages. I'll need to work with @bbevens to fill up the properties file for now, since total list of options are 150+.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
kkhatua: I've removed the deprecated flag for now. I logged [DRILL-6684](https://issues.apache.org/jira/browse/DRILL-6684)  for this purpose.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: Looks like I missed this. Done in the latest commit.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/ExtendedOptionIterator.java
kkhatua: Done

exec/java-exec/src/main/resources/options.describe.properties
parthchandra: We can log a new JIRA for updating the Options validator so that all options have a description and move on.

arina-ielchiieva: Sorry guys but I think it's better not to introduce `options.describe.properties` file at all. Updating Options validator is not that hard, for those options we don't have a description prepared, just leave blanks.

kkhatua: @arina-ielchiieva the purpose of the PR was a UX enhancement. The backend is a temporary change until the proposed change to the OptionValidator is implemented later. Time to change existing OptionValidators would require significant time. Introducing the complete change now, by leaving blanks defeats the purpose of the UX enhancement.

arina-ielchiieva: @kkhatua usually there is nothing more persistent then temporary :)
Current list of options in properties files does not contain full list of system / session options. It means that the rest would have empty description, right? So I am not sure how this is better then updating OptionsValidators. Expanding the constructor and adding descriptions to options, does not seem to take a significant time, might be not the most interesting part of the development but still.

kkhatua: Yes, some options have empty descriptions because I didn't find them on the website. I'm looking to get in the UX enhancement originally intended for 1.14.0. I also wanted to (separately) discuss whether changes to the contructor is suitable or introduction of annotations.

arina-ielchiieva: I would go with constructor.

kkhatua:
@arina-ielchiieva please review.

arina-ielchiieva:
@kkhatua two questions:
1. Will `sys.options` return descriptions along previous information (this can be useful)?
2. Do we need `options.describe.properties` ? We can get all descriptions from option manager now.

kkhatua:
@arina-ielchiieva 
1. The descriptions are, as of now, part of `sys.options_val` (and sys.internal_options_val`) due to the width of `sys.options` . I intend to use [DRILL-6684](https://issues.apache.org/jira/browse/DRILL-6684) to swap the names of the two system tables. That is a bigger exercise because unit tests, etc need to be adjusted accordingly.
2. I've dropped it from the last commit. 

I'll work with Bridget and other developers to update their features' property descriptions. As of now, we've barely got half of these covered.

arina-ielchiieva:
@kkhatua please squash the commits.

kkhatua:
Done. Rebased and squashed. Cleared unit and functional tests.

arina-ielchiieva:
@kkhatua please resolve the conflicts.

kkhatua:
@arina-ielchiieva  Done.

kkhatua:
Rebased again due to Guava shading.

sohami:
@kkhatua - please make PR and JIRA title same.

PR #1280, opened 2018-05-21
DRILL-5990: Drill should not startup with JRE sourced java executable

kkhatua:
Currently, the check for the Java compiler is made only during query execution.
This commit ensures that if the java executable used for instantiating the Drillbit is from a JRE (i.e. the Java compiler is not available), the Drillbit will fail with the exception reported in the logs:
```
Exception in thread "main" org.apache.drill.exec.exception.DrillbitStartupException: JDK Java compiler not available. Ensure Drill is running with the java executable from a JDK and not a JRE
        at org.apache.drill.exec.server.Drillbit.<init>(Drillbit.java:138)
        at org.apache.drill.exec.server.Drillbit.start(Drillbit.java:409)
        at org.apache.drill.exec.server.Drillbit.start(Drillbit.java:389)
        at org.apache.drill.exec.server.Drillbit.main(Drillbit.java:385)
```

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -20,6 +20,8 @@ package org.apache.drill.exec.server;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import javax.tools.ToolProvider;
+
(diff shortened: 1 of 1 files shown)


kkhatua:
@parthchandra  / @Ben-Zvi  can you quickly review this? 
I've tested this by switching the `JAVA_HOME` env to point to a JRE installation; and also by unsetting the same variable, while having the default `java` executable point to `$JRE_HOME/bin/java exec`

PR #1284, opened 2018-05-23
DRILL-6438: Remove excess logging form tests.

ilooner:
## Changes

 - Removed excess print statements from tests
 - Switched all System.out and System.err statements in the tests to loggers

### DIFF
--- a/common/src/test/java/org/apache/drill/test/Drill2130CommonHamcrestConfigurationTest.java
+++ b/common/src/test/java/org/apache/drill/test/Drill2130CommonHamcrestConfigurationTest.java
@@ -24,6 +24,7 @@ import static org.junit.Assert.fail;
 import static org.hamcrest.CoreMatchers.equalTo;
 
 public class Drill2130CommonHamcrestConfigurationTest {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Drill2130CommonHamcrestConfigurationTest.class);
 
   @SuppressWarnings("unused")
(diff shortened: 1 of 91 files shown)


ilooner:
@arina-ielchiieva

priteshm:
@vvysotskyi could you please review this?

vvysotskyi:
@ilooner, could you please remove other places in tests where results are printed using `System.out.println()`. For example, in TestHiveUDFs it also may be removed; for every HBase test its result is printed for now.

Also, may be changed value of the option `drill.test.query.printing.silent` to reduce output.

ilooner:
Sure I'll try to remove them from the other tests as well. Some of the extra logging I might not be able to remove. For example a lot of the MongoDB logging comes from separate MongoDB server processes the test launches with the flapadoodle library.

arina-ielchiieva:
@ilooner what about banning `System.out` and `e.printStackTrace()`? Will this be addressed in different PR? If yes, could you please point out to the Jira?

vvysotskyi:
Thanks for so many changes!
The output for some tests (`TestSimpleProjection`, `TestConvertFunctions` etc.) is printed using methods from `VectorUtil`. It would be fine to change the output for these methods.

ilooner:
@arina-ielchiieva I made this Jira to track banning those functions. https://issues.apache.org/jira/browse/DRILL-6464

@vvysotskyi I converted the print statements to log statements in VectorUtil

arina-ielchiieva:
@ilooner thanks for creating the Jira and making the changes, looks good to me.

PR #1286, opened 2018-05-23
DRILL-6435: MappingSet is stateful, so it can't be shared between threads

vrozov:


vrozov:
@vdiravka Please review

paul-rogers:
The specific fix may be OK, but I believe it overlooks a broader point. The `MappingSet` is supposed to be a static association of elements used in code generation. This mapping is meant to be static and thus sharable.

If there is an issue, then somewhere we must have added code that alters these objects per-query. We might want to back up and find that use, then ask if it is really necessary.

If so, then making the objects non-static makes sense; but I wonder what other things will be broken since the original design appeared to be that `MappingSet`s are static.

At the very least, perhaps we'd want to add a comment explaining why we need per-query modifications rather than the standard, static definitions.

vrozov:
> The specific fix may be OK, but I believe it overlooks a broader point. The `MappingSet` is supposed to be a static association of elements used in code generation. This mapping is meant to be static and thus sharable.

> If so, then making the objects non-static makes sense; but I wonder what other things will be broken since the original design appeared to be that `MappingSets` are static.

I don't see `MappingSet` being designed to be `static` and thus sharable between threads. The state of the object is modified during code generation (expression compilation) and if there are multiple queries submitted for code generation at the same time, the state machine of the `MappingSet` goes wild as a calls to `enterConstant()` may be followed by a call to `enterConstant()` for a different query on a different thread (I am not even worried about multi-thread safety at this point).

> If there is an issue, then somewhere we must have added code that alters these objects per-query. We might want to back up and find that use, then ask if it is really necessary.

The code that modifies the state of the `MappingSet` object during code generation was always there. It is not a newly introduced code.

> At the very least, perhaps we'd want to add a comment explaining why we need per-query modifications rather than the standard, static definitions.

`static` definition for `MappingSet` is not standard. In the majority of places it was already declared as instance variable (not `static`).

paul-rogers:
Thanks much for the explanation: I learned something new!
+1

PR #1287, opened 2018-05-23
DRILL-6443: Enable Search for both running AND completed queries

kkhatua:
When running a query in Drill, the `/profiles` page will show the search (and pagination) capabilities only for the top most visible table (i.e. Running Queries ).

The _Completed Queries_ table will show the search feature only when there are no _Running Queries_. This is because the backend uses a generalized freemarker macro to define the seach capabilities for the tables being rendered. With running queries, both, _Running_ and _Completed_ queries tables have the same element ID, resulting in the search capability only being applied to the first table.

This modifies the Freemarker macro to take an additional argument for distinguishing between _Running_ and _Completed_ _Queries_ tables.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/profile/list.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/list.ftl
@@ -23,26 +23,28 @@
 <script src="/static/js/jquery.dataTables-1.10.16.min.js"></script>
 <script>
     $(document).ready(function() {
-      $("#profileList").DataTable( {
-        //Preserve order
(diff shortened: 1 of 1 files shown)


kkhatua:
This is how the search bars appear with running queries:
image

kkhatua:
@arina-ielchiieva  since you had reviewed the original PR #1029 for [DRILL-5867](https://issues.apache.org/jira/browse/DRILL-5867), could you do a review of this small change?

arina-ielchiieva:
+1

PR #1295, opened 2018-05-29
DRILL-6450: Visualized plans for profiles querying JDBC sources is broken

kkhatua:
When viewing a profile for a query against a JDBC source, the visualized plan is not rendered. This is because the generated SQL pushed down to the JDBC source has a line break injected just before the FROM clause.

The workaround is to strip away any injected newlines ('\\n') at least for the SQL defined in the text plan, so that the backend Javascript can render it correctly.
In addition, any single line comments are also removed, but any block comments (i.e. /* .. */ ) are retained as they might carry hints.


kkhatua:
@amansinha100  can you review this for this small PR?

PR #1296, opened 2018-05-30
DRILL-5365: Enforce Immutability of DrillFileSystem

ilooner:
Continuing the PR that was opened here https://github.com/apache/drill/pull/796

## The Issue

If a user specifies **fs.defualt.name** in the hive storage plugin configuration, it can override the default file system defined in the FileSystemPlugin when CTAS statements are executed. This issue was occurring on a user's Drill cluster.

Chun's steps for reproducing the issue:

 1. Two or more nodes cluster. 
 2. Enable impersonation.
 3. Set **fs.default.name**: "file:///" in hive storage plugin.
 4. Restart drillbits. 
 5. As a regular user, on node A, drop the table/file. 
 6. Ctas from a large enough hive table as source to recreate the table/file. 
 7. Query the table from node A should work. 
 8. Query from node B as same user should reproduce the issue.

## The Cause

The root cause is unclear, somehow hive store plugin settings are finding their way into the Configuration object passed to the DrillFileSystem. I'm not sure if this is by design or not. If it is not by design then we should investigate the root cause.

## The Solution (Pending QA Verification)

- Force the correct value of **fs.default.name** in the FileSystemPlugin.
- Enforce immutability of the DrillFileSystem.
- Add further logging to assist in debugging similar issues in the future.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
@@ -36,7 +36,6 @@ import org.apache.hadoop.fs.ContentSummary;
 import org.apache.hadoop.fs.CreateFlag;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
-import org.apache.hadoop.fs.FileAlreadyExistsException;
 import org.apache.hadoop.fs.FileChecksum;
 import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
@@ -44,11 +43,9 @@ import org.apache.hadoop.fs.FsServerDefaults;
 import org.apache.hadoop.fs.FsStatus;
 import org.apache.hadoop.fs.LocatedFileStatus;
 import org.apache.hadoop.fs.Options.ChecksumOpt;
-import org.apache.hadoop.fs.ParentNotDirectoryException;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.PathFilter;
 import org.apache.hadoop.fs.RemoteIterator;
-import org.apache.hadoop.fs.UnsupportedFileSystemException;
 import org.apache.hadoop.fs.XAttrSetFlag;
 import org.apache.hadoop.fs.permission.AclEntry;
 import org.apache.hadoop.fs.permission.AclStatus;
@@ -65,7 +62,8 @@ import org.apache.drill.shaded.guava.com.google.common.annotations.VisibleForTes
 import org.apache.drill.shaded.guava.com.google.common.collect.Maps;
 
 /**
- * DrillFileSystem is the wrapper around the actual FileSystem implementation.
+ * DrillFileSystem is the wrapper around the actual FileSystem implementation. The {@link DrillFileSystem} is
+ * immutable.
  *
  * If {@link org.apache.drill.exec.ops.OperatorStats} are provided it returns an instrumented FSDataInputStream to
  * measure IO wait time and tracking file open/close operations.
@@ -88,23 +86,42 @@ public class DrillFileSystem extends FileSystem implements OpenFileTracker {
   }
 
   public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws IOException {
+    // Configuration objects are mutable, and the underlying FileSystem object may directly use a passed in Configuration.
+    // In order to avoid scenarios where a Configuration can change after a DrillFileSystem is created, we make a copy
+    // of the Configuration.
+    fsConf = new Configuration(fsConf);
     this.underlyingFs = FileSystem.get(fsConf);
     this.codecFactory = new CompressionCodecFactory(fsConf);
     this.operatorStats = operatorStats;
   }
 
-  @Override
-  public void setConf(Configuration conf) {
-    // Guard against setConf(null) call that is called as part of superclass constructor (Configured) of the
-    // DrillFileSystem, at which point underlyingFs is null.
-    if (conf != null && underlyingFs != null) {
-      underlyingFs.setConf(conf);
-    }
+  private void throwUnsupported() {
+    throw new UnsupportedOperationException(DrillFileSystem.class.getCanonicalName() + " is immutable and should not be changed after creation.");
   }
 
+  /**
+   * This method should never be used on {@link DrillFileSystem} since {@link DrillFileSystem} is immutable.
+   * {@inheritDoc}
+   * @throws UnsupportedOperationException when called.
+   */
+  @Override
+  public void setConf(Configuration conf) {
+    if (underlyingFs != null) {
+      throwUnsupported();
+    }
+
+    // The parent class's constructor FileSystem() calls Configured(null) which calls setConf(null).
+    // We want to let that first call to setConf succeed. Any subsequent calls would be made by a user and are not allowed.
+  }
+
+  /**
+   * Returns a copy of {@link Configuration} for this {@link DrillFileSystem}.
+   * <b>Note: </b> a copy of the {@link Configuration} is returned in order to enforce immutability.
+   * @return A copy of {@link Configuration} for this {@link DrillFileSystem}.
+   */
   @Override
   public Configuration getConf() {
-    return underlyingFs.getConf();
+    return new Configuration(this.underlyingFs.getConf());
   }
 
   /**
@@ -143,9 +160,14 @@ public class DrillFileSystem extends FileSystem implements OpenFileTracker {
     return new DrillFSDataInputStream(underlyingFs.open(f), operatorStats);
   }
 
+  /**
+   * This method should never be used on {@link DrillFileSystem} since {@link DrillFileSystem} is immutable.
+   * {@inheritDoc}
+   * @throws UnsupportedOperationException when called.
+   */
   @Override
-  public void initialize(URI name, Configuration conf) throws IOException {
-    underlyingFs.initialize(name, conf);
+  public void initialize(URI name, Configuration conf) {
+    throwUnsupported();
   }
 
   @Override
@@ -205,13 +227,12 @@ public class DrillFileSystem extends FileSystem implements OpenFileTracker {
   }
 
   @Override
-  public void createSymlink(Path target, Path link, boolean createParent) throws AccessControlException, FileAlreadyExistsException, FileNotFoundException, ParentNotDirectoryException, UnsupportedFileSystemException, IOException {
+  public void createSymlink(Path target, Path link, boolean createParent) throws IOException {
     underlyingFs.createSymlink(target, link, createParent);
   }
 
   @Override
-  public FileStatus getFileLinkStatus(Path f) throws AccessControlException, FileNotFoundException,
-      UnsupportedFileSystemException, IOException {
+  public FileStatus getFileLinkStatus(Path f) throws IOException {
     return underlyingFs.getFileLinkStatus(f);
   }
 
@@ -230,14 +251,24 @@ public class DrillFileSystem extends FileSystem implements OpenFileTracker {
     return underlyingFs.getFileChecksum(f);
   }
 
+  /**
+   * This method should never be used on {@link DrillFileSystem} since {@link DrillFileSystem} is immutable.
+   * {@inheritDoc}
+   * @throws UnsupportedOperationException when called.
+   */
   @Override
   public void setVerifyChecksum(boolean verifyChecksum) {
-    underlyingFs.setVerifyChecksum(verifyChecksum);
+    throwUnsupported();
   }
 
+  /**
+   * This method should never be used on {@link DrillFileSystem} since {@link DrillFileSystem} is immutable.
+   * {@inheritDoc}
+   * @throws UnsupportedOperationException when called.
+   */
   @Override
   public void setWriteChecksum(boolean writeChecksum) {
-    underlyingFs.setWriteChecksum(writeChecksum);
+    throwUnsupported();
   }
 
   @Override
@@ -567,9 +598,14 @@ public class DrillFileSystem extends FileSystem implements OpenFileTracker {
     return underlyingFs.getHomeDirectory();
   }
 
+  /**
+   * This method should never be used on {@link DrillFileSystem} since {@link DrillFileSystem} is immutable.
+   * {@inheritDoc}
+   * @throws UnsupportedOperationException when called.
+   */
   @Override
   public void setWorkingDirectory(Path new_dir) {
-    underlyingFs.setWorkingDirectory(new_dir);
+    throwUnsupported();
   }
 
   @Override
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
@@ -212,6 +212,6 @@ public class FileSystemPlugin extends AbstractStoragePlugin {
   }
 
   public Configuration getFsConf() {
-    return fsConf;
+    return new Configuration(fsConf);
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/ExternalSortBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/ExternalSortBatch.java
@@ -161,7 +161,7 @@ public class ExternalSortBatch extends AbstractRecordBatch<ExternalSort> {
     this.incoming = incoming;
     DrillConfig config = context.getConfig();
     Configuration conf = new Configuration();
-    conf.set("fs.default.name", config.getString(ExecConstants.EXTERNAL_SORT_SPILL_FILESYSTEM));
+    conf.set(FileSystem.FS_DEFAULT_NAME_KEY, config.getString(ExecConstants.EXTERNAL_SORT_SPILL_FILESYSTEM));
     try {
       this.fs = FileSystem.get(conf);
     } catch (IOException e) {


ilooner:
@vdiravka @paul-rogers Please review.

paul-rogers:
@ilooner, the fix will probably work, but seems a bit of a hack. It is not clear why Hive changes the default file system, though we can speculate. Drill likely uses Hive as a metastore. (Later versions of Hive split out the metastore into Hive Meta Store or HMS, so, ideally, that's what Drill would call it...)

t is unfortunate that Drill requires users to copy their Hive config from a file into a Drill JSON config object. Two copies, in two formats, is typically frowned upon.

So, we have HMS, describing data on disk. Presumably HMS wants to state the file system that contains the file, and does so as part of its config. 

Now, we want readers to read those files. One would expect that the "hive" storage plugin replace Drill's format plugin mechanisms with its own file-to-format mapping. When reading HMS files, Drill would use the "hive" options and formats. Is that how it works? Not sure.

I believe that "hive" has its own set of readers. I've seen indications that one cannot, say, use the Drill native CSV or Parquet readers (say) for "hive" files. The question is, how is all this wired together. (I don't know, haven't looked at this code.)

Supposedly, if we query a "hive" table, we need to use the "hive" file system info.

Where does this "hive" info leak into a non-"hive" reader? Perhaps joining a "hive" file with a "dfs" file?

With this fix, would that join work?

In short, I think we need to understand the above to ensure we don't actually play Whack-a-Mole and introduce a new bug by fixing another.

ilooner:
@paul-rogers I have added Chun's steps for reproducing the issue to the PR description. The issue happens when doing a table drop and CTAS. My understanding is that this would not happen when doing joins.

I agree that we should find the root cause. I will work with Chun to add extra debugging to his repro to chase down how the information is leaking into the FileSystem configuration.

However, I would like to keep the changes that enforce the immutability of DrillFileSystem. It would make things easier moving forward since a developer only has to remember two things about the DrillFileSystem. 1.) It's immutable. 2.) It's impossible for them to screw up the DrillFileSystem after it's created. Currently the DrillFileSystem's contract is ambiguous and unclear, and there are many accidental ways for a developer to corrupt the DrillFileSystem after it's created. I would rather make all of these things explicit and enforced.

I'll update the PR with an RCA in a while.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -83,28 +87,60 @@
...
+    fsConf = new Configuration(fsConf);
     this.underlyingFs = FileSystem.get(fsConf);
+
+    logger.trace("Configuration for the DrillFileSystem {} {}, underlyingFs: {}",
+      fsConf.getTrimmed(FS_DEFAULT_NAME),
+      fsConf.getTrimmed(FS_DEFAULT_NAME_KEY),
+      this.underlyingFs.getUri());
+
+    if (logger.isTraceEnabled()) {
+      logger.trace("Who made me? {}", new RuntimeException("Who made me?"));
```
vdiravka: I assume it was a temp change?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -83,28 +87,60 @@
...
-    if (conf != null && underlyingFs != null) {
-      underlyingFs.setConf(conf);
+    if (!initialized) {
+      // The parent class's constructor FileSystem() calls Configured(null) which calls setConf(null).
+      // We want to let that first call to setConf succeed. Any subsequent calls would be made by a user and are not allowed.
+      return;
     }
+
+    // A programmer using this class should never call this method.
+    throw new UnsupportedOperationException(DrillFileSystem.class.getCanonicalName() + " is immutable and should not be changed after creation.");
```
vdiravka: Do you have this exception in the process of reproducing the issue with the changes from this PR?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
```diff
@@ -54,6 +54,9 @@
  * references to the FileSystem configuration and path management.
  */
 public class FileSystemPlugin extends AbstractStoragePlugin {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FileSystemPlugin.class);
+
+  public static final String FS_DEFAULT_NAME = "fs.default.name";
```
vdiravka: It was mentioned that `fs.default.name` is deprecated. Can we replace it with `fs.defaultFS`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: Thanks for catching this. I'll remove it after I do an RCA.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: This was intentional. The idea is that if a developer makes a change where they call the setter their build will fail, since they violated the contract.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
ilooner: This was part of the magic in @chunhui-shi 's original fix. For some reason if a user defined **fs.default.name** in the hive storage plugin this was passed on to the DrillFileSystem. We even set *fs.defaultFS** on line 79, but if the deprecated **fs.default.name** option is set as well it's value wins.

After doing an RCA I will delete this.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -83,28 +87,63 @@
...
   }
 
   public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws IOException {
+    Preconditions.checkNotNull(fsConf);
+
+    // Configuration objects are mutable, and the underlying FileSystem object may directly use a passed in Configuration.
+    // In order to avoid scenarios where a Configuration can change after a DrillFileSystem is created, we make a copy
+    // of the Configuration.
+    fsConf = new Configuration(fsConf);
     this.underlyingFs = FileSystem.get(fsConf);
```
vdiravka: Looks like it can return underlyingFs from cache. I think it means that someone else can get this Fs and change it. In the result Drill will rely on changed underlyingFs. Possibly we need copy of underlyingFs too.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
vdiravka: It's good. I just was interested how immutability of DFS solves the original issue, where and how DFS was changing?
Is `setConf()` method the main fix of the issue? Maybe this method is called and fall into return statement?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: There is a complication with this. The readers currently don't close the FileSystem after they are done with it. If we were to create a new underlyingFs each time we make a DrillFileSystem we would have to make sure that we close it after we are done reading and writing. This would require updating the readers and writers that use DrillFileSystem.

I feel that would be a big change to do and I would like to do that incrementally as separate Jiras. What are your thoughts?

ilooner:
I made the FileSystem immutable to clarify my understanding of the DrillFileSystem contract. At this point I actually don't know if it has any impact on fixing the original issue or not, since the original fix was treating the symptoms not the root cause. I will update you with my findings after Chun reproduces the issue with extra debugging enable. Then we should be able to have a better idea about the true root cause.

ilooner:
## Problem

@paul-rogers @vdiravka I have an update. Chun was unable to reproduce the issue, however after staring at the code for several days, I think I see the source of the issue. I believe the issue mostly stems from two bugs in the FileSystem.get method provided to us by hadoop.

  1. **Bug # 1** is that FileSystem.get does not actually take into account the Configuration object you give it. This means that if you ask for two different FileSystem system objects with different configurations for dfs, FileSystem.get assumes both are the same and the first dfs FileSystem and config requested wins and is the one that is always returned. That's pretty crazy to me, but you can check the equals method for **FileSystem.Cache.Key** yourself.
 2. **Bug # 2** The **fs.default.name** property is not honored when doing FileSystem.get, only **fs.defaultFS** is honored. You can see this by tracing through the code:
    1. FileSystem.get(Configuration conf) calls FileSystem.getDefaultUri(Configuration conf) 
    2. FileSystem.getDefaultUri(Configuration conf) looks up FileSystem.FS_DEFAULT_NAME_KEY in the Configuration
    3. FileSystem.FS_DEFAULT_NAME_KEY is **fs.defaultFS**.
    4. If **fs.defaultFS** is not defined then the uri for caching purposes is assumed to be **file://**

These bugs would basically completely break things if hive plugin and the filesystem plugin were storing files in different places. I suspect we aren't seeing any issues because most people only have one dfs cluster to store their data for both hive and other applications.

## Solution

 * Bypass the caching in FileSystem.get and do our own caching in DrillFileSystem, which takes into account Configuration objects.
 * Set **fs.defaultFS**  as well as **fs.default.name** in our configuration objects.
 * Make DrillFileSystem immutable and make things cleaner and more explicit in general.

ilooner:
I also forgot to mention a third possible bug, which is that FileSystemConfigurations defined in the HiveStoragePlugin are passed to DrillFileSystem in HiveDrillNativeParquetRowGroupScan.getFsConf(). I'm not sure if that is by design or not, so I think we need to think about that one a bit more.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
vdiravka: Agree. Just leave TODO here with a note and number of Jira

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
```diff
@@ -76,6 +79,16 @@ public FileSystemPlugin(FileSystemConfig config, DrillbitContext context, String
           fsConf.set(s, config.config.get(s));
         }
       }
+
+      logger.info("Original FileSystem default fs configuration {} {}",
+        fsConf.getTrimmed(FS_DEFAULT_NAME),
+        fsConf.getTrimmed(FileSystem.FS_DEFAULT_NAME_KEY));
+
+      if (logger.isInfoEnabled()) {
+        logger.info("Who made me? {}", new RuntimeException("Who made me?"));
```
vdiravka: Is it proper message or you just forgot to delete it?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -179,9 +182,16 @@ public FSDataInputStream open(Path f) throws IOException {
     return new DrillFSDataInputStream(underlyingFs.open(f), operatorStats);
   }
 
+  /**
+   * This method should never be used on {@link DrillFileSystem} since {@link DrillFileSystem} is immutable.
+   * @param name
```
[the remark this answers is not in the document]
ilooner: done

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -102,13 +102,13 @@ public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws
...
-    logger.trace("Configuration for the DrillFileSystem {} {}, underlyingFs: {}",
+    logger.info("Configuration for the DrillFileSystem {} {}, underlyingFs: {}",
       fsConf.getTrimmed(FS_DEFAULT_NAME),
       fsConf.getTrimmed(FS_DEFAULT_NAME_KEY),
       this.underlyingFs.getUri());
 
-    if (logger.isTraceEnabled()) {
-      logger.trace("Who made me? {}", new RuntimeException("Who made me?"));
+    if (logger.isInfoEnabled()) {
+      logger.info("Who made me? {}", new RuntimeException("Who made me?"));
```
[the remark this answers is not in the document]
ilooner: removed

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
```diff
@@ -80,12 +80,12 @@ public FileSystemPlugin(FileSystemConfig config, DrillbitContext context, String
...
 
-      logger.trace("Original FileSystem default fs configuration {} {}",
+      logger.info("Original FileSystem default fs configuration {} {}",
         fsConf.getTrimmed(FS_DEFAULT_NAME),
         fsConf.getTrimmed(FileSystem.FS_DEFAULT_NAME_KEY));
 
-      if (logger.isTraceEnabled()) {
-        logger.trace("Who made me? {}", new RuntimeException("Who made me?"));
+      if (logger.isInfoEnabled()) {
+        logger.info("Who made me? {}", new RuntimeException("Who made me?"));
```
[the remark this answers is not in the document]
ilooner: removed

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: Added a TODO to the javadoc in DrillFileSystemCache.java and created https://issues.apache.org/jira/browse/DRILL-6608.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
ilooner: removed

ilooner:
For some reason github won't let me respond next to one of your comments, so putting response here:

Searched for usages of **fs.default.name** there was one other usage in ExternalSortBatch, so I replaced it to use the new constant in DrillFileSystem.

ilooner:
Created a jira for the HiveDrillNativeParquetRowGroupScan issue here https://issues.apache.org/jira/browse/DRILL-6609

ilooner:
@vdiravka addressed your comments, and  I pushed a few more changes after your last pass. I added the DrillFileSystemCache class to use instead of the hadoop FileSystem cache.

priteshm:
@vdiravka any more review comments?

vdiravka:
+1 LGTM

arina-ielchiieva:
@ilooner removed ready-to-commit label since I want to make sure we address comment about concurrency.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
```diff
@@ -0,0 +1,99 @@
...
+ * <ul>
+ *   <li>
+ *     Drill currently keeps a {@link org.apache.drill.exec.store.dfs.DrillFileSystem} open indefinitely. This will be corrected
+ *     in DRILL-6608. As a result this cache currently has no methods to remove {@link org.apache.hadoop.fs.FileSystem} objects
+ *     after they are created.
+ *   </li>
+ * </ul>
+ */
+class DrillFileSystemCache {
+  private Map<Map<String, String>, FileSystem> cache = new HashMap<>();
```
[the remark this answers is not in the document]
ilooner: @arina-ielchiieva Thanks for catching this. 

Making the maps concurrent wouldn't be enough, there would still be a race condition in **get** since we are checking and then setting a file system. So we could create two copies of the same file system. I made the get method synchronized instead.

Please take another look.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
```diff
@@ -0,0 +1,99 @@
...
+
+    return cache.get(map);
+  }
+
+  static Map<String, String> configToMap(final Configuration configuration) {
+    Preconditions.checkNotNull(configuration);
+    final Map<String, String> map = new HashMap<>();
+
+    for (Map.Entry<String, String> entry: configuration) {
+      map.put(entry.getKey().trim(), entry.getValue());
```
sohami: In this implementation we are only considering `Configuration` object in Cache key. How about other parameters which [HadoopFileSystem cache](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2812) is using ? Specially I am concerned about the UGI object from which the user information is retrieved which is later checked for access control.

Also for handling `fs.default.name` and `fs.defaultFS` once it's set in the configuration parameter and FileSystem instance is created then [FileSystemCache does use information from there based on UR](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2806)I (which is created using those parameters).

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -65,46 +62,105 @@
...
     this(fsConf, null);
   }
 
   public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws IOException {
-    this.underlyingFs = FileSystem.get(fsConf);
+    // Configuration objects are mutable, and the underlying FileSystem object may directly use a passed in Configuration.
+    // In order to avoid scenarios where a Configuration can change after a DrillFileSystem is created, we make a copy
+    // of the Configuration.
+    this.fsConf = new Configuration(fsConf);
+    normalize(fsConf);
```
sohami: Why not use `Configuration.setDeprecateProperties` instead ? See [here](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L577)

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -65,46 +62,105 @@
...
+
+    this.underlyingFs = CACHE.get(fsConf);
     this.codecFactory = new CompressionCodecFactory(fsConf);
     this.operatorStats = operatorStats;
+    this.initialized = true;
```
arina-ielchiieva: @ilooner if this flag is set in constructor to true do we need it at all?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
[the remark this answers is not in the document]
arina-ielchiieva: static final?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: This is required. The case is somewhat strange and is the following:

  1. The super constructor is called FileSystem()
  1. The FileSystem constructor calls its super constructor Configured and passes null for the configuration.
  1. The Configured constructor tries to set a null configuration by doing setConf(null)

The problem is we want to throw an exception if someone accidentally calls setConf. However, we don't want setConf to throw an exception when called from this chain of super constructor calls. Therefore we need the initialized flag to distinguish between the two cases. 

Frankly I don't like setting the flag myself, but it is required due to design flaws in the hadoop API. The specific flaw in the api is that class constructors should never call overridable methods.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
ilooner: Thanks for showing me this, never knew about it. 

I think setDeprecateProperties does some of what we need, but would still have a problem. Consider the following input and output mappings for setDeprecateProperties.

(fs.default.name: val1 ) -> (fs.default.name: val1 )
(fs.default.name: val1, fs.defaultFS: val2 ) -> (fs.default.name: val1, fs.defaultFS: val2 )
(fs.default.name: val1, fs.defaultFS: val1 ) -> (fs.default.name: val1, fs.defaultFS: val1 )

Ideally we would want all this outputs to be the same and we would want a warning if our given config has different values for fs.default.name and fs.defaultFS.

The normalize method handles this cases. I am trying to think of a way to use setDeprecateProperties to replace some of the existing code in normalize. But I can't think of a way to use setDeprecateProperties with the extra case handling to make the code any simpler than what it is now. 

If you have any ideas on a way we could use setDeprecateProperties and handle the other cases described above more simply than the current normalize method, let me know and I will do it. Otherwise if we can't think of anything now, I'd prefer to keep the normalize method as is.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
ilooner: - The UGI object is only created and used in the Key when FileSystem.getUnique is called. The call DrillFileSystem was using previously was FileSystem.get, so the cache never used UGI objects as part of the key for our case.  The behavior is pre-existing and I'm not sure if it's a bug or now. We should determine if it is a bug or not, if it is I would like to fix that as a separate Jira. What are your thoughts?

- For your second point, you are right. However, there is a bug in the logic of FileSystem.getDefaultUri (which is used to construct the key) that if both fs.defaultFS and fs.default.name are defined in a configuration the deprecated option (fs.defaultFS) wins. Also it does not take into account the rest of the configuration object in the key. So if you require two different configurations for the same file system you are out of luck, only the first one wins.

ilooner:
@arina-ielchiieva @sohami Replied to comments. Please let me know if you have more comments.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
sohami: - Usageof UGI object is done in such a way that FileSystem instance is created in doAs block of it. UGI holds the information about the trusted user. Now when DrillFilesystem constructor is called it calls `FileSystem.get`. This internally calls `CACHE.get(uri, conf)` which creates an instance of `CACHE.Key` (See [here](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L371)). In the `Key` constructor the call to `UGI.getCurrentUser()` returns the context of user under which the application called the constructor. Later this user context is also used in `initialize` of specific implementation of FileClient. Like for HDFS client it is used [here](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DFSClient.java#L647) again. But for future call if DrillFileSystem is called under same user block then the cache version of correct filesystem object will be returned. So in short we have to use it otherwise there will be security regression.

- I don't understand why are we preferring deprecated property over the new one ? `fs.default.name` is deprecate one and `fs.defaultFS` is new property. (See [here](https://hadoop.apache.org/docs/r2.7.2/hadoop-project-dist/hadoop-common/DeprecatedProperties.html)) . Also [Configuration is a static field of FileSystem class](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L3301) so for same scheme there cannot be 2 different configuration instances. Looks like that was the intention and hence CACHE also doesn't use it.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
sohami: Please see comment below `fs.defaultFS` is new property and f`s.default.name` is the deprecated one. I can look more on how to actually use deprecatedContext but my assumption was internally FileSystem library must be initializing the mapping between deprecated and new properties and use new properties value correctly. Also `fs.defaultFS` will always be set since it has a default value of `file:///`

ilooner: Wow, I had it completely mixed up. You are right, thanks for catching this before it went in! I will fix the normalize method, and revisit trying to use setDeprecateProperties.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
ilooner: - The code for CACHE.get calls the Key(uri, conf) constructor https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2667 . That constructor does not initialize the ugi object, so it is not used as part of the key. https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2801 
The UGI object is only used when CACHE.getUnique is called. In that method the Key(uri, conf, unique) constructor is called https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2673 . Only in that case does the Key constructor call UGI.getCurrentUser() . https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L2801 . So if not using the UGI object is a security issue, it is a pre-existing issue. Please double check the code I referenced to make sure my analysis is correct.
 - You are right I had the deprecated and non-deprecated property names mixed up. Thank you very much for catching this! 
 - While there is static conf field in the FileSystem class, no one seems to use it. For example DistrubutedFileSystem calls the setConf method https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/DistributedFileSystem.java#L137, which sets a conf variable only for that instance of the class https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configured.java (DistributedFileSystem extends FileSystem which extends Configured). In fact as far as I can tell from my IDE the static conf variable is never used. Whenever the FileSystem class uses the configuration it calls getConf() which is a method of Configured and which returns the configuration that is stored on a per instance basis. So I think the configuration should still be used in the key for caching.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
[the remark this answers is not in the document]
ilooner: Some how missed this comment last pass. Fixing it now.

ilooner:
@sohami I unit tested setDeprecatedProperties which you suggested. It does exactly what is needed, so I removed the normalize method and used that instead. Doing this fixes the deprecated option mixup as well. I responded to your other comments as well. Please let me know your thoughts.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -65,46 +62,75 @@
...
     this(fsConf, null);
   }
 
   public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws IOException {
-    this.underlyingFs = FileSystem.get(fsConf);
+    // Configuration objects are mutable, and the underlying FileSystem object may directly use a passed in Configuration.
+    // In order to avoid scenarios where a Configuration can change after a DrillFileSystem is created, we make a copy
+    // of the Configuration.
+    this.fsConf = new Configuration(fsConf);
+    fsConf.setDeprecatedProperties();
```
sohami: I don't think this is needed because deprecation properties is handled for each `set` call on configuration object. See [here](https://github.com/apache/hadoop/blob/branch-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/conf/Configuration.java#L1138)

ilooner:
@sohami Thanks for saving the day on this PR and pointing out the issues before the changes got merged. With the flaws you pointed out, I think my proposed solution has been debunked. There was one other possible source of the issue I identified earlier, but thought was less likely.

"FileSystemConfigurations defined in the HiveStoragePlugin are passed to DrillFileSystem in HiveDrillNativeParquetRowGroupScan.getFsConf(). " 

I will explore that possible bug, and try to replicate an error scenario in a unit test. If I find anything I will update this PR with a fix for that issue. If that is not a viable bug, I will repurpose this PR to only enforce the immutability of DrillFileSystem and include no additional changes.

Thanks again!

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystemCache.java
sohami: Makes sense and probably that's the reason `Cache` only uses `Scheme` and `authority` properties which are derived from `fs.defaultFS` parameter. I guess the proxy servers are provided as `authority` strings. All Kerberos and security related parameters are encapsulated within `UGI` object.

ilooner:
@sohami I have repurposed DRILL-5365 to focus only on documenting and enforcing the immutability of the DrillFileSystem. Please take a look at the latest changes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -79,32 +77,56 @@
...
+   * {@inheritDoc}
+   * @throws UnsupportedOperationException when called.
+   */
   @Override
   public void setConf(Configuration conf) {
-    // Guard against setConf(null) call that is called as part of superclass constructor (Configured) of the
-    // DrillFileSystem, at which point underlyingFs is null.
-    if (conf != null && underlyingFs != null) {
-      underlyingFs.setConf(conf);
+    if (!initialized) {
```
sohami: We can still use `(underlyingFs != null)` check and get rid of the `initialized` flag. New check can be:

```
if (underlyingFs != null) {
   throwUnsupported();
}
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/DrillFileSystem.java
```diff
@@ -79,32 +77,56 @@
...
     this(fsConf, null);
   }
 
   public DrillFileSystem(Configuration fsConf, OperatorStats operatorStats) throws IOException {
-    this.underlyingFs = FileSystem.get(fsConf);
-    this.codecFactory = new CompressionCodecFactory(fsConf);
+    // Configuration objects are mutable, and the underlying FileSystem object may directly use a passed in Configuration.
+    // In order to avoid scenarios where a Configuration can change after a DrillFileSystem is created, we make a copy
+    // of the Configuration.
+    this.fsConf = new Configuration(fsConf);
```
sohami: We can convert `this.fsConf` to be a local variable and in `getConf()` do 
`return new Configuration(underlyingFs.getConf());`

ilooner:
@sohami applied your review comments.

PR #1302, opened 2018-06-05
DRILL-6340: Output Batch Control in Project using the RecordBatchSizer

bitblender:
Changes required to implement Output Batch Sizing in Project using the RecordBatchSizer.

### DIFF
--- a/common/src/test/resources/logback-test.xml
+++ b/common/src/test/resources/logback-test.xml
@@ -50,7 +50,7 @@
   </appender>
 
   <root>
-    <level value="error"/>o
+    <level value="error"/>
     <if condition='property("drill.lilith.enable").equalsIgnoreCase("true")'>
       <then>
         <appender-ref ref="SOCKET"/>
--- a/common/src/test/resources/logback-test.xml
+++ b/common/src/test/resources/logback-test.xml
@@ -50,7 +50,7 @@
   </appender>
 
   <root>
-    <level value="error"/>
+    <level value="error"/>o
     <if condition='property("drill.lilith.enable").equalsIgnoreCase("true")'>
       <then>
         <appender-ref ref="SOCKET"/>
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
@@ -210,7 +210,7 @@ public class OutputWidthVisitor extends AbstractExecExprVisitor<OutputWidthExpre
             TypedFieldId fieldId = varLenReadExpr.getReadExpression().getTypedFieldId();
             columnName =  TypedFieldId.getPath(fieldId, state.manager.getIncomingBatch());
         }
-        final RecordBatchSizer.ColumnSize columnSize = state.manager.getComplexColumnSize(columnName);
+        final RecordBatchSizer.ColumnSize columnSize = state.manager.getColumnSize(columnName);
 
         int columnWidth = columnSize.getNetSizePerEntry();
         return new FixedLenExpr(columnWidth);
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
@@ -105,8 +105,8 @@ public class OutputWidthVisitor extends AbstractExecExprVisitor<OutputWidthExpre
         final DrillFuncHolder holder = ((DrillFuncHolderExpr) holderExpr).getHolder();
 
         // If the user has provided a size estimate, use it
-        int estimate = holder.variableOuputSizeEstimate();
-        if (estimate != FunctionTemplate.VARIABLE_OUTPUT_SIZE_ESTIMATE_DEFAULT) {
+        int estimate = holder.variableOutputSizeEstimate();
+        if (estimate != FunctionTemplate.OUTPUT_SIZE_ESTIMATE_DEFAULT) {
             return new FixedLenExpr(estimate);
         }
         // Use the calculator provided by the user or use the default
@@ -124,7 +124,7 @@ public class OutputWidthVisitor extends AbstractExecExprVisitor<OutputWidthExpre
 
     /**
      * Records a variable width write expression. This will be converted to a {@link FixedLenExpr} expression by walking
-     * the tree of expression attached to the write expression .
+     * the tree of expression attached to the write expression.
      * @param writeExpr
      * @param state
      * @return
@@ -135,7 +135,7 @@ public class OutputWidthVisitor extends AbstractExecExprVisitor<OutputWidthExpre
                                                                  OutputWidthVisitorState state) throws RuntimeException {
         TypedFieldId fieldId = writeExpr.getFieldId();
         ProjectMemoryManager manager = state.getManager();
-        OutputWidthExpression outputExpr = null;
+        OutputWidthExpression outputExpr;
         if (manager.isFixedWidth(fieldId)) {
             outputExpr = getFixedLenExpr(fieldId.getFinalType());
         } else {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
@@ -1,0 +1,278 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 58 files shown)


common/src/test/resources/logback-test.xml
```diff
@@ -50,7 +50,7 @@
   </appender>
 
   <root>
-    <level value="error"/>
+    <level value="error"/>o
```
ilooner: typo

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -193,6 +209,13 @@ protected IterOutcome doWork() {
             }
           }
           incomingRecordCount = incoming.getRecordCount();
+          memoryManager.update();
+          if (logger.isTraceEnabled()) {
+            logger.trace("doWork():[1] memMgr RC " + memoryManager.getOutputRowCount()
```
ilooner: use {} instead of concatenating strings with +

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -206,12 +229,24 @@ protected IterOutcome doWork() {
     first = false;
     container.zeroVectors();
 
-    if (!doAlloc(incomingRecordCount)) {
+
+    int maxOuputRecordCount = memoryManager.getOutputRowCount();
+    if (logger.isTraceEnabled()) {
+      logger.trace("doWork():[2] memMgr RC " + memoryManager.getOutputRowCount()
```
ilooner: Same as above.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -206,12 +229,24 @@ protected IterOutcome doWork() {
...
+    }
+    if (!doAlloc(maxOuputRecordCount)) {
       outOfMemory = true;
       return IterOutcome.OUT_OF_MEMORY;
     }
+    long projectStartTime = System.currentTimeMillis();
+    final int outputRecords = projector.projectRecords(this.incoming,0, maxOuputRecordCount, 0);
+    long projectEndTime = System.currentTimeMillis();
+    if (logger.isTraceEnabled()) {
+      logger.trace("doWork(): projection: " + " records " + outputRecords + ", time " + (projectEndTime - projectStartTime) + " ms");
```
ilooner: Use {}

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -230,18 +265,37 @@ protected IterOutcome doWork() {
...
+    assert this.memoryManager.incomingBatch == incoming;
+    final int recordsToProcess = Math.min(remainingRecordCount, memoryManager.getOutputRowCount());
+
+    if (!doAlloc(recordsToProcess)) {
       outOfMemory = true;
       return;
     }
-    final int projRecords = projector.projectRecords(remainderIndex, remainingRecordCount, 0);
+    if (logger.isTraceEnabled()) {
+      logger.trace("handleRemainder: remaining RC " + remainingRecordCount + " toProcess " + recordsToProcess
```
ilooner: {}

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
```diff
@@ -497,30 +561,36 @@ private void setupNewSchemaFromInput(RecordBatch incomingBatch) throws SchemaCha
...
+      //codeGen.saveCodeForDebugging(true);
       this.projector = context.getImplementationClass(codeGen);
       projector.setup(context, incomingBatch, this, transfers);
     } catch (ClassTransformationException | IOException e) {
       throw new SchemaChangeException("Failure while attempting to load generated class", e);
     }
+
+    long setupNewSchemaEndTime = System.currentTimeMillis();
+    if (logger.isTraceEnabled()) {
+      logger.trace("setupNewSchemaFromInput: time " + (setupNewSchemaEndTime - setupNewSchemaStartTime) + " ms" + ", proj " + this + " incoming " + incomingBatch);
```
ilooner: {}

bitblender:
I don't think my tests are contributing a whole lot to the slowdown. With this change, Project will split large batches in tests(like 45 M batches in TestLargeFileCompilation) to ~16M batches. This somehow seems to be slowing down operators like Sort and TopN a fair bit compared to the no-split runs. This needs further investigation to see if these slow-down are un-avoidable .

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
bitblender: Ok. Will do. But what is the advantage of using '{}' over static string concatenation with a check? I think format strings are more error(argument ordering errors) prone than concatenation.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
```diff
@@ -0,0 +1,278 @@
...
+        // If the user has provided a size estimate, use it
+        int estimate = holder.variableOuputSizeEstimate();
+        if (estimate != FunctionTemplate.VARIABLE_OUTPUT_SIZE_ESTIMATE_DEFAULT) {
+            return new FixedLenExpr(estimate);
+        }
+        // Use the calculator provided by the user or use the default
+        OutputWidthCalculator widthCalculator = holder.getOutputWidthCalculator();
+        final int argSize = holderExpr.args.size();
+        ArrayList<OutputWidthExpression> arguments = null;
+        if (argSize != 0) {
```
Ben-Zvi: Trivial question: Would it also work without that if() check ?  I.e., use an empty ArrayList in case of zero args.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectMemoryManager.java
```diff
@@ -0,0 +1,315 @@
...
+ *
+ * Setup phase: As and when ProjectRecordBatch creates or transfers a field, it registers the field with PMM.
+ * If the field is a variable width field, PMM records the expression that produces the variable
+ * width field. The expression is a tree of LogicalExpressions. The PMM walks this tree of LogicalExpressions
+ * to produce a tree of OutputWidthExpressions. The widths of Fixed width fields are just accumulated into a single
+ * total. Note: The PMM, currently, cannot handle new complex fields, it just uses a hard-coded estimate for such fields.
+ *
+ *
+ * Execution phase: Just before a batch is processed by Project, the PMM walks the tree of OutputWidthExpressions
+ * and conversts them to FixedWidthExpressions. It uses the RecordBatchSizer and the function annotations to do this conversion.
```
Ben-Zvi: "converts"

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectMemoryManager.java
```diff
@@ -0,0 +1,315 @@
...
+                assert width >= 0;
+            }
+            totalVariableColumnWidth += width;
+        }
+        rowWidth += totalFixedWidthColumnWidth;
+        rowWidth += totalComplexColumnWidth;
+        rowWidth += totalVariableColumnWidth;
+        int outPutRowCount;
+        if (rowWidth != 0) {
+            setOutputRowCount(getOutputBatchSize(), rowWidth);
```
Ben-Zvi: Maybe add a little comment here; this is the important place in the code where the output size for Project is being set ....

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectRecordBatch.java
ilooner: It's easier to read for the programmer. Also the string concatenation is only executed if the debug statement is executed. If you use + you have to have if statements around your debug statements to avoid any performance impact.

```
if (logger.isDebug()) {
 logger.debug("my " + variable + " message");
}
```

If you use the {} you don't have to have if statements in order to prevent a performance impact

```
logger.debug("my {} message", variable);
```

It's typically considered best practice to do the latter since it's much cleaner.

ilooner:
@bitblender Looks like this is pretty close to being done and has already been +1'ed. Could you just fix conflicts and push your latest changes addressing the review comments, and indicate which comments you addressed and which comments you decided to leave as is. After that I'll add the **ready-to-commit** tag.

bitblender:
@ilooner I have addressed all the comments and I am running the changes through the pre-commit framework. You can go ahead and mark it as ready-to-commit

bitblender:
@ilooner @Ben-Zvi let me know when you are done looking at my latest round of changes in response to the review comments. I will rebase on master, fix the conflict and force push.

ilooner:
Thanks @bitblender +1 LGTM

PR #1303, opened 2018-06-05
DRILL-6463 : Fix integer overflow in MockGroupScanPOP

gparai:
@vrozov @Ben-Zvi please review the PR. Thanks!

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/ScanStats.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/ScanStats.java
@@ -23,27 +23,27 @@ public class ScanStats {
 
   public static final ScanStats ZERO_RECORD_TABLE = new ScanStats(GroupScanProperty.EXACT_ROW_COUNT, 0, 1, 1);
 
-  private final long recordCount;
-  private final float cpuCost;
-  private final float diskCost;
+  private final double recordCount;
+  private final double cpuCost;
+  private final double diskCost;
   private final GroupScanProperty property;
 
-  public ScanStats(GroupScanProperty property, long recordCount, float cpuCost, float diskCost) {
+  public ScanStats(GroupScanProperty property, double recordCount, double cpuCost, double diskCost) {
     this.recordCount = recordCount;
     this.cpuCost = cpuCost;
     this.diskCost = diskCost;
     this.property = property;
   }
 
-  public long getRecordCount() {
+  public double getRecordCount() {
     return recordCount;
   }
 
-  public float getCpuCost() {
+  public double getCpuCost() {
     return cpuCost;
   }
 
-  public float getDiskCost() {
+  public double getDiskCost() {
     return diskCost;
   }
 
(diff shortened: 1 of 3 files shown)


gparai:
@Ben-Zvi the Calcite interface RelOptCost uses doubles for rows, io, cpu. This translates to DrillCostBase ending up using doubles. ScanStats feeds into DrillCostBase, hence the rationale of using doubles. One reason for using doubles is rowcounts are often estimated and end up as not being integers. I guess doubles were used to keep the cost model as accurate as possible.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
```diff
@@ -121,9 +121,9 @@ public void onMatch(RelOptRuleCall call) {
 
     final RelDataType scanRowType = constructDataType(agg, result.keySet());
 
-    final DynamicPojoRecordReader<Long> reader = new DynamicPojoRecordReader<>(
+    final DynamicPojoRecordReader<Double> reader = new DynamicPojoRecordReader<>(
```
vrozov: This will change return type for "select count(*) ...", let's keep it to `Long`.

vrozov:
LGTM

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
gparai: Yes, agree. Done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ConvertCountToDirectScan.java
```diff
@@ -154,7 +154,7 @@ public void onMatch(RelOptRuleCall call) {
   private Map<String, Long> collectCounts(PlannerSettings settings, DrillAggregateRel agg, DrillScanRel scan, DrillProjectRel project) {
     final Set<String> implicitColumnsNames = ColumnExplorer.initImplicitFileColumns(settings.getOptions()).keySet();
     final GroupScan oldGrpScan = scan.getGroupScan();
-    final long totalRecordCount = oldGrpScan.getScanStats(settings).getRecordCount();
+    final long totalRecordCount = (long)oldGrpScan.getScanStats(settings).getRecordCount();
```
Ben-Zvi: getRecordCount() (in ScanStats.java) already returns a long, so a cast here is not needed ....

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/ScanStats.java
```diff
@@ -23,27 +23,27 @@
...
-  private final long recordCount;
-  private final float cpuCost;
-  private final float diskCost;
+  private final double recordCount;
+  private final double cpuCost;
+  private final double diskCost;
   private final GroupScanProperty property;
 
-  public ScanStats(GroupScanProperty property, long recordCount, float cpuCost, float diskCost) {
+  public ScanStats(GroupScanProperty property, double recordCount, double cpuCost, double diskCost) {
```
sohami: There are other GroupScanPop where the rowCount stats for example is converted from double to long while creating ScanStats. Whereas in most of the cases rowCount is long only. 
Since we are making this change for treating rowCount or other stats as double, shouldn't we update other GroupScanPop's as well to make sure they are creating stats using correct type ?

[Example of rowCount with double type but converted to long.](https://github.com/apache/drill/blob/master/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcGroupScan.java#L82)
[Example of rowCount with long type.](https://github.com/apache/drill/blob/master/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/JsonTableGroupScan.java#L188)

gparai: Yes, you are correct. There are more examples of where we should do so e.g. HBaseGroupScan. However, the original JIRA was to fix the integer overflow so I skipped these changes. I will create another JIRA to address these issues.

sohami: LGTM +1

PR #1305, opened 2018-06-05
DRILL-6467: Percentage usage of memory is reported as zero by the WebUI

kkhatua:
The memory reported as a percentage is incorrectly reported as 0%, irrespective of the actual usage.
image

This is because of a snippet in the Javascript that computes the percentage:
```javascript
      //Compute Usage
      function computeMemUsage(used, max) {
        let percent = 0;
        if ( max > 0) {
          let percent = Math.round((100 * used / max), 2);
        }
        let usage   =  bytesInGB(used, 2) + "GB (" + Math.max(0, percent) + "% of "+ bytesInGB(max, 2) +"GB)";
        return usage;
      }
```

Turns out, the `let` keyword creates a new `percent` variable within the scope of the `if (max > 0) { .. }` block, so the original variable never got updated. The fix was to simply remove the `let` preceding the percentage computation.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/index.ftl
+++ b/exec/java-exec/src/main/resources/rest/index.ftl
@@ -554,7 +554,7 @@
       function computeMemUsage(used, max) {
         let percent = 0;
         if ( max > 0) {
-          let percent = Math.round((100 * used / max), 2);
+          percent = Math.round((100 * used / max), 2);
         }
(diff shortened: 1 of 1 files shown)


kkhatua:
Verified with screenshot
image

sohami:
+1 LGTM

PR #1306, opened 2018-06-05
DRILL-6468: OOMs trigger graceful shutdown when terminating Drill. This can cause a hang.

ilooner:
See JIRA for details https://issues.apache.org/jira/browse/DRILL-6468 .

### DIFF
--- a/common/src/main/java/org/apache/drill/common/CatastrophicFailure.java
+++ b/common/src/main/java/org/apache/drill/common/CatastrophicFailure.java
@@ -30,7 +30,7 @@ public class CatastrophicFailure {
    * Exit the VM as we hit a catastrophic failure.
    * @param e
    *          The Throwable that occurred
-   * @param name
+   * @param message
    *          A descriptive message
    * @param code
    *          An error code to exit the JVM with.
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -360,6 +360,11 @@ public class Drillbit implements AutoCloseable {
 
     @Override
     public void run() {
+      if (FailureUtils.hadUnrecoverableFailure()) {
+        // We cannot close drill cleanly in this case.
+        return;
+      }
+
       logger.info("Received shutdown request.");
       try {
         /*
@@ -453,5 +458,4 @@ public class Drillbit implements AutoCloseable {
     // return as-is
     return s;
   }
-
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/FailureUtils.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/FailureUtils.java
@@ -1,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.server;
+
+import io.netty.util.internal.OutOfDirectMemoryError;
+import org.apache.drill.common.CatastrophicFailure;
+import org.apache.drill.exec.exception.OutOfMemoryException;
+
+public final class FailureUtils {
+  private static volatile boolean unrecoverableFailure;
+
+  public static final int EXIT_CODE_HEAP_OOM = -1;
+  public static final int EXIT_CODE_JAVA_ERROR = -2;
+
+  /**
+   * This message is used to distinguish between direct memory and heap memory {@link OutOfMemoryError}s.
+   */
+  public static final String DIRECT_MEMORY_OOM_MESSAGE = "Direct buffer memory";
+
+  private FailureUtils() {
+    // Don't instantiate
+  }
+
+  public static boolean isDirectMemoryOOM(Throwable e) {
+    if (e instanceof OutOfDirectMemoryError || e instanceof OutOfMemoryException) {
+      // These are always direct memory errors
+      return true;
+    }
+
+    return (e instanceof OutOfMemoryError) && DIRECT_MEMORY_OOM_MESSAGE.equals(e.getMessage());
+  }
+
+  public static boolean isHeapOOM(Throwable e) {
+    return (e instanceof OutOfMemoryError) && !DIRECT_MEMORY_OOM_MESSAGE.equals(e.getMessage());
+  }
+
+  public static void unrecoverableFailure(Throwable e, String message, int exitCode) {
+    unrecoverableFailure = true;
+    CatastrophicFailure.exit(e, message, exitCode);
+  }
+
+  public static boolean hadUnrecoverableFailure() {
+    return unrecoverableFailure;
+  }
+}
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
@@ -23,7 +23,6 @@ import com.google.protobuf.InvalidProtocolBufferException;
 import io.netty.channel.ChannelFuture;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
-import org.apache.drill.common.CatastrophicFailure;
 import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.logical.LogicalPlan;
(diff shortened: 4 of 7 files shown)


ilooner:
@vrozov @dvjyothsna Please review.

dvjyothsna:
LGTM +1

common/src/main/java/org/apache/drill/common/CatastrophicFailure.java
```diff
@@ -46,6 +46,7 @@ public static void exit(Throwable e, String message, int code) {
       Thread.sleep(1000);
     } catch (InterruptedException e2) {
     }
-    System.exit(code);
+
+    Runtime.getRuntime().halt(code);
```
vrozov: I am not 100% convinced that using `halt()` it is the right approach. Note that by default, if OOM Error is thrown installed hooks are invoked by a JVM:
```
public class OOMTest {
  public static void main(String[] args) {
    Runtime.getRuntime().addShutdownHook(new Thread() {
      @Override
      public void run() {
        System.out.println("Hook");
      }
    });
    Object[] a = new Object[Integer.MAX_VALUE];
  }
}
```

ilooner: Yes. That is because the OOM is not caught. Drill catches OOMs and calls CatastrophicFailure.exit. If you catch the OOM and then hault the hooks don't get run. I have verified this behavior with this

```
  public static void main(String[] args) {
    Runtime.getRuntime().addShutdownHook(new Thread() {
      @Override
      public void run() {
        System.out.println("Hook");
      }
    });
    
    try {
      Object[] a = new Object[Integer.MAX_VALUE];
    } catch (OutOfMemoryError e) {
      Runtime.getRuntime().halt(0);
    }
  }
```

vrozov: My point is that default JVM behavior is that OOM does not bypass shutdown hook and other libraries may install shutdown hook that they expect will be called on JVM exit.

From Java documentation "This (halt) method should be used with extreme caution.", so I would prefer a solution that bypasses Drill close sequence, but still gives other libraries chance to invoke shutdown hooks.

ilooner: I think this boils down to what you consider a Heap OOM. In my mind if we experience a heap OOM we cannot continue executing any code safely, this includes any shutdown sequences. The reason why it is not safe is because the JVM can get stuck in gc during a memory allocation. So if we run any shutdown code we can hang, and that was the issue we were trying to fix in the first place. Additionally I'm not totally convinced we would want to bypass Drill's close sequence but run the Shutdown hooks for the libraries that Drill uses. That could lead to a dangerous situation where Drill continues running when some underlying resource are defunct and we can get even more errors and hangs.

At the end of the day though it's just your word against mine. So let's do an experiment. I'll go your route and add a flag to bypass only Drill's ShutdownThread in the event of a CatastrophicFailure. If the issue of hanging shutdowns never occurs again for users, great. If it continues to occur for users after this change, then we should agree to make the change to forcibly terminate the JVM with halt.

vrozov: Heap OOM means that JVM tried to allocate a new instance of some class and it failed for whatever reason (not enough heap available, GC taking too long and etc). It does not mean that JVM is fully unstable and will fail during GC. It means that any attempt to allocate another class may (but is not required as in my example if you handle OOM and instantiate a small object) fail as well. JVM always has reserved memory to handle OOM.

Your plan to try first a flag to bypass only Drill close sequence in the shutdown hook works for me.

ilooner: Also there is one more issue with both approaches I thought of. Currently, CatastrophicFailure.exit is only called when an OOM is explicitly caught in the Foreman and FragmentExecutor. However, these are not all the places where catastrophic failures can happen. If an OOM happens anywhere else we will continue to do the wrong thing and try to run the normal shutdown sequence. We should use https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#setUncaughtExceptionHandler-java.lang.Thread.UncaughtExceptionHandler- to set the default exception handler on all threads to properly call CatastrophicFailure.exit in the event of a heap OOM.

So I will implement this as well.

ilooner:
Spoke offline with Vlad. We should not be overriding the thread pools to handle uncaught exceptions. This is because if we execute callables and get a future back the exceptions will automatically be thrown in the parent thread when future.get is called. However, in some places we are using custom runnable implementations like in the PartitionDecorator. In this case exceptions thrown in the thread pool are caught and set on the executor state, but are not handled by the parent thread. We will have to address this gap in error and exception handling in https://issues.apache.org/jira/browse/DRILL-6484 .

ilooner:
@vrozov Updated titles. Moved DrillExceptionHandler to FailureUtils. I can't make it private though because I need to do `Thread.currentThread().setUncaughtExceptionHandler(new FailureUtils.DrillExceptionHandler());` in the Drillbit class.

Please let me know if you have any other comments.

vrozov:
Move `Thread.currentThread().setUncaughtExceptionHandler()` inside static  `setUncaughtExceptionHandler` method of`FailureUtils`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -335,8 +336,9 @@ private void javaPropertiesToSystemOptions() {
    * Shutdown hook for Drillbit. Closes the drillbit, and reports on errors that
    * occur during closure, as well as the location the drillbit was started from.
    */
-  private static class ShutdownThread extends Thread {
+  public static class ShutdownThread extends Thread {
     private final static AtomicInteger idCounter = new AtomicInteger(0);
+    private static volatile boolean unrecoverableFailure;
```
vrozov: Consider moving `unrecoverableFailure` to `FailureUtils.java` and keeping `ShutdownThread` as `private`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/FailureUtils.java
```diff
@@ -0,0 +1,91 @@
...
+    // Don't instantiate
+  }
+
+  public static boolean isDirectMemoryOOM(Throwable e) {
+    if (e instanceof OutOfDirectMemoryError || e instanceof OutOfMemoryException) {
+      // These are always direct memory errors
+      return true;
+    }
+
+    if (e instanceof OutOfMemoryError) {
```
vrozov: `return (e instanceof OutOfMemoryError) && DIRECT_MEMORY_OOM_MESSAGE.equals(e.getMessage())`?

ilooner: done

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
ilooner: done

ilooner:
@vrozov Applied review comments. Please let me know if you have anymore comments.

exec/java-exec/src/main/java/org/apache/drill/exec/server/FailureUtils.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    public static void handleException(Throwable e) {
+      if (isHeapOOM(e)) {
+        // We can't recover cleanly from these.
+        unrecoverableFailure(e, "Unable to handle heap OOM.", EXIT_CODE_HEAP_OOM);
+      } else if (isDirectMemoryOOM(e)) {
+        // It's a problem, but we might be able to recover when we free memory.
+        UserException.memoryError(e).build(logger);
+      } else if (e instanceof Error) {
```
vrozov: Note that `AssertionError` is an instance of an `Error` and it triggers `unrecoverableFailure()` that always bypass Drillbit close sequence. Is it intentional?

exec/java-exec/src/main/java/org/apache/drill/exec/server/FailureUtils.java
```diff
@@ -0,0 +1,97 @@
...
+    if (e instanceof OutOfDirectMemoryError || e instanceof OutOfMemoryException) {
+      // These are always direct memory errors
+      return true;
+    }
+
+    return (e instanceof OutOfMemoryError) && DIRECT_MEMORY_OOM_MESSAGE.equals(e.getMessage());
+  }
+
+  public static boolean isHeapOOM(Throwable e) {
+    if (e instanceof OutOfMemoryError) {
```
vrozov: The same as line 46.

ilooner: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/FailureUtils.java
ilooner: @vrozov I didn't think of this. What do you think the right way to handle it is? In some cases I think we could still gracefully shutdown after an assert fails. But in other cases an AssertionError may occur because the Drillbit's state is corrupt and it cannot reliably perform a graceful shutdown.

priteshm:
Marked as ready-to-commit since the comments are addressed.

vrozov:
There is still one open question and I also need to clarify one additional issue. Consider a scenario where a user tries to shutdown a drillbit and this drillbit is a foreman for a query. A graceful shutdown will be initiated, but due to running query, shutdown sequence will wait for the query to complete. Now suppose that the query runs out of heap. Will not that still lead to a deadlock or an attempt to close the Drillbit?

ilooner:
@vrozov Good point. I think there are two possible approaches to fixing this (possible more):

  1. If we are in the middle of the shutdown and we experience and OOM we give up and forcibly terminate with System.halt
  2. We interrupt the shtudown thread and add logic to check for fatal errors during graceful shutdown. If a fatal error is detected during graceful shutdown we can stop waiting for queries and finish.

Regardless of the approach we take, we should resolving this corner case as a separate jira https://issues.apache.org/jira/browse/DRILL-6497. The original issue that users experienced was a Drillbit hanging when an OOM was encountered during normal operation, not during shutdown. Also looking at the stacktrace that users provided, this PR fixes the issue in the problematic code path. I think it's a good step forward. And as we discussed previously there are many other issues with child threads not properly propagating exceptions and errors, so there much more work do be done in this area that is outside the scope of this PR.

vrozov:
OK. +1

ilooner:
@vrozov Thanks for catching this. I'll remove the UncaughtExceptionHandler since it is a noop, perhaps we can resurrect it for the work in https://issues.apache.org/jira/browse/DRILL-6484.

PR #1309, opened 2018-06-07
DRILL-6477: Drillbit crashes with OOME (Heap) for a large WebUI query

kkhatua:
For queries submitted through the WebUI and retrieving a large result-set, the Drillbit often hangs or crashes due to the (foreman) Drillbit running out of Heap memory.

This is because the Web client translates the result set into a massive object in the heap-space and tries to send that back to the browser. This results in the VM thread actively trying to perform GC if the memory is not sufficient.

The workaround is to have the active webConnection of the query periodically timeout to allow for checking the consumed heap-space. A level of 0.85 (i.e. 85%) is set as default threshold, crossing which, a query submitted through the REST API is marked and failed. 
In addition, a user exception is thrown, indicting the cause of the query failing, along with alternative suggestions for re-executing the query.

This is the example of a query that tried to scan an entire 60M row table through the browser. The query failed with the following error message:
image

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -204,6 +204,8 @@ public final class ExecConstants {
   public static final String SERVICE_KEYTAB_LOCATION = SERVICE_LOGIN_PREFIX + ".keytab";
   public static final String KERBEROS_NAME_MAPPING = SERVICE_LOGIN_PREFIX + ".auth_to_local";
 
+  /* Provide resiliency on web server for queries submitted via HTTP */
+  public static final String HTTP_QUERY_FAIL_LOW_HEAP_THRESHOLD = "drill.exec.http.query.fail.low_heap.threshold";
 
   public static final String USER_SSL_ENABLED = "drill.exec.security.user.encryption.ssl.enabled";
   public static final String BIT_ENCRYPTION_SASL_ENABLED = "drill.exec.security.bit.encryption.sasl.enabled";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -204,9 +204,6 @@ public final class ExecConstants {
   public static final String SERVICE_KEYTAB_LOCATION = SERVICE_LOGIN_PREFIX + ".keytab";
   public static final String KERBEROS_NAME_MAPPING = SERVICE_LOGIN_PREFIX + ".auth_to_local";
 
-  /* Provide resiliency on web server for queries submitted via HTTP */
-  public static final String HTTP_QUERY_FAIL_LOW_HEAP_THRESHOLD = "drill.exec.http.query.fail.low_heap.threshold";
-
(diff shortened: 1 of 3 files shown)


kkhatua:
@parthchandra could you please review this?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -68,8 +78,37 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
...
+      }
+    } while (!isComplete && !nearlyOutOfHeapSpace);
+
+    //Fail if nearly out of heap space
+    if (nearlyOutOfHeapSpace) {
+      workManager.getBee().getForemanForQueryId(queryId)
+        .addToEventQueue(QueryState.FAILED,
+            UserException.resourceError(
+                new Throwable(
+                    "Query submitted through the Web interface was failed due to diminishing free heap memory ("+ Math.floor(((1-usagePercent)*100)) +"% free). "
```
parthchandra: We could make this friendlier :)
"There is not enough heap memory to run this query using the web interface. Please try a query with fewer columns or with a filter or limit condition to limit the data returned. You can also try an ODBC/JDBC client"

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -204,6 +204,8 @@ private ExecConstants() {
   public static final String SERVICE_KEYTAB_LOCATION = SERVICE_LOGIN_PREFIX + ".keytab";
   public static final String KERBEROS_NAME_MAPPING = SERVICE_LOGIN_PREFIX + ".auth_to_local";
 
+  /* Provide resiliency on web server for queries submitted via HTTP */
+  public static final String HTTP_QUERY_FAIL_LOW_HEAP_THRESHOLD = "drill.exec.http.query.fail.low_heap.threshold";
```
parthchandra: I would have just put this as a constant in the QueryWrapper class. I don't expect the user to ever modify this and when the QueryWrapper is updated to address large result sets, then we don't have to worry about removing this.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
kkhatua: I thought it would be useful to show the level at which the free memory was prior to cancellation. I suspect that people will see a GC'ed Drillbit after the cancellation and wonder why Drill is complaining of no sufficient heap.  Hence, the wording.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
kkhatua: I am inclined towards having this as an drill-override.conf property, just so that we have a tuning mechanism. But 85% is a reasonable threshold, so I'll hard-code it as a constant within QueryWrapper.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
parthchandra: Oh yes. That would be useful information to add to the message. But since the error is triggered at 0.85, the number would always be 15% :)

kkhatua:
@parthchandra  updated the PR based on your review.

PR #1311, opened 2018-06-08
DRILL-6479: Support EMIT for the Hash Aggr

Ben-Zvi:
The EMIT IterOutput functions both like OK and like NONE (w/o the final cleanup). Most of the changes for Hash-Aggr involve adapting that operator's state-machine for the new EMIT state -- first set a special local flag "handleEmit" to remember that EMIT was seen, then start working like an OK (i.e., process the current incoming), and afterwards start working like a NONE - i.e., return the results downstream.

  In several code places the special flag needs to be checked: 
(1) Do not set "build complete".
(2) Produce output in the batch (similar to "build complete").
(3) After emitting, need to get the next incoming batch (in doWork() ).
(4) After handling like OK, switch to like NONE.
(5) When needs to return EMIT, instead of OK.

Also the cases of empty input are handled. In outputCurrentBatch(), create and return an empty incoming batch. Also delayedSetup() may need to be called later if the initial input (before first EMITs) was empty.

Two unit tests were added: One with regular group by, and the other using distinct (and null input).


sohami:
@Ben-Zvi - There is a test failure. Can you please look into it ? 

```
Results :

Tests in error: 
  TestE2EUnnestAndLateral.testMultipleBatchesLateral_WithHashAgg:281 Â»  at posit...

Tests run: 2363, Failures: 0, Errors: 1, Skipped: 103
```

Ben-Zvi:
The test failure was a result of assuming that "ordered" would always work; changed the test to use the more conservative choice ( "unOrdered" ).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -573,6 +574,26 @@ public AggOutcome doWork() {
         delayedSetup();
       }
 
+      // If the prior call finished handling a prior incoming EMIT, need to call next() again
+      if ( !handleEmit && outcome == IterOutcome.EMIT ) {
```
sohami: As discussed please try following:
1) Try to remove this code as its a duplicate and you will end up handling all the IterOutcomes again. Instead try setting `currentBatchRecordCount` to 0 once you are done processing all the incoming records. That way this code can be totally removed.
2) Please add few unit tests
3) Add an exception in case spilling happens in EMIT scenario.

Ben-Zvi:
@sohami - (1) Removed the duplicate code. (2) Added **TestHashAggEmitOutcome** (basically a rework of *TestTopNEmitOutcome* ) .
As for (3) - will try to catch this condition in the planner, once there is support for "am I between a Lateral and an Unnest". No point checking this at runtime (after every EMIT).

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/agg/TestHashAggEmitOutcome.java
```diff
@@ -0,0 +1,789 @@
...
+
+  private void decodeAndAddValue(Object currentValue, List<String> listToAdd) {
+    if (currentValue == null) {
+      listToAdd.add("null");
+    } else if (currentValue instanceof byte[]) {
+      listToAdd.add(new String((byte[]) currentValue));
+    } else {
+      listToAdd.add(currentValue.toString());
+    }
+  }
```
sohami: Actually I just found out that all the above functions is not needed instead we can use below methods for verification. I am also updating TopN tests to use it.
```
RowSet actualRowSet = DirectRowSet.fromContainer(topNBatch.getContainer());
new RowSetComparison(expectedRowSet).verify(actualRowSet);
```

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/agg/TestHashAggEmitOutcome.java
```diff
@@ -0,0 +1,789 @@
...
+   * Verifies that if HashAggr receives multiple non-empty record batch with EMIT outcome in between then it produces
+   * output for those input batch correctly. In this case it receives first non-empty batch with OK_NEW_SCHEMA in
+   * buildSchema phase followed by an empty batch with EMIT outcome. For this combination it produces output for the
+   * record received so far along with EMIT outcome. Then it receives second non-empty batch with OK outcome and
+   * produces output for it differently. The test validates that for each output received the order of the records are
+   * correct.
+   * @throws Exception
+   */
+  @Test
+  public void testHashAggrMultipleOutputBatchWithLowerLimits() {
```
sohami: Please correct the test name there is no limit in case of HashAgg :)

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/agg/TestHashAggEmitOutcome.java
```diff
@@ -0,0 +1,789 @@
...
+    assertTrue(haBatch.next() == RecordBatch.IterOutcome.NONE);
+
+    nonEmptyInputRowSet2.clear();
+  }
+
+  /**
+   *
+   */
+  @Test
+  public void testHashAggrMultipleInputToMultipleOutputBatch_LowerLimits() {
```
sohami: Same as above to change test name

sohami:
Thanks for adding the tests. 

Regarding 3: I think the idea for catching Spill condition was moved to planner always generating 1 phase HashAgg when used inside lateral/unnest plan. And since 1 phase doesn't support spilling it will always error out in that condition. But shouldn't we add in the operator the check that if it's 2 phase and EMIT is seen then throw proper error ? The point is we should not fully rely on planner and execution should have its own check for error conditions.

amansinha100:
I agree that the operator should not have to perform the EMIT-2-Phase check.  The performance impact may be small but why even incur that.   Secondly, it would mask an incorrect plan that has been generated.  This is not a sub-optimal plan created due to statistics mis-estimates but rather an incorrect plan.

vdiravka:
This causes unit test failures:
`Results :

Tests in error: 
  TestHashAggrSpill.testNoPredictHashAggrSpill » UserRemote SYSTEM ERROR: IndexO...
  TestHashAggrSpill.testHashAggrSecondaryTertiarySpill » UserRemote SYSTEM ERROR...
  TestHashAggrSpill.testSimpleHashAggrSpill » UserRemote SYSTEM ERROR: IndexOutO...

Tests run: 3211, Failures: 0, Errors: 3, Skipped: 156`


https://issues.apache.org/jira/browse/DRILL-6479?focusedCommentId=16520713&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16520713

sohami:
LGTM from code changes. But please make sure all the pre-commit tests are passing.

PR #1313, opened 2018-06-08
DRILL-6474: Don't use TopN when order by and offset are used without a limit specified.

ilooner:
See jira for problem and solution

https://issues.apache.org/jira/browse/DRILL-6474

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLimitPlanning.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLimitPlanning.java
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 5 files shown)


ilooner:
@amansinha100 @HanumathRao @gparai Please review

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLimitPlanning.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.Test;
+
+public class TestLimitPlanning extends PlanTestBase {
+
+  // DRILL-6474
+  @Test
+  public void dontPushdownIntoTopNWhenNoLimit() throws Exception {
+    String query = "select full_name from cp.`employee.json` order by full_name offset 10";
+
+    PlanTestBase.testPlanMatchingPatterns(query, new String[]{}, new String[]{".*TopN\\(.*"});
```
amansinha100: Just to be sure, can you include 'Sort' in the inclusion string of plan matching.

ilooner: Done

PR #1316, opened 2018-06-11
DRILL-6486: BitVector split and transfer does not work correctly for non byte-multiple transfer lengths

bitblender:
Fix for the bug in BitVector splitAndTransfer. The logic for handling copy of last-n bits was incorrect for none byte-multiple transfer lengths. This PR fixes this problem and adds new tests which simulate the failing condition and also exercise other changes in this PR.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestValueVector.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestValueVector.java
@@ -573,8 +573,7 @@ the interface to load has changed
       assertEquals(1, accessor.get(1));
 
       // Ensure unallocated space returns 0
-      // Not a valid check since bounds checking is enabled
-     //assertEquals(0, accessor.get(3));
+      assertEquals(0, accessor.get(3));
     }
   }
 
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestValueVector.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestValueVector.java
@@ -548,6 +548,7 @@ the interface to load has changed
       m.set(1, 0);
       m.set(100, 0);
       m.set(1022, 1);
+      m.setValueCount(1023);
 
       final BitVector.Accessor accessor = vector.getAccessor();
       assertEquals(1, accessor.get(0));
@@ -560,17 +561,20 @@ the interface to load has changed
       m.set(0, 1);
       m.set(1, 0);
       m.set(1, 0);
+      m.setValueCount(2);
       assertEquals(1, accessor.get(0));
       assertEquals(0, accessor.get(1));
 
       // test toggling the values
       m.set(0, 0);
       m.set(1, 1);
+      m.setValueCount(2);
       assertEquals(0, accessor.get(0));
       assertEquals(1, accessor.get(1));
 
       // Ensure unallocated space returns 0
-      assertEquals(0, accessor.get(3));
+      // Not a valid check since bounds checking is enabled
+     //assertEquals(0, accessor.get(3));
     }
   }
 
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
@@ -119,8 +119,7 @@ public class TestSplitAndTransfer {
 
   int getBit(TestBitPattern pattern, int index) {
     if (pattern == TestBitPattern.RANDOM) {
-      final int randomBit = (int) (Math.random() * 2);
-      return randomBit;
+      return (int) (Math.random() * 2);
     }
     return (pattern == TestBitPattern.ALTERNATING) ? (index % 2) : ((pattern == TestBitPattern.ONE) ? 1 : 0);
   }
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
@@ -167,7 +167,7 @@ public class TestSplitAndTransfer {
       final int start = startLength[0];
       final int length = startLength[1];
       tp.splitAndTransfer(start, length);
-      newBitVector.getMutator().setValueCount(length);
+      assertEquals(newBitVector.getAccessor().getValueCount(), length);
       for (int i = 0; i < length; i++) {
         final int expectedValue = compareArray[start + i];
         assertEquals(expectedValue, accessor.get(i));
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
@@ -17,10 +17,6 @@
  */
 package org.apache.drill.exec.vector;
 
-import static org.junit.Assert.assertArrayEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertTrue;
-
 import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.common.types.TypeProtos.MinorType;
 import org.apache.drill.common.types.Types;
@@ -30,6 +26,12 @@ import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.TransferPair;
 import org.apache.drill.exec.vector.NullableVarCharVector.Accessor;
 import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertArrayEquals;
+
 
 public class TestSplitAndTransfer {
   @Test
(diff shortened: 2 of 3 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
```diff
@@ -77,4 +79,84 @@ public void test() throws Exception {
...
+
+    testBitVectorImpl(3448, new int[][] {{0, 2048}, {2048, 1400}}, TestBitPattern.ZERO);
+    testBitVectorImpl(3448, new int[][] {{0, 2048}, {2048, 1400}}, TestBitPattern.ONE);
+    testBitVectorImpl(3448, new int[][] {{0, 2048}, {2048, 1400}}, TestBitPattern.ALTERNATING);
+    testBitVectorImpl(3448, new int[][] {{0, 2048}, {2048, 1400}}, TestBitPattern.RANDOM);
+  }
+
+  int getBit(TestBitPattern pattern, int index) {
+    if (pattern == TestBitPattern.RANDOM) {
+      final int randomBit = (int) (Math.random() * 2);
```
HanumathRao: I was just thinking if bitValue is better than randomBit?

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
```diff
@@ -285,35 +285,55 @@ public void transferTo(BitVector target) {
 
   public void splitAndTransferTo(int startIndex, int length, BitVector target) {
     assert startIndex + length <= valueCount;
-    int firstByte = getByteIndex(startIndex);
-    int byteSize = getSizeFromCount(length);
-    int offset = startIndex % 8;
-    if (offset == 0) {
+    int firstByteIndex = getByteIndex(startIndex);//byte offset of the first src byte
+    int bytesToRead = getSizeFromCount(length); //src bytes to read (including start/end bytes that might not be fully copied)
```
HanumathRao: bytesToRead is also being used for setting target byte so is it better to use numOfBytes or something like that?

HanumathRao:
@bitblender  Thank you for fixing the bug in this complex code. I have only few variable naming comments. LGTM.

paul-rogers:
Have we considered replacing BitVector with UINT1 vector? This swap was done in most places in Drill before 1.0, probably because BitVector was buggy and slow. Perhaps we should just remove it from the remaining places?

Or, if we get it to work, consider using it where it was originally intended, as the "bits vector" in Nullable types? If we do that, we have cross-version compatibility to worry about with old clients...

bitblender:
@paul-rogers  I have opened https://issues.apache.org/jira/browse/DRILL-6493 to follow up on your suggestion. Until that gets prioritized and completed, the changes in this PR will be necessary to make sure that BitVector splitAndTranfer() works correctly.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
bitblender: Agreed. Changing it to 'numBytesHoldingSourceBits'

exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
bitblender: The variable has been removed.

ilooner:
+1 LGTM Thanks for adding thorough unit tests.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
```diff
@@ -285,35 +285,55 @@ public void transferTo(BitVector target) {
...
+    int numBytesHoldingSourceBits = getSizeFromCount(length); //src bytes to read (including start/end bytes that might not be fully copied)
+    int firstBitOffset = startIndex % 8; //Offset of first src bit within the first src byte
+    if (firstBitOffset == 0) {
       target.clear();
       // slice
       if (target.data != null) {
         target.data.release();
       }
-      target.data = data.slice(firstByte, byteSize);
+      target.data = data.slice(firstByteIndex, numBytesHoldingSourceBits);
```
sohami: I think this is true if `(firstBitOffset == 0 && length%8 == 0)` `if length%8 !=0` then we are keeping reference of more bits then requested. But as memory for underlying Drillbuf is always allocated at byte level then I guess it doesn't make any difference even if we copy extra bits.

However in target vector we are setting the `valueCount` same as `length` but it's not being used while getting any value from Vector (or in `get` apis). I think we should put that check as bounds checking. The internal Drillbuf bounds checking will not catch it since length of buffer is larger than actual data which will allow accessing a value beyond `ValueCount`.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
```diff
@@ -285,35 +285,55 @@ public void transferTo(BitVector target) {
...
+      //Copying the last n bits
+
+      //Copy length is not a byte-multiple
       if (length % 8 != 0) {
-        target.data.setByte(byteSize - 1, ((this.data.getByte(firstByte + byteSize - 1) & 0xFF) >>> offset));
+        byte lastButOneByte = 0;
+        byte bitsFromLastButOneByte = 0;
+        // if start is not byte aligned then we have to copy some bits from the last full byte read in the
+        // previous loop
+        if (firstBitOffset != 0) {
```
sohami: this will always be true here.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
bitblender: Good point. I have added a check.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
bitblender: Yes. Removed the check.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
```diff
@@ -285,35 +285,55 @@ public void transferTo(BitVector target) {
...
+          lastButOneByte = byteIPlus1;
+          bitsFromLastButOneByte = (byte)((lastButOneByte & 0xFF) >>> firstBitOffset);
+        }
+        final int lastByte = this.data.getByte(firstByteIndex + numBytesHoldingSourceBits);
```
sohami: This can go beyond the bound of the data buffer. Consider a case where there are only 3 bytes in SrcVector which will contain bits from 0 to 23. Now in target we want to put bits from index 5 and length 17 which means last bit to be read is 21st from src vector. `firstByteIndex` will be 0 and `numBytesHoldingSourceBits` will be 3 so it will end up calling `getByte` for index 3 which is not present.

bitblender:
@sohami I have fixed the last-byte issue. I removed the bounds check because some of the operators are hitting the IOB, most likely because they are not setting valueCount correctly. Those issues are best fixed in a different PR, so I am removing that check.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
```diff
@@ -328,7 +328,9 @@ public void splitAndTransferTo(int startIndex, int length, BitVector target) {
         lastButOneByte = byteIPlus1;
         bitsFromLastButOneByte = (byte)((lastButOneByte & 0xFF) >>> firstBitOffset);
 
-        final int lastByte = this.data.getByte(firstByteIndex + numBytesHoldingSourceBits);
+
+        final int lastByte = (8 - firstBitOffset >= length % 8) ?
+                0 : this.data.getByte(firstByteIndex + numBytesHoldingSourceBits);
```
sohami: Check looks good. Can you please add the motivation of this check in the comment. Based on my understanding `length%8` gives the number of bits to be copied from last byte and `(8-firstBitOffset)` is the bits left to be copied from `lastButOneByte`. So if bits left to be copied is smaller than bits needed from last byte then we need to look at next byte in those cases.

exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestValueVector.java
```diff
@@ -560,17 +561,20 @@ public void testBitVector() {
...
       m.set(0, 0);
       m.set(1, 1);
+      m.setValueCount(2);
       assertEquals(0, accessor.get(0));
       assertEquals(1, accessor.get(1));
 
       // Ensure unallocated space returns 0
-      assertEquals(0, accessor.get(3));
+      // Not a valid check since bounds checking is enabled
+     //assertEquals(0, accessor.get(3));
```
sohami: since you have removed the bounds check let's uncomment this assert. Also comment says unallocated space shouldn't it be unassigned index instead ?

exec/java-exec/src/test/java/org/apache/drill/exec/vector/TestSplitAndTransfer.java
```diff
@@ -77,4 +79,102 @@ public void test() throws Exception {
...
+
+    final TransferPair tp = bitVector.getTransferPair(allocator);
+    final BitVector newBitVector = (BitVector) tp.getTo();
+    final BitVector.Accessor accessor = newBitVector.getAccessor();
+
+    for (final int[] startLength : startLengths) {
+      final int start = startLength[0];
+      final int length = startLength[1];
+      tp.splitAndTransfer(start, length);
+      newBitVector.getMutator().setValueCount(length);
```
sohami: instead of `setValueCount` in test I think we should check if `splitAndTransfer` is setting it correctly or not ?

bitblender:
Yes.



On June 21, 2018 at 10:25:44 AM, Pritesh Maker (notifications@github.com<mailto:notifications@github.com>) wrote:

@bitblender<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_bitblender&d=DwMCaQ&c=cskdkSMqhcnjZxdQVpwTXg&r=HlugibuI4IVjs-VMnFvNTcaBtEaDDqE4Ya96cugWqJ8&m=hlQMTw-fCidZaB1T0_96G3hUXi6jCqarmxAWz-qfT2M&s=sTmHTnIdq5BE9dp8wuPzCyuLG4B7r8-xKX4p4f8_LX4&e=> would you be able to address the remaining review comments in time for the batch commit this week?

bitblender:
@sohami  Addressed all your review comments. Please take a look again.

PR #1319, opened 2018-06-13
DRILL-6212: Prevent recursive cast expressions

gparai:
@vrozov / @amansinha100  please review the PR. Thanks!

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
@@ -17,22 +17,37 @@
  */
 package org.apache.drill.exec.planner.logical;
 
-
-import org.apache.calcite.plan.Convention;
-import org.apache.calcite.plan.ConventionTraitDef;
+import org.apache.calcite.plan.RelOptRule;
 import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelOptUtil;
+import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Project;
-import org.apache.calcite.rel.rules.ProjectMergeRule;
+import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.tools.RelBuilder;
+import org.apache.calcite.util.Permutation;
 import org.apache.drill.exec.expr.fn.FunctionImplementationRegistry;
 import org.apache.calcite.rel.core.RelFactories.ProjectFactory;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
 import org.apache.drill.exec.planner.DrillRelBuilder;
 
-public class DrillMergeProjectRule extends ProjectMergeRule {
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Rule for merging two projects provided the projects aren't projecting identical sets of
+ * input references.
+ *
+ * NOTE: This rules does not extend the Calcite ProjectMergeRule
+ * because of CALCITE-2223. Once, fixed this rule be changed accordingly. Please see DRILL-6501.
+ */
+public class DrillMergeProjectRule extends RelOptRule {
 
   private FunctionImplementationRegistry functionRegistry;
   private static DrillMergeProjectRule INSTANCE = null;
+  private final boolean force;
 
   public static DrillMergeProjectRule getInstance(boolean force, ProjectFactory pFactory,
       FunctionImplementationRegistry functionRegistry) {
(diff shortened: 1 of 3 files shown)


priteshm:
Since @HanumathRao opened the bug, maybe he can also review it?

vrozov:
I am not familiar with the functionality to review functional changes, java code LGTM.

HanumathRao:
@gparai  Code changes look good to me.

ilooner:
@gparai It looks like this change copies the contents of ProjectMergeRule.onMatch into DrillMergeProjectRule. Can we avoid doing that?

Also can you add a test that reproduces the original issue that you were seeing.

vrozov:
@ilooner There is a call to `simplifyCast()` in the replaced implementation of `ProjectMergeRule.onMatch()`.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
```diff
@@ -77,4 +85,100 @@ private boolean checkComplexOutput(Project project) {
...
+          operand = ((RexCall) rex).getOperands().get(0);
+        }
+
+      }
+      list.add(rex);
+    }
+    return list;
+  }
+
+  public static Project replace(Project topProject, Project bottomProject) {
```
vrozov: Where is this method used?

gparai: It is unnecessary. Removed it.

gparai:
@ilooner / @vrozov I have addressed your review comments. Please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
```diff
@@ -77,4 +84,79 @@ private boolean checkComplexOutput(Project project) {
     }
     return false;
   }
+
+  @Override
+  public void onMatch(RelOptRuleCall call) {
+    final Project topProject = call.rel(0);
```
ilooner: I'm still a little concerned about this change because lines 90 - 115 are copied from ProjectMergeRule, lines 119 - 124 are new, and lines 126 - 142 are logically equivalent except for the added call to simplifyCast. Considering the similarity of the code and the fact that ProjectMergeRule should do this logic correctly, I have a few questions.

Is this a bug in Calcite ProjectMergeRule or in Drill? If it is a bug in Calcite, shouldn't we be contributing the fix there? If it is a Drill specific bug why is the ProjectMergeRule provided by Calcite not sufficient and why does it cause issues for us when a cast is used?

gparai: Please see https://issues.apache.org/jira/browse/CALCITE-2223. This fix is meant to be in temporarily until the right fix is agreed upon and implemented in Calcite.

ilooner: @gparai I'm okay with the temporary fix since it has been battle tested with users, but could you  create a Jira to for undoing the changes once the fix in Calcite is made. Also please include some details in the jira about what would need to be done, and add a note in the javadoc of DrillMergeProjectRule pointing to CALCITE-2223 and the Drill Jira. I'm guessing that we would be able to extend from ProjectMergeRule again when the fix for CALCITE-2223 is available?

While it's good to have this fixed in Drill in the short term, in the long term it is not sustainable to accumulate custom fixes for Calcite in Drill. Looking at CALCITE-2223 they haven't settled on a solution yet, and there hasn't been any activity there for two months. I think you should propose this solution for CALCITE-2223 and open a PR against Calcite as well.

gparai: @ilooner thanks for the comment. I have create DRILL-6501 for the same.

exec/java-exec/src/test/java/org/apache/drill/TestProjectWithFunctions.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill;
+
+// Test the optimizer plan in terms of projecting different functions e.g. cast
```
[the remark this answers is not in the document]
gparai: Done

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
gparai: Done

exec/java-exec/src/test/java/org/apache/drill/TestProjectWithFunctions.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.drill.categories.PlannerTest;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(PlannerTest.class)
+public class TestProjectWithFunctions extends PlanTestBase {
+
+  @Test
+  public void testCastFunctions() throws Exception {
+    test("use dfs.tmp");
```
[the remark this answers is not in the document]
gparai: Test is changed so not required.

gparai:
@ilooner / @vrozov I have addressed the review comments. Please take a look.

ilooner:
@gparai Thanks for making changes LGTM +1.
@vrozov Please let us know if things look good to you.

PR #1320, opened 2018-06-13
DRILL-6491: Prevent merge join for full outer join at planning stage

KazydubB:


vvysotskyi:
Could you please add unit tests to check that plan with full outer merge join will be omitted for the case when it was chosen before your changes? 

Also, it would be fine to add more detailed description of the problem into the Jira.

KazydubB:
Done.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
```diff
@@ -255,6 +260,42 @@ public void testDrill4196() throws Exception {
...
+
+    String query = "select * " +
+        " from (select employee_id from cp.`employee.json` order by employee_id) e1 " +
+        " full outer join (select employee_id from cp.`employee.json` order by employee_id) e2 " +
+        " on e1.employee_id = e2.employee_id " +
+        " limit 10";
+    try {
+      testPlanMatchingPatterns(query, new String[] {"HashJoin"}, new String[] {"MergeJoin"});
+    } finally {
+      test("ALTER SESSION RESET ALL");
```
arina-ielchiieva: Reset only options, you have set.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
```diff
@@ -255,6 +260,42 @@ public void testDrill4196() throws Exception {
...
+    }
+  }
+
+  @Test // DRILL-6491
+  public void testFullJoinIsNotSupported() throws Exception {
+    thrown.expect(UserRemoteException.class);
+    thrown.expectMessage(CoreMatchers.containsString("SYSTEM ERROR: CannotPlanException"));
```
arina-ielchiieva: Declare thrown right before the query.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
```diff
@@ -255,6 +260,42 @@ public void testDrill4196() throws Exception {
...
+
+    String query = "select * " +
+        " from (select employee_id from cp.`employee.json` order by employee_id) e1 " +
+        " full outer join (select employee_id from cp.`employee.json` order by employee_id) e2 " +
+        " on e1.employee_id = e2.employee_id " +
+        " limit 10";
+    try {
+      test(query);
+    } finally {
+      test("ALTER SESSION RESET ALL");
```
arina-ielchiieva: Do not reset all sessions options.

arina-ielchiieva: Do we need finally if we have thrown?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
KazydubB: What if default values for the (explicitly) set options are changed in the future?

arina-ielchiieva: Use ` resetSessionOption` method.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
```diff
@@ -255,6 +260,42 @@ public void testDrill4196() throws Exception {
       .go();
   }
 
+  @Test // DRILL-6491
+  public void testMergeIsNotSelectedForFullJoin() throws Exception {
+    setSessionOption(PlannerSettings.HASHJOIN.getOptionName(), true);
```
arina-ielchiieva: Set options in try so they will be reset in finally.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
KazydubB: Yes, we still need it as the Exception is thrown for real (though, expected).

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
KazydubB: Done.

arina-ielchiieva: I meant right before executing the query.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
```diff
@@ -255,6 +260,41 @@ public void testDrill4196() throws Exception {
...
+    String query = "select * " +
+        " from (select employee_id from cp.`employee.json` order by employee_id) e1 " +
+        " full outer join (select employee_id from cp.`employee.json` order by employee_id) e2 " +
+        " on e1.employee_id = e2.employee_id " +
+        " limit 10";
+    try {
+      setSessionOption(PlannerSettings.HASHJOIN.getOptionName(), true);
+      testPlanMatchingPatterns(query, new String[] {"HashJoin"}, new String[] {"MergeJoin"});
+    } finally {
+      resetSessionOption(PlannerSettings.HASHJOIN.getOptionName());
```
vvysotskyi: This option is set in `@BeforeClass` method, but here you are resetting it to the default value. It may cause failures for other tests in the class which depends on the value of this option.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestMergeJoinAdvanced.java
```diff
@@ -255,6 +260,41 @@ public void testDrill4196() throws Exception {
...
+    thrown.expect(UserRemoteException.class);
+    thrown.expectMessage(CoreMatchers.containsString("SYSTEM ERROR: CannotPlanException"));
+
+    String query = "select * " +
+        " from (select employee_id from cp.`employee.json` order by employee_id) e1 " +
+        " full outer join (select employee_id from cp.`employee.json` order by employee_id) e2 " +
+        " on e1.employee_id = e2.employee_id " +
+        " limit 10";
+    try {
+      setSessionOption(PlannerSettings.HASHJOIN.getOptionName(), false);
```
vvysotskyi: The same as above.

arina-ielchiieva:
Looks good, @vvysotskyi?

PR #1321, opened 2018-06-14
 DRILL-5188: Expand sub-queries using rules

vvysotskyi:
Before this change, sub-queries were rewritten in `SqlToRelConverter`. During such expansion, excessive aggregate calls were created. Due to these excessive aggregate calls, some rel nodes weren't able to transpose to Aggregate rel node and the resulting plan was inefficient and in the most cases with joins with true in condition. As discussed in [CALCITE-2158](https://issues.apache.org/jira/browse/CALCITE-2158), a better approach to expand sub-queries is to use `SubQueryRemoveRule` rules.

This PR contains two commits:
- the first one contains changes to use `SubQueryRemoveRule` rules instead of expanding in `SqlToRelConverter`.
- the second commit contains introducing single_value aggregate function which is used when filter contains a comparison of value and subquery and in some `SubQueryRemoveRule` rules.

### DIFF
--- a/exec/java-exec/src/main/codegen/templates/SingleValueAgg.java
+++ b/exec/java-exec/src/main/codegen/templates/SingleValueAgg.java
@@ -1,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+<@pp.dropOutputFile />
+
+<@pp.changeOutputFile name="/org/apache/drill/exec/expr/fn/impl/gaggr/SingleValueFunctions.java" />
+
+<#include "/@includes/license.ftl" />
+
+package org.apache.drill.exec.expr.fn.impl.gaggr;
+
+import org.apache.drill.exec.expr.DrillAggFunc;
+import org.apache.drill.exec.expr.annotations.FunctionTemplate;
+import org.apache.drill.exec.expr.annotations.FunctionTemplate.FunctionScope;
+import org.apache.drill.exec.expr.annotations.Output;
+import org.apache.drill.exec.expr.annotations.Param;
+import org.apache.drill.exec.expr.annotations.Workspace;
+import org.apache.drill.exec.expr.holders.*;
+
+import javax.inject.Inject;
+import io.netty.buffer.DrillBuf;
+
+/*
+ * This class is generated using freemarker and the ${.template_name} template.
+ */
+@SuppressWarnings("unused")
+public class SingleValueFunctions {
+<#list singleValue.types as type>
+
+  @FunctionTemplate(name = "single_value",
+                  <#if type.major == "VarDecimal">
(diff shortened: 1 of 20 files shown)


vvysotskyi:

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -478,7 +474,7 @@ public boolean isExplain() {
 
     @Override
     public boolean isExpand() {
-      return SqlToRelConverterConfig.DEFAULT.isExpand();
+      return false;
```
amansinha100: Does this change affect expansion of subqueries in the WHERE clause only or also in the FROM clause ? If it does affect the latter, we would want to test it..

exec/java-exec/src/test/java/org/apache/drill/TestDisabledFunctionality.java
```diff
@@ -47,28 +47,21 @@ private static void throwAsUnsupportedException(UserException ex) throws Excepti
...
-    try {
-      test("select n_name from cp.`tpch/nation.parquet` " +
-           "where n_nationkey = " +
-           "(select r_regionkey from cp.`tpch/region.parquet` " +
-           "where r_regionkey = 1)");
-    } catch(UserException ex) {
-      throwAsUnsupportedException(ex);
-    }
+  @Test
+  public void testComparisonWithSingleValueSubQuery() throws Exception {
```
amansinha100: Since this is now a positive test, you could move it out of the TestDisabledFunctionality class which is meant for unsupported operations.

vvysotskyi:
@amansinha100, I made single PR because sub-query transformation rules for some cases, which worked before without SINGLE_VALUE function, use this function. So in order to avoid blocking one PR by another, I created one PR with these changes.

exec/java-exec/src/main/codegen/templates/SingleValueAgg.java
```diff
@@ -0,0 +1,144 @@
...
+    </#if>
+
+    public void setup() {
+      nonNullCount = new BigIntHolder();
+      nonNullCount.value = 0;
+      value = new ${type.runningType}Holder();
+    }
+
+    @Override
+    public void add() {
```
amansinha100: My understanding of SINGLE_VALUE is that the output is single value, not the input.  For instance, consider the following query: 
   SELECT SINGLE_VALUE(a) FROM T GROUP BY b ; 
Here, suppose all values of 'b' are the same.  Then the SINGLE_VALUE(a) will produce only 1 value and should be allowed.

amansinha100: Actually,  I take that back.  SINGLE_VALUE does in fact require input being single value.  I am not sure if SINGLE_VALUE is permitted with a GROUP BY.

exec/java-exec/src/test/java/org/apache/drill/TestCorrelation.java
```diff
@@ -58,4 +58,50 @@ public void testScalarAggAndFilterCorrelatedSubquery() throws Exception {
...
+    testBuilder()
+        .sqlQuery(query)
+        .unOrdered()
+        .baselineColumns("employee_id")
+        .baselineValues(1L)
+        .go();
+  }
+
+  @Test
+  public void testSeveralExistsCorrelateSubquery() throws Exception {
```
amansinha100: This query pattern reminds me of TPC-H query 21, although that one is doing EXISTS and NOT-EXISTS.  Currently, that test is marked Ignored [1].  Would this PR also handle that case ? 

[1] https://github.com/apache/drill/blob/master/exec/java-exec/src/test/java/org/apache/drill/TestTpchDistributed.java#L141

amansinha100:
@vvysotskyi Fair enough to keep the two changes together in this PR.   For my understanding, can you pls provide an Explain plan for a sample query based on this PR ?

exec/java-exec/src/test/java/org/apache/drill/TestCorrelation.java
amansinha100: That's good to know and thanks for enabling those tests.  Depending on who is testing this enhancement, you could let them know that the functional regression tests also need to be updated for higher scale factors.

amansinha100:
Modified plan and code changes LGTM.  +1

PR #1322, opened 2018-06-15
DRILL-6487: Limit estimateRowCount should not return negative rowcount

gparai:
@amansinha100 please review the PR. Thanks!


amansinha100:
+1

PR #1323, opened 2018-06-15
DRILL-6498: Support for EMIT outcome in ExternalSortBatch

sohami:
@parthchandra / @Ben-Zvi - please help to review.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/PriorityQueueCopierWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/PriorityQueueCopierWrapper.java
@@ -34,6 +34,7 @@ import org.apache.drill.exec.ops.OperatorContext;
 import org.apache.drill.exec.physical.impl.xsort.managed.SortImpl.SortResults;
 import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.MaterializedField;
+import org.apache.drill.exec.record.RecordBatch;
 import org.apache.drill.exec.record.VectorAccessible;
 import org.apache.drill.exec.record.VectorAccessibleUtilities;
 import org.apache.drill.exec.record.VectorContainer;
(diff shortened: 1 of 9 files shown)


priteshm:
@Ben-Zvi, @parthchandra any chance of reviewing this soon?

sohami:
@Ben-Zvi / @parthchandra  - I have written a description in the JIRA to give idea about current implementation of Sort, motivation behind current approach for supporting EMIT outcome and what future enhancements can be done. Please let me know if it looks good.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/xsort/managed/PriorityQueueCopierWrapper.java
```diff
@@ -345,6 +348,33 @@ public void close() {
...
+      VectorContainer dataContainer = getContainer();
+      // First output batch, let's create the container.
+      if (container.getNumberOfColumns() == 0) {
+        for (VectorWrapper<?> vw : dataContainer) {
+          container.add(vw.getValueVector());
+        }
+        // In future when we want to support spilling with EMIT outcome then we have to create SV4 container all the
+        // time. But that will have effect of copying data again by SelectionVectorRemover from SV4 to SV_None. Other
+        // than that we have to send OK_NEW_SCHEMA each time. There can be other operators like StreamAgg in downstream
+        // as well, so we cannot have special handlnig in SVRemover for EMIT phase.
```
Ben-Zvi: (1) Typo: "handling"
(2) The usage of the External Sort between Unnest and Lateral is very limited (maybe only if under a Streaming-Aggr), thus the multiple OK_NEW_SCHEMA sounds like a good solution (and StreamAgg needs to be updated to match).

sohami: Fixed (1). For (2) it can also be used just for OrderBy ? The issue with producing OK_NEW_SCHEMA with each batch is that we don't handle this case properly in all the operators and hence there can be pre-mature failures of the query.

PR #1324, opened 2018-06-15
DRILL-6310: limit batch size for hash aggregate

ppadma:
batch sizing for hash aggregate is done by changing sizes of batches we are holding in hash aggregate partitions for aggr values and hash table for keys. Earlier, batch size was always 64K rows. Now, these batches will be sized for 16MB based on actual data. 
This means that the way we index the batches and rows with in the batch have to change. This is done by saving size of the batch in batchHolder. For given keys, based on hash value, sizes of the batches saved in batch holders is used to figure out the batch number and row with in the batch.  

Also, for figuring out sizing information for outgoing values columns, a new map which maintains mapping between input and output columns is added. 

This PR also has fix for 
DRILL-6499: No need to calculate stdRowWidth for every batch by default.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
@@ -19,15 +19,19 @@ package org.apache.drill.exec.physical.impl.aggregate;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 
 import com.google.common.collect.Lists;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.expression.ErrorCollector;
 import org.apache.drill.common.expression.ErrorCollectorImpl;
+import org.apache.drill.common.expression.FunctionCall;
 import org.apache.drill.common.expression.FunctionHolderExpression;
 import org.apache.drill.common.expression.IfExpression;
 import org.apache.drill.common.expression.LogicalExpression;
+import org.apache.drill.common.expression.SchemaPath;
 import org.apache.drill.common.logical.data.NamedExpression;
+import org.apache.drill.common.map.CaseInsensitiveMap;
 import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.compile.sig.GeneratorMapping;
 import org.apache.drill.exec.compile.sig.MappingSet;
@@ -49,11 +53,14 @@ import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.BatchSchema.SelectionVectorMode;
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.RecordBatch;
+import org.apache.drill.exec.record.RecordBatchMemoryManager;
+import org.apache.drill.exec.record.RecordBatchSizer;
 import org.apache.drill.exec.record.TypedFieldId;
 import org.apache.drill.exec.record.VectorWrapper;
 import org.apache.drill.exec.record.selection.SelectionVector2;
 import org.apache.drill.exec.record.selection.SelectionVector4;
 import org.apache.drill.exec.vector.AllocationHelper;
+import org.apache.drill.exec.vector.FixedWidthVector;
 import org.apache.drill.exec.vector.ValueVector;
 
 import com.sun.codemodel.JExpr;
@@ -71,6 +78,12 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
   private BatchSchema incomingSchema;
   private boolean wasKilled;
 
+  private int numGroupByExprs, numAggrExprs;
+
+  // This map saves the mapping between outgoing column and incoming column.
+  private Map<String, String> columnMapping;
+  private final HashAggMemoryManager hashAggMemoryManager;
+
   private final GeneratorMapping UPDATE_AGGR_INSIDE =
       GeneratorMapping.create("setupInterior" /* setup method */, "updateAggrValuesInternal" /* eval method */,
           "resetValues" /* reset */, "cleanup" /* cleanup */);
@@ -84,6 +97,67 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
           "htRowIdx" /* workspace index */, "incoming" /* read container */, "outgoing" /* write container */,
           "aggrValuesContainer" /* workspace container */, UPDATE_AGGR_INSIDE, UPDATE_AGGR_OUTSIDE, UPDATE_AGGR_INSIDE);
 
+  public int getOutputRowCount() {
+    return hashAggMemoryManager.getOutputRowCount();
+  }
+
+  public RecordBatchMemoryManager getRecordBatchMemoryManager() {
+    return hashAggMemoryManager;
+  }
+
+  private class HashAggMemoryManager extends RecordBatchMemoryManager {
+    private int valuesRowWidth = 0;
+
+    HashAggMemoryManager(int outputBatchSize) {
+      super(outputBatchSize);
+    }
+
+    @Override
+    public void update() {
+      // Get sizing information for the batch.
+      setRecordBatchSizer(new RecordBatchSizer(incoming));
+
+      int fieldId = 0;
+      int newOutgoingRowWidth = 0;
+      for (VectorWrapper<?> w : container) {
+        if (w.getValueVector() instanceof FixedWidthVector) {
+          newOutgoingRowWidth += ((FixedWidthVector) w.getValueVector()).getValueWidth();
+          if (fieldId >= numGroupByExprs) {
+            valuesRowWidth += ((FixedWidthVector) w.getValueVector()).getValueWidth();
+          }
+        } else {
+          int columnWidth;
+          if (columnMapping.get(w.getValueVector().getField().getName()) == null) {
+             columnWidth = TypeHelper.getSize(w.getField().getType());
+          } else {
+            RecordBatchSizer.ColumnSize columnSize = getRecordBatchSizer().getColumn(columnMapping.get(w.getValueVector().getField().getName()));
+            if (columnSize == null) {
+              columnWidth = TypeHelper.getSize(w.getField().getType());
+            } else {
+              columnWidth = columnSize.getAllocSizePerEntry();
+            }
+          }
+          newOutgoingRowWidth += columnWidth;
+          if (fieldId >= numGroupByExprs) {
+            valuesRowWidth += columnWidth;
+          }
+        }
+        fieldId++;
+      }
+
+      if (updateIfNeeded(newOutgoingRowWidth)) {
+        // There is an update to outgoing row width.
+        // un comment this if we want to adjust the batch row count of in flight batches.
+        // To keep things simple, we are not doing this adjustment for now.
+        // aggregator.adjustOutputCount(getOutputBatchSize(), getOutgoingRowWidth(), newOutgoingRowWidth);
+      }
+
+      updateIncomingStats();
+      if (logger.isDebugEnabled()) {
+        logger.debug("BATCH_STATS, incoming: {}", getRecordBatchSizer());
+      }
+    }
+  }
 
   public HashAggBatch(HashAggregate popConfig, RecordBatch incoming, FragmentContext context) {
     super(popConfig, context);
@@ -103,6 +177,13 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
 
     boolean allowed = oContext.getAllocator().setLenient();
     logger.debug("Config: Is allocator lenient? {}", allowed);
+
+    // get the output batch size from config.
+    int configuredBatchSize = (int) context.getOptions().getOption(ExecConstants.OUTPUT_BATCH_SIZE_VALIDATOR);
+    hashAggMemoryManager = new HashAggMemoryManager(configuredBatchSize);
+    logger.debug("BATCH_STATS, configured output batch size: {}", configuredBatchSize);
+
+    columnMapping = CaseInsensitiveMap.newHashMap();
   }
 
   @Override
@@ -135,6 +216,9 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
     }
     for (VectorWrapper<?> w : container) {
       AllocationHelper.allocatePrecomputedChildCount(w.getValueVector(), 0, 0, 0);
+    }
+    if (incoming.getRecordCount() > 0) {
+      hashAggMemoryManager.update();
     }
   }
 
@@ -239,8 +323,8 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
     // top.saveCodeForDebugging(true);
     container.clear();
 
-    int numGroupByExprs = (popConfig.getGroupByExprs() != null) ? popConfig.getGroupByExprs().size() : 0;
-    int numAggrExprs = (popConfig.getAggrExprs() != null) ? popConfig.getAggrExprs().size() : 0;
+    numGroupByExprs = (popConfig.getGroupByExprs() != null) ? popConfig.getGroupByExprs().size() : 0;
+    numAggrExprs = (popConfig.getAggrExprs() != null) ? popConfig.getAggrExprs().size() : 0;
     aggrExprs = new LogicalExpression[numAggrExprs];
     groupByOutFieldIds = new TypedFieldId[numGroupByExprs];
     aggrOutFieldIds = new TypedFieldId[numAggrExprs];
@@ -263,13 +347,13 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
 
       // add this group-by vector to the output container
       groupByOutFieldIds[i] = container.add(vv);
+      columnMapping.put(outputField.getName(), ne.getExpr().toString().replace('`',' ').trim());
     }
 
     int extraNonNullColumns = 0; // each of SUM, MAX and MIN gets an extra bigint column
     for (i = 0; i < numAggrExprs; i++) {
       NamedExpression ne = popConfig.getAggrExprs().get(i);
-      final LogicalExpression expr =
-          ExpressionTreeMaterializer.materialize(ne.getExpr(), incoming, collector, context.getFunctionRegistry());
+      final LogicalExpression expr = ExpressionTreeMaterializer.materialize(ne.getExpr(), incoming, collector, context.getFunctionRegistry());
 
       if (expr instanceof IfExpression) {
         throw UserException.unsupportedError(new UnsupportedOperationException("Union type not supported in aggregate functions")).build(logger);
@@ -283,16 +367,28 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
         continue;
       }
 
-      if ( expr instanceof FunctionHolderExpression ) {
-         String funcName = ((FunctionHolderExpression) expr).getName();
-         if ( funcName.equals("sum") || funcName.equals("max") || funcName.equals("min") ) {extraNonNullColumns++;}
-      }
       final MaterializedField outputField = MaterializedField.create(ne.getRef().getAsNamePart().getName(), expr.getMajorType());
-      @SuppressWarnings("resource")
-      ValueVector vv = TypeHelper.getNewVector(outputField, oContext.getAllocator());
+      @SuppressWarnings("resource") ValueVector vv = TypeHelper.getNewVector(outputField, oContext.getAllocator());
       aggrOutFieldIds[i] = container.add(vv);
 
       aggrExprs[i] = new ValueVectorWriteExpression(aggrOutFieldIds[i], expr, true);
+
+      if (expr instanceof FunctionHolderExpression) {
+        String funcName = ((FunctionHolderExpression) expr).getName();
+        if (funcName.equals("sum") || funcName.equals("max") || funcName.equals("min")) {
+          extraNonNullColumns++;
+        }
+        if (((FunctionCall) ne.getExpr()).args.get(0) instanceof SchemaPath) {
+          columnMapping.put(outputField.getName(), ((SchemaPath) ((FunctionCall) ne.getExpr()).args.get(0)).getAsNamePart().getName());
+        }  else if (((FunctionCall) ne.getExpr()).args.get(0) instanceof FunctionCall) {
+          FunctionCall functionCall = (FunctionCall) ((FunctionCall) ne.getExpr()).args.get(0);
+          if (functionCall.args.get(0) instanceof SchemaPath) {
+            columnMapping.put(outputField.getName(), ((SchemaPath) functionCall.args.get(0)).getAsNamePart().getName());
+          }
+        }
+      } else {
+        columnMapping.put(outputField.getName(), ne.getRef().getAsNamePart().getName());
+      }
     }
 
     setupUpdateAggrValues(cgInner);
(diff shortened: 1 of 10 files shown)


ppadma:
@Ben-Zvi please review.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableAllocationTracker.java
```diff
@@ -46,28 +46,26 @@ protected HashTableAllocationTracker(final HashTableConfig config,
...
       Preconditions.checkState(remainingCapacity > 0);
-      return computeNextBatchHolderSize();
+      return computeNextBatchHolderSize(batchSize);
     }
   }
 
-  private int computeNextBatchHolderSize() {
-    return Math.min(remainingCapacity, maxBatchHolderSize);
+  private int computeNextBatchHolderSize(int batchSize) {
+    return Math.min(batchSize, Math.min(remainingCapacity, maxBatchHolderSize));
```
ilooner: Since we are using batchSize now, we can remove the maxBatchHolderSize field.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -1317,7 +1364,7 @@ private void checkGroupAndAggrValues(int incomingRowIdx) {
 
         useReservedValuesMemory(); // try to preempt an OOM by using the reserve
 
-        addBatchHolder(currentPartition);  // allocate a new (internal) values batch
+        addBatchHolder(currentPartition, getBatchSize());  // allocate a new (internal) values batch
```
ilooner: I don't really agree with this approach. We are dynamically adjusting the BatchHolder size based on new records we see coming in, but does this even make sense to do?

We don't really have control over which BatchHolder incoming records get hashed into, so they may get hashed into a BatchHolder that was size based on different input data. Therefore we don't actually get any tangible benefit for independently sizing batch holders dynamically, because the data they contain gets distributed randomly based on a hash. If there was a close correlation between the data in an input batch and the data in a BatchHolder, there would actually be a benefit, but that is not the case here.

I guess this was done because with the current code the size of an output batch is equivalent to the size of a batch holder since an entire batch holder is emitted at once. But instead of trying to work within that limitation I think we should actually change to code to allow the output of a partial batch holder. Additionally we should not allow heterogenous batch holder sizes. We could use the statistics from the first input batch to do a reasonable sizing for batch holders instead of the current 64k default, which would be a great change. However, after a BatchHolder size is selected it should not be changed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -111,6 +111,31 @@
 
   private int resizingTime = 0;
 
+
+  // Given hashtable index, figure out the corresponding index in the batch.
+  public int idxInBatch(int idx) {
```
ilooner: This is complexity comes from the heterogenous BatchHolder sizes. Additionally there is a performance penalty for calculating indexes as the HashTable gets larger. As my comment above mentions, I don't think there is any benefit to having different sizes for batch holders. We should compute a reasonable size based on the first set of input data, and then use the same BatchHolder size everywhere.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ppadma: Adjusting batch holder size here means adjusting number of rows in the batch, based on average row width. Idea is to limit size in terms of memory, not in terms of number of rows. Batches are limited to 16MB (or whatever configured output batch size).  By allocating huge batches and partially transmitting them, we might be able to limit output batch size, but that does not produce much benefit. We want to avoid huge memory allocations.
Why we should not change batch holder size ? 
If we size just based on first batch, it creates the exact problem you mentioned i.e. they will be sized based on older input data and they may not make much sense for new data.
What I have is not exact perfect solution. In fact, I don't even know if such a solution is possible or exists. This will work fine with law of averages.

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -823,7 +817,15 @@ public static int safeDivide(int num, double denom) {
   }
 
   public int rowCount() { return rowCount; }
-  public int stdRowWidth() { return stdRowWidth; }
+
```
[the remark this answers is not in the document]
ppadma: Yes, it is equivalent. We just don't want to calculate it by default and save it. There is a performance penalty to that. Instead, we will calculate when needed.

ppadma:
Regarding changes to updateEstMaxBatchSize(RecordBatch incoming), I actually tried to fix that. But, it is kind of messy. All the calculations there are done based on MAX_BATCH_SIZE, which is basically being  on the conservative side. That is why I did not see OOM errors in any of the tests.  I will address that part in a separate PR or we can discuss and you can fix that as part of refactoring you are doing.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Padma I agree we want to limit the size of output batches, and that reducing the batch holder size is a great change. Having BatchHolders with 64K always 64k rows is not practical. My issue is with changing the batch holder size dynamically. I think it adds complexity without a concrete benefit. Since new data will be added to old BatchHolders data will never really go into a BatchHolder that was appropriately sized for it. Since we can't really have an accurate solution by taking the complex approach with dynamically changing BatchHolder sizes, I think we should go with the simpler approach. We can still use all your changes, I just think we shouldn't continue updating the BatchHolder size.

The complexity added is in the added overhead for computing indexes in the hashtable, and there will be more complexity in doing the refactored memory calculations I am adding.

ilooner:
I'm okay with delaying changes to updateEstMaxBatchSize as long as all the QA tests pass.

I think there are two problems at play here. 

 1. BatchHolders default to 64k rows. This change addresses this issue, my only comment is so not change the batch holder size dynamically.
 2. Limit output batch size.

I think limiting output batch size should be treated independently from the BatchHolder size. When we output batches we must do some additional math to pick the correct number of rows from the BatchHolders. This is because the BatchHolders contains a variety of different data that our statistics did not accurately capture at the time we picked the BatchHolder size. I agree that the solution for problem (2) should be dynamic, and in fact we can have an accurate solution for it since we know the BatchHolders we are getting the records from to create output batches.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -86,6 +99,56 @@
...
+
+    @Override
+    public void update() {
+      // Get sizing information for the batch.
+      setRecordBatchSizer(new RecordBatchSizer(incoming));
+
+      int fieldId = 0;
+      int avgOutgoingRowWidth = 0;
+      for (VectorWrapper<?> w : container) {
+        if (w.getValueVector() instanceof FixedWidthVector) {
```
ilooner: Do these calculations take into account that when we aggregate an INT value it turns into a BIGINT?

ppadma: yes, these are based on types of outgoing container.

ilooner: K thx

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ppadma: agree. there is complexity with variable size batch holders. but, I already dealt with it and made the changes. idea is whenever we are allocating a new batch, we want to use the most recent information. Let us discuss offline with @Ben-Zvi also and decide what we want to do here.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -193,17 +201,43 @@ public int metricId() {
     }
   }
 
+  // Given hashtable index, figure out the corresponding index in the batch.
+  public int idxInBatch(int parition, int idx) {
```
Ben-Zvi: public --> private  , and "parition" is misspelled ....

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -193,17 +201,43 @@ public int metricId() {
...
+      if (idx < (totalSize + batchHolders[parition].get(i).batchSize)) {
+        return (idx - totalSize);
+      }
+      totalSize += batchHolders[parition].get(i).batchSize;
+    }
+    throw new IllegalArgumentException("idx must be less than total size");
+  }
+
+  // Given hashtable index, figure out the corresponding batch.
+  public int idxOfBatch(int partition, int idx) {
```
Ben-Zvi: public --> private , and same comment about the for() loop below

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -193,17 +201,43 @@ public int metricId() {
     }
   }
 
+  // Given hashtable index, figure out the corresponding index in the batch.
+  public int idxInBatch(int parition, int idx) {
+    int totalSize = 0;
+    for (int i=0; i < batchHolders[parition].size(); i++) {
```
Ben-Zvi: A little nicer (instead of  for( int i=0 .....) is:
```
for ( BatchHolder bh : batchHolders[partition] )
```

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -252,6 +286,7 @@ private void setup() {
     }
 
     private void outputValues(IndexPointer outStartIdxHolder, IndexPointer outNumRecordsHolder) {
+      int batchOutputCount = 0;
```
Ben-Zvi: The original code had the intention to start at any offset in the batch. This change is hard-coding to always start at _zero_. If so, the first parameter (and the global **outStartIdxHolder**) can be eliminated.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -193,17 +201,43 @@ public int metricId() {
...
+    }
+    throw new IllegalArgumentException("idx must be less than total size");
+  }
+
   public class BatchHolder {
 
     private VectorContainer aggrValuesContainer; // container for aggr values (workspace variables)
     private int maxOccupiedIdx = -1;
-    private int batchOutputCount = 0;
+    private int batchSize = 0;
```
Ben-Zvi: The word **size** causes some confusion (e.g. size in bytes ?); this is why **count** was originally chosen.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
ilooner: Yeah let's talk about it offline.

Ben-Zvi: Again confusion between **size** and **count** -- The new BatchHolder is allocated based on a **count** of rows, this count is computed (based on sizing the incoming) such that the result batch size would not exceed the given (e.g. 16MB). 
   Now worst case, the following incoming (or a size deviation within the averaged rows) yields bigger values (e.g. longer varchars) -- Looks like the current code would continue to fill the previous batch based on the original **count**. Maybe a check can be added so in this case fewer rows would be inserted into the batch (i.e., less than the original count), such that the **size** restriction would be kept.
   Hopefully such a check would be cheap ....

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
Ben-Zvi: The original code referred to location as a pair: < batch-num ; index-in-batch >
This was condensed into a 4 byte "int", each element in a 2 byte.
With this model, any batch can be of a different size, and access be done as before (e.g. **& BATCH_MASK**). Only "did we reach the limit" checks need to be done against the per-batch limit.
The code that accesses the hash table is super hot code (per each row inserted, searched, etc). Thus running thru such loops is a big performance penalty !!!!!

ppadma:
@Ben-Zvi @ilooner I addressed all code review comments. I got rid of the while loops. Instead, I am adjusting the idx based on batch size when a new batch gets added.  This should address the performance concerns while keeping heterogenous batch sizes. Also, I am adjusting the output row count if incoming batch memory sizing changes. 
Please take a look and give me feedback.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -251,17 +271,12 @@ private void setup() {
...
-      outNumRecordsHolder.value = 0;
-      for (int i = batchOutputCount; i <= maxOccupiedIdx; i++) {
-        try { outputRecordValues(i, batchOutputCount); }
-        catch (SchemaChangeException sc) { throw new UnsupportedOperationException(sc);}
-        if (EXTRA_DEBUG_2) {
-          logger.debug("Outputting values to output index: {}", batchOutputCount);
+    private void outputValues() {
+      for (int i = 0; i <= maxOccupiedIdx; i++) {
+        try {
+          outputRecordValues(i, i);
```
Ben-Zvi: Just a comment: Now that the values output is 1-to-1 from the internal vectors to the output vectors, it looks inefficient to loop and copy/compute the output one row at a time, making no use of any vectorization. (Also relevant to DRILL-5588).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -86,6 +99,63 @@
...
+      int fieldId = 0;
+      int newOutgoingRowWidth = 0;
+      for (VectorWrapper<?> w : container) {
+        if (w.getValueVector() instanceof FixedWidthVector) {
+          newOutgoingRowWidth += ((FixedWidthVector) w.getValueVector()).getValueWidth();
+          if (fieldId >= numGroupByExprs) {
+            valuesRowWidth += ((FixedWidthVector) w.getValueVector()).getValueWidth();
+          }
+        } else {
+          RecordBatchSizer.ColumnSize columnSize = getRecordBatchSizer().getColumn(columnMapping.get(w.getValueVector().getField().getName()));
```
Ben-Zvi: Even for pretty consistent varchars, this size is likely to vary a little from one incoming batch to the next (as it is computed by dividing the size by the number). However the check below (line 146) for "row width did not change" assumes a perfect match. So small variations in varchars average size would keep adjusting.   
  *Question:* Can this check be refined (e.g., use the "next power of two") such that small changes could often be ignored ?

ppadma:
I did not understand why  50% of output batches will exceed the configured output batch size. 
Let me setup a meeting to discuss.

ilooner:
@ppadma Here is a good summary of how an average of normally distributed data relates to the data itself https://www.mathsisfun.com/data/standard-normal-distribution.html .

If you are computing an average of normally distributed data 50% of the data will be less than the average and 50% will be greater than the average. Therefore if we are using averages to compute row counts for output batches, 50% of the output batches will be less than the target size and 50% will be larger. The degree to which output batch sizes deviate from the average is dictated by the standard deviation.

Such situations are unavoidable with a statistics based approach, which is fine. But we should handle the degenerate case where output batches exceed the target batch size.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -86,6 +99,67 @@
...
 
+  public int getOutputRowCount() {
+    return hashAggMemoryManager.getOutputRowCount();
+  }
+
+  public RecordBatchMemoryManager getRecordBatchMemoryManager() {
+    return hashAggMemoryManager;
+  }
+
+  private class HashAggMemoryManager extends RecordBatchMemoryManager {
```
ilooner: Can we pull this out into a separate file and add some unit tests for it?

ppadma:
@ilooner that is generalizing way too much. What we are trying to accomplish here is use real data to figure out the memory needed. Before, we would use standard size of 50 for varchar, which can differ from average actual varchar column size (could be by a big factor).  Typically, within a column, data will be around the same size and that could differ from standard size we are using by a big factor. What we are trying to optimize here is for accounting for that factor. In my opinion, distribution you are referring to is not applicable here. Also, assuming that data will be more or less evenly distributed across partitions, again, I believe stating that 50% of the batches will be more than average and 50% less than average is also wrong. 

Since we are calculating batch size upfront, emitting partial batches and keeping track of all that is unnecessary complication that can be avoided. I don't think it will provide much benefit. 

Allowing the HashAgg operator to decrease the output batch size configured by the user - Yes, that we have to do. In fact, that is the problem I am having with hash join right now which is causing hash join to spill.

ppadma:
@ilooner @Ben-Zvi I removed the code that dynamically updates the row count. Please review.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
ppadma: @ilooner I want to keep it consistent with the way I did for other operators.  Can I file a JIRA and do that later for all operators ? This code gets tested from tests written in testOutputBatchSize.java.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -84,6 +97,63 @@
...
+      int newOutgoingRowWidth = 0;
+      for (VectorWrapper<?> w : container) {
+        if (w.getValueVector() instanceof FixedWidthVector) {
+          newOutgoingRowWidth += ((FixedWidthVector) w.getValueVector()).getValueWidth();
+          if (fieldId >= numGroupByExprs) {
+            valuesRowWidth += ((FixedWidthVector) w.getValueVector()).getValueWidth();
+          }
+        } else {
+          RecordBatchSizer.ColumnSize columnSize = getRecordBatchSizer().getColumn(columnMapping.get(w.getValueVector().getField().getName()));
+          newOutgoingRowWidth += columnSize.getAllocSizePerEntry();
```
Ben-Zvi: I rebased and tested, and there was a failure here `columnSize == null` in some of the new EMIT unit test. Initially I thought that this has to do with the first outgoing batch (which always has OK_NEW_SCHEMA) being non-empty ; but eventually I made a fix (see below) where the 'name' and the 'expr' are different.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -283,16 +363,23 @@ private HashAggregator createAggregatorInternal() throws SchemaChangeException,
...
       aggrExprs[i] = new ValueVectorWriteExpression(aggrOutFieldIds[i], expr, true);
+
+      if (expr instanceof FunctionHolderExpression) {
+        String funcName = ((FunctionHolderExpression) expr).getName();
+        if (funcName.equals("sum") || funcName.equals("max") || funcName.equals("min")) {
+          extraNonNullColumns++;
+        }
+        if (((FunctionCall) ne.getExpr()).args.get(0) instanceof SchemaPath) {
+          columnMapping.put(outputField.getName(), ((SchemaPath) ((FunctionCall) ne.getExpr()).args.get(0)).getAsNamePart().getName());
+        }
```
Ben-Zvi: When I tested (TestHashAggEmitOutcome) there were cases that did not match **SchemaPath**.
So I added here:
```
      else if (((FunctionCall) ne.getExpr()).args.get(0) instanceof FunctionCall) {
          columnMapping.put(outputField.getName(), ((FunctionCall) ((FunctionCall) ne.getExpr()).args.get(0)).getName());
        }
```
The execution did get there, but aI did not check further (e.g., if this was used right in computing the row size).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -263,13 +343,13 @@ private HashAggregator createAggregatorInternal() throws SchemaChangeException,
 
       // add this group-by vector to the output container
       groupByOutFieldIds[i] = container.add(vv);
+      columnMapping.put(outputField.getName(), ne.getRef().getAsNamePart().getName());
```
Ben-Zvi: The regressions *TestHashAggEmitOutcome* set the "name" and the "expr" to different strings, hence the failure above. So I changed this to:
```
      columnMapping.put(outputField.getName(), ne.getExpr().toString().replace('`',' ').trim() );
```
To make the regressions (e.g. **testHashAggrMultipleEMITOutcome()**) pass.
It is a little ugly (removing the '`' before and after), but I could not think of something nicer.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
```diff
@@ -84,6 +97,63 @@
...
+        }
+        fieldId++;
+      }
+
+      updateIncomingStats();
+      if (logger.isDebugEnabled()) {
+        logger.debug("BATCH_STATS, incoming: {}", getRecordBatchSizer());
+      }
+
+      // We do not want to keep adjusting batch holders target row count
```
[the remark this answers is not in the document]
ppadma: ok. made the change.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -646,14 +647,18 @@ public PutStatus put(int incomingRowIdx, IndexPointer htIdxHolder, int hashCode)
...
-      addedBatch = addBatchIfNeeded(currentIdx);
+      addedBatch = addBatchIfNeeded(currentIdx, targetBatchRowCount);
+      if (addedBatch) {
+        // If we just added the batch, update the current index to point to beginning of new batch.
+        currentIdx = (batchHolders.size() - 1) * BATCH_SIZE;
+        freeIndex = currentIdx + 1;
+      }
     } catch (OutOfMemoryException OOME) {
-      retryAfterOOM( currentIdx < batchHolders.size() * BATCH_SIZE );
+      retryAfterOOM( currentIdx < totalBatchHoldersSize);
```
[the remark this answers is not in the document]
ppadma: yes, I agree. I got rid of totalBatchHolderSize.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
ppadma: ok. I took your fix. Also, I changed the code to check for null column map entry and null column size.

ppadma:
@Ben-Zvi addressed review comments. please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -568,8 +576,10 @@ private void retryAfterOOM(boolean batchAdded) throws RetryAfterSpillException {
     if ( batchAdded ) {
       logger.trace("OOM - Removing index {} from the batch holders list",batchHolders.size() - 1);
       BatchHolder bh = batchHolders.remove(batchHolders.size() - 1);
+      prevIndexSize = batchHolders.size() > 1 ? (batchHolders.size()-1) * BATCH_SIZE : 0;
+      totalIndexSize = batchHolders.size() * BATCH_SIZE;
       // update freeIndex to point to end of last batch + 1
-      freeIndex = batchHolders.size() * BATCH_SIZE + 1;
+      freeIndex = totalIndexSize + 1;
```
Ben-Zvi: No need for "+ 1" -- the free index starts at batch offset of zero.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -684,7 +694,7 @@ public PutStatus put(int incomingRowIdx, IndexPointer htIdxHolder, int hashCode,
     }
     htIdxHolder.value = currentIdx;
     return  addedBatch ? PutStatus.NEW_BATCH_ADDED :
-        (freeIndex + 1 > (batchHolders.size() * BATCH_SIZE)) ?
+        (freeIndex + 1 > totalIndexSize) ?
```
Ben-Zvi: Problem: The original code "overflew" the (64K - 1) by adding one, hence this check returned KEY_ADDED_LAST, which signaled to the caller that a batch just got filled -- a good opportunity to check if need to spill (since a new batch would surely be allocated soon).
   By going below 64K, this would not work, hence the code misses on some of the checks of "is spill needed now".

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -694,7 +694,7 @@ public PutStatus put(int incomingRowIdx, IndexPointer htIdxHolder, int hashCode,
     }
     htIdxHolder.value = currentIdx;
     return  addedBatch ? PutStatus.NEW_BATCH_ADDED :
-        (freeIndex + 1 > totalIndexSize) ?
+        (freeIndex + 1 > prevIndexSize + batchHolders.get(batchHolders.size()-1).getTargetBatchRowCount()) ?
```
Ben-Zvi: `prevIndexSize` is used in two places, here and when checking if a new batch is needed. Both use it the same way (by adding the size of the last batch). Both places are part of the **HOT** code, called for every row.
   So maybe if this variable could be set to include the size of the last batch, a lot of computations would be saved.
   Will it be simple?  Need to know the size of the last batch when setting this variable.

ppadma:
@Ben-Zvi Thank you Boaz for the thorough review and giving me a +1 finally.

vvysotskyi:
Unit tests in travis build for this PR failed, @ppadma, could you please take a look at this?

ppadma:
@vvysotskyi sorry about that. fixed the problem.updated the PR.

PR #1329, opened 2018-06-19
DRILL-6513: Max query memory per node set to Drillbit's max direrct memory

sachouche:
Changed the "planner.memory.max_query_memory_per_node" configuration parameter maximum value to be the Drillbit maximum direct memory. 

**NOTE -** This change (as described in the JIRA) will not address all issues related to this configuration value as this is out-of-scope and currently not possible; future enhancements shall happen along with the Drill Resource Management feature.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec;
 
+import org.apache.drill.common.config.DrillConfig;
 import org.apache.drill.exec.physical.impl.common.HashTable;
 import org.apache.drill.exec.rpc.user.InboundImpersonationManager;
 import org.apache.drill.exec.server.options.OptionValidator;
@@ -440,7 +441,7 @@ public final class ExecConstants {
    * DEFAULT: 2048 MB
    */
   public static final String MAX_QUERY_MEMORY_PER_NODE_KEY = "planner.memory.max_query_memory_per_node";
-  public static final LongValidator MAX_QUERY_MEMORY_PER_NODE = new RangeLongValidator(MAX_QUERY_MEMORY_PER_NODE_KEY, 1024 * 1024, Long.MAX_VALUE);
+  public static final LongValidator MAX_QUERY_MEMORY_PER_NODE = new RangeLongValidator(MAX_QUERY_MEMORY_PER_NODE_KEY, 1024 * 1024, DrillConfig.getMaxDirectMemory());
 
   /**
    * Alternative way to compute per-query-per-node memory as a percent


sachouche:
@sohami and @ilooner, can you please review this PR request?

Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -440,7 +441,7 @@ private ExecConstants() {
    * DEFAULT: 2048 MB
    */
   public static final String MAX_QUERY_MEMORY_PER_NODE_KEY = "planner.memory.max_query_memory_per_node";
-  public static final LongValidator MAX_QUERY_MEMORY_PER_NODE = new RangeLongValidator(MAX_QUERY_MEMORY_PER_NODE_KEY, 1024 * 1024, Long.MAX_VALUE);
+  public static final LongValidator MAX_QUERY_MEMORY_PER_NODE = new RangeLongValidator(MAX_QUERY_MEMORY_PER_NODE_KEY, 1024 * 1024, DrillConfig.getMaxDirectMemory());
```
vrozov: There is an assumption that all Drillbits in a cluster are started with the same amount of direct memory. What if this is not the case (cluster is composed of different size boxes)?

sachouche: I have raised this point to the team and they were fine with the change as currently the Planner assumes a Drill cluster is made of homogeneous nodes.

rhou1: The current implementation enables the customer to set MAX_QUERY_MEMORY_PER_NODE to unreasonable values, such as 1 TB.  This has lead to out-of-memory issues in some of our testing.

Drill should validate user settings to a reasonable point.  This is an attempt to move in that direction.  There is more work needed here.

I think it is challenging for the planner to make the best optimizations if each node in the cluster is a different size, and they want to use each node to its best capabilities.  I.e. make some nodes work harder than others.

One work-around is for the customer to set this option on the node with the largest Direct Memory if they want to set the option higher than recommended.  This could cause instability on the nodes that have less Direct Memory available.

vrozov: To achieve the goal of avoiding OOM errors, it is necessary to use node with the lowest amount of memory available, but what node will be used to validate max query memory per node depends on where the setting is applied.

rhou1: Yes, this change is not perfect by any means.  It is just a step towards a better solution.

A better solution is to have a global Resource Manager determine the right setting, where the Resource Manager knows what each node is capable of.  But we don't have that yet.

sachouche: @vrozov 

- The current code was not performing any validation (the maximum value was Long.MAX_VALUE)
- Now we added validation with the caveat that we are assuming the Drillbit are configured homogeneously
- Our code assumes the same configuration (cores-per-node, memory-per-node, and so on)

I suggest the following:
- Let's proceed with this new validation since it is an improvement (albeit not perfect); from the get go I emphasized the scope of this JIRA is limited
- Create a new JIRA to validate that all Drillbits are homogeneous; this will address your current feedback and many other validation deficiencies

sachouche:
Thank you all for your valuable reviews!

Ben-Zvi:
One issue not covered is memory used by non-buffered operators. This change seems to advocate use of " max_query_memory_per_node == DirectMemory ", which can easily lead to an OOM -- All the buffered instances get a (promised) share of the MQMPN, but at runtime some memory is taken by other operators (e.g., scanners), so if **all** the buffered operator instances would try to use **almost all** of their promised memory, the JVM would OOM.

Maybe I'd open another Jira for that; introducing another new option to **reserve** a percentage of the direct memory, hence forcing the MQMPN to be lower than the DirectMemory.

PR #1294, opened 2018-05-25
DRILL-6147: Adding Columnar Parquet Batch Sizing functionality

sachouche:
Implemented the Parquet Batching Sizing functionality for the Flat Reader; please find the associated functional and design specification [here](https://docs.google.com/document/d/1A6zFkjxnC_-9RwG4h0sI81KI5ZEvJ7HzgClCUFpB5WE/edit#heading=h.g2id8coli1bb).

@parthchandra, can you please review this PR? 

Thanks!

PR #1330, opened 2018-06-20
DRILL-6147: Adding Columnar Parquet Batch Sizing functionality

sachouche:
Implemented the Parquet Batching Sizing functionality for the Flat Reader; please find the associated functional and design specification [here](https://docs.google.com/document/d/1A6zFkjxnC_-9RwG4h0sI81KI5ZEvJ7HzgClCUFpB5WE/edit#heading=h.g2id8coli1bb).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -316,6 +316,13 @@ public final class ExecConstants {
   public static final String PARQUET_FLAT_READER_BULK = "store.parquet.flat.reader.bulk";
   public static final OptionValidator PARQUET_FLAT_READER_BULK_VALIDATOR = new BooleanValidator(PARQUET_FLAT_READER_BULK);
 
+  // Controls the flat parquet reader batching constraints (number of record and memory limit)
+  public static final String PARQUET_FLAT_BATCH_NUM_RECORDS = "store.parquet.flat.batch.num_records";
+  public static final OptionValidator PARQUET_FLAT_BATCH_NUM_RECORDS_VALIDATOR = new RangeLongValidator(PARQUET_FLAT_BATCH_NUM_RECORDS, 1, Integer.MAX_VALUE);
+  public static final String PARQUET_FLAT_BATCH_MEMORY_SIZE = "store.parquet.flat.batch.memory_size";
+  // This configuration is used to overwrite the common memory batch sizing configuration property
+  public static final OptionValidator PARQUET_FLAT_BATCH_MEMORY_SIZE_VALIDATOR = new RangeLongValidator(PARQUET_FLAT_BATCH_MEMORY_SIZE, 0, Integer.MAX_VALUE);
+
   public static final String JSON_ALL_TEXT_MODE = "store.json.all_text_mode";
   public static final BooleanValidator JSON_READER_ALL_TEXT_MODE_VALIDATOR = new BooleanValidator(JSON_ALL_TEXT_MODE);
   public static final BooleanValidator JSON_EXTENDED_TYPES = new BooleanValidator("store.json.extended_types");
@@ -688,5 +695,14 @@ public final class ExecConstants {
   public static final String DRILL_PORT_HUNT = "drill.exec.port_hunt";
 
   public static final String ALLOW_LOOPBACK_ADDRESS_BINDING = "drill.exec.allow_loopback_address_binding";
+
+  /** Enables batch size statistics logging */
+  public static final String STATS_LOGGING_BATCH_SIZE_OPTION = "drill.exec.stats.logging.batch_size";
+  public static final BooleanValidator STATS_LOGGING_BATCH_SIZE_VALIDATOR = new BooleanValidator(STATS_LOGGING_BATCH_SIZE_OPTION);
+
+  /** Enables fine-grained batch size statistics logging */
+  public static final String STATS_LOGGING_FG_BATCH_SIZE_OPTION = "drill.exec.stats.logging.fine_grained.batch_size";
+  public static final BooleanValidator STATS_LOGGING_BATCH_FG_SIZE_VALIDATOR = new BooleanValidator(STATS_LOGGING_FG_BATCH_SIZE_OPTION);
+
 
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
@@ -47,6 +47,8 @@ import org.apache.drill.exec.store.RecordReader;
 import org.apache.drill.exec.testing.ControlsInjector;
 import org.apache.drill.exec.testing.ControlsInjectorFactory;
 import org.apache.drill.exec.util.CallBack;
+import org.apache.drill.exec.util.record.RecordBatchStats;
+import org.apache.drill.exec.util.record.RecordBatchStats.RecordBatchStatsContext;
 import org.apache.drill.exec.vector.AllocationHelper;
 import org.apache.drill.exec.vector.NullableVarCharVector;
 import org.apache.drill.exec.vector.SchemaChangeCallBack;
@@ -80,6 +82,8 @@ public class ScanBatch implements CloseableRecordBatch {
   private final BufferAllocator allocator;
   private final List<Map<String, String>> implicitColumnList;
   private String currentReaderClassName;
+  private final RecordBatchStatsContext batchStatsLogging;
+
   /**
    *
    * @param context
@@ -117,6 +121,7 @@ public class ScanBatch implements CloseableRecordBatch {
       this.implicitColumnList = implicitColumnList;
       addImplicitVectors();
       currentReader = null;
+      batchStatsLogging = new RecordBatchStatsContext(context, oContext);
     } finally {
       oContext.getStats().stopProcessing();
     }
@@ -174,6 +179,7 @@ public class ScanBatch implements CloseableRecordBatch {
         boolean isNewSchema = mutator.isNewSchema();
         populateImplicitVectorsAndSetCount();
         oContext.getStats().batchReceived(0, recordCount, isNewSchema);
+        logRecordBatchStats();
 
         if (recordCount == 0) {
           currentReader.close();
@@ -290,6 +296,45 @@ public class ScanBatch implements CloseableRecordBatch {
   public VectorWrapper<?> getValueAccessorById(Class<?> clazz, int... ids) {
     return container.getValueAccessorById(clazz, ids);
   }
+
+  private void logRecordBatchStats() {
+    final int MAX_FQN_LENGTH = 50;
+
+    if (recordCount == 0) {
+      return; // NOOP
+    }
+
+    RecordBatchStats.logRecordBatchStats(
+      batchStatsLogging.getContextOperatorId(),
+      getFQNForLogging(MAX_FQN_LENGTH),
+      this,
+      batchStatsLogging,
+      logger);
+  }
+
+  /** Might truncate the FQN if too long */
+  private String getFQNForLogging(int maxLength) {
+    final String FQNKey = "FQN";
+    final ValueVector v = mutator.implicitFieldVectorMap.get(FQNKey);
+
+    final Object fqnObj;
+
+    if (v == null
+     || v.getAccessor().getValueCount() == 0
+     || (fqnObj = ((NullableVarCharVector) v).getAccessor().getObject(0)) == null) {
+
+      return "NA";
+    }
+
+    String fqn = fqnObj.toString();
+
+    if (fqn != null && fqn.length() > maxLength) {
+      fqn = fqn.substring(fqn.length() - maxLength, fqn.length());
+    }
+
+    return fqn;
+  }
+
 
   /**
    * Row set mutator implementation provided to record readers created by
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
@@ -21,7 +21,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
-
+import com.google.common.base.Preconditions;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Lists;
 
@@ -38,13 +38,13 @@ public abstract class BatchReader {
 
   public int readBatch() throws Exception {
     ColumnReader<?> firstColumnStatus = readState.getFirstColumnReader();
-    long recordsToRead = Math.min(getReadCount(firstColumnStatus), readState.getRecordsToRead());
+    int currBatchNumRecords = readState.batchSizerMgr().getCurrentRecordsPerBatch();
+    long recordsToRead = Math.min(currBatchNumRecords, readState.getRemainingValuesToRead());
     int readCount = readRecords(firstColumnStatus, recordsToRead);
+
     readState.fillNullVectors(readCount);
     return readCount;
   }
-
-  protected abstract long getReadCount(ColumnReader<?> firstColumnStatus);
 
   protected abstract int readRecords(ColumnReader<?> firstColumnStatus, long recordsToRead) throws Exception;
 
@@ -59,14 +59,14 @@ public abstract class BatchReader {
   }
 
   protected void readAllFixedFieldsSerial(long recordsToRead) throws IOException {
-    for (ColumnReader<?> crs : readState.getColumnReaders()) {
+    for (ColumnReader<?> crs : readState.getFixedLenColumnReaders()) {
       crs.processPages(recordsToRead);
     }
   }
 
   protected void readAllFixedFieldsParallel(long recordsToRead) throws Exception {
     ArrayList<Future<Long>> futures = Lists.newArrayList();
-    for (ColumnReader<?> crs : readState.getColumnReaders()) {
+    for (ColumnReader<?> crs : readState.getFixedLenColumnReaders()) {
       Future<Long> f = crs.processPagesAsync(recordsToRead);
       if (f != null) {
         futures.add(f);
@@ -106,15 +106,6 @@ public abstract class BatchReader {
     }
 
     @Override
-    protected long getReadCount(ColumnReader<?> firstColumnStatus) {
-      if (readState.recordsRead() == readState.schema().getGroupRecordCount()) {
-        return 0;
-      }
-      return Math.min(ParquetRecordReader.DEFAULT_RECORDS_TO_READ_IF_VARIABLE_WIDTH,
-                      readState.schema().getGroupRecordCount() - readState.recordsRead());
-    }
-
-    @Override
     protected int readRecords(ColumnReader<?> firstColumnStatus, long recordsToRead) {
       readState.updateCounts((int) recordsToRead);
       return (int) recordsToRead;
@@ -133,15 +124,15 @@ public abstract class BatchReader {
     }
 
     @Override
-    protected long getReadCount(ColumnReader<?> firstColumnStatus) {
-      return Math.min(readState.schema().getRecordsPerBatch(),
-                      firstColumnStatus.columnChunkMetaData.getValueCount() - firstColumnStatus.totalValuesRead);
-    }
-
-    @Override
     protected int readRecords(ColumnReader<?> firstColumnStatus, long recordsToRead) throws Exception {
       readAllFixedFields(recordsToRead);
-      return firstColumnStatus.getRecordsReadInCurrentPass();
+
+      Preconditions.checkNotNull(firstColumnStatus != null);
+      readState.setValuesReadInCurrentPass(firstColumnStatus.getRecordsReadInCurrentPass()); // get the number of rows read
+
+      readState.updateCounts((int) recordsToRead); // update the shared Reader State
+
+      return readState.getValuesReadInCurrentPass();
     }
   }
 
@@ -157,15 +148,21 @@ public abstract class BatchReader {
     }
 
     @Override
-    protected long getReadCount(ColumnReader<?> firstColumnStatus) {
-      return ParquetRecordReader.DEFAULT_RECORDS_TO_READ_IF_VARIABLE_WIDTH;
-    }
-
-    @Override
     protected int readRecords(ColumnReader<?> firstColumnStatus, long recordsToRead) throws Exception {
+      // We should not rely on the "firstColumnStatus.getRecordsReadInCurrentPass()" when dealing
+      // with variable length columns as each might return a different number of records. The batch size
+      // will be the lowest value. The variable column readers will update the "readState" object to
+      // reflect the correct information.
       long fixedRecordsToRead = readState.varLengthReader().readFields(recordsToRead);
       readAllFixedFields(fixedRecordsToRead);
-      return firstColumnStatus.getRecordsReadInCurrentPass();
+
+      // Sanity check the fixed readers read the expected number of rows
+      Preconditions.checkArgument(firstColumnStatus == null
+        || firstColumnStatus.getRecordsReadInCurrentPass() == readState.getValuesReadInCurrentPass());
+
+      readState.updateCounts((int) fixedRecordsToRead);
+
+      return readState.getValuesReadInCurrentPass();
     }
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ParquetSchema.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ParquetSchema.java
@@ -48,7 +48,7 @@ import com.google.common.collect.Lists;
  * to the schema that Drill and the Parquet reader uses.
  */
 
-public class ParquetSchema {
+public final class ParquetSchema {
   /**
    * Set of columns specified in the SELECT clause. Will be null for
    * a SELECT * query.
@@ -74,7 +74,6 @@ public class ParquetSchema {
   private int bitWidthAllFixedFields;
   private boolean allFieldsFixedLength;
   private long groupRecordCount;
-  private int recordsPerBatch;
 
   /**
    * Build the Parquet schema. The schema can be based on a "SELECT *",
@@ -108,22 +107,13 @@ public class ParquetSchema {
    * Build the schema for this read as a combination of the schema specified in
    * the Parquet footer and the list of columns selected in the query.
    *
-   * @param batchSize target size of the batch, in rows
    * @throws Exception if anything goes wrong
    */
 
-  public void buildSchema(long batchSize) throws Exception {
+  public void buildSchema() throws Exception {
     groupRecordCount = footer.getBlocks().get(rowGroupIndex).getRowCount();
     loadParquetSchema();
     computeFixedPart();
-
-    if (! selectedColumnMetadata.isEmpty() && allFieldsFixedLength) {
-      recordsPerBatch = (int) Math.min(Math.min(batchSize / bitWidthAllFixedFields,
-          footer.getBlocks().get(0).getColumns().get(0).getValueCount()), ParquetRecordReader.DEFAULT_RECORDS_TO_READ_IF_FIXED_WIDTH);
-    }
-    else {
-      recordsPerBatch = ParquetRecordReader.DEFAULT_RECORDS_TO_READ_IF_VARIABLE_WIDTH;
-    }
   }
 
   /**
@@ -168,7 +158,6 @@ public class ParquetSchema {
   public boolean isStarQuery() { return selectedCols == null; }
   public ParquetMetadata footer() { return footer; }
   public int getBitWidthAllFixedFields() { return bitWidthAllFixedFields; }
-  public int getRecordsPerBatch() { return recordsPerBatch; }
   public boolean allFieldsFixedLength() { return allFieldsFixedLength; }
   public List<ParquetColumnMetadata> getColumnMetadata() { return selectedColumnMetadata; }
 
@@ -195,9 +184,6 @@ public class ParquetSchema {
    */
 
   private boolean fieldSelected(MaterializedField field) {
-    // TODO - not sure if this is how we want to represent this
-    // for now it makes the existing tests pass, simply selecting
-    // all available data if no columns are provided
     if (isStarQuery()) {
       return true;
     }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBinaryReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBinaryReader.java
@@ -19,39 +19,43 @@ package org.apache.drill.exec.store.parquet.columnreaders;
 
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.Lists;
-import org.apache.drill.common.exceptions.DrillRuntimeException;
-import org.apache.drill.exec.vector.ValueVector;
-
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import org.apache.drill.common.exceptions.DrillRuntimeException;
+import org.apache.drill.exec.store.parquet.columnreaders.batchsizing.RecordBatchOverflow;
+import org.apache.drill.exec.store.parquet.columnreaders.batchsizing.RecordBatchOverflow.FieldOverflowDefinition;
+import org.apache.drill.exec.store.parquet.columnreaders.batchsizing.RecordBatchSizerManager;
+import org.apache.drill.exec.store.parquet.columnreaders.batchsizing.RecordBatchSizerManager.FieldOverflowState;
+import org.apache.drill.exec.store.parquet.columnreaders.batchsizing.RecordBatchSizerManager.FieldOverflowStateContainer;
+import org.apache.drill.exec.store.parquet.columnreaders.batchsizing.RecordBatchSizerManager.VarLenColumnBatchStats;
+import org.apache.drill.exec.vector.ValueVector;
 
+/** Class which handles reading a batch of rows from a set of variable columns */
 public class VarLenBinaryReader {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(VarLenBinaryReader.class);
 
-  ParquetRecordReader parentReader;
+  final ParquetRecordReader parentReader;
+  final RecordBatchSizerManager batchSizer;
   final List<VarLengthColumn<? extends ValueVector>> columns;
+  /** Sorting columns to minimize overflow */
+  final List<VLColumnContainer> orderedColumns;
+  private final Comparator<VLColumnContainer> comparator = new VLColumnComparator();
   final boolean useAsyncTasks;
-  private final long targetRecordCount;
   private final boolean useBulkReader;
 
   public VarLenBinaryReader(ParquetRecordReader parentReader, List<VarLengthColumn<? extends ValueVector>> columns) {
     this.parentReader = parentReader;
+    this.batchSizer = parentReader.getBatchSizesMgr();
     this.columns = columns;
+    this.orderedColumns = populateOrderedColumns();
     this.useAsyncTasks = parentReader.useAsyncColReader;
     this.useBulkReader = parentReader.useBulkReader();
-
-    // Can't read any more records than fixed width fields will fit.
-    // Note: this calculation is very likely wrong; it is a simplified
-    // version of earlier code, but probably needs even more attention.
-
-    int totalFixedFieldWidth = parentReader.getBitWidthAllFixedFields() / 8;
-    if (totalFixedFieldWidth == 0) {
-      targetRecordCount = 0;
-    } else {
-      targetRecordCount = parentReader.getBatchSize() / totalFixedFieldWidth;
-    }
   }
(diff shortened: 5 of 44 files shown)


sachouche:
@ppadma, @Ben-Zvi 

I have accidentally closed PR #1294 after performing a rebase. Can you please comment using this new PR since I wasn't able to re-open the previous one. Sorry for the inconvenience.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -315,6 +315,13 @@ private ExecConstants() {
   public static final String PARQUET_FLAT_READER_BULK = "store.parquet.flat.reader.bulk";
   public static final OptionValidator PARQUET_FLAT_READER_BULK_VALIDATOR = new BooleanValidator(PARQUET_FLAT_READER_BULK);
 
+  // Controls the flat parquet reader batching constraints (number of record and memory limit)
+  public static final String PARQUET_FLAT_BATCH_NUM_RECORDS = "store.parquet.flat.batch.num_records";
+  public static final OptionValidator PARQUET_FLAT_BATCH_NUM_RECORDS_VALIDATOR = new RangeLongValidator(PARQUET_FLAT_BATCH_NUM_RECORDS, 1, Integer.MAX_VALUE);
+  public static final String PARQUET_FLAT_BATCH_MEMORY_SZ = "store.parquet.flat.batch.memory_sz";
```
vrozov: Please be consistent with other drill settings, do not abbreviate size to sz.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -680,4 +687,13 @@ public static String bootDefaultFor(String name) {
 
   public static final String ALLOW_LOOPBACK_ADDRESS_BINDING = "drill.exec.allow_loopback_address_binding";
 
+  /** Enables batch size statistics logging */
+  public static final String STATS_LOGGING_BATCH_SZ_OPTION = "drill.exec.stats.logging.batch_sz";
```
vrozov: Usually the amount of logging is configured using logger level.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/PageReader.java
```diff
@@ -335,7 +335,7 @@ public boolean next() throws IOException {
       readPosInBytes = in.position();
       repetitionLevels.readInteger();
     }
-    if (parentColumnReader.columnDescriptor.getMaxDefinitionLevel() != 0) {
+    if (parentColumnReader.columnDescriptor.getMaxDefinitionLevel() != 0){
```
vrozov: Please avoid format only changes.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -20,6 +20,10 @@
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import io.netty.buffer.DrillBuf;
+import java.util.Collections;
```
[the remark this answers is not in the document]
sachouche: Done! sorry for that as you and I already configured my Eclipse to not re-organize the imports but I guess the default settings have changed after minor update.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
sachouche: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
sachouche: We are using both:
- Record batch will automatically show up when debug level is used
- or we explicitly this configuration; this basically help us minimize the amount of logging
- Will also provide the ability to enable record batch stats per operator type

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/PageReader.java
sachouche: done!

ilooner:
@ppadma @Ben-Zvi Do you guys have time to look at this PR? If not, I can help review pieces of it to help speed things up.

ppadma:
@ilooner yes Tim. that will certainly be helpful. thx a lot.

priteshm:
@ilooner thanks for helping out here. Please do prioritize this review.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenEntryReader.java
```diff
@@ -64,23 +66,23 @@ private final VarLenColumnBulkEntry getEntryBulk(int valuesToRead) {
         break;
       }
 
-      final int data_len = getInt(srcBuff, srcPos);
+      final int dataLen = getInt(srcBuff, srcPos);
```
ilooner: Thanks for fixing variable name.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/BatchSizingMemoryUtil.java
```diff
@@ -0,0 +1,327 @@
...
+
+/** Helper class to assist the Flat Parquet reader build batches which adhere to memory sizing constraints */
+public final class BatchSizingMemoryUtil {
+
+  /** BYTE in-memory width */
+  public static final int BYTE_VALUE_WIDTH = 1;
+  /** INT in-memory width */
+  public static final int INT_VALUE_WIDTH  = 4;
+  /** Default variable length column average precision;
+   * computed in such away that 64k values will fit within one MB to minimize internal fragmentation
```
ilooner: "away" -> "a way"

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/BatchSizingMemoryUtil.java
```diff
@@ -0,0 +1,327 @@
...
+import org.apache.drill.exec.vector.VarBinaryVector;
+import org.apache.drill.exec.vector.VarCharVector;
+import org.apache.drill.exec.vector.VarDecimalVector;
+import org.apache.drill.exec.vector.VariableWidthVector;
+
+/** Helper class to assist the Flat Parquet reader build batches which adhere to memory sizing constraints */
+public final class BatchSizingMemoryUtil {
+
+  /** BYTE in-memory width */
+  public static final int BYTE_VALUE_WIDTH = 1;
```
ilooner: We could reuse some preexisting constants. `UInt1Vector.VALUE_WIDTH`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/BatchSizingMemoryUtil.java
```diff
@@ -0,0 +1,327 @@
...
+  /** INT in-memory width */
+  public static final int INT_VALUE_WIDTH  = 4;
```
ilooner: `UInt4Vector.VALUE_WIDTH`

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -632,6 +632,10 @@ public void setSafe(int index, byte[] bytes) {
       // Let's process the input
       while (input.hasNext()) {
         T entry = input.next();
+
+        if (entry == null || ((VarLenBulkEntry) entry).getNumValues() == 0) {
```
ilooner: Cast is unnecessary

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -897,7 +901,7 @@ public BufferedMutator(int startIdx, int buffSz, ${minor.class}Vector parent) {
       this.parent = parent;
       this.dataBuffOff = this.parent.offsetVector.getAccessor().get(startIdx);
       this.totalDataLen = this.dataBuffOff;
-      this.offsetsMutator = new UInt4Vector.BufferedMutator(startIdx, buffSz * 4, parent.offsetVector);
+      this.offsetsMutator = new UInt4Vector.BufferedMutator(startIdx, buffSz, parent.offsetVector);
```
ilooner: Why remove ` * 4 `?

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
```diff
@@ -1008,6 +1013,11 @@ private void setOffsets(int[] lengths, int numValues, boolean hasNulls) {
         remaining -= toCopy;
 
       } while (remaining > 0);
+
+      // We need to flush as offset data can be accessed during loading to
+      // figute out current payload size.
```
ilooner: "figute" -> "figure"

ilooner:
@sachouche I still need to read things more in depth, so I will continue to add comments as I get a better understanding of the code.

At a high level, I didn't see any unit tests added for the new classes and code. Since this is a significant new feature, we need to add tests. Let's try to come up with a plan for what we can add unit tests for and how. Perhaps we could start with BatchSizingMemoryUtil.java and work from there.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/OverflowSerDeUtil.java
```diff
@@ -0,0 +1,355 @@
...
+    }
+  }
+
+  /**
+   * Helper class for handling variable length {@link ValueVector} overflow serialization logic
+   */
+  private static final class VLVectorSerializer {
+    private static final int BYTE_VALUE_WIDTH = 1;
+    private static final int INT_VALUE_WIDTH  = 4;
+
```
ilooner: We could use `UInt1Vector.VALUE_WIDTH` and `UInt4Vector.VALUE_WIDTH` here as well instead of recreating these constants.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/OverflowSerDeUtil.java
```diff
@@ -0,0 +1,355 @@
...
+    void clear() {
+      dataByteLen = 0;
+      totalByteLength = 0;
+    }
+  }
+
+  /**
+   * Helper class for handling variable length {@link ValueVector} overflow serialization logic
+   */
```
ilooner: Could you add some javadoc about why the serialization used in `SendBatchAsyncListen` and `VectorSerializer.Writer` is not sufficient. My guess is because they blindly dump all the data in the value vectors but you want to dump only a chunk of the data in the value vector?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
```diff
@@ -0,0 +1,657 @@
...
+   * @param schema current reader schema
+   * @param totalRecordsToRead total number of rows to read
+   */
+  public RecordBatchSizerManager(OptionManager options,
+    ParquetSchema schema,
+    long totalRecordsToRead) {
+
+    this.schema = schema;
+    this.totalRecordsToRead = totalRecordsToRead;
+    this.configRecordsPerBatch = options.getOption(ExecConstants.PARQUET_FLAT_BATCH_NUM_RECORDS).num_val.intValue();
```
ilooner: `options.getLong()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
```diff
@@ -0,0 +1,657 @@
...
+    fieldOverflowMap.clear();
+  }
+
+// ----------------------------------------------------------------------------
+// Internal implementation logic
+// ----------------------------------------------------------------------------
+
+  private int getConfiguredMaxBatchMemory(OptionManager options) {
+    // Use the parquet specific configuration if set
+    int maxMemory = options.getOption(ExecConstants.PARQUET_FLAT_BATCH_MEMORY_SIZE).num_val.intValue();
```
ilooner: `options.getLong()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
```diff
@@ -0,0 +1,657 @@
...
+public final class RecordBatchSizerManager {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(RecordBatchSizerManager.class);
+  public static final String BATCH_STATS_PREFIX = "BATCH_STATS";
+
+  /** Minimum column memory size */
+  private static final int MIN_COLUMN_MEMORY_SZ = VarLenColumnBulkInput.getMinVLColumnMemorySize();
+  /** Default memory per batch */
+  private static final int DEFAULT_MEMORY_SZ_PER_BATCH = 16 * 1024 * 1024;
+  /** Default records per batch */
+  private static final int DEFAULT_RECORDS_PER_BATCH = 32 * 1024 -1;
```
ilooner: Instead of having these defaults in the code, you can set them in drill-module.conf. And set the appropriate RangeValidator for your System option. Then you could delete this code

```
    if (normalizedMemorySize <= 0) {
      normalizedMemorySize = DEFAULT_MEMORY_SZ_PER_BATCH;

      final String message = String.format("Invalid Parquet memory per batch [%d] byte(s); using default [%d] bytes",
          configMemorySizePerBatch, normalizedMemorySize);
      logger.warn(message);
    }
```
from normalizeNumRecordsPerBatch since the SystemOptions framework will handle setting the default and validating the option value for you.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
```diff
@@ -0,0 +1,657 @@
...
+    if (memorySizePerColumn < MIN_COLUMN_MEMORY_SZ) {
+      final int prevValue   = normalizedMemorySize;
+      normalizedMemorySize  = MIN_COLUMN_MEMORY_SZ * numColumns;
+
+      final String message = String.format("The Parquet memory per batch [%d] byte(s) is too low for this query ; using [%d] bytes",
+        prevValue, normalizedMemorySize);
+      logger.warn(message);
+    }
+
+    if (logger.isDebugEnabled()) {
```
ilooner: Use

```
logger.debug("{}: The Parquet reader batch memory has been set to [{}] byte(s)", BATCH_STATS_PREFIX, normalizedMemorySize)
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
```diff
@@ -156,16 +151,23 @@ public VariableWidthReader(ReadState readState) {
...
+      // with variable length columns as each might return a different number of records. The batch size
+      // will be the lowest value. The variable column readers will update the "readState" object to
+      // reflect the correct information.
       long fixedRecordsToRead = readState.varLengthReader().readFields(recordsToRead);
       readAllFixedFields(fixedRecordsToRead);
-      return firstColumnStatus.getRecordsReadInCurrentPass();
+
+      // Sanity check the fixed readers read the expected number of rows
+      if (firstColumnStatus != null) {
+        assert firstColumnStatus.getRecordsReadInCurrentPass() == readState.getValuesReadInCurrentPass();
```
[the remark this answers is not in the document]
sachouche: Done!

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
vrozov: Please point me to other places where Drill uses both log level and configuration settings to configure the amount of logging.

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
```diff
@@ -0,0 +1,225 @@
...
+    /** Unique Operator Identifier */
+    private final String contextOperatorId;
+
+    /**
+     * @param options options manager
+     */
+    public RecordBatchStatsContext(FragmentContext context, OperatorContext oContext) {
+      enableBatchSzLogging   = context.getOptions().getOption(ExecConstants.STATS_LOGGING_BATCH_SIZE_OPTION).bool_val;
+      enableFgBatchSzLogging = context.getOptions().getOption(ExecConstants.STATS_LOGGING_FG_BATCH_SIZE_OPTION).bool_val;
+      contextOperatorId      = new StringBuilder()
```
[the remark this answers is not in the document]
vrozov: We discussed such format style in the past and agreed not to use it. Please format using a single space around `=`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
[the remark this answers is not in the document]
vrozov: Consider using `Preconditions` unless it is an extremely "hot" place. Note that Java `assert` translates to a check on a `final boolean` variable.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/OverflowSerDeUtil.java
sachouche: This is what I added:
NOTE - We use a specialized implementation for overflow SERDE (instead of reusing existing ones) because of the following reasons:

- We want to only serialize a subset of the VV data
- Other SERDE methods will not copy the data contiguously and instead rely on the RPC layer to write the drill buffers in the correct order so that they are de-serialized as a single contiguous drill buffer

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/OverflowSerDeUtil.java
sachouche: done!

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
```diff
@@ -0,0 +1,225 @@
...
+    /** Fine grained batch size logging */
+    private final boolean enableFgBatchSzLogging;
+    /** Unique Operator Identifier */
+    private final String contextOperatorId;
+
+    /**
+     * @param options options manager
+     */
+    public RecordBatchStatsContext(FragmentContext context, OperatorContext oContext) {
+      enableBatchSzLogging   = context.getOptions().getOption(ExecConstants.STATS_LOGGING_BATCH_SIZE_OPTION).bool_val;
```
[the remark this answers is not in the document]
sachouche: done.

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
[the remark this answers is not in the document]
sachouche: We could but we might want to get other members feedbacks. I want to avoid doing such a change as part of this JIRA to avoid slowing down the review process.

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * @return the contextOperatorId
+     */
+    public String getContextOperatorId() {
+      return contextOperatorId;
+    }
+
+    private String getQueryId(FragmentContext _context) {
```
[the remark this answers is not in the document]
sachouche: Correct (please see previous comment)

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
sachouche: done.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
sachouche: Performance benchmarks indicated that 4kb is an optimal value; larger values will have the side effect of overflowing from the L1 cache (32kb). Actually, I meant to do this change a while back but somehow missed this one.

exec/vector/src/main/codegen/templates/VariableLengthVectors.java
sachouche: nice catch!

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ParquetSchema.java
```diff
@@ -195,9 +184,6 @@ public BlockMetaData getRowGroupMetadata() {
    */
 
   private boolean fieldSelected(MaterializedField field) {
-    // TODO - not sure if this is how we want to represent this
```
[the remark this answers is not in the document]
sachouche: I couldn't make sense of this TODO:
- The code goes through all columns found in a Parquet file metadata
- The code will read a Parquet column when a) query is a  SELECT_STAR or b) the column has been referenced in the query

My point, why keep a TODO when its meaning cannot be understood?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
sachouche: done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
sachouche: The goal is to allow users per operator custom values and if not then a common value is instead used; to do that, we need to specify a ZERO value in the configuration to figure out whether the user has set a value.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
sachouche: done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
sachouche: done.

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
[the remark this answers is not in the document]
sachouche: @vrozov 
This code was written few months ago before the format feedback. I went through the code and did my best to fix formatting but I have missed some of them.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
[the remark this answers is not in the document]
sachouche: Done and thank you for the tip!

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBinaryReader.java
```diff
@@ -90,22 +91,161 @@ public long readFields(long recordsToReadInThisPass) throws IOException {
...
+
+      // Enqueue this column entry information to handle overflow conditions; we will not know
+      // whether an overflow happened till all variable length columns have been processed
+      columnStats.add(new VarLenColumnBatchStats(columnReader.valueVec, readColumns));
 
+      // Decrease the number of records to read when a column returns less records (minimize overflow)
+      if (batchNumRecords > readColumns) {
+        batchNumRecords = readColumns;
+        // it seems this column caused an overflow (higher layer will not ask for more values than remaining)
+        ++columnContainer.numCausedOverflows;
```
ilooner: Why are we incrementing numCausedOverflows here? If our batchNumRecords is 200 initially, but we only read 100 values from each column is that considered an overflow?

ilooner: Also it looks like it is possible to read different numbers of values from different columns since we are checking `prevReadColumns != readColumns`. Is that correct? If so couldn't we miss setting the overflow flag. Consider the case where `batchNumRecords = 200` and we read 200 from every column except the last one where we read 100. In that case we overflowed, but we didn't set the flag because the loop terminated.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBinaryReader.java
```diff
@@ -90,22 +91,161 @@ public long readFields(long recordsToReadInThisPass) throws IOException {
...
+    // returns less values than num-values-i.
+    RecordBatchOverflow.Builder builder = null;
+
+    // We need to collect all columns which are subject to an overflow (except for the ones which are already
+    // returning values from previous batch overflow)
+    for (VarLenColumnBatchStats columnStat : columnStats) {
+      if (columnStat.numValuesRead > batchNumRecords) {
+        // We need to figure out whether this column was already returning values from a previous batch
+        // overflow; if it is, then this is a NOOP (as the overflow data is still available to be replayed)
+        if (fieldHasAlreadyOverflowData(columnStat.vector.getField().getName())) {
```
ilooner: Thanks for the comments they are helpful. I'm still a little confused about this case though perhaps we can talk offline. Specifically what if a column returned some values from a previous overflow, and some newly read values? Is that not possible?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBinaryReader.java
```diff
@@ -90,22 +91,161 @@ public long readFields(long recordsToReadInThisPass) throws IOException {
...
+          continue;
+        }
+
+        // We need to set the value-count as otherwise some size related vector APIs won't work
+        columnStat.vector.getMutator().setValueCount(batchNumRecords);
```
ilooner: if we added more than batchNumRecords to the vector, the extra records will still be in the vector consuming memory. So even though we don't want to pass down the overflowed records down to the next operator, we are still technically doing it since they are still in the memory occupied by the value vector.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/BatchSizingMemoryUtil.java
```diff
@@ -0,0 +1,327 @@
...
+  }
+
+  /**
+   * Load memory usage information for a variable length value vector
+   *
+   * @param vector source value vector
+   * @param currValueCount current value count
+   * @param vectorMemory result object which contains source vector memory usage information
+   */
+  public static void getMemoryUsage(ValueVector sourceVector,
```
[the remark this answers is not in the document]
ilooner: For posterity. We spoke offline and agreed to keep this code in parquet for now. Things can be generalized and moved into common classes and areas as they are needed in separate changes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/BatchSizingMemoryUtil.java
```diff
@@ -0,0 +1,329 @@
...
+    int currValueCount,
+    VectorMemoryUsageInfo vectorMemoryUsage) {
+
+    assert sourceVector instanceof VariableWidthVector;
+
+    vectorMemoryUsage.reset(); // reset result container
+
+    final MajorType type = sourceVector.getField().getType();
+
+    switch (type.getMinorType()) {
```
ilooner: Are we not going to support repeated types?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/OverflowSerDeUtil.java
```diff
@@ -0,0 +1,366 @@
...
+
+      bufferLength += fieldVLSerDe.getBytesUsed(fieldOverflowEntry.firstValueIdx, fieldOverflowEntry.numValues);
+    }
+    assert bufferLength >= 0;
+
+    // Allocate the required memory to serialize the overflow fields
+    final DrillBuf buffer = allocator.buffer(bufferLength);
+
+    if (logger.isDebugEnabled()) {
+      logger.debug(String.format("Allocated a buffer of length %d to handle overflow", bufferLength));
```
ilooner: No need for String.format

```
logger.debug("Allocated a buffer of length {} to handle overflow", bufferLength)
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenColumnBulkInput.java
```diff
@@ -160,11 +191,41 @@ public final void done() {
...
-     || !pageInfo.dictionaryValueReader.isDefined()) {
+        || !pageInfo.dictionaryValueReader.isDefined()) {
       parentInst.pageReader.readyToReadPosInBytes = oprReadState.pageReadPos;
     }
     parentInst.pageReader.valuesRead = oprReadState.numPageFieldsProcessed;
-    parentInst.totalValuesRead += oprReadState.batchFieldIndex;
+    parentInst.totalValuesRead += oprReadState.batchNumValuesReadFromPages;
+
+    if (logger.isDebugEnabled()) {
+      String message = String.format("requested=%d, returned=%d, total-returned=%d",
```
ilooner:
```
logger.debug(requested={}, returned={}, total-returned={}",recordsToRead, oprReadState.batchFieldIndex, parentInst.totalValuesRead);
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/OverflowSerDeUtil.java
```diff
@@ -0,0 +1,366 @@
...
+    }
+
+    /**
+     * The number of bytes used (by the {@link ValueVector}) to store a value range delimited
+     * by the parameters "firstValueIdx" and "numValues".
+     *
+     * @param firstValueIdx start index of the first value to copy
+     * @param numValues number of values to copy
+     */
+    private int getBytesUsed(int firstValueIdx, int numValues) {
```
ilooner: We don't have to do this now, but we should probably move this method into the value vectors eventually. They already have

```
valueVector.getPayloadByteCount(numValues);
```

Which does basically the same thing except it does not let you specify the start index.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
sachouche: The default values have been added previously before we agreed on a common operator configuration. I have removed these default values.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/batchsizing/RecordBatchSizerManager.java
```diff
@@ -0,0 +1,667 @@
...
+          AllocationHelper.allocate(v, recordsPerBatch, columnMemoryInfo.columnPrecision, 0);
+        } else {
+          // This column was found in another Parquet file but not the current one; so we inject
+          // a null value. At this time, we do not account for such columns. Why? the right design is
+          // to create a ZERO byte all-nulls value vector to handle such columns (there could be hundred of these).
+          AllocationHelper.allocate(v, recordsPerBatch, 0, 0); // the helper will still use a precision of 1
+        }
+      }
+    } catch (NullPointerException e) {
+      throw new OutOfMemoryException();
```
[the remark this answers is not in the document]
sachouche: This code is the original code that I have moved from the Schema class into this one.

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
```diff
@@ -0,0 +1,225 @@
...
+     */
+    public boolean isEnableFgBatchSzLogging() {
+      return enableFgBatchSzLogging || logger.isDebugEnabled();
+    }
+
+    /**
+     * @return indicates whether stats messages should be logged in info or debug level
+     */
+    public boolean useInfoLevelLogging() {
+      return isEnableBatchSzLogging() && !logger.isDebugEnabled();
```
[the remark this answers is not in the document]
sachouche: please see response above.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -482,6 +482,8 @@ drill.exec.options: {
     exec.storage.enable_new_text_reader: true,
     exec.udf.enable_dynamic_support: true,
     exec.udf.use_dynamic: true,
+    drill.exec.stats.logging.batch_size: false,
+    drill.exec.stats.logging.fine_grained.batch_size: false,
```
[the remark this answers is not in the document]
sachouche: Please refer to the response above.

vrozov:
I am not convinced that a user should be able to change the amount of logging from a client session. A user may cause an excessive logging and bring Drill cluster down. In addition, an end user may not have access to drillbit logs, so changing log level is useless. Ability to change log levels must be granted only to an admin who has access to logback.xml and logs.

sachouche:
@vrozov 
The logging functionality is not exposed to end users. All batch stats flags are internal.

vrozov:
@sachouche I refer to your following comment:
>User should be able to control batch statistics behavior within a client session: a) what operator stats to log and b) coarse vs fine grained stats

In any case, logging should be managed by a single entity (slf4j and logback.xml). Having two different ways to manage logging is confusing and error prone.

sachouche:
- Updated code based on most recent comments
- Fixed a compilation failure due to changes in the Batch Sizer APIs made in recently committed PRs

ilooner:
@vrozov I discussed the logging with @sachouche. The issue is that some information needed to be logged for QA purposes on a per query basis, and for ease of testing needed to be toggled from the sqlline. This was the agreed on approach by the batch sizing team, so at this point we're coming in at the end of a series of discussions. Nevertheless I spent about half a day to find a way this use case could be met in a logback centric way, and I could not find a reasonable alternative that could be done without a large amount of work.

The main issue is that slf4j / logback has global logging configurations, which makes it impossible to turn on logging on a per query bases. We can use MDC to include a query id in the log messages and we could filter by query, but even that is not ideal since you are generating logs you don't need for all queries. Then there is the issue of updating the logging configuration. The only logback native method is to enable periodic scanning of the logback.xml file and to update the logback.xml file on all nodes when you want to enable certain logging. That is not a great experience for someone automating tests, or debugging.

The conclusion we came to was that Drill needs a real logging solution, and to really do it right you would have to go down the rabbit hole.

  1. Enable per query logging configurations, not just a global logging configuration. To do this we would have to use thread local loggers instead of static loggers, or something like https://clogr.io/, or a custom ContextSelector https://stackoverflow.com/questions/38208617/practical-use-of-logback-context-selectors.
  2. Allow either sql line configuration of log levels or web ui configuration of log levels (like Apex had).
  3. Include a logging context with the query context to communicate the logging config for a query.

These are non trivial, have a lot of corner cases, and require design discussions with the community.

For the interim Salim is making the options controlling logging internal, so users will not see them. Then he is going to have a follow up PR to consolidate the logging across all the batch sizing operators. Later when we have a real logging solution we can delete these options and do the right thing.

vrozov:
If logging needs to be enabled just for QA purposes, I still don't see why it needs to be enabled per query or be dynamic and can't be managed by logging.xml. Additionally, if I understand "QA purpose", batch sizes needs to be validated by unit testing, not by looking at the logs.

ilooner:
@vrozov My understanding was the following. QA has setup automatic tests of both the performance of batch sizing as well as correctness on a real cluster. Each batch sizing change has unit tests to validate batch size. But on a real cluster with real data, the only viable way to validate right now for QA is to check the batch sizes output by an operator is through logging. Since Drill takes testing on real clusters seriously and aims to do more than just unit tests, I think this is perfectly acceptable.

Since logging has overhead, and QA wanted to automate both the performance and correctness tests, they required the ability to turn logging off via sql line. This was the approach agreed on by developers and testers in the Drill community including @sachouche, @bitblender, @ppadma, robert (don't know his github id), and @priteshm.

Given the scope of agreement in the community, the fact that similar changes have already been merged, and also to minor impact on the drill code itself ~20 lines; I suggest moving this discussion to a separate change. In my investigation I was not able to find a viable alternative to this approach, @vrozov perhaps you could present an alternative approach on the dev list and lead the proposal. It would be a great help moving forward.

In the meantime the changes proposed here represent a valuable performance improvement for the Drill community, so let's not hold up this change over this.

PR #1333, opened 2018-06-25
DRILL-6410: Memory leak in Parquet Reader during cancellation

vrozov:
@parthchandra @sachouche @ilooner Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
@@ -23,16 +23,13 @@ import java.io.EOFException;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.common.exceptions.UserException;
+import org.apache.drill.exec.util.ExecutableTasksLatch;
 import org.apache.drill.exec.util.filereader.DirectBufInputStream;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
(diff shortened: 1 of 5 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -83,21 +80,18 @@
...
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(AsyncPageReader.class);
 
-  private ExecutorService threadPool;
-  private long queueSize;
-  private LinkedBlockingQueue<ReadStatus> pageQueue;
-  private ConcurrentLinkedQueue<Future<Void>> asyncPageRead;
-  private long totalPageValuesRead = 0;
-  private Object pageQueueSyncronize = new Object(); // Object to use to synchronize access to the page Queue.
-                                                     // FindBugs complains if we synchronize on a Concurrent Queue
+  private final ExecutableTasksLatch<AsyncPageReaderTask> executableTasksLatch;
```
sachouche: The ExecutableTasksLatch replaces the ExecutorService for concurrent task manipulation:
- Is this the recommended way moving forward?
- There is no javadoc for the ExecutableTasksLatch; you need to document this new class APIs and execution semantics. This is especially true if we want to encourage the use of this implementation pattern.

vrozov: `ExecutableTasksLatch` encapsulates `ExecutorService` for parallel tasks that require a synchronization point at cancellation. For all other tasks `ExecutorService` can be used directly.

I'll add more documentation after the first round of review.

priteshm:
@ilooner can you also take a look at it?

ilooner:
@priteshm will take a look monday.
@vrozov please fix conflict and travis failures.

vrozov:
@ilooner I'll rebase after review. Travis CI failure is not related to the change, it failed due to the build exceeding Travis CI limit.

vrozov:
@ilooner I guess by "last discussion" you refer to the discussion between you, me and @sachouche, where "majority" does not mean the community majority. In the Apache, any contributor can provide a solution that (s)he considers to be the best solution possible and then it can either be accepted by the community/contributor or blocked with -1 (requires technical justification). If another contributor provides an alternative solution, a community may decide to go with the alternate solution as long as it addresses technical concerns of the initial contribution. For this particular case, my requirements are a) a unified approach (@parthchandra has the same requirement) and b) the ability to cancel tasks asynchronously. If that can be done with the approach outlined in PR #1257 and a contributor will change it to address all the issues, let's move forward with the alternate approach.

A note regarding the complexity of the implementation. This implementation uses public java concurrency classes as well. It does not rely on unsupported or unsafe to use Java classes and/or API. Basically, `LockSupport` is the same first-class concurrency construct as `Thread` or 'CountDownLatch` classes. The primary use case for using those constructs is to create a combination of `ExecutorService` and a `CountDownLatch` that is not provided by the Java itself.

To summarize, I am perfectly fine to go with an alternate solution or with another committer to review the PR, it will be against Apache way to force a committer to review or commit a change, that (s)he is not comfortable with.

vrozov:
@parthchandra Please review

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
parthchandra: The documentation would be useful for the reviewer too. It is easier to check the code if one knows what what guarantees ExecutableTasksLatch is supposed to provide.

parthchandra: Note that the class comment for AsyncPageReader needs to be updated too.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -284,30 +278,9 @@ protected void nextInternal() throws IOException {
...
-        } else {
-          f.get(1, TimeUnit.MILLISECONDS);
-        }
-      } catch (RuntimeException e) {
-        // Do Nothing
-      } catch (Exception e) {
-        // Do nothing.
-      }
-    }
+    executableTasksLatch.await(() -> true);
```
parthchandra: Isn't the original code doing the same thing that `await` is doing? TBH, I'd really like to understand where the memory leak was occurring. (Just trying to understand how this PR, does, in fact, fix the issue)

exec/java-exec/src/main/java/org/apache/drill/exec/util/ExecutableTasksLatch.java
```diff
@@ -0,0 +1,255 @@
...
+        LockSupport.park();
+      }
+    }
+  }
+
+  /**
+   * Helper class to wrap {@linkplain Callable}{@literal <Void>} with cancellation and waiting for completion support
+   *
+   */
+  public static final class ExecutableTask<C extends Callable<Void>> implements Runnable {
```
parthchandra: Can you elaborate what this class has that is not already available in `FutureTask`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
vrozov: No, `ExecutableTasksLatch.await()` guarantees that when it returns all tasks submitted for execution are either done or canceled. `Future.cancel()` does not wait for the `FutureTask` to be canceled as it merely interrupts the thread where `FutureTask` is running (in the case it is already running). Note that after `Future` is canceled it is not possible to check whether it is finished or not (`Future.get` throws `CancellationException`).

Missing guarantee that all tasks are finished when `clear()` returns, so some tasks continue to run and reference vectors, so allocator reports a memory leak.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ExecutableTasksLatch.java
vrozov: Please see my comment for `ExecutableTasksLatch.await()`.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ExecutableTasksLatch.java
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  /**
+   * submits a task for execution by {@linkplain ExecutorService}
+   * @param callable task to execute
+   * @return newly created instance of {@linkplain ExecutableTask}
+   */
+  public ExecutableTask<C> execute(C callable) {
+    ExecutableTasksLatch.ExecutableTask<C> task = new ExecutableTasksLatch.ExecutableTask<>(callable, this);
+    executor.execute(task);
```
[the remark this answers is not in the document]
vrozov: I doubt that `Future` would be faster, but it is definitely more tested. Please see my comment why `Future` and `FutureTask` can't be used.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
vrozov: - IMO, most of the time, the source code should document itself and java documentation is necessary for an API distributed as an already compiled library/jar.
- I would prefer reviewers to point to obscure code rather than to rely on the documentation.
- Waiting for a review comments helps to avoid inconsistency between the code and the documentation (quite a common problem) as usually code evolves faster and documentation lags behind. 
- I already added documentation for `ExecutableTasksLatch` and `ExecutableTask`.
- I'll change comments for `AsyncPageReader` during rebase and merge conflict resolution.

vrozov:
@parthchandra rebased and resolved conflicts. Please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
parthchandra: Seems to me that we could have extended FutureTask to provide this guarantee. Perhaps that is what @ilooner would have liked too.
See this : https://stackoverflow.com/questions/6040962/wait-for-cancel-on-futuretask

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
vrozov: Please be more specific. Do you refer to `AsyncPageReader` doc? It was already changed during the merge conflict resolution.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
vrozov: I know and saw that post before the fix for the same problem was implemented only for `PartitionerDecorator`. I decided to go with a similar approach outlined in the post, but that does not use `FutureTask`to support additional functionality that I outlined in my response to @ilooner. This implementation supports asynchronous cancellation (call to `cancel()` does not block waiting for a task to complete allowing a faster cancellation) that the solution that extends `FutureTask` does not provide.

priteshm:
@vrozov Dechang is running a perf test to confirm the question from @parthchandra. Once that is done, we should merge to the branch. Was there any other reason to close it?

PR #1334, opened 2018-06-25
DRILL-6385: Support JPPD feature

weijietong:
@amansinha100 @parthchandra The join predicate push down feature is basically ready to review. Please give some review advices, thanks.


weijietong:
ping @amansinha100 @parthchandra or someone else would be pleased to review this PR.

amansinha100:
@weijietong thanks for the PR.  Yes, this is high on my list to review.  I am tied up until tomorrow but will review during the weekend.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/config/HashJoinPOP.java
```diff
@@ -32,35 +33,53 @@
...
 import org.apache.drill.exec.physical.base.AbstractJoinPop;
+import org.apache.drill.exec.work.filter.RuntimeFilterDef;
+
 
 @JsonTypeName("hash-join")
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class HashJoinPOP extends AbstractJoinPop {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(HashJoinPOP.class);
 
+  private RuntimeFilterDef runtimeFilterDef;
```
chunhui-shi: Who is responsible for constructing this runtimeFilterDef?

chunhui-shi: After physical plan is generated, RuntimeFilterManager is is responsible for constructing this runtimeFilterDef. So I guess your estimation of the cost sending bloom filter is trivial.

weijietong: RuntimeFilterDef was created after the physical plan was generated. Constructing the BloomFilter at HashJoin node is a heavy work， to a light weight join query (i.e. two small tables), it's not applicable to construct a bloom filter.So I left some TODO work at the RuntimeFilterManager.quelifiedHolders method to value whether applicable to apply the RuntimeFilter according to the NDV from both sides. The NDV statistics is not supported now. I will fire another work about the Metadata to support it.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -226,6 +244,96 @@ public IterOutcome next() {
...
+      recordCount = svIndex;
+      BatchSchema batchSchema = this.schema;
+      VectorContainer backUpContainer = new VectorContainer(this.oContext.getAllocator(), batchSchema);
+      int fieldCount = batchSchema.getFieldCount();
+      for (int i = 0; i < fieldCount; i++) {
+        ValueVector from = this.getContainer().getValueVector(i).getValueVector();
+        ValueVector to = backUpContainer.getValueVector(i).getValueVector();
+        to.setInitialCapacity(svIndex);
+        for (int r = 0; r < svIndex; r++) {
+          to.copyEntry(r, from, selectionVector2.getIndex(r));
```
amansinha100: To summarize the way you have implemented this part:  suppose the original ScanBatch contained 100 rows and 10 of them qualified the bloom filter, you create an SV2 of size 10, set the original qualifying row's index in the SV2, then for each ValueVector you copy the qualifying row's data into the backupContainer, followed by exchanging it with the current output container.   A couple of thoughts about this: 
  - Why not produce the SV2 in the output batch and let the downstream operator handle it ? Quite often there may be a Filter operator above the Scan  which would be applying other filters (i.e not the run-time filters) and it can combine its own SV2 with the SV2 produced by the Scan.  That way you avoid the extra copying and let Filter handle it (note that Filter does code-gen so it is more efficient way to handle bulk filtering). 
 - There is clearly trade-offs with the extra copy approach : it depends on selectivity, i.e how many rows actually get eliminated by the run-time filter.  I suppose you have mentioned this as a TODO depending on the NDV statistics ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -696,6 +780,18 @@ public void executeBuildPhase() throws SchemaChangeException {
...
           read_right_HV_vector = (IntVector) buildBatch.getContainer().getLast();
         }
+        //create runtime filter
+        if (cycleNum == 0 && enableRuntimeFilter) {
+          //create runtime filter and send out async
+          int condFieldIndex = 0;
+          for (BloomFilter bloomFilter : bloomFilters) {
+            for (int ind = 0; ind < currentRecordCount; ind++) {
+              long hashCode = hash64.hash64Code(ind, 0, condFieldIndex);
+              bloomFilter.insert(hashCode);
```
amansinha100: Currently, the hash join relies on memory calculations (both build and probe sides) to do accounting for spilling purposes.  The bloom filter memory use should also be included in that calculation, although it would fine if you create an enhancement JIRA and address it separately.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
weijietong: @aman thanks for other valuable reviews ,will soon update that. To this point, I initially tend to output the SV2. But to fellow reasons, I give up: 

* Not all the possible operators support the SelectionModel. If users’ rule pushed down the filter conditions,the output SV2 maybe not processed by the not supported operators(i.e. it’s not definitive  to have a operator which supports the SelectionModel above the Scan node).

* The BatchSchema’s SelectionModel also becomes a runtime var.This will also affect the upper filter node’s code-gen logic to dynamically generate fresh filter codes to the Scan’s SV2.

I agree that the memory copy cost will be less if the above filter node can filter more rows over the Scan’s SV2.So what your opinion about this?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/config/HashJoinPOP.java
```diff
@@ -32,35 +33,53 @@
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
 import org.apache.drill.exec.physical.base.AbstractJoinPop;
+import org.apache.drill.exec.work.filter.RuntimeFilterDef;
+
 
 @JsonTypeName("hash-join")
+@JsonIgnoreProperties(ignoreUnknown = true)
```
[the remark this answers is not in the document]
weijietong: The primary reason is to backward compatible. PhysicalPlans without `runtimeFilterDef` property (maybe not support RuntimeFilter) can still work out. That is the `ignoreUnknown=true` usage.  The `@JsonIgnore` is to not output the property content as the final json. That's not our target.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
amansinha100: Hi @weijietong ,  the physical plan operators are supposed to implement 2 APIs:  `getSupportedEncodings()` API (see [1]) which indicates whether or not it accepts an SelectionVector2 (other options are SV4 and NONE) and the corresponding `getEncoding()` API.  If an operator does not accept an SV2 or SV4, the planner will insert a SelectionVectorRemover just below that node which essentially does the copying of qualified rows.  Note that the SVRemover is implemented by RemovingRecordBatch [2] which uses a `StraightCopier` whenever the child has NONE, so in that case it just does a simple transfer instead of copy.  

In the case of Filter,  it can accept both NONE and SV2 because it is possible in some cases to have a filter on top of another filter (with some intermediate non-blocking operator).  

Let me know if this makes sense.   The main reason I am proposing this is the Filter-Scan is a very common pattern and we would want to minimize the copy overhead as much as possible.  Unfortunately, I am tied up with some other work otherwise I could make the changes on top of your branch to experiment. 

[1] https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/Prel.java#L51
[2] https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/RemovingRecordBatch.java#L57

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
weijietong: I will enhance this at another JIRA.

weijietong:
@amansinha100  The scan node's memory copy logic has removed. Thanks for the knowledge of `SelectionVectorPrelVisitor.addSelectionRemoverWhereNecessary`. I appreciate what you payed out. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,755 @@
...
+    private int probeSideScanMajorId;
+
+    private int probeSideScanOpId;
+
+    private PhysicalOperator probeSideScanOp;
+
+    private List<CoordinationProtos.DrillbitEndpoint> probeSideScanEndpoints;
+
+    //probe side's children have blocked nodes.
+    private boolean meetBlockedOperator;
```
amansinha100: Better name would be 'probeHasBlockingOperator'

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,755 @@
...
+
+    //whethe this join operator is a partitioned HashJoin or broadcast HashJoin
```
amansinha100: In addition to broadcast and partitioned hash join, there's a third state also which is non-distributed hash join (this depends on slice_target, estimated row count etc.).  Is the JPPD not applied for such cases ?

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,755 @@
...
+    private boolean meetBroadcastExchange;
+
+    private boolean meetPartitionedExchange;
```
amansinha100: Better names would be `encounteredPartitionExchange`,  `encounteredBroadcastExchange`.   But do you really need these variables declared in the HashJoinOpHolder class since these are temporary states encountered during the traversal of the operator tree and you are anyways having another class variable `isBroadcastHashJoin`.

amansinha100:
@weijietong I have couple of design questions that are better discussed in the JIRA, so I will add them there.. pls take a look.  Thanks.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ScanPrel.java
```diff
@@ -151,12 +151,12 @@ public RelOptCost computeSelfCost(final RelOptPlanner planner, RelMetadataQuery
 
   @Override
   public SelectionVectorMode[] getSupportedEncodings() {
-    return SelectionVectorMode.DEFAULT;
+    return SelectionVectorMode.NONE_AND_TWO;
```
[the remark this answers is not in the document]
weijietong: Yes, I will update that.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ScanPrel.java
```diff
@@ -151,12 +151,12 @@ public RelOptCost computeSelfCost(final RelOptPlanner planner, RelMetadataQuery
...
   }
 
   @Override
   public SelectionVectorMode getEncoding() {
-    return SelectionVectorMode.NONE;
+    return SelectionVectorMode.TWO_BYTE;
```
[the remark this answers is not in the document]
weijietong: What do you mean by `NONE_AND_TWO ` ,since the return is `SelectionVectorMode` not an array. If JPPD is not applicable, to `ScanBatch` ,the output container's build schema would always be `SelectionVectorMode.NONE` , and will not generate SV2.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/OptionValue.java
```diff
@@ -76,7 +77,7 @@ public boolean inScopeOf(OptionScope scope) {
   }
 
   public enum Kind {
-    BOOLEAN, LONG, STRING, DOUBLE
+    BOOLEAN, LONG, STRING, DOUBLE, INTEGER
```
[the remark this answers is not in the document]
weijietong: The reason to add an INTEGER option is that the largest BF's byte size is supposed to be Integer.MAX_VALUE which is limited by DrillBuf's size.  So INTEGER option would be precisely.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -375,6 +378,30 @@ protected void cleanup() {
...
     }
+
+    public void receiveRuntimeFilter(final RuntimeFilterWritable runtimeFilter) {
+      BitData.RuntimeFilterBDef runtimeFilterDef = runtimeFilter.getRuntimeFilterBDef();
+      boolean toForeman = runtimeFilterDef.getToForeman();
+      QueryId queryId = runtimeFilterDef.getQueryId();
+      //to foreman
+      if (toForeman) {
+        Foreman foreman = queries.get(queryId);
+        if (foreman != null) {
```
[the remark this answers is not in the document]
weijietong: Yes, it's supposed to handle the cases: the foreman has crashed or the whole query has completed ahead of time.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,755 @@
...
+    }
+  }
+
+
+  private static List<BloomFilterDef> constructBloomFilterDefs(HashJoinOpHolder holder, int bloomFilterSizeInBytes, boolean sendToSelf) {
+    List<JoinCondition> joinConditions = holder.hashJoinPOP.getConditions();
+    List<BloomFilterDef> bloomFilterDefs = new ArrayList<>();
+    for (JoinCondition joinCondition : joinConditions) {
+      LogicalExpression leftProExp = joinCondition.getLeft();
+      FieldReferenceFinder fieldReferenceFinder = new FieldReferenceFinder();
```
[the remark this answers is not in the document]
weijietong: You are right, will update that to consider a join condition with function.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
weijietong: You mean there's only one HashJoin execution node? To that case , I guess JPPD is not necessary, since non-distributed one means less data, it's not cost efficient to generate a BF.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/ScanPrel.java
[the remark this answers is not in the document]
amansinha100: Makes sense.  I was thinking of the other API which returns an array.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
amansinha100: Yeah, it would be fine to not do JPPD for that case .. we can document the expected behavior.

amansinha100:
@weijietong thanks for making these changes.  I will do one final round of review hopefully soon and let you know.

weijietong:
ping @amansinha100

amansinha100:
Hey @weijietong,  sorry for the delay; I will send you my final comments by tomorrow.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilterCreator.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.work.filter;
+
+import io.netty.buffer.DrillBuf;
+import org.apache.drill.exec.memory.BufferAllocator;
+
+public class BloomFilterCreator {
```
amansinha100: Could you pls add a description either here or another appropriate place about how the BF creation works for a generalized multi-column join.. e.g  t1.a1 = t2.a2 AND t1.b1 = t2.b2.   In this case, assume t2 is the build side.  My understanding is as follows: 
  - create a compound hash value using both build side columns: hash(a2, hash(b2))   (this may not be the exact way it is being created but same general idea).
  - add this to the BF, setting appropriate bits to 1
  - on probe side, do the same for a1, b1 and check BF for either false/possibly true

Let me know if this is correct.   This also implies that the size of BF is independent of the number of columns being joined.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -455,6 +455,8 @@ drill.exec.options: {
     exec.hashjoin.num_partitions: 32,
     exec.hashjoin.num_rows_in_batch: 1024,
     exec.hashjoin.max_batches_in_memory: 0,
+    exec.hashjoin.enable.runtime_filter: true,
```
amansinha100: I would suggest to keep the default value as false until the feature goes through performance and functional testing.

amansinha100:
@weijietong  thanks for your patience.  I have submitted a final round of review comments.   Once you address them we can get the feature merged into 1.15.0 such that it gives it enough time to go through the normal functional and performance testing.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,586 @@
...
+  }
+
+  /**
+   * Find a join condition's left input source scan Prel. If we can't find a target scan Prel then this
+   * RuntimeFilter can not pushed down to a probe side scan Prel.
+   * @param fieldName left join condition field Name
+   * @param leftRelNode left RelNode of a BiRel or the SingleRel
+   * @return a left scan Prel which contains the left join condition name or null
+   */
+  private static ScanPrel findLeftScanPrel(String fieldName, RelNode leftRelNode) {
```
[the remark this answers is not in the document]
weijietong: Due to the fact that HashJoin's buildSchema() method will prefetch both sides first batch data from both sides, if there's a blocking node under the probe side one , the corresponding probe side scan node will already run over before the build side bloom filter generated. So I will update that to still keep the blocking node checking logic.

exec/java-exec/src/main/resources/drill-module.conf
weijietong: agree

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,586 @@
...
+    if (leftRelNode instanceof ScanPrel) {
+      RelDataType scanRowType = leftRelNode.getRowType();
+      RelDataTypeField field = scanRowType.getField(fieldName, true, true);
+      if (field != null) {
+        //found
+        return (ScanPrel)leftRelNode;
+      } else {
+        return null;
+      }
+    } else {
```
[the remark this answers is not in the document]
weijietong: will update that.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilterCreator.java
weijietong: Current implementation is one bloom filter one join column.  To your example, multi-column join , will generate two bloom filters.  The reason to this implementation is to achieve one vector column memory access by one  hash computation. But the Murmur hash 's complex computation ate up the pipeline performance , the result performance does not so good. So I will change it to your assumes implementation.

weijietong:
@amansinha100 thanks for your valuable review. Since being on vacation , others will be commented and updated later.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
```diff
@@ -0,0 +1,176 @@
...
+      numBytes = DEFAULT_MAXIMUM_BLOOM_FILTER_SIZE;
+    }
+
+    // 32 bytes alignment, one bucket.
+    numBytes = (numBytes + 0x1F) & (~0x1F);
+    return numBytes;
+  }
+
+  private void setMask(int key) {
+    final int SALT[] = {0x47b6137b, 0x44974d91, 0x8824ad5b, 0xa2b7289d, 0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31};
```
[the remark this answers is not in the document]
weijietong: those 8 odd numbers was picked from Impala's implementation(see bloom-filter.h) . They act as salt values to participate in calculating the mask. I will add some comments to this method to make it clear.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
```diff
@@ -0,0 +1,176 @@
...
+ * According to Putze et al.'s "Cache-, Hash- and Space-Efficient BloomFilter
+ * Filters", the main theory is to construct tiny bucket bloom filters which
+ * benefit to the cpu cache and SIMD opcode.
+ */
+
+public class BloomFilter {
+  // Bytes in a bucket.
+  private static final int BYTES_PER_BUCKET = 32;
+  // Minimum bloom filter data size.
+  private static final int MINIMUM_BLOOM_SIZE = 256;
```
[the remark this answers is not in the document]
weijietong: BYTES is right. will update that naming.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
```diff
@@ -0,0 +1,176 @@
...
+
+  private static final int DEFAULT_MAXIMUM_BLOOM_FILTER_SIZE = 16 * 1024 * 1024;
+
+  private DrillBuf byteBuf;
+
+  private int numBytes;
+
+  private int mask[] = new int[8];
+
+  private byte[] tempBucket = new byte[32];
```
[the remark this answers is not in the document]
weijietong: The reason to keep it as a member variable is to avoid heap memory GC while invoking the corresponding methods frequently. If move it to the locally methods, it will break down the performance .

weijietong:
To the cleanup point, I currently think there's no problem. Once the query was canceled, the foreman or the drillbit will run out , the in-flight runtime filter will failed due to the wire termination of one side.  The off-heap memory occupied by the bloom filter will be released by the promise of using `SendingAccountor` to send out `RuntimeFilter` through DataTunnel. If we find any potential issues later , we can fixed them at the following JIRAs.

weijietong:
@amansinha100 please review the changes made according to your advices.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilterCreator.java
amansinha100: Hmm...if the 2 columns create separate bloom filters, where does the ANDing (intersection) of the two BFs happen ?  In the current implementation there's an ORing of the BFs but I did not see ANDing (maybe I missed, so let me know).  As far as the memory usage is concerned, we can continue down this path and evaluate both CPU cost and memory usage for multi-column joins where the cardinality of the columns ranges from low to high.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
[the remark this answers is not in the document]
amansinha100: Sounds good..but do you need to clear the tempBucket array across invocations ?

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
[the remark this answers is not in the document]
amansinha100: Although I understand these are open source software but in general it is better to avoid referencing another project's implementation.  In this case, it is probably ok since picking these values is similar to picking prime numbers for hash computations and typically they are common across implementations.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilterCreator.java
weijietong: There's no intersection of the two separate BFs. The intersection meaning is implemented indirectly at the probe side `ScanBatch`. The separated BFs filter the probe side columns independently and set a BitSet (which indicates the possible rows matched the join condition ) together. This part of codes is at the `ScanBatch`'s `applyRuntimeFilter` and `computeBitSet` methods.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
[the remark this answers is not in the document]
weijietong: The `DrillBuf.getBytes(int index, byte[] dst)` will override the dst array's content. And we always invoke the getBytes method before usage the dst array. So there's no risk to use a polluted byte array.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilterCreator.java
weijietong: Btw, if there's too much multi-column joins , we can add an option to limit the generated  BloomFilter numbers. That's to say there's no need to generate one BF one joint column. We can choose to abandon generating some BFs to save the CPU cost and memory usage.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -528,6 +590,27 @@ private void initializeBuild() {
...
+  private void initializeRuntimeFilter() {
+    if (!enableRuntimeFilter) {
+      return;
+    }
+    if (runtimeFilterReporter != null) {
+      return;
+    }
+    runtimeFilterReporter = new RuntimeFilterReporter((ExecutorFragmentContext) context);
+    RuntimeFilterDef runtimeFilterDef = popConfig.getRuntimeFilterDef();
+    if (runtimeFilterDef != null) {
```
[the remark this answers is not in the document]
weijietong: `enableRuntimeFilter` is a global option to decide whether a HashJoin could generate RuntimeFilter. But it does not indicate the HashJoin must have a RuntimeFilterDef. There's some preconditions defined at the `RuntimeFilterManager` such as Left join is not allowed to use a BF to filter data. So we can't use `Preconditions.checkState(runtimeFilterDef != null)`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
```diff
@@ -0,0 +1,176 @@
...
+
+
+  public BloomFilter(DrillBuf byteBuf) {
+    this.byteBuf = byteBuf;
+    this.numBytes = byteBuf.capacity();
+    this.byteBuf.writerIndex(numBytes);
```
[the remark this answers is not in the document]
weijietong: the DrillBuf's content will be changed by calling setBytes and getBytes which not affect the readerIndex and writerIndex. When the BloomFilter was created, the DrillBuf's size is sure. The writerIndex setting here in advance is also clear and sure.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilterCreator.java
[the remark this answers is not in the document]
weijietong: make sense.

amansinha100: It would be good to add javadoc with this description of multi-column join and how the implicit intersection is done later during BF application in the Scan rather than at the time of BF creation.   This is different from normal hash join where a single hash value is created for multi-column during creation and probing.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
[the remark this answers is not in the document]
amansinha100: This would also depend on the NDV once we have proper stats.

amansinha100:
Since most of my comments have been addressed,  I am +1.   Thanks @weijietong.  
Once you address @sohami 's comments,  the next steps would be: 
  - pls file JIRAs for items we identified during code review. 
 - rebase on master and run unit tests
 - mark the JIRA ready-to-commit
 - we'll need a volunteer to run the regression tests on QA test cluster (I will ask for volunteer)
 - presumably with the default FALSE setting for the JPPD feature, tests are clean, then it can be merged to master. 
 - continue further tests with a TRUE setting

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilterCreator.java
weijietong: This description will be add at the `ScanBatch`'s `applyRuntimeFilter `.

weijietong:
@amansinha100 The following JIRAs: DRILL-6573 ,DRILL-6572 would solve what we identified before .  @sohami 's comments have been addressed.

amansinha100:
@sohami pls see if your comments are addressed.  @weijietong just FYI, the ongoing 1.14.0 release is yet to be finished (it is in voting stage), so there may be some delays in  running regression tests on the QA cluster with your PR.

sohami:
Sorry for delayed response but I promise to complete review by this weekend.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -375,6 +378,30 @@ protected void cleanup() {
...
+          foreman.getRuntimeFilterManager().registerRuntimeFilter(runtimeFilter);
```
sohami: the `registerRuntimeFilter` and eventual broadcast all is happening on rpc thread which is not good since Rpc thread should be released ASAP. Please follow a model similar to **cancelForeman** in WorkManager to submit the `registerRuntimeFilter` task as new Runnable to WorkManager executor

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -408,6 +424,9 @@ private void runPhysicalPlan(final PhysicalPlan plan) throws ExecutionSetupExcep
     fragmentsRunner.setFragmentsInfo(work.getFragments(), work.getRootFragment(), work.getRootOperator());
 
     startQueryProcessing();
+    if (enableRuntimeFilter) {
+      runtimeFilterManager.waitForComplete();
```
sohami: Why is this needed here ?

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,666 @@
...
+      }
+      if (currentPrel instanceof StreamAggPrel) {
+        encounteredBlockNode = true;
+        return null;
+      }
+
+      if (currentPrel instanceof HashAggPrel) {
+        encounteredBlockNode = true;
+        return null;
+      }
```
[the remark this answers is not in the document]
weijietong: The blocking operators we care about here are those will wait for all the next batches to come in to output the result batches. So the `Sort`, `TopN` need to be considered. Thanks for noticing that, will update that. `Join` I think is not proper here . The `HashJoin`'s probe stage is not a blocking process. The `MergeJoin` is also not a blocking one.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,666 @@
...
+          if (holder == null) {
+            holder = new RFHelperHolder();
+            holders.add(holder);
+          }
+          holder.setRuntimeFilterDef(runtimeFilterDef);
+          GroupScan probeSideScanOp = runtimeFilterDef.getProbeSideGroupScan();
+          Wrapper container = findPhysicalOpContainer(rootWrapper, hashJoinPOP);
+          int majorFragmentId = container.getMajorFragmentId();
+          holder.setJoinMajorId(majorFragmentId);
+          Wrapper probeSideScanContainer = findPhysicalOpContainer(container, probeSideScanOp);
```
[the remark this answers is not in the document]
weijietong: Could you explain why this will return null. To my understanding , once the code reaches here,  the HashJoinPoP is certain to own a `Wrapper`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
weijietong: To make sure the `RuntimerFilterManager`'s sending out  `ByteBuf`s to be safely cleared out no matter the network successed or failed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -190,11 +213,21 @@ public IterOutcome next() {
         if (isNewSchema) {
           // Even when recordCount = 0, we should return return OK_NEW_SCHEMA if current reader presents a new schema.
           // This could happen when data sources have a non-trivial schema with 0 row.
-          container.buildSchema(SelectionVectorMode.NONE);
+          if (firstRuntimeFiltered) {
+            container.buildSchema(SelectionVectorMode.TWO_BYTE);
+            runtimeFiltered = true;
+          } else {
+            container.buildSchema(SelectionVectorMode.NONE);
+          }
```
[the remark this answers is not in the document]
weijietong: I will take into account this suggestion.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/BloomFilter.java
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Merge this bloom filter with other one
+   * @param other
+   */
+  public void or(BloomFilter other) {
+    int otherLength = other.byteBuf.capacity();
+    int thisLength = this.byteBuf.capacity();
+    Preconditions.checkArgument(otherLength == thisLength);
+    byte[] otherTmpBucket = new byte[BYTES_PER_BUCKET];
+    for (int i = 0; i < thisLength / BYTES_PER_BUCKET; i++) {
```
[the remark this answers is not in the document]
weijietong: Yes, it's not possible to have not aligned buffers to aggregate.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
sohami: But based on my understanding `RuntimeFilterManager` only sends out something over network during broadcasting of aggregated bloomfilter. Broadcast currently happens on the Rpc thread which needs to be changed. So at what time will this Foreman thread will end up also sending the buffers over network ?

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,666 @@
...
+      DataTunnel dataTunnel = drillbitContext.getDataConnectionsPool().getTunnel(drillbitEndpoint);
+      Consumer<RpcException> exceptionConsumer = new Consumer<RpcException>() {
+        @Override
+        public void accept(final RpcException e) {
+          //logger.error("fail to broadcast a runtime filter to the probe side scan node", e);
+        }
+
+        @Override
+        public void interrupt(final InterruptedException e) {
+          //logger.error("fail to broadcast a runtime filter to the probe side scan node", e);
```
sohami: should we log this failure here and above ? Looks like if there is any failure in broadcasting aggregated filter we are treating it as minor event since query can still proceed without it.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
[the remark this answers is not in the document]
sohami: Comment was based on `findPhysicalOpContainer` return values. If null will never be returned then we should modify the method `findPhysicalOpContainer` to not do so. In cases when nothing is found to return that's an IllegalState

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,666 @@
...
+    UserBitShared.QueryId queryId = runtimeFilterB.getQueryId();
+    List<String> probeFields = runtimeFilterB.getProbeFieldsList();
+    logger.info("RuntimeFilterManager receives a runtime filter , majorId:{}, queryId:{}", majorId, QueryIdHelper.getQueryId(queryId));
+    int size;
+    synchronized (this) {
+      size = joinMjId2scanSize.get(majorId);
+      if (aggregatedRuntimeFilter == null) {
+        aggregatedRuntimeFilter = runtimeFilterWritable;
+      } else {
+        this.aggregatedRuntimeFilter.aggregate(runtimeFilterWritable);
```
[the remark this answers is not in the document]
weijietong: `aggregatedRuntimeFilter` is unique to a majorId.  But still has bugs here ,I have updated to have a majorId-aggregatedRuntimeFilter map. Thanks.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
weijietong: There is various cases that the network between the foreman and receivers break down. Also the norm case is the receiver finished earlier before receiving the bloom filter. So the security way is to use the `SendingAccountor` to release the allocated `ByteBuf`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
weijietong: As the runtime filter processing is async , the error log here is not so important. I am not sure to log it or not ,  so I comment  it here.

weijietong:
@sohami I have changed the logic to generate SV2 in the ScanBatch. It is comparatively easy now. 

Initially, I worried about the performance of the `RemovingRecordBatch`'s `setupNewSchema()` while generating different output schema of `ScanBatch` by applying `RuntimeFilter` with an assumption implementation of the `RemovingRecodBatch`'s `setupNewSchema()` to generate dynamic codes.But actually `RemovingRecodBatch`'s `setupNewSchema()` does not generate dynamic codes. So the changed implementation has reduced some complexity to judge whether to output a schema with SV2 or NONE.

I am not sure I fully understand your proposed strategy , maybe a implementation like adding `RemovingRecodBatch` on a node . But the whole design is to connect the `Scan` node and `HashJoin` one. That changing will make more work to do. So if you have no doubt about the current changed  implementation , let's it go.

Btw, others you mentioned has been addressed. Please review them, tks!

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
sohami: `SendingAccountor` on which this Foreman thread waits for completion is local to the `RunTimeFilterManager`. The job of `SendingAccountor` is to wait the caller thread iff there is any pending message to be sent unless the send is completed successfully or failed.  The buffer release is handled in the failure path of the mesaage not in sending accountor itself. The only time this `SendingAccountor` is incremented is when there is aggregated filter to be broadcasted. 
By the time this `Foreman` thread is exited there is no guarantee that remote minor fragments may have completed sending all their respective `Bloomfilters` and that `aggregated filter` is also broadcasted, hence it will be no-op in those cases. And more over if there is any failure case as you mentioned above there are chances of memory leak. To avoid that it will make more sense to put this `runtimeFilterManager.waitForComplete` inside `ForemanResult::close` method right before `closeFuture.removeListener(closeListener)`

sohami:
Overall other changes looks good but `ScanBatch` still has issues as before. I will comment in JIRA for ScanBatch Issue since github is very slow. Also for future reference please add separate commit for changes made to address review comments, that will make it easier to review just those changes. :)

exec/java-exec/src/main/java/org/apache/drill/exec/work/filter/RuntimeFilterManager.java
```diff
@@ -0,0 +1,677 @@
...
+  }
+
+  private Wrapper findPhysicalOpContainer(Wrapper wrapper, PhysicalOperator op) {
+    boolean contain = containsPhysicalOperator(wrapper, op);
+    if (contain) {
+      return wrapper;
+    }
+    List<Wrapper> dependencies = wrapper.getFragmentDependencies();
+    if (CollectionUtils.isEmpty(dependencies)) {
+      return null;
```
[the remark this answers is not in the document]
weijietong: No ,it's a right case to a join node with one side dependency having container,one side dependency not having .

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
weijietong: sorry for the wrong state. will change it to the right position .

weijietong:
ScanBatch’s discussion is on JIRA,please see that @sohami.

weijietong:
@sohami  this commit has added a `RuntimeFilterRecordBath` to replace the related work of the `ScanBatch`. Please review, thanks.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
```diff
@@ -0,0 +1,222 @@
...
+          this.field2id.put(toFilterField, typedFieldId.getFieldIds()[0]);
+          typedFieldIds.add(typedFieldId);
+          ValueVectorReadExpression toHashFieldExp = new ValueVectorReadExpression(typedFieldId);
+          hashFieldExps.add(toHashFieldExp);
+        }
+        hash64 = hashHelper.getHash64(hashFieldExps.toArray(new LogicalExpression[hashFieldExps.size()]), typedFieldIds.toArray(new TypedFieldId[typedFieldIds.size()]));
+      } catch (Exception e) {
+        throw UserException.internalError(e).build(logger);
+      }
+    }
```
sohami: above setup of `Hash64` should not happen for each `incoming` record batch, since that will be very expensive. It should happen only once during `setupNewSchema` phase until next `OK_NEW_SCHEMA`.

sohami:
@weijietong - Thanks for making the changes. But I am seeing issues in implementation related to how internals work, following protocol of iterative model of Drill and few other things. I have left few comments for now and was wondering that it will be easier to provide a commit myself to address those issues and help you get this PR pushed in faster.

Also I was working on a change for SV2 optimization [here](https://github.com/sohami/drill/commits/SV2_Optimization). Given you also have subset of the change for it with some issues, it would be great if you can rebase on top of my change.
Please let me know what you think.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   *
+   * @return True means rows are filtered by the RuntimeFilter , SelectionVector2 is set.
+   * False means not affected by the RuntimeFilter, SelectionVector2 is no set.
+   * @throws SchemaChangeException
+   */
+  private void applyRuntimeFilter() throws SchemaChangeException {
+    RuntimeFilterWritable runtimeFilterWritable = context.getRuntimeFilter();
+    if (runtimeFilterWritable == null) {
+      sv2.setRecordCount(incoming.getRecordCount());
```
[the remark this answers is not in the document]
weijietong: Since the refactored  `GenericSV2Copier` treat the `originalRecodCount == recordCount` case special. There's no need to do this addition work.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
weijietong: When the `setupNewSchema` was called, the `RuntimeFilterWritable` may have not been received . So the `hash64` should be initialized at the `doWork`. Since there's `hash64 == null` checking and initializing `hash64` logic, this part of code will only run one time.

weijietong:
@sohami Since we move the filter logic from `ScanBatch` into the `RuntimeFilterRecordBatch` , I refactor the `RuntimeFilterManager` to move the JPPD judgement logic into `RuntimeFilterVisitor`. These part of codes have fewer changes than before.  

Your SV2 optimization is more general than mine. I don't know how to operate to rebase your own branch into mine. Maybe you can guide me or I fire another JIRA to refactor the implementation to yours one after this PR and yours were merged.

weijietong:
Btw, to your optimization, I suggest the initial value `batchActualRecordCount` of `SelectionVector2` to be -1 ,not a positive number.

protocol/src/main/protobuf/UserBitShared.proto
```diff
@@ -343,6 +343,7 @@ enum CoreOperatorType {
   IMAGE_SUB_SCAN = 52;
   SEQUENCE_SUB_SCAN = 53;
   PARTITION_LIMIT = 54;
+  RUNTIME_FILTER = 55;
```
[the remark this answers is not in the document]
parthchandra: It would be OK to not generate c++ protobufs in this PR since the C++ client doesn't actually (need to) use these values. We can lump the protobuf regeneration with some other C++ client PR.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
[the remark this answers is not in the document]
weijietong: make sense ,will update.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
weijietong: make sense, but could you explain at which case ,to a query , which will have consecutive incoming `RecordBatch` but some with different schema? This really confuse me and make the solving logic complex.

weijietong:
@sohami @parthchandra  thanks for the help. I have addressed the comments according to your branch. Btw, I suggest the SV2's `batchActualRecordCount` be renamed as `originalRecordCount`.  The `recordCount` member is the truly final actual record count. Otherwise ,this naming will make others confused. If you agree with it, I will update that.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
sohami: a simple example will be Scan is reading 2 json files in 2 different record batches. Both files have a column of type map lets say m. File1 have children in map a,b of type int whereas file2 have children a,b,c. In this case scan will generate OK_NEW_SCHEMA

sohami:
@weijietong - Would have been easier to just cherry-pick the commits on top of your rebased branch. But I guess it's fine. 
`recordCount` is not the actual record count, its the selected number of records. I have also added comments to differentiate between the two variables.

Few things to resolve:
1) Since I have already shared a commit with generated protobuf files for c++ client. It would be good to take that as well.
2) Please resolve the travis build failure.
3) Meanwhile I ran tests on my shared branch yesterday and there are unit tests failures. Can you run unit tests locally and make sure all of them are passing.

weijietong:
@sohami tests have passed by removing the Kind.Integer from the OptionValue. Still keep the `IntegerValidator` to validate the number which should be a integer range value.

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/TypeValidators.java
```diff
@@ -156,9 +156,19 @@ public DoubleValidator(String name) {
...
-      super(name, Kind.INTEGER);
+      super(name);
+    }
+
+    @Override
+    public void validate(final OptionValue v, final OptionMetaData metaData, final OptionSet manager) {
+      super.validate(v, metaData, manager);
+      if (v.num_val > Integer.MAX_VALUE || v.num_val < Integer.MIN_VALUE) {
+        throw UserException.validationError()
+          .message(String.format("Option %s does not have a validate integer value", getOptionName()))
```
sohami: does not have a **valid** integer value

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/AbstractSV2Copier.java
```diff
@@ -32,6 +32,7 @@
   private SelectionVector2 sv2;
   protected List<TransferPair> transferPairs = new ArrayList<>();
 
+
```
sohami: please revert back this change.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/RemovingRecordBatch.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.drill.exec.record.VectorWrapper;
 import org.apache.drill.exec.record.WritableBatch;
 
+
```
sohami: please revert back this extra line here and below.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/svremover/StraightCopier.java
```diff
@@ -26,6 +26,7 @@
 
 import java.util.List;
 
+
```
sohami: please revert back extra line

exec/java-exec/src/main/java/org/apache/drill/exec/record/selection/SelectionVector2.java
```diff
@@ -179,6 +179,7 @@ public void close() {
     clear();
   }
 
+
```
sohami: remove this extra line

pom.xml
```diff
@@ -389,6 +389,7 @@
         <groupId>com.mycila</groupId>
         <artifactId>license-maven-plugin</artifactId>
       </plugin>
+
```
sohami: revert this change.

sohami:
Thanks for all the changes. I ran the regression tests and everything was green. 
+1 LGTM from my side.

amansinha100:
@sohami thanks for the detailed follow-up review and running the regression tests with the PR (I assume with default setting of run-time pushdown disabled).  @weijietong sorry I have not been responsive lately due to personal reasons.   Thanks for addressing review comments. 
@sohami , it's a +1 from me as well, so when you get a chance pls go ahead and merge the PR.

arina-ielchiieva:
@weijietong please rebase and resolve the conflicts (would you need to regenerate the protobufs).

arina-ielchiieva:
@weijietong thanks!

arina-ielchiieva:
@weijietong run the tests on your branch and have noticed that

 though all unit tests pass there are a huge amount of warnings during the run, checked on master such warnings are absent. Checked a couple of times and got consistent results. Attaching output of two runs for comparison. You may see the difference even in files sizes (2.53 MB vs 6.45 MB).

[consoleText_master.txt](https://github.com/apache/drill/files/2310225/consoleText_master.txt)
[consoleText_jppd.txt](https://github.com/apache/drill/files/2310229/consoleText_jppd.txt)

Please take a look and fix.

sohami:
@arina-ielchiieva  - Nice catch.
@weijietong  - Please remove [this line](https://github.com/apache/drill/pull/1334/files#diff-e58119193708a202639717568263672fR43) and line 78 from same file. The extra WARN lines are because of enabling logger in those tests cases. Since logger is not closed it's not restored back, hence for all future tests WARN level log is enabled.

weijietong:
@arina-ielchiieva @sohami  thanks for the detail catch, done.

PR #1335, opened 2018-06-25
DRILL-6529: Project Batch Sizing causes two LargeFileCompilation tests to timeout

bitblender:
Re-enabled LargeFileCompilation External Sort and increased timeout to 500s

The tests were running very close to the timeout limit without Project Batch Sizing. The batch sizing in Project causes the sort operator to take more time. Increasing the limit to 500s allows the tests to pass.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestLargeFileCompilation.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestLargeFileCompilation.java
@@ -29,7 +29,7 @@ import org.junit.rules.TestRule;
 
 @Category({SlowTest.class})
 public class TestLargeFileCompilation extends BaseTestQuery {
-  @Rule public final TestRule TIMEOUT = TestTools.getTimeoutRule(150000); // 150secs
+  @Rule public final TestRule TIMEOUT = TestTools.getTimeoutRule(200000); // 200 secs
 
   private static final String LARGE_QUERY_GROUP_BY;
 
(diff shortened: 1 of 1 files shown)


vvysotskyi:
Value for `NUM_PROJECT_COLUMNS` was increased from 2000 to 5000 in DRILL-5140 to check that generated class is splitted to several classes to avoid CompileException when class constant pool has too many items. 

In the case without the fix, these tests failed for `NUM_PROJECT_COLUMNS=5000`. But for 2500 they still passed. So I think, we should leave this value as it is and increase the timeout, or move these tests to drill-test-framework.

ilooner:
@vvysotskyi Thanks for catching this. Would decreasing the number of columns for just testEXTERNAL_SORT and testTOP_N_SORT to 2500, but keeping a column count of 5000 for all the other tests be sufficient to test that corner case?

bitblender:
@vvysotskyi @ilooner  NUM_PROJECT_COLUMNS controls 3 other tests besides testTopN and testExternalSort - testPARQUET_WRITER(), testTEXT_Writer and TestProject.  How about we set NUM_PROJECT_COLUMNS=2500 and then introduce a new constant NUM_PROJECT_TEST_COLUMNS=10000 for testProject? Basically, reduce the stress on SORTers and WRITERs but bump up the column count on testProject to a number which will push the code generated for Project over the constant pool limit. testProject takes about 130s on my Mac with 10000 columns.

arina-ielchiieva:
@bitblender removed ready-to-commit label since PR needs to be updated.

bitblender:
@ilooner @vvysotskyi I have implemented the change that I proposed. Please take a look.

PR #1337, opened 2018-06-26
DRILL-6534: Upgrade ZooKeeper patch version to 3.4.12 and add Apache Curator to dependencyManagement

KazydubB:


pom.xml
```diff
@@ -51,6 +51,7 @@
     <sqlline.version>1.1.9-drill-r7</sqlline.version>
     <jackson.version>2.9.5</jackson.version>
     <jackson.databind.version>2.9.5</jackson.databind.version>
+    <zookeeper.version>3.4.11</zookeeper.version>
```
vdiravka: Why do not use last stable version 3.4.12?
https://issues.apache.org/jira/browse/DRILL-6534?focusedCommentId=16522622&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16522622

pom.xml
```diff
@@ -1552,6 +1553,11 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>org.apache.zookeeper</groupId>
+        <artifactId>zookeeper</artifactId>
```
vdiravka: Do we need update of `curator-framework` too? Since the Curator Framework is a high-level API that simplifies using ZooKeeper.
You can try 2.12.0 version of `curator-framework`.

vdiravka:
@KazydubB  Do you know why the PR isn't added to Jira automatically? Please add the link there.

KazydubB:
@vdiravka, added the link to Jira.

pom.xml
KazydubB: Please see my comment on https://issues.apache.org/jira/browse/DRILL-6540.

KazydubB: I've tried upgrading all curator dependencies to 2.12.0, it seems to work as expected but curator-client's size is much larger than current one's, so one needs to update max size of drill-jdbc-all module by extra 1-2 MB. It is better to be handled as another task.
However I did add curator dependencies to dependencyManagement explicitly.

KazydubB:
WIP: going to exclude unneeded dependencies from dependencies added to dependencyManagement.

KazydubB:
Done.

exec/java-exec/pom.xml
```diff
@@ -309,7 +309,7 @@
     <dependency>
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-x-discovery</artifactId>
-      <version>2.7.1</version>
+      <version>${curator.version}</version>
```
vdiravka: `<version>` and `<exclusions>` aren't necessary, if you specified them in the dependencyManagement block

pom.xml
vdiravka: You have specified `Zookeeper` and `Curator` dependencies in dependencyManagement block and it's fine. It is good for controlling the versions of these libraries and their exclusions over the whole project.
But still this dependencies explicitly are used in the Drill project modules as transitive dependencies. Could you specify them explicitly in dependencies for Drill modules to avoid of using them as transitive dependencies?

pom.xml
```diff
@@ -244,7 +245,17 @@
             <dependency>
               <groupId>org.apache.curator</groupId>
               <artifactId>curator-test</artifactId>
-              <version>${curator.test.version}</version>
+              <version>${curator.version}</version>
```
vdiravka: Why do we need the exclusions in dependency for `maven-javadoc-plugin`?
https://github.com/apache/drill/blob/master/pom.xml#L211
The exclusions for this dependency in dependencyManagement block will take an affect to this too.

contrib/storage-kafka/pom.xml
```diff
@@ -88,14 +88,7 @@
     <dependency>
       <groupId>org.apache.curator</groupId>
       <artifactId>curator-test</artifactId>
-      <version>${curator.test.version}</version>
       <scope>test</scope>
```
vdiravka: It is already specified in dependencyManagement. No need to use it for every `curator-test` in every module anymore.

pom.xml
```diff
@@ -1555,6 +1566,71 @@
...
+          </exclusion>
+        </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.curator</groupId>
+        <artifactId>curator-test</artifactId>
+        <version>${curator.version}</version>
+        <scope>test</scope>
+        <exclusions>
+          <exclusion>
```
vdiravka: Do we need to exclude `netty` here? You have removed it from dependencies in other modules, so it means it isn't excluded from `curator-test` anymore, isn't it?

pom.xml
```diff
@@ -60,7 +61,7 @@
     <junit.version>4.11</junit.version>
     <hamcrest.core.version>1.3</hamcrest.core.version>
     <maven.embedder.version>3.5.3</maven.embedder.version>
-    <curator.test.version>2.7.1</curator.test.version>
+    <curator.version>2.7.1</curator.version>
```
vdiravka: Did you create a ticket with the a description for the issues of update onto curator 2.12.0 and 4.0.1 versions?
Could you post here the link for it?

exec/jdbc-all/pom.xml
```diff
@@ -636,6 +636,7 @@
                     <exclude>org.objenesis:objenesis</exclude>
                     <exclude>org.eclipse.jetty:*</exclude>
                     <exclude>org.apache.hadoop:*</exclude>
+                    <exclude>org.jboss.netty:netty</exclude>
```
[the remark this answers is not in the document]
vdiravka: Does it mean that this dependency is necessary for `zookeeper` in Drill project?

pom.xml
KazydubB: Removed exclusions block and version.

pom.xml
KazydubB: curator-test does not have it as a dependency.

exec/jdbc-all/pom.xml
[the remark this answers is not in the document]
KazydubB: Yes, for some modules this is included.

KazydubB:
@vdiravka, here is a task for upgrading curator: https://issues.apache.org/jira/browse/DRILL-6580.

PR #1341, opened 2018-06-26
DRILL-6512: Remove unnecessary processing overhead from RecordBatchSizer

ppadma:
Idea is to not computing everything by default. Do only the minimum work needed in the constructor and provide methods to get different metrics as needed. 

@ilooner please review.


bitblender:
Thanks @ppadma ! These change eliminate quite a bit of useless computation. Is there any need/plan to allow concurrent access to the RecordBatchSizer in the future? With these changes, access to the RecordBatchSizer will have more synchronization points for concurrent access. Earlier, ouputRowCount was the only field that was mutated after creating a RecordBatch object. Now many more fields are being changed.  I just wanted to raise that one point. Otherwise, LGTM +1

ppadma:
@bitblender thanks Karthik. RecordBatchSizer is being created by each operator when they get a new batch. I don't see any need for concurrent access, atleast in the near future.

PR #1346, opened 2018-06-28
DRILL-6546: Allow unnest function with nested columns and complex expressions

vvysotskyi:
- Added new rule `ProjectComplexRexNodeCorrelateTransposeRule` which takes a complex expression from the `Project` below `Uncollect` rel node and creates new project with expressions from the left side of `Correlate` and this complex expression. 
For example, part of the plan before applying the rule:
```
LogicalCorrelate(correlation=[$cor0], joinType=[inner], requiredColumns=[{1}]): rowcount = 1.0, cumulative cost = {inf}, id = 100
  EnumerableTableScan(subset=[rel#94:Subset#0.ENUMERABLE.ANY([]).[]], table=[[cp, lateraljoin/nested-customer.parquet]]): rowcount = 100.0, cumulative cost = {100.0 rows, 101.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 7
  Uncollect(subset=[rel#99:Subset#3.NONE.ANY([]).[]]): rowcount = 1.0, cumulative cost = {1.0 rows, 1.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 98
    LogicalProject(subset=[rel#97:Subset#2.NONE.ANY([]).[]], EXPR$0=[ITEM($cor0.orders, 'items')]): rowcount = 1.0, cumulative cost = {1.0 rows, 1.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 96
      LogicalValues(subset=[rel#95:Subset#1.NONE.ANY([]).[0]], tuples=[[{ 0 }]]): rowcount = 1.0, cumulative cost = {1.0 rows, 1.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 8
```
Plan after applying the rule:
```
LogicalProject(**=[$0], orders=[$1], $complexRexNode0=[$3]): rowcount = 1.0, cumulative cost = {inf}, id = 116
  LogicalCorrelate(correlation=[$cor1], joinType=[inner], requiredColumns=[{2}]): rowcount = 1.0, cumulative cost = {inf}, id = 115
    LogicalProject(**=[$0], orders=[$1], $complexRexNode=[ITEM($1, 'items')]): rowcount = 100.0, cumulative cost = {100.0 rows, 300.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 112
      EnumerableTableScan(subset=[rel#94:Subset#0.ENUMERABLE.ANY([]).[]], table=[[cp, lateraljoin/nested-customer.parquet]]): rowcount = 100.0, cumulative cost = {100.0 rows, 101.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 7
    Uncollect: rowcount = 1.0, cumulative cost = {2.0 rows, 2.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 114
      LogicalProject($complexRexNode=[$cor1.$complexRexNode]): rowcount = 1.0, cumulative cost = {1.0 rows, 1.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 113
        LogicalValues(subset=[rel#95:Subset#1.NONE.ANY([]).[0]], tuples=[[{ 0 }]]): rowcount = 1.0, cumulative cost = {1.0 rows, 1.0 cpu, 0.0 io, 0.0 network, 0.0 memory}, id = 8
```
- Made change to convert `DrillCompoundIdentifier` inside unnest to the item call to avoid column not found error when nested column is used.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
@@ -21,6 +21,7 @@ import org.apache.drill.categories.OperatorTest;
 import org.apache.drill.exec.planner.physical.PlannerSettings;
 import org.apache.drill.test.ClusterFixture;
 import org.apache.drill.test.ClusterTest;
+import org.apache.drill.test.TestBuilder;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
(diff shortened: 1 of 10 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
```diff
@@ -114,6 +115,57 @@ public void testMultiUnnestAtSameLevel() throws Exception {
...
+        .sqlQuery(sql)
+        .unOrdered()
+        .baselineColumns("o_amount")
+        .baselineValues(4.5)
+        .go();
+  }
+
+  @Test
+  public void testUnnestWithMap() throws Exception {
+    String sql = "select u.item from\n" +
```
HanumathRao: Can you also add multi unnest query to the testing. (something like this)
select u.item from cp.`lateraljoin/nested-customer.parquet` c, unnest(c.orders.items) as u(item), unnest(c.orders.items) as u1(item1) limit 1

HanumathRao:
@vvysotskyi Thanks for making the code changes. Code changes look good to me. +1.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
vvysotskyi: Thanks, added `testMultiUnnestWithMap()` test.

vvysotskyi:
After the rebase on current master, unit tests are failed. Fixing now.

vvysotskyi:
Unit tests failures after rebase were caused by another bug. It appears when a query has a single expression in project list and this expression is taken from the unnest. In this case, name of this field in unnest rel node is lost due to the change made in DRILL-6545.

I have fixed this bug and added one more unit test for this case.

vvysotskyi:
@amansinha100, could you please do the final review?

amansinha100:
@vvysotskyi one question about the `$complexRexNode` ..suppose there are 2 or more such nested columns, e.g `a.b.c`  and `a.b.d`  that need to be projected from the left side of Lateral, then I would think that there would be 2 such fields: `$complexRexNode1`  and `$complexRexNode2`.   Is that correct ?

vvysotskyi:
AFAIK, currently allowed only unnest with single column. When several unnests are used, they will have their own Correlate rel nodes. Pp

vvysotskyi:
AFAIK, currently allowed only unnest with single column. When several unnests are used, they will have their own Correlate rel nodes. Pp

amansinha100:
Yes, each unnest will operate on single column, so the plan will have 2 LateralJoin and as long as there is a Project inserted on the left side of each one, then I suppose the code will work fine.   

Overall, I am +1.

PR #1347, opened 2018-06-28
DRILL-6545: Projection Push down into Lateral Join operator.

HanumathRao:
@amansinha100  Please review this PR.

This PR includes.
1) Two new Rules for Push Project into lateral join and push project past lateral join( this rule is only to avoid triggering for un-necessary cases).
2) Change in LateralPop so as to pass the excluded columns to runtime operator.
3) Constructing row type for LateralJoin based after removing the correlated column.


HanumathRao:
@amansinha100  Thank you for the review.

PR #1348, opened 2018-06-28
DRILL-6346: Create an Official Drill Docker Container

Agirish:
Added apache-release profile in distribution/pom.xml to build a Docker image once the project is built successfully. While the Docker image can be built anytime, I assumed that an official Docker image would only be generated during the release. 

This requires having Docker configured on the environment (works with Docker for Mac). 

Usage: 
Build the project by specifying the apache-release profile
```
$ mvn clean install -Papache-release -DskipTests
```
Pushing image to Docker Hub (or any Container Registry):
```
cd distribution
mvn dockerfile:build -Papache-release
mvn dockerfile:push -Papache-release
```
Run the Docker container -- which should connect and start Drill in embedded mode
```
$ docker run -i -t drill/apache-drill-centos:1.14.0-SNAPSHOT /bin/bash
Jun 29, 2018 3:28:21 AM org.glassfish.jersey.server.ApplicationHandler initialize
apache drill 1.14.0-SNAPSHOT 
"json ain't no thang"
0: jdbc:drill:zk=local> select version from sys.version;
+------------------+
|     version      |
+------------------+
| 1.14.0-SNAPSHOT  |
+------------------+
1 row selected (0.28 seconds)
```

### DIFF
--- a/distribution/Dockerfile
+++ b/distribution/Dockerfile
@@ -1,0 +1,35 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+FROM centos:7
+
+# Project version defined in pom.xml is passed as an argument
+ARG VERSION
+
+# JDK 8 is a pre-requisite to run Drill ; 'which' package is needed for drill-config.sh
+RUN yum install -y java-1.8.0-openjdk-devel which ; yum clean all ; rm -rf /var/cache/yum
+
+# The drill tarball is generated upon building the Drill project
+COPY target/apache-drill-$VERSION.tar.gz /tmp
+
+# Drill binaries are extracted into the '/opt/drill' directory
+RUN mkdir /opt/drill
+RUN tar -xvzf /tmp/apache-drill-$VERSION.tar.gz --directory=/opt/drill --strip-components 1
+
+# Starts Drill in embedded mode and connects to Sqlline
+ENTRYPOINT /opt/drill/bin/drill-embedded
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -485,6 +485,34 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>docker</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.spotify</groupId>
+            <artifactId>dockerfile-maven-plugin</artifactId>
+            <version>1.4.3</version>
+            <executions>
+              <execution>
+                <id>docker-image</id>
+                <goals>
+                  <goal>build</goal>
+                  <goal>push</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <repository>${docker.repository}</repository>
+              <tag>${project.version}</tag>
+              <buildArgs>
+                <VERSION>${project.version}</VERSION>
+              </buildArgs>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
 
 </project>
--- a/docs/dev/Docker.md
+++ b/docs/dev/Docker.md
@@ -1,0 +1,97 @@
+# How to build, publish and run a Apache Drill Docker image
+
+## Prerequisites
+
+   To build an Apache Drill docker image, you need to have the following software installed on your system to successfully complete a build. 
+  * [Java 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
+  * [Maven 3.3.1 or greater](https://maven.apache.org/download.cgi)
+  * [Docker CE](https://store.docker.com/search?type=edition&offering=community)
+  
+   If you are using an older Mac or PC, additionally configure [docker-machine](https://docs.docker.com/machine/overview/#what-is-docker-machine) on your system
+
+## Checkout
+```
+git clone https://github.com/apache/drill.git
+```
+## Build Drill
+```
+$ cd drill
+$ mvn clean install
+```   
+## Build Docker Image
+```
+$ cd distribution
+$ mvn dockerfile:build -Pdocker
+```
+## Push Docker Image
+   
+   By default, the docker image built above is configured to be pushed to [Drill Docker Hub](https://hub.docker.com/r/drill/apache-drill/) to create official Drill Docker images.
+```   
+$ cd distribution
+$ mvn dockerfile:push -Pdocker
+```    
+  You can configure the repository in pom.xml to point to any private or public container registry, or specify it in your mvn command.
+```  
+$ cd distribution
+$ mvn dockerfile:push -Pdocker -Pdocker.repository=<my_repo>
+```
+## Run Docker Container
+   
+   Running the Docker container should start Drill in embedded mode and connect to Sqlline. 
+```    
+$ docker run -i --name drill-1.14.0 -p 8047:8047 -t drill/apache-drill:1.14.0 /bin/bash
+Jun 29, 2018 3:28:21 AM org.glassfish.jersey.server.ApplicationHandler initialize
+INFO: Initiating Jersey application, version Jersey: 2.8 2014-04-29 01:25:26...
+apache drill 1.14.0 
+"json ain't no thang"
+0: jdbc:drill:zk=local> select version from sys.version;
++------------+
+|  version   |
++------------+
+| 1.14.0     |
++------------+
+1 row selected (0.28 seconds)
+```  
+
+   You can also run the container in detached mode and connect to sqlline using drill-localhost. 
+```    
+$ docker run -i --name drill-1.14.0 -p 8047:8047 --detach -t drill/apache-drill:1.14.0 /bin/bash
+<displays container ID>
+
+$ docker exec -it drill-1.14.0 bash
+<connects to container>
+
+$ /opt/drill/bin/drill-localhost
+apache drill 1.14.0 
+"json ain't no thang"
+0: jdbc:drill:drillbit=localhost> select version from sys.version;
++------------+
+|  version   |
++------------+
+| 1.14.0     |
++------------+
+1 row selected (0.28 seconds)
+```
+
+## Querying Data
+
+   By default, you can only query files which are accessible within the container. For example, the sample data which ships with Drill. 
+```
+> select first_name, last_name from cp.`employee.json` limit 1;
++-------------+------------+
+| first_name  | last_name  |
++-------------+------------+
+| Sheri       | Nowmer     |
++-------------+------------+
+1 row selected (0.256 seconds)
+```
+
+   To query files outside of the container, you can configure [docker volumes](https://docs.docker.com/storage/volumes/#start-a-service-with-volumes)
+   
+## Drill Web UI
+
+   Drill web UI can be accessed using http://localhost:8047 once the Drill docker container is up and running. On Windows, you may need to specify the IP address of your system instead of 'localhost'.
+
+## More information 
+
+   For more information including how to run Apache Drill in a Docker container, visit the [Apache Drill Documentation](http://drill.apache.org/docs/)
--- a/pom.xml
+++ b/pom.xml
@@ -87,6 +87,7 @@
     <additionalparam>-Xdoclint:none</additionalparam>
     <rat.skip>true</rat.skip>
     <license.skip>true</license.skip>
+    <docker.repository>drill/apache-drill</docker.repository>
   </properties>
 
   <scm>


distribution/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+FROM centos:7
+ARG VERSION
+
+RUN yum install -y java-1.8.0-openjdk-devel which ; yum clean all ; rm -rf /var/cache/yum
+
+COPY target/apache-drill-$VERSION.tar.gz /tmp
+RUN mkdir /opt/drill
+RUN tar -xvzf /tmp/apache-drill-$VERSION.tar.gz --directory=/opt/drill --strip-components 1
+
+ENTRYPOINT /opt/drill/bin/drill-embedded
```
arina-ielchiieva: Will user be able to use sqlline if drill if already started in embedded mode? Also how to stop drill in this case?

distribution/pom.xml
```diff
@@ -485,6 +485,36 @@
...
+              <execution>
+                <id>docker-image</id>
+                <goals>
+                  <goal>build</goal>
+                  <goal>push</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <repository>drill/apache-drill-centos</repository>
```
arina-ielchiieva: 1. We might want to make this configurable? 
2. Why docker image build will be available only in apaceh-release profile? What if user want to create image and push it to its own repo?
3. `drill/apache-drill-centos` does committers have access to this repo to do the push?
4. https://github.com/parthchandra/drill/wiki/Drill-Release-Process should be updated with instructions about Docker.

distribution/pom.xml
```diff
@@ -485,6 +485,36 @@
...
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>com.spotify</groupId>
+            <artifactId>dockerfile-maven-plugin</artifactId>
+            <version>1.4.3</version>
+            <executions>
+              <execution>
+                <id>docker-image</id>
+                <goals>
```
arina-ielchiieva: Will it push without phase being specified? Example: `<phase>package</phase>`

distribution/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+FROM centos:7
+ARG VERSION
+
+RUN yum install -y java-1.8.0-openjdk-devel which ; yum clean all ; rm -rf /var/cache/yum
```
arina-ielchiieva: Should `which` be removed?

distribution/pom.xml
Agirish: 1. This is configurable - anyone can update it to any private or public Docker hub or any Container Registry. Do you have a different approach in mind?
2. I briefly touched upon this in the description. Since the ask was for an official Docker image for AD, it makes sense to build this once during the release. Anyone can download the Docker image at anytime, using a simple pull. Everything else remains the same. I also chose this approach to prevent Docker images from being generated everytime. 

3. Tim and I have. We intend to give access to the 'drill' Docker Hub Organization to all committers. Perhaps, starting with release managers for every release. It's a good central place to have all related docker images for Drill.

4. Sure.

distribution/pom.xml
Agirish: We can get rid of 'push' if you'd like. 'build' is good enough for most purposes. I'm not sure about 'package'.

distribution/Dockerfile
Agirish: Yes, that's right. I can add comments.

Agirish: Done. Please take a look and suggest if any of the comments can be rephrased.

distribution/pom.xml
ilooner: I suggest not keeping a separate repository based on os. Typically there is one repository and the os is included in the tag for the image.

```
drill/apache-drill:1.14-centos
drill/apache-drill:1.14-alpine
```

ilooner: We could move `drill/apache-drill-centos` into a maven property so that the user could override it from the command line.

Agirish: Sure. How about even simpler tag: 
drill/apache-drill:1.14.0
The default tag will be for CentOS. We can have additional tags like you mentioned:
drill/apache-drill:1.14.0-centos (alias for 1.14.0)
drill/apache-drill:1.14.0-ubuntu
drill/apache-drill:1.14.0-alpine 
etc..

Also sure, the Docker Hub repo can be moved into a maven property. Should we keep the drill docker hub repo as the default, or just make it a local repository by default instead.

distribution/pom.xml
Agirish: Here is how we can explicitly build, tag and push (any one action at at time):

Inside the distribution directory, executing:
`mvn -Papache-release dockerfile:build`
`mvn -Papache-release dockerfile:tag`
`mvn -Papache-release dockerfile:push`

Build and Tag happens on the user's system. Push pushes the image to the container registry on Docker Hub: https://hub.docker.com/r/drill/apache-drill-centos/tags/

On the root directory, one can simply run the below command to build and tag the container. Can be optionally pushed to Docker Hub separately. 
 `mvn clean install -Papache-release -DskipTests`

Agirish:
I'll work on adding the instructions into both the docs dir on this repo and the write-up on Parth's repo once the PR is in.

Agirish:
@arina-ielchiieva , on second thoughts, i think having a seperate profile is better. Created a new profile called 'docker'. 
Steps remain mostly similar:
Build Project (any profile can be provided. If it's 'docker' profile, then docker image is built)
`$ mvn clean install`
Build Docker image Push image to Docker Hub (or any Container Registry):
```
cd distribution
mvn dockerfile:build -Pdocker
mvn dockerfile:push -Pdocker
```

distribution/Dockerfile
arina-ielchiieva: When user uses run command (without -d / --detach), he will get Sqlline. But if he restarts the container (stop + start) or runs container with -d / --detach, he won't get Sqlline. In this case he needs to use `drill-localhost` command to access Sqlline. I think we need to document this.

distribution/pom.xml
arina-ielchiieva: `Tim and I have. We intend to give access to the 'drill' Docker Hub Organization to all committers. Perhaps, starting with release managers for every release. It's a good central place to have all related docker images for Drill.`
I think you might consider giving access to several PMCs as well :)

arina-ielchiieva:
@Agirish I think having separate docker profile is good idea.

distribution/pom.xml
Agirish: Sure! We'll need everyone to create an account on Docker Hub: https://hub.docker.com/u/drill/ is the repository. Once I have the usernames, I can add them to the 'drill' organization - that should give write access. How do you suggest I proceed with this? For now, I was planning on asking Boaz (current release manager) and anyone interested their usernames.

distribution/Dockerfile
Agirish: But drill-localhost will not work right? That requires support for standalone mode - which requires ZK. Or am I missing something? 

Yes, with the current approach, the container is only alive until the embedded sqlline session is active. If they exit and reattach, they will get a new session (all history is lost). 

I could add an init script with an infinite wait, and suggest going the docker exec route, to keep the container alive and maintain history - but since it would still only support embedded mode (until we add support for standalone using ZK), i went with this approach. What do you suggest?

One possibility, is we can document instructions on how to make changes to get what I suggested above working.

arina-ielchiieva: Actually, it worked. I have started drill-embedded in one console. And in other connected to drill using drill-localhost.

Agirish: It worked for me too! I think I understood why - drill-localhost connects to a locally running Drillbit (doesn't matter embedded or distributed). Since the run command which stars Drill in embedded mode keeps the Drillbit running, we can use drill-localhost to launch another session. Cool, i'll document this.

priteshm:
Since @arina-ielchiieva is the batch committer, I marked the JIRA as ready-to-commit, so she can review it as well.

arina-ielchiieva:
+1 but @Agirish please follow up on the following items:
1. document `drill-localhost` usage;
2. document docker build in docs/dev/DevDocs.md;
3. apply to make drill docker repository official (https://docs.docker.com/docker-hub/official_repos/#how-do-i-create-a-new-official-repository).

Agirish:
Thanks all, for the review!

I'll work with @bbevens to document usage of Docker. I'll follow-up on the Docker official images in a few days. Regarding dev docs in this repo, I have a draft here: https://github.com/Agirish/drill/blob/docker_doc/docs/dev/Docker.md. If it looks good, I can add it to the same PR. If not, I can create a new PR on a later date.

Agirish:
@cgivre , sure. Rough ETA on the drill.apache.org/docs documentation is mid July.

ppadma:
@Agirish This is very useful Abhishek. Thanks for doing this.

arina-ielchiieva:
@Agirish doc looks good and sure please include it in this PR.
One minor suggestion in `Push Docker Image` section, please add example how to define custom repo. 
I suppose it should be something like this: `mvn dockerfile:push -Pdocker -Pdocker.repository=<my_repo>`.

distribution/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# The drill tarball is generated upon building the Drill project
+COPY target/apache-drill-$VERSION.tar.gz /tmp
+
+# Drill binaries are extracted into the '/opt/drill' directory
+RUN mkdir /opt/drill
+RUN tar -xvzf /tmp/apache-drill-$VERSION.tar.gz --directory=/opt/drill --strip-components 1
+
+# Starts Drill in embedded mode and connects to Sqlline
+ENTRYPOINT /opt/drill/bin/drill-embedded
```
paul-rogers: This `Dockerfile` is only for an embedded Drillbit, which is find for playing around.

Should we also offer an example production `Dockerfile`? In such a file, the following would also be needed:

* Copy in custom jars (UDFs custom storage plugins.)
* Copy in custom libraries (PAM, etc.)
* Specify custom config.

Doing the above will be easier if the user's files are not placed in the `$DRILL_HOME/conf` and other directories. That is, we want to use the site directory  feature.

Maybe allow a `/opt/drill-site` location and pass `--site /opt/drill-site` on the command line for the entry point.

Then, the `Dockerfile` can provide examples of how to copy the various kinds of files to the proper site directory locations:

* `/opt/drill-site/` - `drill-override.conf`, `core-site.xml`, `logback.xml`
* `/opt/drill-site/jar/` - UDF and storage-plugin jars
* `/opt/drill-site/lib/` - native libraries

Note that the functionality for Drill to find things in these site directory locations already exists. All we're doing here is showing the user how to use them.

distribution/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# JDK 8 is a pre-requisite to run Drill ; 'which' package is needed for drill-config.sh
+RUN yum install -y java-1.8.0-openjdk-devel which ; yum clean all ; rm -rf /var/cache/yum
+
+# The drill tarball is generated upon building the Drill project
+COPY target/apache-drill-$VERSION.tar.gz /tmp
+
+# Drill binaries are extracted into the '/opt/drill' directory
+RUN mkdir /opt/drill
+RUN tar -xvzf /tmp/apache-drill-$VERSION.tar.gz --directory=/opt/drill --strip-components 1
```
paul-rogers: A handy trick I learned is to do the container setup in a script, say `setup.sh` that is copied into the container and run. This avoids creating a bunch of extra Docker layers as the set of steps gets larger.

distribution/Dockerfile
paul-rogers: We've all learned that Drill logs are essential to figure out what went wrong. As configured, these will end up in `/opt/drill/log` in the container's writeable layer. The data is lost when the container exits.

Might want to configure the container to write logs to, say, `/var/log/drill`, then encourage the user to do a bind or volume mount to that location in order to persist logs after the container exits.

Alternatively, for use in a system such as K8s, configure the logs to write to stdout so that the K8s log system can capture the log output, display it to the user (`kubectl logs <pod>`) or route it to the log aggregation system.

paul-rogers: Wondering how best to configure a true Drill server? Would the user really want to bind the ZK address, say, at container build time (as we suggested above, copying in the user's `drill-override.conf` file.)

Or, should we think about how to pass in config? In K8s, config can be passed in via config maps and mounted into the proper location.

Doing so, however, points out a flaw in the site directory structure: config files go into `$DRILL_SITE` but jars go into `$DRILL_SITE/jars`. If `$DRILL_SITE` is mounted from a config map, we hide the jars.

Maybe need to modify the site directory so that the scripts first look in `$DRILL_SITE/conf`, then in `$DRILL_SITE`. In this case, the config map would be mounted into `$DRILL_SITE/conf`.

paul-rogers: How will secrets be handled? MapR tickets or Kerberos certificates?

What other configuration is needed?

paul-rogers: The `Dockerfile` does not expose ports:

```
EXPOSE 8047/tcp
```

The above is for the web console port. Should we expose others?

paul-rogers: Should we provide a `README.md` file? Explain that this particular `Dockerfile`:

* Runs an embedded Drillbit
* Works only for data already on the class path
* That the user must ssh(?) into the container?

Explain how to do additional configuration (as discussed in other comments).

Explain how to create a production `Dockerfile`.

Else, the user must be a bit of a Docker and Drill expert to work out what would be required, an it won't be clear what goal this particular file is trying to achieve.

paul-rogers: The entry point launches an embedded Drillbit. That in turn fires up Sqlline which has an interactive console.

It would seem that the only way to use this image is to launch it in an interactive session:
```
docker run -it <image-name>
```

This seems like a handy trick for trying Drill. But, very limited for production. In any event, the `README.md` file should probably explain how to use the image.

arina-ielchiieva:
Another point we should consider is that there are two ways that author can publish images on Docker Hub: build locally and push or make Dockerfile publicly available and use Docker Hub's continuous build system. The second way is more preferred and marked as trusted.

paul-rogers:
If running a server, we may want to let the Drillbit shutdown gracefully. To do this, intercept the `SIGTERM` coming into the container and forward it to the process. If the `start.sh` script `exec`s the Drillbit, then the Drillbit may handle the shutdown. Otherwise, you can have the `start.sh` [handle the SIGTERM](https://medium.com/@gchudnov/trapping-signals-in-docker-containers-7a57fdda7d86).

The normal `drillbit.sh` script appears to do an `exec`, so the Drill server should end up as pid 1 (if your `start.sh` script does an `exec` to `drillbit.sh`.) But, check this in case anything has changed since we looked into this stuff for Drill-on-YARN.

distribution/Dockerfile
ilooner: I'm actually in favor doing setup within the Dockerfile, not in a `setup.sh`. The layers are actually advantageous because if you change a step at the end of the Dockerfile or add a step at the end, only the diff is rebuilt. All the layers before that new step can be reused from the docker's cache. Furthermore it speeds up deploying updated docker containers since only the diff has to be downloaded. If you use a master `setup.sh` script there will be essentially only one layer, so even small changes to the setup script will require a complete rebuild of your container, and then a complete redownload by clients.

The trick is that you want the big infrequently changing setup tasks to be toward the beginning of your docker file to take full advantage of layer caching. And you want the small lightweight setup tasks to be towards the end of the Dockerfile, so that if you change them hardly anything needs to be rebuilt, uploaded, and downloaded.

distribution/Dockerfile
ilooner: I've typically seen docker images use environmental variables to pass in parameters. If zookeeper is deployed on Kubernetes, wouldn't the nodes have static hostnames and ports like node.zk1, node.zk2? If so we could have a ZK_HOSTS variable which overrides drill config, and users can set the appropriate values for ZK_HOSTS in their drill pod definition yaml.

Behind the scenes a script could update the drill config file appropriately depending on whether ZK_HOSTS is defined or not before launching the drillbit.

Agirish: Hey Paul, thanks for the comments. This PR is to support Drill on Docker in embedded mode, to help people quickly get started with Drill. If they have Docker on their Mac or PC, they can just issue the Docker run command to pull the latest image of Drill and run in embedded mode. 

But, I agree that supporting a standalone Drill Dockerfile is necessary. I already have it working, but there is some more effort required - which i'll track in a different PR (tracked by a new JIRA). That would require documentation on how to pull in a ZK image (we already have it part of the Drill Docker Hub repo) and launch Drill in distributed mode. 

Regarding the drill site directory, that's a good suggestion - i'll try how to get that working seamlessly with Docker. Will reach out to you for your comments then.

distribution/Dockerfile
Agirish: @paul-rogers My initial implementation did have a init script - i removed it for this PR - to make it as simple as possible. I'll need to include it later for distributed mode support, if I cannot figure out a different approach for a keep alive probe. 

@ilooner , agree with regard to layers - i find it very useful. I don't intend to move the tar download to setup.sh - just the config and start-up parts (eventually, in a new PR).

distribution/Dockerfile
Agirish: Agree - this PR introduces embedded support only. But when we support stand-alone mode, either under Docker or K8S, we'll work on preserving configurations and contents. 

I have Drill working under K8S already - and the profiles have been moved to DFS (which is preserved). For logs, we'll need to figure out K8S volume plugin mounts may be. 

For a vanilla Docker env, i'll need to lookup if that's possible.

Agirish: I will introduce support for K8S in a later PR. In my implementation, I bring up ZK as a service and use the auto-resolving service name in Drill configuration (I have had to edit the drill-override.conf with the service name when creating the Docker container). 

@ilooner , yes Config maps is K8S specific - where we can pass in variables defined elsewhere.

Agirish: For embedded mode, i've not needed to expose the ports. For K8S, i've needed to expose the HTTP port (8047) and the server port (31010) to enable the loadbalancer.

Agirish: I've not worked on enabling security as yet. In the K8S implementation for MapR, we're planning on use config maps for this.

Agirish: I do have a readme file - https://github.com/Agirish/drill/blob/docker_doc/docs/dev/Docker.md

I'll be adding that to this PR very shortly (once I clean-up).

Agirish: I also have some basic info in the description

Agirish: Yes, we support both in this PR. I've added some details in the readme: https://github.com/Agirish/drill/blob/docker_doc/docs/dev/Docker.md (will add it to this PR shortly).

Agirish:
@paul-rogers , regarding graceful shutdown, i think we can support it under K8S. Currently under K8S, the Drill container isn't even aware that it's being kicked out.

Agirish:
@arina-ielchiieva , i'll work on implementing the Docker continuous integration support (it requires configurations on the Docker Hub website). It's pretty helpful - users will be able to get to try out the latest master on Docker. For now, it's a manual process.

ilooner:
@Agirish @paul-rogers @arina-ielchiieva Is this going to be merged? Or do we want to make the production enhancements Paul identified first?

I suggest we merge this now, since having a docker image for people to play with locally is a step forward, and create JIRAs for the production enhancements Paul pointed out and do them separately. Thoughts?

Agirish:
I agree we should get this merged. You all suggested to include docs, so i'll quickly pull in this file: https://github.com/Agirish/drill/blob/docker_doc/docs/dev/Docker.md into this PR - that's the only pending change. 

Standalone mode support on Docker, and the whole of K8S support can tracked by separate JIRAs (i'll file them shortly).

Agirish:
Kindly take a look at the Docker.md file and let me know if any changes are required.

distribution/Dockerfile
paul-rogers: For vanilla Docker, I suppose the user could use a Docker volume mount onto `/opt/drill/log` to capture logs... The problem is, the more directories that must be mounted this way, the messier it gets. 

As noted above, it is fine to address this in another PR.

paul-rogers: Env vars are fine: both Docker and K8s support them. The trick here is that the set of such configs will be fixed and the code here will need to edit the values into the Drill config files, which makes it hard to add customer-specific config.

Drill configs, in their full glory, are quite robust (with all the `drill-env.sh` and `drill-override.conf `options considered.) Far too much stuff to expose as env vars.

Maybe have the "stock" Dockerfile handle the most obvious config settings (memory, ZK address, ...) and edit them into the files. (Luckily memory is already designed for this, so no work needed.) Then, advanced users can create their own Dockerfiles for the advanced cases, using the "stock" one as an example to get started.

K8s config maps are just an easy way to pass whole files. Probably the same can be done in Docker by putting the files on a local disk, then mounting that directory into Docker. Hence, the idea of separating the site directory from the Drill distribution; the user can just mount their own site directory. (And, it can be from a config map in K8s.)

paul-rogers: I keep forgetting, MapR has its own way to do security... For those of us in HDFS-land, we need to pass in Kerberos tickets or S3 certificates.

I wonder, can we put all these items into a "certs" directory that can be mounted either as a K8s secret or as a host volume in plain Docker? Maybe the folks who've worked on security have a suggestion.

paul-rogers: Even in embedded mode, I'd definitely want to see the query profile, so no harm in exposing the web port.

Yes, for production, we need to expose the three Drill ports. (As an aside, this means the ports can't be customized. Which is fine: if someone wants to run multiple Drillbits per node, they need to use an overlay network such as Weave, Calico, etc.)

paul-rogers: The PR description won't follow the code. Look forward to the readme file. I can point to it from the Docker section in the Drill book. (And, you'll have done the writing for me... Actually, by pointing to the readme, the reader can get the latest info, since this is a young feature and will change.)

distribution/pom.xml
paul-rogers: Sounds like our Maven build is getting complex enough that it could benefit from a readme as well. (I've accumulated notes for how to generate code, how to rebuild protobufs, how to run single tests, etc. This information should be in a shared form...)

docs/dev/Docker.md
```diff
@@ -0,0 +1,104 @@
...
+      API version:  1.37 (minimum version 1.12)
+      Go version:   go1.9.5
+      Git commit:   9ee9f40
+      Built:        Thu Apr 26 07:22:38 2018
+      OS/Arch:      linux/amd64
+      Experimental: true
+
+## Checkout
+
+    git clone https://github.com/apache/drill.git
```
paul-rogers: For commands and preformatted tables/lists, might want to use the triple-back-tick format so that the stuff looks like a command and is not HTML-formatted.

Agirish: @paul-rogers, done. Please take a look.

distribution/Dockerfile
Agirish: @paul-rogers, I meant to say that I have a separate ZK image which can be used to bring up ZK standalone, and use it for Drill. For example, see https://github.com/Agirish/drill-containers/blob/master/dockerfiles/apache-zookeeper-centos-3.4.10/Dockerfile

And yes, I agree having a consolidated image would be good. It will require us to bring up ZK (using the independent ZK image) and then bring up Drill using the ZK information. With K8S, it can be made seamless, but with Docker, it'll require some manual configurations. Hence pushed it out to a different PR. 

Regarding referencing it for the Drill book, may be we could connect over email. I have existing docs and instructions which i've used to successfully bring up Drill in distributed modes on Docker and K8S env. It could be faster than getting it into Apache :)

Agirish: Sure, i'll keep this in mind.

Agirish: Yeah, will need some effort on security.

Agirish: I meant to say that i've not needed to expose the ports to have the web UI or any functionality working. I've tried on Mac and Linux. I've only needed to expose the ports in K8S env when using an externally accessible IP via LoadBalancer. 

And yes, I use weave for K8S. It's helpful for multi-node Docker env on K8S.

Agirish: I've added the readme. I'm also working with @bbevens on a user facing doc. I'm not sure if there is anything else you would like me to add. And yes, surely we can update the doc as we add / modify features.

distribution/pom.xml
Agirish: Yeah, I agree - but there is no clean way to do it. We could add some notes in the dev docs later perhaps? Or do you have a different suggestion?

docs/dev/Docker.md
```diff
@@ -0,0 +1,73 @@
...
+  You can configure the repository in pom.xml to point to any private or public container registry, or specify it in your mvn command:
+```  
+$ cd distribution
+$ mvn dockerfile:push -Pdocker -Pdocker.repository=<my_repo>
+```
+## Run Docker Container
+   
+   Running the Docker container should start Drill in embedded mode and connect to Sqlline
+```    
+$ docker run -i -t drill/apache-drill:1.14.0 /bin/bash
```
arina-ielchiieva: Please add the following clarifications:
1. Detach: When starting this way (without detach), once you exit the sqline, container will be also killed. To persist the container use `--detach`: `docker run -i --detach -t drill/apache-drill:1.14.0 /bin/bash /bin/bash`.
2. Port forwarding: In order to access Drill WebUI you need to open 8047 port: `docker run -i -p 8047:8047 -t drill/apache-drill:1.14.0 /bin/bash`.
3. Name: Suggest optionally to add name to the container (`--name drill`): `docker run --name drill -i -p 8047:8047 -t drill/apache-drill:1.14.0 /bin/bash`.
4. Docker machine: On Windows and Mac default docker machine might not have enough resources.
On my env I had to create new docker machine. Example:
```
# stop default
docker-machine stop default
# add drill-machine
docker-machine create -d virtualbox --virtualbox-cpu-count "2" --virtualbox-memory "16384" drill-machine
# activate drill-machine
eval $(docker-machine env drill-machine)
# check status
docker-machine active
```
5. Querying custom files: To query custom files (which I believe will be the major use case), user needs to mount a volume (https://docs.docker.com/storage/volumes/) and then update file storage plugin to look at that volume. I did not try that but it's worth mentioning and even adding a simple manual.

arina-ielchiieva:
To summarize in this Jira we are aiming to add docker support only for Drill in embedded mode.
@Agirish 
1. Please create follow up Jira with future docker enhancements (support for distributed mode, making docker official and ability to build on docker env etc).
2. Please add clarifications I have mentioned in PR comment about detach, port forwarding, naming, docker machine and querying custom files.

Once all that is done, I believe we can approve the PR and merge it.

docs/dev/Docker.md
Agirish: (1) --detach option is useful. Users will then need to use drill-localhost to connect to Sqlline once they connect into the container. Will update the doc.
(2) I don't think it's necessary. Also, it's preferred to use localhost:8047 to connect to Web UI if brought up on Mac or PC.  
(3) Name will be useful - this way we'll not need to provide the container ID during Docker exec. Will add this and update the doc. 
(4) I tried on Docker for Mac - there were no issues. Looking at the doc (https://docs.docker.com/machine/overview/#what-is-docker-machine), it's for older Mac systems which do not support Docker for Mac (probably similarly for PCs). If someone faces an issue, they'll likely face it for any container they try to bring up - independent of Drill. So this is more generic Docker troubleshooting. I'm thinking, we can just have a link to this instead? 
(5) Volumes will be helpful - for now, i'll share a link to this in the doc. Will try it out and include examples in the doc Bridget will add - as it's a more user facing doc (ex: how to query data on your Mac when using Drill Docker container)

arina-ielchiieva: (1) - Sure, thanks.
(2) - I had to do port forwarding on my machine (Windows 10), it did not work without it, neither  is with localhost, had to use ip + port.
(3) - Sure, thanks
(4) - had an issue on my env Windows 10, you just can mention it for Windows users. Drill is resource consuming, for example, I was able to bring up mysql with default docker machine.
(5) - Sure, thanks

Agirish: Done, please check. 
Regarding (2), i'll update the doc with port forwarding (specific to Windows perhaps). 
For (4), I've mentioned it in the pre-requisites.

arina-ielchiieva:
+1, LGTM. Please don't forget to create a follow up Jira for the enhancements.

Agirish:
I've created DRILL-6600, DRILL-6599, DRILL-6598 & DRILL-6597 to track the enhancements. 

@arina-ielchiieva, thanks for your review!

PR #1349, opened 2018-06-28
DRILL-6554: Minor code improvements in parquet statistics handling

vrozov:
@arina-ielchiieva Please review

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/ParquetIsPredicate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/ParquetIsPredicate.java
@@ -113,9 +113,9 @@ public class ParquetIsPredicate<C extends Comparable<C>> extends LogicalExpressi
    * IS TRUE predicate.
    */
   private static LogicalExpression createIsTruePredicate(LogicalExpression expr) {
-    return new ParquetIsPredicate<Boolean>(expr,
+    return new ParquetIsPredicate<Boolean>(expr, (exprStat, evaluator) ->
         //if max value is not true or if there are all nulls  -> canDrop
-        (exprStat, evaluator) -> !((BooleanStatistics)exprStat).getMax() || isAllNulls(exprStat, evaluator.getRowCount())
+        isAllNulls(exprStat, evaluator.getRowCount()) || exprStat.hasNonNullValue() && !((BooleanStatistics) exprStat).getMax()
     );
   }
 
@@ -123,9 +123,9 @@ public class ParquetIsPredicate<C extends Comparable<C>> extends LogicalExpressi
    * IS FALSE predicate.
    */
   private static LogicalExpression createIsFalsePredicate(LogicalExpression expr) {
-    return new ParquetIsPredicate<Boolean>(expr,
+    return new ParquetIsPredicate<Boolean>(expr, (exprStat, evaluator) ->
         //if min value is not false or if there are all nulls  -> canDrop
-        (exprStat, evaluator) -> ((BooleanStatistics)exprStat).getMin() || isAllNulls(exprStat, evaluator.getRowCount())
+        isAllNulls(exprStat, evaluator.getRowCount()) || exprStat.hasNonNullValue() && ((BooleanStatistics) exprStat).getMin()
     );
   }
 
@@ -133,9 +133,9 @@ public class ParquetIsPredicate<C extends Comparable<C>> extends LogicalExpressi
    * IS NOT TRUE predicate.
    */
   private static LogicalExpression createIsNotTruePredicate(LogicalExpression expr) {
-    return new ParquetIsPredicate<Boolean>(expr,
+    return new ParquetIsPredicate<Boolean>(expr, (exprStat, evaluator) ->
         //if min value is not false or if there are no nulls  -> canDrop
-        (exprStat, evaluator) -> ((BooleanStatistics)exprStat).getMin() && hasNoNulls(exprStat)
+        hasNoNulls(exprStat) && exprStat.hasNonNullValue() && ((BooleanStatistics) exprStat).getMin()
     );
   }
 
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/ParquetPredicatesHelper.java
```diff
@@ -39,22 +40,21 @@ static boolean isNullOrEmpty(Statistics stat) {
    *
    * @param stat parquet column statistics
    * @param rowCount number of rows in the parquet file
-   * @return True if all rows are null in the parquet file
-   *          False if at least one row is not null.
+   * @return <tt>true</tt> if all rows are null in the parquet file and <tt>false</tt> otherwise
    */
   static boolean isAllNulls(Statistics stat, long rowCount) {
-    return stat.isNumNullsSet() && stat.getNumNulls() == rowCount;
+    Preconditions.checkArgument(rowCount >= 0, String.format("negative rowCount %d is not valid", rowCount));
```
arina-ielchiieva: Why do you want to fail in this case? Maybe just warning?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -417,16 +417,9 @@ public static DateCorruptionStatus checkForCorruptDateValuesInStatistics(Parquet
...
-          Integer max = (Integer) statistics.genericGetMax();
-          if (statistics.hasNonNullValue()) {
-            if (max > ParquetReaderUtility.DATE_CORRUPTION_THRESHOLD) {
-              return DateCorruptionStatus.META_SHOWS_CORRUPTION;
-            }
-          } else {
-            // no statistics, go check the first page
-            return DateCorruptionStatus.META_UNCLEAR_TEST_VALUES;
-          }
+          IntStatistics statistics = (IntStatistics)footer.getBlocks().get(rowGroupIndex).getColumns().get(colIndex).getStatistics();
```
arina-ielchiieva: 1. space after `(IntStatistics)`: `(IntStatistics) footer.getBlocks()...`
2. What if statistics is not int one? Maybe check for instanceOf?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/ParquetIsPredicate.java
```diff
@@ -113,39 +113,39 @@ public boolean canDrop(RangeExprEvaluator<C> evaluator) {
    * IS TRUE predicate.
    */
   private static LogicalExpression createIsTruePredicate(LogicalExpression expr) {
-    return new ParquetIsPredicate<Boolean>(expr,
+    return new ParquetIsPredicate<Boolean>(expr, (exprStat, evaluator) ->
         //if max value is not true or if there are all nulls  -> canDrop
-        (exprStat, evaluator) -> !((BooleanStatistics)exprStat).getMax() || isAllNulls(exprStat, evaluator.getRowCount())
+        isAllNulls(exprStat, evaluator.getRowCount()) || exprStat.hasNonNullValue() && !((BooleanStatistics)exprStat).getMax()
```
arina-ielchiieva: Please add space after `(BooleanStatistics)` here in in three cases below.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/ParquetPredicatesHelper.java
vrozov: When can rowCount be negative?

arina-ielchiieva: I don't know, why you have added this check then? :)

vrozov: To avoid bugs :). It is invalid to call `isAllNulls()` with negative `rowCount`.

arina-ielchiieva: I think this check is not needed here :) Why this method should check what was given to it? It's job just to compare row count and number of nulls.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
vrozov: Please see parquet format spec. `DATE` is always `int32`.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/stat/ParquetPredicatesHelper.java
vrozov: To validate input. It can't give the correct answer if an input is a junk (`rowCount` is negative).

arina-ielchiieva: Ok, then just please describe what will happen during query execution if negative row comes into this method, will query fail or statistics will be ignored and there will not filter push down? I hope for the second.

vrozov: I hope for the reverse. Negative row count passed to this method indicates a bug, and bug means incorrect result. I'd rather fail a query than give a wrong result.

arina-ielchiieva: But this negative rowCount could only come from the statistics, I don't think that bad statistics should prevent user from executing the query.

vrozov: It does not matter where it comes from (it actually comes from `RowGroupInfo`). The condition needs to be intercepted prior to calling `isAllNulls()` as `isAllNulls()` would return the wrong result (whether true or false) for a negative row count.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
vrozov: I don't see any specific code style in regards to spacing for casting (grep -I -R \(\([a-zA-Z0-9_]*\)[a-zA-Z0-9_] * | grep -c java). It seems to be a preference of a contributor.

arina-ielchiieva:

PR #1351, opened 2018-06-29
DRILL-6543: Disable Hash Join fallback, add percent_reserved_allowance_from_direct

Ben-Zvi:
(1) Turned the default for Hash Join Fallback to **false**.
(2) Added a new option: **planner.memory.percent_reserved_allowance_from_direct**
     This new option enforces (when a query with buffered operators is run) reserving part of the Direct Memory for the non-buffered operators. It checks against the two ways to set the memory (for the buffered operators): Memory per query per node, or Percent per query.
Also added a test for the enforcement.


exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -490,6 +491,31 @@ private ExecConstants() {
...
+   * </p>
+   * <p>
+   * Note that this enforcement is only good for the non concurrent case. For multiple
+   * concurrent queries, other queries may grab some unaccounted for Direct Memory. For
+   * concurrent work, better set the option "planner.memory.percent_per_query" correctly.
+   * </p>
+   * <p>
+   * DEFAULT: 25%
+   * </p>
+   */
```
ilooner: Thanks for adding this documentation.

exec/java-exec/src/test/java/org/apache/drill/exec/util/TestQueryMemoryAlloc.java
```diff
@@ -120,6 +123,53 @@ public void testCustomPercent() throws Exception {
     }
   }
 
+  /**
+   *  Test that the percent_reserved_allowance_from_direct limitation works
+   */
+  @Test
+  public void testReservedAllowanceCheck() throws Exception {
+    OperatorFixture.Builder builder = OperatorFixture.builder(dirTestWatcher);
```
ilooner: Thanks for adding a unit test.

exec/java-exec/src/main/java/org/apache/drill/exec/util/MemoryAllocationUtilities.java
```diff
@@ -138,16 +139,36 @@ public static long computeOperatorMemory(OptionSet optionManager, long maxAllocP
   @VisibleForTesting
   public static long computeQueryMemory(DrillConfig config, OptionSet optionManager, long directMemory) {
 
+    // Get the options
+    double percent_per_query = optionManager.getOption(ExecConstants.PERCENT_MEMORY_PER_QUERY);
+    long max_query_per_node = optionManager.getOption(ExecConstants.MAX_QUERY_MEMORY_PER_NODE);
+    double percent_allowance = optionManager.getOption(ExecConstants.PERCENT_RESERVED_ALLOWANCE_FROM_DIRECT);
+
+    // verify that the allowance is kept
+    if ( percent_per_query + percent_allowance > 1.0 ) {
```
ilooner: Why do we need to make sure these add to one? Couldn't we just reserve our allowance from the previously computed maxAllocPerNode value. For example

```
    double percentAllowance = optionManager.getOption(ExecConstants.PERCENT_RESERVED_ALLOWANCE_FROM_DIRECT);
    // Memory computed as a percent of total memory.
    long perQueryMemory = Math.round(directMemory *
        optionManager.getOption(ExecConstants.PERCENT_MEMORY_PER_QUERY));

    // But, must allow at least the amount given explicitly for
    // backward compatibility.

    perQueryMemory = Math.max(perQueryMemory,
        optionManager.getOption(ExecConstants.MAX_QUERY_MEMORY_PER_NODE));

    // Compute again as either the total direct memory, or the
    // configured maximum top-level allocation (10 GB).

    long maxAllocPerNode = Math.min(directMemory,
        config.getLong(RootAllocatorFactory.TOP_LEVEL_MAX_ALLOC));

    // Final amount per node per query is the minimum of these two.
    maxAllocPerNode = Math.min(maxAllocPerNode, perQueryMemory);

    // Deduct non buffered allowance
    return Math.round(maxAllocPerNode * (1.0 - percentAllowance));
```

exec/java-exec/src/main/java/org/apache/drill/exec/util/MemoryAllocationUtilities.java
```diff
@@ -138,16 +139,36 @@ public static long computeOperatorMemory(OptionSet optionManager, long maxAllocP
   @VisibleForTesting
   public static long computeQueryMemory(DrillConfig config, OptionSet optionManager, long directMemory) {
 
+    // Get the options
+    double percent_per_query = optionManager.getOption(ExecConstants.PERCENT_MEMORY_PER_QUERY);
```
ilooner: Minor nitpick, but the convention in Java is to use Camel case for names

`percentPerQuery`

Ben-Zvi: Fixed ....

PR #1354, opened 2018-06-30
DRILL-6570: Fixed IndexOutofBoundException in Parquet Reader

sachouche:
Reserving same size intermediary buffers to handle the case of false-positive; that is, a column is first thought to be fixed length (after sampling), then reverted to variable length as we read more values.


sachouche:
@kkhatua, can you please review this fix?

Thanks!

kkhatua:
@sachouche  The change looks simple, but I might be missing the bigger picture. For e.g., `ColumnPrecisionInfo.precision` is passed as an unused argument. Not clear as to why you would not need it still.
Perhaps someone else who is more familiar could review this.

sachouche:
The reason I didn't remove the ColumnPrecisionInfo.precision is that I am allowing myself another chance, in the future, to optimize the fixed length path; at this time, the savings are very less (about 2kb). Though, I can remove this input object if it helps code clarity.

kkhatua:
The IDE (or Maven, itself) might report this as unused. It might be worth mentioning as a that this is a placeholder for the future. 
That said, I'd still advice @Ben-Zvi  or someone else familiar with the original PR review this.

PR #1357, opened 2018-07-02
DRILL-6557: Use size in bytes during Hive statistics calculation if present

arina-ielchiieva:
1. Check size in bytes presence in stats before fetching input splits and use it if present.
2. Add log trace suggesting to use ANALYZE command before running queries if statistics is unavailable and Drill had to fetch all input splits.
3. Minor refactoring /  cleanup in HiveMetadataProvider class.


arina-ielchiieva:
@vvysotskyi please review.

PR #1360, opened 2018-07-03
DRILL-6578: Handle query cancellation in Parquet reader

sachouche:
Goal -
- The optimized Parquet reader uses an iterator style to load column data 
- We need to ensure the code can properly handle query cancellation even in the presence of bugs within the hasNext() .. next() calls

Fix Details -
- Added a check within the hasNext() and next() to detect a thread interrupt
- If this is the case, these methods will throw a runtime exception and keep the thread's interrupted state set
- This will ensure that any blocking call will get the InterruptedException to be thrown

### DIFF
--- a/common/src/main/java/org/apache/drill/common/exceptions/DrillRuntimeException.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/DrillRuntimeException.java
@@ -48,4 +48,22 @@ public class DrillRuntimeException extends RuntimeException {
   public static void format(Throwable cause, String format, Object...args) {
     throw new DrillRuntimeException(String.format(format, args), cause);
   }
+
+  /**
+   * This method can be called within loops to check whether the current thread has been
+   * interrupted; it ensures that operator implementation can respond to query cancellation
+   * in a timely manner.
+   *
+   * <p>Calling this method will result in the following behavior:
+   * <ul>
+   * <li>Throws a runtime exception if current thread interrupt flag has been set
+   * <li>Clears current thread interrupt flag
+   * </ul>
+   */
+  public static void checkInterrupted() {
+    if (Thread.interrupted()) {
+      // This exception will ensure the control layer will immediately get back control
+      throw new DrillRuntimeException("Interrupt received; aborting current operation");
+    }
+  }
 }
--- a/exec/vector/src/main/codegen/templates/VariableLengthVectors.java
+++ b/exec/vector/src/main/codegen/templates/VariableLengthVectors.java
@@ -19,7 +19,7 @@ import java.lang.Override;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.util.Set;
-
+import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.exec.exception.OutOfMemoryException;
 import org.apache.drill.exec.memory.AllocationManager.BufferLedger;
 import org.apache.drill.exec.vector.BaseDataValueVector;
@@ -641,6 +641,8 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements V
         if (callback != null) {
           callback.onNewBulkEntry(entry);
         }
+
+        DrillRuntimeException.checkInterrupted(); // Ensures fast handling of query cancellation
       }
 
       // Flush any data not yet copied to this VL container


sachouche:
@vrozov, can you please review this fix?

Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenColumnBulkInput.java
```diff
@@ -109,6 +109,8 @@
   /** {@inheritDoc} */
   @Override
   public boolean hasNext() {
+    checkCancellation(); // Checks whether query cancellation has been called
```
vrozov: consider `checkInterrupted`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenColumnBulkInput.java
```diff
@@ -479,6 +483,17 @@ private void deinitOverflowData() {
     fieldOverflowStateContainer = null;
   }
 
+  private void checkCancellation() {
```
vrozov: move to a common utility class. Make it static.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenColumnBulkInput.java
```diff
@@ -479,6 +483,17 @@ private void deinitOverflowData() {
...
+    // - The methods hasNext() & next() are usually invoked within a loop
+    // - We need to ensure that bugs do not cause such loops to execute forever
+    // - To do that, will check the interrupted status (e.g., cancellation) to break the infinite loop
+    if (Thread.currentThread().isInterrupted()) {
```
vrozov: use `Thread.interruped()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenColumnBulkInput.java
sachouche: do you mean rename the method to checkInterrupted?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenColumnBulkInput.java
sachouche: within Parquet or under exec/util?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenColumnBulkInput.java
sachouche: @vrozov,
- Thread.interruped() will clear the thread interrupted status; I want to propagate the thread interruption status till the higher level framework code clears it to state "ok; received the cancellation request and now ready to handle other requests"
- Are you suggesting that it is not necessary to still continue setting this flag since we are throwing an exception?

please clarify!

vrozov: It is not necessary to keep the flag after the exception is thrown. It will be better to throw InterruptedException outside of Iterator.

sachouche:
@vrozov and @Ben-Zvi  can you please review this PR?

Thanks!

vrozov:
@sachouche If you prefer to throw `DrillRuntimeException`, make `checkInterrupted` a static method of `DrillRuntimeException`, but I would prefer to avoid throwing `DrillRuntimeException` in the iterator as the iterator does not know when to throw this exception. Instead, throw `InterrupedException` when the iterator is used in a loop.

sachouche:
@vrozov, please confirm if this is what you are suggesting:

boolean hasNext() {
try {
  checkInterrupted();
} catch (InterruptedException) {
  throw new DrillRuntimeException(...);
}

same for next()

vrozov:
@sachouche No, there is no point to throw `InterrupedException` and catch it in the same method. I would prefer to check for a thread to be interrupted in a method that calls `hasNext()` and `next()`.

sachouche:
So you suggest that I update the setSafe() method to throw the checked exception InterruptedException. Can we find a middle ground as the main point here is to handle cancellation? what about going back to my original implementation which though a DrillException when interrupt flag is set.

Ben-Zvi:
Is the special file/class RuntimeUtils needed ? (I.e., instead of calling Thread.interrupted() directly).
I may be a bit biased against proliferation of files/classes that don't do much work ....

sachouche:
Suggestion made by @vrozov.. though I agree, the new method doesn't do much.

vrozov:
I did not suggest to introduce a new class. My suggestion was to move the method for example to DrillRuntimeException.

vrozov:
Yes, I would prefer to check for an interrupt in setSafe() (outside of an iterator).

vrozov:
@sachouche I guess, you mean raise, not handle. For the rest
a) if you throw DrillRuntimeException, please see c)
b) yes (move it to a method that uses an iterator, if it is `setSafe()`, then move it to that method)
c) I would prefer InterruptedException over DrillRuntimeException, but if it is too much to propagate InterruptedException, I am OK with DrillRuntimeException.
d) yes, when an exception is raised, the interrupt flag needs to be cleared to avoid throwing exception again.

sachouche:
@vrozov , thanks Vlad, working on it!

sachouche:
@vrozov and @Ben-Zvi 
Updated the PR according to latest feedbacks.

sachouche:
@vrozov,
can you please review this PR? I have implemented the changes as agreed (last week).

sachouche:
Thanks @vrozov for the review!
@Ben-Zvi, can you please review this PR?

sachouche:
Thank you both for the review!

PR #1361, opened 2018-07-03
DRILL-6579: Added sanity checks to the Parquet reader to avoid infini…

sachouche:
…te loops

Added sanity checks to avoid infinite loops.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
@@ -40,9 +40,9 @@ public abstract class BatchReader {
     ColumnReader<?> firstColumnStatus = readState.getFirstColumnReader();
     int currBatchNumRecords = readState.batchSizerMgr().getCurrentRecordsPerBatch();
     long recordsToRead = Math.min(currBatchNumRecords, readState.getRemainingValuesToRead());
-    int readCount = readRecords(firstColumnStatus, recordsToRead);
-
+    int readCount = recordsToRead > 0 ? readRecords(firstColumnStatus, recordsToRead) : 0;
     readState.fillNullVectors(readCount);
+
     return readCount;
   }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBulkPageReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBulkPageReader.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.store.parquet.columnreaders;
 
+import com.google.common.base.Preconditions;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
@@ -106,11 +107,13 @@ final class VarLenBulkPageReader {
     pageInfo.dictionaryValueReader = pageInfoInput.dictionaryValueReader;
     pageInfo.numPageValues = pageInfoInput.numPageValues;
     if (clear) {
-    buffer.clear();
-  }
+      buffer.clear();
+    }
   }
 
   final VarLenColumnBulkEntry getEntry(int valuesToRead) {
+    Preconditions.checkArgument(valuesToRead > 0, "Number of values to read [%s] should be greater than zero", valuesToRead);
+
     VarLenColumnBulkEntry entry = null;
 
     // If there is overflow data, then we need to consume it first
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenFixedEntryReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenFixedEntryReader.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.store.parquet.columnreaders;
 
+import com.google.common.base.Preconditions;
 import java.nio.ByteBuffer;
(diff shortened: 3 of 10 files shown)


sachouche:
@Ben-Zvi 

Can you please review this PR?

Thanks!

sachouche:
@Ben-Zvi 
A Drill user (version 1.12) reported a Parquet issue which indicated two different problems
a) The Parquet code was running within an infinite loop; we don't have any repro for this issue. Code inspection showed that if the number-of-rows to load is zero than this condition could arise. So fixed this possibility and added sanity checks 
b) When an infinite loop arises, the user could not cancel the query; so I created another JIRA DRILL-6578 to detect thread interruption within the main processing loop to ensure query cancellation always succeeds

sachouche:
Thank you @Ben-Zvi for the review!

sachouche:
@vrozov, updated PR with your feedback.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenFixedEntryReader.java
```diff
@@ -37,14 +38,15 @@
   /** {@inheritDoc} */
   @Override
   final VarLenColumnBulkEntry getEntry(int valuesToRead) {
-    assert columnPrecInfo.precision >= 0 : "Fixed length precision cannot be lower than zero";
+    Preconditions.checkArgument(columnPrecInfo.precision >= 0, "Fixed length precision cannot be lower than zero");
```
vrozov: checkState?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenNullableFixedEntryReader.java
```diff
@@ -38,14 +39,16 @@
   /** {@inheritDoc} */
   @Override
   final VarLenColumnBulkEntry getEntry(int valuesToRead) {
-    assert columnPrecInfo.precision >= 0 : "Fixed length precision cannot be lower than zero";
+    Preconditions.checkArgument(columnPrecInfo.precision >= 0, "Fixed length precision cannot be lower than zero");
```
vrozov: Should the check be here or in the constructor?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenFixedEntryReader.java
sachouche: Pushed the check to the CTOR.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenNullableFixedEntryReader.java
sachouche: You are correct; pushed the check to the CTOR.

sachouche:
@vrozov 
can you please approve this PR as there is a deadline for drill 1.14?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/BatchReader.java
```diff
@@ -40,8 +40,11 @@ public int readBatch() throws Exception {
     ColumnReader<?> firstColumnStatus = readState.getFirstColumnReader();
     int currBatchNumRecords = readState.batchSizerMgr().getCurrentRecordsPerBatch();
     long recordsToRead = Math.min(currBatchNumRecords, readState.getRemainingValuesToRead());
-    int readCount = readRecords(firstColumnStatus, recordsToRead);
+    int readCount = 0;
```
vrozov: use `?`: `int readCount = recordsToRead > 0 ? ...`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBulkPageReader.java
```diff
@@ -111,6 +112,8 @@ final void set(PageDataInfo pageInfoInput, boolean clear) {
   }
 
   final VarLenColumnBulkEntry getEntry(int valuesToRead) {
+    Preconditions.checkArgument(valuesToRead > 0, "Number values to read cannot be zero");
```
vrozov: Consider `String.format("Number of values to read %d must be positive", valuesToRead);`

sachouche:
@vrozov 
Performed the requested changes; can you please approve this PR?

Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenBulkPageReader.java
```diff
@@ -106,11 +107,13 @@ final void set(PageDataInfo pageInfoInput, boolean clear) {
...
     pageInfo.numPageValues = pageInfoInput.numPageValues;
     if (clear) {
-    buffer.clear();
-  }
+      buffer.clear();
+    }
   }
 
   final VarLenColumnBulkEntry getEntry(int valuesToRead) {
+    Preconditions.checkArgument(valuesToRead > 0, String.format("Number of values to read [%d] should be greater than zero", valuesToRead));
```
vrozov: Please apply the same to all other places where there is a check for argument or state to be greater than zero.

vrozov: It may also be better to use Preconditions template instead of String template (sorry for the wrong example in my prior comment).

sachouche:
@vrozov 
Made the requested modifications.

PR #1363, opened 2018-07-05
DRILL-6549: batch sizing for nested loop join

ppadma:


ppadma:
@sohami @Ben-Zvi can one of you review this ? Pretty straightforward changes.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/unit/TestOutputBatchSize.java
```diff
@@ -2287,6 +2292,341 @@ public void testHashAggMax() throws ExecutionSetupException {
...
 
+  @Test
+  public void testNestedLoopJoinMultipleOutputBatches() throws Exception {
+    LogicalExpression functionCallExpr = new FunctionCall("equal",
+            ImmutableList.of((LogicalExpression) new FieldReference("c1", ExpressionPosition.UNKNOWN),
+                    (LogicalExpression) new FieldReference("c2", ExpressionPosition.UNKNOWN)),
+            ExpressionPosition.UNKNOWN);
+
+    NestedLoopJoinPOP nestedLoopJoin = new NestedLoopJoinPOP(null, null, JoinRelType.INNER, functionCallExpr);
+    mockOpContext(nestedLoopJoin, initReservation, maxAllocation);
```
Ben-Zvi: Just a comment: This test calls `mockOpContext()`, while the other new NLJ tests don't. I tested all with and without this call, and it seems to make no difference anyway ....

PR #1366, opened 2018-07-06
[DRILL-6581] C++ Client SSL Implementation Fixes/Improvements

superbstreak:
- Fix: Hostname verification doesnt function as expected. 
       > Host and port passed in to the callback are always empty. Parsing the connection string before we use the host name.
       > Connection string port (31010) is not required for the hostname verification.
- Fix: Certificate load verification exceptions are swallowed and not propagated.
- Improvement: SSL V3 is not disabled.
       > Disabled SSLv2 and SSLv3.
- Improvement: Hostname verification will throw the same error message as certificate exception.  
       > Separated them to allow better user error handling/debugging.
- Improvement: Create individual error messages to allow error handling of the application using the client and follows the standard of the rest of the errors
- Improvement: [DRILL-6586] Add SSL Hostname verification with zookeeper connection mode support
- Improvement: [DRILL-6587] Added support for custom SSL CTX Options


contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -60,11 +66,24 @@ class UserProperties;
...
         public:
-            ChannelContext(DrillUserProperties* props):m_properties(props){};
+            ChannelContext(DrillUserProperties* props):m_properties(props),m_hasCertHostnameVerificationErr(false){};
             virtual ~ChannelContext(){};
             const DrillUserProperties* getUserProperties() const { return m_properties;}
+
+            /// @brief Check the certiticate hostname verification status.
+            /// 
+            /// @return if the verification has failed.
+            const bool HasCertHostnameVerificationError() const {return m_hasCertHostnameVerificationErr;}
```
superbstreak: I think we should put these inside SSLChannelContext since these are SSL related items

priteshm:
@parthchandra can you please review this?

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -199,6 +248,29 @@ class UserProperties;
...
+            /// @brief Handle protocol handshake exceptions for SSL specific failures.
+            /// 
+            /// @param in_errmsg                The error message.
+            /// 
+            /// @return the connectionStatus.
+            connectionStatus_t HandleProtocolHandshakeException(const char* errmsg) {
+                if (!(((SSLChannelContext_t *)m_pContext)->GetCertificateHostnameVerificationStatus())){
+                    return handleError(
+                        CONN_HANDSHAKE_FAILED,
+                        getMessage(ERR_CONN_SSL_CN));
```
sohami: Would be useful to preserve the original error message in this case as well. Please change to `getMessage(ERR_CONN_SSL_CN, errmsg)`

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -21,6 +21,17 @@
...
+#include "errmsgs.hpp"
+
+#if defined(IS_SSL_ENABLED)
+#include <openssl/ssl.h>
+#endif
+
+namespace
+{
+// The error message to indicate certificate verification failure.
+#define DRILL_BOOST_SSL_CERT_VERIFY_FAILED  "handshake: certificate verify failed\0"
```
sohami: I don't think we can rely on this error string. Instead it would be good to use something like below for decoding ssl errors.
https://stackoverflow.com/questions/9828066/how-to-decipher-a-boost-asio-ssl-error-code

contrib/native/client/src/clientlib/channel.cpp
```diff
@@ -211,6 +211,21 @@ ChannelContext* ChannelFactory::getChannelContext(channelType_t t, DrillUserProp
             }
 
             pChannelContext = new SSLChannelContext(props, tlsVersion, verifyMode);
+
+            if (props->isPropSet(USERPROP_CUSTOM_SSLCTXOPTIONS)){
```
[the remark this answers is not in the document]
superbstreak: This is an optional settings that the client application can set since the default configuration from boost is somewhat limited. And also based on the version of boost the client is compiled against, some configurations are not available.

contrib/native/client/src/clientlib/channel.hpp
```diff
@@ -215,6 +287,52 @@ class UserProperties;
...
+                bool verified = verifier(in_preverified, in_ctx);
+
+                DRILL_LOG(LOG_DEBUG) 
+                    << "DrillSSLHostnameVerifier::operator(): Verification Result: " 
+                    << verified 
+                    << std::endl;
+
+                // Sets the result back to the context.
+                context->SetCertHostnameVerificationStatus(verified);
+                return verified && in_preverified;
```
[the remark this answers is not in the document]
superbstreak: Oh yes, redundant! Thanks!

contrib/native/client/src/clientlib/channel.hpp
superbstreak: Will do Thanks!

contrib/native/client/src/clientlib/channel.hpp
superbstreak: Yup, I was concern about this bit, too. Will do, thanks!

superbstreak:
@sohami updated. Please review.

PR #1367, opened 2018-07-09
DRILL-6585: PartitionSender clones vectors, but shares field metdata

paul-rogers:
See the discussion for [PR #1244 for DRILL-6373](https://github.com/apache/drill/pull/1244).

The PartitionSender clones vectors. But, it does so by reusing the MaterializedField from the original vector. Though the original authors of MaterializedField apparently meant it to be immutable, later changes for maps and unions ended up changing it to add members.

When cloning a map, we get the original map materialized field, then start doctoring it up as we add the cloned map members. This screws up the original map vector's metadata.

The solution is to clone an empty version of the materialized field when creating a new vector.
But, since much code creates vectors by giving a perfectly valid, unique materialized field, we want to add a new method for use by the ill-behaved uses, such as PartitionSender, that ask to create a new vector without cloning the materialized field.

The solution is to add a new method, `TypeHelper.getClonedVector()`, which handles the "new vector from the materialized field of an existing vector" case. Modified the partition sender to use this version.

Moved an existing method to group the "new vector" functions.

My IDE is set up to mark variables as final when possible; so the edited files also have these changes.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java
@@ -376,9 +376,8 @@ public abstract class PartitionerTemplate implements Partitioner {
      */
     public void initializeBatch() {
       for (VectorWrapper<?> v : incoming) {
-        // create new vector
-        @SuppressWarnings("resource")
-        ValueVector outgoingVector = TypeHelper.getNewVector(v.getField(), allocator);
+        // create new vector by cloning the incoming vector's type
+        ValueVector outgoingVector = TypeHelper.getNewVector(v.getField().clone(), allocator);
         outgoingVector.setInitialCapacity(outgoingRecordBatchSize);
         vectorContainer.add(outgoingVector);
       }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java
@@ -377,7 +377,7 @@ public abstract class PartitionerTemplate implements Partitioner {
     public void initializeBatch() {
       for (VectorWrapper<?> v : incoming) {
         // create new vector by cloning the incoming vector's type
-        ValueVector outgoingVector = TypeHelper.getNewVector(v.getField().clone(), allocator);
+        ValueVector outgoingVector = TypeHelper.getNewVector(v.getField().cloneEmpty(), allocator);
         outgoingVector.setInitialCapacity(outgoingRecordBatchSize);
         vectorContainer.add(outgoingVector);
       }


paul-rogers:
@vrozov, please review. This is the fix for the issue we discussed a few weeks ago.

exec/vector/src/main/codegen/templates/BasicTypeHelper.java
```diff
@@ -233,10 +227,54 @@ public static ValueVector getNewVector(String name, BufferAllocator allocator, M
...
+   * @param allocator allocator to use to allocate memory for the vector
+   * @param callBack callBack when the schema of the vector changes (not typically
+   * used in modern Drill)
+   * @return the new vector with no memory allocated
+   * @see #getNewVector(MaterializedField, BufferAllocator, CallBack) for when
+   * the metadata need not be cloned
+   */
+  
+  public static ValueVector getClonedVector(MaterializedField field, BufferAllocator allocator, CallBack callBack) {
+    MaterializedField newField = MaterializedField.create(field.getName(), field.getType());
```
Ben-Zvi: Isn't `field.getType()` just returning the type as is ?  So where does the cloning happen ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/orderedpartitioner/OrderedPartitionRecordBatch.java
```diff
@@ -23,6 +23,7 @@
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.calcite.rel.RelFieldCollation.Direction;
```
vrozov: I'd prefer to avoid any formatting changes including `import` reorder unless a method, a class or a file is modified to include a fix. I don't see anything being fixed in `OrderedPartitionRecordBatch.java` except for `final` declarations.

exec/vector/src/main/codegen/templates/BasicTypeHelper.java
```diff
@@ -233,10 +227,54 @@ public static ValueVector getNewVector(String name, BufferAllocator allocator, M
       throw new UnsupportedOperationException(buildErrorMessage("get holder reader implementation", type, mode));
   }
   
-  public static ValueVector getNewVector(MaterializedField field, BufferAllocator allocator){
+  public static ValueVector getNewVector(String name, BufferAllocator allocator, MajorType type, CallBack callBack) {
+    MaterializedField field = MaterializedField.create(name, type);
+    return getNewVector(field, allocator, callBack);
+  }  
+  
+  public static ValueVector getNewVector(MaterializedField field, BufferAllocator allocator) {
```
vrozov: `getNewVector(MaterializedField field, ...` promotes `field` sharing between vectors and possibly threads. What are the use cases where `MaterializedField` needs to be shared between vectors? I mean, when is it necessary that modifications to `MaterializedField` that belongs to one vector are reflected transparently in another vector `MaterializedField`?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/orderedpartitioner/OrderedPartitionRecordBatch.java
paul-rogers: I'll back out the changes my editor made. I'll redo the changes with the current project options turned off. Stay tuned.

exec/vector/src/main/codegen/templates/BasicTypeHelper.java
paul-rogers: No sharing should ever occur. Vectors under construction are not sharable; only completed, finalized vectors are shareable. (None of the vector builder methods are synchronized. Not sure what it even means for two threads to write to the same vector as they must coordinate on rows, and thread T1 can't write its VARCHAR to row 2 until thread T2 writes its value to row 1.)

exec/vector/src/main/codegen/templates/BasicTypeHelper.java
paul-rogers: The cloning is creating a new `MaterializedField`. Though you do point out a good point: the `MajorType` includes subtypes for a union vector; those need to be dropped here. I'll add code that rebuilds the `MajorType` from the `MinorType` and `Mode`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/orderedpartitioner/OrderedPartitionRecordBatch.java
paul-rogers: Turned out this file was the wrong one: this code is never used. Should have fixed the `PartitionSenderTemplate` rather than this one. Reverted changes.

exec/vector/src/main/codegen/templates/BasicTypeHelper.java
paul-rogers: This is a red herring. This file is never used. Please see the correct fix in `PartionSenderTemplate`.

exec/vector/src/main/codegen/templates/BasicTypeHelper.java
paul-rogers: Fixed.

exec/vector/src/main/codegen/templates/BasicTypeHelper.java
paul-rogers: I perhaps misunderstood the original comment.

There are two ways that `getNewVector` is used:

* The caller builds up a `MaterializedField` and passes it to the new vector. No sharing occurs. This is the normal case used by, as far as I can tell, all but one place in the code.
* The `PartitionSenderTemplate` creates a new vector using the type of a previous vector. Here, the `MaterializedField` is already in use and must be cloned.

There seems no good reason to clone every `MaterializedField` since most uses are correct. In fact, the new result set loader code rather counts on the fact that the passed-in `MaterializeField` is the one that the vector uses.

We could possibly insert code into the `getNewVector()` that triggers an assert if the `MaterializedField` already has children (or its minor type has subtypes). That will catch field sharing if it occurs again.

Doing so might uncover other improper usages, so is a bit out of the scope of this PR. Good exercise for someone who has access to the full pre-commit test suite.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java
```diff
@@ -376,9 +376,8 @@ public void updateStats(FragmentWritableBatch writableBatch) {
      */
     public void initializeBatch() {
       for (VectorWrapper<?> v : incoming) {
-        // create new vector
-        @SuppressWarnings("resource")
-        ValueVector outgoingVector = TypeHelper.getNewVector(v.getField(), allocator);
+        // create new vector by cloning the incoming vector's type
+        ValueVector outgoingVector = TypeHelper.getClonedVector(v.getField(), allocator);
```
paul-rogers: This is the heart of the change: this is the place where we were reusing previously-owned materialized fields.

paul-rogers:
@vrozov, can you take another look at this PR to see if I have addressed your concerns?

Turned out that I was able to find an existing `MaterializedField.clone()` method. Since only the `PartitionerTemplate` seems to have an issue, I just inserted the extra call directly there, avoiding the need to modify any other files.

If we can test this fix using PR #1244, then perhaps we can get this fix into this week's batch commit, PR #1244 into next week's, after which I can do the final PR for this batch of changes.

vrozov:
@paul-rogers I am somewhat OK with the code changes granted that they fix the issue. My concerns are (in random order)

- Usage of the `clone()` where copy constructor should be used (see [Copy Constructor versus Cloning](https://www.artima.com/intv/bloch13.html))
- Improper implementation of `clone()` by  the `MaterializedField`
- Unclear semantics when `MaterializedField` needs to be cloned
- Should not by the time `InitializedBatch` is called, `MaterializedField` for incoming vectors be fully initialized? If they are, what causes changes in outgoing vectors `MaterializedField` and if they are not, how changes to `MaterializedField` in incoming vectors will be propagated to outgoing vectors `MaterializedField` once they refer to a deep copy?

paul-rogers:
@vrozov, as usual, you zeroed in on two issues.

First, it turns out that we have both a `clone()` and `cloneEmpty()`. I should have used `cloneEmpty()`. This is a problem with not being able to run the pre-commit tests, I can't catch such bone-headed errors...

Second, the `clone()` method does not even make sense. It will clear out child fields for the `MaterializedField`, but *not* the subtypes for the major type. This is very likely a bug. Why have we never caught it? Turns out that `clone()` is not called.

The same bug exists for all of the `withPathAndType()` variations: we preserve subtypes. But, the semantics of the (mostly broken) union vectors appear to want a new vector to be given a materialized field without subtypes; subtypes are added later as the types are needed. Is that behavior itself a bug or a feature? We can't be sure because union vectors are not supported my much of Drill.

This makes this a hard case. Since the behavior is murky, I'd suggest leaving well enough alone, and just fixing the one problem we know about: passing a fully-populated materialized field to a new vector.

Might be worthwhile filing a JIRA to track this issue. (Though, we never seem to fix such tickets...)

Running tests with the other PR (with the above fix) should tell us if the fix works, since that PR found the issue.

To your fundamental question, please see the discussion in the original PR and in the description for details. The story is, in essence:

* There is a fully-built vector coming into the partitioner.
* When that incoming vector was built, the code for a map added children to the `MaterializedField`, and for unions, added a subtype to the `MajorType`.
* For whatever reason, the partition wants to make a new copy of that vector using the same structure.
* If we reuse the same vector, then the new code will modify it as we build up the maps or unions, causing concurrent modifications to the source vector's field while building the destination.
* If we do a simple copy, we'll carry over child fields and subtypes, which will likely cause problems as we build up the new vector's structure.

The original code would only work for non-nullable, fixed-width types (with no internal structure) I have no idea why it started to fail with my result set loader changes. Probably because, to get things to work, I had to fix bugs in the vector code in which we were not properly maintaining the material field structure, which caused problems with the new code tried to copy a vector structure to create the "overflow vector."

sohami:
@paul-rogers - I have couple of questions related to this change and your last comment:
1) I don't understand what you mean by `clone` method will clear out the child field. Isn't `MaterializedField::clone` is taking care of [cloning the children fields](https://github.com/apache/drill/blob/master/exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java#L145) as well ? Instead `cloneEmpty` is throwing away the children fields.

2) The main issue with the original PR and the logic of building outgoing batch from incoming batch of Partition Sender is something like below:

- In original PR there is a change for NullableValueVectors to add the `values` and `bits` vector materialized field as child field of parent vector field. See [here](https://github.com/apache/drill/pull/1244/files#diff-646d5ac291e14369d72634752edaae84R111). From your comment it looks like because the internal `values` ValueVector mode needs to be required so you are creating another Materialized Field with that mode for internal `values` vector and adding it as child of parent vector field. But AFAIK based on the usage seen, the children of a Materialized Field is only used in case of container Type vector like Map not in case of composite vector like NullableValueVectors. For those the internal `values` and `bits` vector holds on to the parent Materialized Field. Now with change even for Nullable ValueVectors the children component of parent field will be populated. This will probably affect the serde component and compatibility too ?

- Now coming to the main issue. Earlier for each incoming of the partition sender, there are multiple threads which are trying to create outgoing batches but using the same Materialized from incoming batch vectors. Since outgoing batch will have same schema as incoming batch.

- When a ValueVector in incoming batch is of Map type then while creating a new ValueVector of Map type in outgoing batch it passes the Materialized Field from input Map Vector. In the constructor of [AbstractMapVector](https://github.com/apache/drill/blob/master/exec/vector/src/main/java/org/apache/drill/exec/vector/complex/AbstractMapVector.java#L49), it clones the input field and then iterate over the children of **not** the cloned field but the original input field (which is from same incoming vector). This looks to be a bug where it should [iterate on cloned field children rather than input field children](https://github.com/apache/drill/blob/master/exec/vector/src/main/java/org/apache/drill/exec/vector/complex/AbstractMapVector.java#L51).

- With above if there is a child of the Map vector which is of Nullable type, then that child vector will get reference to the children field from actual input field to parent MapVector (not the reference to cloned children field).

- With [new change](https://github.com/apache/drill/pull/1244/files#diff-646d5ac291e14369d72634752edaae84R111) in NullableValueVector it will try to modify it's input Materialized Field to add children for `bit` and `values` vector field.

- So combination of above bug in Map vector and the new change will cause the concurrent modification exception as NullableValueVector child of MapVector will modify the Materialized field of incoming ValueVector not the new clone Materialized Field created for the outgoing vector.

With your current fix I think you should still use `clone` instead of `cloneEmpty` since we need to preserve the children information too in outgoing batches of PartitionSender.

paul-rogers:
@sohami, thanks for your comments and questions. Unfortunately, I cannot debug the use case and so you may have a deeper understanding than I do. I'm working from experience gained some six months ago when working with the result set loader, and that knowledge is getting rusty.

> In original PR there is a change for NullableValueVectors to add the values and bits vector materialized field as child field of parent vector field. ... From your comment it looks like because the internal values ValueVector mode needs to be required so you are creating another Materialized Field with that mode for internal values vector and adding it as child of parent vector field.

The reason for that change is that the result set loader code that clones a vector needs to know the actual type. That code walks the vector tree, using the `MaterializedField` to get the type. If a `values` vector (which has no `bits` vector) reports its type as `Nullable`, then the clone will create a `bits` vector, which causes havoc.

I'm thinking that I should change the cloning code. Rather than believing the `MaterializedField`, I can use the vector class type itself. That will be more clunky and slow, but it will eliminate the need to change the existing vector code.

Given how long this discussion has gone on, that I can't do the required tests, and that we can't we discuss this in person, I'm thinking that the alternative approach may be more expedient.

I suppose a larger question is whether the final bits of the result set loader are even still useful. Much work has been done on batch sizing since this work started. Is it still worth while finishing up this code so we can control the batch size for readers? Parquet has its own solution. Is it worth worrying about the others?

paul-rogers:
Closing this PR. Will try a different approach since we're not converging on a solution.

sohami:
> The reason for that change is that the result set loader code that clones a vector needs to know the actual type. That code walks the vector tree, using the MaterializedField to get the type. If a values vector (which has no bits vector) reports its type as Nullable, then the clone will create a bits vector, which causes havoc.

As of now NullableVectors takes care of creating it's internal `values` and `bits` vector correctly. That is to say it creates correct `values` vector of required type even though the MaterializedField of that vector points to parent field. 
When you traverse the MaterializedField tree to construct a new vector based on it, for Nullable Vector without your change will not have any children fields. So probably based on DataMode of Materialized Field you will know that the vector is of Nullable Type and you will create a NullableValueVector.  Internally this NullableValueVector should create correct `values` vector (i.e. without bit vector) even though it's field will be same as parent NullableVector field. I don't think you have to create `values` vector separately ? May be I am missing something, would be helpful if you can point me to cloning code inside ResultSetLoader.

paul-rogers:
@sohami, the problem is simply that each value vector can stand alone: the `values` vector is, itself, a perfectly valid vector. However, the `MaterializedField` metadata for the `values` vector is inconsistent with the vector class. I considered this a bug. But, given the hassle with this PR, we may want to declare that inconsistency a feature.

In the result set loader code, I tried to use the vector metadata to recursively manage vectors: a required int vector (say) was to be handled the same whether it is a "top-level" or nested `values` vector. The result was nice, simple code.

I will instead revise the code use the class of the vector as accurate metadata, and not rely on the unreliable metadata.

PR #1370, opened 2018-07-10
DRILL-5797: Use Parquet new reader in all non-complex column queries

okalinin:
# DRILL-5797: use Parquet new reader more often
## Background
This PR is follow up on previous work done by @dprofeta and documented in the JIRA. Previously new reader was only used if file schema did not contain any single complex column. With this change, new reader will be used on a complex schema in case queried column list does not contain any complex one which should make new reader usage more frequent.

## Change description
In order to make usage of new reader possible on complex schema, following modifications had to be made:

- `ParquetReaderUtility` class - modified and added several functions to enable it working with nested schema. E.g. one limitation was explicitly referencing top level schema element path with `column.getPath()[0]` in several locations. Top level schema element path was also used in building path to `SchemaElement` map which caused map corruption for cases when schema contained columns `a` and `b.a` (for both schema elements key `a` was used overwriting the map entry).
- `ParquetSchema` - `fieldSelected()` function replaced with `columnSelected()` in order to enable it functioning with full paths. Previously, it would fail on cases when schema contains columns a and b.a as both schema paths would be classified as selected.
- `ParquetColumnMetadata` - replaced top level path reference with full path; also, replaced parameter passed to `ParquetToDrillTypeConverter.toMajorType()` from `se.getType_length()` to `column.getTypeLength()`. Reason behind is `se.getType_length()` returning 0 on FIXED_LEN_BYTE_ARRAY column and subsequent failure in minor type conversion that was failing complex parquet tests. `column.getTypeLength()` provides correct result. In fact, I am not sure if this is Parquet bug - possibly TBD item.
- `AbstractParquetScanBatchCreator` - added function which utilizes `ParquetReaderUtility` functions to identify whether query columns list contains complex columns and thus whether query qualifies for new reader.

## Level of testing
build tests and complex*q query tests from Drill test framework. Tests added for newly introduced methods except for `ParquetReaderUtility.buildFullColumnPath()`.


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -191,4 +201,42 @@ protected AbstractDrillFileSystemManager(OperatorContext operatorContext) {
     protected abstract DrillFileSystem get(Configuration config, String path) throws ExecutionSetupException;
   }
 
+  /**
+   * Check whether any of queried columns is nested or repetitive.
+   *
+   * @param footer  Parquet file schema
+   * @param columns list of query SchemaPath objects
+   */
+  public static boolean containsComplexColumn(ParquetMetadata footer, List<SchemaPath> columns) {
```
arina-ielchiieva: Why it's public and static?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -191,4 +201,42 @@ protected AbstractDrillFileSystemManager(OperatorContext operatorContext) {
...
+
+    Map<String, ColumnDescriptor> colDescMap = ParquetReaderUtility.getColNameToColumnDescriptorMapping(footer);
+    Map<String, SchemaElement> schemaElements = ParquetReaderUtility.getColNameToSchemaElementMapping(footer);
+
+    for (SchemaPath schemaPath : columns) {
+      // non-nested column check: full path must be equal to root segment path
+      if (!schemaPath.getUnIndexed().toString().replaceAll("`", "")
```
arina-ielchiieva: That's definitely not a good choice of comparing by replacing "`". Please find a different way to compare or write helper method that would compare without replacing, maybe traversing schema path. Schema path has plenty of handy methods, if needed is missing implement one. Basically you cannot rely on to string representation it can change, you should rely on object info instead.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -191,4 +201,42 @@ protected AbstractDrillFileSystemManager(OperatorContext operatorContext) {
...
+          .equalsIgnoreCase(schemaPath.getRootSegment().getPath())) {
+        logger.debug("Forcing 'old' reader due to nested column: {}", schemaPath.getUnIndexed().toString());
```
arina-ielchiieva: This method has nothing to do with forcing the reader, it's just tells if given schema path contain complex columns.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -161,20 +167,24 @@ private ParquetMetadata readFooter(Configuration conf, String path) throws IOExc
...
-      if (col.getMaxRepetitionLevel() > 0) {
-        return true;
+      for (ColumnDescriptor col : schema.getColumns()) {
+        if (col.getMaxRepetitionLevel() > 0) {
+          return true;
+        }
       }
+      return false;
+    } else {
+      return containsComplexColumn(footer, columns);
```
arina-ielchiieva: Basically `isComplex` and `containsComplexColumn` does the same, why would split it into semantically the same methods?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -191,4 +201,42 @@ protected AbstractDrillFileSystemManager(OperatorContext operatorContext) {
...
+      ColumnDescriptor column = colDescMap.get(schemaPath.getUnIndexed().toString().toLowerCase());
+
+      if (column == null) {
+        SchemaElement se = schemaElements.get(schemaPath.getUnIndexed().toString().toLowerCase());
+        if (se != null) {
+          return true;
+        }
+      } else {
+        if (column.getMaxRepetitionLevel() > 0) {
+          logger.debug("Forcing 'old' reader due to repetitive column: {}", schemaPath.getUnIndexed().toString());
```
arina-ielchiieva: The same as above.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -191,4 +201,42 @@ protected AbstractDrillFileSystemManager(OperatorContext operatorContext) {
...
+        return true;
+      }
+
+      // following column descriptor lookup failure may mean two cases, depending on subsequent SchemaElement lookup:
+      // 1. success: queried column is complex => use old reader
+      // 2. failure: queried column is not in schema => use new reader
+      ColumnDescriptor column = colDescMap.get(schemaPath.getUnIndexed().toString().toLowerCase());
+
+      if (column == null) {
+        SchemaElement se = schemaElements.get(schemaPath.getUnIndexed().toString().toLowerCase());
```
arina-ielchiieva: Not in favor of naming variables as `se`, does not any any meaning.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.ArrayList;
+
+import java.io.IOException;
+import java.nio.file.Paths;
+
+/**
+ * DRILL-5797 introduces more granularity on new reader use cases. This test is aimed at
+ * checking correctness of function used for new reader usage decision making.
+ */
+public class TestParquetReaderDecision {
```
arina-ielchiieva: Again method you are testing has nothing to do with reader decision, it's just tells if there are complex columns...

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
```diff
@@ -0,0 +1,99 @@
...
+  private static ParquetMetadata footer;
+
+  @BeforeClass
+  public static void setUpBeforeClass() throws Exception {
+    conf = new Configuration();
+
+    try {
+      footer = ParquetFileReader.readFooter(conf, new Path(path));
+    } catch (IOException ioe) {
+      fail("Could not read Parquet file '" + path + "', error message: " + ioe.getMessage()
```
arina-ielchiieva: Please use `String.format`. Also you just posing error message but not the full stack trace. Which might be useful. I would not catch the exception at all and just add it to method signature. When the test fails, it would be obvious what hapend.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
```diff
@@ -0,0 +1,99 @@
...
+          + " cwd: " + Paths.get(".").toAbsolutePath().normalize().toString());
+      throw(ioe);
+    }
+  }
+
+  @Test
+  public void testParquetReaderDecision() {
```
arina-ielchiieva: Please split into several unit tests, unit tests names should show what cases they are testing.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
```diff
@@ -0,0 +1,99 @@
...
+
+    assertTrue("Complex column not detected",
+        AbstractParquetScanBatchCreator.containsComplexColumn(footer, caseOldReader1));
+    assertTrue("Complex column not detected",
+        AbstractParquetScanBatchCreator.containsComplexColumn(footer, caseOldReader2));
+    assertTrue("Complex column not detected",
+        AbstractParquetScanBatchCreator.containsComplexColumn(footer, caseOldReader3));
+
+    assertFalse("No complex column, but complex column is detected",
+        AbstractParquetScanBatchCreator.containsComplexColumn(footer, caseNewReader1));
```
arina-ielchiieva: Though you have made method public static for testing, it's good practice, move method to utilities then. Semantically `AbstractParquetScanBatchCreator` definitely does not need to have this method as public static. As I have mentioned before `isComplex` and `containsComplexColumn` should be one method, might be utility one to ease testing.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,83 @@
...
+    } catch (IOException ioe) {
+      fail("Could not read Parquet file '" + path + "', error message: " + ioe.getMessage()
+          + " cwd: " + Paths.get(".").toAbsolutePath().normalize().toString());
+      throw(ioe);
+    }
+  }
+
+  @Test
+  public void testSchemaElementsMap() {
+    assertTrue(footer != null);
```
arina-ielchiieva: Why we check footer nullability in each test, can be done in `@BeforeClass` method.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,83 @@
...
+    Map<String, SchemaElement> schemaElements = ParquetReaderUtility.getColNameToSchemaElementMapping(footer);
+    assertTrue(schemaElements.size() == 14);
+
+    SchemaElement se = schemaElements.get("`marketing_info`.`camp_id`");
+    assertTrue(se != null);
+    assertTrue("camp_id".equals(se.getName()));
```
arina-ielchiieva: Use `assertEquals` here and below.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @Test
+  public void testSchemaElementsMap() {
+    assertTrue(footer != null);
+    Map<String, SchemaElement> schemaElements = ParquetReaderUtility.getColNameToSchemaElementMapping(footer);
+    assertTrue(schemaElements.size() == 14);
+
+    SchemaElement se = schemaElements.get("`marketing_info`.`camp_id`");
+    assertTrue(se != null);
```
arina-ielchiieva: `assertNotNull`

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.Map;
+
+public class TestParquetReaderUtility {
+
+  private static final String path = "src/test/resources/store/parquet/complex/complex.parquet";
+  private static Configuration conf;
```
arina-ielchiieva: Do you use `conf` elsewhere except of  `setUp`?

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,83 @@
...
+  private static ParquetMetadata footer;
+
+  @BeforeClass
+  public static void setUpBeforeClass() throws Exception {
```
arina-ielchiieva: I think annotation makes it obvious that this method is before class, no need to add this in method name.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,83 @@
...
+    conf = new Configuration();
+
+    try {
+      footer = ParquetFileReader.readFooter(conf, new Path(path));
+    } catch (IOException ioe) {
+      fail("Could not read Parquet file '" + path + "', error message: " + ioe.getMessage()
```
arina-ielchiieva: `String.format`

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
```diff
@@ -0,0 +1,99 @@
...
+import java.nio.file.Paths;
+
+/**
+ * DRILL-5797 introduces more granularity on new reader use cases. This test is aimed at
+ * checking correctness of function used for new reader usage decision making.
+ */
+public class TestParquetReaderDecision {
+
+  private static final String path = "src/test/resources/store/parquet/complex/complex.parquet";
+  private static Configuration conf;
```
arina-ielchiieva: You do need to store conf at class level.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -140,15 +140,87 @@ public static int getIntFromLEBytes(byte[] input, int start) {
...
-    HashMap<String, SchemaElement> schemaElements = new HashMap<>();
+    Map<String, SchemaElement> schemaElements = new HashMap<>();
     FileMetaData fileMetaData = new ParquetMetadataConverter().toParquetMetadata(ParquetFileWriter.CURRENT_VERSION, footer);
-    for (SchemaElement se : fileMetaData.getSchema()) {
-      schemaElements.put(se.getName(), se);
+
+    Iterator iter = fileMetaData.getSchema().iterator();
+
+    // skip first default 'root' element
+    if (iter.hasNext()) {
```
arina-ielchiieva: Why we need to skip root element?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -140,15 +140,87 @@ public static int getIntFromLEBytes(byte[] input, int start) {
...
+   * @return       schema full path to SchemaElement map
+   */
   public static Map<String, SchemaElement> getColNameToSchemaElementMapping(ParquetMetadata footer) {
-    HashMap<String, SchemaElement> schemaElements = new HashMap<>();
+    Map<String, SchemaElement> schemaElements = new HashMap<>();
     FileMetaData fileMetaData = new ParquetMetadataConverter().toParquetMetadata(ParquetFileWriter.CURRENT_VERSION, footer);
-    for (SchemaElement se : fileMetaData.getSchema()) {
-      schemaElements.put(se.getName(), se);
+
+    Iterator iter = fileMetaData.getSchema().iterator();
```
arina-ielchiieva: `Iterator<SchemaElement> iterator`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -140,15 +140,87 @@ public static int getIntFromLEBytes(byte[] input, int start) {
...
+  /**
+   * Populate full path to SchemaElement map by recursively traversing schema elements referenced by the given iterator
+   *
+   * @param iter file schema values iterator
+   * @param path parent schema element path
+   * @param schemaElements schema elements map to insert next iterator element into
+   */
+  private static void addSchemaElementMapping(Iterator iter, StringBuilder path,
+      Map<String, SchemaElement> schemaElements) {
+    SchemaElement se = (SchemaElement)iter.next();
```
arina-ielchiieva: No need for casting if you pass parametrized iterator...

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -140,15 +140,87 @@ public static int getIntFromLEBytes(byte[] input, int start) {
...
+    path.append('`').append(se.getName().toLowerCase()).append('`');
+    schemaElements.put(path.toString(), se);
+
+    int remainingChildren = se.getNum_children();
+
+    while (remainingChildren > 0 && iter.hasNext()) {
```
arina-ielchiieva: Why we need to count remaining children? Why `iterator.hasNext()` is not sufficient?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
okalinin: First element in `FileMetaData.getSchema()` is default element named `root`. If we don't skip it, map keys would have format:
`root`.`a`
`root`.`b`
etc.
Correspondingly, comparison with `SchemaPath.toString()` representation would need to be implemented with that in mind (`root` part would have to be cut anyway). If this is considered to be better approach, I could implement it that way indeed.

arina-ielchiieva: I think then you may skip the first element, but please add comment describing the reason.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
okalinin: `iterator.hasNext()` relates to whether there are more elements in flat tree representation and isin itself not enough to maintain correct recursion level. If we don't maintain/check remaining children count for parent element, schema
`a`.`b`
`a`.`c`
will result into keys in the map:
`a`
`a`.`b`
`a`.`b`.`c`
since function will erratically enter recursion where it should not.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
okalinin: Actually I took it from another test as I saw it in many other locations so considered maintaining the same naming convention. I'll correct it.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
okalinin: I based this one on original PR, but indeed it's a valid point. Will regroup accordingly.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
okalinin: I will fix this and all other test issues you highlighted. When fixing them I realised that existing `complex.parquet` test file possibly doesn't provide desired coverage for changes in this PR. Bulk part of changes are aimed at making utility functions work with schemas like 
`a`
`b`.`a`
while schema in `complex.parquet` doesn't contain such corner cases.
Does it make sense to add new resource to cover such cases and base tests on that file?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
okalinin: Replaced with `SchemaPath.isLeaf()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
okalinin: Fixed

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
arina-ielchiieva: Yes, just please generate small file to cover all needed cases.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
okalinin: Methods combined in one and moved to utility class.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
okalinin: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
okalinin: Fixed

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
okalinin: Class renamed to `TestComplexColumnInSchema`

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
okalinin: Removed exception handling as failure reason would indeed be obvious.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
okalinin: `containsComplexColumn` method moved to ParquetReaderUtility class.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
okalinin: Done

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
okalinin: Fixed

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
okalinin: Exception handling removed as per above comment.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
okalinin: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
okalinin: Fixed both

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
okalinin: Method moved to ParquetReaderUtility as per discussion below.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderDecision.java
okalinin: Done, schema is:
```
file schema:   root
--------------------------------------------------------------------------------
id:            OPTIONAL INT64 R:0 D:1
a:             OPTIONAL BINARY O:UTF8 R:0 D:1
repeated:      REPEATED INT64 R:1 D:1
VariableCase:  OPTIONAL BINARY O:UTF8 R:0 D:1
nested:        OPTIONAL F:3
.id:           OPTIONAL INT64 R:0 D:2
.repeated:     REPEATED INT64 R:1 D:2
.VaRiAbLeCaSe: OPTIONAL BINARY O:UTF8 R:0 D:2
```

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestComplexColumnInSchema.java
```diff
@@ -0,0 +1,141 @@
...
+import java.io.IOException;
+
+/**
+ * This test checks correctness of complex column detection in the Parquet file schema.
+ */
+public class TestComplexColumnInSchema {
+
+  private static final String path = "src/test/resources/store/parquet/complex/complex_special_cases.parquet";
+  private static ParquetMetadata footer;
+  List<SchemaPath> columns;
```
arina-ielchiieva: No need to have it as class variable. Let's each method have it's own columns variable.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private static final String path = "src/test/resources/store/parquet/complex/complex.parquet";
+  private static ParquetMetadata footer;
+
+  @BeforeClass
+  public static void setup() throws IOException {
+    Configuration conf = new Configuration();
+
+    footer = ParquetFileReader.readFooter(conf, new Path(path));
+    assertNotNull("footer must be not null", footer);
```
arina-ielchiieva: You are checking Footer for null, here but not in prev test.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,73 @@
...
+
+    schemaElement = schemaElements.get("`marketing_info`");
+    assertNotNull("Schema element must be not null", schemaElement);
+    assertEquals("Schema element name match lookup key", schemaElement.getName(), "marketing_info");
+  }
+
+  @Test
+  public void testColumnDescriptorMap() {
+    Map<String, ColumnDescriptor> colDescMap = ParquetReaderUtility.getColNameToColumnDescriptorMapping(footer);
+    assertTrue(colDescMap.size() == 11);
```
arina-ielchiieva: assertEquals...

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetReaderUtility.java
```diff
@@ -0,0 +1,73 @@
...
+    Configuration conf = new Configuration();
+
+    footer = ParquetFileReader.readFooter(conf, new Path(path));
+    assertNotNull("footer must be not null", footer);
+  }
+
+  @Test
+  public void testSchemaElementsMap() {
+    Map<String, SchemaElement> schemaElements = ParquetReaderUtility.getColNameToSchemaElementMapping(footer);
+    assertTrue(schemaElements.size() == 14);
```
arina-ielchiieva: assertEquals...

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestComplexColumnInSchema.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.ArrayList;
+
+import java.io.IOException;
+
+/**
+ * This test checks correctness of complex column detection in the Parquet file schema.
+ */
+public class TestComplexColumnInSchema {
+
+  private static final String path = "src/test/resources/store/parquet/complex/complex_special_cases.parquet";
```
arina-ielchiieva: You can add description of the schema and data in the comment.
Example: https://github.com/apache/drill/blob/9a6cb59b9b7a5b127e5f60309ce2f506ede9652a/exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetFilterPushDownForComplexTypes.java

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -140,15 +142,88 @@ public static int getIntFromLEBytes(byte[] input, int start) {
...
+   * @param path parent schema element path
+   * @param schemaElements schema elements map to insert next iterator element into
+   */
+  private static void addSchemaElementMapping(Iterator<SchemaElement> iter, StringBuilder path,
+      Map<String, SchemaElement> schemaElements) {
+    SchemaElement schemaElement = iter.next();
+    path.append('`').append(schemaElement.getName().toLowerCase()).append('`');
+    schemaElements.put(path.toString(), schemaElement);
+
+    int remainingChildren = schemaElement.getNum_children();
```
arina-ielchiieva: Please add comment why remaining children are important in while loop.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -107,7 +105,9 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
           ParquetReaderUtility.detectCorruptDates(footer, rowGroupScan.getColumns(), autoCorrectCorruptDates);
         logger.debug("Contains corrupt dates: {}", containsCorruptDates);
 
-        if (!context.getOptions().getBoolean(ExecConstants.PARQUET_NEW_RECORD_READER) && !isComplex(footer)) {
+        if (!context.getOptions().getBoolean(ExecConstants.PARQUET_NEW_RECORD_READER)
+            && !ParquetReaderUtility.containsComplexColumn(footer, rowGroupScan.getColumns())) {
+          logger.debug("Query qualifies for new Parquet reader");
```
arina-ielchiieva: Which query? We might add query id or something to help identification in the log.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -525,4 +599,57 @@ public static long getDateTimeValueFromBinary(Binary binaryTimeStampValue, boole
...
+        ColumnDescriptor column = colDescMap.get(schemaPath.getUnIndexed().toString().toLowerCase());
+
+        if (column == null) {
+          SchemaElement schemaElement = schemaElements.get(schemaPath.getUnIndexed().toString().toLowerCase());
+          if (schemaElement != null) {
+            return true;
+          }
+        } else {
+          if (column.getMaxRepetitionLevel() > 0) {
+            logger.debug("rowGroupScan contains repetitive column: {}", schemaPath.getUnIndexed().toString());
```
arina-ielchiieva: trace?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -525,4 +599,57 @@ public static long getDateTimeValueFromBinary(Binary binaryTimeStampValue, boole
...
+      }
+      return false;
+    } else {
+      Map<String, ColumnDescriptor> colDescMap = ParquetReaderUtility.getColNameToColumnDescriptorMapping(footer);
+      Map<String, SchemaElement> schemaElements = ParquetReaderUtility.getColNameToSchemaElementMapping(footer);
+
+      for (SchemaPath schemaPath : columns) {
+        // Schema path which is non-leaf is complex column
+        if (!schemaPath.isLeaf()) {
+          logger.debug("rowGroupScan contains complex column: {}", schemaPath.getUnIndexed().toString());
```
arina-ielchiieva: trace?

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestComplexColumnInSchema.java
okalinin: This and all other outstanding comments have been addressed.

PR #1371, opened 2018-07-10
DRILL-6588: Make Sys tables of nullable datatypes

kkhatua:
This addresses the issue of columns in the System tables being marked as non-nullable. While these tables are immutable, they can carry nullable values as well. (e.g. `num_val` in `sys.options`)
This commit introduces an annotation for the `PojoReader` that applies the nullable property if explicitly defined for a column (i.e. explicitly defined `isNullable` for a member of the POJO instance).
```
apache drill 1.14.0-SNAPSHOT 
"just drill it"
0: jdbc:drill:schema=sys> select count(*) from sys.options where num_val is null;
+---------+
| EXPR$0  |
+---------+
| 108     |
+---------+
1 row selected (2.703 seconds)
0: jdbc:drill:schema=sys> select count(*) from sys.options where isnull(num_val);
+---------+
| EXPR$0  |
+---------+
| 108     |
+---------+
1 row selected (0.3 seconds)
0: jdbc:drill:schema=sys> select distinct is_nullable, count(*) from INFORMATION_SCHEMA.`COLUMNS` where table_schema = 'sys' group by is_nullable;
+--------------+---------+
| is_nullable  | EXPR$1  |
+--------------+---------+
| NO           | 36      |
| YES          | 50      |
+--------------+---------+
2 rows selected (0.69 seconds)
0: jdbc:drill:schema=sys> describe options;
+-------------------+--------------------+--------------+
|    COLUMN_NAME    |     DATA_TYPE      | IS_NULLABLE  |
+-------------------+--------------------+--------------+
| name              | CHARACTER VARYING  | NO           |
| kind              | CHARACTER VARYING  | NO           |
| accessibleScopes  | CHARACTER VARYING  | NO           |
| optionScope       | CHARACTER VARYING  | NO           |
| status            | CHARACTER VARYING  | NO           |
| num_val           | BIGINT             | YES          |
| string_val        | CHARACTER VARYING  | YES          |
| bool_val          | BOOLEAN            | YES          |
| float_val         | DOUBLE             | YES          |
+-------------------+--------------------+--------------+
9 rows selected (0.221 seconds)
0: jdbc:drill:schema=sys> describe internal_options;
+-------------------+--------------------+--------------+
|    COLUMN_NAME    |     DATA_TYPE      | IS_NULLABLE  |
+-------------------+--------------------+--------------+
| name              | CHARACTER VARYING  | NO           |
| kind              | CHARACTER VARYING  | NO           |
| accessibleScopes  | CHARACTER VARYING  | NO           |
| optionScope       | CHARACTER VARYING  | NO           |
| status            | CHARACTER VARYING  | NO           |
| num_val           | BIGINT             | YES          |
| string_val        | CHARACTER VARYING  | YES          |
| bool_val          | BOOLEAN            | YES          |
| float_val         | DOUBLE             | YES          |
+-------------------+--------------------+--------------+
9 rows selected (0.185 seconds)
0: jdbc:drill:schema=sys> describe options_val;
+-------------------+--------------------+--------------+
|    COLUMN_NAME    |     DATA_TYPE      | IS_NULLABLE  |
+-------------------+--------------------+--------------+
| name              | CHARACTER VARYING  | NO           |
| kind              | CHARACTER VARYING  | NO           |
| accessibleScopes  | CHARACTER VARYING  | NO           |
| val               | CHARACTER VARYING  | YES          |
| optionScope       | CHARACTER VARYING  | NO           |
+-------------------+--------------------+--------------+
5 rows selected (0.183 seconds)
0: jdbc:drill:schema=sys> describe profiles_json;
+--------------+--------------------+--------------+
| COLUMN_NAME  |     DATA_TYPE      | IS_NULLABLE  |
+--------------+--------------------+--------------+
| queryId      | CHARACTER VARYING  | NO           |
| json         | CHARACTER VARYING  | YES          |
+--------------+--------------------+--------------+
2 rows selected (0.727 seconds)
0: jdbc:drill:schema=sys> describe profiles;
+--------------+--------------------+--------------+
| COLUMN_NAME  |     DATA_TYPE      | IS_NULLABLE  |
+--------------+--------------------+--------------+
| queryId      | CHARACTER VARYING  | NO           |
| startTime    | TIMESTAMP          | YES          |
| foreman      | CHARACTER VARYING  | NO           |
| fragments    | BIGINT             | YES          |
| user         | CHARACTER VARYING  | YES          |
| queue        | CHARACTER VARYING  | YES          |
| planTime     | BIGINT             | YES          |
| queueTime    | BIGINT             | YES          |
| executeTime  | BIGINT             | YES          |
| totalTime    | BIGINT             | YES          |
| state        | CHARACTER VARYING  | YES          |
| query        | CHARACTER VARYING  | YES          |
+--------------+--------------------+--------------+
12 rows selected (0.309 seconds)
```

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
@@ -17,12 +17,13 @@
  */
 package org.apache.drill.exec.store;
 
-import com.google.common.collect.Lists;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayList;
+import java.util.List;
+
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.sql.type.SqlTypeName;
-
-import java.util.List;
 
 /**
  * RecordDataType defines names and data types of columns in a static drill table.
@@ -30,9 +31,9 @@ import java.util.List;
 public abstract class RecordDataType {
 
   /**
-   * @return the {@link org.apache.calcite.sql.type.SqlTypeName} of columns in the table
+   * @return the {@link org.apache.calcite.sql.type.SqlTypeName} of columns in the table as a pair with its nullability
    */
-  public abstract List<SqlTypeName> getFieldSqlTypeNames();
+  public abstract List<SimpleImmutableEntry<SqlTypeName, Boolean>> getFieldSqlTypeNames();
 
   /**
    * @return the column names in the table
@@ -47,17 +48,21 @@ public abstract class RecordDataType {
    * @return the constructed type
    */
   public final RelDataType getRowType(RelDataTypeFactory factory) {
-    final List<SqlTypeName> types = getFieldSqlTypeNames();
+    final List<SimpleImmutableEntry<SqlTypeName, Boolean>> types = getFieldSqlTypeNames();
     final List<String> names = getFieldNames();
-    final List<RelDataType> fields = Lists.newArrayList();
-    for (final SqlTypeName typeName : types) {
+    final List<RelDataType> fields = new ArrayList<>();
+    for (SimpleImmutableEntry<SqlTypeName, Boolean> sqlTypePair : types) {
+      final SqlTypeName typeName = sqlTypePair.getKey();
+      final RelDataType tempDataType;
       switch (typeName) {
         case VARCHAR:
-          fields.add(factory.createSqlType(typeName, Integer.MAX_VALUE));
+          tempDataType = factory.createSqlType(typeName, Integer.MAX_VALUE);
           break;
         default:
-          fields.add(factory.createSqlType(typeName));
+          tempDataType = factory.createSqlType(typeName);
       }
+      //Add [Non]Nullable RelDataType
+      fields.add(factory.createTypeWithNullability(tempDataType, sqlTypePair.getValue()));
     }
     return factory.createStructType(fields, names);
   }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
@@ -21,11 +21,11 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.math.BigDecimal;
 import java.sql.Timestamp;
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.calcite.sql.type.SqlTypeName;
-
-import com.google.common.collect.Lists;
 import org.apache.drill.exec.store.RecordDataType;
 
 /**
@@ -34,8 +34,8 @@ import org.apache.drill.exec.store.RecordDataType;
 public class PojoDataType extends RecordDataType {
 //  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PojoDataType.class);
 
-  private final List<SqlTypeName> types = Lists.newArrayList();
-  private final List<String> names = Lists.newArrayList();
+  private final List<SimpleImmutableEntry<SqlTypeName, Boolean>> types = new ArrayList<>();
+  private final List<String> names = new ArrayList<>();
   private final Class<?> pojoClass;
 
   public PojoDataType(Class<?> pojoClass) {
@@ -48,22 +48,25 @@ public class PojoDataType extends RecordDataType {
       Class<?> type = f.getType();
       names.add(f.getName());
 
+      //Absence of annotation @NonNullable => (isNullable=true)
+      final boolean isNullable = !(f.isAnnotationPresent(NonNullable.class));
+
       if (type == int.class || type == Integer.class) {
-        types.add(SqlTypeName.INTEGER);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.INTEGER, isNullable));
       } else if(type == boolean.class || type == Boolean.class) {
-        types.add(SqlTypeName.BOOLEAN);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.BOOLEAN, isNullable));
       } else if(type == long.class || type == Long.class) {
-        types.add(SqlTypeName.BIGINT);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.BIGINT, isNullable));
       } else if(type == double.class || type == Double.class) {
-        types.add(SqlTypeName.DOUBLE);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.DOUBLE, isNullable));
       } else if(type == BigDecimal.class) {
-        types.add(SqlTypeName.DECIMAL);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.DECIMAL, isNullable));
       } else if(type == String.class) {
-        types.add(SqlTypeName.VARCHAR);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.VARCHAR, isNullable));
       } else if(type.isEnum()) {
-        types.add(SqlTypeName.VARCHAR);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.VARCHAR, isNullable));
       } else if (type == Timestamp.class) {
-        types.add(SqlTypeName.TIMESTAMP);
+        types.add(new SimpleImmutableEntry<SqlTypeName, Boolean>(SqlTypeName.TIMESTAMP, isNullable));
       } else {
         throw new RuntimeException(String.format("PojoDataType doesn't yet support conversions from type [%s].", type));
       }
@@ -75,7 +78,7 @@ public class PojoDataType extends RecordDataType {
   }
 
   @Override
-  public List<SqlTypeName> getFieldSqlTypeNames() {
+  public List<SimpleImmutableEntry<SqlTypeName, Boolean>> getFieldSqlTypeNames() {
     return types;
   }
 
@@ -83,5 +86,4 @@ public class PojoDataType extends RecordDataType {
   public List<String> getFieldNames() {
     return names;
   }
-
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
@@ -21,8 +21,8 @@ import java.sql.Timestamp;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Set;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.TimeZone;
 
 import org.apache.drill.exec.ExecConstants;
@@ -32,6 +32,7 @@ import org.apache.drill.exec.rpc.user.UserServer.BitToUserConnectionConfig;
 import org.apache.drill.exec.rpc.user.UserSession;
 import org.apache.drill.exec.server.options.OptionManager;
 import org.apache.drill.exec.server.rest.profile.SimpleDurationFormat;
+import org.apache.drill.exec.store.pojo.NonNullable;
 import org.apache.drill.exec.util.ImpersonationUtil;
 import org.joda.time.DateTime;
 
@@ -99,14 +100,18 @@ public class BitToUserConnectionIterator implements Iterator<Object> {
 
   public static class ConnectionInfo {
     public String user;
+    @NonNullable
     public String client;
+    @NonNullable
     public String drillbit;
+    @NonNullable
     public Timestamp established;
     public String duration;
     public int queries;
     public boolean isAuthenticated;
     public boolean isEncrypted;
     public boolean usingSSL;
+    @NonNullable
     public String session;
 
     public ConnectionInfo(Entry<BitToUserConnection, BitToUserConnectionConfig> connectionConfigPair, String hostname) {
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/store/sys/TestSystemTable.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/store/sys/TestSystemTable.java
@@ -17,6 +17,8 @@
  */
 package org.apache.drill.exec.store.sys;
 
+import static org.junit.Assert.assertEquals;
+
 import org.apache.drill.PlanTestBase;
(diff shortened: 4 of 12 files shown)


kkhatua:
@amansinha100 can you review this?

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
```diff
@@ -49,15 +55,25 @@
   public final RelDataType getRowType(RelDataTypeFactory factory) {
     final List<SqlTypeName> types = getFieldSqlTypeNames();
     final List<String> names = getFieldNames();
+    final List<Boolean> nullables = getFieldNullability();
     final List<RelDataType> fields = Lists.newArrayList();
-    for (final SqlTypeName typeName : types) {
+    Iterator<SqlTypeName> typesIter = types.listIterator();
+    Iterator<Boolean> nullabilityIter = nullables.listIterator();
```
paul-rogers: We now four correlated lists. This is often difficult to reason about and maintain.

Is it possible to have a single list of, say, `FieldDefn` objects, each of which has a name, type, nullable and `RelDataType`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/Nullability.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates Nullability
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
+public @interface Nullability {
```
paul-rogers: Turns out that Java has a existing `@Nonnull` and `@Nullable` annotations in `java.annotation`. Can we use those?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
```diff
@@ -36,6 +36,7 @@
 
   private final List<SqlTypeName> types = Lists.newArrayList();
   private final List<String> names = Lists.newArrayList();
+  private final List<Boolean> nullables = Lists.newArrayList();
```
paul-rogers: Again, single list of objects rather than three correlated list of primitives?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
```diff
@@ -48,6 +49,11 @@ public PojoDataType(Class<?> pojoClass) {
       Class<?> type = f.getType();
       names.add(f.getName());
 
+      //Look up @Nullability for nullable property
+      Nullability nullability = f.getDeclaredAnnotation(Nullability.class);
+      nullables.add(nullability == null ? //Absence of annotation => (isNullable=true)
```
paul-rogers: `f.isAnnotationPresent(Nullability.class)` returns a `boolean` result; may simplify the code here

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
```diff
@@ -84,4 +90,8 @@ public PojoDataType(Class<?> pojoClass) {
     return names;
   }
 
+  @Override
+  public List<Boolean> getFieldNullability() {
```
paul-rogers: Same comment about the awkwardness of correlated lists...

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
kkhatua: I wanted to avoid changing existing implementation, so I added an extra list and iterated in parallel.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
kkhatua: It seemed very verbose, but that makes sense. I'll make this change.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/Nullability.java
kkhatua: That would help. Let me take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
kkhatua: RecordDataType uses the `names` and `types` lists separately. So, I'll need atleast 2 lists.

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
paul-rogers: This is, in fact, the gist of the comment. While it is understandable to want to leave he existing code as-is, adding more lists to the existing parallel lists is not an improvement: it has caused a semi-wrong initial implementation to become overly cumbersome.

If the implementation is local to this class or module, then leaving the code cleaner after your change than when you found it is generally a good thing.

kkhatua:
@paul-rogers I've made the changes to use Java's Annotation class `Nonnull`
Please review the rest of the changes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
```diff
@@ -47,17 +48,22 @@
    * @return the constructed type
    */
   public final RelDataType getRowType(RelDataTypeFactory factory) {
-    final List<SqlTypeName> types = getFieldSqlTypeNames();
+    final List<ImmutablePair<SqlTypeName, Boolean>> types = getFieldSqlTypeNames();
```
paul-rogers: Better, but we still have three correlated lists. Can you combine the four fields into a single struct (class) and have a single list?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
```diff
@@ -99,14 +102,18 @@ public void remove() {
 
   public static class ConnectionInfo {
     public String user;
+    @Nonnull
```
paul-rogers: Nice use of the Java native annotation!

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
kkhatua: There are just 2 lists : `names` and `types`. 
The third list (`fields`) is used as `factory.createStructType(fields, names)`
I'm reluctant to change abstract methods because they might have usage beyond the current SystemTable context.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
kkhatua: Thanks for pointing out the availability of these annotations. Didn't want to reinvent the wheel. :)

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
arina-ielchiieva: I also agree with Paul, using Pair is not a good choice. Better to introduce holder class.

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
```diff
@@ -47,17 +48,22 @@
...
     final List<String> names = getFieldNames();
     final List<RelDataType> fields = Lists.newArrayList();
-    for (final SqlTypeName typeName : types) {
+
+    for (ImmutablePair<SqlTypeName, Boolean> sqlTypePair : types) {
+      final SqlTypeName typeName = sqlTypePair.getLeft();
+      final RelDataType tmpRDT;
```
arina-ielchiieva: tmpRTD -> tmpDataType

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
```diff
@@ -48,22 +51,25 @@ public PojoDataType(Class<?> pojoClass) {
       Class<?> type = f.getType();
       names.add(f.getName());
 
+      //Absence of annotation @Nonnull => (isNullable=true)
+      final boolean isNullable = f.isAnnotationPresent(Nonnull.class) ? false : true;
```
arina-ielchiieva: This can be simplified to `! f.isAnnotationPresent(Nonnull.class)`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
arina-ielchiieva: The thing is that this is not native java annotation. It's absent in java 8 SE specification.
https://docs.oracle.com/javase/8/docs/api/javax/annotation/package-summary.html
http://mvnrepository.com/artifact/javax.annotation/javax.annotation-api/1.3.2

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
kkhatua: Agreed. Will do this change

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
kkhatua: Will do

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
kkhatua: That is odd. I actually was referring to this: https://blogs.oracle.com/java-platform-group/java-8s-new-type-annotations
I suspect the documentation is missing, because the `javax.annotation.Nonnull` is meant for runtime code generation
```javax.annotation.Nonnull

@TypeQualifier
@Retention(value=RUNTIME)
@Documented
```

kkhatua: Ok. Got it. This is part of the annotations-2.0.1 dependency that drill has in its 3rd party library (`com.google.code.findbugs`). It might be used here in a context different from what it was intended to be used for.
So, @arina-ielchiieva / @paul-rogers , should I go ahead with this usage or revert to implementing the annotation class I originally had?

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
kkhatua: For a one-time usage of creating a nullable RelDataType (since Calcite doesn't provide a convenient way of doing so), I think a holder class is an overkill. 
I have this option from Calcite: 
`https://calcite.apache.org/apidocs/org/apache/calcite/rel/type/RelDataTypeFactory.html#createStructType-java.util.List-`
which also makes use of a list of Map.Entry objects. 
So, unless I change Calcite, there is no real value in creating a holder class.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
paul-rogers: Sorry, I was misled by the "javax" package and thought this was a Java thing. Good catch @arina-ielchiieva.

Since it is not, go ahead and define your own. Today we assume all fields are nullable, and must keep that default. So, since the default is nullable, you only need the `@NonNullable` annotation.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
```diff
@@ -75,13 +81,12 @@ public PojoDataType(Class<?> pojoClass) {
   }
 
   @Override
-  public List<SqlTypeName> getFieldSqlTypeNames() {
+  public List<ImmutablePair<SqlTypeName,Boolean>> getFieldSqlTypeNames() {
```
arina-ielchiieva: `SqlTypeName,Boolean` -> space

exec/java-exec/src/main/java/org/apache/drill/exec/store/pojo/PojoDataType.java
```diff
@@ -34,7 +37,7 @@
 public class PojoDataType extends RecordDataType {
 //  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PojoDataType.class);
 
-  private final List<SqlTypeName> types = Lists.newArrayList();
+  private final List<ImmutablePair<SqlTypeName, Boolean>> types = Lists.newArrayList();
```
arina-ielchiieva: Please don't use guava, use `new ArrayList<>()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/BitToUserConnectionIterator.java
arina-ielchiieva: +1 for @NotNullable

exec/java-exec/src/main/java/org/apache/drill/exec/store/RecordDataType.java
arina-ielchiieva: Ok, don't have strong opinion about this right now, so you may leave as is. But I would prefer Java native implementation in this case:
```
      AbstractMap.SimpleImmutableEntry<Integer, String> entry
          = new AbstractMap.SimpleImmutableEntry<>(1, "one");
```

kkhatua:
@arina-ielchiieva / @paul-rogers  Done the final changes.

arina-ielchiieva:
@kkhatua could you please add unit tests, similar to the queries you showed in description.

kkhatua:
Added a unit test that groups and counts the number of IS_NULLABLE values. Originally, the query would have returned only 1 group (`IS_NULLABLE = false`). After the patch, there are 2 records for both boolean values.

exec/java-exec/src/test/java/org/apache/drill/exec/store/sys/TestSystemTable.java
```diff
@@ -90,4 +92,11 @@ public void testProfilesLimitPushDown() throws Exception {
     String numFilesPattern = "maxRecordsToRead=10";
     testPlanMatchingPatterns(query, new String[] {numFilesPattern}, new String[] {});
   }
+
+  @Test
+  public void testColumnNullability() throws Exception {
+    String query = " select distinct is_nullable, count(*) from INFORMATION_SCHEMA.`COLUMNS` where table_schema = 'sys' group by is_nullable";
```
arina-ielchiieva: `" select` -> please remove space

arina-ielchiieva:
Once minor comment to remove space, also please squash the commits. +1, LGTM.

PR #1373, opened 2018-07-10
DRILL-6517: Hash-Join: handling uninitialized vector container

Ben-Zvi:
When a running query is cancelled (e.g., when Disk is full while spilling), many next() calls return a STOP outcome. Any Hash-Join that is "sniffing" its incoming batches (by repeatedly calling next()) would proceed to update the memory manager regardless of the outcomes returned.
   In some abnormal outcomes (like STOP), the batch's container is **not initialized**. The update call needs the batch's row count, which for *RemovingRecordBatch* is implemented (differently) by invoking the container's getRowCount() -- so in this case the call fails.
  The Fix: Move all the checks for the abnormal conditions immediately after the "sniffing", thus returning early and avoiding the update of the memory manager.
  Longer term question: How to implement getRowCount() for the record batch consistently.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
@@ -82,6 +82,7 @@ import org.apache.drill.exec.work.filter.RuntimeFilterReporter;
 
 
 import static org.apache.drill.exec.record.RecordBatch.IterOutcome.EMIT;
+import static org.apache.drill.exec.record.RecordBatch.IterOutcome.NONE;
 import static org.apache.drill.exec.record.RecordBatch.IterOutcome.OK_NEW_SCHEMA;
 
 /**
@@ -449,7 +450,8 @@ public class HashJoinBatch extends AbstractBinaryRecordBatch<HashJoinPOP> {
 
         prefetchFirstProbeBatch();
 
-        if (leftUpstream.isError()) {
+        if (leftUpstream.isError() ||
+            ( leftUpstream == NONE && joinType != JoinRelType.FULL && joinType != JoinRelType.RIGHT )) {
           // A termination condition was reached while prefetching the first probe side data holding batch.
           // We need to terminate.
           return leftUpstream;
@@ -496,9 +498,7 @@ public class HashJoinBatch extends AbstractBinaryRecordBatch<HashJoinPOP> {
            * Either case build the output container's schema and return
            */
           if (outputRecords > 0 || state == BatchState.FIRST) {
-            if (state == BatchState.FIRST) {
-              state = BatchState.NOT_FIRST;
-            }
+            state = BatchState.NOT_FIRST;
 
             return IterOutcome.OK;
           }
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinOutcome.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinOutcome.java
@@ -1,0 +1,212 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.join;
+
+import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.drill.categories.OperatorTest;
+import org.apache.drill.common.expression.FieldReference;
+import org.apache.drill.common.logical.data.JoinCondition;
+import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.physical.base.PhysicalOperator;
+import org.apache.drill.exec.physical.config.HashJoinPOP;
+import org.apache.drill.exec.physical.impl.MockRecordBatch;
+import org.apache.drill.test.PhysicalOpUnitTestBase;
+import org.apache.drill.exec.record.BatchSchema;
+import org.apache.drill.exec.record.RecordBatch;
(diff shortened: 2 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -258,6 +251,13 @@ protected boolean prefetchFirstBatchFromBothSides() {
       return false;
```
sohami: `checkForEarlyFinish` checks for NONE from both side. What if only one side receives NONE? In that case also `BatchMemoryManager.update()` should not be called

Ben-Zvi: The case of only one side receiving NONE is legit -- for those RIGHT/LEFT/FULL OUTER joins, where the other side is still returned.
As for the update() on a NONE -- probably any setting of NONE also initializes the container's record count to zero.

ppadma:
+1. LGTM. Thanks for fixing this.

Ben-Zvi:
Thanks @ppadma ; lets have @ilooner look at this too since he made some of the relevant changes (e.g., *getRecordCount()* for *RemovingRecordBatch*).
The initial thesis was that this bug was triggering DRILL-6453 (query fails after 2h11m). Unfortunately now with this fix it seems the other way around (that failure caused those STOP outcomes).

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
sohami: AFAIK NONE is not supposed to have a valid incoming container associated with it

ilooner: @Ben-Zvi I agree the case where only one side receives NONE is legit. I think the issue @sohami was referring to is that **batchMemoryManager.update** will be called for both sides even when only one side has NONE. Looking at the javadoc for IterOutcome.NONE there is no guarantee that we will have a container with a value record count set on it. Also looking at other operators such as project, they do not seem to set the record count in the case where they process an empty table, so we cannot assume that a container will have a record count of 0 when receiving NONE.

Could you add a check to call update and debug statements for the left and right side only if their IterOutcome is not NONE?

Ben-Zvi:
An alternative is to set the record count to zero on the vector container when the outcome is not OK or OK_NEW_SCHEMA. (Instead of checking each time on update() )

ppadma:
@Ben-Zvi How about we move the update to sniffNonEmptyBatch under the case OK when recordCount is not zero ?

Ben-Zvi:
@ilooner please review. Latest changes: Applied suggestion from @ppadma (call *update()* inside *sniffNonEmptyBatch()* ) . And added tests that send records with uninitialized Vector Container to either side (for either STOP or NONE outcomes).

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinOutcome.java
```diff
@@ -0,0 +1,204 @@
...
+
+    conditions.add(new JoinCondition( SqlKind.EQUALS.toString(),
+      FieldReference.getWithQuotedRef("leftcol"),
+      FieldReference.getWithQuotedRef("rightcol")));
+
+    HashJoinPOP hjConf = new HashJoinPOP(null, null, conditions, JoinRelType.INNER);
+
+    HashJoinBatch hjBatch = new HashJoinBatch(hjConf,operatorFixture.getFragmentContext(), mockInputBatchLeft, mockInputBatchRight );
+
+    assertTrue(hjBatch.next() == returnedOutcome ); // verify returned outcome
```
ppadma: may be rename this to expectedOutcome instead of returnedOutcome ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -289,7 +283,13 @@ private IterOutcome sniffNonEmptyBatch(int inputIndex, RecordBatch recordBatch)
           if (recordBatch.getRecordCount() == 0) {
             continue;
           }
-          // We got a non empty batch
+          // We got a non empty batch; update the memory manager
+          final boolean isBuildSide = inputIndex == 1;
+          final int side = isBuildSide ? RIGHT_INDEX : LEFT_INDEX;
+          // For the build side, use aggregate i.e. average row width across batches
+          batchMemoryManager.update(side, 0, isBuildSide);
```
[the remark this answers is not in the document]
Ben-Zvi: The "? true : false" part is redundant :-)

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinOutcome.java
Ben-Zvi: Good idea. Also the new test file name was conceived in a rush, in case there's a better one. This is the first set of unit tests for the Hash-Join operator (vs. executing a whole join query). Maybe we could add more HJ unit tests in this file later.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -289,7 +283,13 @@ private IterOutcome sniffNonEmptyBatch(int inputIndex, RecordBatch recordBatch)
           if (recordBatch.getRecordCount() == 0) {
             continue;
           }
-          // We got a non empty batch
+          // We got a non empty batch; update the memory manager
+          final boolean isBuildSide = inputIndex == 1;
+          final int side = isBuildSide ? RIGHT_INDEX : LEFT_INDEX;
```
ilooner: isn't 0 / 1 and LEFT_INDEX / RIGHT_INDEX the same thing? Similarly isn't **side** the same thing as **inputIndex**? Could we make things consistent to avoid confusion? Or if there is a good reason for using different names for things could you add a comment explaining the differences.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/join/TestHashJoinOutcome.java
```diff
@@ -0,0 +1,204 @@
...
+// import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+/**
+ *  Unit tests of the Hash Join getting various outcomes as input
+ *  with uninitialized vector containers
+ */
+@Category(OperatorTest.class)
+public class TestHashJoinOutcome extends PhysicalOpUnitTestBase {
+
```
ilooner: Thanks for adding tests.

Ben-Zvi:
@ilooner please review -- This PR was changed, reduced from its original scope as other work (e.g. DRILL-6453) handled the main issue (a STOP outcome with uninitialized container).
   The main addition is the test file, checking that incoming uninitialized input containers (for STOP or NONE outcomes) are handled correctly. (Basically the same test as before, only now calling `next()` twice (as the first call always returns OK_NEW_SCHEMA)). 
   A small fix was needed for the hash join, as in the case the first real incoming batch was NONE, the second `next()` was returning an OK (with 0 rows). With the change it returns NONE.
  And a minor change in the MockRecordBatch to avoid getting record count on uninitialized containers.

PR #1374, opened 2018-07-10
DRILL-6542 : IndexOutOfBoundsException for multilevel lateral queries…

sohami:
… with schema changed partitioned complex data

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
@@ -25,6 +25,7 @@ import org.apache.drill.exec.physical.base.LateralContract;
 import org.apache.drill.exec.record.BatchSchema.SelectionVectorMode;
 import org.apache.drill.exec.record.RecordBatch;
 import org.apache.drill.exec.record.TransferPair;
+import org.apache.drill.exec.vector.SchemaChangeCallBack;
 import org.apache.drill.exec.vector.ValueVector;
 import org.apache.drill.exec.vector.complex.RepeatedValueVector;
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
```diff
@@ -97,8 +99,16 @@ public final int unnestRecords(final int recordCount) {
...
+    final SchemaChangeCallBack callBack = new SchemaChangeCallBack();
     for (TransferPair t : transfers) {
       t.splitAndTransfer(innerValueIndex, count);
+
+      // Get the corresponding ValueVector in output container and transfer the data
+      final ValueVector vectorWithData = t.getTo();
+      final ValueVector outputVector = outgoing.getContainer().addOrGet(vectorWithData.getField(), callBack);
+      Preconditions.checkState(!callBack.getSchemaChangedAndReset(), "Outgoing container doesn't have " +
+        "expected ValueVector of type %s, present in TransferPair of unnest field", vectorWithData.getClass());
+      vectorWithData.makeTransferPair(outputVector).transfer();
```
parthchandra: So we do two transfers? Cheaper to transfer directly to output vector.  We can create a new transfer pair with the source vector in the transfer pair we set up.

sohami: Yes we have to do 2 transfers since there is no api for getTransferPairToSingleMap to take target vector. For New Schema there are 2 cases, in one we have to preserve the output container vector (when schema change is for non-unnest column) but reset transfers whereas in other case we have to create a new vector in the container (when schema change is on unnest column). That's the reason it is done this way.

PR #1376, opened 2018-07-12
DRILL-6592: Unnest record batch size is called too frequently

parthchandra:
@sohami  please review


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestRecordBatch.java
```diff
@@ -216,7 +225,7 @@ public IterOutcome innerNext() {
         boolean isNewSchema = schemaChanged();
```
sohami: I think `isNewSchema` flag can be deprecated and replaced by `hasNewSchema`.

PR #1379, opened 2018-07-13
DRILL-6591: Show Exception for failed queries submitted in WebUI

kkhatua:
When query fails on Web UI result page no error is shown, only "No result found." 
This was because DRILL-6477 (PR #1309) switched to `WebUserConnection.await(long timeoutInMillis)` . Unlike the original `WebUserConnection.await()`, this method did not throw any UserException generated by a query failure.
The fix was to add a new WebUser-only method - `WebUserConnection.timedWait(long timeoutInMillis)` 
This ensures that other callers to the `WebUserConnection.await(timeoutInMillis)` are unaffected by the expected behaviour.

This is what the UI returns after this patch:
_When running query which hits memory limits (DRILL-6477), we get this:_
```
org.apache.drill.common.exceptions.UserException: RESOURCE ERROR: There is not enough heap memory to run this query using the web interface. Please try a query with fewer columns or with a filter or limit condition to limit the data returned. You can also try an ODBC/JDBC client. [Error Id: 702884d0-4e90-46c0-a861-71b7d384084b on kk127.qa.lab:31010]
```

_When querying a non-existent schema:_
```
org.apache.drill.common.exceptions.UserRemoteException: VALIDATION ERROR: Schema [[sys1]] is not valid with respect to either root schema or current default schema. Current default schema: No default schema selected [Error Id: 9fd6041c-b4f1-4f63-912e-d8a44ed318b0 on kk127.qa.lab:31010]
```

_When running an incorrect SQL:_
```
Non-query expression encountered in illegal context (Non-query expression encountered in illegal context) org.apache.drill.common.exceptions.UserException: PARSE ERROR: Non-query expression encountered in illegal context [Error Id: 22238845-2514-4bb0-af8a-be5b6c922a63 on kk127.qa.lab:31010]
```

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
@@ -84,13 +84,11 @@ public class QueryWrapper {
     float usagePercent = getHeapUsage();
 
     // Wait until the query execution is complete or there is error submitting the query
-    if (logger.isDebugEnabled()) {
-      logger.debug("Wait until the query execution is complete or there is error submitting the query");
-    }
+    logger.debug("Wait until the query execution is complete or there is error submitting the query");
     do {
       try {
-        isComplete = webUserConnection.await/*timedWait*/(TimeUnit.SECONDS.toMillis(1)); //periodically timeout 1sec to check heap
-      } catch (InterruptedException e) { /* Timed Out => Check heap usage */ }
+        isComplete = webUserConnection.await(TimeUnit.SECONDS.toMillis(1)); //periodically timeout 1 sec to check heap
+      } catch (InterruptedException e) {}
       usagePercent = getHeapUsage();
       if (usagePercent >  HEAP_MEMORY_FAILURE_THRESHOLD) {
         nearlyOutOfHeapSpace = true;
@@ -99,14 +97,11 @@ public class QueryWrapper {
 
     //Fail if nearly out of heap space
     if (nearlyOutOfHeapSpace) {
-      UserException almostOutOfHeapException = UserException.resourceError(
-          new Throwable(
-              "There is not enough heap memory to run this query using the web interface. "
-              + "Please try a query with fewer columns or with a filter or limit condition to limit the data returned. "
-              + "You can also try an ODBC/JDBC client. "
-              )
-          )
-        .build(logger);
+      UserException almostOutOfHeapException = UserException.resourceError()
+          .message("There is not enough heap memory to run this query using the web interface. ")
+          .addContext("Please try a query with fewer columns or with a filter or limit condition to limit the data returned. ")
+          .addContext("You can also try an ODBC/JDBC client. ")
+          .build(logger);
       //Add event
       workManager.getBee().getForemanForQueryId(queryId)
         .addToEventQueue(QueryState.FAILED, almostOutOfHeapException);
@@ -114,9 +109,7 @@ public class QueryWrapper {
       throw almostOutOfHeapException;
     }
 
-    if (logger.isTraceEnabled()) {
-      logger.trace("Query {} is completed ", queryId);
-    }
+    logger.trace("Query {} is completed ", queryId);
 
     if (webUserConnection.getError() != null) {
       throw new UserRemoteException(webUserConnection.getError());
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
@@ -83,12 +83,17 @@ public class QueryWrapper {
     float usagePercent = getHeapUsage();
 
     // Wait until the query execution is complete or there is error submitting the query
-    logger.debug("Wait until the query execution is complete or there is error submitting the query");
+    if (logger.isDebugEnabled()) {
+      logger.debug("Wait until the query execution is complete or there is error submitting the query");
+    }
     do {
       try {
-        isComplete = webUserConnection.await(TimeUnit.SECONDS.toMillis(1)); /*periodically timeout to check heap*/
-      } catch (Exception e) { }
-
+        isComplete = webUserConnection.timedWait(TimeUnit.SECONDS.toMillis(1)); /*periodically timeout 1sec to check heap */
+      } catch (Exception e) { /* Timed Out => Check usage */
+        if (e instanceof UserException) { //Throw UserException only
+          throw e;
+        }
+      }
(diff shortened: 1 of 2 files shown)


kkhatua:
@arina-ielchiieva / @sohami  could you please review this?
(Thanks for finding the bug and giving pointers to the fix)

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -83,12 +83,17 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
...
-    logger.debug("Wait until the query execution is complete or there is error submitting the query");
+    if (logger.isDebugEnabled()) {
+      logger.debug("Wait until the query execution is complete or there is error submitting the query");
+    }
     do {
       try {
-        isComplete = webUserConnection.await(TimeUnit.SECONDS.toMillis(1)); /*periodically timeout to check heap*/
-      } catch (Exception e) { }
-
+        isComplete = webUserConnection.timedWait(TimeUnit.SECONDS.toMillis(1)); /*periodically timeout 1sec to check heap */
```
sohami: Actually you don't need this `timedWait` method. You can still use same `await(timeout)` as you were using earlier. But outside while loop you can do below check to get the exception in error condition. Check [here](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/work/prepare/PreparedStatementProvider.java#L158) 

```
// check if query completed with exception
if (isComplete && webUserConnection.getError() != null) {
      throw new UserRemoteException(webUserConnection.getError());
}
```

kkhatua: For some strange reason, the webUserConnection.getError() was not giving me the Exception message for invalid schemas. Let me see if I missed something.

kkhatua: Yep. It's sufficient to just wrap the DrillPBError into the UserException. I'm getting the invalid schema exception reported as well.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -83,12 +84,13 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
     float usagePercent = getHeapUsage();
 
     // Wait until the query execution is complete or there is error submitting the query
-    logger.debug("Wait until the query execution is complete or there is error submitting the query");
+    if (logger.isDebugEnabled()) {
```
arina-ielchiieva: No need do this check, it's done internally.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -83,12 +84,13 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
...
+      logger.debug("Wait until the query execution is complete or there is error submitting the query");
+    }
     do {
       try {
-        isComplete = webUserConnection.await(TimeUnit.SECONDS.toMillis(1)); /*periodically timeout to check heap*/
-      } catch (Exception e) { }
-
+        isComplete = webUserConnection.await/*timedWait*/(TimeUnit.SECONDS.toMillis(1)); //periodically timeout 1sec to check heap
```
arina-ielchiieva: `/*timedWait*/` -> please remove this is obvious
`1sec` -> space

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -97,23 +99,29 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
 
     //Fail if nearly out of heap space
     if (nearlyOutOfHeapSpace) {
+      UserException almostOutOfHeapException = UserException.resourceError(
+          new Throwable(
```
arina-ielchiieva: I know it was here before, but still why Throwable?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -97,23 +99,29 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
...
-                    )
-                )
-              .build(logger)
-            );
+        .addToEventQueue(QueryState.FAILED, almostOutOfHeapException);
+      //Return NearlyOutOfHeap exception
+      throw almostOutOfHeapException;
     }
 
     if (logger.isTraceEnabled()) {
```
arina-ielchiieva: I know it was before but this check is not necessary.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
kkhatua: My bad. That was actually a method I introduced before and reverted in the last commit.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
kkhatua: Ok.

kkhatua:
Done with the change

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -97,21 +97,22 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
...
-                    "There is not enough heap memory to run this query using the web interface. "
-                    + "Please try a query with fewer columns or with a filter or limit condition to limit the data returned. "
-                    + "You can also try an ODBC/JDBC client. "
-                    )
-                )
-              .build(logger)
-            );
+        .addToEventQueue(QueryState.FAILED, almostOutOfHeapException);
+      //Return NearlyOutOfHeap exception
+      throw almostOutOfHeapException;
```
arina-ielchiieva: We did not throw exception before, why we are throwing it now?

kkhatua: I added the exception originally to the event queue, but I'm not sure if that will necessarily propagate the exception back. If it isn't thrown back, there is a possibility that before the eventQueue is handled, the resultSet will make it back to the WebServer  that will start constructing the JSONResponse object and run out of memory there.

PR #1382, opened 2018-07-13
DRILL-5495: convert_from function on top of int96 data results in Arr…

vdiravka:
…ayIndexOutOfBoundsException

The only issue is the wrong parameter was used for iteration in the process of reading values of the Nullable Fixed Binary field.


arina-ielchiieva:
Looks like is was mechanical error. @vdiravka nice catch! Though I am not sure to validate this we need to add that many files, total with 100000 rows? Can we create unit test using only one small file? If not, I would say, we might consider removing the test.

vdiravka:
@arina-ielchiieva Yes, it is mechanical issue. 
`recordsToReadInThisPass` is `numNonNullValues` (see `NullableColumnReader#processPagesBulk():284`)
But `recordsReadInThisIteration` is `numNullValues` + `numNonNullValues`. 
In `NullableColumnReader#readField()` the only `numNonNullValues` should be used.

It is hard to reproduce the issue with one small file. But since it was the mechanical error, I think it is fine do not add unit test for this issue. I have updated the PR.

arina-ielchiieva:
@vdiravka, thanks for the explanation. LGTM, +1.

PR #1383, opened 2018-07-17
DRILL-6613: Refactor MaterializedField

vrozov:


exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/ColumnState.java
```diff
@@ -123,7 +123,7 @@ public static MapArrayColumnState build(ResultSetLoaderImpl resultSetLoader,
       // Create the map's offset vector.
 
       UInt4Vector offsetVector = new UInt4Vector(
-          BaseRepeatedValueVector.OFFSETS_FIELD,
+          MaterializedField.OFFSETS_FIELD,
```
paul-rogers: While it is good idea to centralize these names, not sure the `MaterializedField` is the place, since the metadata is logically a bit below the level of vector structure.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
```diff
@@ -453,22 +453,15 @@ protected boolean setupNewSchema() throws SchemaChangeException {
...
-          // outputField is taken from the incoming schema to avoid the loss of nested fields
-          // when the first batch will be empty.
-          if (incomingVector != null) {
-            outputField = incomingVector.getField().clone();
-          } else {
-            outputField = MaterializedField.create(outputName, expr.getMajorType());
-          }
-        } else {
-          outputField = MaterializedField.create(outputName, expr.getMajorType());
+          incomingVector = incoming.getValueAccessorById(id.getIntermediateClass(), id.getFieldIds()).getValueVector();
```
paul-rogers: Are we sure that the dropped code is not needed? This looks like one of those fixes added after a session with the school of hard knocks. That is, added to fix some specific bug.

Can we track down the JIRA ticket to see what the problem is that the code solved and run a test to ensure that removing the code doesn't cause a regression?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -792,7 +792,7 @@ private void setupOutputContainerSchema() {
         }
 
         // make sure to project field with children for children to show up in the schema
-        final MaterializedField projected = field.withType(outputType);
+        final MaterializedField projected = field.copy(outputType);
```
paul-rogers: The `withType` is clearer. With `copy`, we have to infer that this is "copy fields from the give field except for type, which we are changing." `withType` is similar to usages of this pattern elsewhere.

exec/java-exec/src/test/java/org/apache/drill/exec/record/TestMaterializedField.java
```diff
@@ -58,32 +58,24 @@ public void initialize() {
   }
 
   @Test
-  public void testClone() {
-    final MaterializedField cloneParent = parent.clone();
-    final boolean isParentEqual = parent.equals(cloneParent);
-    assertTrue("Cloned parent does not match the original", isParentEqual);
+  public void testCopy() {
+    final MaterializedField cloneParent = parent.copy();
+    assertEquals("Parent copy does not match the original", parent, cloneParent);
```
paul-rogers: Call me paranoid, but this is using one of the functions I'd be worried about. If someone omits some fields from the `equals()` implementation, we'll get a false positive.

Would seem better to do a full test of `equals()`, and only then use it in tests.

exec/java-exec/src/test/java/org/apache/drill/exec/record/TestMaterializedField.java
```diff
@@ -58,32 +58,24 @@ public void initialize() {
...
-        assertTrue("Cloned children do not match the original", isChildrenEqual);
-      }
+    for (Object[] args : matrix) {
+      assertTypeAndPath(parent, (String)args[0], (TypeProtos.MajorType)args[1]);
+      assertTypeAndPath(child, (String)args[0], (TypeProtos.MajorType)args[1]);
     }
+  }
+
+  private void assertTypeAndPath(MaterializedField field, String path, TypeProtos.MajorType type) {
+    final MaterializedField clone = field.copy(path, type);
```
paul-rogers: What are we copying if we change the name and the type? Are we not just keeping the mode? Seems odd... When would we do that? Isn't there a cleaner way to express that idea?

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -38,6 +41,8 @@
  */
 
 public class MaterializedField {
+  public final static MaterializedField OFFSETS_FIELD = new MaterializedField(ValueVector.OFFSETS_VECTOR_NAME, Types.required(MinorType.UINT4), 0);
```
paul-rogers: OK to put in a value vector, but not here. The idea of an offsets field is a vector concept, not a metadata concept.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -49,39 +54,79 @@ private MaterializedField(String name, MajorType type, LinkedHashSet<Materialize
...
+  public MaterializedField copy(MajorType type) {
+    return copy(name, type);
+  }
+
+  public MaterializedField copy(String name) {
+    return copy(name, getType());
+  }
+
+  public MaterializedField copy(String name, final MajorType type) {
+    if (this == OFFSETS_FIELD) {
```
paul-rogers: You could play the same trick with the `bits` field. Just need to make sure we use the same name everywhere. Vectors are notoriously inconsistent in things like this.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -49,39 +54,79 @@ private MaterializedField(String name, MajorType type, LinkedHashSet<Materialize
...
+  public int getPrecision() {
+    return type.getPrecision();
   }
 
-  public MaterializedField newWithChild(MaterializedField child) {
-    MaterializedField newField = clone();
-    newField.addChild(child);
-    return newField;
+  public boolean isNullable() {
+    return type.getMode() == DataMode.OPTIONAL;
```
paul-rogers: Would be a good addition to `Types`.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -115,56 +175,17 @@ public void removeChild(MaterializedField field) {
    */
 
   public void replaceType(MajorType newType) {
-    assert type.getMinorType() == newType.getMinorType();
-    assert type.getMode() == newType.getMode();
+    Preconditions.checkArgument(type.getMinorType() == newType.getMinorType());
+    Preconditions.checkArgument(type.getMode() == newType.getMode());
```
paul-rogers: Any reason to pay this cost for all queries? Isn't a test-time assert sufficient?

Also, what the heck does it mean to replace a type if we assert that the new type is the same as the old one? We are actually setting the precision and scale fields. Maybe the child types.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -49,39 +54,79 @@ private MaterializedField(String name, MajorType type, LinkedHashSet<Materialize
...
-    SerializedField.Builder serializedFieldBuilder = getAsBuilder();
-    for(MaterializedField childMaterializedField : getChildren()) {
-      serializedFieldBuilder.addChild(childMaterializedField.getSerializedField());
+  public MaterializedField copy() {
+    return copy(getName(), getType());
+  }
+
+  public MaterializedField copy(MajorType type) {
+    return copy(name, type);
+  }
```
paul-rogers: As noted above, `withType` and 'withName` are cleaner.

Might even be cleaner to find the users and replace with something clearer.

When would we replace the name? When cloning a vector? Is there standard for the new name?

When would we replace the type? When changing, say, Nullable to Required or visa-versa?

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -49,39 +54,79 @@ private MaterializedField(String name, MajorType type, LinkedHashSet<Materialize
...
+    return copy(name, getType());
+  }
+
+  public MaterializedField copy(String name, final MajorType type) {
+    if (this == OFFSETS_FIELD) {
+      return this;
     }
-    return serializedFieldBuilder.build();
+    MaterializedField field = new MaterializedField(name, type, getChildren().size());
+    field.copyFrom(getChildren(), MaterializedField::copy);
```
paul-rogers: The copy always copies children, which, as we saw in another PR, is a bit murky.

1. Is this ever used for maps? If not, then whether we copy children is ambiguous. (Doesn't matter if we copy children for vectors without them.)
2. If used maps, will copy the children. But, when we create that map clone, we'll add the child vectors one by one, which will create the child fields. Will we end up with duplicates?

Since we are touching this area, should we clean it up?

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -49,39 +54,79 @@ private MaterializedField(String name, MajorType type, LinkedHashSet<Materialize
...
   }
 
   public Collection<MaterializedField> getChildren() {
-    return new ArrayList<>(children);
+    return children;
+  }
+
+  public int getWidth() {
+    return type.getWidth();
+  }
```
paul-rogers: Width is deprecated. We use precision now where width used to be used.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
```diff
@@ -49,39 +54,79 @@ private MaterializedField(String name, MajorType type, LinkedHashSet<Materialize
...
+
```
paul-rogers: Handy functions, but we seem to omit one the two most useful attributes: `getMinorType()` and `getMode()`.

And, if we add these, we can replace the zillions of `getType().getMinorType()` calls with `getMinorType()`. Same for mode.

Also, since these functions are new, they are not called from anywhere. You've indicated in the past you don't like functions added that are not used. Should we hunt down uses of `field.getType().getScale()` and so on to replace them with your new methods?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/flatten/FlattenRecordBatch.java
vrozov: There is no dropped code, the code is re-organized to avoid nested if/else.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
vrozov: `copy` refers to a copy constructor with a deep copy semantics. `with...` usually applies to a builder and that pattern is secondary here.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/ColumnState.java
vrozov: It is not a name. It is an immutable static `MaterializedField` that can be shared between vectors. The name still belongs to `ValueVector`.

exec/java-exec/src/test/java/org/apache/drill/exec/record/TestMaterializedField.java
vrozov: The goal of the PR is to address usage of the `clone()` instead of the copy constructor with a few minor fixes of badly written Java code. The change here addresses the issue of using `assertTrue` where it should assert for equality.

exec/java-exec/src/test/java/org/apache/drill/exec/record/TestMaterializedField.java
vrozov: I am not sure what is the concern. Is it with the test or with the `copy`?

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: I disagree. I don't see any problem with static `OFFSET_FIELD` belonging to `MaterializedField`.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: Please see my comment regarding `clone()`, copy constructor and a builder design pattern.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: I agree that it will be good to unify `bits` field as well as part of another JIRA/PR.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: `children` in `MaterializedField` is a `Set`. Please see `equals()` and `hash()` for `MaterializedField`.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: `getWidth()` is still used.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: The PR does not introduce new functions.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: I would be concerned with the cost only in case `replaceType()` is called for every row (tuple) (note that in java there is no test-time assert, asserts are optimized out only when bytecode is compiled by the hotspot) and I have no idea what it means to replace a type.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
vrozov: In such case one of the `copy` should be named `copyWithNameAndType`. As parameters already provide a hint that `name` and/or `type` will be changed I don't see much need to use the method name for the same.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: I don't see what is wrong with the semantics here, please elaborate. Do you mean that it is not OK to reuse `OFFSETS_FIELD`, why? Is not it immutable (name, type and children are final)? Or do you mean that other `MaterializedField`s can be re-used as well? In the latter case, extending `OFFSET_FIELD` pattern to other `MaterializedField`s should be done in a separate PR. This PR is limited to fixing existing code, including part of the code that compared `OFFSET_FIELD`.

I don't see a need to rename `copy` to `copySafe` as there is no `copyUnsafe` and one should not exists.

paul-rogers: Yes, I mean that other instances can be reused. By "wrong semantics", I mean that we have singled out one particular field instance to be "atomic" and reusable. Nothing wrong with that. It is simply awkward:

* Not only is the offsets field reusable, so is the bits field. So, we' see to be making a statement that there is something special about offsets.
* Offsets is reusable because it is fixed size, required. So, if we handle the case for one fixed size, required field, we should handle all of them. Else, again, we're saying that something is special about offsets.
* Further, since the resulting code copies sometimes, but not others, this is not a true "copy" operation. It is a "copy if necessary" operation. Implementation-wise, the "copy if necessary" should either return the same field (if no copy is needed) or call a "copy" method to make an actual copy when needed.

None of this is wrong; the code here works. But, if we are going to the trouble to improve working code, we might want to handle the other cases as well.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
paul-rogers: My point is not how things are implemented. The point is: we almost never want to copy children when copying a `MaterializedField`. Why? Because the consumer of that copy may be code that creates a new vector. When it does, it will add child vectors (copies of the source vectors), and that action will add new child fields.

This is why I said that "copy" is misleading: we need to understand the context in which we make the copy, and possibly name the method accordingly: "copyForNewVector", "fullCopy", "copyIfNeeded", and so on. All of these should have comments that explain the use case that they serve.

Sorry that this is so complex; it is just the way the code has evolved. As I recently found, it is quite hard to change entrenched code behavior, even when it is not quite right.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
paul-rogers: It should not be. @arina-ielchiieva did a PR a year ago that deprecated width. References to width represent bugs.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
paul-rogers: I see. You moved one-liners from elsewhere in the file and turned them into four-liners...

Any reason to do this? Drill has an informal pattern to prefer one-liners where they make sense.

Obviously this is a trivial discussion, but seems odd to make perfectly good code longer...

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
paul-rogers: OK. Naming can lead to a big discussion: the code works just the same either way, so consider this one resolved.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
sohami: There is a case in `UnnestRecordBatch` where we have to keep a copy of Materialized field along with it's children of incoming repeated type vector which may be repeated map as well. This is to detect if there is any schema change with new incoming batch as compared to previous one.

Also in case of creating a Map vector in output container one can call `container.addOrGet(field, null) ` if the field contains children information then this call will take care of adding all the children vectors in Map vector as well. So you don't have to explicitly add child vectors.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: > Not only is the offsets field reusable, so is the bits field. So, we' see to be making a statement that there is something special about offsets.

Offset field was already handled differently compared to other fields including bits field prior to the PR. The only new behavior that the PR introduces is a singleton pattern for offset field during copy operation to avoid compare by name where compare by identity should be used in the first place. As there was no compare by name for bits, bits field is not handled. I do agree that bits can be handled similar to offset field, but it is not a goal for this PR. The goal of the PR outlined in the JIRA description is to avoid using `clone()` for `MaterializedField`.

> Offsets is reusable because it is fixed size, required. So, if we handle the case for one fixed size, required field, we should handle all of them. Else, again, we're saying that something is special about offsets.

Offset field is reusable because it is immutable (name, type and children are all final). Bits field can be also made reusable, but it is not a goal of the PR. The PR does not make offset field special as it was already handled differently prior to the PR.

>Further, since the resulting code copies sometimes, but not others, this is not a true "copy" operation. It is a "copy if necessary" operation. Implementation-wise, the "copy if necessary" should either return the same field (if no copy is needed) or call a "copy" method to make an actual copy when needed.

It is the only `copy` operation that is available outside of `MaterializedField` class and no any other `copy` should exist. The `copy` decides whether or not it needs to create another instance or it can return a singleton.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: @paul-rogers In cases where `deep` copy is not necessary, `MaterializedField` already provides `create()` method that takes name and type. There is no need to introduce `copyForNewVector`, `fullCopy` or `copyIfNeeded` IMO.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: I don't know why it was deprecated instead of being completely removed. If it needs to be removed, somebody has to open another JIRA/PR.

exec/vector/src/main/java/org/apache/drill/exec/record/MaterializedField.java
vrozov: I don't see such "informal" pattern. I do see more cases where drill prefers 3 lines over a single line pattern, so is the change.

PR #1386, opened 2018-07-18
DRILL-6574: Add option to push LIMIT(0) on top of SCAN (late limit 0 optimization)

KazydubB:


arina-ielchiieva:
@vvysotskyi could you please review?

arina-ielchiieva:
Please note currently Travis checks do not pass since #1384 should be merged first.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/FindLimit0Visitor.java
```diff
@@ -126,9 +161,84 @@ public static boolean containsLimit0(final RelNode rel) {
     return visitor.isContains();
   }
 
-  private boolean contains = false;
+  /**
+   * TODO(DRILL-3993): Use RelBuilder to create a limit node to allow for applying this optimization in potentially
```
vvysotskyi: Please remove this comment. I don't think that it should be done in `RelBuilder` since rel node should be checked starting from the top, but `RelBuilder` may be used at any level.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLateLimit0Optimization.java
```diff
@@ -0,0 +1,111 @@
...
+public class TestLateLimit0Optimization extends BaseTestQuery {
+
+  @Test
+  public void convertFromJson() throws Exception {
+    checkThatQueryIsNotOptimized("SELECT CONVERT_FROM('{x:100, y:215.6}' ,'JSON') AS MYCOL FROM (VALUES(1))");
+  }
+
+  private static void checkThatQueryIsNotOptimized(final String query) throws Exception {
+    PlanTestBase.testPlanMatchingPatterns(wrapLimit0(query),
+        new String[]{},
```
vvysotskyi: No need to create an empty array, `null` may be passed.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLateLimit0Optimization.java
```diff
@@ -0,0 +1,111 @@
...
+    return "SELECT * FROM (" + query + ") LZT LIMIT 0";
+  }
+
+  @Test
+  public void convertToIntBE() throws Exception {
+    checkThatQueryIsOptimized("SELECT CONVERT_TO(r_regionkey, 'INT_BE') FROM cp.`tpch/region.parquet`");
+  }
+
+  private static void checkThatQueryIsOptimized(final String query) throws Exception {
+    PlanTestBase.testPlanMatchingPatterns(wrapLimit0(query),
```
vvysotskyi: `PlanTestBase` has a method which takes two arguments, please use that method here.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/limit/TestLateLimit0Optimization.java
```diff
@@ -0,0 +1,111 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.limit;
+
+import org.apache.drill.test.BaseTestQuery;
+import org.apache.drill.PlanTestBase;
+import org.junit.Test;
+
+public class TestLateLimit0Optimization extends BaseTestQuery {
```
vvysotskyi: It would be good to add a test which checks that `limit 0` wasn't inserted when the option is disabled.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/FindLimit0Visitor.java
```diff
@@ -126,9 +161,84 @@ public static boolean containsLimit0(final RelNode rel) {
...
+   * any of the transformations, but currently this can be applied after Drill logical transformation, and before
+   * Drill physical transformation.
+   */
+  public static DrillRel addLimitOnTopOfLeafNodes(final DrillRel rel) {
+    final Pointer<Boolean> isUnsupported = new Pointer<>(false);
```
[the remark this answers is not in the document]
KazydubB: Because the variable is re-assigned and is used in anonymous classes (unsupportedFunctionsVisitor, unsupportedOperationsVisitor variables) and should be (effectively) final.

PR #1387, opened 2018-07-18
DRILL-6603: Set num_nulls for parquet statistics to -1 when actual number is not defined.

arina-ielchiieva:
Details in [DRILL-6603](https://issues.apache.org/jira/browse/DRILL-6603).


arina-ielchiieva:
@vvysotskyi could you please review?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataBase.java
```diff
@@ -108,6 +105,18 @@
...
   public static abstract class ColumnMetadata {
+
+    /**
+     * Number of nulls is considered to be valid if its value is not null and -1.
+     *
+     * @return true if nulls value is defined, false otherwise
+     */
+    @JsonIgnore
+    public boolean isNullsDefined() {
+      Long nulls = getNulls();
```
vrozov: Why is it necessary to deal both with `null` and `-1`? Why is `getNulls` a `Long` and not a `long`? Consider renaming it to `getNumNulls`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataBase.java
```diff
@@ -108,6 +105,18 @@
...
 
   public static abstract class ColumnMetadata {
+
+    /**
+     * Number of nulls is considered to be valid if its value is not null and -1.
+     *
+     * @return true if nulls value is defined, false otherwise
+     */
+    @JsonIgnore
+    public boolean isNullsDefined() {
```
vrozov: Consider `isNumNullsSet()`.

exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestConvertCountToDirectScan.java
```diff
@@ -143,7 +143,7 @@ public void ensureConvertForSeveralColumns() throws Exception {
 
       testPlanMatchingPatterns(
```
vvysotskyi: It may be replaced by a method which takes two arguments.

exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestConvertCountToDirectScan.java
```diff
@@ -159,4 +159,30 @@ public void ensureConvertForSeveralColumns() throws Exception {
...
 
+  @Test
+  public void ensureCorrectCountWithMissingStatistics() throws Exception {
+    dirTestWatcher.copyResourceToRoot(Paths.get("parquet", "wide_string.parquet"));
+
+    test("use dfs.tmp");
+    String tableName = "wide_str_table";
+    try {
+      // table will contain two partitions: one - with null value, second - with non null value
+      test("create table %s partition by (col_str) as select * from dfs.`parquet/wide_string.parquet`", tableName);
```
vvysotskyi: I think table `wide_string.parquet` may be accessible from `cp` workspace, so please remove line `dirTestWatcher.copyResourceToRoot()` above and change `dfs` to `cp` in the query if it will work.

exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestConvertCountToDirectScan.java
```diff
@@ -159,4 +159,30 @@ public void ensureConvertForSeveralColumns() throws Exception {
...
+
+      String query = String.format("select count(col_str) as cnt_str, count(*) as cnt_total from %s", tableName);
+
+      // direct scan should not be applied since we don't have statistics
+      testPlanMatchingPatterns(query, new String[]{}, new String[]{"DynamicPojoRecordReader"});
```
vvysotskyi: No need to create an empty array `new String[]{}`, just null may be passed, since this method has a check for nulls.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataBase.java
arina-ielchiieva: Previously created Drill metadata files might have nulls defined as null. Agree that renaming is good idea but since it will affect several versions of `ColumnMetadata` class which is used in ser /de when creating Drill parquet metadata files, I'd rather not make such changes right now. Since PR resolves release blocker, I suggest we keep changes at minimum to avoid possible regressions. Created Jira for the improvement - https://issues.apache.org/jira/browse/DRILL-6620.

vrozov:
@vvysotskyi In Apache, when one of the community members decides to review a PR and submits his/her comments, as a courtesy to the community and that member, the PR should not be merged without giving the other reviewer and other community members an opportunity to take a look at the updated PR and PR contributor's responses.

arina-ielchiieva:
@vrozov sorry about that but it was a release blocker and you had only minor style comments which I have addressed. I have asked  Volodymyr to merge changes to unblock the release since I'll be out for couple of days. If you see anything critical please file a Jira.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataBase.java
vrozov: @arina-ielchiieva My major concern is not the name of the method, but using `Long` when `long` should be used. Using `long` will also address my second major concern: using `null` and `-1` for "not set" definition.

arina-ielchiieva: I know and totally agree that `long` usage is more correct choice here but this was done before and I left `Long` for backward compatibility plus did not want to change all `ColumnMetadata` successors yo avoid any problems with deserialization. I think we can make this changes in subsequent Jira DRILL-6620, added comment in Jira.

vrozov:
@arina-ielchiieva
- `Long` vs `long` is not a "minor style" concern.
- In Apache a contributor/PMC/VP can't ask another committer to merge his/her changes, it is clearly against "Apache way" where a committer represents him/herself only.
- In Apache, while there may be a blocker for a release, there are no milestones.

arina-ielchiieva:
@vrozov 
1. I did not introduce Long and just kept previous code to ensure backward compatibility, I don't think it's a major concern plus I filed Jira for future enhancement. If you fill we should revert the commit, we can do that.
2. Volodymyr did the review and merged the changes, though he should have waiting for your approval. Again sorry about that.

vrozov:
@arina-ielchiieva The goal of the comments is to avoid the same happening in the future (not to fix something that is already done, so no need to revert the commit, especially that it fixes DRILL-6603) whether it is for PR that I review or it is reviewed by somebody else (in Apex, for example, there is a [merging PR rule](http://apex.apache.org/contributing.html) that a PR can't be merged unless all reviewers sign off with a grace period of 48 hours.)

arina-ielchiieva:
@vrozov, thank you! Lesson learnt.

PR #1390, opened 2018-07-20
DRILL-6624: Fix loss of the table row type when the same schema name …

vvysotskyi:
…was specified as single path and as a complex path in the same query

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
+++ b/exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
@@ -17,7 +17,6 @@
  */
 package org.apache.calcite.jdbc;
 
-import com.google.common.collect.Lists;
 import org.apache.calcite.DataContext;
 
 import org.apache.calcite.linq4j.tree.Expression;
@@ -33,6 +32,7 @@ import org.apache.drill.exec.store.StoragePluginRegistry;
 import org.apache.drill.exec.store.SubSchemaWrapper;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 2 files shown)


vvysotskyi:
@vdiravka, could you please take a look?

exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
```diff
@@ -91,15 +91,17 @@ public void loadSchemaFactory(String schemaName, boolean caseSensitive) {
           return;
         }
 
-        // Found the storage plugin for first part(e.g. 'dfs') of schemaName (e.g. 'dfs.tmp')
-        // register schema for this storage plugin to 'this'.
-        plugin.registerSchemas(schemaConfig, thisPlus);
-
+        // Looking for the SchemaPlus for the top level (e.g. 'dfs') of schemaName (e.g. 'dfs.tmp')
+        SchemaPlus firstLevelSchema = thisPlus.getSubSchema(paths.get(0));
```
vdiravka: What is thisPlus? Could we consider renaming it?

vvysotskyi: It is also `SchemaPlus` instance which corresponds to the current schema. Thanks for pointing this, renamed `thisPlus` into `schemaPlus`.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/TestSchema.java
```diff
@@ -84,4 +84,42 @@ public void testUseBrokenStorage() throws Exception {
...
+          .sql("create table dfs.tmp.t2 as select 1 as c2")
+          .run();
+
+      client.queryBuilder()
+          .sql("create table dfs.tmp.`t3` as SELECT 1 as c2,1 as c1")
+          .run();
+
+      client.testBuilder()
+          .sqlQuery("select a.c2 from dfs.tmp.`t2` a\n" +
+            "join  dfs.tmp.`t3` b ON b.c2 = a.c2 \n" +
```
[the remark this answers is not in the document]
vvysotskyi: Experimented with a query and tables number. This bug is not reproduced when a single table is joining itself, but it is reproduced when two tables are joined. So I rewrote the query and used two tables in the test.

PR #1391, opened 2018-07-22
DRILL-6622: Fixed a NullPointerException in a query with Union

sachouche:
@Ben-Zvi, can you please review the fix?
Thanks!

There were two bugs in the Aggregator batch sizing logic:
Issue I
- The aggregator runs in a loop to consume all input batches
- The loop was updating the batch sizing stats after they were consumed
- Assume output-row-count is 1 and we receive a batch with at least 32k + 1 records
- The code would create 32k output batches (one per incoming record) and then fails because of overflow
- Fix - Now updating the batch sizing logic when a non-empty batch is received and before the processing loop

Issue II
- The Aggregator has two main modules: AggregatorBatch and Aggregator objects
- Both share the same "incoming" record batch instance
- Though there is logic to spill incoming batches when under pressure
- The batch sizing logic was not aware that when batches are spilled the shared "incoming" object instance will diverge; that is, the Aggregator object will mutate the incoming object
- The batch sizer was being invoked with a stale "incoming" object (the one from the AggregatorBatch)
- Fix - Update the  Aggregator code to always pass the active incoming object explicitly

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggBatch.java
@@ -114,8 +114,13 @@ public class HashAggBatch extends AbstractRecordBatch<HashAggregate> {
 
     @Override
     public void update() {
+      update(incoming);
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


sachouche:
@Ben-Zvi , thank you for the review!

PR #1392, opened 2018-07-23
DRILL-6611 : Enable meta-enter query submission in web query interface

hrbrmstr:
As the git info shows, one file —`query.ftl` has been modified with a small addition of vanilla javascript to add an event listener for meta-enter which will submit the query vs needing to use the mouse.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/query/query.ftl
+++ b/exec/java-exec/src/main/resources/rest/query/query.ftl
@@ -111,6 +111,12 @@
       enableBasicAutocompletion: true,
       enableLiveAutocompletion: false
     });
+    // meta-enter to submit query
+    document.getElementById('queryForm')
+            .addEventListener('keydown', function(e) {
+      if (!(e.keyCode == 13 && e.metaKey)) return;
+      if (e.target.form) e.target.form.submit();
+    });
   </script>
 
 </#macro>


exec/java-exec/src/main/resources/rest/query/query.ftl
```diff
@@ -111,6 +111,12 @@
       enableBasicAutocompletion: true,
       enableLiveAutocompletion: false
     });
+    // meta-enter to submit query
+    document.getElementById('queryForm')
+            .addEventListener('keydown', function(e) {
+      if (!(e.keyCode == 13 && e.metaKey)) return;
```
kkhatua: For Windows machines, the `metaKey` equivalent is the Windows key. Interestingly, this is not supported by Firefox 48+ (Ref: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey )
For Windows, the key combination is `Ctrl+Enter` .  (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/ctrlKey)
So, you'll need to do a check for the key events something like this:
```
if (! ((e.ctrlKey || e.metaKey) &&  e.keyCode == 13) ) return;
```

kkhatua: Folks familiar with Mac tell me the Meta key has been deprecated and since Ctrl is available on Mac, let's go with that combination: `Ctrl + Enter`
Also, check for `keyCode == 10` (Refer: https://bugs.chromium.org/p/chromium/issues/detail?id=79407 )
This doesn't seem to be listed anymore, but might be worth having to ensure compatibility.

kkhatua:
@hrbrmstr can you also change the PR's title to 
**DRILL-6611: Enable meta-enter query submission in web query interface**
This will facilitate the Apache JIRA system to pick up the PR and automatically link to it.

arina-ielchiieva:
@hrbrmstr any update on the PR?

kkhatua:
@hrbrmstr I've made a fork of your PR and made the additional changes there.
https://github.com/kkhatua/drill/commits/DRILL-6611
I've tested the PR with the combined changes and no issues on Mac and Windows' browsers.

Do you want to make changes to this PR ? Alternatively, I could close this PR and commit the above branch's changes instead.

arina-ielchiieva:
@kkhatua since there is no reply, please go ahead and open alternative PR but keep the commit with original author name.

kkhatua:
Closing this PR as it is a part of #1521 , which addressed some impersonation bugs.

PR #1394, opened 2018-07-23
Drill 6626

sachouche:
The IndexOutOfBoundException was happening during Hash Table rehash:

The was a regression when doing batch sizing
Each outgoing-batch needed to fix its hash values (which were cached)
Each outgoing batch should start at index (num-out-batches - 1) * MAX_BATCH_SZ; this is based on the insertion logic
The code instead used the real row count instead of MAX_BATCH_SZ
Fix - Put back the original code since the indexing scheme didn't change


sachouche:
@Ben-Zvi , can you please review the fix. Thanks!

PR #1395, opened 2018-07-23
DRILL-6629 BitVector split and transfer does not work correctly for transfer length < 8

bitblender:
The first src byte was not being read when the transfer length was < 0

### DIFF
--- a/exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
+++ b/exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
@@ -323,7 +323,8 @@ public final class BitVector extends BaseDataValueVector implements FixedWidthVe
       if (length % 8 != 0) {
         // start is not byte aligned so we have to copy some bits from the last full byte read in the
         // previous loop
-        byte lastButOneByte = byteIPlus1;
+        // if numBytesHoldingSourceBits == 1, lastButOneByte is the first byte, but we have not read it yet, so read it
+        byte lastButOneByte = (numBytesHoldingSourceBits == 1) ? this.data.getByte(firstByteIndex) : byteIPlus1;
         byte bitsFromLastButOneByte = (byte)((lastButOneByte & 0xFF) >>> firstBitOffset);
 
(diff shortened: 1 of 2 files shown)


bitblender:
@HanumathRao @sohami Previous fix for the BitVector split and transfer was missing a case. Please review this.

exec/vector/src/main/java/org/apache/drill/exec/vector/BitVector.java
```diff
@@ -323,7 +323,8 @@ public void splitAndTransferTo(int startIndex, int length, BitVector target) {
       if (length % 8 != 0) {
         // start is not byte aligned so we have to copy some bits from the last full byte read in the
         // previous loop
-        byte lastButOneByte = byteIPlus1;
+        // if numBytesHoldingSourceBits == 1, lastButOneByte is the first byte, but we have not read it yet, so read it
+        byte lastButOneByte = (numBytesHoldingSourceBits == 1) ? this.data.getByte(firstByteIndex) : byteIPlus1;
```
ppadma: @bitblender I think there could be a problem here. please check. If you are copying say from firstBitOffset 2, length 4. We want to copy 4  bits only. But, this might copy 6 bits. bitsFromLastButOneByte will be all bits from firstBitOffset to the end of the byte.

bitblender: Yes. This can happen. Good observation!

I did not handle it in the initial PR because this can happen only in the single-byte (length < 8) case, so there is always allocated memory available(at least 1 byte is alloced in the target vector) to write the extra bits and since valueCount of the target was being set correctly, this was not a concern. Writing the extra bits would have no observable impact if the buffer was not zero-initialized. 

But since we are zero-initializing buffers, it makes sense to be precise and leave bits beyond valueCount as 0. I have made changes to mask off the extra bits.

ppadma: ok. LGTM. +1.

PR #1401, opened 2018-07-26
DRILL-6616: Batch Processing for Lateral/Unnest

sohami:
This PR contains the both planning and execution side changes for batch processing support in Lateral and Unnest. It has 3 different commits (planning side change, lateral execution changes and unnest execution changes) and needs to be merged together for tests to pass. Before closing or merging this PR please make sure there is 3 +1's from different committers.
@amansinha100  - to review planning side changes.
@parthchandra  - to review lateral execution changes
@sohami - to review unnest execution changes

### DIFF
--- a/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcPrel.java
+++ b/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcPrel.java
@@ -20,7 +20,6 @@ package org.apache.drill.exec.store.jdbc;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.List;
 
 import org.apache.calcite.adapter.java.JavaTypeFactory;
 import org.apache.calcite.adapter.jdbc.JdbcImplementor;
@@ -132,9 +131,4 @@ public class JdbcPrel extends AbstractRelNode implements Prel {
     return false;
   }
 
-  @Override
-  public Prel addImplicitRowIDCol(List<RelNode> children) {
-    throw new UnsupportedOperationException("Adding Implicit RowID column is not supported for " +
-                                            this.getClass().getSimpleName() + " operator ");
-  }
 }
--- a/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcPrel.java
+++ b/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcPrel.java
@@ -20,6 +20,7 @@ package org.apache.drill.exec.store.jdbc;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.List;
 
 import org.apache.calcite.adapter.java.JavaTypeFactory;
 import org.apache.calcite.adapter.jdbc.JdbcImplementor;
@@ -129,5 +130,11 @@ public class JdbcPrel extends AbstractRelNode implements Prel {
   @Override
   public boolean needsFinalColumnReordering() {
     return false;
+  }
+
+  @Override
+  public Prel addImplicitRowIDCol(List<RelNode> children) {
+    throw new UnsupportedOperationException("Adding Implicit RowID column is not supported for " +
+                                            this.getClass().getSimpleName() + " operator ");
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
@@ -21,10 +21,10 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import org.apache.drill.exec.exception.SchemaChangeException;
 import org.apache.drill.exec.ops.FragmentContext;
-import org.apache.drill.exec.physical.base.LateralContract;
 import org.apache.drill.exec.record.BatchSchema.SelectionVectorMode;
 import org.apache.drill.exec.record.RecordBatch;
 import org.apache.drill.exec.record.TransferPair;
+import org.apache.drill.exec.vector.IntVector;
 import org.apache.drill.exec.vector.SchemaChangeCallBack;
 import org.apache.drill.exec.vector.ValueVector;
 import org.apache.drill.exec.vector.complex.RepeatedValueVector;
@@ -36,9 +36,10 @@ import java.util.List;
 import static org.apache.drill.exec.record.BatchSchema.SelectionVectorMode.NONE;
 
 /**
- * Contains the actual unnest operation. Unnest is a simple transfer operation in this impelementation.
- * For use as a table function, we will need to change the logic of the unnest method to operate on
- * more than one row at a time and remove any dependence on Lateral
+ * Contains the actual unnest operation. Unnest is a simple transfer operation in this implementation.
+ * Additionally, unnest produces an implicit rowId column that allows unnest to output batches with many
+ * rows of incoming data being unnested in a single call to innerNext(). Downstream blocking operators need
+ * to be aware of this rowId column and include the rowId as the sort or group by key.
  * This class follows the pattern of other operators that generate code at runtime. Normally this class
  * would be abstract and have placeholders for doSetup and doEval. Unnest however, doesn't require code
  * generation so we can simply implement the code in a simple class that looks similar to the code gen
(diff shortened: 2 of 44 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
```diff
@@ -48,24 +51,47 @@
...
+   *
+   *
+   *
+                valueIndex  0         1       2   3
+                            |- - - - -|- - - -|- -|- - - -|
+                            | | | | | | | | | | | | | | | |
+                            |- - - - -|- - - -|- -|- - - -|
+           innerValueIndex  0 1 2 3 4 0 1 2 3 0 1 0 1 2 3 |
+     runningInnerValueIndex 0 1 2 3 4 5 6 7 8 9 ...
+   *
```
sohami: Thanks for ascii art :)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
```diff
@@ -48,24 +51,47 @@
...
+   *
+   *
+   */
+  private int valueIndex; // index in the incoming record being processed
+  // The index in the unnest column that is being processed. We start at zero and continue until
   // InnerValueCount is reached or  if the batch limit is reached
   // this allows for groups to be written between batches if we run out of space, for cases where we have finished
   // a batch on the boundary it will be set to 0
```
sohami: please revisit this comment, it's little confusing. How about just saying _"It's the index of the array element in the unnest column at row pointed by valueIndex which is currently being processed. It starts at zero and continue until InnerValueCount is reached or the batch limit is reached. It allows for groups to be written across batches if we run out of space"_

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
```diff
@@ -83,55 +109,80 @@ public void setOutputCount(int outputCount) {
...
+            // Flush this batch.
+            break outer;
+          }
+          try {
+            // rowId starts at 1, so the value for rowId is valueIndex+1
+            rowIdVectorMutator.setSafe(outputIndex, valueIndex + 1);
+
+          } finally {
+            outputIndex++;
+            //currentInnerValueIndexLocal++;
```
sohami: please remove commented code if not used.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestRecordBatch.java
```diff
@@ -103,16 +109,19 @@ public void update() {
...
+
       // Average rowWidth of single element in the unnest list.
       // subtract the offset vector size from column data size.
       final int avgRowWidthSingleUnnestEntry = RecordBatchSizer
           .safeDivide(columnSize.getTotalNetSize() - (getOffsetVectorWidth() * columnSize.getValueCount()), columnSize
               .getElementCount());
 
       // Average rowWidth of outgoing batch.
-      final int avgOutgoingRowWidth = avgRowWidthSingleUnnestEntry;
+      final int avgOutgoingRowWidth = avgRowWidthSingleUnnestEntry + rowIdColumnSize.getDataSizePerEntry();
```
sohami: If still using RecordBatchSize then `getDataSizePerEntry() ---> getStdDataSizePerEntry() . `First one will return zero since the vector is empty to start with.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestRecordBatch.java
```diff
@@ -103,16 +109,19 @@ public void update() {
       final MaterializedField field = incoming.getSchema().getColumn(typedFieldId.getFieldIds()[0]);
 
       // Get column size of unnest column.
+
       RecordBatchSizer.ColumnSize columnSize = getRecordBatchSizer().getColumn(field.getName());
 
+      RecordBatchSizer.ColumnSize rowIdColumnSize = getRecordBatchSizer().new ColumnSize(rowIdVector, null);
```
sohami: Since we know `RowId` is always going to be an `IntVector`, we can directly use `TypeHelper.getSize()` instead of going through `RecordBatchSizer`.

parthchandra: Good catch. Thanks.

contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcPrel.java
```diff
@@ -130,4 +131,10 @@ public SelectionVectorMode getEncoding() {
   public boolean needsFinalColumnReordering() {
     return false;
   }
+
+  @Override
+  public Prel addImplicitRowIDCol(List<RelNode> children) {
```
amansinha100: I am thinking a cleaner interface would be to have an additional method `supportsImplicitRIDCol()`  which would return FALSE by default.  You could use the Java 8 feature for default interface methods [1].   Only call `addImplicitRowIdCol()` if adding the rowid is supported.  

[1] https://dzone.com/articles/interface-default-methods-java

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
```diff
@@ -317,6 +319,33 @@ public Void visitCall(RexCall call) {
...
   }
 
+  /**
+   * Given a list of rexnodes it transforms the rexnodes by changing the expr to use new index mapped to the old index.
+   * @param builder : RexBuilder from the planner.
+   * @param exprs: RexNodes to be transformed.
+   * @param corrMap: Mapping between old index to new index.
+   * @return
+   */
+  public static List<RexNode> transformExprs(RexBuilder builder, List<RexNode> exprs, Map<Integer, Integer> corrMap) {
```
amansinha100: Doesn't Calcite already provide similar transformation in RexUtil ?

sohami:
+1 for changes related to DRILL-6618: Unnest changes for implicit column.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/FilterPrel.java
```diff
@@ -81,4 +86,17 @@ public boolean needsFinalColumnReordering() {
...
+  @Override
+  public Prel addImplicitRowIDCol(List<RelNode> children) {
+    RexBuilder builder = this.getCluster().getRexBuilder();
+    return (Prel) this.copy(this.traitSet, children.get(0), DrillRelOptUtil.transformExpr(builder, condition, buildMap()));
+  }
+
+  private Map<Integer, Integer> buildMap() {
+    Map<Integer, Integer> map = new HashMap<>();
+    for (int i=0;i<this.getInput().getRowType().getFieldCount();i++) {
+      map.put(i, i+1);
```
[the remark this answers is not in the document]
HanumathRao: Yes. I did consider this option but for aggregations calcite expects to place the group by columns at the beginning of the row type. Hence, I thought it is better to do this uniformly across all other operators as well.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
HanumathRao: AFAIK, calcite does something like this but it is done in a private method.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/FilterPrel.java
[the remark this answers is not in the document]
amansinha100: ok, could you pls add a comment about this or better still make the right shifting buildMap() a common function.

HanumathRao:
@amansinha100  Thank you for the review. I have done the needed changes.

amansinha100:
@HanumathRao  Updated changes LGTM.  +1

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unnest/UnnestImpl.java
vlsi: Unfortunately it won't be visible in javadoc.
I guess it should be wrapped with `<pre>{@code ...}</pre>`:  https://stackoverflow.com/a/542142/1261287

PR #1403, opened 2018-07-26
DRILL-6634: Add udf module under contrib directory and move some udfs into it

arina-ielchiieva:
1. Created new contrib/udf module.
2. Moved distance, phonetic, networking, crypto functions from java-exec to contrib/udf module.
3. Moved functions from gis module to contrib/udf module. Removed gis module.
4. Removed unnecessary dependencies from java-exec module.
5. Minor refactoring of moved functions code.

Jira - [DRILL-6634](https://issues.apache.org/jira/browse/DRILL-6634).


paul-rogers:
@arina-ielchiieva: question: as you know, Drill functions are quite difficult to debug -- they rely on code generation and code rewriting, one cannot simply step through the code.

UDFs are also hard to test: one needs to set up the inputs to the functions just right to test all cases.

I've found that using the Drill unit test framework is very helpful, especially for setting up input data and verifying output data. However, as some dev list user pointed out, the test classes were not historically available to code outside the `exec` module. Perhaps Tim's recent changes fixed this so that unit test classes are available in `contrib`?

In short, the request is that we ensure that code in `contrib` can be debugged and tested the same way as code in `exec`. Else, people will have to do what I found I had to do recently: develop the code in `exec`, then, once it is done, move it to `contrib`. If I had to make changes, I had to temporarily move it back to `exec`. Let's make sure that this is no longer necessary.

arina-ielchiieva:
@paul-rogers not sure to which issue you are referring to, never heard about it. Anyway new udf module has test dependencies to exec module including tests part, so all handly test classes available in exec are available in udf. When I moved udfs, I have moved unit tests testing these udfs as well. You can checkout my branch and check in case there are some concerns or I did not understand possible issue correctly.

paul-rogers:
@arina-ielchiieva, if the test dependencies are available than all is good. Any remaining issues can be addressed if/when they occur.

PR #1404, opened 2018-07-27
DRILL-6639: Exception happens while displaying operator profiles for …

vdiravka:
…some queries

Details in Jira ticket

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
@@ -253,6 +253,10 @@ public class OperatorWrapper {
    * @return index of spill metric
    */
   private int getSpillCycleMetricIndex(CoreOperatorType operatorType) {
+    // TODO: DRILL-6642, replace null values for ProtocolMessageEnum with UNRECOGNIZED NullValue to avoid null checks
+    if (operatorType == null) {
+      return NO_SPILL_METRIC_INDEX;
+    }
     String metricName;
 
     switch (operatorType) {
--- a/protocol/src/main/java/org/apache/drill/exec/proto/beans/CoreOperatorType.java
+++ b/protocol/src/main/java/org/apache/drill/exec/proto/beans/CoreOperatorType.java
@@ -67,7 +67,15 @@ public enum CoreOperatorType implements com.dyuproject.protostuff.EnumLite<CoreO
     UNNEST(42),
     HIVE_DRILL_NATIVE_PARQUET_ROW_GROUP_SCAN(43),
     JDBC_SCAN(44),
-    REGEX_SUB_SCAN(45);
+    REGEX_SUB_SCAN(45),
+    MAPRDB_SUB_SCAN(46),
+    MONGO_SUB_SCAN(47),
+    KUDU_WRITER(48),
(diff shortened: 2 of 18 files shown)


arina-ielchiieva:
@vdiravka adding operators names into proto fixes part of the issue. Query profile page should be displayed correctly even if operator name is not in proto. Show NA or something else. Actually, it was this way before, don’t know which commit changed this behavior.

vdiravka:
@arina-ielchiieva The operator type was always (in earlier Drill versions) null, if operator doesn't exist in protobuf file. But DRILL-6279 (`OperatorWrapper.getSpillCycleMetricIndex()`) enabled using of `operatorType` without handling `null` cases. It is possible to handle everywhere such null cases. But the proper way is to update to the latest protobuf (3.6.0 for now) version, which allows to set `UNRECOGNIZED` Enum NullValue.
If it is fine I will proceed with a new Jira ticket.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
```diff
@@ -255,6 +255,10 @@ public void addSummary(TableBuilder tb, HashMap<String, Long> majorFragmentBusyT
   private int getSpillCycleMetricIndex(CoreOperatorType operatorType) {
     String metricName;
```
arina-ielchiieva: Maybe move metricName after if statement?

arina-ielchiieva:
@vdiravka could you please also check children of `EasyFormatPlugin` for missing operator types: `public int getReaderOperatorType() {` method. Looks like `HttpdLogFormatPlugin`, `ImageFormatPlugin`, `SequenceFileFormatPlugin` are missing value in proto enum.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/OperatorWrapper.java
vdiravka: Done

arina-ielchiieva:
+1, LGTM.

protocol/src/main/java/org/apache/drill/exec/proto/beans/CoreOperatorType.java
```diff
@@ -131,6 +139,14 @@ public static CoreOperatorType valueOf(int number)
             case 43: return HIVE_DRILL_NATIVE_PARQUET_ROW_GROUP_SCAN;
             case 44: return JDBC_SCAN;
             case 45: return REGEX_SUB_SCAN;
+            case 46: return MAPRDB_SUB_SCAN;
```
sohami: I think for future we should come up with a strategy to avoid missing OperatorType from protobuf and also probably separate the contrib related format plugins or operator to have a different range of values. This will help different distro's to also define their own operator in that specific range rather than having conflict with Drill's internal core operators. Having different range is still a short term solution.

For longer term we can discuss other solutions like some form of registration techniques for all format plugin and operator which will advertise their own unique id (thanks to @parthchandra) for suggesting it. But we can discuss more for future improvements.

arina-ielchiieva: Good point, also once I have created Jira to allow operators names to be in String rather with dependency on protobuf but this would be a breaking change (https://issues.apache.org/jira/browse/DRILL-5405).

PR #1405, opened 2018-07-27
DRILL-6640: Drill takes long time in planning when there are large number of files in views/tables DFS parent directory

kr-arjun:
Includes
1.  Modified DotDrillUtil.getDotDrills implementation to avoid using FS globStatus call with GLOB for a given base file name.
2. Added unit test cases for the new method.
3. Code refactoring to include additional comments.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillUtil.java
@@ -18,10 +18,14 @@
 package org.apache.drill.exec.dotdrill;
 
 import java.io.IOException;
+import java.io.FileNotFoundException;
 import java.util.List;
+import java.util.Arrays;
+import java.util.ArrayList;
 
 import org.apache.drill.exec.store.dfs.DrillFileSystem;
 import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.GlobPattern;
 import org.apache.hadoop.fs.Path;
 
 import com.google.common.collect.Lists;
@@ -29,7 +33,15 @@ import com.google.common.collect.Lists;
 public class DotDrillUtil {
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DotDrillUtil.class);
 
-  private static List<DotDrillFile> getDrillFiles(DrillFileSystem fs, FileStatus[] statuses, DotDrillType... types){
+  /**
+   * Returns List of DotDrillFile objects for given list of FileStatus objects matching the given Dot Drill File Types.
+   * Return an empty list if no FileStatus matches the given Dot Drill File Types.
+   * @param fs DrillFileSystem instance
+   * @param statuses List of FileStatus objects
+   * @param types Dot Drill Types to be matched
+   * @return List of matched DotDrillFile objects
+   */
+  private static List<DotDrillFile> getDrillFiles(DrillFileSystem fs, List<FileStatus> statuses, DotDrillType... types){
     List<DotDrillFile> files = Lists.newArrayList();
     for(FileStatus s : statuses){
       DotDrillFile f = DotDrillFile.create(fs, s);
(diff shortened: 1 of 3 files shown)


HanumathRao:
@kr-arjun There seems to be some test failures. Can you verify if these test failures are because of your latest code.

kr-arjun:
@HanumathRao I have fixed the test failures. Could you please review now.

priteshm:
@HanumathRao can you complete the review today?

exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillUtil.java
```diff
@@ -48,16 +59,70 @@
...
+  private static List<FileStatus> getDrillFileStatus(DrillFileSystem fs, Path root, String name, DotDrillType... types) throws IOException{
+    List<FileStatus> statusList = new ArrayList<FileStatus>();
+
+    if (name.endsWith(".drill")) {
+      FileStatus[] status = fs.globStatus(new Path(root, name));
+      if (status != null) {
+        statusList.addAll(Arrays.asList(status));
+      }
+    } else {
+      DotDrillType[] typeArray;
```
HanumathRao: Instead of typeArray you may want  to use "types".

exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillUtil.java
```diff
@@ -48,16 +59,70 @@
...
+   * @param name name/pattern of the file
+   * @param types Dot Drill Types to be matched
+   * @return List of FileStatuses for files matching name and  Drill Dot file types.
+   * @throws IOException  if any I/O error occurs when fetching file status
+   */
+  private static List<FileStatus> getDrillFileStatus(DrillFileSystem fs, Path root, String name, DotDrillType... types) throws IOException{
+    List<FileStatus> statusList = new ArrayList<FileStatus>();
+
+    if (name.endsWith(".drill")) {
+      FileStatus[] status = fs.globStatus(new Path(root, name));
```
HanumathRao: Why do you want to change the existing logic for this case?

HanumathRao: In this case does it mean that types should not matter? If so is it good to have some assert.

exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillUtil.java
```diff
@@ -48,16 +59,70 @@
...
+   * Return an empty list if no files matches the pattern and Drill Dot file types.
+   * @param fs DrillFileSystem instance
+   * @param root parent Path
+   * @param name name/pattern of the file
+   * @param types Dot Drill Types to be matched
+   * @return List of FileStatuses for files matching name and  Drill Dot file types.
+   * @throws IOException  if any I/O error occurs when fetching file status
+   */
+  private static List<FileStatus> getDrillFileStatus(DrillFileSystem fs, Path root, String name, DotDrillType... types) throws IOException{
+    List<FileStatus> statusList = new ArrayList<FileStatus>();
```
HanumathRao: filesStatus does it make sense?

exec/java-exec/src/test/java/org/apache/drill/exec/dotdrill/TestDotDrillUtil.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class TestDotDrillUtil {
+
+  private static final String TEST_ROOT_DIR = System.getProperty("java.io.tmpdir") + "/dot_drill_util";
```
HanumathRao: I think you may want to look at some existing tests to inherit baseclass like BaseTestQuery. This will make sure that the test directories are cleaned up properly.

@ilooner  Can you please take a look at the test file.

kr-arjun:
@HanumathRao  I have modified variable names as per review comments. I didn't include assert check for types length while checking name ending with '.drill' as it is causing test failures. This assertion check fails for the scenario where complete view name is used in querying as in below test.

https://github.com/apache/drill/blob/master/exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java#L748

 For '*.drill' files, the type check is being done later in the getDrillFiles method. I have modified the method comments to indicate this behavior that types checking is done only for case name does not end with .drill. Please let me know if this is okay.

exec/java-exec/src/test/java/org/apache/drill/exec/dotdrill/TestDotDrillUtil.java
kr-arjun: @ilooner Could you please review the Test file. This test scenario is more related to FileSystem related tests. Please let me know if there are any existing base class that can be used here.

HanumathRao:
@kr-arjun Thanks for the changes. Code changes look good to me.

kr-arjun:
Thanks, @HanumathRao . @ilooner  Could you please review these changes?

priteshm:
@ilooner can you please review this as well?

kr-arjun:
@ilooner I found a possible issue with current logic. It may fail if the path has a wildcard ( case where not ending with .drill. In this case list status may fail). I am pushing the changes for handling this scenario. Thanks!

kr-arjun:
@ilooner  I have pushed the latest changes. Could you please review. Thanks!

ilooner:
Sorry for the delayed review @kr-arjun. Please use BaseDirTestWatcher for creating temp directories in tests. There is information in the javadoc for the class, an example in ExampleTest, and you can take a look at the other tests that use it for more examples.

arjun-rajan:
@ilooner  I have made the changes as per the review comment. Could you please review. Thanks!

kr-arjun:
@ilooner  I have made the changes as per the review comment. Could you please review. Thanks!

arjun-rajan:
@ilooner Could you please review the changes.

arina-ielchiieva:
@kr-arjun please squash commits into one.

kr-arjun:
@arina-ielchiieva I have squashed the commits into one. Thanks!

sohami:
@kr-arjun - Please make the PR and JIRA title same.

kr-arjun:
@sohami I have corrected the PR title. Thanks!

PR #1406, opened 2018-07-27
DRILL-6641: Fix columnValueCounts in ParquetGroupScanStatistics when ParquetGroupScan has RowGroupInfo without column statistics

vvysotskyi:


vvysotskyi:
@arina-ielchiieva, could you please review this PR?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScanStatistics.java
```diff
@@ -88,19 +88,21 @@ public void collect(List<RowGroupInfo> rowGroupInfos, ParquetTableMetadataBase p
       for (ColumnMetadata column : rowGroup.getColumns()) {
         SchemaPath schemaPath = SchemaPath.getCompoundPath(column.getName());
         Long previousCount = columnValueCounts.get(schemaPath);
+        Long newCount;
```
arina-ielchiieva: @vvysotskyi definitely case worth fixing! 
Maybe we can optimize `if (previousCount != null)` to something like `prevoiusCount == null ? 0 : previousCount`? This would help to remove else part and duplicated logic.

vvysotskyi: Thanks for the advice, with this change the code looks prettier!

arina-ielchiieva:
+1, LGTM.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScanStatistics.java
```diff
@@ -88,19 +88,14 @@ public void collect(List<RowGroupInfo> rowGroupInfos, ParquetTableMetadataBase p
       for (ColumnMetadata column : rowGroup.getColumns()) {
         SchemaPath schemaPath = SchemaPath.getCompoundPath(column.getName());
         Long previousCount = columnValueCounts.get(schemaPath);
-        if (previousCount != null) {
-          if (previousCount != GroupScan.NO_COLUMN_STATS && column.isNumNullsSet()) {
-            Long newCount = rowCount - column.getNulls();
-            columnValueCounts.put(schemaPath, columnValueCounts.get(schemaPath) + newCount);
-          }
+        previousCount = previousCount == null ? 0 : previousCount;
```
vrozov: Unnecessary boxing/unboxing and hash lookups. Consider
```
        long newCount = GroupScan.NO_COLUMN_STATS;
        if (column.isNumNullsSet()) {
          newCount = rowCount - column.getNulls();
          Long previousCount = columnValueCounts.get(schemaPath);
          if (previousCount != null) {
            if (GroupScan.NO_COLUMN_STATS != previousCount) {
              newCount += previousCount;
            } else {
              continue;
            }
          }
        }
        columnValueCounts.put(schemaPath, newCount);
```

vvysotskyi: @vrozov, thanks for taking a look at this change. I think we should leave this code as it is for the next reasons:
- boxing of `0` is happened only once;
- unboxing happens in both cases: when `+=` is produced for your case or when `+` is done.
- `continue` in the `else` block will cause problems for detecting that column is partitioned (see usage of `first` local variable);
- existing code looks clearer, since the logic is just splitted only into two branches.

vrozov: @vvysotskyi I mentioned not only boxing/unboxing but also hash lookup. In any case, I have a better proposal to use `MutableLong` or `AtomicLong`:
```
        final MutableLong emptyCount = new MutableLong();
        MutableLong count = columnValueCounts.putIfAbsent(schemaPath, emptyCount);
        if (count == null) {
          count = emptyCount;
        }
        if (count.longValue() != GroupScan.NO_COLUMN_STATS && !column.isNumNullsSet()) {
          count.add(rowCount - column.getNulls());
        } else {
          count.setValue(GroupScan.NO_COLUMN_STATS);
        }
```

vvysotskyi: @vrozov, changed the code to use `MutableLong`.

I appreciate the proposal to improve my code, but this optimization does not bring an essential performance increase. The previous code was not so bad and indeed it was better than before the changes and the main thing that it fixes the bug without breaking anything.

During the time we have this conversation, we might fix more bugs or do more useful things for the project. In the future, I would recommend concentrating on the approach, checking that the code will still work as expected, rather than on such minor things.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScanStatistics.java
```diff
@@ -87,19 +91,14 @@ public void collect(List<RowGroupInfo> rowGroupInfos, ParquetTableMetadataBase p
...
         SchemaPath schemaPath = SchemaPath.getCompoundPath(column.getName());
-        Long previousCount = columnValueCounts.get(schemaPath);
-        if (previousCount != null) {
-          if (previousCount != GroupScan.NO_COLUMN_STATS && column.isNumNullsSet()) {
-            Long newCount = rowCount - column.getNulls();
-            columnValueCounts.put(schemaPath, columnValueCounts.get(schemaPath) + newCount);
-          }
+        MutableLong emptyCount = new MutableLong();
+        MutableLong previousCount =
+            Optional.ofNullable(columnValueCounts.putIfAbsent(schemaPath, emptyCount))
```
vrozov: Usage of `Optional` is not justified here (a fancy way to avoid explicit comparison to `null` that leads to new object allocation).

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScanStatistics.java
vrozov: @vvysotskyi Following your approach "that the code will still work as expected", code style issues should never be raised as a concern during PR review, so I disagree with your recommendation. In this particular case, usage of `Long` and unnecessary hash lookups catch my eye the same way how an inconsistently formatted code does not look good to everyone.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScanStatistics.java
vvysotskyi: Thanks, replaced with `if (previousCount == null)`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScanStatistics.java
```diff
@@ -61,7 +63,9 @@ public ParquetGroupScanStatistics(ParquetGroupScanStatistics that) {
   }
 
   public long getColumnValueCount(SchemaPath column) {
-    return columnValueCounts.containsKey(column) ? columnValueCounts.get(column) : 0;
+    return Optional.ofNullable(columnValueCounts.get(column))
```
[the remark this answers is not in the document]
vvysotskyi: Thanks, done.

PR #1408, opened 2018-07-30
DRILL-6453: Resolve deadlock when reading from build and probe sides simultaneously in HashJoin

ilooner:
# The Problem

Originally hash join sniffed the first data holding batch in the probe and build side. Using the size statistics from both sides, memory calculations were performed in order to determine when to spill data. 

The issue with this is that fetching the first data holding batch from both sides can cause a deadlock in the exchange operators. Details of how this can happen have been included by others on the Jira.

# Theory of Operation

## Sniffing Batches

Batch sniffing is done in three phases.

 1. Schema sniffing is done in buildSchema()
 2. Before executing the build phase we sniff the first data holding build side batch and use the stats to decide the number of partitions and do memory calculations.
 3. Before executing the probe phase we sniff the first data holding probe side batch, and use the size statistics to do memory calculations that decide when to spill.

## Memory Estimation

When sniffing the schema for the build and probe side, we may get lucky and get data for the probe side. If this is the case then we cause use the probe side data to estimate the optimal number of partitions to use in the join operator. If we don't have probe side data when computing the number of partitions to use we assume that the incoming probe batches will be less than or equal to the configured batch size.

Since the number of partitions must be configured upfront before we may have probe data, we may get stuck in a situation where we have too many partitions to effectively process the probe side. In order to avoid this scenario we also adjust the number of records in probe side partition batches after we receive data from the probe side.

## Corner cases

While implementing this many corner cases had to be handled.

- Empty build side
- Empty probe side
- Empty probe and build sides
- Getting probe side data when retrieving the probe schema
- Not getting probe side data when retrieving the probe schema

## Testing

I added unit tests for all the corner cases, and have extracted logic for predicting incoming and partition batch sizes into BatchSizePredictorImpl. In unit tests various corner cases are tested by providing mock implementation of BatchSizePredictorImpl.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/BatchSizePredictor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/BatchSizePredictor.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.join;
+
+import org.apache.drill.exec.record.RecordBatch;
+
+/**
+ * This class predicts the sizes of batches given an input batch.
+ *
+ * <h4>Invariants</h4>
+ * <ul>
+ *   <li>The {@link BatchSizePredictor} assumes that a {@link RecordBatch} is in a state where it can return a valid record count.</li>
+ * </ul>
+ */
+public interface BatchSizePredictor {
+  /**
+   * Gets the batchSize computed in the call to {@link #updateStats()}. Returns 0 if {@link #hadDataLastTime()} is false.
+   * @return Gets the batchSize computed in the call to {@link #updateStats()}. Returns 0 if {@link #hadDataLastTime()} is false.
+   * @throws IllegalStateException if {@link #updateStats()} was never called.
+   */
+  long getBatchSize();
+
+  /**
+   * Gets the number of records computed in the call to {@link #updateStats()}. Returns 0 if {@link #hadDataLastTime()} is false.
+   * @return Gets the number of records computed in the call to {@link #updateStats()}. Returns 0 if {@link #hadDataLastTime()} is false.
+   * @throws IllegalStateException if {@link #updateStats()} was never called.
+   */
+  int getNumRecords();
+
+  /**
+   * True if the input batch had records in the last call to {@link #updateStats()}. False otherwise.
+   * @return True if the input batch had records in the last call to {@link #updateStats()}. False otherwise.
+   */
+  boolean hadDataLastTime();
+
+  /**
+   * This method can be called multiple times to collect stats about the latest data in the provided record batch. These
+   * stats are used to predict batch sizes. If the batch currently has no data, this method is a noop. This method must be
+   * called at least once before {@link #predictBatchSize(int, boolean)}.
+   */
+  void updateStats();
+
+  /**
+   * Predicts the size of a batch using the current collected stats.
+   * @param desiredNumRecords The number of records contained in the batch whose size we want to predict.
+   * @param reserveHash Whether or not to include a column containing hash values.
+   * @return The size of the predicted batch.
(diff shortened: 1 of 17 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -381,16 +409,14 @@ public HashJoinMemoryCalculator getCalculatorImpl() {
...
+    if (!prefetchedBuild) {
       // If we didn't retrieve our first data hold batch, we need to do it now.
-      prefetched = true;
-      prefetchFirstBatchFromBothSides();
+      prefetchedBuild = true;
+      prefetchFirstBuildBatch();
 
       // Handle emitting the correct outcome for termination conditions
-      // Use the state set by prefetchFirstBatchFromBothSides to emit the correct termination outcome.
+      // Use the state set by prefetchFirstBuildBatch to emit the correct termination outcome.
```
Ben-Zvi: Code cleaning: The check below (switch statement) can be done as part of the identical check after executeBuildPhase() (which is skipped anyway if STOP).  Also maybe move the "wasKilled" check first thing in innerNext().

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -248,32 +254,54 @@ protected void buildSchema() throws SchemaChangeException {
...
       // Got our first batche(s)
       state = BatchState.FIRST;
+    }
+  }
+
+  /**
+   *
+   * @return True terminate. False continue.
+   */
+  private boolean prefetchFirstProbeBatch() {
```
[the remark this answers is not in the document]
ilooner: Refactored this code.

ilooner:
@Ben-Zvi I've refactored the code, should be much cleaner now. Please take another look.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -248,95 +257,134 @@ protected void buildSchema() throws SchemaChangeException {
     }
   }
```
Ben-Zvi: It would be nicer if *buildSchema()* was calling the "standard" *prefetchFirstBatchFromBothSides()* like the other binary operators (sans MJ). Instead of having specialized different code, and modifying the RecordBatch just for the hash join.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -248,95 +257,134 @@ protected void buildSchema() throws SchemaChangeException {
...
+      });
+
+    // Left and right sides must return a valid response and both sides cannot be NONE.
+    return (!leftUpstream.isError() && !rightUpstream.isError()) &&
+      (leftUpstream != IterOutcome.NONE && rightUpstream != IterOutcome.NONE);
+  }
+
+  private IterOutcome sniffNewSchema(final int index,
+                                     final RecordBatch batch,
+                                     final Runnable schemaSetter) {
```
Ben-Zvi: The use of lambda function as a parameter just makes this code harder to read, and adds no real value.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinMemoryCalculatorImpl.java
```diff
@@ -365,17 +299,20 @@ protected void initialize(boolean autoTune,
                               double loadFactor) {
       Preconditions.checkState(!firstInitialized);
       Preconditions.checkArgument(initialPartitions >= 1);
+      // If we had probe data before there should still be probe data now.
+      // If we didn't have probe data before we could get some new data now.
+      Preconditions.checkState(probeSizePredictor.hasData() && !probeEmpty || !probeSizePredictor.hasData());
```
Ben-Zvi: A simpler rewrite of the condition:  **! (probeEmpty && ! probeSizePredictor.hasData())**

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/BatchSizePredictor.java
```diff
@@ -0,0 +1,79 @@
...
+   * @return Gets the number of records computed in the call to {@link #updateStats()}. Returns 0 if {@link #hasData()} is false.
+   * @throws IllegalStateException if {@link #updateStats()} was never called.
+   */
+  int getNumRecords();
+
+  /**
+   * True if the input batch had records in the last call to {@link #updateStats()}. False otherwise.
+   * @return True if the input batch had records in the last call to {@link #updateStats()}. False otherwise.
+   */
+  boolean hasData();
```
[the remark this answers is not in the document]
ilooner: Renamed to hadDataLastTime

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinMemoryCalculatorImpl.java
ilooner: Done

ilooner:
@Ben-Zvi Thanks for the review and suggestions. The code is much cleaner now.

Ben-Zvi:
OK - tag it ready-to-commit ...

PR #1414, opened 2018-08-01
DRILL-6101: Optimized implicit columns handling within scanner

sachouche:
Problem Description -

File based implicit columns are projected only if explicitly requested within the query
Note that Partition Columns are not included in this discussion (only referring about FILENAME, FILEPATH, FQN, and SUFFIX)
The scanner operator is called with three sets of columns to handle: Table Columns, Partition Columns, and Implicit Columns
When a SELECT_STAR is used, the operator doesn't receive the original query selection (only '**' is received)
This behavior mandates that the Scanner operator projects all file based Implicit Columns only for these to be filtered out later on by the Project Operator
Performance tests indicates this behavior introduces a 30% degradation within the scanning phase for some TPCH queries (this degradation is larger for tables with long paths)
Fix -

Noticed the code uses a Utility to figure out whether a selection is a STAR_QUERY; this utility expects a list of columns and attempts to detect the presence of the STAR selection keyword
Modified the code to include all selection columns (including the ones in the where clause)
This allowed the execution layer to invoke the Scan operator with the correct implicit columns (the ones explicitly listed within the query) and thus addressing this performance issue
Note that readers are not impacted with the newly added metadata as the reader code doesn't use the columns list when a STAR_QUERY is involved

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillScanRel.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillScanRel.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.planner.logical;
 
+import java.util.ArrayList;
 import java.io.IOException;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


sachouche:
@HanumathRao and @ilooner can you please review this PR.

sachouche:
@ilooner fixed the formatting issue; thanks for the review!

paul-rogers:
As it turns out, optimized processing for implicit columns is available in the long-stalled result set loader pull requests. That code grabs only the data columns (which it packs optimally into a result set), then adds implicit columns in a later step, avoiding the thrashing that can occur if the implicit columns are populated concurrent with incoming data.

What is really needed is for the planner to recognize the implicit columns and project them separately from the data. That is `SELECT *, filename` should result in projecting `*` from the table, and only the `filename` implicit column. Else, the best we can do is generate the columns, then throw them away, which is obviously suboptimal.

HanumathRao:
@sachouche  code changes look good to me.

sachouche:
@paul-rogers 
I agree the planner should have been responsible for optimally exposing the implicit & partition columns since it has all the necessary metadata. This feature was not implemented that way.. The execution layer included logic to optimize implicit column handling if the query was not a STAR_QUERY. My fix was mainly to pass the missing metadata in the case of STAR_QUERY and thus trigger the existing optimization. Our testing showed up to 30% improvement within the Parquet scanner for SELECT_STAR queries and big reduction in batch memory utilization.

ilooner:
@paul-rogers I suggest we go with this change for now since the code change is minimal, needed by some of our users, and provides a big speed up. I see your result set loader PRs are blocked on testing. I will be your personal tester and try to help you get those PRs in. When those PRs are ready we can switch to using your comprehensive solution.

PR #1419, opened 2018-08-03
DRILL-6662: Access AWS access key ID and secret access key using Cred…

KazydubB:
…ential Provider API for S3 storage plugin

### DIFF
--- a/distribution/src/resources/core-site-example.xml
+++ b/distribution/src/resources/core-site-example.xml
@@ -30,4 +30,18 @@
         <value>ENTER_YOUR_SECRETKEY</value>
     </property>
 
+    <!--Use this property to specify one or more credential provider URIs instead of
+        configuring above credentials in plain text-->
+    <!--<property>
+        <name>hadoop.security.credential.provider.path</name>
+        <value>ENTER_YOUR_PROVIDER_PATH</value>
+    </property>-->
+
+    <!--Set this property to true to avoid caching of S3 file system configuration properties,
+        so when you add/update a property (e.g. fs.s3a.secret.key) in S3 storage plugin its new value will be taken-->
+    <!--<property>
+        <name>fs.s3a.impl.disable.cache</name>
+        <value>true</value>
+    </property>-->
+
 </configuration>
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
@@ -20,6 +20,7 @@ package org.apache.drill.exec.store.dfs;
 import static org.apache.drill.exec.store.dfs.FileSystemSchemaFactory.DEFAULT_WS_NAME;
 
 import java.io.IOException;
+import java.net.URI;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
```diff
@@ -104,6 +108,31 @@ public FileSystemPlugin(FileSystemConfig config, DrillbitContext context, String
...
+   * is absent in providers, it will conditionally fallback to configuration setting. The fallback will occur unless
+   * {@link org.apache.hadoop.security.alias.CredentialProvider#CLEAR_TEXT_FALLBACK} is set to false.
+   * @throws IOException thrown if a credential cannot be retrieved from provider
+   */
+  private void handleS3Credentials() throws IOException {
+    final String[] credentialKeys = {"fs.s3a.secret.key", "fs.s3a.access.key"};
+    for (String key : credentialKeys) {
+      char[] credentialChars = fsConf.getPassword(key);
+      if (credentialChars != null) {
+        fsConf.set(key, String.valueOf(credentialChars));
```
steveloughran: s3a connector actually calls trim() on the created string, so that you can have some whitespace around any inline declaration. Worth retaining for consistency

KazydubB: Hm, what do you call an "inline declaration"? In s3 storage plugin and with hadoop credential CLI's -value (though this option is for testing because it's unsecure) the value should be wrapped with quotes if it is designed to contain spaces. If set in core-site.xml, the value is wrapped with tags, so any space is considered to be intentional.

distribution/src/resources/core-site-example.xml
```diff
@@ -30,4 +30,14 @@
         <value>ENTER_YOUR_SECRETKEY</value>
     </property>
 
+    <property>
```
arina-ielchiieva: Please comment out this section and add comment explaining that user should use one of those.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
```diff
@@ -104,6 +109,33 @@ public FileSystemPlugin(FileSystemConfig config, DrillbitContext context, String
...
+   * {@link org.apache.hadoop.security.alias.CredentialProviderFactory#CREDENTIAL_PROVIDER_PATH} property)
+   * credential providers and set it into {@code conf}. If provider path is not configured or credential
+   * is absent in providers, it will conditionally fallback to configuration setting. The fallback will occur unless
+   * {@link org.apache.hadoop.security.alias.CredentialProvider#CLEAR_TEXT_FALLBACK} is set to {@code false}.
+   *
+   * @param conf {@code Configuration} which will be updated with credentials from provider
+   * @throws IOException thrown if a credential cannot be retrieved from provider
+   */
+  private void handleS3Credentials(Configuration conf) throws IOException {
+    String[] credentialKeys = {"fs.s3a.secret.key", "fs.s3a.access.key"};
```
arina-ielchiieva: Consider using org.apache.hadoop.fs.s3a.Contants class.

KazydubB: I am aware of the Constants, but artifact (hadoop-aws), containing this class is not among the module's dependencies (however it is present in distribution's (compile-scope) and drill-root's (test-scope) dependencies). Is there a need to add the dependency?

PR #1420, opened 2018-08-03
DRILL-6664: Limit the maximum parquet reader batch rows to 64k

sachouche:
Changed the internal configuration of the Parquet reader to cap the maximum number of rows per batch to 64k. This is precautionary as downstream operators expect batches of no more than 64k rows.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -34,6 +34,7 @@ import org.apache.drill.exec.server.options.TypeValidators.StringValidator;
 import org.apache.drill.exec.server.options.TypeValidators.AdminUsersValidator;
 import org.apache.drill.exec.server.options.TypeValidators.AdminUserGroupsValidator;
 import org.apache.drill.exec.testing.ExecutionControls;
+import org.apache.drill.exec.vector.ValueVector;
 
 public final class ExecConstants {
   private ExecConstants() {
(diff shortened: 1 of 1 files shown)


sachouche:
@ilooner or @Ben-Zvi can you please review this PR?

Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -322,7 +322,7 @@ private ExecConstants() {
 
   // Controls the flat parquet reader batching constraints (number of record and memory limit)
   public static final String PARQUET_FLAT_BATCH_NUM_RECORDS = "store.parquet.flat.batch.num_records";
-  public static final OptionValidator PARQUET_FLAT_BATCH_NUM_RECORDS_VALIDATOR = new RangeLongValidator(PARQUET_FLAT_BATCH_NUM_RECORDS, 1, Integer.MAX_VALUE);
+  public static final OptionValidator PARQUET_FLAT_BATCH_NUM_RECORDS_VALIDATOR = new RangeLongValidator(PARQUET_FLAT_BATCH_NUM_RECORDS, 1, 65535);
```
Ben-Zvi: Should the number be the actual power of 2 -- 2^16 ?  That is - 65536  , without the "minus one" ....

ilooner: I was thinking the same thing, so I double checked and the RangeValidator uses an inclusive range. So I think 65535 is correct.

Ben-Zvi: This figure is used to set the `maxRecordsPerBatch` in `RecordBatchSizerManager`; hence every batch sent downstream from the Parquet reader would have at most 64K-1 records. What limit do other operators use ?  
  Though a minor space waste, going "minus one" does have some benefit for variable length columns (see DRILL-5446), where the offsets vector need not double.

sachouche: Yes, you are correct. VL columns use an offset vector which always requires an extra entry; limiting the number of rows to 2^^N-1 guarantees the offset vector will not double in size when the last entry is added.

sachouche: Thanks @Ben-Zvi and @ilooner for the review!

ilooner: Yeah I think your right. I was getting confused between max count and max index value. The Max batch record count used elsewhere is indeed 2^16 as you said. Thx :)

ilooner:
@sachouche spoke with @Ben-Zvi offline. Some operators use 2^16 -1 to avoid offset vector doubling while others don't care. It would be good to decide on standard number to use across the board. Could you guys discuss briefly? I will temporarily remove the ready-to-commit label for now.

paul-rogers:
`ValueVector.MAX_ROW_COUNT` is what you want to use. Easier to remember than repeating the magic number everywhere.

sachouche:
*What Parquet Used to do*
- The parquet reader used to hardcode this information along with a comment:
DEFAULT_RECORDS_TO_READ_IF_FIXED_WIDTH = 64*1024 - 1; // 64K - 1, max SV2 can address
- Unfortunately, the SelectionVector2 only specifies "64k" as part of a comment (no mention of 64k-1)

*Memory Optimization*
- The memory optimization for a VL column can save 64k memory + avoids a reset and copy of the offset vector
- Fixed length columns would wait few bytes of space since the last entry is unoccupied
- Is this optimization super important? I would say no, as the current default batch memory is set to 16MB; in practice record batches will have less than 64k rows when VL columns are involved.

*Conclusion*
- I like @paul-rogers suggestion to use the ValueVector.MAX_ROW_COUNT as it satisfies the goal of this JIRA (64k) and brings us one step further to standardization. I'll update the changes shortly.

Thanks for the feedback!

sachouche:
@ilooner I have updated the PR with @paul-rogers suggestion.

PR #1426, opened 2018-08-07
DRILL-6671: Multi level lateral unnest join is throwing an exception …

HanumathRao:
…during materializing the plan.

This PR includes changes to correctly attach the concerned UnnestPOP to LateralJoinPOP. 
The logic is to push the UnnestPOP as we are visiting the lateral join nodes and pop the unnest pop and attach it to the lateraljoin. 

 @gparai  @sohami  Can you please review this change.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/Materializer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/Materializer.java
@@ -18,6 +18,7 @@
 package org.apache.drill.exec.planner.fragment;
 
 import java.util.List;
+import java.util.Stack;
 
 import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.exec.exception.FragmentSetupException;
@@ -116,14 +117,13 @@ public class Materializer extends AbstractPhysicalVisitor<PhysicalOperator, Mate
 
     children.add(op.getLeft().accept(this, iNode));
     children.add(op.getRight().accept(this, iNode));
-    UnnestPOP unnestInLeftInput = iNode.getUnnest();
+    UnnestPOP unnestForThisLateral = iNode.getUnnest();
 
     PhysicalOperator newOp = op.getNewWithChildren(children);
     newOp.setCost(op.getCost());
     newOp.setOperatorId(Short.MAX_VALUE & op.getOperatorId());
 
-    ((LateralJoinPOP)newOp).setUnnestForLateralJoin(unnestInLeftInput);
-
+    ((LateralJoinPOP)newOp).setUnnestForLateralJoin(unnestForThisLateral);
     return newOp;
   }
 
@@ -138,7 +138,7 @@ public class Materializer extends AbstractPhysicalVisitor<PhysicalOperator, Mate
     final Wrapper info;
     final int minorFragmentId;
 
-    UnnestPOP unnest = null;
+    Stack<UnnestPOP> unnest = new Stack<>();
 
     public IndexedFragmentNode(int minorFragmentId, Wrapper info) {
       super();
@@ -163,11 +163,11 @@ public class Materializer extends AbstractPhysicalVisitor<PhysicalOperator, Mate
     }
 
     public void addUnnest(UnnestPOP unnest) {
-      this.unnest = unnest;
+      this.unnest.push(unnest);
     }
 
     public UnnestPOP getUnnest() {
-      return this.unnest;
+      return this.unnest.pop();
     }
 
   }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/Materializer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/Materializer.java
@@ -17,8 +17,9 @@
  */
 package org.apache.drill.exec.planner.fragment;
 
+import java.util.ArrayDeque;
+import java.util.Deque;
 import java.util.List;
-import java.util.Stack;
 
 import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.exec.exception.FragmentSetupException;
@@ -138,7 +139,7 @@ public class Materializer extends AbstractPhysicalVisitor<PhysicalOperator, Mate
     final Wrapper info;
     final int minorFragmentId;
 
-    Stack<UnnestPOP> unnest = new Stack<>();
+    private final Deque<UnnestPOP> unnest = new ArrayDeque<>();
 
     public IndexedFragmentNode(int minorFragmentId, Wrapper info) {
       super();
@@ -163,11 +164,11 @@ public class Materializer extends AbstractPhysicalVisitor<PhysicalOperator, Mate
     }
 
     public void addUnnest(UnnestPOP unnest) {
-      this.unnest.push(unnest);
+      this.unnest.addFirst(unnest);
     }
 
     public UnnestPOP getUnnest() {
-      return this.unnest.pop();
+      return this.unnest.removeFirst();
     }
 
   }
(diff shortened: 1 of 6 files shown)


HanumathRao:
@vvysotskyi  Please can you help review this change.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
```diff
@@ -169,13 +169,31 @@ public void testLeftLateral_WithFilterAndLimitInSubQuery() throws Exception {
   @Test
   public void testMultiUnnestAtSameLevel() throws Exception {
     String Sql = "EXPLAIN PLAN FOR SELECT customer.c_name, customer.c_address, U1.order_id, U1.order_amt," +
```
vvysotskyi: If this fix allows execution for the query from the test, I think we should remove `EXPLAIN PLAN FOR` and check the result.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/Materializer.java
```diff
@@ -138,7 +138,7 @@ public PhysicalOperator visitUnnest(UnnestPOP unnest, IndexedFragmentNode value)
     final Wrapper info;
     final int minorFragmentId;
 
-    UnnestPOP unnest = null;
+    Stack<UnnestPOP> unnest = new Stack<>();
```
vvysotskyi: Could you please replace `Stack` usage by `Deque` and `ArrayDeque` and make fields private final

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/LateralJoinPrel.java
```diff
@@ -48,7 +48,7 @@
 public class LateralJoinPrel extends DrillLateralJoinRelBase implements Prel {
 
 
-  protected LateralJoinPrel(RelOptCluster cluster, RelTraitSet traits, RelNode left, RelNode right, boolean excludeCorrelateCol,
+  public LateralJoinPrel(RelOptCluster cluster, RelTraitSet traits, RelNode left, RelNode right, boolean excludeCorrelateCol,
```
vvysotskyi: I think there is no need to make this constructor public. In `LateralUnnestRowIDVisitor` may be used method `copy()` instead of using the constructor.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/LateralUnnestRowIDVisitor.java
```diff
@@ -64,7 +66,14 @@ public Prel visitLateral(LateralJoinPrel prel, Boolean value) throws RuntimeExce
     children.add(((Prel)prel.getInput(0)).accept(this, value));
     children.add(((Prel) prel.getInput(1)).accept(this, true));
 
-    return (Prel) prel.copy(prel.getTraitSet(), children);
+    if (!value) {
```
vvysotskyi: As I understand, `value` argument is used to specify that the current `prel` is the right input of the `LateralJoinPrel` if it is true, I recommend to rename it for better readability.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/LateralUnnestRowIDVisitor.java
```diff
@@ -64,7 +66,14 @@ public Prel visitLateral(LateralJoinPrel prel, Boolean value) throws RuntimeExce
...
+      return (Prel) prel.copy(prel.getTraitSet(), children);
+    } else {
+      CorrelationId corrId = new CorrelationId(prel.getCorrelationId().getId() + 1);
```
vvysotskyi: Passing increased `id` to `CorrelationId` may cause problems since inputs of `LateralJoinPrel` also may contain both `CorrelationId`s with old and new `id`s and it may cause problems for some queries. I would recommend receiving `CorrelationId` by calling `prel.getCluster().createCorrel()`, since it guarantees that the index will be unique.

Also, just creating and passing a new `CorrelationId` instance to the `LateralJoinPrel` is not enough, because as I mentioned above, inputs refer to it, so all these references should be replaced. As an example of this approach, you may use code from [ProjectCorrelateTransposeRule](https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/rel/rules/ProjectCorrelateTransposeRule.java#L124).

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/LateralUnnestRowIDVisitor.java
```diff
@@ -64,7 +66,14 @@ public Prel visitLateral(LateralJoinPrel prel, Boolean value) throws RuntimeExce
...
+      ImmutableBitSet requiredColumns = prel.getRequiredColumns().shift(1);
```
vvysotskyi: Could you please add a comment that shift is required for taking into account newly added `$drill_implicit_field$`

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
HanumathRao: @vvysotskyi This is an existing testcase. I just want to keep it intact. I cannot remove the EXPLAIN PLAN and use it for executing the query because this query will result in an execution error (unnest is not supported for non list types) and this is the right behaviour. For testing the usecase of this PR an additional testcase has been introduced which also tests for the correctness of the result.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/LateralUnnestRowIDVisitor.java
```diff
@@ -59,16 +65,37 @@ public Prel visitPrel(Prel prel, Boolean isRightOfLateral) throws RuntimeExcepti
...
+        requiredColsMap.put(corrColIndex, corrColIndex+1);
+      }
+      ImmutableBitSet requiredColumns = prel.getRequiredColumns().shift(1);
+
+      CorrelationId corrId = prel.getCluster().createCorrel();
+      RexCorrelVariable rexCorrel =
+              (RexCorrelVariable) prel.getCluster().getRexBuilder().makeCorrel(
+                      children.get(0).getRowType(),
+                      corrId);
+      RelNode rightChild = children.get(1).accept(
```
vvysotskyi: Looks like `ProjectCorrelateTransposeRule.RelNodesExprsHandler` does not replace RexCorrelVariable for the case when RelNode does not have any input. In our case, this is `UnnestPrel`. So we need to extend this class and override method `visit(RelNode other)`:
```
    @Override
    public RelNode visit(RelNode other) {
      return super.visit(other.accept(rexVisitor));
    }
```
Also, it does not work since `UnnestPrel` use `accept(RexShuttle shuttle)` method from `AbstractRelNode` which does nothing, so we should override it somehow like:
```
  @Override
  public RelNode accept(RexShuttle shuttle) {
    RexNode ref = shuttle.apply(this.ref);
    if (this.ref == ref) {
      return this;
    }
    return new UnnestPrel(getCluster(), traitSet, rowType, ref);
  }
```

HanumathRao:
@vvysotskyi I have made the changes.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/visitor/LateralUnnestRowIDVisitor.java
```diff
@@ -59,16 +66,55 @@ public Prel visitPrel(Prel prel, Boolean isRightOfLateral) throws RuntimeExcepti
...
+              new CorrelateVarReplacer(
+                      new ProjectCorrelateTransposeRule.RexFieldAccessReplacer(corrId,
```
vvysotskyi: The first argument in `RexFieldAccessReplacer` is a `CorrelationId` which should be replaced, but you have specified the new one. 
Please replace here `corrId` with `prel.getCorrelationId()`.

HanumathRao: Done.

PR #1427, opened 2018-08-08
DRILL-6674: Minor fixes to avoid auto boxing cost in logging in Later…

sohami:
…alJoinBatch


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/LateralJoinBatch.java
```diff
@@ -728,7 +730,6 @@ private void finalizeOutputContainer() {
 
     // Set the record count in the container
     container.setRecordCount(outputIndex);
-    container.buildSchema(BatchSchema.SelectionVectorMode.NONE);
```
parthchandra: Did you mean to remove this line?

PR #1428, opened 2018-08-09
DRILL-6670: align Parquet TIMESTAMP_MICROS logical type handling with earlier versions + minor fixes

okalinin:
### Background
Parquet dependencies were upgraded with DRILL-6353 which changed behavior in how Parquet handles `TIMESTAMP_MICROS` logical type. Previously, calling `SchemaElement.getConverted_type()` was returning null and logical type was ignored. With updated Parquet version above call returns `TIMESTAMP_MICROS` which triggers query exception.

Issue impacts both ParquetRecordReader and DrillParquetReader.

This change aims at restoring original functionality - handling `TIMESTAMP_MICROS` as `INT64` with no logical type in both Parquet readers. It doesn't seem to make sense to do more since `TIMESTAMP_MICROS` is deprecated logical type as per Parquet [current documentation](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md).

Additionally, PR fixes query failure for query on files with non-dictionary enabled encoding. 

### Change description
- Added `TIMESTAMP_MICROS` logical type in both Parquet readers in order to handle it as regular `INT64`
- Modified `ParquetSimpleTestFileGenerator` to include `TIMESTAMP_MICROS` columns in Parquet test files and updated existing test data files
- Modified respective tests with minor regrouping
- Fixed `TestDrillParquetReader` to make sure correct reader is actually used to handle test queries
- Improved debug lines to avoid using ambiguous 'new/old Parquet reader' terminology
- Fixed `ColumnReaderFactory` non-dictionary enabled file handing (it was missing a null check leading to query failure).
- Added Parquet tests on non-dictionary enabled files.
### Level of testing
- Sample file and query provided by issue reporter
- Unit tests


okalinin:
I will push another ParquetRecordReader fix for logical type handling on Parquet files with dictionary encoding disabled. Please do not review PR yet.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -108,7 +108,7 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
 
         if (!context.getOptions().getBoolean(ExecConstants.PARQUET_NEW_RECORD_READER)
             && !ParquetReaderUtility.containsComplexColumn(footer, rowGroupScan.getColumns())) {
-          logger.debug("Query {} qualifies for new Parquet reader",
+          logger.debug("Query {} qualifies for ParquetRecordReader",
```
vrozov: IMO, it will be much better to log `PARQUET_NEW_RECORD_READER` option and whether or not there are complex columns prior to creating a reader and reader instance class after the reader is created:
```
log.debug("PARQUET_NEW_RECORD_READER is {}. Complex columns {}.", "enabled"/"disabled", "found"/"not found");
Reader reader;
if (...) {
  reader = new ParquetRecordReader()l
} else {
  reader = new DrillParquetReader();
}
log.debug("Query {} uses {});
readers.add(reader);
```

okalinin: Thanks for the suggestion. Done.

okalinin: @vrozov I realised that your proposed wording sounds better. 'not enforced' may cause confusion too. Reverted back to 'enabled/disabled'.

okalinin:
Comment addressed, conflicts resolved. PR should be OK for further review.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet2/TestDrillParquetReader.java
```diff
@@ -33,20 +33,20 @@
 import org.junit.Test;
 
 public class TestDrillParquetReader extends BaseTestQuery {
-  // enable decimal data type
+  // enable decimal data type and make sure DrillParquetReader is used to handle test queries
   @BeforeClass
   public static void enableDecimalDataType() throws Exception {
     test(String.format("alter session set `%s` = true", PlannerSettings.ENABLE_DECIMAL_DATA_TYPE_KEY));
```
arina-ielchiieva: Please use alterSession / resetSession methods.

okalinin: Done.

arina-ielchiieva:
+1, LGTM.

PR #1430, opened 2018-08-10
DRILL-6680: Expose show files command into INFORMATION_SCHEMA

arina-ielchiieva:
Details in [DRILL-6680](https://issues.apache.org/jira/browse/DRILL-6680).


exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -605,4 +605,5 @@ drill.exec.options: {
     store.kafka.poll.timeout: 200,
     web.logs.max_lines: 10000,
     window.enable: true,
+    storage.list_files_recursively: true
```
Agirish: I would suggest we keep the default false for backward compatibility - if not, show files command output may explode for very high level directories, and may be less helpful for exploratory purposes.

Agirish: ah...Is this setting only applicable when querying the INFORMATION_SCHEMA.`FILES` table? And SHOW FILES command will continue to work as it does today? If so, sorry - please ignore my previous comment.

arina-ielchiieva: Sounds fair, made the changes. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/Records.java
```diff
@@ -543,4 +550,38 @@ public Schema(String catalog, String name, String owner, String type, boolean is
       this.IS_MUTABLE = isMutable ? "YES" : "NO";
     }
   }
-}
+
+  public static class File {
```
vvysotskyi: Could you please add Javadoc similar to Javadocs for other nested classes in this class.

arina-ielchiieva:
@vvysotskyi thanks for the code review, addressed CR comments.

PR #1431, opened 2018-08-14
DRILL-6687: Improve RemovingRecordBatch to do transfer when all recor…

sohami:
…ds needs to be copied

Add optimization in SelectionVector2 to enable RemovingRecordBatch to transfer ValueVectors from incoming to output container when all records needs to be copied.
Modified FilterRecordBatch and LimitRecordBatch to play by this optimization

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/record/selection/SelectionVector2.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/record/selection/SelectionVector2.java
@@ -36,7 +36,10 @@ import org.apache.drill.exec.record.DeadBuf;
 public class SelectionVector2 implements AutoCloseable {
 
   private final BufferAllocator allocator;
+  // Indicates number of indexes stored in the SV2 buffer which may be less than actual number of rows stored in
+  // RecordBatch container owning this SV2 instance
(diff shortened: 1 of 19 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/record/selection/SelectionVector2.java
```diff
@@ -37,6 +37,7 @@
 
   private final BufferAllocator allocator;
   private int recordCount;
+  private int batchActualRecordCount = -1;
```
parthchandra: put a comment to differentiate between recordCount and batchActualRecordCount

exec/java-exec/src/main/java/org/apache/drill/exec/record/selection/SelectionVector2.java
```diff
@@ -151,6 +159,18 @@ public void setRecordCount(int recordCount){
     this.recordCount = recordCount;
   }
 
+  public boolean doFullTransfer() {
```
parthchandra: doFullTransfer --> canDoFullTransfer

parthchandra:
+1

PR #1433, opened 2018-08-15
DRILL-6685: Fixed exception when reading Parquet data

sachouche:
Fixed a regression when addressing DRILL-6570: 
- When fixing DRILL-6570, we unified a bulk entry's max-values so that a false-positive (from fixed length to variable length) could happen smoothly
- The regression was that the fixed length algorithm was relying on the previous bulk-entry max-value constraint

Fix -
- I have re-introduced the constraint within the fixed-length reader
- Added a test-suite using @robert Hou parquet data to prevent such regressions


sachouche:
@Ben-Zvi and @ilooner can either of you review this PR?

Thanks!

sachouche:
@ilooner, I already brought down the size from 26MB to 1MB; what is an acceptable size?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/VarLenFixedEntryReader.java
```diff
@@ -43,7 +43,7 @@ final VarLenColumnBulkEntry getEntry(int valuesToRead) {
 
     final int expectedDataLen = columnPrecInfo.precision;
     final int entrySz = 4 + columnPrecInfo.precision;
-    final int readBatch = Math.min(entry.getMaxEntries(), valuesToRead);
+    final int readBatch = getFixedLengthMaxRecordsToRead(valuesToRead);
```
Ben-Zvi: Would it be nicer to pass the *entrySz* as a parameter to *getFixedLengthMaxRecordsToRead()* , instead of recreating it ?
  E.g., in case it ever changes. Or maybe it'll look clearer.

ilooner:
It would be ideal if we could get it down on the order of 10 kbs or less, or generate it.

sachouche:
- @Ben-Zvi , refactored the getFixedLengthMaxRecordsToRead() method to take the entrySz as input 
- @ilooner re-generated the file and now the size is ~100kb

sachouche:
Thank you @ilooner and @Ben-Zvi  for the review!

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetBulkReader.java
```diff
@@ -0,0 +1,115 @@
...
+
+import org.apache.drill.test.BaseTestQuery;
+import org.apache.drill.exec.ExecConstants;
+import org.apache.drill.exec.planner.physical.PlannerSettings;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+/** Tests the Parquet bulk reader */
+@SuppressWarnings("deprecation")
```
arina-ielchiieva: One minor comment: since `BaseTestQuery` is deprecated, I suppose we should try to write unit tests using `ClusterTest` rather then suppression the warning.

sachouche: @arina-ielchiieva,
- Majority of the tests are using BaseTestQuery; I even found existing tests easier to write than with ClusterTest (setting session flags, no need to define upfront schema, result-set validation). My guess, we should create a JIRA to move all existing BaseTestQuery to ClusterTest so that DEV get a chance to get more familiar with this test harness.
- I use Eclipse; the UI shows warnings all over the place
- Suppressing the deprecation allows me to focus on more important warnings

sachouche: @arina-ielchiieva 
Thank you for the info and the sample test! that really helped.. I have now ported the new Parquet test to using the ClusterTest base class.

sachouche:
@arina-ielchiieva,
- Agreed, but I liked better the pattern used in the sample test (that you have sent) as it provides more flexibility
- Each test can invoke separate flags without increasing complexity
- Combining global settings along with overrides can become tricky (an override needs to ensure that all global settings are restituted to their original state after the test finishes)

arina-ielchiieva:
@sachouche I don't have strong opinion of that, so PR can go as is, though since all your tests using the same options, before / after could save you from logic duplication.

PR #1435, opened 2018-08-17
Parquet compression is disabled by default

diegocaro:
Compression on parquet files are disabled by default, check line 577 on drill-module.conf ->
https://github.com/apache/drill/blob/master/exec/java-exec/src/main/resources/drill-module.conf#L577


arina-ielchiieva:
@bbevens could you please review?

bbevens:
I've updated the doc to change default type from snappy to none. This pull request can be closed. Thanks!

PR #1437, opened 2018-08-21
DRILL-6702: Disable CPU Reporting for non-HotSpot JDKs

kkhatua:
When running Drill on the IBM JDK (J9), the webUI throws an ClassCastException :
Caused by: java.lang.ClassCastException: com.ibm.lang.management.ExtendedOperatingSystem incompatible with com.sun.management.OperatingSystemMXBean

This PR simply disables that, since Drill should ideally be recompiled with these alternative JDKs.

### DIFF
--- a/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
@@ -40,14 +40,7 @@ public class CpuGaugeSet implements MetricSet {
   public CpuGaugeSet() {
     this.rtMXBean = ManagementFactory.getRuntimeMXBean();
     //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
-    OperatingSystemMXBean tempOSMXBean;
-    try {
-      tempOSMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
-    } catch (ClassCastException cce) {
-      logger.warn("{}. Detected non-Supported JVM [{}]. CPU Metrics in the WebUI will not be available!", cce.getMessage(), System.getProperty("java.vm.name"));
-      tempOSMXBean = null;
-    }
-    this.osMXBean = tempOSMXBean;
+    this.osMXBean = getOSMXBeanForCpuMetrics();
   }
 
   @Override
@@ -57,6 +50,15 @@ public class CpuGaugeSet implements MetricSet {
     metric.put("drillbit.load.avg", new DrillbitProcessLoad(osMXBean));
     metric.put("drillbit.uptime", new DrillbitUptime(rtMXBean));
     return metric;
+  }
+
+  private static OperatingSystemMXBean getOSMXBeanForCpuMetrics() {
+    try {
+      return (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    } catch (Exception ex) {
+      logger.warn("{}. Detected non-Supported JVM [{}]. CPU Metrics in the WebUI will not be available!", ex.getMessage(), System.getProperty("java.vm.name"));
+    }
+    return null;
   }
 }
 
@@ -72,7 +74,7 @@ final class OperatingSystemLoad implements Gauge<Double> {
 
   @Override
   public Double getValue() {
-    return (osMXBean != null) ? osMXBean.getSystemLoadAverage(): null;
+    return (osMXBean != null) ? osMXBean.getSystemLoadAverage() : null;
   }
 
 }
--- a/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
@@ -32,13 +32,22 @@ import com.sun.management.OperatingSystemMXBean;
  */
 @SuppressWarnings("restriction")
 public class CpuGaugeSet implements MetricSet {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CpuGaugeSet.class);
 
-  private OperatingSystemMXBean osMXBean;
-  private RuntimeMXBean rtMXBean;
+  private final OperatingSystemMXBean osMXBean;
+  private final RuntimeMXBean rtMXBean;
 
   public CpuGaugeSet() {
-    this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
     this.rtMXBean = ManagementFactory.getRuntimeMXBean();
+    //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
+    OperatingSystemMXBean tempOSMXBean;
+    try {
+      tempOSMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    } catch (ClassCastException cce) {
+      logger.warn("{}. Detected non-Supported JVM [{}]. CPU Metrics in the WebUI will not be available!", cce.getMessage(), System.getProperty("java.vm.name"));
+      tempOSMXBean = null;
+    }
+    this.osMXBean = tempOSMXBean;
   }
 
   @Override
@@ -63,7 +72,7 @@ final class OperatingSystemLoad implements Gauge<Double> {
 
   @Override
   public Double getValue() {
-    return osMXBean.getSystemLoadAverage();
+    return (osMXBean != null) ? osMXBean.getSystemLoadAverage(): null;
   }
 
 }
@@ -80,7 +89,7 @@ final class DrillbitProcessLoad implements Gauge<Double> {
 
   @Override
   public Double getValue() {
-    return osMXBean.getProcessCpuLoad();
+    return (osMXBean != null) ? osMXBean.getProcessCpuLoad() : null;
   }
 
 }
--- a/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
+++ b/common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
@@ -55,7 +55,7 @@ public class CpuGaugeSet implements MetricSet {
   private static OperatingSystemMXBean getOSMXBeanForCpuMetrics() {
     try {
       return (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
-    } catch (Exception ex) {
+    } catch (ClassCastException ex) {
       logger.warn("{}. Detected non-Supported JVM [{}]. CPU Metrics in the WebUI will not be available!", ex.getMessage(), System.getProperty("java.vm.name"));
     }
     return null;


common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -32,13 +32,24 @@
...
+    if (System.getProperty("java.vm.name").contains("HotSpot")) {
+      this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    } else {
+      this.osMXBean = null;
+      nonHotSpotJVMName = System.getProperty("java.vm.name");
+    }
     this.rtMXBean = ManagementFactory.getRuntimeMXBean();
+    if (nonHotSpotJVMName != null) {
+      logger.warn("Detected non-HotSpot JVM [{}]. CPU Metrics in the WebUI will not be available!", nonHotSpotJVMName);
+    }
```
arina-ielchiieva:
```
public class CpuGaugeSet implements MetricSet {
  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(CpuGaugeSet.class);

  private final RuntimeMXBean rtMXBean;
  private final OperatingSystemMXBean osMXBean;

  public CpuGaugeSet() {
    //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
    this.rtMXBean = ManagementFactory.getRuntimeMXBean();
    java.lang.management.OperatingSystemMXBean operatingSystemMXBean = ManagementFactory.getOperatingSystemMXBean();
    if (operatingSystemMXBean instanceof OperatingSystemMXBean) {
      this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
    } else {
      this.osMXBean = null;
      logger.warn("Detected non-HotSpot JVM [{}]. CPU Metrics in the WebUI will not be available!", System.getProperty("java.vm.name"));
    }
  }
...
```

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -63,7 +70,10 @@ public OperatingSystemLoad(OperatingSystemMXBean osBean) {
 
   @Override
   public Double getValue() {
-    return osMXBean.getSystemLoadAverage();
+    if (osMXBean != null) {
```
vrozov: Use `?` operator

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -32,13 +32,20 @@
...
-  private OperatingSystemMXBean osMXBean;
-  private RuntimeMXBean rtMXBean;
+  private final OperatingSystemMXBean osMXBean;
+  private final RuntimeMXBean rtMXBean;
 
   public CpuGaugeSet() {
-    this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
     this.rtMXBean = ManagementFactory.getRuntimeMXBean();
+    //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
+    if (System.getProperty("java.vm.name").contains("HotSpot")) {
```
arina-ielchiieva: 1.  We can extract `System.getProperty("java.vm.name")` into variable and not call it twice.
2. Can `System.getProperty("java.vm.name")` not be set and return null? In this case we'll get NPE.

vrozov: Where the guarantee that `ManagementFactory.getOperatingSystemMXBean()` returns an instance of `OperatingSystemMXBean` if `""java.vm.name"` contains HotSpot comes from? Why not to try the assignment and handle an exception?

vrozov: I'd recommend `try/catch` over `instanceOf`.

arina-ielchiieva: Any particular reason why?

vrozov: - It is a common design pattern in the case when a check will result in a positive outcome 99.9% of the time.
- it can handle a JDK that decides not to implement `ManagementFactory.getOperatingSystemMXBean()` and throw `UnsupportedOperationException` instead

vrozov: @kkhatua The reverse of the condition is also problematic, it is possible that `"java.vm.name"` does not contain HotSpot while cast to `OperatingSystemMXBean` may still succeed.

kkhatua: I like the `try/catch` approach. The instanceOf actually passed for the IBM JDK and failed in the cast. The `java.vm.name` appears to be standard, so I doubt it would be missing. Other JDKs might not implement it because it is a `com.sun. ...` package. (Remembering in context of Oracle vs Google, over Android)

arina-ielchiieva: Sounds reasonable, let then proceed with try / catch approach.

vrozov: @kkhatua try GraalVM, it is a successor for HotSpot.

kkhatua: Ok. I'm not able to repro the ClassCastException with the IBM JDK running an Oracle compilation. 
The IBM JDK on my Linux box is using `com.ibm.lang.management.internal.UnixExtendedOperatingSystem` which implements the `com.sun.management.OperatingSystemMXBean` interface. The JIRA reports a class cast exception from `com.ibm.lang.management.ExtendedOperatingSystem`. So, it might be an issue with the platform's implementation. 
I'll make the change however, and see if the original JIRA filer can confirm the fix.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -32,13 +32,23 @@
...
+    //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
+    OperatingSystemMXBean tempOSMXBean = null;
+    try {
+      tempOSMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    } catch (ClassCastException cce) {
+      logger.warn("Detected non-HotSpot JVM [{}]. CPU Metrics in the WebUI will not be available!", System.getProperty("java.vm.name"));
+      tempOSMXBean = null;
+    }
+    this.osMXBean = tempOSMXBean;
+    System.out.println("osMXBean::" + osMXBean);
```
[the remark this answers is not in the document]
arina-ielchiieva: In this case I suggest not to push intermediate results, since it's quite confusing for the reviewer.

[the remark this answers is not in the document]
kkhatua: Agreed. The fix has been verified and I've done the clean up.

kkhatua:
@arina-ielchiieva / @vrozov The fix has been verified. Let me know if I need to do any other changes, else I can squash and push the commit.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -32,13 +32,22 @@
...
 
   public CpuGaugeSet() {
-    this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
     this.rtMXBean = ManagementFactory.getRuntimeMXBean();
+    //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
+    OperatingSystemMXBean tempOSMXBean;
+    try {
+      tempOSMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    } catch (ClassCastException cce) {
+      logger.warn("{}. Detected non-Supported JVM [{}]. CPU Metrics in the WebUI will not be available!", cce.getMessage(), System.getProperty("java.vm.name"));
```
arina-ielchiieva: @kkhatua 
1. Could you please give example of warn message? I am not sure that `e.getMessage` will give enough information.
2. Vlad's suggestion to use ternary operator should be also implemented.
3. Please the squash the commits.

kkhatua: @arina-ielchiieva the only message you get is
``` java.lang.ClassCastException: com.ibm.lang.management.ExtendedOperatingSystem incompatible with com.sun.management.OperatingSystemMXBean```
Not very useful, but you still get the simplified message without stack trace. There isn't much value in adding more information than what is already there, because the issue only emerged when JDBC drivers were being built using the IBM JDK, something that a regular user will not face. (my testing showed that an OracleJDK compiled class ran without issue for an IBM JDK Runtime).

I'll apply Vlad's suggestion & squash the commits.

File: common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java [hunk 6]
```diff
@@ -63,7 +72,7 @@ public OperatingSystemLoad(OperatingSystemMXBean osBean) {
 
   @Override
   public Double getValue() {
-    return osMXBean.getSystemLoadAverage();
+    return (osMXBean != null) ? osMXBean.getSystemLoadAverage(): null;
```
arina-ielchiieva: What you would need braces here?

kkhatua: For readability. Helps delineate the condition and the 2 values.

vrozov: I am OK with parentheses, but please add a space before `:`.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -32,13 +32,22 @@
...
+  private final RuntimeMXBean rtMXBean;
 
   public CpuGaugeSet() {
-    this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
     this.rtMXBean = ManagementFactory.getRuntimeMXBean();
+    //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
+    OperatingSystemMXBean tempOSMXBean;
+    try {
+      tempOSMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    } catch (ClassCastException cce) {
```
vrozov: catch `RuntimeException` or `Exception`.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -32,13 +32,22 @@
...
-  private OperatingSystemMXBean osMXBean;
-  private RuntimeMXBean rtMXBean;
+  private final OperatingSystemMXBean osMXBean;
+  private final RuntimeMXBean rtMXBean;
 
   public CpuGaugeSet() {
-    this.osMXBean = (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
     this.rtMXBean = ManagementFactory.getRuntimeMXBean();
+    //DRILL-6702: Instead of worrying about compiling with IBM JDK, for now, we shall provide no CPU metrics for non-HotSpot JVMs
+    OperatingSystemMXBean tempOSMXBean;
```
vrozov: consider `private static` function that returns `OperatingSystemMXBean` instead of the temp variable.

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
kkhatua: I primarily expect only the ClassCastException, but I've converted this to catch Exception.

kkhatua:
Done the changes. Will squash and commit when I have +1 (s)

common/src/main/java/org/apache/drill/exec/metrics/CpuGaugeSet.java
```diff
@@ -49,6 +51,15 @@ public CpuGaugeSet() {
     metric.put("drillbit.uptime", new DrillbitUptime(rtMXBean));
     return metric;
   }
+
+  private static OperatingSystemMXBean getOSMXBeanForCpuMetrics() {
+    try {
+      return (OperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean();
+    } catch (Exception ex) {
```
arina-ielchiieva: Not sure how it can throw other exception instead of class cast. I liked version with more narrow exception better.

PR #1441, opened 2018-08-23
DRILL-6703: Query with complex expressions in lateral and unnest fails with CannotPlanException

vvysotskyi:
Initially to handle unnest with complex calls was introduced `ProjectComplexRexNodeCorrelateTransposeRule`. But this rule only matched `LogicalCorrelate -> Uncollect -> LogicalProject` pattern. So all other combinations, which have `LogicalCorrelate`, `Uncollect` and `LogicalProject` with rel nodes between `LogicalCorrelate` and `Uncollect` weren't handled.

The solution proposed in this PR is to make visitor based on `ProjectComplexRexNodeCorrelateTransposeRule` and apply it before applying rules, so all possible cases will be handled correctly.

For details please see [DRILL-6703](https://issues.apache.org/jira/browse/DRILL-6703).


arina-ielchiieva:
@HanumathRao could you please review?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ComplexUnnestVisitor.java
```diff
@@ -0,0 +1,195 @@
...
+      leftProjExprs.add(rexBuilder.makeInputRef(left, field.getIndex()));
+      fieldNames.add(field.getName());
+    }
+    fieldNames.add(COMPLEX_FIELD_NAME);
+
+    builder.push(left);
+
+    // Adds complex expression with replaced correlation
+    // to the projected list from the left
+    leftProjExprs.addAll(new RexFieldAccessReplacer(builder).apply(project.getChildExps()));
```
HanumathRao: Assuming that only one project expression is present at this point, Is it better to have an assert here to make sure that only one proj expression is added?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ComplexUnnestVisitor.java
```diff
@@ -0,0 +1,195 @@
...
+      return visit((Uncollect) other);
+    }
+    return super.visit(other);
+  }
+
+  public RelNode visit(Uncollect uncollect) {
+    RelBuilder builder = DrillRelFactories.LOGICAL_BUILDER.create(uncollect.getCluster(), null);
+    RexBuilder rexBuilder = builder.getRexBuilder();
+
+    assert uncollect.getInput() instanceof Project : "Uncollect should have Project input";
```
HanumathRao: why is this assert required? Is it better to have a if clause and return instead of assert?

vvysotskyi: In the case, if at this point we have something that differs from Project, then something went wrong.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
```diff
@@ -656,4 +656,29 @@ public void testUnnestNestedStarSubquery() throws Exception {
         .go();
   }
 
+  @Test
```
HanumathRao: Can you add a test case with multi lateral.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/ComplexUnnestVisitor.java
vvysotskyi: Agree with you. I remember some of checks at the validation stage, but one more wouldn't be excessive.
Added in the code above.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/lateraljoin/TestE2EUnnestAndLateral.java
vvysotskyi: Added `testMultiLateralWithComplexProject`.

HanumathRao:
Thanks for making the changes. Changes look good to me.

vvysotskyi:
@HanumathRao thanks for the review!

PR #1442, opened 2018-08-23
DRILL-6688 Data batches for Project operator exceed the maximum specified

bitblender:
This change separates the metadata-width and data-width of a variable-width column such that the data-width is used in all intermediate calculations and the meta-data width is added finally when the column's width is accumulated into the total width.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
@@ -212,7 +212,7 @@ public class OutputWidthVisitor extends AbstractExecExprVisitor<OutputWidthExpre
         }
         final RecordBatchSizer.ColumnSize columnSize = state.manager.getColumnSize(columnName);
 
-        int columnWidth = columnSize.getNetSizePerEntry();
+        int columnWidth = columnSize.getDataSizePerEntry();
         return new FixedLenExpr(columnWidth);
     }
 
@@ -256,12 +256,12 @@ public class OutputWidthVisitor extends AbstractExecExprVisitor<OutputWidthExpre
                                                         throws RuntimeException {
         OutputWidthExpression ifReducedExpr = ifElseWidthExpr.expressions[0].accept(this, state);
         assert ifReducedExpr instanceof FixedLenExpr;
-        int ifWidth = ((FixedLenExpr)ifReducedExpr).getWidth();
+        int ifWidth = ((FixedLenExpr)ifReducedExpr).getDataWidth();
         int elseWidth = -1;
         if (ifElseWidthExpr.expressions[1] != null) {
             OutputWidthExpression elseReducedExpr = ifElseWidthExpr.expressions[1].accept(this, state);
             assert elseReducedExpr instanceof FixedLenExpr;
-            elseWidth = ((FixedLenExpr)elseReducedExpr).getWidth();
+            elseWidth = ((FixedLenExpr)elseReducedExpr).getDataWidth();
         }
         int outputWidth = Math.max(ifWidth, elseWidth);
         return new FixedLenExpr(outputWidth);
@@ -270,8 +270,10 @@ public class OutputWidthVisitor extends AbstractExecExprVisitor<OutputWidthExpre
     private OutputWidthExpression getFixedLenExpr(MajorType majorType) {
         MajorType type = majorType;
         if (Types.isFixedWidthType(type)) {
-            int fixedWidth = ProjectMemoryManager.getWidthOfFixedWidthType(type);
-            return new OutputWidthExpression.FixedLenExpr(fixedWidth);
+            // Use only the width of the data. Metadata width will be accounted for at the end
+            // This is to avoid using metadata size in intermediate calculations
+            int fixedDataWidth = ProjectMemoryManager.getDataWidthOfFixedWidthType(type);
+            return new OutputWidthExpression.FixedLenExpr(fixedDataWidth);
         }
         return null;
     }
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
```diff
@@ -270,8 +270,10 @@ public OutputWidthExpression visitIfElseWidthExpr(IfElseWidthExpr ifElseWidthExp
     private OutputWidthExpression getFixedLenExpr(MajorType majorType) {
         MajorType type = majorType;
         if (Types.isFixedWidthType(type)) {
-            int fixedWidth = ProjectMemoryManager.getWidthOfFixedWidthType(type);
-            return new OutputWidthExpression.FixedLenExpr(fixedWidth);
+            // Use only the width of the data. Metadata width will be accounted for at the end
+            // This is to avoid using metadata size in intermediate calculations
+            int fixedDataWidth = ProjectMemoryManager.getDataWidthOfFixedWidthType(type);
```
Ben-Zvi: Since here we're guaranteed to get a fixed with type, we could just call TypeHelper.getSize() and eliminate getDataWidthOfFixedWidthType() .   This does not address the MinorType.NULL case, which is not clear how we get there ...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectMemoryManager.java
```diff
@@ -267,8 +284,13 @@ public void update() {
                 OutputWidthExpression savedWidthExpr = columnWidthInfo.getOutputExpression();
                 OutputWidthVisitorState state = new OutputWidthVisitorState(this);
                 OutputWidthExpression reducedExpr = savedWidthExpr.accept(new OutputWidthVisitor(), state);
-                width = ((FixedLenExpr)reducedExpr).getWidth();
+                width = ((FixedLenExpr)reducedExpr).getDataWidth();
                 Preconditions.checkState(width >= 0);
+                int metadataWidth = getMetadataWidth(columnWidthInfo.outputVV);
+                Preconditions.checkState(metadataWidth >= 0);
```
Ben-Zvi: This check is an overkill; seems that all the computations in getMetadataWidth() are bound to return >= 0 values.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectMemoryManager.java
```diff
@@ -301,4 +323,21 @@ public void update() {
...
+        int width = 0;
+        if (vv instanceof NullableVector) {
+            width += ((NullableVector)vv).getBitsVector().getPayloadByteCount(1);
+        }
+
+        if (vv instanceof VariableWidthVector) {
+            width += ((VariableWidthVector)vv).getOffsetVector().getPayloadByteCount(1);
+        }
+
+        if (vv instanceof BaseRepeatedValueVector) {
```
Ben-Zvi: Should this class be changed to the interface "RepeatedValueVector" to cover repeated maps and lists as well ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
bitblender: The MinorType.NULL check is required for handling cases where there can be a function with a null argument. This can happen, for instance, in convert_to_JSON like in TestComplexTypeReader testNonExistentFieldConverting() when trying to convert a non existent field.
See below for the object graph.

this = {DrillFuncHolderExpr@8281} 
 holder = {DrillSimpleFuncHolder@8279} "DrillSimpleFuncHolder [functionNames=[convert_toJSON, convert_toSIMPLEJSON], returnType=MajorType[minor_type: VARBINARY mode: REQUIRED], nullHandling=NULL_IF_NULL, parameters=[ValueReference [type=MajorType[minor_type: LATE mode: REQUIRED], name=input]]]"
 majorType = {TypeProtos$MajorType@8317} "minor_type: VARBINARY\nmode: OPTIONAL\n"
 interpreter = null
 args = {SingletonImmutableList@8315}  size = 1
  0 = {NullExpression@8285} 
   t = {TypeProtos$MajorType@8287} "minor_type: NULL\nmode: OPTIONAL\n"
 nameUsed = "convert_tojson"
 fieldReference = null
 pos = {ExpressionPosition@8322} "org.apache.drill.common.expression.ExpressionPosition@748c42c3[charIndex = -1, expression = --UNKNOWN EXPRESSION--]"

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectMemoryManager.java
bitblender: Will remove.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectMemoryManager.java
bitblender: Yes. I missed it. Thanks for noticing it.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/OutputWidthVisitor.java
Ben-Zvi: Still can get rid of * getDataWidthOfFixedWidthType() *, as all the checks there for variable width are not needed.

bitblender:
@Ben-Zvi Thanks for your review. I have addressed your review comments.

bitblender:
@Ben-Zvi I am looking at the CI failures.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectMemoryManager.java
Ben-Zvi: Sorry for misleading -- thanks for the fix .....

PR #1444, opened 2018-08-24
DRILL-6709: Extended the batch stats utility to other operators

sachouche:
- Enhanced the Batch Stats utility to handle input / output operators as it was being used only by the Parquet reader
- Modified all batch stats logging (from other operators which support batch sizing) to use the Batch Sizing utility

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.util.record;
 
+import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
 import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.ops.FragmentContext;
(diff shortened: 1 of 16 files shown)


sachouche:
@ilooner and @bitblender, can you please review this PR?

Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
```diff
@@ -134,25 +135,66 @@ private boolean isBatchStatsEnabledForOperator(FragmentContext context, Operator
     }
   }
 
+  /** Indicates whether a record batch is Input or Output */
+  public enum BatchIOType {
```
bitblender: I feel "RecordBatchType" describes the enum better then "BatchIOType"

exec/java-exec/src/main/java/org/apache/drill/exec/util/record/RecordBatchStats.java
```diff
@@ -279,4 +324,19 @@ private static void logBatchStatsMsg(RecordBatchStatsContext batchStatsContext,
     }
   }
 
+  private static String toString(BatchIOType ioType) {
+    Preconditions.checkNotNull(ioType, "The record batch IO type cannot be null");
+
+    switch (ioType) {
```
bitblender: 'case' is generally indented one level into the 'switch'. 

Also, Java allows Enums to have constructors and methods that  can be used to associate a string with an enum value. This makes modification of enum names and attributes easier. See https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

sachouche:
@bitblender  thank you for your review! I have implemented all the feedback that you requested.

sachouche:
@ilooner can you please review this PR?
thanks!

sachouche:
Thank you @bitblender and @ilooner for performing the review!

PR #1445, opened 2018-08-26
DRILL-6706: fixed null pointer exception in HashJoin

sachouche:
**Findings after Code Inspection -**
Looking at the code, having back-ticks within SchemaPath is not necessary (this is my understanding)
- Back-ticks are useful in the context of a compound name such as T.`column.with.a.dot`.another_column
- As soon as individual parts are parsed then the back-ticks can be omitted

Then why Aggregator was able to handle columns having back-ticks (and whatever I could throw at it )
- I found logic to strip the extra back-ticks within the aggregator code

**Suggested Fix -**
- Stripping the back-tick from the MaterializedField names within Parquet seems to fix the HashJoin issue
- The Aggregator didn't seem bothered by this change either

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
@@ -18,8 +18,8 @@
 package org.apache.drill.exec.record;
 
 import java.util.Set;
+import java.util.Collection;
 import java.util.Map;
-
 import org.apache.drill.common.map.CaseInsensitiveMap;
 import org.apache.drill.common.types.TypeProtos;
 import org.apache.drill.common.types.TypeProtos.DataMode;
@@ -631,7 +631,7 @@ public class RecordBatchSizer {
 
   // This keeps information for only top level columns. Information for nested
   // columns can be obtained from children of topColumns.
-  private Map<String, ColumnSize> columnSizes = CaseInsensitiveMap.newHashMap();
+  private Map<String, ColumnSize> columnSizes = new QuoteInsensitiveMap(CaseInsensitiveMap.newHashMap());
 
   /**
    * Number of records (rows) in the batch.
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/ParquetSchema.java
```diff
@@ -228,7 +228,7 @@ public void createNonExistentColumns(OutputMutator output, List<NullableIntVecto
 
   private NullableIntVector createMissingColumn(SchemaPath col, OutputMutator output) throws SchemaChangeException {
     // col.toExpr() is used here as field name since we don't want to see these fields in the existing maps
-    MaterializedField field = MaterializedField.create(col.toExpr(),
+    MaterializedField field = MaterializedField.create(col.toExpr().replace('`',' ').trim(),
```
vrozov: Why is it necessary to replace \` with a space to trim it later?

sachouche:
Modified the fix to avoid regression due to DRILL-4264:
- This should be a temporary fix up to @vvysotskyi  clarifies and improves his DRILL-4264 fix
- Instead of quoting not found columns, we should improve the metadata so that operators could handle such use-cases explicitly
- Not also this behavior is based on readers; it should apply to all readers (not just Parquet)

Summary - the current fix keeps the Parquet reader behavior while fixing the NullPointerException.

sachouche:
@ilooner can you please review the fix?

Thanks!

sachouche:
Thanks @ilooner! The RecordBatchSizer doesn't have to change since it seems that nested columns (e.g.; a.b.c) will be passed as leaf columns because of the Project operators. I was talking in general, that column references should be passed consistently to all operators.

ilooner:
Thanks for the clarification

vvysotskyi:
@ilooner, I don't have any objection regarding this PR. If this bug is critical, it may be merged and DRILL-6714 with more proper fix may be done later.

sohami:
@sachouche - Please make PR and JIRA title same.

PR #1450, opened 2018-08-28
DRILL-6717: lower and upper functions not works with national characters

oleg-zinovev:


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java
```diff
@@ -529,7 +538,16 @@ public void eval() {
```
Ben-Zvi: Isn't this too complicated ?   I tried and changed this code by adding here:
```
      String orig = new String(buffer.unsafeGetMemory(0,out.end));
      String upp = orig.toUpperCase(/*Locale.GERMAN*/);
      buffer.setBytes(0,upp.getBytes(),0,out.end);
```
and then commented out the whole following *for* loop, and the test passed!

(not sure which one performs better)

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java
```diff
@@ -1441,11 +1442,12 @@ public void testUpper() throws Exception {
             "upper('AbC aBc') as col_space,\n" +
             "upper('123ABC$!abc123.') as col_special,\n" +
             "upper('') as col_empty,\n" +
-            "upper(cast(null as varchar(10))) as col_null\n" +
+            "upper(cast(null as varchar(10))) as col_null,\n" +
+            "upper('привет, мир!') as col_national\n" +
```
Ben-Zvi: Also added here  
```+ "upper('ßßßßßßßßßß') as col_ss\n" + ```
 which should yield 
```"SSSSSSSSSSSSSSSSSSSS"``` 
as uppercase (even without Locale.GERMAN) . This was in part to check for a buffer overflow ....

Comment (arina-ielchiieva) [reply exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java]: @oleg-zinovev there ignored unit test tow tests below (https://github.com/apache/drill/pull/1450/files#diff-01e4581e7fc41d7bc289363e5d7ab661L1471), just enable it and it should cover the use cases.

Comment (oleg-zinovev) [reply exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java]: @Ben-Zvi 
The test works, but I'm confused by the reason:
`ß` - 2 byte character, converting it to the upper case returns 2 single-byte `S` characters. Are there any other examples of such symbols?

Comment (arina-ielchiieva) [reply exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java]: Awesome. Please remove changes in the test case since `testMultiByteEncoding` covers all that is needed.

Comment (oleg-zinovev) [reply exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/impl/TestStringFunctions.java]: @Ben-Zvi 
I can not find any string function supporting UTF-16

File: exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java (hunk 1 above)
Comment (oleg-zinovev) [reply exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java]: I don't know reason (maybe performance or allocation of big strings ?), but In test classes you can find simple implemenations:
org.apache.drill.udf.dynamic.LowerFunction 
/ org.apache.drill.udf.dynamic.CustomUpperFunction

Comment (arina-ielchiieva) [reply exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java]: These functions are generated to test dynamic UDFs, they are not shipped with Drill.

**Comment by arina-ielchiieva** [comment]
Looks good, @Ben-Zvi any comments from your side?

**Comment by Ben-Zvi** [comment]
@arina-ielchiieva - I'm still undecided; the Java technology is mature enough to handle national characters Up/Down case conversions, so why should we reimplement and add complexity ?  Any benefit to the PR code (over using _toUpperCase()_ , etc) ?
   I compared the performance (by adding Stopwatch calls around the conversions) by converting the same string in a loop -- the PR code initially performs poorly, then becomes on par with the _toUpperCase()_ code (likely the JIT improves).

**Comment by arina-ielchiieva** [comment]
@oleg-zinovev what do you think about Boaz comment?

**Comment by oleg-zinovev** [comment]
@arina-ielchiieva , @Ben-Zvi 
Shortly: String.toUpperCase faster on non-ascii strings and slower on ascii-only strings. I will provide my JMH code + new PR with "merged" idea tomorrow.

**Comment by oleg-zinovev** [comment]
Bechmark:
https://github.com/oleg-zinovev/DRILL-6717-benchmark
Charts:
ASCII - lower: https://live.amcharts.com/kNTU4
ASCII - upper: https://live.amcharts.com/xYzIx
NON_ASCII - lower: https://live.amcharts.com/lkYjI
NON_ASCII - upper: https://live.amcharts.com/QyYTI
MIXED (ASCII + NON_ASCII) - lower: https://live.amcharts.com/jOWEz
MIXED (ASCII + NON_ASCII) - upper: https://live.amcharts.com/hhNGI
Description:
BenZvi - Solution proposed by Ben-Zvi
OZ - Old PR code
Merged - New PR code
Value axis is logarithmic

P.S. I apologize for the delay.

File: log.txt [hunk 3]
```diff
@@ -0,0 +1,1697 @@
+[[1;34mINFO[m] Scanning for projects...
```
Ben-Zvi: This log file should not have been added to git !!

Ben-Zvi:
@oleg-zinovev - thank you for the thorough performance comparison.
Though I'm confused - the graphs seem similar, but comparing the numbers from the log (I looked at the ops/s figure before the [Average]) seem to "jump around" - sometimes similar, sometimes double or more the other figure, with no clear "better" side.

The other question was about "simple, standard code" vs. "complex code".  The initial comment suggested using the simple code, unless the complex code has a significant performance advantage.

arina-ielchiieva:
@oleg-zinovev did you have a chance to take a look at Boaz question?

arina-ielchiieva:
@Ben-Zvi looks like Oleg changed the code. Is it ok now? Could you please take a look.
@oleg-zinovev please do not forget to tag reviewers when you have made the changes, otherwise nobody will know that you have addressed code review comments.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/StringFunctions.java
```diff
@@ -786,10 +784,11 @@ public void setup() {
 
     @Override
     public void eval() {
-      out.buffer = buffer = buffer.reallocIfNeeded(input.end - input.start);
+      byte[] result = org.apache.drill.exec.expr.fn.impl.StringFunctionHelpers.initCap(input.start, input.end, input.buffer);
```
paul-rogers: The earlier functions handle the DrillBuf to string conversion and the String to byte[] conversion. Might want to do that here, changing the initCap() method to:

```
String initCap(String input)
```

oleg-zinovev: @paul-rogers review, please

PR #1453, opened 2018-08-30
DRILL-6710: Disallow negative scale for decimal data type

vvysotskyi:
- Added adjustment to decimal return type inference rules to force setting non-negative scale and max numeric precision even for the cases when the actual resulting value may overflow expected precision.
- Added check at the execution stage for fitting decimal value into the expected precision.
This PR is based on the https://github.com/apache/drill/pull/1451

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/output/DecimalReturnTypeInference.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/output/DecimalReturnTypeInference.java
@@ -277,7 +277,7 @@ public class DecimalReturnTypeInference {
 
       return TypeProtos.MajorType.newBuilder()
           .setMinorType(attributes.getReturnValue().getType().getMinorType())
-          .setScale(scale)
+          .setScale(Math.max(scale, 0))
           .setPrecision(precision)
           .setMode(mode)
           .build();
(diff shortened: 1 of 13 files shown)


vvysotskyi:
@sohami, could you please review?

vvysotskyi:
With the change in this PR, if aggregate functions: 
- `AVG`
- `STDDEV_POP`
- `STDDEV_SAMP`
- `VAR_POP`
- `VAR_SAMP`
are applied to the decimal column without cast may fail if any of the intermediate calculations for them (all of them may be rewritten using `SUM` and `COUNT` aggregate functions) exceeds max allowed numeric precision.
For the case, when planner knows that this column is decimal (for example, cast to decimal is used inside agg function), they wouldn't be rewritten and exact implementation will be used.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/output/DecimalReturnTypeInference.java
```diff
@@ -277,7 +277,7 @@
 
       return TypeProtos.MajorType.newBuilder()
           .setMinorType(attributes.getReturnValue().getType().getMinorType())
-          .setScale(scale)
+          .setScale(Math.max(scale, 0))
```
sohami: How about adjusting precision/scale for other `ReturnTypeInference` like `DecimalAggReturnTypeInference, DecimalMaxScaleReturnTypeInference, etc`

vvysotskyi:
Thanks for the review!

Currently, such behavior of aggregate functions cannot be fixed because Drill does not know the input type of the function and rewrites it using `SUM` and `COUNT` aggregate functions. For regular types, it is ok, since results of these intermediate agg calls may be reused, but for decimals due to type inference rules, intermediate calculations on the results of agg calls may cause overflow.

vvysotskyi:
Note for batch committer: this PR causes functional tests failures, so please don't merge it until PR with the fix for baselines is prepared.

priteshm:
@vvysotskyi any update on the PR for the functional tests?

denysord88:
@priteshm PR is ready, but I have some issues in drill-test-framework after baselines updates.
In negative tests test-framework can't parse exceptions for 7 tests. Now I'm trying to fix it.

PR #1454, opened 2018-08-31
DRILL-6684: Swap sys.options and sys.options_val tables

kkhatua:
(As a follow up to DRILL-5735)

The current sys.options table has a verbose layout, because of which sys.options_internal was introduced. The latter supports `descriptions`, which means it makes sense to have that table as the new `sys.options`. The recommendation is to deprecate the old format, so that any dependencies continue to make use of it as long as required.
Existing tests have been patched to account for the swap.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
@@ -31,28 +31,30 @@ import org.apache.drill.exec.store.sys.OptionIterator.OptionValueWrapper;
  * </p>
  */
 public enum SystemTable {
-  OPTION("options", false, OptionValueWrapper.class) {
+  OPTIONS_OLD("options_old", false, OptionValueWrapper.class) {
+    @Deprecated
     @Override
     public Iterator<Object> getIterator(final ExecutorFragmentContext context, final int maxRecords) {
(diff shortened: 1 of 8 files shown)


kfaraaz:
@kkhatua Is it not possible to add descriptions to the existing sys.options table ?
Will users no longer be able to query from sys.options table, with this new change ?

kkhatua:
@kfaraaz 
The existing `sys.options` table is very verbose because of multiple 'datatype_val' columns (all except one will be alway null). Injecting a descriptions column breaks the visual formatting within SQLLine.
The existing `sys.options_val` is better representative of the options. For usability sake, we are renaming the latter table as `sys.options` , while the old table becomes `sys.options_old`. So, both tables will still be available.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTable.java
```diff
@@ -31,28 +31,30 @@
  * </p>
  */
 public enum SystemTable {
-  OPTION("options", false, OptionValueWrapper.class) {
+  OPTION_OLD("options_old", false, OptionValueWrapper.class) {
```
arina-ielchiieva: Maybe rename as `OPTIONS_OLD` and `OPTIONS` for consistency? (making it plural).

kkhatua: Done.

PR #1455, opened 2018-08-31
DRILL-6724: Dump operator context to logs when error occurs during query execution

KazydubB:


exec/java-exec/src/main/java/org/apache/drill/exec/store/avro/AvroRecordReader.java
```diff
@@ -398,4 +399,21 @@ public void close() {
...
+  @Override
+  public UserException createDataReadException(Throwable cause, String message) {
+    UserException.Builder builder = UserException.dataReadError(cause)
+        .message(message)
+        .addContext("File ", hadoop);
+    long currentPosition = -1L;
+    try {
+      currentPosition = reader.tell();
+    } catch (IOException e) {
+      // Ignore as this is to provide information only
```
vdiravka: `logger.trace()`

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/sequencefile/SequenceFileRecordReader.java
```diff
@@ -163,4 +163,23 @@ public void close() {
...
+    UserException.Builder builder = UserException.dataReadError(cause)
+        .message(message)
+        .addContext("File ", split.getPath().toString());
+    long position = -1L;
+    try {
+      if (reader != null) {
+        position = reader.getPos();
+      }
+    } catch (IOException e) {
+      // Ignore as this is to provide information only
```
vdiravka: `logger.trace()`

paul-rogers:
I have a basic question on this one. Is an Index Out Of Bounds Exception (IOBE) an expected behavior for this context? Or, does it indicate a bug in the code? Normally, code is expected to know the array size, and index only within that size, meaning that an IOBE is a program error.

By presenting the error to the user in the nicely-formatted way introduced in this code, we seem to be implying that this is something the user needs to worry about. Yet, if this is a program error, we should say that this is a System Error and they need to contact Support for resolution.

Note also that for unexpected program errors, there is a well-defined generic way we catch exceptions in the Fragment Executor. The Fragment Executor is the handler of last resort, will report an error, and will shut down the fragment.

So, basic question: what is the source of the IOBE? If the IOBE is a program error, do we need the fancy handling if the Fragment Executor handles such errors? And, if the IOBE indicates a problem the user can resolve, should the error messages indicate a possible resolution? What might that problem be? Corrupted database? Incorrect SQL? Update done concurrently with a query? Something else?

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JSONRecordReader.java
```diff
@@ -256,4 +256,13 @@ public void close() throws Exception {
       stream.close();
     }
   }
+
+  @Override
+  public UserException createDataReadException(Throwable cause, String message) {
```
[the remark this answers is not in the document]
KazydubB: Used this method inside handleAndRaise.

exec/java-exec/src/main/java/org/apache/parquet/hadoop/ColumnChunkIncReadStore.java
```diff
@@ -269,4 +270,10 @@ public PageReader getPageReader(ColumnDescriptor descriptor) {
   public long getRowCount() {
     return rowCount;
   }
+
+  public UserException.Builder getExceptionWithContext(Throwable cause, String message) {
+    return UserException.dataReadError(cause)
```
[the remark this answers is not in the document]
KazydubB: Unfortunetely, this DrillRuntimeException is created with data from inner class (ColumnChunkIncPageReader) and there is no way to extract it at the moment.

KazydubB:
@paul-rogers, source of IOBE here is when allocating value vector buffers or reading records from input. Currently, every Exception (except OutOfMemoryException and UserException) will be wrapped into UserException with ErrorType INTERNAL_ERROR (and will remain being such after FragmentExecutor handling).
Purpose of this improvement is to enhance error messages to provide more useful information to users, while original cause is preserved. Provided context may be useful for users as it may provide some ways of resolving the issue or to make a workaround while the issue is being investigated etc.
You're right that IOBE indicates a program error so I've made changes to wrap IOBE into UserException of type INTERNAL_ERROR and not DATA_READ_ERROR (as it was before in the PR).

paul-rogers:
@KazydubB, thanks for the explanation. You are inserting code everywhere you believe an IOBE could be thrown. Since the exception is due to a coding error, you should, in theory, wrap all possible places. But, one could apply this same argument to NPE, division by zero and other programing errors. Soon, every line of code must be wrapped to catch programming errors.

And, what can the user do? Only call support or post to the dev list. We then look at logs: the stack trace, an create a test case.

So, we could adopt a more general approach:

* If the error is something the user can correct (bad name, external system problem, resource problem), log a detailed error for the user. Catch the problem close to the cause so we can provide context.
* If the error is a programming problem, catch it generically in the fragment executor. The original exception has the stack trace with is what the developer needs to fix the issue. Tell the user, "System error -- contact your vendor. Preserve drillbit.log for host foo.bar.com and provide it to the developer."

If we do that, then your wrapper code is not needed unless it indicates an error the user can fix. Instead, improve the wording of the generic "something went wrong" error in the fragment executor.

That is, there is no reason to add complexity to the code to try to handle all possible programming errors. Instead, focus that effort on better testing to find and fix such errors. Said another way, the code being added assumes errors exist, we don't know about them, and the user would like lots of details about how we screwed up in the code. That is, IMHO, an incorrect read. Uses want working code, and want to know how to correct errors due to their application or environment.

Note that this is policy is not unique to each plugin. Rather, it is a project-wide guideline that might want to be discussed on the dev list.

arina-ielchiieva:
@paul-rogers 
one thing I am not sure about is that stack trace contains all needed information, sometimes having file name, line number is really useful for debugging or providing devs the info, so at least all this can be logged.

paul-rogers:
@arina-ielchiieva, very good point. Having that information is very useful regardless of the source of error. How could we provide that generically without the need to wrap each bit of code in a "in case something goes wrong" try/catch block? For example, suppose we get an NPE or OOM. Should we catch all those?

Could each operator (including scanners) provide a context which can be dumped to the log along with the stack trace? Often, we try to recreate the fragment operators from the stack trace. Would be better to see a stack of operators, each with it's critical info.

The fragment exec walks the operator tree to shut down each operator. I could also do so to create the operator stack with related info.

Is this something we could consider? Would have been very helpful the many times I tried to track down a failure found in a QA test or user case.

vdiravka:
@KazydubB It would be useful to show the use case how the error was displayed and how it is changed after these changes. Without any real use case it is not clear whether it is necessary. Please find some cases to show the additional info in Exception message and post it in Jira description.
Also as I have mentioned, to go further with these changes it is necessary to have a unit test with verifying the new Exception content.

paul-rogers:
Note that is is, in general, very hard to test an exception such as these. We are catching programming errors. How does one trigger such exceptions? By running a test that finds a programming bug. What is the proper response? Fix the bug so that it never occurs again. This is why it is so hard to catch specific programming errors: by definition, they should never occur.

However, if done at the fragment executor level, then Drill does have a way to trigger an exception using the error injection framework. Basically, pick a point in the code (perhaps in one of the readers) an call the error injector. If it is enabled, it will throw the requested exception. You can then verify that the error is as expected. And, since the error is intentionally injected, you can repeat this test as often as you like.

Error handling is a tricky topic, which is why I'm jumping in to offer suggestions based on my time in "the school of hard knocks."

Ben-Zvi:
The IBM JVM has a special call: `com.ibm.jvm.Dump.SystemDump()` that generates a dump (similar to a Linux core dump) that can be analyzed later (e.g., check values of fields, heap. Probably not direct memory). It can even garbage collect the heap before dumping !!
  Unfortunately the standard JVMs offer no such call yet, hence not applicable for Drill now.

arina-ielchiieva:
Agree with @paul-rogers, we should consider catching exception at a higher level and provide user general explanation and suggest to refer to the logs, meanwhile in logs do operator dump (including as much info as possible, like file name, line, exception etc).

KazydubB:
@vdiravka and @paul-rogers I've reimplemented PR, please take a look.

Cc: @arina-ielchiieva

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaRecordReader.java
```diff
@@ -116,6 +118,7 @@ public int next() {
         if (++messageCount >= DEFAULT_MESSAGES_PER_BATCH) {
           break;
         }
+        currentMessageCount = messageCount;
```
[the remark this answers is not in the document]
KazydubB: Good point, removed local variable.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JSONRecordReader.java
```diff
@@ -205,36 +191,32 @@ public int next() {
...
+          if (write == ReadState.JSON_RECORD_PARSE_EOF_ERROR) {
+            break;
           }
+        } else {
+          break;
         }
-        else{
-          break outside;
-        }
+      } catch(IOException ex) {
```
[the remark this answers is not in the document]
KazydubB: Done.

KazydubB:
@vdiravka regarding your concerns:

- I suppose you are right that there is no need to dump all operators up to a failed one (though having this information may help recreating fragment operators in some cases). Dump from the failed batch might suffice in most cases. Maybe we should consider configuration parameter for number of batches to be dumped?

- This `DumpCat` relies on Trace operator which should be injected to physical plan (see comments in [Implement Diagnostic Operator](https://issues.apache.org/jira/browse/DRILL-256)) and it seems that TraceInjector, which traverses the set of operators and injects trace operator before every operator is not invoked anywhere, since the trace operator does not support hyper vectors. Moreover this feature is poorly documented (and may appear unreliable as it is not being tested/accounted for since it was added almost 5 years ago and there are no recent jiras for the subject). And lastly, script to run this tool (DumpCat) is not shipped in Drill's distribution ([Remove dumpcat script from Drill distribution](https://issues.apache.org/jira/browse/DRILL-4751)) rendering it hard for users to use it.

vdiravka:
@KazydubB 
* The last regular (successful) batch and failed batch can be useful (in case of Schema Changes and so on). I think they are enough. Or when dumping of the more batches can be useful?
* It would be useful to enable (for instance with additional session option) dumping the additional info by leveraging these Trace operators. It could be future enhancement, could you please create a Jira ticket with description how it is possible to improve dump operator context with `DumpCat`.
* What about other comments in CR?
* Did you check your functionality? How does the new info in logs look like? Please create some test cases with mocking exception throwing from different operators.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/BaseRootExec.java
```diff
@@ -123,6 +123,21 @@ public void receivingFragmentFinished(final FragmentHandle handle) {
     logger.warn("Currently not handling FinishedFragment message");
   }
 
+  @Override
+  public void dumpOperators() {
+    if (!operators.isEmpty()) {
+      logger.info("Operator dump started.");
+      for (CloseableRecordBatch batch : operators) {
+        batch.dump();
```
KazydubB: You are right, I suppose a failed and its successful parent operators should do. Thanks.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/ScanBatch.java
```diff
@@ -199,11 +205,13 @@ public IterOutcome next() {
             continue; // Skip to next loop iteration if reader returns 0 row and has same schema.
         } else {
           // return OK if recordCount > 0 && ! isNewSchema
+          lastOutcome = IterOutcome.OK;
```
KazydubB: Sure.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -1603,6 +1603,26 @@ private void updateStats(HashTable[] htables) {
     }
   }
 
+  @Override
+  public String toString() {
+    return "HashAggTemplate[numPartitions=" + numPartitions
```
KazydubB: Thank you for suggestion, will do.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
```diff
@@ -247,4 +247,10 @@ private void computeBitSet(int fieldId, BloomFilter bloomFilter, BitSet bitSet)
       }
     }
   }
+
+  @Override
+  public void dump() {
+    logger.info("RuntimeFilterRecordBatch[selectionVector={}, toFilterFields={}, originalRecordCount={}, " +
```
KazydubB: Agreed, will change.

KazydubB:
@vdiravka I've addressed your comments, could you take a look, please?

exec/java-exec/src/test/java/org/apache/drill/TestOperatorDump.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+    assertTrue(String.format("Entries %s were not found in logs.", entryList), entryList.isEmpty());
+
+    Set<String> loggerNames = appender.getLoggerNames();
+    assertTrue(String.format("Entry for class %s was not found", expectedClassName),
+        loggerNames.contains(expectedClassName));
+  }
+
+  // ConsoleAppender which stores logged events
+  private static class TestAppender extends ConsoleAppender<ILoggingEvent> {
```
vdiravka: Consider rename class

exec/java-exec/src/test/java/org/apache/drill/test/LogFixture.java
```diff
@@ -204,10 +210,14 @@ private void setupConsole(LogFixtureBuilder builder) {
     ple.setContext(lc);
     ple.start();
 
-    appender = new ConsoleAppender<>( );
+    if (builder.appender == null) {
```
vdiravka: `appender = builder.appender == null ? new ConsoleAppender<>() : builder.appender;`

common/src/main/java/org/apache/drill/common/exceptions/UserExceptionContext.java
```diff
@@ -141,7 +141,7 @@ String generateContextMessage(boolean includeErrorIdAndIdentity, boolean include
     }
 
     if (includeSeeLogsMessage) {
-      sb.append("Please, refer to logs for more information.\n");
+      sb.append("\nPlease, refer to logs for more information.\n");
```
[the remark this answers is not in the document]
vdiravka: Looks like on Windows some editors like Notepad don't consider `\n` as line separator.
For instance, `Intelij Idea` terminal on Windows shows:
```
0: jdbc:drill:zk=local> select * from cp.`employee.json` where full_name = 5;
Error: SYSTEM ERROR: NumberFormatException: Sheri Nowmer

Fragment 0:0

Please, refer to logs for more information.

[Error Id: bdf70e30-4b36-48cf-8d97-39b2d31dbece on 10.2.14.127:31010] (state=,code=0)
```
but `sqlline.log` in NotePad shows:
```
2018-09-21 18:49:11,972 [245aed07-be42-1f73-4905-f1bbc400f50d:frag:0:0] ERROR o.a.d.e.w.fragment.FragmentExecutor - SYSTEM ERROR: NumberFormatException: Sheri NowmerFragment 0:0Please, refer to logs for more information.[Error Id: bdf70e30-4b36-48cf-8d97-39b2d31dbece on 10.2.14.127:31010]
org.apache.drill.common.exceptions.UserException: SYSTEM ERROR: NumberFormatException: Sheri NowmerFragment 0:0Please, refer to logs for more information. [Error Id: bdf70e30-4b36-48cf-8d97-39b2d31dbece on 10.2.14.127:31010]
```

PR #1456, opened 2018-09-05
DRILL-6729: Enable Unnest/Lateral Join feature by default

sohami:
@Ben-Zvi - Please help to review this PR.


sohami:
@Ben-Zvi - It was used in tests but looks like it was redundant since same key was defined in PlannerSettings as well. Replace it from ExecConstants.

PR #1457, opened 2018-09-05
Integration with InfluxDb via InfluxQL

mircotamburini:
It'll be very interesting having integration with this performant TimeSeries DB and the opportunity join other datasources (InfluxDb don't do it)


arina-ielchiieva:
@mircotamburini if you want to discuss integration please write email to Drill mailing list, please do not open dummy PR.

PR #1462, opened 2018-09-10
DRILL-3853: Upgrade to SqlLine 1.5.0

arina-ielchiieva:
Details in [DRILL-3853](https://issues.apache.org/jira/browse/DRILL-3853).


exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -70,6 +71,9 @@
      */
     GuavaPatcher.patch();
     Environment.logEnv("Drillbit environment: ", logger);
+    // Jersey uses java.util.logging - create bridge: jul to slf4j
+    SLF4JBridgeHandler.removeHandlersForRootLogger();
```
vdiravka: Is it for logging info from libs, which `use java.util.logging`? Is the default level suitable for Drill? How the log file is changed?

Does it makes sense to change the log level for it in some cases, possibly via logback.xml?

exec/java-exec/src/main/java/org/apache/drill/exec/client/DrillSqlLineApplication.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  public Map<String, OutputFormat> getOutputFormats(SqlLine sqlLine) {
+    return sqlLine.getOutputFormats();
+  }
+
+  @Override
+  public Collection<String> getConnectionUrlExamples() {
+    return Arrays.asList("jdbc:drill:", "jdbc:drill:zk=local");
```
vdiravka: What about `jdbc:drill:drillbit=localhost` instead of `jdbc:drill:`?

arina-ielchiieva: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
arina-ielchiieva: This was done to remove the following, especially when starting Drill in embedded mode.
```
Sep 10, 2018 3:27:01 PM org.glassfish.jersey.server.ApplicationHandler initialize
```
I don't think we need Jersey info logs in Drill log, error will be logged as usual.

arina-ielchiieva:
@vdiravka the PR was update with the new functionality: added ability to config Drill SqlLine properties via drill-sqlline.conf and override using drill-sqlline-override.conf. Please review.

arina-ielchiieva:
@vdiravka, thanks for the code review, addressed code review comments.

vdiravka:
+1 LGTM

PR #1463, opened 2018-09-11
DRILL-6625: Intermittent failures in Kafka unit tests

aravi5:
### Fix producer errors
- Increase the value of `ProducerConfig.REQUEST_TIMEOUT_MS_CONFIG` to accomadate slower systems.
- Increase the value of producer `ProducerConfig.RETRIES_CONFIG` to 3 (from 0).
- Prevent producer to send duplicate messages due to retries by setting `ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG` to `true`.

### Fix consumer errors
- Increase `ExecConstants.KAFKA_POLL_TIMEOUT` (from 200 ms).

### Stabilize KafkaFilterPushdownTest
- Catch the issue with setup early by checking if expected number of records exist before running tests in `KafkaFilterPushdownTest` class.

Please refer to [DRILL-6625](https://issues.apache.org/jira/browse/DRILL-6625) for more details about the fix.

PR #1464, opened 2018-09-11
DRILL-6733: Unit tests from KafkaFilterPushdownTest are failing in so…

agozhiy:
…me environments.

Added a check that prevents the cluster tear down if TestKafkaSuit is not running.

### DIFF
--- a/contrib/storage-kafka/src/test/java/org/apache/drill/exec/store/kafka/KafkaTestBase.java
+++ b/contrib/storage-kafka/src/test/java/org/apache/drill/exec/store/kafka/KafkaTestBase.java
@@ -86,7 +86,9 @@ public class KafkaTestBase extends PlanTestBase {
 
   @AfterClass
   public static void tearDownKafkaTestBase() throws Exception {
-    TestKafkaSuit.tearDownCluster();
+    if (TestKafkaSuit.isRunningSuite()) {
+      TestKafkaSuit.tearDownCluster();
+    }
   }
 
 }


arina-ielchiieva:
@aravi5 could you please take a look at this PR?

vdiravka:
I have BUILD SUCCESS with this solution after `mvn test -pl contrib/storage-kafka` with 1, 2 and 3 number of `-DforkCount`.
I think the design of `TestKafkaSuit` could be improved to avoid such kind of issues. But with current design the solution looks good.
@aravi5 Do you agree?

aravi5:
@agozhiy - Thank you for fixing this, the fix looks good.

@vdiravka  - The design of `TestKafkaSuit` is very similar to design of `MongoTestSuit` and hence would require changes similar to the ones made in [storage-mongo/pom.xml](https://github.com/apache/drill/pull/923/commits/f5dfa56f33a46b92e2f9de153d82a16a77642ddf#diff-e110e2cbfd77d27e85d5121529c612bfR83).

Current behavior is surefire runs tests classes twice - once as a part of `TestKafkaSuit` and the other by directly running classes. To prevent the latter from happening, changes were made in `pom.xml` for `storage-mongo` plugin. 

These changes can be addressed as a part of a separate PR or I am happy to include the changes as a part of my open PR https://github.com/apache/drill/pull/1463.

vdiravka:
@aravi5 Sounds reasonable. Looks like your proposal fits DRILL-6625 and #1463 PR, so you can include it there.

sohami:
Merged in master

PR #1466, opened 2018-09-13
DRILL-6381: Add support for index based planning and execution

amansinha100:
- Please see [DRILL-6381](https://issues.apache.org/jira/browse/DRILL-6381) for a description of the feature.  For additional details please see the commit messages.  Wherever appropriate, Javadoc has been provided in the new code. 

- This PR contains 5 commits and has contributions from several people who have been listed as co-authors in individual commits.

- This PR introduces a dependency on MapR 6.0.1 (current apache master is on v5.2.1) which in turn requires CentOS 7.x


amansinha100:
@Ben-Zvi since the code touches HashJoin and a few other run time operators, could you help review the operator changes ? In particular RowKeyJoin, RangePartition, ScanBatch and related changes. 

@arina-ielchiieva do you think you can help review the planner changes and storage plugin changes ? 

Note: All unit and functional tests (including advanced tests) are passing with this branch. 

Thanks in advance..

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/JoinControl.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package org.apache.drill.exec.planner.common;
+
+/**
+ * For the int type control,
+ * the meaning of each bit start from lowest:
+ * bit 0: intersect or not, 0 -- default(no intersect), 1 -- INTERSECT (DISTINCT as default)
+ * bit 1: intersect type, 0 -- default (DISTINCT), 1 -- INTERSECT_ALL
+ */
+public class JoinControl {
```
Ben-Zvi: Looks like the JoinControl currently is not being actually used by the runtime operators; is it just a preparation for future use ?

(A little off the subject)  And it is used to convey information from the planner to the operator. I was recently planning passing other information (e.g., semi-join, anti-semi, broadcast, between lateral and unnest ...) -- any such new bit of information requires so many changes, and extending method signatures; there should be an easier way to do that - like passing the whole plan (but that's another project ....).

amansinha100: The JoinControl is actually used in IndexIntersectPlanGenerator, NonCoveringIndexPlanGenerator and a few other places.  You may need to expand those files since GitHub does not automatically load the diffs by default. 

For the broader question of passing around the join type and related information in a more componentized way, yes this has been a pending item that planner needs to provide.  Pls feel free to create a JIRA and if you have any initial thoughts you can start proposing it there.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/planner/index/MapRDBIndexDiscover.java
```diff
@@ -0,0 +1,374 @@
...
+      collationMap.put(indexFields.get(i), indexFieldCollations.get(i));
+    }
+    CollationContext collationContext = new CollationContext(collationMap, indexFieldCollations);
+    return collationContext;
+  }
+
+  private DrillIndexDescriptor buildIndexDescriptor(String tableName, IndexDesc desc)
+      throws InvalidIndexDefinitionException {
+    if (desc.isExternal()) {
+      //XX: not support external index
```
HanumathRao: Please change this comment to "External index is not supported"

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/RestrictedMapRDBSubScanSpec.java
```diff
@@ -0,0 +1,219 @@
...
+    if (hasRowKey()) {
+      // get the entry at the current index within this batch
+      Object o = rowKeyVector.getAccessor().getObject(currentIndex++);
+      if (o == null) {
+        throw new DrillRuntimeException("Encountered a null row key during restricted subscan !");
+      }
+
+      // this is specific to the way the hash join maintains its entries. once we have reached the max
+      // occupied index within a batch, move to the next one and reset the current index to 0
+      // TODO: we should try to abstract this out
```
HanumathRao: Please try to simplify this comment.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/generators/NonCoveringIndexPlanGenerator.java
```diff
@@ -0,0 +1,335 @@
...
+    final RelDataTypeFactory.FieldInfoBuilder leftFieldTypeBuilder =
+        dbScan.getCluster().getTypeFactory().builder();
+
+    //we are applying the same index condition to primary table's restricted scan, the reason
+    // for this is, the scans on index table and primary table are not a transaction, meaning that _after_ index scan,
+    // primary table might already have data get updated, thus some rows picked by index were modified and no more satisfy the
+    // index condition. By applying the same index condition again here, we will avoid the possibility to have some
+    //not-wanted records get into downstream operators in such scenarios.
+    //the remainder condition will be applied on top of RowKeyJoin.
+    FilterPrel leftIndexFilterPrel = new FilterPrel(dbScan.getCluster(), dbScan.getTraitSet(),
```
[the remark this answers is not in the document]
amansinha100: Yes, synchronous indexes (like in RDBMS) obviate the need for re-applying this filter.  I can add an interface method isAsyncIndex() or some such to the IndexCollection and handle this accordingly.. is that ok ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/RowKeyJoinPrel.java
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  @Override
+  public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
+    if(PrelUtil.getSettings(getCluster()).useDefaultCosting()) {
+      return super.computeSelfCost(planner).multiplyBy(.1);
+    }
+    double rowCount = mq.getRowCount(this.getRight());
+    DrillCostFactory costFactory = (DrillCostFactory) planner.getCostFactory();
+    return costFactory.makeCost(rowCount, 0, 0, 0,
```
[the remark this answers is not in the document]
amansinha100: RowKeyJoin operator by itself incurs negligible CPU and I/O cost since it is not doing a real join.  The actual cost is attributed to the skip-scan (random I/O).  The RK join will hold 1 batch in memory but since it is not a buffering operator, we typically don't assign memory costs to them.   I can add the explanation.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/DrillDistributionTraitDef.java
```diff
@@ -86,7 +86,10 @@ public RelNode convert(
         return new HashToRandomExchangePrel(rel.getCluster(), planner.emptyTraitSet().plus(Prel.DRILL_PHYSICAL).plus(toDist), rel,
                                              toDist.getFields());
       case RANGE_DISTRIBUTED:
-        return new OrderedPartitionExchangePrel(rel.getCluster(), planner.emptyTraitSet().plus(Prel.DRILL_PHYSICAL).plus(toDist), rel);
+        // return new OrderedPartitionExchangePrel(rel.getCluster(), planner.emptyTraitSet().plus(Prel.DRILL_PHYSICAL).plus(toDist), rel);
```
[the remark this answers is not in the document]
amansinha100: The OrderedPartitionExchange was not being used anywhere.  Its original purpose was to do range-based sorting; however this was not completely functional.  It was also a much more complex operator (see its implementation in OrderedPartitionRecordBatch). I think if we go back to fixing it and using it in future, we will probably add a new type of distribution trait.  I will add explanation along these lines.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/generators/NonCoveringIndexPlanGenerator.java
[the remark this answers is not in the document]
gparai: Yes, adding it in the interface would be good. We would have to make a similar change in index intersection plan generator as well.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexGroup.java
```diff
@@ -0,0 +1,63 @@
...
+  public IndexGroup() {
+    indexProps = Lists.newArrayList();
+  }
+
+  public boolean isIntersectIndex() {
+    if (indexProps.size() > 1) {
+      return true;
+    } else {
+      return false;
+    }
```
Ben-Zvi: c'mon - why not   `return indexProps.size() > 1 ;` instead ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/JoinControl.java
Ben-Zvi: I cherry-picked the 5 commits on top of the (almost) latest master, and searching through the code I could not find who **consumes** the `joinControl` , that is - that value is being passed around in many method calls, and initializes local fields, but no where in the code a **decision** is made based on that value ... 
   Many methods that seem to analyze this value are unused. Like `isIntersect()`, `isIntersectDistinct()`, `isIntersectAll()` in `IndexGroup`,  or `getJoinControl()` in `HashTableConfig`.
    And also unused are the fields `joinControl` in `HashJoinTableTemplate`, or `joinControl` in `HashJoinPrel`.

amansinha100: Hmm...it seems a small part of the code in HashJoinBatch may not have been merged correctly .. either when I did the recent merge or when previously we merged the index code with the hash join spilling code.  On a separate private branch, I do see `JoinControl` being used for a decision point.  I will talk to you offline about this. 

I am also tagging @chunhui-shi who is a co-author and had added the `JoinControl` specifically to distinguish between INTERSECT_DISTINCT and INTERSECT_ALL.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexGroup.java
amansinha100: Agreed.  Done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/JoinControl.java
chunhui-shi: JoinControl has to be used in HashJoin runtime operator, basically it is to allow the operator to skip some operations when it is index join(so we are expecting INTERSECT and nothing else). I don't have the original code at hand, but I think it should be in the first MapR 6.0 release. Since this part is only about performance and correctness won't be impacted, if it was lost during some merge process people might not notice it.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/RestrictedMapRDBSubScanSpec.java
amansinha100: This method seems to be a left-over from a previous version of the implementation and it turns out is not needed anymore.  I have removed it.

amansinha100:
@gparai , @HanumathRao I have addressed most of your code review comments.  Working on @Ben-Zvi 's  comment on JoinControl.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexPlanUtils.java
```diff
@@ -0,0 +1,872 @@
...
+    }
+    return false;
+  }
+
+  /**
+   * For a particular table scan for table T1 and an index on that table, find out if it is a covering index
+   * @return
+   */
+  static public boolean isCoveringIndex(IndexCallContext indexContext, FunctionalIndexInfo functionInfo) {
+    if(functionInfo.hasFunctional()) {
```
vdiravka: space

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexPlanUtils.java
```diff
@@ -0,0 +1,872 @@
...
+   * @return false if the query could not be covered by the index (should not create covering index plan)
+   */
+  static private boolean queryCoveredByIndex(IndexCallContext indexContext,
+                              FunctionalIndexInfo functionInfo) {
+    //for indexed functions, if relevant schemapaths are included in index(in indexed fields or non-indexed fields),
+    // check covering based on the local information we have:
+    //   if references to schema paths in functional indexes disappear beyond capProject
+
+    if (indexContext.getFilter() != null && indexContext.getUpperProject() == null) {
+      if( !isFullQuery(indexContext)) {
```
vdiravka: formatting

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexPlanUtils.java
```diff
@@ -0,0 +1,872 @@
...
+      }
+    }
+    else {//upperProject and lowerProject both are null, the only place to find columns being used in query is scan
+      exprs.addAll(indexContext.getScanColumns());
+    }
+
+    Map<LogicalExpression, Set<SchemaPath>> exprPathMap = functionInfo.getPathsInFunctionExpr();
+    PathInExpr exprSearch = new PathInExpr(exprPathMap);
+
+    for(LogicalExpression expr: exprs) {
```
vdiravka: space

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexPlanUtils.java
```diff
@@ -0,0 +1,872 @@
...
+
+  /**
+   *
+   * @param expr
+   * @param context
+   * @return if there is filter and expr is only in equality condition of the filter, return true
+   */
+  private static boolean exprOnlyInEquality(LogicalExpression expr, IndexCallContext context) {
+    //if there is no filter, expr wont be in equality
+    if(context.getFilter() == null) {
```
vdiravka: space

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexPlanUtils.java
```diff
@@ -0,0 +1,872 @@
...
+    SimpleRexRemap remap = new SimpleRexRemap(indexContext.getScan(), newRowType, indexContext.getScan().getCluster().getRexBuilder());
+    remap.setExpressionMap(functionInfo.getExprMap());
+    return remap.rewriteWithMap(toRewriteRex, mapRexExpr);
+  }
+
+  public static RexNode getLeadingPrefixMap(Map<LogicalExpression, RexNode> leadingPrefixMap,
+                                  List<LogicalExpression> indexCols,
+                                  IndexConditionInfo.Builder builder, RexNode condition) {
+    boolean prefix = true;
+    int i=0;
```
vdiravka: spaces

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexPlanUtils.java
```diff
@@ -0,0 +1,872 @@
...
+
+    RexNode initCondition = condition.isAlwaysTrue() ? null : condition;
+    while (prefix && i < indexCols.size()) {
+      LogicalExpression p = indexCols.get(i++);
+      List<LogicalExpression> prefixCol = ImmutableList.of(p);
+      IndexConditionInfo info = builder.indexConditionRelatedToFields(prefixCol, initCondition);
+      if(info != null && info.hasIndexCol) {
```
vdiravka: space

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/IndexSelector.java
```diff
@@ -0,0 +1,766 @@
...
+              indexContext.getScan(), functionInfo,indexContext);
+    }
+
+    if (indexContext.hasUpperProject()) {
+      inputCollation =
+          IndexPlanUtils.buildCollationProject(indexContext.getUpperProject().getProjects(), indexContext.getLowerProject(),
+              indexContext.getScan(), functionInfo, indexContext);
+    }
+
+    if ( (inputCollation != null) &&
```
vdiravka: formatting

protocol/src/main/java/org/apache/drill/exec/proto/beans/CoreOperatorType.java
```diff
@@ -78,7 +78,8 @@
     SEQUENCE_SUB_SCAN(53),
     PARTITION_LIMIT(54),
     PCAPNG_SUB_SCAN(55),
-    RUNTIME_FILTER(56);
+    RUNTIME_FILTER(56),
+    ROWKEY_JOIN(57);
```
vdiravka: Please regenerate protobuf files for C++ native client as well

pom.xml
```diff
@@ -53,8 +53,8 @@
     <jackson.version>2.9.5</jackson.version>
     <jackson.databind.version>2.9.5</jackson.databind.version>
     <zookeeper.version>3.4.12</zookeeper.version>
-    <mapr.release.version>5.2.1-mapr</mapr.release.version>
-    <ojai.version>1.1</ojai.version>
+    <mapr.release.version>6.0.1-mapr</mapr.release.version>
```
vdiravka: Use 6.1.0-mapr version
It will be introduced after merging this PR:
https://github.com/apache/drill/pull/1489/files#diff-600376dffeb79835ede4a0b285078036R56

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/PlannerSettings.java
```diff
@@ -114,6 +114,28 @@
   public static final String UNIONALL_DISTRIBUTE_KEY = "planner.enable_unionall_distribute";
   public static final BooleanValidator UNIONALL_DISTRIBUTE = new BooleanValidator(UNIONALL_DISTRIBUTE_KEY, null);
 
+  // ------------------------------------------- Index planning related options BEGIN --------------------------------------------------------------
+  public static final String USE_SIMPLE_OPTIMIZER_KEY = "planner.use_simple_optimizer";
+  public static final BooleanValidator USE_SIMPLE_OPTIMIZER = new BooleanValidator(USE_SIMPLE_OPTIMIZER_KEY, null);
+  public static final BooleanValidator INDEX_PLANNING = new BooleanValidator("planner.enable_index_planning", null);
+  public static final BooleanValidator ENABLE_STATS = new BooleanValidator("planner.enable_statistics", null);
```
vdiravka: Is it used for obtaining MapR-DB and Parquet table statistics?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillOptiq.java
```diff
@@ -22,7 +22,7 @@
 import java.util.LinkedList;
 import java.util.List;
 
-import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
+import org.apache.calcite.rel.type.RelDataType;
```
vdiravka: Optiq is an old name of Calcite. Does it make sense to rename this class?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/InvalidIndexDefinitionException.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.planner.index;
+
+import org.apache.drill.common.exceptions.DrillRuntimeException;
+
+public class InvalidIndexDefinitionException extends DrillRuntimeException {
```
vdiravka: Consider adding some description to the exception

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/SimpleRexRemap.java
```diff
@@ -0,0 +1,300 @@
...
+        if (operands.get(0) instanceof RexLiteral) {
+          oplit = (RexLiteral) operands.get(0);
+          if(oplit.getTypeName() == SqlTypeName.CHAR) {
+            newLiteral = builder.makeLiteral(((NlsString) oplit.getValue()).getValue() + "%");
+            input = operands.get(1);
+          }
+        }
+        else if (operands.get(1) instanceof RexLiteral) {
+          oplit = (RexLiteral) operands.get(1);
+          if(oplit.getTypeName() == SqlTypeName.CHAR) {
```
vdiravka: space

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/generators/IndexIntersectPlanGenerator.java
```diff
@@ -0,0 +1,350 @@
...
+    final RelDataTypeField rightRowKeyField = indexPlan.getRowType().getFieldList().get(0);
+    final RelNode rangeDistRight = createRangeDistRight(indexPlan, rightRowKeyField,
+        (DbGroupScan)IndexPlanUtils.getGroupScan(origScan));
+
+    //now with index plan constructed, build plan of left(probe) side to use restricted db scan
+
+    Pair<RelNode, DbGroupScan> leftRelAndScan = buildRestrictedDBScan(remnant, isAnyIndexAsync);
+
+    RelNode finalRel = buildRowKeyJoin(leftRelAndScan.left, rangeDistRight, true, JoinControl.DEFAULT);
+    if ( upperProject != null) {
```
vdiravka: formatting

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/rules/AbstractMatchFunction.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public static boolean containsStar(DbGroupScan dbscan) {
+    for (SchemaPath column : dbscan.getColumns()) {
+      if (column.getRootSegment().getPath().startsWith("*")) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
```
vdiravka: new line

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/rules/MatchFunction.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.apache.drill.exec.planner.index.rules;
+
+import org.apache.calcite.plan.RelOptRuleCall;
+
+public interface MatchFunction<T> {
+  boolean match(RelOptRuleCall call);
+  T onMatch(RelOptRuleCall call);
+}
```
vdiravka: new line

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
```diff
@@ -166,4 +169,25 @@ public void onMatch(RelOptRuleCall call) {
     return list;
   }
 
+  public static Project replace(Project topProject, Project bottomProject) {
+    final List<RexNode> newProjects =
+        RelOptUtil.pushPastProject(topProject.getProjects(), bottomProject);
+
+    // replace the two projects with a combined projection
+    if(topProject instanceof DrillProjectRel) {
```
vdiravka: space

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
```diff
@@ -166,4 +169,25 @@ public void onMatch(RelOptRuleCall call) {
     return list;
   }
 
+  public static Project replace(Project topProject, Project bottomProject) {
```
vdiravka: Can `ProjectAllowDup` and `Project` rels be merged?
For instance:
`00-01      ProjectAllowDup(MONTHID=[$0], NOFOCONTRACTS=[$1]) : rowType = RecordType(INTEGER MONTHID, INTEGER NOFOCONTRACTS)`
`00-02        Project(MonthId=[CAST($1):INTEGER], No of Contracts=[CAST($2):INTEGER]) : rowType = RecordType(INTEGER MonthId, INTEGER No of Contracts)`

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
```diff
@@ -517,6 +388,15 @@ private static FieldPath getFieldPathForProjection(SchemaPath column) {
     return new FieldPath(child);
   }
 
+  public static boolean includesIdField(Collection<FieldPath> projected) {
+    return Iterables.tryFind(projected, new Predicate<FieldPath>() {
```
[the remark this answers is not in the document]
amansinha100: @vdiravka could you elaborate on this a bit ?  I haven't done much with lambda expressions in Java 8, but if you show how to rewrite this statement and assuming it has an advantage over the existing implementation, I would be happy to.  Thanks.

amansinha100:
@vdiravka I think I have addressed all the formatting issues in commit 5e0fa72 and added some missing javadocs (btw, thanks for pointing out the formatting issues..I don't think I introduced them but    should have caught them before creating the PR).  I will go ahead and 'hide' your comments related to these and leave your other comments open.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/DrillIndexDescriptor.java
```diff
@@ -0,0 +1,110 @@
...
+      }
+      return (IndexGroupScan)scan;
+    }
+    catch(IOException e) {
+      logger.error("Error in getIndexGroupScan ", e);
+    }
+    return null;
+  }
+
+  public void attach(String storageName, DrillTable inTable) {
```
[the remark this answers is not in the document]
amansinha100: Removed attach() since it was not being used.  Added Javadoc or `@Override` for few others.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushProjectIntoScan.java
```diff
@@ -0,0 +1,145 @@
...
+import org.apache.drill.exec.planner.physical.ProjectPrel;
+import org.apache.drill.exec.planner.physical.ScanPrel;
+import org.apache.drill.exec.store.StoragePluginOptimizerRule;
+import org.apache.drill.exec.store.mapr.db.binary.BinaryTableGroupScan;
+import org.apache.drill.exec.store.mapr.db.json.JsonTableGroupScan;
+import org.apache.drill.exec.util.Utilities;
+
+import java.util.List;
+
+public abstract class MapRDBPushProjectIntoScan extends StoragePluginOptimizerRule {
```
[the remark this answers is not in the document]
amansinha100: The main difference is that DrillPushProjectIntoScanRule is applied during logical planning phase whereas the MapRDBPushProjectIntoScan is applied during physical planning because we want to ensure that even after new physical plans are created (such as index based plans), the projection pushdown is done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillOptiq.java
amansinha100: It probably does..can you create a JIRA for the renaming since it is independent of this PR.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/generators/IndexIntersectPlanGenerator.java
```diff
@@ -0,0 +1,350 @@
...
+    List<Integer> rightJoinKeys = ImmutableList.of(rightRowKeyIdx);
+
+    logger.trace(String.format(
+        "buildRowKeyJoin: leftIdx: %d, rightIdx: %d",
+        leftRowKeyIdx, rightRowKeyIdx));
+    RexNode joinCondition =
+        RelOptUtil.createEquiJoinCondition(left, leftJoinKeys,
+            right, rightJoinKeys, builder);
+
+    if (isRowKeyJoin == true) {
```
Ben-Zvi: How about `if (isRowKeyJoin) {` instead ?

contrib/format-maprdb/src/main/java/org/apache/drill/exec/planner/index/MapRDBFunctionalIndexInfo.java
```diff
@@ -0,0 +1,168 @@
...
+  //map of SchemaPath involved in a functional field
+  private Map<LogicalExpression, Set<SchemaPath>> pathsInExpr;
+
+  private Set<SchemaPath> newPathsForIndexedFunction;
+
+  private Set<SchemaPath> allPathsInFunction;
+
+  public MapRDBFunctionalIndexInfo(IndexDescriptor indexDesc) {
+    this.indexDesc = indexDesc;
+    columnToConvert = Maps.newHashMap();
```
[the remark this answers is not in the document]
vdiravka: You can use it, but from the last Guava version used by Drill (23.0 version) there is a note for this factory method:
> <p><b>Note for Java 7 and later:</b> this method is now unnecessary and
> should be treated as deprecated. Instead, use the {@code HashMap}
> constructor directly, taking advantage of the new
> <a href="http://goo.gl/iz2Wi">"diamond" syntax</a>.
https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/collect/Maps.html#newHashMap--

There was no this note in the previous Guava version used by Drill (19.0 version), therefore some usages of this method are left.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushLimitIntoScan.java
```diff
@@ -0,0 +1,203 @@
...
+import org.apache.drill.exec.planner.physical.ProjectPrel;
+import org.apache.drill.exec.planner.physical.RowKeyJoinPrel;
+import org.apache.drill.exec.planner.physical.ScanPrel;
+import org.apache.drill.exec.store.StoragePluginOptimizerRule;
+import org.apache.drill.exec.store.hbase.HBaseScanSpec;
+import org.apache.drill.exec.store.mapr.db.binary.BinaryTableGroupScan;
+import org.apache.drill.exec.store.mapr.db.json.JsonTableGroupScan;
+import org.apache.drill.exec.store.mapr.db.json.RestrictedJsonTableGroupScan;
+
+public abstract class MapRDBPushLimitIntoScan extends StoragePluginOptimizerRule {
```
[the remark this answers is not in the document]
amansinha100: The `DrillPushLimitToScanRule` works for the logical plan.  We introduced the MapR-DB specific rule to apply similar pushdown for the physical plan because after index planning is done additional pushdowns may be possible.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushProjectIntoScan.java
```diff
@@ -0,0 +1,145 @@
...
+      RelOptHelper.some(ProjectPrel.class, RelOptHelper.any(ScanPrel.class)), "MapRDBPushProjIntoScan:Proj_On_Scan") {
+    @Override
+    public void onMatch(RelOptRuleCall call) {
+      final ScanPrel scan = (ScanPrel) call.rel(1);
+      final ProjectPrel project = (ProjectPrel) call.rel(0);
+      if (!(scan.getGroupScan() instanceof MapRDBGroupScan)) {
+        return;
+      }
+      doPushProjectIntoGroupScan(call, project, scan, (MapRDBGroupScan) scan.getGroupScan());
+      if (scan.getGroupScan() instanceof BinaryTableGroupScan) {
```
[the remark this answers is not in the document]
amansinha100: I have made the changes here and rest of this class to only check for JsonTableGroupScan.  Binary table's projection pushdown was already handled by the logical planning rule `DrillPushProjectIntoScan`.   In future, if we want to extend this rule to binary tables, we would need to ensure it is tested.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushFilterIntoScan.java
```diff
@@ -187,7 +186,7 @@ protected void doPushFilterIntoBinaryGroupScan(final RelOptRuleCall call,
                                                                         groupScan.getTableStats());
     newGroupsScan.setFilterPushedDown(true);
 
-    final ScanPrel newScanPrel = ScanPrel.create(scan, filter.getTraitSet(), newGroupsScan, scan.getRowType());
+    final ScanPrel newScanPrel = ScanPrel.create(scan, filter.getTraitSet(), newGroupsScan, scan.getRowType(), scan.getTable());
```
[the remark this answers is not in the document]
amansinha100: Changed to use the constructor.  @vdiravka why not remove the `create` method altogether ?

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBFormatPlugin.java
```diff
@@ -57,34 +64,74 @@ public MapRDBFormatPlugin(String name, DrillbitContext context, Configuration fs
     hbaseConf = HBaseConfiguration.create(fsConf);
     hbaseConf.set(ConnectionFactory.DEFAULT_DB, ConnectionFactory.MAPR_ENGINE2);
     connection = ConnectionFactory.createConnection(hbaseConf);
+    jsonTableCache = new MapRDBTableCache(context.getConfig());
+    int scanRangeSizeMBConfig = context.getConfig().getInt(PluginConstants.JSON_TABLE_SCAN_SIZE_MB);
+    if (scanRangeSizeMBConfig < 32 || scanRangeSizeMBConfig > 8192) {
```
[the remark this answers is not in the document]
amansinha100: Done

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/InvalidIndexDefinitionException.java
amansinha100: Done

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/PlannerSettings.java
amansinha100: I will let @gparai elaborate on this but my understanding is that by keeping a general name such as `planner.enable_statistics` option we could potentially use it for any type of data source (including Parquet) once the broader stats capability is added to Drill.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/index/generators/IndexIntersectPlanGenerator.java
amansinha100: done

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushFilterIntoScan.java
```diff
@@ -137,11 +137,10 @@ protected void doPushFilterIntoJsonGroupScan(RelOptRuleCall call,
       return; //no filter pushdown ==> No transformation.
     }
 
-    // clone the groupScan with the newScanSpec.
-    final JsonTableGroupScan newGroupsScan = groupScan.clone(newScanSpec);
+    final JsonTableGroupScan newGroupsScan = (JsonTableGroupScan) groupScan.clone(newScanSpec);
```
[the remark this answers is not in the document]
amansinha100: Currently, the `clone()` method returns an instance of `GroupScan`, so this cast is needed. Did you mean change the original `clone` implementation itself ?  That will require some additional changes since both binary table and json table implement it.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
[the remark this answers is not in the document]
amansinha100: Since this is a syntactic modification, I am inclined to leave it as-is .. partly because I might cause an unintentional side effect/regression and partly because such functional utilities appear elsewhere in the Drill and Calcite code.  Ideally, if we want to move to Lambdas, we should re-visit all such implementations.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/PlannerSettings.java
gparai: Yes, that was the intent - allow statistics use for any source with one option.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/JoinControl.java
amansinha100: @Ben-Zvi I have incorporated the joinControl logic as part of the hash join probe phase.  Please see changes in [1].  I haven't done it for build phase since it will be superseded by the semi-join changes.  Can you pls review this change ?

[1] https://github.com/apache/drill/pull/1466/commits/5330fd684a20587e9e860d7894e3d3602a6d0495

contrib/format-maprdb/src/test/java/com/mapr/drill/maprdb/tests/index/IndexPlanTest.java
```diff
@@ -0,0 +1,1715 @@
...
+        .run();
+  }
+
+  @Test
+  public void testWithFilterGroupBy() throws Exception {
+    String query = " select t1.driverlicense from hbase.`index_test_primary` t1" +
+            " where t1.driverlicense > 100000001 group by t1.driverlicense limit 2";
+    try {
+      test(defaultHavingIndexPlan);
+      test(disableHashAgg);
```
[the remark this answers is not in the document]
amansinha100: @vdiravka since this is a fairly large test file (close to 75 tests) this suggestion and couple of other suggestions for the test changes seem nice-to-have but perhaps not must-have ?  In the interest of getting the main code changes merged, would it be okay to defer this ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
amansinha100: They can be merged into the top level Project that allows duplicates...however, the purpose of the `replace()` method here is to simply allow the caller to replace a project with another with the assumption that callers know exactly what they are doing.  This is not applying the full fledged `DrillMergeProjectRule`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/Partitioner.java
```diff
@@ -29,6 +29,8 @@
 import org.apache.drill.exec.record.RecordBatch;
 
 public interface Partitioner {
+  int DEFAULT_RECORD_BATCH_SIZE = (1 << 10) - 1;
```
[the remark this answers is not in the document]
amansinha100: The DEFAULT_RECORD_BATCH_SIZE is already in the current master branch here [1].  I simply moved it into the `Partitioner` interface. 

[1] https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/PartitionerTemplate.java#L58

chunhui-shi:
+1. I think you already addressed all the comments and just need to solve the rebase conflicts.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -568,25 +603,13 @@ public IterOutcome innerNext() {
...
-          probeBatch.kill(true);
-          leftUpstream = next(HashJoinHelper.LEFT_INPUT, probeBatch);
-          while (leftUpstream == IterOutcome.OK_NEW_SCHEMA || leftUpstream == IterOutcome.OK) {
-            for (final VectorWrapper<?> wrapper : probeBatch) {
-              wrapper.getValueVector().clear();
-            }
-            leftUpstream = next(HashJoinHelper.LEFT_INPUT, probeBatch);
-          }
-        }
+        drainLeft();
```
Ben-Zvi: Note the expected merge conflict here (from DRILL-6755 = #1480):  The new code defined similar methods (e.g. `killAndDrainLeftUpstream()`). So please follow the #1480 new code.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -1212,6 +1238,81 @@ private void updateStats() {
     this.stats.setLongStat(Metric.SPILLED_PARTITIONS, numSpilled);
   }
 
+  private void drainLeft() {
```
Ben-Zvi: Follow on (after #1480) - these two methods can be eliminated ( `drainLeft()` and `drainRight()` )

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
amansinha100: I have resolved this merge conflict after rebasing on my local branch.  I haven't pushed it remote yet (need to see if I can avoid force-push).

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushProjectIntoScan.java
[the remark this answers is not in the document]
vdiravka: It makes sense, thank you for clarifications.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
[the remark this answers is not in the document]
vdiravka: Agreed. 
Vova was doing that, but has not finished in the result.

contrib/format-maprdb/src/test/java/com/mapr/drill/maprdb/tests/index/IndexPlanTest.java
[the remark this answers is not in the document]
vdiravka: Surely, there are a lot of places in Drill unit tests, which can be refactored. At least `BaseTestQuery` is marked as `@Deprecated`, but is widely used.
Possibly we can gather all this small issues and make the global refactoring in the process of some task, for example in the process of updating JUnit to 5 version.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/Partitioner.java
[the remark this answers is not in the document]
vdiravka: I see, thanks for clarification.
Is that comment regarding `recordCount` isn't necessary anymore?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
vdiravka: It makes sense. A small javadoc can be useful here.

protocol/src/main/java/org/apache/drill/exec/proto/beans/CoreOperatorType.java
amansinha100: I have regenerated the protobuf files for the native C++ client.  (As an aside, the steps for doing this seems complicated unlike the Java version. In either case, we really should not need to rebuild the client side if a new operator has been added on the server side).

contrib/format-maprdb/src/main/java/org/apache/drill/exec/planner/index/MapRDBFunctionalIndexInfo.java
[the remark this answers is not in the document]
amansinha100: The checkstyle-validation had identified a bunch of non-shaded Guava classes in the core classes when I created the PR and I had fixed those by using the shaded version. But oddly, it did not complain for the MapR-DB plugin classes. Does this enforcement not occur for the plugins ? I am not sure..but in any case I looked through the new code added in MapR-DB plugin and have modified the imports to use shaded Guava classes.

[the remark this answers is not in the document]
amansinha100: Ah, that explains it.  Presumably we will fix it for 1.15.  Thanks.

pom.xml
amansinha100: For this version update, I will let the the PR you referenced [1] handle it.  There are functional test failures with that PR that need to be addressed. 

[1] https://github.com/apache/drill/pull/1489

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
```diff
@@ -535,7 +535,10 @@ public FragmentContext getContext() {
 
   @Override
   public BatchSchema getSchema() {
-    return outgoingContainer.getSchema();
+    if (outgoingContainer.hasSchema()) {
+      return outgoingContainer.getSchema();
+    }
+    return null;
```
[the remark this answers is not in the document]
amansinha100: The null value could occur for example when even though the `VectorContainer` has vectors but the operator has not yet called `buildSchema()` on the container.  There's a `Preconditions.checkNotNull` in `VectorContainer.getSchema()`.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdSelectivity.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.util.List;
+
+public class DrillRelMdSelectivity extends RelMdSelectivity {
+  private static final DrillRelMdSelectivity INSTANCE = new DrillRelMdSelectivity();
+
+  public static final RelMetadataProvider SOURCE = ReflectiveRelMetadataProvider.reflectiveSource(BuiltInMethod.SELECTIVITY.method, INSTANCE);
+
+
+  public Double getSelectivity(RelNode rel, RexNode predicate) {
```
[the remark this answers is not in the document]
amansinha100: There were customizations needed for RelSubset, DrillScanRel and ScanPrel selectivities.  For all other nodes, we are calling the `super.getSelectivity()`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/partitionsender/Partitioner.java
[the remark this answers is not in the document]
amansinha100: I think that was an omission.  I have re-inserted the comment but with some modifications because the batch-sizing project now allows operators to set the output batch size in terms of Mbytes rather than `recordCount`.  It is not yet applied across the board, so this `DEFAULT_RECORD_BATCH_SIZE` is still relevant for the exchange operators.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/MapRDBPushProjectIntoScan.java
```diff
@@ -0,0 +1,145 @@
...
+      final ProjectPrel newProj =
+          new ProjectPrel(project.getCluster(),
+              project.getTraitSet().plus(Prel.DRILL_PHYSICAL),
+              newScan,
+              newProjects,
+              project.getRowType());
+
+      if (ProjectRemoveRule.isTrivial(newProj) &&
+          // the old project did not involve any column renaming
+          sameRowTypeProjectionsFields(project.getRowType(), newScan.getRowType())) {
```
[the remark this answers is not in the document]
amansinha100: This may have to do with the fact that `MapRDBPushProjectIntoScan` is applied during physical planning unlike the `DrillPushProjectIntoScanRule` so there may be some differences in how the pushdown occurs.  We would have added the extra check `sameRowTypeProjectionsFields` to prevent an incorrect pushdown, although I don't recall exact details now.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillMergeProjectRule.java
amansinha100: Added javadoc.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/BroadcastExchangePrel.java
```diff
@@ -57,7 +57,9 @@ public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
 
     final int  rowWidth = child.getRowType().getFieldCount() * DrillCostBase.AVG_FIELD_WIDTH;
     final double cpuCost = broadcastFactor * DrillCostBase.SVR_CPU_COST * inputRows;
-    final double networkCost = broadcastFactor * DrillCostBase.BYTE_NETWORK_COST * inputRows * rowWidth * numEndPoints;
+
+    //we assume localhost network cost is 1/10 of regular network cost
+    final double networkCost = broadcastFactor * DrillCostBase.BYTE_NETWORK_COST * inputRows * rowWidth * (numEndPoints - 0.9);
```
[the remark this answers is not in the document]
amansinha100: @gparai,  forgot to respond to this.  The cost formula is: 
    (cost of broadcasting num_bytes to N - 1 nodes)  +  (cost of local broadcast to all minor fragments on my own node)
   = (C * num_bytes * (N - 1) )  +  (C * num_bytes * 0.1)   where the 0.1 factor comes from the assumption that local broadcast is 10% network cost of the remote broadcast. 
  = C * num_bytes * (N - 0.9)

While the formula seems reasonable, it is biasing the cost in favor of Broadcast compared to HashPartition.  We should re-visit this and ideally a similar change should be done for HashPartition also.  I don't recall the exact use case which motivated the change.. it may have been the Index Intersection.  I can create a JIRA to re-visit this.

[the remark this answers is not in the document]
amansinha100: I have added comments in the code regarding this.

amansinha100:
I will close this PR and open a separate one that is rebased on apache master.  The reason is after rebase and squashing some commits, I am having to force-push but that may clobber some of the review comments which I want to avoid.

amansinha100:
Opened the new PR:  https://github.com/apache/drill/pull/1512   to do the final merge.

PR #1468, opened 2018-09-17
DRILL-6630: Extra spaces are ignored while publishing results in Dril…

agozhiy:
…l Web UI

Added a style option to the table.


agozhiy:
Before the patch:
screenshot from 2018-09-18 11-38-29
After the patch:
screenshot from 2018-09-18 11-37-42

PR #1470, opened 2018-09-20
DRILL-6746: Query can hang when PartitionSender task thread sees a co…

sohami:
…nnection failure while sending data batches to remote fragment


sohami:
@sachouche - Please help to review this PR.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/SpoolingRawBatchBuffer.java
```diff
@@ -119,6 +119,19 @@ public RawFragmentBatch take() throws IOException, InterruptedException {
       return buffer.take().get();
     }
 
+    @Override
+    public RawFragmentBatch poll(long timeout, TimeUnit timeUnit) throws InterruptedException, IOException {
+      RawFragmentBatchWrapper batchWrapper = buffer.poll(timeout, timeUnit);
+      if (batchWrapper != null) {
+        try {
+          return batchWrapper.get();
+        } catch (InterruptedException e) {
```
sachouche: You should not catch InterruptedException as it is being handled by caller; the handling is also different:
- Upper code is having more logic on how to deal with an interrupted exception
- This code is erasing the interrupted flag whereas caller code doesn't

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/UnlimitedRawBatchBuffer.java
```diff
@@ -63,6 +64,15 @@ public RawFragmentBatch take() throws IOException, InterruptedException {
       return batch;
     }
 
+    @Override
+    public RawFragmentBatch poll(long timeout, TimeUnit timeUnit) throws InterruptedException, IOException {
+      RawFragmentBatch batch = buffer.poll(timeout, timeUnit);
```
sachouche: Does poll return null when timeout is reached? if not, then you should handle timeout exception.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/BaseRawBatchBuffer.java
```diff
@@ -167,7 +169,25 @@ public RawFragmentBatch getNext() throws IOException {
...
+        // here waiting for next batch to arrive. Meanwhile when next batch arrived and was enqueued it sees
+        // FragmentExecutor failure state and doesn't enqueue the batch and cleans up the buffer queue. Hence this
+        // thread will stuck forever. So we pool for 5 seconds until we get a batch or FragmentExecutor state is in
+        // error condition.
+        while (b == null) {
+          b = bufferQueue.poll(5, TimeUnit.SECONDS);
+          if (!context.getExecutorState().shouldContinue()) {
+            kill(context);
+            if (b != null) {
+              assertAckSent(b);
```
sachouche: can this assert fail when the minor fragment is getting closed? My point, assertions should be avoided during cleanup path.

exec/java-exec/src/main/java/org/apache/drill/exec/work/batch/SpoolingRawBatchBuffer.java
sohami: Thanks for catching this. Specifically for spooling buffer implementation even `poll()` method was consuming the InterruptedException. I have made both polls to throw `InterruptedException`.

sohami:
@sachouche - Thanks for review. I have address the review comments, please take another look.

PR #1472, opened 2018-09-20
DRILL-6747: Fixed IOB in HashJoin for inner and right joins on empty probe side

ilooner:
- Fixed IOB in HashJoin for inner and right joins on empty probe side.
 - Documentated expected behavior of RecordBatch after returning NONE.
 - Made UnorderedReceiver obey contract after returning NONE.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatch.java
@@ -115,6 +115,14 @@ public interface RecordBatch extends VectorAccessible {
      *   This value will be returned only after {@link #OK_NEW_SCHEMA} has been
      *   returned at least once (not necessarily <em>immediately</em> after).
      * </p>
+     * <p>
+     *   Also after a RecordBatch returns NONE a RecordBatch should:
+     *   <ul>
+     *     <li>Contain the last valid schema seen by the operator.</li>
+     *     <li>Contain a VectorContainer with empty columns corresponding to the last valid schema.</li>
+     *     <li>Return a record count of 0.</li>
+     *   </ul>
+     * </p>
      */
     NONE(false),
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
@@ -174,7 +174,7 @@ public class UnorderedReceiverBatch implements CloseableRecordBatch {
       first = false;
 
       if (batch == null) {
-        batchLoader.clear();
+        batchLoader.zero();
         if (!context.getExecutorState().shouldContinue()) {
           return IterOutcome.STOP;
         }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchLoader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchLoader.java
@@ -277,6 +277,14 @@ public class RecordBatchLoader implements VectorAccessible, Iterable<VectorWrapp
   public void resetRecordCount() { valueCount = 0; }
 
   /**
+   * Removes an data from the loader, but maintains the schema and empty vectors.
+   */
+  public void zero() {
+    container.zeroVectors();
+    resetRecordCount();
+  }
+
+  /**
    * Clears this loader, which clears the internal vector container (see
    * {@link VectorContainer#clear}) and resets the record count to zero.
    */


ilooner:
@Ben-Zvi

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatch.java
```diff
@@ -115,6 +115,14 @@
      *   This value will be returned only after {@link #OK_NEW_SCHEMA} has been
      *   returned at least once (not necessarily <em>immediately</em> after).
      * </p>
+     * <p>
+     *   Also after a RecordBatch returns NONE a RecordBatch should:
+     *   <ul>
+     *     <li>Contain the last valid schema seen by the operator.</li>
+     *     <li>Contain an empty VectorContainer with empty columns corresponding to the last valid schema.</li>
```
Ben-Zvi: Is `empty` required ?

ilooner: Fixed thanks for catching and thanks for the review!

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatch.java
```diff
@@ -115,6 +115,14 @@
...
+     * <p>
+     *   Also after a RecordBatch returns NONE a RecordBatch should:
+     *   <ul>
+     *     <li>Contain the last valid schema seen by the operator.</li>
+     *     <li>Contain a VectorContainer with empty columns corresponding to the last valid schema.</li>
+     *     <li>Return a record count of 0.</li>
+     *   </ul>
+     * </p>
      */
     NONE(false),
```
sohami: NONE doesn't guarantee a valid container associated with it. No operator (or batch) should rely on VectorContainer from upstream once NONE outcome is returned. Why is this change in behavior needed ?

ilooner: @sohami I agree that guarantee was not provided before. However, I am proposing to change the contract in the way described in the comment. Note no data would be associated with a record batch after it returns NONE. It would only have a VectorContainer with **EMPTY** columns corresponding to the last seen valid schema, and return a record count of NONE. In practice almost all operators already do this (except they don't always zero their container).

The advantage of adding this new guarantee is that downstream operators do not have to do book keeping when the upstream can do it for us very easily.

The details of why this was done are in the JIRA. There were many ways to fix the issue in HashJoin but this is by far the cleanest, and less risky since most of the code in HashJoin has no unit tests. It also has the advantage of clarifying the behavior of RecordBatches for future operator implementations. So I felt this approach was a win win and the best way to go about fixing the issue. However, there should be a follow up Jira to validate the new contract I am proposing for the rest of the operators, not just unordered reciever.

sohami: I agree that it will not affect current operators behavior since anyways none of operator use the RecordBatch information once NONE is seen. It takes this outcome as End of Stream and completes all its processing.

My concern is if you do this then looks like with NONE outcome in HashJoin case we are using the batch information to make some decision. In that sense it is treating NONE similar to OK with empty batch ? Based on JIRA explanation why didn't HashJoin used the schema related information from batch with OK_NEW_SCHEMA since NONE will come atleast after OK_NEW_SCHEMA?

ilooner: HashJoin needs to first sniff the first data holding batch from the probe side in order to compute stats for the memory calculations used to spill. If the probe side is empty we will hit NONE. This is unavoidable and this is where the problem begins.

In the case of a right join HashJoin still builds HashTables, and the HashTables require build and probe batches to be updated simultaneously before being built. When the build and probe sides are updated the Hashtable expects a vector container with valid columns. If the columns are not present an IOB is thrown.

Then later the probing logic is called in HashJoinProbeTemplate, which expects the upstream probeBatch to return a valid record count.

There are a few ways to fix this problem:

  1. We can refactor the HashTable to update probe and build sides independently. This will have some corner cases though, and the HashTable has no unit tests, so it's risky. Or we can remove the code generation from the HashTable, which will also be somewhat time consuming. Or we could change the join logic to not build the HashTable in this case, which is another major change. Then we can also refactor HashJoinProbeTemplate to handle an empty probe side gracefully, but again HashJoinProbeTemplate is not unit tested so this will be time consuming.

  2. We leave the HashTable and HashJoinProbeTemplate as is, but use the last schema we got from the probe side to build dummy empty vector containers for building the HashTable, and for probing in HashJoinProbeTemplate. IMO this is a hack. Also there are some corner cases because some Drill operators violate the contract that they must first send OK_NEW_SCHEMA and then NONE. I've observed in some unit tests some readers directly skip to sending NONE without OK_NEW_SCHEMA if there is no data.

 3. We define the state of a RecordBatch after it returns NONE to have a record count of zero, the last valid schema, and an empty VectorContainer. This would be defined should a downstream operator choose to query it. This resolves the issue in HashJoin without doing hacks and risky refactoring in a rush. It also defines state that was previously undefined, which IMO is a good thing. I would also do follow up PRs to make all operators obey this behavior, since it will be trivial to do.

 -  Of the three options, option 1 is the strictly correct thing to do if you don't consider time costs. However, due to the technical debt we've accumulated it will be time consuming to do.
 - Option 2 is a hack IMO and just makes things more confusing for the future generation of Drillers.
 - Option 3 is a win win. We get a clean resolution to the issue, and we define the state of operators after they return NONE. Defining state that was previously undefined is always a good thing, especially when it simplifies the logic of the system. And I can go through the operators and make sure they do this, it should be a trivial change involving zeroing VectorContainers before returning NONE and updating record counts to be 0.

ilooner: We discussed offline and agreed to go with this solution for now.

PR #1476, opened 2018-09-21
DRILL-5782: Web UI: do not attempt to build visualized plan when plan…

agozhiy:
… is absent

Added checks for undefined values. Now there are no js errors and also a JSON profile is now displayed for incorrect queries.


arina-ielchiieva:
+1

PR #1477, opened 2018-09-21
DRILL-6753: Fix show files command to return result the same way as before

arina-ielchiieva:
1. Add ACCESS_TIME column.
2. Re-write show files command to return result using ShowFilesCommandResult to maintain backward compatibility.

Details in [DRILL-6753](https://issues.apache.org/jira/browse/DRILL-6753).


exec/java-exec/src/test/java/org/apache/drill/exec/store/ischema/TestFilesTable.java
```diff
@@ -86,62 +102,46 @@ public void testSelectWithRecursion() throws Exception {
...
+      assertTrue(querySummary.succeeded());
+      assertEquals(2, querySummary.recordCount());
+      // option has no effect
       client.alterSession(ExecConstants.LIST_FILES_RECURSIVELY, true);
-      client.testBuilder()
-          .sqlQuery("show files in dfs.`files`.folder")
-          .unOrdered()
-          .sqlBaselineQuery("select * from INFORMATION_SCHEMA.`FILES` where schema_name = 'dfs.files' and relative_path like 'folder/%'")
-          .go();
+      assertTrue(querySummary.succeeded());
```
vvysotskyi: Looks like after the setting the option, the query wasn't run again and the previous result is checking.

arina-ielchiieva: @vvysotskyi nice catch, thanks. Fixed.

PR #1480, opened 2018-09-26
DRILL-6755: Avoid building Hash Table for inner/left join when probe side is empty

Ben-Zvi:
This PR is split into two commits to help the reviewers:
(1) Preparations and cleanups: No change to the code's logic, only new definitions and minor cleaning.
(2) The actual change: Define the flag `skipHashTableBuild` and use it to skip initial hash-table setting, and later to kill the build side upstream and skip the hash table build. (This also saves the useless work in the upstream; e.g., more scanning).
    Also included is a test of an inner join, with a NONE probe input, and multiple input batches on the build side.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
@@ -117,6 +117,7 @@ public class HashJoinBatch extends AbstractBinaryRecordBatch<HashJoinPOP> {
   private final JoinRelType joinType;
   private boolean joinIsLeftOrFull;
   private boolean joinIsRightOrFull;
+  private boolean skipHashTableBuild; // when outer side is empty, and the join is inner or left (see DRILL-6755)
 
   // Join conditions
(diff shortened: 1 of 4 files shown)


Ben-Zvi:
Indeed this PR only addresses the "NONE 1st" case. As for the "NONE 2nd" case (following a 1st OK_NEW_SCHEMA) - this would open up the can of worms we saw before - early sniffing the probe side cranks the actual execution there (e.g., another Hash Join below an Exchange) while other parts of the query are still in schema discovery.

Not sure if in all cases where an operator finds it has no data to return during schema discovery, a NONE is returned first. Note that the NONE does have a schema (e.g., so a Right Join can return nulls for those missing columns). Maybe we can increase the impact of this PR by checking all these cases, and documenting/testing that our protocol allows for a NONE to be returned first (i.e. during schema discovery).

ilooner:
@Ben-Zvi This was missed during the last batch commit for some reason. Could you fix the merge conflict and merge it?

Ben-Zvi:
This PR should be closed. Apparently was not closed due to a typo in the commit message:
```
clsoes #1480
```

PR #1481, opened 2018-09-26
DRILL-6763: Codegen optimization of SQL functions with constant values

lushuifeng:
Details in [DRILL-6763](https://issues.apache.org/jira/projects/DRILL/issues/DRILL-6763):

Here is the descriptions of the change:
<del>1. add system option `exec.optimize_function_compilation` to toggle to state of this functionality.</del>
2. codegen is changed by declaring setter method in EvaluationVisitor#visitXXXconstants
3. the member declared in step2 is initialized when the instance of the class is created in XXBatch and others
4. attachment is the code of the same query mentioned in DRILL-6763 generated by setting the value of exec.optimize_function_compilation to true
@arina-ielchiieva @vdiravka Would you please take a look?
What kind of unit tests should be added?
[query.txt](https://github.com/apache/drill/files/2418035/query.txt)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
@@ -22,10 +22,13 @@ import static org.apache.drill.exec.compile.sig.GeneratorMapping.GM;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 
+import io.netty.buffer.DrillBuf;
+import org.apache.calcite.util.Pair;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.expression.LogicalExpression;
 import org.apache.drill.common.types.TypeProtos;
@@ -39,7 +42,9 @@ import org.apache.drill.exec.compile.sig.MappingSet;
 import org.apache.drill.exec.compile.sig.SignatureHolder;
 import org.apache.drill.exec.exception.SchemaChangeException;
 import org.apache.drill.exec.expr.fn.WorkspaceReference;
+import org.apache.drill.exec.expr.holders.ValueHolder;
 import org.apache.drill.exec.record.TypedFieldId;
+import org.apache.drill.shaded.guava.com.google.common.base.Function;
 import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
 import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
 import org.apache.drill.shaded.guava.com.google.common.collect.Maps;
@@ -65,6 +70,7 @@ public class ClassGenerator<T>{
 
   public static final GeneratorMapping DEFAULT_SCALAR_MAP = GM("doSetup", "doEval", null, null);
   public static final GeneratorMapping DEFAULT_CONSTANT_MAP = GM("doSetup", "doSetup", null, null);
+  public static final String INNER_CLASS_FIELD_NAME = "innerClassField";
 
   static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ClassGenerator.class);
 
@@ -76,6 +82,7 @@ public class ClassGenerator<T>{
   private final Map<String, ClassGenerator<T>> innerClasses = Maps.newHashMap();
   private final List<TypedFieldId> workspaceTypes = Lists.newArrayList();
   private final Map<WorkspaceReference, JVar> workspaceVectors = Maps.newHashMap();
+  private final Map<Pair<Integer, JVar>, Function<DrillBuf, ? extends ValueHolder>> constantVars;
   private final CodeGenerator<T> codeGenerator;
 
   public final JDefinedClass clazz;
@@ -86,6 +93,8 @@ public class ClassGenerator<T>{
   private ClassGenerator<T> innerClassGenerator;
   private LinkedList<SizedJBlock>[] blocks;
   private LinkedList<SizedJBlock>[] oldBlocks;
+
+  private JVar innerClassField;
 
   /**
    * Assumed that field has 3 indexes within the constant pull: index of the CONSTANT_Fieldref_info +
@@ -135,6 +144,7 @@ public class ClassGenerator<T>{
     this.evaluationVisitor = eval;
     this.model = model;
     this.optionManager = optionManager;
+    constantVars = new HashMap<>();
 
     blocks = (LinkedList<SizedJBlock>[]) new LinkedList[sig.size()];
     for (int i =0; i < sig.size(); i++) {
@@ -370,6 +380,7 @@ public class ClassGenerator<T>{
         innerClassGenerator.maxIndex += index;
         // blocks from the inner class should be used
         setupInnerClassBlocks();
+        innerClassField = clazz.field(JMod.NONE, model.ref(innerClassGenerator.clazz.name()), INNER_CLASS_FIELD_NAME);
         return true;
       }
       return innerClassGenerator.createNestedClass();
@@ -425,10 +436,8 @@ public class ClassGenerator<T>{
    * Creates methods from the signature {@code sig} with body from the appropriate {@code blocks}.
    */
   void flushCode() {
-    JVar innerClassField = null;
     if (innerClassGenerator != null) {
       blocks = oldBlocks;
-      innerClassField = clazz.field(JMod.NONE, model.ref(innerClassGenerator.clazz.name()), "innerClassField");
       innerClassGenerator.flushCode();
     }
     int i = 0;
@@ -531,9 +540,46 @@ public class ClassGenerator<T>{
 
   public JVar declareClassField(String prefix, JType t, JExpression init) {
     if (innerClassGenerator != null && hasMaxIndexValue()) {
-      return innerClassGenerator.clazz.field(JMod.NONE, t, prefix + index++, init);
+      return innerClassGenerator.declareClassField(prefix, t, init);
     }
     return clazz.field(JMod.NONE, t, prefix + index++, init);
+  }
+
+  public Pair<Integer, JVar> declareClassConstField(String prefix, JType t,
+                                                    Function<DrillBuf, ? extends ValueHolder> function) {
+    return declareClassConstField(prefix, t, null, function);
+  }
+
+  /**
+   * declare a constant field for the class.
+   * argument {@code function} holds the constant value which
+   * returns a value holder must be set to the class field when the class instance created.
+   * the class field innerClassField will be created if innerClassGenerator exists.
+   *
+   * @param prefix the prefix name of class field
+   * @param t the type of class field
+   * @param init init expression
+   * @param function the function holds the constant value
+   * @return the depth of nested class, class field
+   */
+  public Pair<Integer, JVar> declareClassConstField(String prefix, JType t, JExpression init,
+                                                    Function<DrillBuf, ? extends ValueHolder> function) {
+    JVar var;
+    int depth = 1;
+    if (innerClassGenerator != null) {
+      Pair<Integer, JVar> nested = innerClassGenerator.declareClassConstField(prefix, t, init, function);
+      depth = nested.getKey() + 1;
+      var = nested.getValue();
+    } else {
+      var = clazz.field(JMod.NONE, t, prefix + index++, init);
+    }
+    Pair<Integer, JVar> depthVar = Pair.of(depth, var);
+    constantVars.put(depthVar, function);
+    return depthVar;
+  }
+
+  public Map<Pair<Integer, JVar>, Function<DrillBuf, ? extends ValueHolder>> getConstantVars() {
+    return constantVars;
   }
 
   public HoldingContainer declare(MajorType t) {
@@ -646,7 +692,7 @@ public class ClassGenerator<T>{
         Class<?> p = params[i];
         childNew.arg(shim.param(model._ref(p), "arg" + i));
       }
-      shim.body()._return(childNew);
+      shim.body()._return(JExpr._this().invoke("injectMembers").arg(childNew));
     }
   }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
@@ -38,6 +38,7 @@ import org.apache.drill.exec.compile.sig.RuntimeOverridden;
 import org.apache.drill.exec.exception.ClassTransformationException;
 import org.apache.drill.exec.exception.OutOfMemoryException;
 import org.apache.drill.exec.exception.SchemaChangeException;
+import org.apache.drill.exec.expr.ClassGenerator;
 import org.apache.drill.exec.expr.TypeHelper;
 
 import org.apache.drill.exec.memory.BaseAllocator;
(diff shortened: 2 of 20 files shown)


vvysotskyi:
@lushuifeng, thanks for the contribution!

Besides setters added to the generated code, are there any changes which optimize the generated code, are these setters used somewhere?

Please note, that Drill has its own framework for scalar replacement and passing value holders to the methods as arguments make a scalar replacement for those holders impossible.

Also, Drill can generate too large classes. To avoid problems with the class constant pool overflow was implemented splitting the classes into the smaller ones. Is this change incorporates or at least does not break it?

vvysotskyi:
@lushuifeng, thanks, looks great!

But could you please also provide results for JDK and Janino compilers with and without this feature.
And it would be good to know the results for other literals, except strings and more complex queries.

Are there any cases, where this feature breaks something, or cause performance degradation? If no, then no need to provide the option for disabling this functionality.

Regarding unit tests, it may be written considering compiler limitation for method size - somehow may be passed huge string literals, so query should fail without these changes. This test should be sufficient.

vvysotskyi:
If this fix just improves the performance without any consequences, then there is no need to disable it or specify option for that.
But in either case, it is useful to see numbers.

Regarding the unit test, generate a query with large string literals, so they will be specified in the methods in generated code, and query should fail with "code too large" compile exception.

lushuifeng:
@vvysotskyi unit tests are added according to your advice.
What should I do due to the error in CI of `drill-jdbc-all-1.15.0-SNAPSHOT.jar size (39000409) too large. Max. is 39000000` , the max size is changed temporarily, is it appropriate to change in the PR?
The option will be kept for a while to provide more sql tests

arina-ielchiieva:
@lushuifeng regarding jdbc-all I guess we need to exclude unneeded libs from it to reduce the size or bump the size if there is nothing to exclude. I believe this effects all new PRs so this should be done in separate Jira prior to new PRs.

vvysotskyi:
@lushuifeng, `jdbc-all` jar size will be reduced in the scope of https://github.com/apache/drill/pull/1168.
@sachouche, what is the status of that PR?

sachouche:
This PR is ready; it only needed testing from SQL clients to ensure that all runtime dependencies are satisfied.

sachouche:
PR #1168 is ready; the remaining part is some advanced testing as requested by @parthchandra (using SQL clients) to ensure that all runtime dependencies are there. It seems the Drill internal unit tests do not provide such guarantees. I didn't have the cycles to perform such testing.

lushuifeng:
@vvysotskyi thanks for your suggestions.
I didn't know that Travis only run some of the test.
I'm sorry I've not run all the tests yet, I would like to run them as soon as possible

lushuifeng:
@vvysotskyi thanks for your comments. Changes are made as you suggested.
As you mentioned early, some tests are made to test this functionality:
Both jdk and janio compiler works fine.
Constants with different types(string, double, long, decimal, time, timestamp and so on) are tested.
But some tests beside the Travis are not passed, a lot of `No default schema selected` errors occured, it seems that these errors are not connected to this change.

lushuifeng:
@vvysotskyi Sorry for my misunderstanding.
yes, simply ran with `mvn -T 8 clean install`, the tests cann't be passed.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
```diff
@@ -536,6 +543,19 @@ public JVar declareClassField(String prefix, JType t, JExpression init) {
     return clazz.field(JMod.NONE, t, prefix + index++, init);
   }
 
+  public JMethod declareSetterMethod(JVar var, Function<DrillBuf, ? extends ValueHolder> function) {
```
vvysotskyi: Please add Javadoc here also.

lushuifeng:
the tests without `-T 8` still failed.
Could you please run one of my failed tests: `org.apache.drill.exec.impersonation.TestImpersonationMetadata#testShowFilesInWSWithNoPermissionsForQueryUser`

exec/jdbc-all/pom.xml
```diff
@@ -511,7 +511,7 @@
                   This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                   </message>
-                  <maxsize>39000000</maxsize>
+                  <maxsize>39100000</maxsize>
```
vvysotskyi: Please rebase onto the master and revert this change, since there was PR, were excluded some classes from the JAR.

lushuifeng:
@arina-ielchiieva thanks for your advice.
The tests are not passed with `mvn clean install` on master, got the same errors on branch DRILL-6763.
Single Test `TestImpersonationMetadata#testShowFilesInWSWithNoPermissionsForQueryUser` failed as on the branch DRILL-6763.

vvysotskyi:
This fix causes the next unit tests failures:
```
Failed tests: 
  TestLargeFileCompilation.testJDKHugeStringConstantCompilation:257->BaseTestQuery.resetSessionOption:375 Failed to reset session option `exec.java_compiler`, Error: org.apache.drill.common.exceptions.UserException: SYSTEM ERROR: IllegalArgumentException: Attempted to send a message when connection is no longer valid.

Query submission to Drillbit failed.

[Error Id: 6a397072-825d-4550-a39e-ea4f6eafdc6d ]

Tests in error: 
  TestLargeFileCompilation.testProject:191->BaseTestQuery.testNoResult:384 » Rpc
  TestLargeFileCompilation.testTOP_N_SORT:177->BaseTestQuery.testNoResult:358->BaseTestQuery.testNoResult:384 » Rpc
  TestLargeFileCompilation.testEXTERNAL_SORT:171->BaseTestQuery.testNoResult:358->BaseTestQuery.testNoResult:384 » Rpc
  TestLargeFileCompilation.testPARQUET_WRITER:156->BaseTestQuery.testNoResult:358->BaseTestQuery.testNoResult:384 » Rpc
  TestLargeFileCompilation.testHashJoin:206->BaseTestQuery.testNoResult:358->BaseTestQuery.testNoResult:384 » Rpc
  TestHashJoin.testHashJoinExprInCondition:252 » Rpc org.apache.drill.common.exc...

Tests run: 3545, Failures: 1, Errors: 6, Skipped: 156
```

vvysotskyi:
@lushuifeng, since these tests failures are absent on the latest master, they caused by your changes. Could you please fix them? Also, some of the failures in `TestLargeFileCompilation` may be caused by the previously failed unit tests from this class.

lushuifeng:
@vvysotskyi Could you please take a look at these changes?
> commit 8b44957 fix the compilation error if there are nested splitting classes
> commit 5edecc4 and 413891a fix the compilation error of the innerClasses like BatchHolder and OutgoingRecordBatch contain constant values

I've just found that previous errors in my tests `mvn clean install` are because of there is [a DOT in my user name](https://issues.apache.org/jira/browse/HADOOP-7050)

vvysotskyi:
@lushuifeng, I have a question regarding nested splitting classes: these setters will be added only to the top level class, or they will be separated between both: top and nested classes?

lushuifeng:
Both top and nested classes will be added since the new class member is in inner class if there is a inner class, the inner class member is initialized by invoking top class member function:

top:
`public void setConstant5(IntHolder intHolder) {
  this.innerClassField.setConstant5(intHolder)
}`
inner:
`
public void setConstant5(IntHolder intHolder) {
  this.constant5  = intHolder;
}`

lushuifeng:
I agree with you, it will cause the overflow of the class constant pool in extreme case, that is to say a projection or a filter contains thousands (or maybe larger) of constants

The good news is all tests in `TestLargeFileCompilation ` are succeeded with this change, shall we add an option to disable this functionality or maybe there is a good way to avoid?

I'd do some tests to find out how much constants will cause the overflow of the class constant pool.

vvysotskyi:
I don't think that this is a good idea to add an option. In the case when we have a combination of these two problems, option won't help.

I think it is possible to enhance the way for collecting setter methods for example to map full [nested] class names and its setters, so after that receive nested class instances and call setters for them in `CodeGenMemberInjector.injectMembers()`.

lushuifeng:
In order to collect setter methods of nested class, some changes will be made:
1. `public void setConstant4(IntHolder constant4)` will be changed to `public void setConstant4(IntHolder constant4, String name)`, the second param `name` is to store the name of the first param, here is the string "constant4", unfortunately the second param is useless unless there is a nested class.
2.  methods will be collected during `__DRILL_INIT__`,  Lambda is not supported in JCodeModel 2.6, it seems that anonymousClass in JCodeModel 2.6 can't be narrowed, the <ValueHolder, String> is missing. some casts have to be added
>     Map<String, BiConsumer<ValueHolder, String>> nestedClassFunctions = new HashMap();
>     this.nestedClassFunctions.put("constant355", new BiConsumer<ValueHolder, String>() {
>          @Override
>           public void accept(ValueHolder constant355, String name) {
>              (innerClassField).setConstant355(((IntHolder) constant355), "constant355");
>           }
>      });
3. InnerClassField is referenced in anonymousClass, keyword FINAL should be added to its mods, so it should be initialized in constructor not in `__DRILL_INIT__`

What is your suggestions? @vvysotskyi  thanks.

vvysotskyi:
@lushuifeng, I meant to replace 
`Map<JMethod, Function<DrillBuf, ? extends ValueHolder>> setters`
with something like this: 
`Map<JMethod, Pair<String, Function<DrillBuf, ? extends ValueHolder>>> setters`,
where the `String` is the name of nested class field, or the name, which corresponds to the full path to this clas, for expample like this: `nestedClassField.nestedNestedClassField.nestedNestedNestedClassField`. So in the `CodeGenMemberInjector.injectMembers()`, required nested class instance may be received using its name and setter method may be called on it.

lushuifeng:
I think it is the way I'm doing besides you wrap two params (DrillBuf and String) into a Pair for the first point. The method in  `CodeGenMemberInjector.injectMembers()` will get two params instead of old one.
Have you tried about anonymousClass in JCodeModel 2.6, it is a little confused to me that the narrow function not work correctly.

lushuifeng:
@vvysotskyi All tests have been passed, could you please take a look? thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
```diff
@@ -531,11 +541,42 @@ public JVar declareClassField(String prefix, JType t) {
...
+   * argument {@code function} holds the constant value which
+   * returns a value holder must be invoked when the class instance created.
+   * the setter method of innerClassField will be invoked if innerClassGenerator exists.
+   *
+   * @param var the class member variable
+   * @param function the function holds the constant value
+   * @return the depth of nested class, setter method
+   */
+  public Pair<Integer, JMethod> declareSetterMethod(JVar var, Function<DrillBuf, ? extends ValueHolder> function) {
+    JMethod setter = clazz.method(JMod.PUBLIC, void.class, "set" + StringUtils.capitalize(var.name()));
```
vvysotskyi: Please move the initialization to the else block below and put code which uses this var after the `if...else` block.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
```diff
@@ -646,7 +687,7 @@ public void preparePlainJava() {
         Class<?> p = params[i];
         childNew.arg(shim.param(model._ref(p), "arg" + i));
       }
-      shim.body()._return(childNew);
+      shim.body()._return(JExpr._this().invoke("injectMembers").arg(childNew));
```
vvysotskyi: Could you please explain what is the reason for this change?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
```diff
@@ -1074,7 +1078,12 @@ private void addBatchHolder(int part, int batchRowCount) {
 
   // These methods are overridden in the generated class when created as plain Java code.
   protected BatchHolder newBatchHolder(int batchRowCount) {
-    return new BatchHolder(batchRowCount);
+    return this.injectMembers(new BatchHolder(batchRowCount));
+  }
+
+  protected BatchHolder injectMembers(BatchHolder batchHolder) {
+    CodeGenMemberInjector.injectMembers(cg, batchHolder, context);
```
vvysotskyi: Is it possible to move this method call to the place, general for other similar classes?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/CodeGenMemberInjector.java
```diff
@@ -0,0 +1,79 @@
...
+   * Generated code for a class may have several class members, they
+   * are initialized by invoking this method when the instance created.
+   *
+   * @param cg       the class generator
+   * @param instance the class instance created by the compiler
+   * @param context  the fragment context
+   */
+  public static void injectMembers(ClassGenerator<?> cg, Object instance, FragmentContext context) {
+    Map<Integer, Object> cachedInstances = new HashMap<>();
+    for (Map.Entry<Pair<Integer, JMethod>, Function<DrillBuf, ? extends ValueHolder>> setter : cg.getSetters().entrySet()) {
```
vvysotskyi: We may change the type of collection which used for storing setter methods. Can be used Guava's Table, so we will be able to omit usage of `cachedInstances` map.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
```diff
@@ -87,6 +95,8 @@
   private LinkedList<SizedJBlock>[] blocks;
   private LinkedList<SizedJBlock>[] oldBlocks;
 
+  private JVar innerClassField;
+
   /**
    * Assumed that field has 3 indexes within the constant pull: index of the CONSTANT_Fieldref_info +
```
vvysotskyi: Please note that with introducing new methods in the class, this formula should be revised to take them into account.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
lushuifeng: Is It better by NOT declaring the setter function, these fields can be accessed by reflection directly, so the number of constants in constant pool can be reduced

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
lushuifeng: The innerClasses like BatchHolder may have the constants, BatchHolder instance is created via newBatchHolder dynamically when processing recordBatch, that is the difference between innerClasses and nested splitting classes.

The innerClasses member fields is initialized by invoking injectMembers(innerClassInstance) which is override generated by this code as shown in 
`protected BatchHolder newBatchHolder(int batchRowCount) {
    return this.injectMembers(new BatchHolder(batchRowCount));
  }`

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggTemplate.java
lushuifeng: The ClassGenerator and FragmentContext is needed in this method for general cases, but not for test classes, see example in `ExampleTemplateWithInner.java`

newBatchHolder is override by generated code can be implemented as follows to remove above method 
`protected BatchHolder newBatchHolder(int batchRowCount) { return CodeGenMemberInjector.injectMembers(cg, new BatchHolder(batchRowCount), context); }`
but the `cg` and `context` params are hard coded, not flexable. I think it is better to make a function call to hide the implementation which may differ in each template.

lushuifeng:
@vvysotskyi setter method have been removed, the formula of `maxIndex` should be same as before.

[DRILL-6722](https://issues.apache.org/jira/projects/DRILL/issues/DRILL-6722?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC) seems to be resolved with this change by accident, the root cause is not found yet.

Test `TestLargeFileCompilation#testProject` has 10% improvement in the running time compare to master without cached classes.

lushuifeng:
@vvysotskyi Could you please take a look, thanks.

lushuifeng:
@vvysotskyi Could you please take a look, thanks.

lushuifeng:
@vvysotskyi Could you please take a look, thanks.

lushuifeng:
@vvysotskyi Could you please take a look, thanks.

lushuifeng:
@vvysotskyi Could you please take a look, thanks.

lushuifeng:
done

lushuifeng:
Thank you for your review and comments.

PR #1482, opened 2018-09-27
DRILL-6762: Dynamic UDFs registered on 1 Drillbit not visible on others

kkhatua:
The issue is with the registry version value of -1 being, both a valid version number and indicative of the (remote) registry being unreachable. DRILL-6053 exacerbates this ambiguity. The workaround is to provide a different value for the registry version when it is unreachable. In this case, we picked Integer.MIN_VALUE which is << -1

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
@@ -408,7 +408,7 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
    * @return true is local registry should be refreshed, false otherwise
    */
   private boolean isRegistrySyncNeeded(long remoteVersion, long localVersion) {
-    return remoteVersion != -1 && remoteVersion != localVersion;
+    return remoteVersion != RemoteFunctionRegistry.UNREACHABLE && remoteVersion != localVersion;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/RemoteFunctionRegistry.java
```diff
@@ -84,6 +84,7 @@
   private static final String registry_path = "registry";
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(RemoteFunctionRegistry.class);
   private static final ObjectMapper mapper = new ObjectMapper().enable(INDENT_OUTPUT);
+  public static final int UNREACHABLE = Integer.MIN_VALUE;
```
Muritiku: Maybe better to use long but not int?
Because in that case will be autoconversion in private "boolean isRegistrySyncNeeded(long remoteVersion, long localVersion)"

Just a thought. It is not reccomended in java to compare int and long directly.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
```diff
@@ -408,7 +408,7 @@ private boolean isRegistrySyncNeeded() {
    * @return true is local registry should be refreshed, false otherwise
    */
   private boolean isRegistrySyncNeeded(long remoteVersion, long localVersion) {
-    return remoteVersion != -1 && remoteVersion != localVersion;
+    return remoteVersion != RemoteFunctionRegistry.UNREACHABLE && remoteVersion != localVersion;
```
Muritiku: Look at my another comment. Here is the int and long conversion.

arina-ielchiieva:
Closing the PR, since https://github.com/apache/drill/pull/1484 was opened which contains full fix of the problem.

PR #1483, opened 2018-09-28
DRILL-3988: Expose Drill built-in functions & UDFs  in a system table

kkhatua:
This commit exposes available SQL functions in Drill and also detects UDFs that have been dynamically loaded into Drill. 
An example is shown below for 2 UDFs dynamically loaded into the cluster, along side the existing built-in functions that come with Drill.

```
0: jdbc:drill:schema=sys> select source, count(*) as functionCount from sys.functions group by source;
+-----------------------------------------+----------------+
|                 source                  | functionCount  |
+-----------------------------------------+----------------+
| built-in                                | 2704           |
| simple-drill-function-1.0-SNAPSHOT.jar  | 12             |
| drill-url-tools-1.0.jar                 | 1              |
+-----------------------------------------+----------------+
3 rows selected (0.209 seconds)
```

The system table exposes information as shown. The UDF is initialized, making the `returnType` available.
The `random(FLOAT8-REQUIRED,FLOAT8-REQUIRED)` function is an example of a UDF that has overloaded arguments (see `signature`).
The `url_parse(VARCHAR-REQUIRED)` function is another example of an initialized UDF.
Rest are built-in functions that meet the query's filter criteria.
```
0: jdbc:drill:schema=sys> select * from sys.functions where name like 'random' or name like '%url%';
+-------------+----------------------------------+-------------+-----------------------------------------+
|    name     |            signature             | returnType  |                 source                  |
+-------------+----------------------------------+-------------+-----------------------------------------+
| parse_url   | VARCHAR-REQUIRED                 | LATE        | built-in                                |
| random      |                                  | FLOAT8      | built-in                                |
| random      | FLOAT8-REQUIRED,FLOAT8-REQUIRED  | FLOAT8      | simple-drill-function-1.0-SNAPSHOT.jar  |
| url_decode  | VARCHAR-REQUIRED                 | VARCHAR     | built-in                                |
| url_encode  | VARCHAR-REQUIRED                 | VARCHAR     | built-in                                |
| url_parse   | VARCHAR-REQUIRED                 | LATE        | drill-url-tools-1.0.jar                 |
+-------------+----------------------------------+-------------+-----------------------------------------+
6 rows selected (0.619 seconds)
```

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
@@ -293,10 +293,6 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
     return false;
   }
 
-  public LocalFunctionRegistry getLocalFunctionRegistry() {
-    return localFunctionRegistry;
-  }
-
   public RemoteFunctionRegistry getRemoteFunctionRegistry() {
     return remoteFunctionRegistry;
   }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
@@ -491,13 +491,13 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
 
   /**
    * Retrieve all functions, mapped by source jars (after syncing)
-   * @return Map of source jars and their functions
+   * @return Map of source jars and their functionHolders
    */
-  public Map<String, List<FunctionHolder>> getAllFunctionsHoldersByJar() {
+  public Map<String, List<FunctionHolder>> getAllJarsWithFunctionsHolders() {
     if (useDynamicUdfs) {
       syncWithRemoteRegistry(localFunctionRegistry.getVersion());
     }
-    return localFunctionRegistry.getAllFunctionsHoldersByJar();
+    return localFunctionRegistry.getAllJarsWithFunctionsHolders();
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
@@ -28,6 +28,7 @@ import java.net.URLConnection;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -54,6 +55,7 @@ import org.apache.drill.exec.coord.store.TransientStoreListener;
 import org.apache.drill.exec.exception.FunctionValidationException;
 import org.apache.drill.exec.exception.JarValidationException;
 import org.apache.drill.exec.expr.fn.registry.LocalFunctionRegistry;
+import org.apache.drill.exec.expr.fn.registry.FunctionHolder;
 import org.apache.drill.exec.expr.fn.registry.JarScan;
 import org.apache.drill.exec.expr.fn.registry.RemoteFunctionRegistry;
 import org.apache.drill.exec.planner.sql.DrillOperatorTable;
@@ -291,6 +293,10 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
     return false;
   }
 
+  public LocalFunctionRegistry getLocalFunctionRegistry() {
+    return localFunctionRegistry;
+  }
+
   public RemoteFunctionRegistry getRemoteFunctionRegistry() {
     return remoteFunctionRegistry;
   }
@@ -481,6 +487,17 @@ public class FunctionImplementationRegistry implements FunctionLookupContext, Au
       }
     }
     return missingJars;
+  }
+
+  /**
+   * Retrieve all functions, mapped by source jars (after syncing)
+   * @return Map of source jars and their functions
+   */
+  public Map<String, List<FunctionHolder>> getAllFunctionsHoldersByJar() {
+    if (useDynamicUdfs) {
+      syncWithRemoteRegistry(localFunctionRegistry.getVersion());
+    }
+    return localFunctionRegistry.getAllFunctionsHoldersByJar();
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
@@ -25,6 +25,7 @@ import org.apache.drill.common.concurrent.AutoCloseableLock;
 import org.apache.drill.exec.expr.fn.DrillFuncHolder;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
@@ -97,9 +98,13 @@ public class FunctionRegistryHolder {
   // function name, Map<function signature, function holder>
   private final Map<String, Map<String, DrillFuncHolder>> functions;
 
+  // jar name, List<function holder>
+  private final Map<String, List<FunctionHolder>> jarFunctions;
+
   public FunctionRegistryHolder() {
     this.functions = new ConcurrentHashMap<>();
     this.jars = new ConcurrentHashMap<>();
+    this.jarFunctions = new HashMap<String, List<FunctionHolder>>();
   }
 
   /**
@@ -131,6 +136,7 @@ public class FunctionRegistryHolder {
         Map<String, Queue<String>> jar = new ConcurrentHashMap<>();
         jars.put(jarName, jar);
         addFunctions(jar, newJar.getValue());
+        jarFunctions.put(jarName, newJar.getValue());
       }
       this.version = version;
     }
@@ -158,6 +164,17 @@ public class FunctionRegistryHolder {
   public List<String> getAllJarNames() {
     try (@SuppressWarnings("unused") Closeable lock = readLock.open()) {
       return new ArrayList<>(jars.keySet());
+    }
+  }
+
+  /**
+   * Retrieves all functions associated with all the jars
+   * This is read operation, so several users can perform this operation at the same time.
+   * @return list of all functions, mapped by their sources
+   */
+  public Map<String, List<FunctionHolder>> getAllFunctionHoldersByJar() {
+    try (@SuppressWarnings("unused") Closeable lock = readLock.open()) {
+      return jarFunctions;
     }
   }
 
@@ -365,5 +382,7 @@ public class FunctionRegistryHolder {
         functions.remove(function);
       }
     }
+    jarFunctions.remove(jarName);
+
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
@@ -26,6 +26,7 @@ import org.apache.drill.exec.expr.fn.DrillFuncHolder;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
@@ -98,13 +99,9 @@ public class FunctionRegistryHolder {
   // function name, Map<function signature, function holder>
   private final Map<String, Map<String, DrillFuncHolder>> functions;
 
-  // jar name, List<function holder>
-  private final Map<String, List<FunctionHolder>> jarFunctions;
-
   public FunctionRegistryHolder() {
     this.functions = new ConcurrentHashMap<>();
     this.jars = new ConcurrentHashMap<>();
-    this.jarFunctions = new HashMap<String, List<FunctionHolder>>();
   }
 
   /**
@@ -136,7 +133,6 @@ public class FunctionRegistryHolder {
         Map<String, Queue<String>> jar = new ConcurrentHashMap<>();
         jars.put(jarName, jar);
         addFunctions(jar, newJar.getValue());
-        jarFunctions.put(jarName, newJar.getValue());
       }
       this.version = version;
     }
@@ -168,14 +164,35 @@ public class FunctionRegistryHolder {
   }
 
   /**
-   * Retrieves all functions associated with all the jars
+   * Retrieves all functions (holders) associated with all the jars
    * This is read operation, so several users can perform this operation at the same time.
    * @return list of all functions, mapped by their sources
    */
-  public Map<String, List<FunctionHolder>> getAllFunctionHoldersByJar() {
+  public Map<String, List<FunctionHolder>> getAllJarsWithFunctionHolders() {
+    Map<String, List<FunctionHolder>> allFunctionHoldersByJar = new HashMap<>();
+
     try (@SuppressWarnings("unused") Closeable lock = readLock.open()) {
-      return jarFunctions;
+      for (String jarName : jars.keySet()) {
+        //Capture functionHolders here
+        List<FunctionHolder> drillFuncHolderList = new LinkedList<>();
+
+        Map<String, Queue<String>> functionsInJar = jars.get(jarName);
+        for (Map.Entry<String, Queue<String>> functionEntry : functionsInJar.entrySet()) {
+          String fnName = functionEntry.getKey();
+          Queue<String> fnSignatureList = functionEntry.getValue();
+          //Get all FunctionHolders (irrespective of source)
+          Map<String, DrillFuncHolder> functionHolders = functions.get(fnName);
+          //Iterate for matching entries and populate new Map
+          for (Map.Entry<String, DrillFuncHolder> entry : functionHolders.entrySet()) {
+            if (fnSignatureList.contains(entry.getKey())) {
+              drillFuncHolderList.add(new FunctionHolder(fnName, entry.getKey(), entry.getValue()));
+            }
+          }
+        }
+        allFunctionHoldersByJar.put(jarName, drillFuncHolderList);
+      }
     }
+    return allFunctionHoldersByJar;
   }
 
   /**
@@ -382,7 +399,5 @@ public class FunctionRegistryHolder {
         functions.remove(function);
       }
     }
-    jarFunctions.remove(jarName);
-
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
@@ -18,7 +18,6 @@
 package org.apache.drill.exec.store.sys;
 
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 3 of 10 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,167 @@
...
+        for (String name : registeredNames) {
+          FunctionInfo fnInfo = new FunctionInfo(name, signature, returnType, LocalFunctionRegistry.BUILT_IN);
+          functionMap.put(generateKey(name, signature), fnInfo);
+        }
+      }
+    }
+  }
+
+  //Loads remote-registry functions into functionMap (updates if
+  private void listRemoteFunctionRegistry(Map<String, FunctionInfo> functionMap) {
```
arina-ielchiieva: @kkhatua 
To avoid this unpleasant parsing I would suggest you add method to `FunctionRegistryHolder` which would return `ListMultimap<String, DrillFuncHolder>`, where key would be jar_name, value would be list of function holders. Let's call it `getAllFunctionsByJars` (name can be changed).   Then in `LocalFunctionRegistry` add delegating method:
```
 public ListMultimap<String, DrillFuncHolder> getAllFunctionsByJars() {
   return registryHolder.getAllFunctionsByJars();
 }
```
Then in `FunctionImplementationRegistry` add the following method:
```
public ListMultimap<String, DrillFuncHolder> getAllFunctionsByJars() {
   if (useDynamicUdfs) {
     syncWithRemoteRegistry(localFunctionRegistry.getVersion());
   }
   return localFunctionRegistry.getAllFunctionsByJars();
 }
```
The above you would get you full list of built-in and remote functions
Unit test can be added to `FunctionRegistryHolderTest` for the new method.

kkhatua: Thanks, @arina-ielchiieva . Modified the PR as per your suggestion.

kkhatua:
Refactored as per suggestions. Added tests.

kfaraaz:
Can you please add a unit test to drop a function using, DROP FUNCTION USING JAR '<jar_name>.jar'; 
and then query the sys.functions table and verify that the dropped function is not in the results returned, after it was dropped.

kkhatua:
@kfaraaz  these re unit tests, so we're not using the SQL interface to add or remove UDF jars. It's primarily testing the APIs. Existing functional tests should be covering that.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
```diff
@@ -291,6 +293,10 @@ public boolean isFunctionComplexOutput(String name) {
     return false;
   }
 
+  public LocalFunctionRegistry getLocalFunctionRegistry() {
```
arina-ielchiieva: Why you need access to local function registry here?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -97,9 +98,13 @@
   // function name, Map<function signature, function holder>
   private final Map<String, Map<String, DrillFuncHolder>> functions;
 
+  // jar name, List<function holder>
+  private final Map<String, List<FunctionHolder>> jarFunctions;
```
arina-ielchiieva: I don't see any value of adding one more holder to keep functions. Current two holders have enough information and its unique. Your holder just duplicates it. Please consider approach i have suggested before.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,116 @@
...
+  private Iterator<FunctionInfo> sortedIterator;
+
+  public FunctionsIterator(ExecutorFragmentContext context) {
+    Map<String, FunctionInfo> functionMap = new HashMap<String, FunctionInfo>();
+    //Access Registry for function list
+    FunctionImplementationRegistry funcImplRegistry = (FunctionImplementationRegistry) context.getFunctionRegistry();
+    Map<String, List<FunctionHolder>>  jarFunctionListMap = funcImplRegistry.getAllFunctionsHoldersByJar();
+
+    //Step 1: Load Built-In
+    for (FunctionHolder builtInEntry : jarFunctionListMap.get(LocalFunctionRegistry.BUILT_IN)) {
```
arina-ielchiieva: Why you have to populate built-in and jar functions separately?
`Map<String, List<FunctionHolder>>  jarFunctionListMap` contains `LocalFunctionRegistry.BUILT_IN` for built-in functions.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,116 @@
...
+    for (String jarName : new ArrayList<>(jarFunctionListMap.keySet())) {
+      if (!jarName.equals(LocalFunctionRegistry.BUILT_IN)) {
+        for (FunctionHolder udfEntry : jarFunctionListMap.get(jarName)) {
+          populateFunctionMap(functionMap, jarName, udfEntry.getHolder());
+        }
+      }
+    }
+
+    List<FunctionInfo> functionList = new ArrayList<FunctionsIterator.FunctionInfo>(functionMap.values());
+    functionList.sort(new Comparator<FunctionInfo>() {
```
arina-ielchiieva: You can make comparator class level variable.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  //Populate for give Jar-FunctionHolder Entry
+  private void populateFunctionMap(Map<String, FunctionInfo> functionMap, String jarName, DrillFuncHolder dfh) {
+    String registeredNames[] = dfh.getRegisteredNames();
+    String signature = dfh.getInputParameters();
+    String returnType = dfh.getReturnType().getMinorType().toString();
+    for (String name : registeredNames) {
+      //Generate a uniqueKey to allow for fnName#fnSignature
+      String funcSignatureKey = new StringBuilder(64).append(name).append('#').append(signature).toString();
```
arina-ielchiieva: Please explain the purpose of this unique key.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,116 @@
...
+          return o1.returnType.compareTo(o2.returnType);
+        }
+        return result;
+      }
+    });
+
+    sortedIterator = functionList.iterator();
+  }
+
+  //Populate for give Jar-FunctionHolder Entry
```
arina-ielchiieva: Use java-doc.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,116 @@
...
+  public boolean hasNext() {
+    return sortedIterator.hasNext();
+  }
+
+  @Override
+  public FunctionInfo next() {
+    return sortedIterator.next();
+  }
+
+  public static class FunctionInfo {
```
arina-ielchiieva: java-doc

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionImplementationRegistry.java
kkhatua: For DRILL-6084 , the `WebServer`  needs access to the registry to load the list of functions for rendering in JavaScript.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
kkhatua: The `functions` holder groups functionHolders from different sources as long as they share the same name. As a result, for a given function name, the source of a specific signature cannot be inferred. Since the holders non-primitive datatypes, they are not copied and only references maintained. So their additional footprint should be fairly small within the Drillbit.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: That is because the UDFs are initialized and show up with the `source` as `BUILT_IN`. 
Iterating over holders from the non-`built-in` sources, I update the `functionMap` with the true sources for those UDFs. I noticed this when a function name was overloaded with a different signature.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: Agreed. But the variable isn't used anywhere else in the `FunctionsIterator` class, hence I wrote it in-place.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: The original loading of functions from `built-in` captures functions from all the sources. When iterating over the non-`built-in` sources, the key helps replace `FunctionInfo` instances that are wrongly shown as sourced from `built-in`

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: Ok

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: Ok

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: Just realized.. that's a private method. Doesn't need a java-doc

kfaraaz:
@kkhatua I checked there are no existing functional tests that do a DROP FUNCTION USING ...
We need to add functional tests that drop functions which are defined in jar files and recreate and verify.

/drillTEST_FRAMEWORK/drill-test-framework/framework/resources/Functional
[test@dfc3012-45 Functional]# grep -ir "DROP FUNCTION USING JAR" *
[test@dfc3012-45 Functional]#

kkhatua:
@arina-ielchiieva updated the PR with an additional commit.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: Put it in the comments

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,110 @@
...
+          return o1.returnType.compareTo(o2.returnType);
+        }
+        return result;
+      }
+    });
+
+    sortedIterator = functionList.iterator();
+  }
+
+  //Populate the map for a given Jar-DrillFunctionHolder
```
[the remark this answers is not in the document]
kkhatua: Done.

kkhatua:
Made the suggested changes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.drill.exec.ops.ExecutorFragmentContext;
+import org.apache.drill.exec.store.pojo.NonNullable;
+
+/**
+ * List functions as a System Table
+ */
+public class FunctionsIterator implements Iterator<Object> {
+  //private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FunctionsIterator.class);
+
+  private Iterator<FunctionInfo> sortedIterator;
```
arina-ielchiieva: final?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,111 @@
...
+
+  public FunctionsIterator(ExecutorFragmentContext context) {
+    Map<String, FunctionInfo> functionMap = new HashMap<String, FunctionInfo>();
+    //Access Registry for function list
+    FunctionImplementationRegistry funcImplRegistry = (FunctionImplementationRegistry) context.getFunctionRegistry();
```
arina-ielchiieva: It's not save to cast: either change return type in `getFunctionRegistry()` or add `instanceOf` check.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,111 @@
...
+    Map<String, List<FunctionHolder>> jarFunctionListMap = funcImplRegistry.getAllJarsWithFunctionsHolders();
+    for (String jarName : new ArrayList<>(jarFunctionListMap.keySet())) {
```
arina-ielchiieva: Why we need to wrap into ArrayList?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
arina-ielchiieva: Why `StringBuilder` capacity is 64?

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Representation of an entry in the System table - Functions
+   */
+  public static class FunctionInfo {
+    @NonNullable
+    public final String name;
+    @NonNullable
+    public final String signature;
+    public final String returnType;
```
arina-ielchiieva: Ypu have mentioned that issue with return type if fixed, should it be `@NonNullable`? the same for source.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -161,6 +163,38 @@ public void removeJar(String jarName) {
...
+   * This is read operation, so several users can perform this operation at the same time.
+   * @return list of all functions, mapped by their sources
+   */
+  public Map<String, List<FunctionHolder>> getAllJarsWithFunctionHolders() {
+    Map<String, List<FunctionHolder>> allFunctionHoldersByJar = new HashMap<>();
+
+    try (@SuppressWarnings("unused") Closeable lock = readLock.open()) {
+      for (String jarName : jars.keySet()) {
+        //Capture functionHolders here
+        List<FunctionHolder> drillFuncHolderList = new LinkedList<>();
```
arina-ielchiieva: Why LinkedList?

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolderTest.java
```diff
@@ -202,26 +250,38 @@ public void testGetAllFunctionsWithSignatures() {
...
   }
 
   @Test
   public void testGetHoldersByFunctionName() {
-    List<DrillFuncHolder> expectedResult = newJars.values().stream()
-        .flatMap(Collection::stream)
-        .filter(f -> "lower".equals(f.getName()))
-        .map(FunctionHolder::getHolder)
-        .collect(Collectors.toList());
+    List<DrillFuncHolder> expectedUniqueResult = Lists.newArrayList();
```
arina-ielchiieva: Please use `new ArrayList<>` here and below.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolderTest.java
```diff
@@ -155,6 +162,47 @@ public void testGetAllJarNames() {
     compareTwoLists(expectedResult, registryHolder.getAllJarNames());
   }
 
+  @Test
+  public void testGetAllJarsWithFunctionHolders() {
+    Map<String, List<FunctionHolder>> fnHoldersInRegistry = registryHolder.getAllJarsWithFunctionHolders();
+    //Iterate and confirm lists are same
+    for (String jarName : Lists.newArrayList(newJars.keySet())) {
```
arina-ielchiieva: No need to wrap into ArrayList here and a couple more times below in the code.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: My bad. I was switching between this (list) and a map when implementing, so I missed removing this during refactoring. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: +1

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/FunctionsIterator.java
kkhatua: Ok

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
kkhatua: Since the variable is only used to add and remove the function holders, this seemed most efficient for memory and speed. Some sources (like `built-in`) will have 2K+ functions, while UDFs will have as little as 1 function.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolderTest.java
kkhatua: Done

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolderTest.java
kkhatua: Done. removed all references to Lists and removed the Guava import as well.

exec/java-exec/src/test/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolderTest.java
```diff
@@ -155,6 +162,47 @@ public void testGetAllJarNames() {
...
+
+      compareTwoLists(expectedHolderList, testHolderList);
+    }
+
+    Map<String, String> shuffleFunctionMap = new HashMap<>();
+    // Confirm that same function spans multiple jars with different signatures
+    //Init: Expected Map of items
+    for (String jarName : newJars.keySet()) {
+      for (FunctionHolder funcHolder : newJars.get(jarName)) {
+        if (funcHolder.getName().equals(SHUFFLE_FUNC_NAME)) {
```
arina-ielchiieva: General recommendation is not use equals on constant to avoid NPE: `SHUFFLE_FUNC_NAME.equals(funcHolder.getName())`. Please change here and in the code below.

PR #1484, opened 2018-09-28
DRILL-6762: Fix dynamic UDFs versioning issue

arina-ielchiieva:
1. Added UndefinedVersionDelegatingStore to serve as versioned wrapper for those stores that do not support versioning.
2. Aligned remote and local function registries version type. Type will be represented as int since ZK version is returned as int.
3. Added NOT_AVAILABLE and UNDEFINED versions to DataChangeVersion holder to indicate proper registry state.
4. Added additional trace logging.
5. Minor refactoring and clean up.

Details in [DRILL-6762](https://issues.apache.org/jira/browse/DRILL-6762).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
@@ -21,7 +21,6 @@ import java.io.IOException;
 
 import org.apache.calcite.sql.SqlDescribeSchema;
 import org.apache.calcite.sql.SqlNode;
-import org.apache.calcite.sql.parser.SqlParseException;
 import org.apache.calcite.tools.RelConversionException;
 import org.apache.calcite.tools.ValidationException;
 import org.apache.drill.common.exceptions.UserException;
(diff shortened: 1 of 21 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
```diff
@@ -76,15 +74,17 @@ public static PhysicalPlan getPlan(QueryContext context, String sql) throws SqlP
...
    * @return query physical plan
    */
-  public static PhysicalPlan getPlan(QueryContext context, String sql, Pointer<String> textPlan)
-      throws ForemanSetupException {
+  public static PhysicalPlan getPlan(QueryContext context, String sql, Pointer<String> textPlan) throws ForemanSetupException {
     Pointer<String> textPlanCopy = textPlan == null ? null : new Pointer<>(textPlan.value);
     try {
       return getQueryPlan(context, sql, textPlan);
     } catch (Exception e) {
+      logger.trace("Unable to find function in local function registry. " +
```
vvysotskyi: Is it possible that the exception here may appear for other cases except for the case when the function wasn't found?

arina-ielchiieva: Agree. Updated trace message.

PR #1488, opened 2018-10-04
DRILL-786: Allow CROSS JOIN syntax

ihuzenko:
Details in [DRILL-786](https://issues.apache.org/jira/browse/DRILL-786).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
@@ -51,10 +51,13 @@ import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.planner.logical.DrillLimitRel;
 import org.apache.drill.exec.record.VectorAccessible;
 import org.apache.drill.exec.resolver.TypeCastRules;
+import org.apache.drill.exec.work.foreman.UnsupportedRelOperatorException;
 
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+
+import static org.apache.drill.exec.planner.physical.PlannerSettings.NLJOIN_FOR_SCALAR;
 
 public class JoinUtils {
 
@@ -64,6 +67,11 @@ public class JoinUtils {
     CARTESIAN   // no join condition
   }
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(JoinUtils.class);
+
+  public static final String FAILED_TO_PLAN_CARTESIAN_JOIN = String.format(
+      "This query cannot be planned possibly due to either a cartesian join or an inequality join. %n" +
+          "If a cartesian or inequality join is used intentionally, set the option '%s' to false and try again.",
+      NLJOIN_FOR_SCALAR.getOptionName());
 
   // Check the comparator is supported in join condition. Note that a similar check is also
   // done in JoinPrel; however we have to repeat it here because a physical plan
@@ -125,6 +133,18 @@ public class JoinUtils {
     }
 
     return false;
+  }
+
+  /**
+   * Check if the given RelNode contains any Cartesian join.
+   * Return true if find one. Otherwise, return false.
+   *
+   * @param relNode     {@link RelNode} instance to be inspected
+   * @return            Return true if the given relNode contains Cartesian join.
+   *                    Otherwise, return false
+   */
+  public static boolean checkCartesianJoin(RelNode relNode) {
+    return checkCartesianJoin(relNode, new LinkedList<>(), new LinkedList<>(), new LinkedList<>());
   }
 
   /**
@@ -297,6 +317,16 @@ public class JoinUtils {
    */
   public static boolean hasScalarSubqueryInput(RelNode left, RelNode right) {
     return isScalarSubquery(left) || isScalarSubquery(right);
+  }
+
+  /**
+   * Creates new exception for queries that cannot be planned due
+   * to presence of cartesian or inequality join.
+   *
+   * @return new {@link UnsupportedRelOperatorException} instance
+   */
+  public static UnsupportedRelOperatorException cartesianJoinPlanningException() {
+    return new UnsupportedRelOperatorException(FAILED_TO_PLAN_CARTESIAN_JOIN);
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
@@ -18,7 +18,6 @@
 package org.apache.drill.exec.planner.sql.handlers;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
@@ -106,7 +105,6 @@ import org.apache.drill.exec.server.options.OptionManager;
 import org.apache.drill.exec.util.Pointer;
 import org.apache.drill.exec.work.foreman.ForemanSetupException;
 import org.apache.drill.exec.work.foreman.SqlUnsupportedException;
-import org.apache.drill.exec.work.foreman.UnsupportedRelOperatorException;
 import org.slf4j.Logger;
 
 import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
@@ -294,8 +292,8 @@ public class DefaultSqlHandler extends AbstractSqlHandler {
     } catch (RelOptPlanner.CannotPlanException ex) {
       logger.error(ex.getMessage());
 
-      if (JoinUtils.checkCartesianJoin(relNode, new ArrayList<>(), new ArrayList<>(), new ArrayList<>())) {
-        throw new UnsupportedRelOperatorException("This query cannot be planned possibly due to either a cartesian join or an inequality join");
+      if (JoinUtils.checkCartesianJoin(relNode)) {
+        throw JoinUtils.cartesianJoinPlanningException();
       } else {
         throw ex;
       }
@@ -459,8 +457,8 @@ public class DefaultSqlHandler extends AbstractSqlHandler {
     } catch (RelOptPlanner.CannotPlanException ex) {
       logger.error(ex.getMessage());
 
-      if (JoinUtils.checkCartesianJoin(drel, new ArrayList<>(), new ArrayList<>(), new ArrayList<>())) {
-        throw new UnsupportedRelOperatorException("This query cannot be planned possibly due to either a cartesian join or an inequality join");
+      if (JoinUtils.checkCartesianJoin(drel)) {
+        throw JoinUtils.cartesianJoinPlanningException();
       } else {
         throw ex;
       }
@@ -482,8 +480,8 @@ public class DefaultSqlHandler extends AbstractSqlHandler {
       } catch (RelOptPlanner.CannotPlanException ex) {
         logger.error(ex.getMessage());
 
-        if (JoinUtils.checkCartesianJoin(drel, new ArrayList<>(), new ArrayList<>(), new ArrayList<>())) {
-          throw new UnsupportedRelOperatorException("This query cannot be planned possibly due to either a cartesian join or an inequality join");
+        if (JoinUtils.checkCartesianJoin(drel)) {
+          throw JoinUtils.cartesianJoinPlanningException();
         } else {
           throw ex;
         }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/UnsupportedOperatorsVisitor.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/UnsupportedOperatorsVisitor.java
@@ -36,7 +36,6 @@ import org.apache.calcite.sql.fun.SqlCountAggFunction;
 import org.apache.calcite.sql.SqlCall;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlJoin;
-import org.apache.calcite.sql.JoinType;
 import org.apache.calcite.sql.SqlNode;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.sql.util.SqlShuttle;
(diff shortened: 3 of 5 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
```diff
@@ -33,29 +36,30 @@
...
-import org.apache.drill.common.exceptions.UserException;
-import org.apache.drill.common.logical.data.JoinCondition;
-import org.apache.calcite.rel.RelNode;
-import org.apache.calcite.plan.RelOptUtil;
-import org.apache.calcite.plan.volcano.RelSubset;
-import org.apache.drill.common.types.Types;
-import org.apache.drill.exec.physical.impl.common.Comparator;
-import org.apache.drill.exec.planner.logical.DrillAggregateRel;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
+import org.apache.drill.common.exceptions.UserException;
```
vvysotskyi: It is fine to rearrange imports, but another developer may have different rules for this. Let's postpone this change to the best times when the checkstyle rule for this will be added to the project.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
```diff
@@ -65,6 +69,15 @@
   }
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(JoinUtils.class);
 
+  public static final String FAILED_TO_PLAN_CARTESIAN_JOIN = String.format(
+      "This query cannot be planned possibly due to either a cartesian join or an inequality join. " +
+          "If cartesian or inequality join is used intentionally, set option '%s' to false and try again.",
+      NLJOIN_FOR_SCALAR.getOptionName());
+
+  public static UnsupportedRelOperatorException cartesianJoinPlanningException() {
```
vvysotskyi: Please add Javadoc

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -106,12 +98,17 @@
...
-import org.slf4j.Logger;
-
 import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
 import org.apache.drill.shaded.guava.com.google.common.base.Stopwatch;
+import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableList;
 import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
+import org.apache.drill.shaded.guava.com.google.common.collect.Sets;
+import org.slf4j.Logger;
+
+import java.io.IOException;
```
vvysotskyi: The same for imports.

exec/java-exec/src/test/java/org/apache/drill/exec/planner/sql/CrossJoinTest.java
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  @Test
+  public void testCrossJoinFailsForEnabledOption() throws Exception {
+    enableNlJoinForScalarOnly();
+
+    thrownException.expect(UserRemoteException.class);
+    thrownException.expectMessage(FAILED_TO_PLAN_CARTESIAN_JOIN);
+
+    queryBuilder().sql(
```
vvysotskyi: Please wrap the code into the try block and reset value of the option in finally block in this test and tests below.

exec/java-exec/src/test/java/org/apache/drill/exec/planner/sql/CrossJoinTest.java
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  private static void disableNlJoinForScalarOnly() {
+    client.alterSession(PlannerSettings.NLJOIN_FOR_SCALAR.getOptionName(), false);
+  }
+
+  private static void enableNlJoinForScalarOnly() {
+    client.alterSession(PlannerSettings.NLJOIN_FOR_SCALAR.getOptionName(), true);
+  }
+}
```
vvysotskyi: Please add a new line at the end of the file.

priteshm:
@amansinha100 @HanumathRao can you also review this since you'll have comments on the JIRA.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
```diff
@@ -127,6 +135,18 @@ public static boolean checkCartesianJoin(RelNode relNode, List<Integer> leftKeys
     return false;
   }
 
+  /**
+   * Overloaded version of {@link JoinUtils#checkCartesianJoin(RelNode, List, List, List)}
```
vvysotskyi: Please change the Javadoc to reflect the aim of this method. (just copy it from the overloaded method)

exec/java-exec/src/test/java/org/apache/drill/exec/planner/sql/CrossJoinTest.java
vvysotskyi: Thanks, sounds reasonable.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
```diff
@@ -65,6 +68,11 @@
   }
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(JoinUtils.class);
 
+  public static final String FAILED_TO_PLAN_CARTESIAN_JOIN = String.format(
+      "This query cannot be planned possibly due to either a cartesian join or an inequality join. " +
+          "If cartesian or inequality join is used intentionally, set option '%s' to false and try again.",
```
HanumathRao: "If a cartesian" instead of "If cartesian" and "set the option" instead of "set option"??

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/UnsupportedOperatorsVisitor.java
```diff
@@ -256,14 +255,6 @@ public SqlNode visit(SqlCall sqlCall) {
             "See Apache Drill JIRA: DRILL-1986");
         throw new UnsupportedOperationException();
       }
-
-      // Block Cross Join
-      if(join.getJoinType() == JoinType.CROSS) {
```
HanumathRao: Just to for the clarification is that for implicit cross joins and explicit cross joins do we have joinType set to CROSS by the calcite. If not then can we not distinguish them and only enable the explicit CROSS joins all the time??

exec/java-exec/src/test/java/org/apache/drill/exec/planner/sql/CrossJoinTest.java
HanumathRao: Add a testcase for scalar cross join use case.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/UnsupportedOperatorsVisitor.java
ihuzenko: Yes, at this moment we can detect that it's explicit cross join because here inspected object is SqlNode, but later after conversion to RelNode this `JoinType.CROSS` will be lost, instead we only have `JoinRelType.INNER` in the node. So to preserve original type additional tricks with query context are necessary. 
What you suggest is actually **Option 2** from my comment in [DRILL-786](https://issues.apache.org/jira/browse/DRILL-786) , but as I remember we agreed to go with **Option 3** .  So what are the reasons to move back and use Option 2  ?

exec/java-exec/src/test/java/org/apache/drill/exec/planner/sql/CrossJoinTest.java
ihuzenko: done

ihuzenko:
@HanumathRao I've addressed comments, please take a look.

vvysotskyi:
@HanumathRao, can we merge this PR, or there is something else to do?

HanumathRao:
@ihuzenko  Thank you for making the changes. +1.
@vvysotskyi  Please go ahead with the merge.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
```diff
@@ -65,6 +68,11 @@
   }
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(JoinUtils.class);
 
+  public static final String FAILED_TO_PLAN_CARTESIAN_JOIN = String.format(
+      "This query cannot be planned possibly due to either a cartesian join or an inequality join. " +
```
amansinha100: The descriptive error message is good.  However, since it is close to 200 chars..can you pls add a newline after the first statement (In the JIRA you do show the message split up on 2 lines) ?

amansinha100:
@ihuzenko I have a minor comment about the error message (sorry I just noticed it when I was about to merge your PR).  If you can address it, I will do the merge soon after.

ihuzenko:
> @ihuzenko I have a minor comment about the error message (sorry I just noticed it when I was about to merge your PR). If you can address it, I will do the merge soon after.

Hi, no problems. I've fixed the issue, could you please take a look ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
```diff
@@ -65,6 +68,11 @@
   }
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(JoinUtils.class);
 
+  public static final String FAILED_TO_PLAN_CARTESIAN_JOIN = String.format(
+      "This query cannot be planned possibly due to either a cartesian join or an inequality join. %n" +
```
amansinha100: This should be `\n`.

amansinha100:
The formatting of the message is still incorrect. Also, I ran some test queries and found an issue.  After I set the `enable_nljoin_for_scalar_only` flag to false and set the `slice_target` to a small number (to simulate joining slightly bigger tables),  the cross join planning still fails.  This could be a pre-existing issue but since now we are explicitly supporting cross join, can you check why the query fails.   

0: jdbc:drill:zk=local> alter session set `planner.enable_nljoin_for_scalar_only` = false;
+-------+-------------------------------------------------+
|  ok   |                     summary                     |
+-------+-------------------------------------------------+
| true  | planner.enable_nljoin_for_scalar_only updated.  |
+-------+-------------------------------------------------+

0: jdbc:drill:zk=local> alter session set `planner.slice_target` = 1;
+-------+--------------------------------+
|  ok   |            summary             |
+-------+--------------------------------+
| true  | planner.slice_target updated.  |
+-------+--------------------------------+

0: jdbc:drill:zk=local> explain plan for select count(*) from cp.`tpch/nation.parquet` cross join cp.`tpch/region.parquet`;

Error: UNSUPPORTED_OPERATION ERROR: This query cannot be planned possibly due to either a cartesian join or an inequality join. If a cartesian or inequality join is used intentionally, set the option 'planner.enable_nljoin_for_scalar_only' to false and try again.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/JoinUtils.java
ihuzenko: Hi, %n was used intentionally because it's platform independent. Please check the answer https://stackoverflow.com/questions/7833689/java-string-new-line/38077906#38077906 .

amansinha100: Hi @ihuzenko , have you tested on Sqlline ? (not just through unit tests).  On my MacBook this is how it displays (all one line): 
Error: UNSUPPORTED_OPERATION ERROR: This query cannot be planned possibly due to either a cartesian join or an inequality join. If a cartesian or inequality join is used intentionally, set the option 'planner.enable_nljoin_for_scalar_only' to false and try again.

BTW, can you check the other comment I had provided..we would need to fix that issue.

ihuzenko:
Hi @amansinha100 . Thanks for your active involvement, I investigated the issue and found that:

when option `planner.slice_target` is set to low value, then StreamAggPrule at the first match will try to create 2 phase plan, but method call.transformTo(newRel) won't be called (means that rule results are ignored), because the rule matches at the very early stage of physical planning, there aren't rels matched by new traitSet with physical convention. Currently for simple cross join query with aggregate function, physical rules are applied in order, which is wrong and results into PlanningException:

Prel.ScreenPrule
Prel.ScreenPrule
Prel.ScanPrule
Prel.ScanPrule
ProjectPrule
ProjectPrule
Prel.NestedLoopJoinPrule
ExpandConversionRule
ExpandConversionRule ....

The problem may be fixed either by returning false from StreamAggPrule.create2PhasePlan(call, aggregate) for cross join queries, so then StreamAggPrule will be applied successfully at early stage. Or another option is to change NestedLoopJoinPrule left and join node's distribution trait from ANY to SINGLETON, which then results in activation of ProjectPrule -> StreamAggPrule etc. So then StreamAggPrule is activated again, but now at the right stage when it can be applied successfully.

@vvysotskyi  suggests to proceed with second option, but in order to do this I need to investigate more about how changing of the distribution traits for NestedLoopJoinPrule left and join node impacts physical planning. And then, if possible, try to activate necessary rules without distribution trait change.

@amansinha100 I'd appreciate if you could share your thoughts about the options.

ihuzenko:
I noticed that this issue with aggregate and low slice target doesn't relate only to cross-join queries. As a workaround we can bypass unsuccessful 2-phase aggregation by disabling _planner.enable_multiphase_agg_ option. 

@amansinha100 , @vvysotskyi  can we proceed with merge of this PR ? So then this issue with slice target will be fixed as part of [DRILL-6839](https://issues.apache.org/jira/browse/DRILL-6839) .

amansinha100:
@ihuzenko sorry I missed responding to your earlier comment.  Based on your latest comment, I suppose we can merge this PR but explain the limitation of the Cross Join in the documentation until [DRILL-6839](https://issues.apache.org/jira/browse/DRILL-6839) is fixed.  

Regarding the underlying issue with the 2 phase StreamingAgg, I did not quite understand the comment 
> Currently for simple cross join query with aggregate function, physical rules are applied in order, which is wrong and results into PlanningException:

This is not correct, since rules in Volcano planner are applied iteratively.  So there could be an initial state where a physical rule did not match but subsequently after other transformations, it could match and be triggered.  I would suggest enabling Calcite tracing to get a better idea of what's going on with the distribution traits of the NestedLoopJoin that is causing the downstream StreamingAgg to not get planned with multi-phase aggregation.

amansinha100:
+1  (pls see previous comment about documenting the behavior of cross-join with the limitation).

ihuzenko:
Hi @amansinha100 , thanks for approving the PR and sorry for confusing comment about rule ordering. I've added  documentation note under jira issue.

vvysotskyi:
+1 for this PR and for the proposal to fix this issue in the scope of the separate Jira.

PR #1491, opened 2018-10-04
DRILL-6084: Show Drill functions in WebUI for autocomplete

kkhatua:
Building on top of DRILL-3988 and leveraging DRILL-5868, this allows support for Drill functions to be now available in the WebUI.
If users **wants** UDFs to show up, they should place the UDF jars in the `$DRILL_HOME/jars/3rdparty` directory so that this can be loaded during the Drillbit's startup. 

Here is a snapshot of the WebUI leveraging the list of functions dynamically generated in the internal javascript and available in the WebUI's auto-complete:
image


kkhatua:
@arina-ielchiieva / @ilooner please review only the last commit, as this is rebased off commit #1483 , which (in turn) is rebased off commit #1484 .
Much of the underlying javascript location, etc is identical to what was done for #1279

kkhatua:
@arina-ielchiieva  / @ilooner  can one of you do a review of this minor PR?

exec/java-exec/src/main/resources/ace.mode-sql.template.js
```diff
@@ -0,0 +1,121 @@
+/**
```
arina-ielchiieva: 1. I suggest we keep js scripts / templates under one folder.
Could you please move it rest/static/js folder?
Also could you please do the same for options.dscribe.template.js

2. mode-sql.js has the same code, please factor out the same parts.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -456,21 +471,20 @@ public void close() throws Exception {
...
-    List<OptionValue> options = Lists.newArrayList(publicOptions);
+    List<OptionValue> options = new ArrayList<>(publicOptions);
     Collections.sort(options);
     int numLeftToWrite = options.size();
 
     //Template source Javascript file
     InputStream optionsDescripTemplateStream = Resource.newClassPathResource(OPTIONS_DESCRIBE_TEMPLATE_JS).getInputStream();
     //Generated file
     File optionsDescriptionFile = new File(getTmpJavaScriptDir(), OPTIONS_DESCRIBE_JS);
+    final String FILE_CONTENT_FOOTER = "};";
```
arina-ielchiieva: If variable is in method, then it not necessary to keep its name in upper case.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -495,4 +509,50 @@ private void generateOptionsDescriptionJSFile() throws IOException {
       writer.flush();
     }
   }
+
+  //Generates ACE library javascript populated with list of available SQL functions
+  private void generateFunctionJS() throws IOException {
```
arina-ielchiieva: java-doc

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -495,4 +509,50 @@ private void generateOptionsDescriptionJSFile() throws IOException {
...
+    //Naturally ordered set of function names
+    TreeSet<String> functionSet = new TreeSet<>();
+    //Extracting ONLY builtIn functions (i.e those already available)
+    List<FunctionHolder> builtInFuncHolderList = this.drillbit.getContext().getFunctionImplementationRegistry().getLocalFunctionRegistry()
+        .getAllJarsWithFunctionsHolders().get(LocalFunctionRegistry.BUILT_IN);
+
+    //Build List of usable functions
```
arina-ielchiieva: Please explain what functions are considered to be "usable".

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -495,4 +509,50 @@ private void generateOptionsDescriptionJSFile() throws IOException {
...
+    for (FunctionHolder builtInFunctionHolder : builtInFuncHolderList) {
+      String name = builtInFunctionHolder.getName();
+      if (!name.contains(" ") && name.matches("([a-z]|[A-Z])\\w+")) {
+        functionSet.add(name);
+      } else {
+        logger.debug("Non-alphabetic leading character. Function skipped : {} ", name);
+        skipCount++;
+      }
+    }
+    logger.debug("{} functions will not be available in WebUI : {} ", skipCount);
```
arina-ielchiieva: There are two `{}` but only one parameter.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -495,4 +509,50 @@ private void generateOptionsDescriptionJSFile() throws IOException {
...
+
+    //Generated file
+    File functionsListFile = new File(getTmpJavaScriptDir(), ACE_MODE_SQL_JS);
+    //Template source Javascript file
+    InputStream aceModeSqlTemplateStream = Resource.newClassPathResource(ACE_MODE_SQL_TEMPLATE_JS).getInputStream();
```
arina-ielchiieva: Should we close the stream or `Files.copy` will do this for us?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -495,4 +509,50 @@ private void generateOptionsDescriptionJSFile() throws IOException {
...
+    functionsListFile.deleteOnExit();
```
arina-ielchiieva: It's better to put `functionsListFile.deleteOnExit(); right after `File functionsListFile = new File(getTmpJavaScriptDir(), ACE_MODE_SQL_JS);` file creation for easier tracking.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -495,4 +509,50 @@ private void generateOptionsDescriptionJSFile() throws IOException {
...
+    int numLeftToWrite = functionSet.size();
+    //Create a copy of a template and write with that!
+    java.nio.file.Files.copy(aceModeSqlTemplateStream, functionsListFile.toPath());
+    StringBuilder funcListBldr = new StringBuilder();
```
arina-ielchiieva: Much easier logic can be used here: you just append all functions with `|` and in the end of StringBuilder is not empty substract last char.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: The ACE library cannot support autocomplete for functions like '<' , '!' and '==' . We filter out all functions that don't start with an alphabet to avoid breaking the suggestions that the library will provide after loading. Added this in the comments now.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: 😇  fixed it

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: Agreed! 👍

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: Let's not take a chance. I'll use try resources to manage the autoclose.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
kkhatua: Yeah... this should have been trivial. Silly on my part 🤦‍♂️

kkhatua:
That's a good point. Under the `sys.functions` implementation, there were 2700+ built in functions (when accounting for the different signatures), while  there are over 450 unique function names. From these, I've pruned out non-alphabetical functions, but I am certain there are many functions like `converttonullabledecimal18`.
While I don't know why such functions exist in the registry if they are not expected to be used, my thinking is that I don't think users will use such functions without knowing what they do. We could define a 'black-list' of such function names, but that adds the task of having to maintain this in the long run.

arina-ielchiieva:
@kkhatua what if we add one more parameter to the function template, similar to null handling? This parameter will indicate if function is internal or public. Default is public. This would allow you to filter out all internal functions. You would have to add one more column to the sys.functions table as well.
By the way, does sys.functions table have null handling column? I believe it could be useful.

kkhatua:
@arina-ielchiieva  made some additional changes:
1. Introduced the boolean `internal` field in the `sys.functions` table. For the WebUI, these functions are not available.
2. Added tests to validate the `internal` field.

PR #1496, opened 2018-10-08
DRILL-6783: CAST string literal as INTERVAL MONTH/YEAR works inconsis…

KazydubB:
…tently when selecting from a table with multiple rows


KazydubB:
@vvysotskyi, I've added a unit test, please take a look.

PR #1497, opened 2018-10-08
DRILL-6410: Fixed memory leak in flat Parquet reader

sachouche:
**Problem Description**
- Occasionally, a memory leak is observed within the Parquet reader (flat) when query cancellation is invoked
- I tried to fix this leak in a previous attempt but it seems it is still happening
- Thus far, only QA have been able to observe this issue (and only occasionally)

**Analysis**
- There was a recent breakthrough which gives me hope for addressing this issue
- The leak logged two piece of information: leak size and state of the child allocator
- The state of the child allocator indicated no leak (all allocated bytes were released)
- After code examination, it occurred to me this was happening because the Asynchronous Page Reader task was releasing the Drill buffer while the scan thread was closing the allocator
- The code attempts to cancel asynchronous tasks and then release allocated buffers, though there is one big caveat: the Java FutureTask.cancel(true) doesn't block during the cancellation process; this method merely interrupts the asynchronous task and proceeds
- This means if the asynchronous thread was context switched or doing computation (not blocked waiting), then the fragment cleanup logic can close the allocator before all resources have been released

**Fix**
- There are multiple ways to address this issue; chose the simplest and least intrusive fix
- Created a help class that will be in charge of submitting a callable task C using an ExecutorService E
- The helper class will return a wrapper future that will essentially decorate the cancellation method and nothing else to ensure no performance regressions (or other) are observed
- Added a test to ensure this logic works as expected

**NOTE**
- This fix could be extended to include a set of tasks and implement a blocking wait (unlike the JDK implementation); the fix would also be efficient as it can call cancellation (using no blocking call) and only then perform a wait
- It was felt that providing a non-intrusive & simpler implementation is the way to go

### DIFF
--- a/exec/jdbc-all/pom.xml
+++ b/exec/jdbc-all/pom.xml
@@ -511,7 +511,7 @@
                   This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                   </message>
-                  <maxsize>39000000</maxsize>
+                  <maxsize>39500000</maxsize>
                   <minsize>15000000</minsize>
                   <files>
                    <file>${project.build.directory}/drill-jdbc-all-${project.version}.jar</file>
@@ -571,7 +571,7 @@
                           This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                         </message>
-                        <maxsize>37000000</maxsize>
+                        <maxsize>37500000</maxsize>
                         <minsize>15000000</minsize>
                         <files>
                           <file>${project.build.directory}/drill-jdbc-all-${project.version}.jar</file>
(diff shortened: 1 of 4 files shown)


ilooner:
+1 LGTM. This fix is simple and minimal.

exec/jdbc-all/pom.xml
```diff
@@ -511,7 +511,7 @@
                   This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                   </message>
-                  <maxsize>39000000</maxsize>
+                  <maxsize>39500000</maxsize>
```
vvysotskyi: There was no need to increase the limit for Jar size since in https://github.com/apache/drill/pull/1486 were excluded some classes so the issue with hitting the limit without adding any library was fixed there.

sachouche: @vvysotskyi,
- I rebased several times and didn't see the fix; saw it only after publishing the PR
- Talked internally to the team and was advised to update the pom

BTW - The current way of packaging the jdbc-all jar is unsustainable (unless we accept the fact this jar might indefinitely increase in size). The right fix, is to divide the java-exec project into multiple projects: common, client, and server. This way, the jdbc-all artifact should only depend on common and client.

PR #1498, opened 2018-10-09
DRILL-6775: The schema for empty output is not shown in Drill Web UI

agozhiy:
Removed an excess check for the result emptiness that prevented retrieval of the column names.


agozhiy:
Now it returns a table with no data:
screenshot from 2018-10-09 15-38-06

arina-ielchiieva:
+1

PR #1499, opened 2018-10-10
DRILL-6788: Intermittent unit test failure TestDrillbitResilience.fai…

sohami:
…lsWhenParsing: Query state should be FAILED (and not COMPLETED)


arina-ielchiieva:
+1

PR #1500, opened 2018-10-11
Msgpack format reader

jcmcote:
Implementation of a msgpack format reader

- schema learning
- skip over malformed records
- skip over invalid field names
- skip over records not matching schema
- writing msgpack has not yet been implemented

implementation of a zstandard codec
- only decompression is implemented


vdiravka:
@jcmcote could you add a corresponding JIRA as a prefix in the title of the pull request? Refer the format of other pull requests here: https://github.com/apache/drill/pulls

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackFormatPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+  public RecordWriter getRecordWriter(FragmentContext context, EasyWriter writer) throws IOException {
+    throw new UnsupportedOperationException();
+  }
+
+  @JsonTypeName("msgpack")
+  public static class MsgpackFormatConfig implements FormatPluginConfig {
+
+    public List<String> extensions = ImmutableList.of("mp");
+    private boolean readBinaryAsString = false;
+    private boolean lenient = true;
```
paul-rogers: Would recommend a Javadoc comment for each field. Write these as if speaking to the user (since it is the user that must set these values in the Web console.) Later, you can reuse this information to help the doc. writer produce documentation for your plugin.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackFormatPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+    private boolean printToConsole = true;
```
paul-rogers: This seems like an odd option for a distributed system. If your plugin runs on a dozen hosts, where does the data go?

There is another way to handle this: using a config (not session) option. Just define a new option in your `drill-module.conf` file and set a default value of `false`. Then, at runtime, you can set it to `true` in your `drill-override.conf` file.

I believe this is an area you want to get right the first time. I believe Drill will simply fail to start if you have stored configs that use this field, but your format object is modified to remove this field. Sadly, Drill provides no versioning of these plugin configs, so AFAIK, you can't change them once they are released.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackFormatPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+    private boolean learnSchema = true;
```
paul-rogers: This flag would mean that you have to tinker with the format plugin config to run a query to learn the schema. Since there is no SQL way to do this, it would be a cumbersome manual operation, and won't work well in a multi-user system.

A better (though far from ideal) option is to use a session option: at least that can be set per-user and won't collide with settings by other users. (The JSON and other readers use that trick.)

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackFormatPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+    private boolean useSchema = true;
```
paul-rogers: Again, should this be a format option, or a per-query option? At present, session options (as described above) are the only choice.

Or, this can be automatic: if the root directory (the directory that contains the file, or as identified in the plan for a directory scan) contains a file with a name you specify, read that file to obtain the schema.

The ideal solution would be to use a table function, but, alas, table functions don't inherit values from the plugin config; if you use them, you must specify all options. And, when you do so, the implementation places Draconian restrictions on what you can do in this class. (I think all the members must be simple types and public, if I recall correctly.)

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackFormatPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+    @Override
+    public int hashCode() {
+      final int prime = 31;
+      int result = 1;
+      result = prime * result + ((extensions == null) ? 0 : extensions.hashCode());
+      result = prime * result + (learnSchema ? 1231 : 1237);
+      result = prime * result + (printToConsole ? 1231 : 1237);
+      result = prime * result + (lenient ? 1231 : 1237);
+      result = prime * result + (readBinaryAsString ? 1231 : 1237);
+      result = prime * result + (useSchema ? 1231 : 1237);
```
paul-rogers: There is a handy Guava class for this. Example from this [sample plugin](https://github.com/paul-rogers/drill/blob/5e5dc8652bac7750be075154488394cadc32341d/contrib/format-regex/src/main/java/org/apache/drill/exec/store/easy/regex/RegexFormatConfig.java):

```
  @Override
  public boolean equals(Object obj) {
    if (this == obj) { return true; }
    if (obj == null || getClass() != obj.getClass()) {
      return false;
    }
    final RegexFormatConfig other = (RegexFormatConfig) obj;
    return Objects.equal(regex, other.regex) &&
           Objects.equal(fields, other.fields) &&
           Objects.equal(extension, other.extension);
  }
   @Override
  public int hashCode() {
    return Arrays.hashCode(new Object[] {regex, fields, extension});
  }
```

Now, I have no idea why we hash or compare these objects since doing so makes little sense...

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackFormatPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+      return result;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+      if (this == obj) {
```
paul-rogers: See comment above for the terse way to do this.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackFormatPlugin.java
```diff
@@ -0,0 +1,212 @@
...
+        return false;
+      }
+      return true;
+    }
+
+  }
+
+  @Override
+  public int getReaderOperatorType() {
+    return CoreOperatorType.JSON_SUB_SCAN_VALUE;
```
paul-rogers: This value is used to make sense of metrics. So, either set it to 0 (no information) or go ahead and add a value for your plugin.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackParsingException.java
```diff
@@ -0,0 +1,20 @@
+package org.apache.drill.exec.store.msgpack;
+
+import org.apache.drill.common.exceptions.DrillRuntimeException;
+
+public class MsgpackParsingException extends DrillRuntimeException {
```
paul-rogers: You should not really need this. Instead, use `UserError` to report errors. You'll see it has lots of builder methods to report errors due to the user (wrong config, say) or to external errors (corrupted file, say) or due to internal errors (some invariant is broken indicating a possible bug.) `UserError` has the mechanism to shut down the query and report a meaningful message, along with context, back to the user. See [this example](https://github.com/paul-rogers/drill/blob/5e5dc8652bac7750be075154488394cadc32341d/contrib/format-regex/src/main/java/org/apache/drill/exec/store/easy/regex/RegexRecordReader.java).

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+import org.msgpack.core.MessageUnpacker;
+import org.msgpack.value.MapValue;
+import org.msgpack.value.Value;
+import org.msgpack.value.ValueType;
+
+import io.netty.buffer.DrillBuf;
+
+public class MsgpackReader {
+
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(MsgpackReader.class);
```
paul-rogers: Only LogBack loggers are supported.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+  private final List<SchemaPath> columns;
```
paul-rogers: Let the super class track this for you. It will also parse out any wildcard and similar attributes. See the example cited above.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+//
+//
+//  private void ensure(final int length) {
+//    workBuf = workBuf.reallocIfNeeded(length);
+//  }
+
+  @SuppressWarnings("resource")
+  public void ensureAtLeastOneField(ComplexWriter writer) {
```
paul-rogers: Looks like you happened to look at one of the more complex readers for an example. The example cited above shows a much easier way to do this.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+      while (fieldPath.getChild() != null && !fieldPath.getChild().isArray()) {
+        fieldWriter = fieldWriter.map(fieldPath.getNameSegment().getPath());
+        fieldPath = fieldPath.getChild();
+      }
+      writerList.add(fieldWriter);
+      fieldPathList.add(fieldPath);
+      if (fieldWriter.isEmptyMap()) {
+        emptyStatus.set(i, true);
+      }
+      if (i == 0 && !allTextMode) {
```
paul-rogers: All text mode in JSON means read individual fields, but (since JSON is text), keep them as text to avoid type conflicts.

Not sure it makes sense to do that with MsgPack since fields are typed in the file; you'd have to "un-type" them to convert them to strings.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReaderContext.java
```diff
@@ -0,0 +1,103 @@
...
+
+  public boolean readBinaryAsString;
+
+  public boolean useSchema;
+
+  public long currentRecordNumberInFile() {
+    return parseErrorCount + runningRecordCount + recordCount + 1;
+  }
+
+  public void handleAndRaise(String suffix, Exception e) throws UserException {
```
paul-rogers: Here's where you'd use `UserException` and its builders rather than a custom version.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Collection for tracking empty array writers during reading and storing them
+   * for initializing empty arrays
+   */
+  private final List<ListWriter> emptyArrayWriters = Lists.newArrayList();
+  private boolean allTextMode = false;
+  private MapValueWriter mapValueWriter;
+
+  public MsgpackReader(InputStream stream, MsgpackReaderContext context, DrillBuf managedBuf, List<SchemaPath> columns,
```
paul-rogers: You may find it easier to do the column parsing in the Record reader so you can use Drill's (minimal) built-in helpers and the patterns shown in the example cited above.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+      boolean skipQuery) {
```
paul-rogers: Note also that in a Hadoop system, your reader should handle file blocks. Are MsgPack files splittable? If not, then each file will be read by a single reader; if the file is large, that read will entail remote reads for blocks other than the first.

If MsgPack is splittable, then `FileWork` gives the block offset and you'll need code to handle any file header/footer along with the requested block.

There is a field somewhere that says whether data of this type is splittable; I think in the plugin constuctor.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+    this.skipQuery = skipQuery;
+    rootSelection = FieldSelection.getFieldSelection(columns);
+    EnumMap<ValueType, AbstractValueWriter> valueWriterMap = new EnumMap<>(ValueType.class);
+    valueWriterMap.put(ValueType.ARRAY, new ArrayValueWriter(valueWriterMap, emptyArrayWriters));
+    valueWriterMap.put(ValueType.FLOAT, new FloatValueWriter());
+    valueWriterMap.put(ValueType.INTEGER, new IntegerValueWriter());
+    valueWriterMap.put(ValueType.BOOLEAN, new BooleanValueWriter());
+    valueWriterMap.put(ValueType.STRING, new StringValueWriter());
+    valueWriterMap.put(ValueType.BINARY, new BinaryValueWriter());
+    valueWriterMap.put(ValueType.EXTENSION, new ExtensionValueWriter());
```
paul-rogers: Somewhat confused here; some comments may help.

Presumably, the file can contain any number of FLOAT fields, including 0. Each field needs its own state (reader, parser, whatever.) Each has its own value vector. How does it work to have one "ValueWriter" per type that takes no parameters to say which field or vector is being worked on?

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
```diff
@@ -0,0 +1,192 @@
...
+    if (skipQuery) {
+      writer.rootAsMap().bit("count").writeBit(1);
+    } else {
+      mapValueWriter.writeToMap(value, writer.rootAsMap(), this.rootSelection, schema);
+    }
+  }
+
+//
+//
+//  private void ensure(final int length) {
```
paul-rogers: Not clear that this reader is one that needs an off-heap work buffer.

vvysotskyi:
@jcmcote, in [HADOOP-13578](https://github.com/apache/hadoop/commit/a0a276162147e843a5a4e028abdca5b66f5118da#diff-8983b1157165c3e54c37877bc01da2ec) was added ZStandard Compression to the hadoop library. I think it would be better to collaborate with existing well-tested implementation instead of introducing the custom one.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
```diff
@@ -0,0 +1,114 @@
...
+
+  public void save(MaterializedField mapField, Path schemaLocation) throws IOException {
+    try (FSDataOutputStream out = fileSystem.create(schemaLocation, true)) {
+      SerializedField serializedMapField = mapField.getSerializedField();
+      String data = TextFormat.printToString(serializedMapField);
+      IOUtils.write(data, out);
+    }
+  }
+
+  public MaterializedField merge(MaterializedField existingField, MaterializedField newField) {
```
paul-rogers: This merges field c into map m {a, b}.

Does this code handle merge of types? In file 1, c is a BIGINT. In file 2, it is a DOUBLE. What is the merged type? DOUBLE?

In fact, does this mechanism handle cross-file merges? A scan operator can scan file 1, file 2, file 3 as three distinct readers, but must return a consistent schema. (In fact, file 2 must return the same vectors as file 1.)

The ResultSetLoader has piles of code to handle all this; I'm a bit skeptical that all those cases are handled here. Nor, am I suggesting that you rewrite all that code.

What you may want to do is:

1. Understand the complex requirements when a single scan handles multiple files, when a query has multiple scans, and when those scans are distributed.
2. Understand the kinds of schema evolution that could occur.
3. Look at the implementation of the ResultSetLoader and decide if you want to rewrite that code (maybe you can find a simpler solution), or if you want to leverage the existing code.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
```diff
@@ -0,0 +1,114 @@
...
+    for (MaterializedField newChild : newField.getChildren()) {
+      String newChildName = newChild.getName();
+      MaterializedField foundExistingChild = getFieldByName(newChildName, existingField);
+      if (foundExistingChild != null) {
+        if (foundExistingChild.hasSameTypeAndMode(newChild)) {
+          if (foundExistingChild.getType().getMinorType() == MinorType.MAP) {
+            privateMerge(foundExistingChild, newChild);
+          } // else we already have it
+        } else {
+          // error
```
paul-rogers: MsgPack items are probably all nullable (as in JSON). Does it support only two three as in JSON: Number, String and Boolean?

The challenge here is choosing a numeric format (BIGINT or DOUBLE), and knowing how to convert from one to the other.

It is somewhat difficult to extract the rules being used just from the code. Maybe a Javadoc comment or package-info file to give some background?

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
```diff
@@ -0,0 +1,114 @@
...
+          throw new SchemaChangeRuntimeException("Not the same schema for " + foundExistingChild + " and " + newChild);
+        }
+      } else {
+        existingField.addChild(newChild.clone());
+      }
+    }
+  }
+
+  private MaterializedField getFieldByName(String newChildName, MaterializedField existingField) {
+    for (MaterializedField f : existingField.getChildren()) {
```
paul-rogers: A sequential search won't perform well if called once per row. Or, is this called only once per column?

As it turns out, the TupleSchema is designed with a hash map to speed up name access. Perhaps consider that mechanism if you need frequent access to these columns.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
```diff
@@ -0,0 +1,114 @@
...
+        return f;
+      }
+    }
+    return null;
+  }
+
+  public Path findSchemaFile(Path dir) throws IOException {
+    int MAX_DEPTH = 5;
+    int depth = 0;
+    while (dir != null && depth < MAX_DEPTH) {
```
paul-rogers: Should this be done once in the planner, and passed in the plan to each reader? Otherwise, doing this search for every file might be a bit slow.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchemaWriter.java
```diff
@@ -0,0 +1,166 @@
...
+      break;
+    case VARBINARY:
+      writeBinary(mapWriter, name, listWriter);
+      break;
+    default:
+      logger.warn("UnSupported messagepack type: " + valueType);
+    }
+  }
+
+  private void writeTimestamp(MapWriter mapWriter, String name, ListWriter listWriter) {
```
paul-rogers: Maybe explain how this works... Not clear where the data resides.

contrib/format-msgpack/src/test/java/org/apache/drill/exec/store/msgpack/TestMsgpackRecordReader.java
```diff
@@ -0,0 +1,1415 @@
...
+    rowSetBuilder = newRowSetBuilder();
+    rowSetBuilder.addRow("data");
+    verify(rowSetBuilder.build(), nextRowSet());
+  }
+
+  @Test
+  public void testSchemaMapOfMapShort() throws Exception {
+    learnModel();
+    String sql = "select root.mapOfmap.aMap.sho as w from dfs.data.`secondBatchHasCompleteModel.mp` as root";
+    rowSetIterator = client.queryBuilder().sql(sql).rowSetIterator();
```
paul-rogers: If this is a one-batch result (since only one row), you can use rowSet() to get the row set without an iterator or the need to skip the first, empty, batch.

contrib/codec-zstd/src/main/java/org/apache/hadoop/io/compress/zstd/ZstdDecompressor.java
```diff
@@ -0,0 +1,270 @@
...
+import org.apache.hadoop.io.compress.Decompressor;
+
+import com.github.luben.zstd.ZstdDirectBufferDecompressingStream;
+import com.github.luben.zstd.util.Native;
+
+/**
+ * A {@link Decompressor} based on the zstandard compression algorithm.
+ */
+public class ZstdDecompressor implements Decompressor {
+  private static final Log LOG = LogFactory.getLog(ZstdDecompressor.class.getName());
```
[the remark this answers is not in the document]
jcmcote: These classes are an implementation of the Hadoop codec. For example when drill reads gz files it is relying on the Hadoop library to handle gzip decompression.

The way you contribute a new codec is via the Java Service Plugin mechanism. That is META-INF/services folder.

I've implemented the interface org.apache.hadoop.io.compress.CompressionCodec using the zstandard java library.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
jcmcote: msgpack files are a sequence of messages. In theory it can be split but you would need to scan the file using the msgpack library to know where you could split them. How would a csv file be split? Is there code that scans the file and determines where the offset should be?

jcmcote:
> @jcmcote, in [HADOOP-13578](https://github.com/apache/hadoop/commit/a0a276162147e843a5a4e028abdca5b66f5118da#diff-8983b1157165c3e54c37877bc01da2ec) was added ZStandard Compression to the hadoop library. I think it would be better to collaborate with existing well-tested implementation instead of introducing the custom one.

Agreed. When will drill pickup the new version of Hadoop. Is that a big deal to upgrade the version of Hadoop used?

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackReader.java
jcmcote: I'll put more comments in the code. Basically this is a switch implemented using an EnumMap. In the  ComplexValueWriter I use this switch to lookup what class will handle writing a value type. Here's the line of code from that writeElement method

      valueWriterMap.get(value.getValueType()).write(value, mapWriter, fieldName, listWriter, selection, schema);
So based on the value type I get the corresponding writer class to use.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
```diff
@@ -0,0 +1,114 @@
...
+          throw new DrillRuntimeException("Failed to read schema file: " + schemaLocation, e);
+        }
+        SerializedField read = newBuilder.build();
+        previousMapField = MaterializedField.create(read);
+      }
+    }
+    return previousMapField;
+  }
+
+  public void save(MaterializedField mapField, Path schemaLocation) throws IOException {
```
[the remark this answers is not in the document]
jcmcote: No it does not handle multi-thread cases. Right now I turn on learning mode and submit queries of the form
select * from dfs.root.`dir/aSingleFile.mp`
I can scan more files but only one at a time. I then copy that schema file in all directories container that type of data. That's fine for my purpose right now. It's a bit manual and not ready for prime time as it is.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
jcmcote: I don't handle multi-files in parallel right now.

Schema evolution is also not handled. You will recall from my questions in the dev mailing list, I have come across cases where an array would have VARCHAR and VARBINARY. In that case I would get a schema with array of type VARCHAR (having skipped over the error of not being able to write VARBINARY into that array.

By looking at the logs I can see these values are being skipped. I then manually update the schema to tell it I want that array to be VARBINARY. From now one when I read the VARCHAR values are coerced into VARBINARY.

So for now I use the schema learning to discover the bulk of the structure and fix any edge cases manually.

But I think the bases are there to build a smarter discovery mechanism where it could evolve the schema as it sees more of the data.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
jcmcote: The fields are all nullable, correct. It supports many "sizes" for Number but in the java library I'm using it rolls all those up into Number. So yes there's just a few types to handle. But there is a distinction between Number and Float. So it's not as difficult as in JSON where you might see "1", "2" then "3.1".

Again I will add more comments and documentation.

contrib/format-msgpack/src/main/java/org/apache/drill/exec/store/msgpack/MsgpackSchema.java
jcmcote: yes, I've already changed that. Thanks.

vdiravka:
@jcmcote There is a Jira ticket for Hadoop libs version update: [DRILL-6540](https://issues.apache.org/jira/browse/DRILL-6540).
There is an issue related to `commons-logging`, see [details](https://issues.apache.org/jira/browse/DRILL-6540?focusedCommentId=16611930&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16611930).
Also there is my "work in progress" branch in the ticket.

vvysotskyi:
@jcmcote, Is it possible to split this pull request into two parts: leave here only changes connected with Msgpack format reader, and continue work on Compression codecs in the scope of a separate Jira after upgrade of Hadoop library is done?

jcmcote:
@vvysotskyi Sure I can split them up. Should be easy to do.

arina-ielchiieva:
@jcmcote taking into account that there is ongoing work to provide schema using file (https://issues.apache.org/jira/browse/DRILL-6835). You might consider waiting for those changes to be published to use common approach of reading and writing schema files.

jcmcote:
okay sounds good

cgivre:
Hi @jcmcote 
Are you still interested in completing this PR?  Recently, the enhanced vector format PRs were committed and could make this better and easier.

cgivre:
Hi @jcmcote Are you still interested in completing this PR?

PR #1503, opened 2018-10-14
DRILL-6795: Upgrade Janino compiler from 2.7.6 to 3.0.10

vvysotskyi:
For details please reffer [DRILL-6795](https://issues.apache.org/jira/browse/DRILL-6795).


arina-ielchiieva:
@vdiravka could you please review?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/ModifiedUnparser.java
```diff
@@ -0,0 +1,110 @@
...
+      pw.println(": {");
+      pw.print(AutoIndentWriter.INDENT);
+      unparseStatements(methodDeclarator.optionalStatements);
+      pw.print(AutoIndentWriter.UNINDENT);
+      pw.println("}");
+      pw.print(' ');
+    }
+  }
+
+  private void visitReturnStatement(Java.ReturnStatement returnStatement) {
```
vdiravka: It is a potential ticket for improvement in Janino, to allow usage of default implementations of visitor's methods and possibility to override it, if it is necessary.
Once it will be done in Janino, the overridden `unparseBlockStatement()` method could be removed.
Do you agree?

vvysotskyi: Agree, created https://github.com/janino-compiler/janino/issues/61 for this. But anyway, I think this is not a blocker for this PR.

vdiravka: Thanks for creating the ticket.
Agree, the current solution is acceptable.

PR #1506, opened 2018-10-17
DRILL-6793: FragmentExecutor cannot send its final state for the case…

KazydubB:
… when RootExec root wasn't initialized

### DIFF
--- a/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
+++ b/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
@@ -531,7 +531,7 @@ public class MaprDBJsonRecordReader extends AbstractRecordReader {
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder("MaprDBJsonRecordReader[Table=")
-        .append(table.getPath());
+        .append(table != null ? table.getPath() : null);
     if (reader != null) {
       sb.append(", Document ID=")
           .append(IdCodec.asString(reader.getId()));
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/avro/AvroRecordReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/avro/AvroRecordReader.java
@@ -403,9 +403,11 @@ public class AvroRecordReader extends AbstractRecordReader {
   public String toString() {
     long currentPosition = -1L;
     try {
-      currentPosition = reader.tell();
+      if (reader != null) {
+        currentPosition = reader.tell();
+      }
     } catch (IOException e) {
-      logger.trace("Unable to obtain reader position: " + e.getMessage());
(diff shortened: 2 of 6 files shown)


contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
```diff
@@ -530,8 +530,10 @@ public void close() {
 
   @Override
   public String toString() {
-    StringBuilder sb = new StringBuilder("MaprDBJsonRecordReader[Table=")
-        .append(table.getPath());
+    StringBuilder sb = new StringBuilder("MaprDBJsonRecordReader[Table=");
+    if (table != null) {
```
vvysotskyi: For the case when `table` is null, the string will be formed incorrectly - `Table=` should be appended only if `table != null`

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
```diff
@@ -530,8 +530,10 @@ public void close() {
...
+      sb.append(table.getPath());
+    }
     if (reader != null) {
       sb.append(", Document ID=")
```
vvysotskyi: Also, please move `, `string here and below to the previous appenders.

exec/java-exec/src/main/java/org/apache/drill/exec/store/avro/AvroRecordReader.java
```diff
@@ -403,7 +403,9 @@ public void close() {
   public String toString() {
     long currentPosition = -1L;
     try {
-      currentPosition = reader.tell();
+      if (reader != null) {
+        currentPosition = reader.tell();
+      }
     } catch (IOException e) {
       logger.trace("Unable to obtain reader position: " + e.getMessage());
```
vvysotskyi: Please log the stack trace of the exception.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
KazydubB: Thanks for pointing it out, done.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/MaprDBJsonRecordReader.java
KazydubB: Made previous field to be present in any case. It seems there is no need in doing so.

exec/java-exec/src/main/java/org/apache/drill/exec/store/avro/AvroRecordReader.java
KazydubB: Done.

PR #1508, opened 2018-10-19
DRILL-6804: Simplify usage of OperatorPhase in HashAgg.

ilooner:


ilooner:
@Ben-Zvi please review

Ben-Zvi:
I don't understand the purpose of this Jira - the changed code does not look simpler than the original.   E.g., is was much simpler to read `is2ndPhase` than now `popConfig.getAggPhase().isSecondPhase()` .  
   The enum `OperatorPhase` is not adding any functionality. The only issue is the debug string `getPhaseName()` which can be implemented as a private one line method.

ilooner:
@Ben-Zvi the idea is to use an object oriented approach for structuring the code. Instead of computing values like is2ndPhase from the OperatorPhase object within HashAgg itself, we can make the values a property of the OperatorPhase itself.

It also helps reduce the number of variables present in HashAgg template from these three

```
  private boolean isTwoPhase = false;
  private boolean is2ndPhase = false;	
  private boolean is1stPhase = false;
```

to just a single operatorPhase variable. 

This is useful when we move the code in HashAggTemplate into separate HashAggPartition, PartitionEviction, and HashAggMemoryCalculator classes since it minimizes the number of variables that need to be passed to these classes. This change was done in preparation for that restructuring.

Perhaps I could changes the names to from `operatorPhase.isSecondPhase()` to something like this `phase.is2nd()`. Similarly I could change the line you pointed out from `popConfig.getAggPhase().isSecondPhase()` to two separate lines like

```
OperatorPhase phase = popConfig.getAggPhase();
phase.is2nd()
```

Let me know your thoughts.

ilooner:
@Ben-Zvi I changed the names to be more similar to what they were before. The previous comment includes the technical justification for this change. Please do another round of review and let me know if you have anymore comments.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/physical/AggPrelBase.java
```diff
@@ -47,7 +47,39 @@
...
+      this.is1st = is1st;
+      this.name = name;
+    }
+
+    public boolean hasTwo() {
+      return hasTwo;
+    }
+
+    public boolean is1st() {
+      return is1st;
```
Ben-Zvi: Note that if some future code would call `is1st()` on a single phase, the result would be TRUE !!

ilooner-mapr: Thanks for catching this, fixed. And thanks for the review.

PR #1511, opened 2018-10-25
Edit, per Edit Tracker request

kimbostrom:
Copyedited to increase brevity.
Suggest making calculation descriptions parallel, beginning with the same wording.
Suggest omitting cross-references to validate text.
Suggest omitting final entry and Note: why document what is not useful? Also, the second sentence of the description seems unclear: what is "a value of the commonly shared length"?


kimbostrom:
Edit for Ashwin

kkhatua:
@kimbostrom  could you add a corresponding JIRA as a prefix in the title of the pull request? Refer the format of other pull requests here: https://github.com/apache/drill/pulls

PR #1513, opened 2018-10-28
DRILL-6809: Handle repeated map in schema inference

paul-rogers:
It turns out that the RowSet utilities build a repeated map without including the hidden `$offsets$` vector in the metadata for the map. But, other parts in Drill do include this vector.

The RowSet behavior might be a bug which can be addressed in another PR.

This PR:

* Adds unit tests for map accessors at the row set level. Looks like these were never added originally. They are a simplified form of the ResultSetLoader map tests.
* Verified that the schema inference can infer a schema from a repeated map (using the RowSet style.)
* Added a test to reproduce the case from the bug.
* Made a tweak to the RowSetBuilder to allow access to the RowSetWriter which is needed by the new tests.
* Couple of minor clean-ups.


arina-ielchiieva:
+1, LGTM.

PR #1519, opened 2018-11-01
DRILL-6760: Retain original exception in Verbose Error Message

lushuifeng:
please see [DRILL-6760](https://issues.apache.org/jira/browse/DRILL-6760)
There are some cases that verbose error message in profile page contains no original exception stack trace.

### DIFF
--- a/common/src/main/java/org/apache/drill/common/exceptions/UserException.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/UserException.java
@@ -742,6 +742,9 @@ public class UserException extends DrillRuntimeException {
     if (getCause() != null) {
       // some unit tests use this information to make sure a specific exception was thrown in the server
       builder.setException(ErrorHelper.getWrapper(getCause()));
+    } else {
+      // not a wrapper exception
+      builder.setException(ErrorHelper.getWrapper(this));
     }
     return builder.build();
   }
--- a/common/src/main/java/org/apache/drill/common/exceptions/UserRemoteException.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/UserRemoteException.java
@@ -19,6 +19,8 @@ package org.apache.drill.common.exceptions;
 
 import org.apache.drill.exec.proto.UserBitShared.DrillPBError;
 
+import static org.apache.drill.common.util.DrillExceptionUtil.getThrowable;
+
 /**
  * Wraps a DrillPBError object so we don't need to rebuilt it multiple times when sending it to the client. It also
  * gives access to the original exception className and message.
@@ -28,7 +30,7 @@ public class UserRemoteException extends UserException {
   private final DrillPBError error;
 
   public UserRemoteException(DrillPBError error) {
-    super(error.getErrorType(), "Drill Remote Exception", null);
+    super(error.getErrorType(), "Drill Remote Exception", getThrowable(error.getException()));
     this.error = error;
   }
 
--- a/common/src/main/java/org/apache/drill/common/util/DrillExceptionUtil.java
+++ b/common/src/main/java/org/apache/drill/common/util/DrillExceptionUtil.java
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.common.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.drill.exec.proto.UserBitShared;
+
+import java.lang.reflect.Constructor;
+import java.util.Arrays;
+
+/**
+ * Utility class which contain methods for conversion Drill ProtoBuf Error and Throwable
+ */
+public class DrillExceptionUtil {
+
+  /**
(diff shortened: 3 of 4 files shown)


lushuifeng:
@arina-ielchiieva Could you please review this?
The tests are succeeded but timeout

common/src/main/java/org/apache/drill/common/util/DrillExceptionUtils.java
```diff
@@ -0,0 +1,96 @@
...
+ */
+package org.apache.drill.common.util;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.drill.exec.proto.UserBitShared;
+
+import java.lang.reflect.Constructor;
+import java.util.Arrays;
+
+public class DrillExceptionUtils {
```
arina-ielchiieva: 1. Please add java doc for the whole util class.
2. Please add unit tests for each public static method.
3. DrillExceptionUtil or DrillExceotionUtils?
4. Maybe move this helper class to drill-commons module, package - `org.apache.drill.common.util` ?

common/src/main/java/org/apache/drill/common/util/DrillExceptionUtils.java
```diff
@@ -0,0 +1,96 @@
...
+      }
+      throwable.setStackTrace(stackTrace);
+      return throwable;
+    } catch (Throwable t) {
+      logger.error("Failed to get throwable from PB", t);
+      return null;
+    }
+  }
+
+  private static Throwable getInstance(String className, String message, Throwable inner) throws ReflectiveOperationException {
```
arina-ielchiieva: Please add java doc.

exec/java-exec/src/test/java/org/apache/drill/exec/ProfileTest.java
```diff
@@ -0,0 +1,65 @@
...
+  public static void setup() throws Exception {
+    startCluster(ClusterFixture.builder(dirTestWatcher));
+  }
+
+  @Test
+  public void verboseErrorTest() throws Exception {
+    DrillbitContext context = cluster.drillbit().getContext();
+    PersistentStore<UserBitShared.QueryProfile> completed = context.getProfileStoreContext().getCompletedProfileStore();
+    String sql = "select * from cp.`jsoninput/DRILL-2350.json`";
+    QueryBuilder.QuerySummaryFuture future = client.queryBuilder().sql(sql).futureSummary();
```
arina-ielchiieva: Why are getting future summary? Why not just get `QueryBuilder.QuerySummary testQuerySummary = queryBuilder().sql(testQuery).run();`?

Also you can add check that `assertTrue(testQuerySummary.failed());`

exec/java-exec/src/test/java/org/apache/drill/exec/ProfileTest.java
```diff
@@ -0,0 +1,65 @@
...
+    QueryBuilder.QuerySummary summary = future.get();
+    String queryId = summary.queryIdString();
+    UserBitShared.QueryProfile profile = completed.get(queryId);
+    if (profile == null) {
```
arina-ielchiieva: How do you know that profile is absent sometimes if this does not fail on your laptop? Not sure if we need test that fails randomly and adding logger looks more like a hack. If we cannot stabilize the test, then we might consider just removing it.

exec/java-exec/src/test/java/org/apache/drill/exec/ProfileTest.java
```diff
@@ -0,0 +1,65 @@
...
+      logger.error("No profile found, queryId:" + queryId);
+      return;
+    }
+    String verboseError = profile.getVerboseError();
+    String extraError = verboseError.substring(profile.getError().length());
+    Pattern pattern = Pattern.compile("[a-z0-9]+(\\.[a-zA-Z0-9$]+)+\\(\\):\\d+");
```
arina-ielchiieva: Please add comment explaining the pattern

exec/java-exec/src/test/java/org/apache/drill/exec/ProfileTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.junit.Assert.fail;
+
+public class ProfileTest extends ClusterTest {
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProfileTest.class);
```
arina-ielchiieva: private? or actually remote the logger at all.

common/src/main/java/org/apache/drill/common/util/DrillExceptionUtils.java
```diff
@@ -0,0 +1,96 @@
...
+public class DrillExceptionUtils {
+  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DrillExceptionUtils.class);
+
+  public static Throwable getThrowable(UserBitShared.DrillPBError error) {
+    UserBitShared.ExceptionWrapper wrapper = error.getException();
+    return getThrowable(wrapper);
+  }
+
+  /**
+   * recreate throwable from exception PB which received from remote or local node.
```
arina-ielchiieva: recreate -> Recreate
PB -> protobuf

arina-ielchiieva:
@lushuifeng do you have Apache Jira account? I want to assign Jira (https://issues.apache.org/jira/browse/DRILL-6760) to you.

exec/java-exec/src/test/java/org/apache/drill/exec/ProfileTest.java
lushuifeng: The test always failed and verbose error is absent, What's interesting is that the queryId sometimes is not stored in Completed Store in Travis tests which is not reproducible on my laptop.
You can launch the sqlline and run the same sql test in `ProfileTest` and take a look at the error page.
I will provide more details tomorrow.

arina-ielchiieva: Then just remove the unit test.

lushuifeng:
@arina-ielchiieva Changes are made according to your suggestions, could you please take another look? (Building job timeouts)

Here is the comparison of before and after the change with the same query.
screen shot 2018-11-08 at 11 19 29

arina-ielchiieva:

lushuifeng:
Done.

PR #1524, opened 2018-11-06
DRILL-6830: Remove Hook.REL_BUILDER_SIMPLIFY handler after use

lushuifeng:
Please see [DRILL-6830](https://issues.apache.org/jira/projects/DRILL/issues/DRILL-6830?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC).
Other Hook handlers has been checked, No such case found.


lushuifeng:
@vvysotskyi Could please review this?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -377,11 +377,15 @@ public RelRoot toRel(final SqlNode validatedNode) {
...
 
-    //To avoid unexpected column errors set a value of top to false
-    final RelRoot rel = sqlToRelConverter.convertQuery(validatedNode, false, false);
-    return rel.withRel(sqlToRelConverter.flattenTypes(rel.rel, true));
+    try {
+      //To avoid unexpected column errors set a value of top to false
+      final RelRoot rel = sqlToRelConverter.convertQuery(validatedNode, false, false);
+      return rel.withRel(sqlToRelConverter.flattenTypes(rel.rel, true));
+    } finally {
+      closeable.close();
```
vvysotskyi: Could you please replace it with `try with resources` block?

lushuifeng: fixed and squashed

lushuifeng:
@vvysotskyi There are some type mismatch errors in the tests since RelBuilder.simplify is not always 
 false so simplifier.simplifyPreservingType will be invoked in some tests.
 These errors are reproducible by running one test at a time, say running `TestCaseNullableTypes#testCaseNullableTypesVarchar` alone on master, the tests are passed if running all tests in TestCaseNullableTypes at once.
Should I fixed these errors in this PR or create another ticket? 
I have to say that I am not quite familiar with Calcite.

vvysotskyi:
@lushuifeng, In future, please check that all unit tests passed before opening the pull request [1].

lushuifeng:
@vvysotskyi My apologies! I will pay more attention next time.

vvysotskyi:
We can't merge PR if it causes tests failures. This test fails because after closing `Hook` in `SqlConverter.toRel()` method, rel builder is used during optimizing stage.
I would recommend moving `Hook` usage into the place, where it will be closed only after finishing planning.

lushuifeng:
I think the tests are failed because the misused of Hook, running TestCaseNullableTypes#testCaseNullableTypesVarchar alone on master gets the error either.
we can put `static {
    Hook.REL_BUILDER_SIMPLIFY.add(Hook.propertyJ(false));
  }` somewhere to keep the old way instead of adding a new one to the Hook, but the final solution is to fix the errors in the tests.
I think it would still get the errors when close the hook after finishing planning.

lushuifeng:
All tests are passed with `mvn clean install` if the static block is added.
I wonder if that is OK as a workaround. I will try to fix these errors later.

vvysotskyi:
@lushuifeng, you are right that the problem connected with `TestCaseNullableTypes#testCaseNullableTypesVarchar` is in Calcite. The initial goal for adding `Hook.REL_BUILDER_SIMPLIFY.add(Hook.propertyJ(false));` was to avoid failures connected with treating empty strings as null feature, but looks like it was fixed in another place.

The interesting thing is that problem connected with `TestCaseNullableTypes#testCaseNullableTypesVarchar` failure was fixed after 1.17 version, so when Drill is rebased onto Calcite 1.18, `Hook.REL_BUILDER_SIMPLIFY.add(Hook.propertyJ(false));` may be removed.

@ihuzenko, since you are working on Calcite rebase, could you please also take care of it?

lushuifeng:
@vvysotskyi thanks, that's great if calcite has fixed this issue.

vvysotskyi:
Let's keep this PR opened until [DRILL-6862](https://issues.apache.org/jira/browse/DRILL-6862) is resolved to be sure that there weren't changes in Calcite which will force us to use `Hook.REL_BUILDER_SIMPLIFY`.

ihuzenko:
Hello @vvysotskyi and  @lushuifeng . I think this PR may be closed, because I moved hook initialization to static block in my [DRILL-6862](https://github.com/apache/drill/pull/1631) pull request.

PR #1527, opened 2018-11-08
DRILL-4456: Add Hive translate UDF

vvysotskyi:
- Added renaming of Hive `TRANSLATE` function to `TRANSLATE3` during function registration.

For problem description [DRILL-4456](https://issues.apache.org/jira/browse/DRILL-4456).

### DIFF
--- a/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/expr/fn/HiveFunctionRegistry.java
+++ b/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/expr/fn/HiveFunctionRegistry.java
@@ -20,9 +20,12 @@ package org.apache.drill.exec.expr.fn;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Stream;
 
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.sql.SqlOperatorBinding;
+import org.apache.calcite.sql.fun.OracleSqlOperatorTable;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.type.SqlReturnTypeInference;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.drill.common.config.DrillConfig;
@@ -38,6 +41,7 @@ import org.apache.drill.exec.planner.sql.DrillOperatorTable;
 import org.apache.drill.exec.planner.sql.HiveUDFOperator;
 import org.apache.drill.exec.planner.sql.HiveUDFOperatorWithoutInference;
 import org.apache.drill.exec.planner.sql.TypeInferenceUtils;
+import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableMap;
 import org.apache.hadoop.hive.ql.exec.Description;
 import org.apache.hadoop.hive.ql.exec.UDF;
 import org.apache.hadoop.hive.ql.udf.UDFType;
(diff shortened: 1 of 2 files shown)


arina-ielchiieva:
@vdiravka could you please review.

vdiravka:
@vvysotskyi The PR with two commits make sense if that tickets are interdependent.
Looks like this case is different. So please split current PR to the two different ones.

vvysotskyi:
@vdiravka, done: https://github.com/apache/drill/pull/1528

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/expr/fn/HiveFunctionRegistry.java
```diff
@@ -119,10 +132,26 @@ public void register(DrillOperatorTable operatorTable) {
...
+  }
 
-    for(int i=0; i<names.length;i++) {
-      methods.put(names[i].toLowerCase(), clazz);
+  /**
+   * Checks using predicate from {@link #FUNCTION_REPLACE_MAP}
+   * whether specified function names should be replaced
+   * with the names from the map.
+   */
+  private String[] renameUDF(String[] namesToCheck) {
```
arina-ielchiieva: Please add `@param` and `@return`.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/expr/fn/HiveFunctionRegistry.java
```diff
@@ -48,8 +52,17 @@
...
 
-public class HiveFunctionRegistry implements PluggableFunctionRegistry{
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(HiveFunctionRegistry.class);
+public class HiveFunctionRegistry implements PluggableFunctionRegistry {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(HiveFunctionRegistry.class);
+
+  /**
+   * Map for renaming Hive UDFs whose names satisfy the predicate in the key by names from the map value.
+   */
+  private static final Map<Predicate<String[]>, String[]> FUNCTION_REPLACE_MAP = ImmutableMap.<Predicate<String[]>, String[]> builder()
```
arina-ielchiieva: Please add better description.

vdiravka: @vvysotskyi Try to simplify the logic for `FUNCTION_REPLACE_MAP`. Possibly try to get rid from `String[]` (you can add all function names to this Map), so in result it can be similar to 
https://github.com/apache/drill/blob/master/logical/src/main/java/org/apache/drill/common/expression/fn/FunctionReplacementUtils.java#L42

vvysotskyi: Thanks, reworded a description and simplified the map to use target function names as keys and predicates for checking as values.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/expr/fn/HiveFunctionRegistry.java
vvysotskyi: Thanks, added.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/expr/fn/HiveFunctionRegistry.java
```diff
@@ -102,15 +116,15 @@ public void register(DrillOperatorTable operatorTable) {
     }
   }
 
-  private <C,I> void register(Class<? extends I> clazz, ArrayListMultimap<String,Class<? extends I>> methods) {
+  private <I> void register(Class<? extends I> clazz, ArrayListMultimap<String, Class<? extends I>> methods) {
```
ihuzenko: If it's not required to have separate renameUDF(names) method, then it's possible to simplify body of this method: 

```
private <C, I> void register(Class<? extends I> clazz, ArrayListMultimap<String, Class<? extends I>> methods) {
    Description desc = clazz.getAnnotation(Description.class);
    Stream<String> names;
    if (desc != null) {
      names = Stream.of(desc.name().split(",")).map(String::trim);
    } else {
      names = Stream.of(clazz).map(Class::getName)
          .map(name -> name.replace('.', '_'));
    }
    names.map(String::toLowerCase)
        .map(funName -> FUNCTION_REPLACE_MAP.getOrDefault(funName, funName))
        .forEach(udfFunName -> methods.put(udfFunName, clazz));

    UDFType type = clazz.getAnnotation(UDFType.class);
    if (type != null && !type.deterministic()) {
      nonDeterministicUDFs.add(clazz);
    }
  }

```

vvysotskyi: Thanks, done

PR #1529, opened 2018-11-09
DRILL-6837: Missing reference for dynamically created Javascript library

kkhatua:
A PR for DRILL-6084 broke the auto-complete library required for Edit Query page, with the removal of the static javascript library.
https://github.com/apache/drill/pull/1491/files#diff-2bb3c04f1a9eead1faf57212cce5d2a9

The dynamically loaded script is not defined in the Freemarker template page.
The fix is trivial.

This is the console error:
image


arina-ielchiieva:
+1

PR #1533, opened 2018-11-11
DRILL-6843: Update SchemaBuilder comment to match implementation

paul-rogers:
The schema builder API evolved a while back. The class provides a detailed comment to explain usage. That comment was out of date. This fix updates the comment to match the code.


arina-ielchiieva:
+1

PR #1536, opened 2018-11-12
DRILL-6039: Fixed drillbit.sh script to do graceful shutdown

dvjyothsna:
@sohami please review


exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -433,6 +442,22 @@ public void run() {
     public StatusThread() {
       // assume this thread is created by a non-daemon thread
       setName("WorkManager.StatusThread");
+  }
+  private void pollShutdown(Drillbit drillbit) throws IOException, InterruptedException {
+      final Path path = FileSystems.getDefault().getPath(System.getenv("DRILL_PID_DIR"));
+      try (final WatchService watchService = FileSystems.getDefault().newWatchService()) {
+        path.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY, StandardWatchEventKinds.ENTRY_CREATE);
+        while (true) {
+          final WatchKey wk = watchService.take();
```
sohami: This design is not correct and has issues.

- You are using `WorkManager` `StatusThread` for registering a `WatchService` and then doing a blocking call `take()` which will make this thread stuck until there is any event seen by `WatchService`. Hence `StatusThread` will not be able to perform it's actual job of sending status of `runningFragments` to it's `Foreman`.
- The `pollShutdown` method is calling `Drillbit.close()` which is in `StatusThread`. Now `Drillbit.close() `calls close on `WorkManager` and which calls `interrupt` on `StatusThread`. There is a cycle here. What it means to call interrupt on itself ? Usually interrupt is used by another thread to wake up a blocking thread. I think based on interrupt call the interrupt flag of status thread will be set. Then when it comes out of `pollShutdown` and calls sleep it will hit interrupted exception. It might work but I don't like the idea where WorkManager which is part of Drillbit has a thread which calling `Drillbit.close()`. 

Please create a separate thread within Drillbit class itself to do this. But in `Drillbit::close()` method you need to still `interrupt` this new thread for cases when `close` it called from some other path. But you can maintain a flag which will indicate if `Thread.interrupt()` should be called or not.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -285,6 +286,10 @@ public synchronized void close() {
       if (storeProvider != profileStoreProvider) {
         AutoCloseables.close(profileStoreProvider);
       }
+      File f = new File(System.getenv("DRILL_PID_DIR") + "/.graceful");
```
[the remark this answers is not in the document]
dvjyothsna: Can we get that config param in the script?

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
```diff
@@ -442,6 +467,14 @@ public void run() {
       // StatusThread is started
       final Controller controller = dContext.getController();
       final DrillbitEndpoint localBitEndPoint = dContext.getEndpoint();
+      try {
+        pollShutdown(drillbit);
+      } catch (IOException e) {
+        e.printStackTrace();
```
arina-ielchiieva: I think we should not use `e.printStackTrace();` but rather do proper error handling or logging.

distribution/src/resources/drillbit.sh
```diff
@@ -94,11 +94,24 @@ waitForProcessEnd()
   commandName=$2
   kill_drillbit=$3
   processedAt=`date +%s`
+  triggered_shutdown=false
+  FILE=$DRILL_PID_DIR/.graceful
```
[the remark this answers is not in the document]
kkhatua: I think making it hidden is practical, since this is only serves the purpose of signalling and is not expected to be around visible. 
Also, @dvjyothsna , it might be worth changing the variable name to something like `GRACEFUL_SIGFILE`

distribution/src/resources/drillbit.sh
```diff
@@ -94,11 +94,24 @@ waitForProcessEnd()
...
      echo -n "."
      sleep 1;
+     #Incase of graceful shutdown, create .graceful file and wait till it is deleted to trigger kill command.
+     if [ "$kill_drillbit" = false ]; then
+       if [ "$triggered_shutdown" = false ]; then
+         touch $DRILL_PID_DIR/.graceful
+         triggered_shutdown=true
+       else
+         if [ ! -f "$FILE" ]; then
+           kill $pidKilled > /dev/null 2>&1;
```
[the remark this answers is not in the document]
kkhatua: @sohami I think she is just playing it safe by doing a clean up. 
A suggestion... for all the `kill` commands, provide the signal as part of the arguments.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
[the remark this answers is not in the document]
kkhatua: You could provide this as an env var (full path: `DRILL_PID_DIR+"/"+GRACEFUL_SIGFILE`) within the `drill-config.sh` (ref: https://github.com/apache/drill/blob/master/distribution/src/resources/drill-config.sh#L336 )

distribution/src/resources/drillbit.sh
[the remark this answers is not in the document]
dvjyothsna: Changed the name. Should I make it hidden or not?

distribution/src/resources/drillbit.sh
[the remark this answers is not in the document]
dvjyothsna: Removed this code and drillbit.sh start will delete the sig file.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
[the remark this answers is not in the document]
dvjyothsna: Added the file name(GRACEFUL_SIGFILE) in drill-config.sh

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -285,6 +286,10 @@ public synchronized void close() {
...
+      if (f.exists()) {
+        f.delete();
```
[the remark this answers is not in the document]
dvjyothsna: Deletion of the file is done in drillbit.sh start

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
dvjyothsna: @sohami I have made the changes based on your comments. Please let me know if you have any comments.

exec/java-exec/src/main/java/org/apache/drill/exec/work/WorkManager.java
dvjyothsna: Changed it.

distribution/src/resources/drill-config.sh
```diff
@@ -334,6 +334,7 @@ fi
 # provided in drill-env.sh.
 
 export DRILL_PID_DIR=${DRILL_PID_DIR:-$DRILL_HOME}
+export GRACEFUL_FILE_SUFFIX=GRACEFUL_SIGFILE
```
kkhatua: Where is the value for `GRACEFUL_SIGFILE` defined? Provide a default if not set
```
export FOO=${FOO:-"value"}
```

dvjyothsna: I thought you asked me to change the name of the file graceful to GRACEFUL_SIGFILE. Let me change it.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -335,6 +351,50 @@ private void javaPropertiesToSystemOptions() {
...
+    @Override
+    public void run () {
+      try {
+        pollShutdown(drillbit);
+      } catch (Exception e) {
+        throw new RuntimeException("Caught exception while polling for shutdown\n" + stackTrace, e);
+      }
+    }
+
+    private void pollShutdown(Drillbit drillbit) throws IOException, InterruptedException {
```
sohami: please rename all the variables to be something else rather than using the type for name. Like 
`path-->drillPidDirPath`
`file-->gracefulFileName`

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -335,6 +351,50 @@ private void javaPropertiesToSystemOptions() {
...
+        path.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY, StandardWatchEventKinds.ENTRY_CREATE);
+        while (!triggered_shutdown) {
+          final WatchKey wk = watchService.take();
+          for (WatchEvent<?> event : wk.pollEvents()) {
+            final Path changed = (Path) event.context();
+            if (changed.endsWith(file)) {
+              drillbit.interruptPollShutdown = false;
+              triggered_shutdown = true;
+              drillbit.close();
+              wk.cancel();
```
sohami: put `wk.cancel()` in finally block to handle case when `drillbit.close()` throws exception

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -335,6 +351,50 @@ private void javaPropertiesToSystemOptions() {
     }
   }
 
+
+  // Polls for graceful file to check if graceful shutdown is triggered from the script.
+  private static class PollShutdownThread extends Thread {
```
sohami: Rename to `GracefulShutdownThread`

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -335,6 +351,50 @@ private void javaPropertiesToSystemOptions() {
...
+    private void pollShutdown(Drillbit drillbit) throws IOException, InterruptedException {
+      final Path path = FileSystems.getDefault().getPath(System.getenv("DRILL_PID_DIR"));
+      final String file = System.getenv("GRACEFUL_SIGFILE");
+      boolean triggered_shutdown = false;
+      try (final WatchService watchService = FileSystems.getDefault().newWatchService()) {
+        path.register(watchService, StandardWatchEventKinds.ENTRY_MODIFY, StandardWatchEventKinds.ENTRY_CREATE);
+        while (!triggered_shutdown) {
+          final WatchKey wk = watchService.take();
+          for (WatchEvent<?> event : wk.pollEvents()) {
+            final Path changed = (Path) event.context();
```
sohami: `event.context()` can be null (See [here](https://docs.oracle.com/javase/7/docs/api/java/nio/file/WatchEvent.html#context())). please add a check for it.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -291,6 +302,11 @@ public synchronized void close() {
 
     logger.info("Shutdown completed ({} ms).", w.elapsed(TimeUnit.MILLISECONDS) );
     stateManager.setState(DrillbitState.SHUTDOWN);
+    // Interrupt the polling for shutdown since shutdown is triggered from WebUI.
```
sohami: `Interrupt GracefulShutdownThread since Drillbit close is not called from it`

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -335,6 +351,50 @@ private void javaPropertiesToSystemOptions() {
...
+      this.stackTrace = stackTrace;
+    }
+
+    @Override
+    public void run () {
+      try {
+        pollShutdown(drillbit);
+      } catch (Exception e) {
+        throw new RuntimeException("Caught exception while polling for shutdown\n" + stackTrace, e);
+      }
```
sohami: I think you have to catch `InterruptedException` and consume it after logging rather than throwing `RuntimeException` for it. Since when `Drillbit.close` is called from another path it will interrupt this thread and we should not throw `RuntimeException` in that case.

distribution/src/resources/drillbit.sh
```diff
@@ -192,6 +201,10 @@ start_bit ( )
   echo $procId > $pidFile # Yeah, $pidFile is a file, $procId is the pid...
   echo $! > $pidFile
   sleep 1
+  #remove any previous uncleaned graceful file
+  if [ -f "$GRACEFUL_FILE" ]; then
+    rm $GRACEFUL_FILE
```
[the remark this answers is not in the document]
dvjyothsna: The error thrown by rm is currently logged to console.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
dvjyothsna: Changed it

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
dvjyothsna: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
dvjyothsna: Added a null check for event.context().

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
dvjyothsna: Changed it. Initially, it was GracefulShutdownThread but I thought people might be confused since it only gracefully shuts down from the script.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
dvjyothsna: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
dvjyothsna: Done.

dvjyothsna:
Thank you Sorabh! Squashed all the commits.

PR #1539, opened 2018-11-13
DRILL-6847: Add Query Metadata to RESTful Interface

cgivre:
The Drill RESTful interface does not return the structure of the query results.   This makes integrating Drill with other BI tools difficult because they do not know what kind of data to expect.  
This PR adds a new section to the results called Metadata which contains a list of the minor types of all the columns returned.

The query below will now return the following in the RESTful interface:

```
SELECT CAST( employee_id AS INT) AS employee_id,
full_name,
first_name, 
last_name, 
CAST( position_id AS BIGINT) AS position_id, 
position_title 
FROM cp.`employee.json` LIMIT 2
```
```
{
  "queryId": "2414bf3f-b4f4-d4df-825f-73dfb3a56681",
  "columns": [
    "employee_id",
    "full_name",
    "first_name",
    "last_name",
    "position_id",
    "position_title"
  ],
  "metadata": [
    "INT",
    "VARCHAR",
    "VARCHAR",
    "VARCHAR",
    "BIGINT",
    "VARCHAR"
  ],
  "rows": [
    {
      "full_name": "Sheri Nowmer",
      "employee_id": "1",
      "last_name": "Nowmer",
      "position_title": "President",
      "first_name": "Sheri",
      "position_id": "1"
    },
    {
      "full_name": "Derrick Whelply",
      "employee_id": "2",
      "last_name": "Whelply",
      "position_title": "VP Country Manager",
      "first_name": "Derrick",
      "position_id": "2"
    }
  ]
}
```

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
@@ -115,7 +115,7 @@ public class QueryWrapper {
     }
 
     // Return the QueryResult.
-    return new QueryResult(queryId, webUserConnection.columns, webUserConnection.results);
+    return new QueryResult(queryId, webUserConnection, webUserConnection.results);
   }
 
   //Detect possible excess heap
@@ -127,12 +127,15 @@ public class QueryWrapper {
     private final String queryId;
     public final Collection<String> columns;
     public final List<Map<String, String>> rows;
+    public final List<String> metadata;
 
-    public QueryResult(QueryId queryId, Collection<String> columns, List<Map<String, String>> rows) {
-      this.queryId = QueryIdHelper.getQueryId(queryId);
-      this.columns = columns;
-      this.rows = rows;
-    }
+    //DRILL-6847:  Modified the constructor so that the method has access to all the properties in webUserConnection
+    public QueryResult(QueryId queryId, WebUserConnection webUserConnection, List<Map<String, String>> rows) {
+        this.queryId = QueryIdHelper.getQueryId(queryId);
+        this.columns = webUserConnection.columns;
+        this.metadata = webUserConnection.metadata;
+        this.rows = rows;
+      }
 
     public String getQueryId() {
       return queryId;
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
@@ -37,10 +37,12 @@ import org.apache.drill.exec.rpc.ConnectionThrottle;
 import org.apache.drill.exec.rpc.RpcOutcomeListener;
 import org.apache.drill.exec.rpc.user.UserSession;
 import org.apache.drill.exec.vector.ValueVector.Accessor;
+import org.apache.drill.exec.record.MaterializedField;
 
 import java.net.SocketAddress;
 import java.util.List;
 import java.util.Map;
+import java.util.ArrayList;
 import java.util.Set;
 
 /**
@@ -63,6 +65,8 @@ public class WebUserConnection extends AbstractDisposableUserClientConnection im
   public final List<Map<String, String>> results = Lists.newArrayList();
 
   public final Set<String> columns = Sets.newLinkedHashSet();
+
+  public final List<String> metadata = new ArrayList<>();
 
   WebUserConnection(WebSessionResources webSessionResources) {
     this.webSessionResources = webSessionResources;
@@ -106,7 +110,29 @@ public class WebUserConnection extends AbstractDisposableUserClientConnection im
         // TODO:  Clean:  DRILL-2933:  That load(...) no longer throws
         // SchemaChangeException, so check/clean catch clause below.
         for (int i = 0; i < loader.getSchema().getFieldCount(); ++i) {
-          columns.add(loader.getSchema().getColumn(i).getName());
+          //DRILL-6847:  This section adds query metadata to the REST results
+          MaterializedField col = loader.getSchema().getColumn(i);
+          columns.add(col.getName());
+          StringBuilder dataType = new StringBuilder(col.getType().getMinorType().name());
+
+          //For DECIMAL type
+          if (col.getType().hasPrecision()) {
+            dataType.append("(");
+            dataType.append(col.getType().getPrecision());
+
+            if (col.getType().hasScale()) {
+              dataType.append(", ");
+              dataType.append(col.getType().getScale());
+            }
+
+            dataType.append(")");
+          } else if (col.getType().hasWidth()) {
+            //Case for VARCHAR columns with specified width
+            dataType.append("(");
+            dataType.append(col.getType().getWidth());
+            dataType.append(")");
+          }
+          metadata.add(dataType.toString());
         }
         ValueVectorElementFormatter formatter = new ValueVectorElementFormatter(webSessionResources.getSession().getOptions());
         for (int i = 0; i < rows; ++i) {


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -64,6 +66,8 @@
 
   public final Set<String> columns = Sets.newLinkedHashSet();
 
+  public final ArrayList<String> metadata = Lists.newArrayList();
```
vdiravka: Guava suggests to use the ArrayList constructor directly, taking advantage of the new "diamond" syntax:
https://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Lists.html
Thus `public final List<String> metadata = new ArrayList<>();` will be better.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -115,7 +116,8 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
     }
 
     // Return the QueryResult.
-    return new QueryResult(queryId, webUserConnection.columns, webUserConnection.results);
+    //return new QueryResult(queryId, webUserConnection.columns, webUserConnection.results);
```
vdiravka: no need in leaving this string as a comment

cgivre: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
cgivre: Done

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -106,7 +110,10 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
         // TODO:  Clean:  DRILL-2933:  That load(...) no longer throws
         // SchemaChangeException, so check/clean catch clause below.
         for (int i = 0; i < loader.getSchema().getFieldCount(); ++i) {
-          columns.add(loader.getSchema().getColumn(i).getName());
+
+          MaterializedField col = loader.getSchema().getColumn(i);
+          columns.add(col.getName());
+          metadata.add(col.getType().getMinorType().name());
```
arina-ielchiieva: Some types can have precision and scale, does this information needed?

cgivre: Hi @arina-ielchiieva ,
The use case I had in mind was integrating Drill with SQLPad and Apache Superset.  In these instances basically, the UI needed to know if a field was numeric, temporal of any sort, or text so that it could render visualizations properly.  

I'm sure there are other use cases out there, but I know that for me at least, this was a major blocker in getting Drill to work with the various BI tools.  The JDBC interface provided this information, but the RESTful interface did not, so I had to resort to hackery. 

So to answer your question, it might be useful for other use cases to provide precision and scale, but for the one I had in mind, that would not be helpful.

arina-ielchiieva: I see but though it is not your use case, I think we should consider shipping not only String type but as well information about precision and scale for those who might need it.

cgivre: How would you recommend designing that?  I was trying to keep this PR relatively simple, and backwards compatible, but one option might be to make the metadata a little duplicative so something like:

```
"metadata": [{
   "name": "price",
   "type": "FLOAT4"
   "precision":
   "scale"
   },{
     "name": "customer",
     "type": "VARCHAR"
  ...
]
```
Do you know off hand where the precision/scale or any other attributes of the columns can be accessed?

arina-ielchiieva: 1. Duplicating column name does not make sense.
2. You may not output precision and scale is they are absent, depends in which Object you plan to deserialize this information.
3. Look at major type, for example.

kkhatua:
@cgivre  I didn't build and try this out, but I'm curious on how we manage for `select * from` queries. Especially with schema changes... like fields being added or dropped between rows.
Also, I agree with @arina-ielchiieva 's point of not repeating the field names (unless schema change requires it). But I'm not sure how badly we break backward compatibility (perhaps carry a `version` in the REST response?).

cgivre:
HI @kkhatua 
A few things.  First re: breaking backward compatibility.  Since the RESTful interface returns JSON KV pairs, and since all this is doing is introducing an additional key (metadata) which contains the column types, I don't think it will break anything since it isn't doing anything to the existing keys.  No real way to know unless a lot of people complain, but the tools which I use that use the RESTful interface are not affected.

Secondly, regarding star queries, I did try a few on my local machine and you get a dump of all the column types. IE if your data contains 4 columns and you do a `SELECT *` you get:

```
  "metadata": [
    "INT",
    "VARCHAR",
    "VARCHAR",
    "VARCHAR"
  ],
```
Since this code gets called when results are returned, I don't think schema inconsistencies matter.  IE if the query fails due to schema inconsistencies, you won't get metadata, and if it succeeds you get the results. 

@arina-ielchiieva Would you like to see additional arrays with precision and scale with null values if it is not needed.  IE 
```
"scale": [
4,
null,
null,
null
]
```

arina-ielchiieva:
@cgivre you mean inside of metadata?
```
  "metadata": [
    type: [
       "DECIMAL"
     ],
    precision: [
       5
    ],
   scale: [
     2
   ]
 ]
```

cgivre:
@arina-ielchiieva 
I like that.  If the column type doesn't have precision or scale, would it look like this?

```
"metadata": [
    type: [
       "DECIMAL",
       "VARCHAR"
     ],
    precision: [
       5,
       null
    ],
   scale: [
     2,
     null
   ]
 ]
```

arina-ielchiieva:
Looks ok, try to implement and we'll see.

cgivre:
@arina-ielchiieva 
I added the precision and scale.   
The query:
```
SELECT CAST( employee_id AS DECIMAL) AS employee_id,
full_name,
first_name, 
last_name, 
CAST( position_id AS BIGINT) AS position_id, position_title 
FROM cp.`employee.json` LIMIT 2
```
Will result in the following metadata in the RESTful interface:
```
{
  "queryId": "240f59ad-9846-bee1-0668-575c00add18a",
  "columns": [
    "employee_id",
    "full_name",
    "first_name",
    "last_name",
    "position_id",
    "position_title"
  ],
  "metadata": {
    "precision": [
      38,
      null,
      null,
      null,
      null,
      null
    ],
    "scale": [
      0,
      null,
      null,
      null,
      null,
      null
    ],
    "type": [
      "VARDECIMAL",
      "VARCHAR",
      "VARCHAR",
      "VARCHAR",
      "BIGINT",
      "VARCHAR"
    ]
  },
  "rows": [
    {
      "full_name": "Sheri Nowmer",
      "employee_id": "1",
      "last_name": "Nowmer",
      "position_title": "President",
      "first_name": "Sheri",
      "position_id": "1"
    },
    {
      "full_name": "Derrick Whelply",
      "employee_id": "2",
      "last_name": "Whelply",
      "position_title": "VP Country Manager",
      "first_name": "Derrick",
      "position_id": "2"
    }
  ]
}
```

arina-ielchiieva:
@cgivre GitHub diff show lots of changes which clearly unrelated to your changes. Please revert all unrelated formatting changes that might have caused it since it's hard to review what has been done.

cgivre:
Hi @arina-ielchiieva 
Can you just hide the non-whitespace changes in GitHub for the review?  I think when I ran the code-formatter in IntelliJ it reformatted those files. 

If not, I'll revert and resubmit.

arina-ielchiieva:
@cgivre submitting core reformatting without significant reason is not encouraged. Please use Drill dev guidelines to ensure you use the formatting accepted in Drill project. The same with star imports, this also can be configured in Intellij.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -126,14 +128,23 @@ private float getHeapUsage() {
   public static class QueryResult {
     private final String queryId;
     public final Collection<String> columns;
+    public final HashMap<String, ArrayList> metadata;
```
arina-ielchiieva: Use more generic collections: `public final Map<String, List> metadata;`.
Plus List of what type? Maybe `List<String>`?

Please update here and in the code below.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -126,14 +128,23 @@ private float getHeapUsage() {
...
     public final List<Map<String, String>> rows;
 
     public QueryResult(QueryId queryId, Collection<String> columns, List<Map<String, String>> rows) {
       this.queryId = QueryIdHelper.getQueryId(queryId);
       this.columns = columns;
+      this.metadata = new HashMap<String,ArrayList>();
       this.rows = rows;
     }
 
+    public QueryResult(QueryId queryId, WebUserConnection webUserConnection, List<Map<String, String>> rows) {
```
arina-ielchiieva: Why we have two constructors now?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -106,7 +116,25 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
         // TODO:  Clean:  DRILL-2933:  That load(...) no longer throws
         // SchemaChangeException, so check/clean catch clause below.
         for (int i = 0; i < loader.getSchema().getFieldCount(); ++i) {
-          columns.add(loader.getSchema().getColumn(i).getName());
+            MaterializedField col = loader.getSchema().getColumn(i);
```
arina-ielchiieva: What about varchar? This type has length...

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
cgivre: I left the original constructor because I wasn't sure if it was used elsewhere and I didn't want to break anything.  The new one passes the `webUserConnection` directly to the `QueryResult` and enables you to access all the newly added properties.

I can remove the old one if you'd prefer.

arina-ielchiieva: Please make sure it's not and remove if needed.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -106,7 +116,25 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
...
+            columns.add(col.getName());
+
+            metadata.get("type").add(col.getType().getMinorType().name());
+
+            //minorType.add(col.getType().getMinorType().name());
```
arina-ielchiieva: Please clean up comments.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
cgivre: This change was done for code cleanup.   Since I was referring to the `column` several times, it was easier to create a `col` variable.

cgivre:
@arina-ielchiieva 
Ok, metadata is refomatted.

The query:

```
SELECT CAST( employee_id AS INT) AS employee_id, 
CAST( employee_id AS DECIMAL(8,4)) AS employee_id1, 
CAST(full_name AS VARCHAR(10)) AS full_name,
first_name, 
last_name, 
CAST( position_id AS BIGINT) AS position_id, 
position_title 
FROM cp.`employee.json` 
LIMIT 2
```

Results in the following metadata:
```
{
  "queryId": "240bc0a2-cb41-320d-17cf-072c912c83a8",
  "columns": [
    "employee_id",
    "employee_id1",
    "full_name",
    "first_name",
    "last_name",
    "position_id",
    "position_title"
  ],
  "rows": [
    {
      "full_name": "Sheri Nowm",
      "employee_id1": "1.0000",
      "employee_id": "1",
      "last_name": "Nowmer",
      "position_title": "President",
      "first_name": "Sheri",
      "position_id": "1"
    },
    {
      "full_name": "Derrick Wh",
      "employee_id1": "2.0000",
      "employee_id": "2",
      "last_name": "Whelply",
      "position_title": "VP Country Manager",
      "first_name": "Derrick",
      "position_id": "2"
    }
  ],
  "metadata": [
    "INT",
    "VARDECIMAL(8, 4)",
    "VARCHAR(10)",
    "VARCHAR",
    "VARCHAR",
    "BIGINT",
    "VARCHAR"
  ]
}
```

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -106,7 +110,30 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
...
+            columns.add(col.getName());
+            StringBuilder dataType = new StringBuilder(col.getType().getMinorType().name());
+
+            //For DECIMAL type
+            if (col.getType().hasPrecision()) {
+                dataType.append("(");
+                dataType.append(col.getType().getPrecision());
+
+                if (col.getType().hasScale()) {
+                  dataType.append((", "));
```
arina-ielchiieva: Please remove extra parenthesis and squash the commits.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -106,7 +110,30 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
         // TODO:  Clean:  DRILL-2933:  That load(...) no longer throws
         // SchemaChangeException, so check/clean catch clause below.
         for (int i = 0; i < loader.getSchema().getFieldCount(); ++i) {
-          columns.add(loader.getSchema().getColumn(i).getName());
+            //DRILL-6847:  This section adds query metadata to the REST results
```
arina-ielchiieva: @cgivre sorry for being picky but it seems indent in for loop and if clauses does not qualify to the accepted one in Drill project.

cgivre: No worries... Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
```diff
@@ -115,7 +115,7 @@ public QueryResult run(final WorkManager workManager, final WebUserConnection we
     }
 
     // Return the QueryResult.
-    return new QueryResult(queryId, webUserConnection.columns, webUserConnection.results);
+      return new QueryResult(queryId, webUserConnection, webUserConnection.results);
```
arina-ielchiieva: Please revert back initial indent.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -106,7 +110,29 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
...
+            dataType.append("(");
+            dataType.append(col.getType().getPrecision());
+
+            if (col.getType().hasScale()) {
+              dataType.append(", ");
+              dataType.append(col.getType().getScale());
+            }
+
+            dataType.append(")");
+          } else if( col.getType().hasWidth()) {
```
arina-ielchiieva: `else if( col.getType().hasWidth())` -> `else if (col.getType().hasWidth())`

cgivre: Fixed.  Sorry about all the formatting errors.  Normally, I run the formatter before submitting but that screwed up the rest of the file's formatting.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryWrapper.java
cgivre: Fixed

arina-ielchiieva:
+1

PR #1542, opened 2018-11-15
DRILL-6850: JDBC integration tests failures

vvysotskyi:
In the first commit:
- Fixed RDBMS integration tests (expected decimal output and testCrossSourceMultiFragmentJoin)
- Updated libraries versions
- Resolved NPE for empty result

In the second commit:
- Made changes to force setting DRILL_LOGICAL Convention for DrillRelFactories and DrillFilterRel
- Fixed workspace case insensitivity for JDBC storage plugin

In the third commit:
- Allowed configuring table names case sensitivity for JDBC storage plugin

For problem description please see [DRILL-6850](https://issues.apache.org/jira/browse/DRILL-6850).

### DIFF
--- a/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcStorageConfig.java
+++ b/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcStorageConfig.java
@@ -34,18 +34,21 @@ public class JdbcStorageConfig extends StoragePluginConfig {
   private final String url;
   private final String username;
   private final String password;
+  private final boolean caseInsensitiveTableNames;
 
   @JsonCreator
   public JdbcStorageConfig(
       @JsonProperty("driver") String driver,
       @JsonProperty("url") String url,
       @JsonProperty("username") String username,
-      @JsonProperty("password") String password) {
+      @JsonProperty("password") String password,
+      @JsonProperty("caseInsensitiveTableNames") boolean caseInsensitiveTableNames) {
     super();
     this.driver = driver;
     this.url = url;
     this.username = username;
     this.password = password;
+    this.caseInsensitiveTableNames = caseInsensitiveTableNames;
   }
 
   public String getDriver() {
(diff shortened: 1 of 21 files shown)


contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcStorageConfig.java
```diff
@@ -64,6 +67,10 @@ public String getPassword() {
     return password;
   }
 
+  public boolean areTableNamesCaseInsensitive() {
```
arina-ielchiieva: Please check ser / de. Maybe you need to specify `@JsonProperty('caseInsensitiveTableNames')` over getter?

contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcStoragePlugin.java
```diff
@@ -291,27 +292,41 @@ public Table getTable(String name) {
       if (table != null) {
         return table;
       }
-      return inner.getTable(name.toUpperCase());
+      if (!areTableNamesCaseSensitive()) {
```
arina-ielchiieva: Do we know which db we are using in the plugin? Maybe we can do table search via switch by db name rather than via if's with comments above?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/FindHardDistributionScans.java
```diff
@@ -52,7 +52,12 @@ public RelNode visit(TableScan scan) {
     DrillTable unwrap;
     unwrap = scan.getTable().unwrap(DrillTable.class);
     if (unwrap == null) {
-      unwrap = scan.getTable().unwrap(DrillTranslatableTable.class).getDrillTable();
+      DrillTranslatableTable drillTranslatableTable = scan.getTable().unwrap(DrillTranslatableTable.class);
+      if (drillTranslatableTable == null) {
+        contains = true; // it rejects single mode.
```
arina-ielchiieva: Please add explanation why...

arina-ielchiieva:
+1

PR #1543, opened 2018-11-15
 DRILL-6668: In Web UI, highlight options that are not default values 

kkhatua:
This commit introduces a new button on the options page that allows a user to reset an option to its system default value. 

To simplify things, a tooltip is shown when the mouse hovers over the button. If the option value is already default, the button is disabled.
image

In addition, a second commit refactor Update button functionality to use same code path and reuses what we already are using to set the default (using AJAX) and auto-refreshing

### DIFF
--- a/exec/java-exec/src/main/resources/rest/options.ftl
+++ b/exec/java-exec/src/main/resources/rest/options.ftl
@@ -90,9 +90,9 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
                   </#if>
                     <div class="input-group-btn">
                       <button class="btn btn-default" type="submit">Update</button>
-                      <button class="btn btn-default" onClick="resetToDefault('${option.getName()}','${option.getDefaultValueAsString()}', '${option.getKind()}')" type="button"
-                              <#if option.getDefaultValueAsString() == option.getValueAsString()>disabled="true" style="pointer-events:none" <#else>
-                      title="Reset to ${option.getDefaultValueAsString()}"</#if>>Reset</button>
+                      <button class="btn btn-default" onClick="resetToDefault('${option.getName()}','${option.getDefaultValue()}', '${option.getKind()}')" type="button"
+                              <#if option.getDefaultValue() == option.getValueAsString()>disabled="true" style="pointer-events:none" <#else>
+                      title="Reset to ${option.getDefaultValue()}"</#if>>Reset</button>
                     </div>
                   </div>
                 </div>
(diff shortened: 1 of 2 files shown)


kkhatua:
@paul-rogers since you filed the JIRA, can you review this? Added other reviewers since I couldn't add you.

exec/java-exec/src/main/resources/rest/options.ftl
```diff
@@ -84,7 +104,9 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
                     <input type="text" class="form-control" placeholder="${option.getValueAsString()}" name="value" value="${option.getValueAsString()}">
                   </#if>
                     <div class="input-group-btn">
-                      <button class="btn btn-default" type="submit">Update</button>
+                      <button class="btn btn-default" type="button" onclick="alterSysOptionUsingId('${option.getName()}')">Update</button>
+                      <button class="btn btn-default" type="button" onclick="alterSysOption('${option.getName()}','${option.getDefaultValueAsString()}', '${option.getKind()}')" <#if option.getDefaultValueAsString() == option.getValueAsString()>disabled="true" style="pointer-events:none" <#else>
+                      title="Reset to ${option.getDefaultValueAsString()}"</#if>>Default</button>
```
arina-ielchiieva: Maybe better rename button to Reset?

exec/java-exec/src/main/resources/rest/options.ftl
```diff
@@ -21,6 +21,26 @@
...
+    <script>
+        //Alter System Values
+        function alterSysOption(optionName, optionValue, optionKind) {
+            $.post("/option/"+optionName, {kind: optionKind, name: optionName, value: optionValue}, function () {
+                location.reload(true);
+            });
+        }
+
+        //Read Values and apply
+        function alterSysOptionUsingId(optionRawName) {
```
arina-ielchiieva: Looks like before we just had POST method on the Update button and did not have to do replacements. I would prefer if e stick to previous approach or define one without parsing and replacements.

exec/java-exec/src/main/resources/rest/options.ftl
kkhatua: Will do.

exec/java-exec/src/main/resources/rest/options.ftl
kkhatua: It does make the changes a little more complex, but the primary motivation was that the same code path is being exercised for making any changes to the options. It's also why I kept it as 2 commits. We can revert it if we don't want a common execution.

arina-ielchiieva: Looks like common execution brings more complexity than should, please revert.

kkhatua: Ok. Will do that and run the tests.

kkhatua:
@arina-ielchiieva I've reverted the commit and changed the button label.
However, for some strange reason, the `StatusResourcesTest` kept failing during `LocalFunctionRegistry` initialization.  I had to make some strange changes to the `OptionWrapper` class, and that resolved the issue, and all unit tests passed.

arina-ielchiieva:
Overall, looks good.

arina-ielchiieva:
+1

PR #1549, opened 2018-11-21
DRILL-6834: Introduce option to disable result set on CTAS, create vi…

KazydubB:
…ew and drop table/view etc. for JDBC connection

- Added session-scoped option `drill.exec.fetch_resultset_for_ddl` to control whether update count or result set should be returned for JDBC connection session. Eligible queries are CTAS, CREATE VIEW, CREATE FUNCTION, DROP TABLE, DROP VIEW, DROP FUNCTION, USE schema, SET option, REFRESH METADATA TABLE. By default the option is set to `true` which ensures that result set is returned;
- Updated Drill JDBC: `DrillCursor` and `DrillStatement` to achieve desired behaviour;

### DIFF
--- a/exec/jdbc/src/test/java/org/apache/drill/jdbc/test/TestJdbcQuery.java
+++ b/exec/jdbc/src/test/java/org/apache/drill/jdbc/test/TestJdbcQuery.java
@@ -18,6 +18,7 @@
 package org.apache.drill.jdbc.test;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 import java.nio.file.Paths;
 import java.sql.Connection;
@@ -26,6 +27,7 @@ import java.sql.Statement;
 import java.sql.Types;
 
 import org.apache.drill.categories.JdbcTest;
+import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.jdbc.JdbcTestBase;
 import org.junit.BeforeClass;
 import org.junit.Ignore;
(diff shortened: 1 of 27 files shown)


exec/jdbc/src/test/java/org/apache/drill/jdbc/test/TestJdbcQuery.java
```diff
@@ -403,4 +407,147 @@ public void testConvertFromInEmptyInputSql() throws Exception {
...
+  @Test
+  public void testResultSetIsNotReturnedSet() throws Exception {
+    Statement s = null;
+    try (Connection conn = connect()) {
+      s = conn.createStatement();
+      boolean hasResult = s.execute(String.format("alter session set `%s` = false", ExecConstants.RETURN_RESULT_SET_FOR_DDL));
+
+      assertFalse("SET query should not return result set", hasResult);
+      assertNull("No result", s.getResultSet());
+      assertNotEquals("Update count should be >= 0", -1, s.getUpdateCount());
```
vvysotskyi: Not sure that we should also verify `updateCount` value. It is required to check that there is no result set.

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/TestJdbcQuery.java
```diff
@@ -403,4 +407,147 @@ public void testConvertFromInEmptyInputSql() throws Exception {
...
+  @Test
+  public void testResultSetIsNotReturnedCTAS() throws Exception {
+    final String tableName = "dfs.tmp.`ctas`";
+
+    Statement s = null;
+    try (Connection conn = connect()) {
+      s = conn.createStatement();
+      s.execute(String.format("alter session set `%s` = false", ExecConstants.RETURN_RESULT_SET_FOR_DDL));
+
+      s.execute(String.format("CREATE TABLE %s AS SELECT * FROM cp.`employee.json`", tableName));
```
vvysotskyi: Please add `drop table if exists` to the `finally` block.

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/TestJdbcQuery.java
```diff
@@ -403,4 +407,147 @@ public void testConvertFromInEmptyInputSql() throws Exception {
...
+  @Test
+  public void testResultSetIsNotReturnedCreateView() throws Exception {
+    final String tableName = "dfs.tmp.`cv`";
+
+    Statement s = null;
+    try (Connection conn = connect()) {
+      s = conn.createStatement();
+      s.execute(String.format("alter session set `%s` = false", ExecConstants.RETURN_RESULT_SET_FOR_DDL));
+
+      s.execute(String.format("CREATE VIEW %s AS SELECT * FROM cp.`employee.json`", tableName));
```
vvysotskyi: Please add `drop view if exists`  to the `finally` block.

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/TestJdbcQuery.java
```diff
@@ -403,4 +407,147 @@ public void testConvertFromInEmptyInputSql() throws Exception {
         .sql("SELECT CONVERT_FROM(columns[1], 'JSON') as col1 from cp.`empty.csv`")
         .returns("");
   }
+
+  @Test
+  public void testResultSetIsNotReturnedSet() throws Exception {
+    Statement s = null;
+    try (Connection conn = connect()) {
```
vvysotskyi: Please rewrite tests, since looks like a connection is closed before closing the statement [1].

[1] https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3.2

KazydubB:
@vvysotskyi, thanks for your revire. I've updated the PR

PR #1551, opened 2018-11-23
DRILL-6867: WebUI Query editor cursor position

oleg-zinovev:
Few additional font families added for `query-editor-format` element


arina-ielchiieva:
@kkhatua could you please review?

oleg-zinovev:
Unfortunately, I can check this only on wednesday or Thursday. Is this ok?

PR #1553, opened 2018-11-23
DRILL-6868: Upgrade Janino compiler to 3.0.11

vvysotskyi:
- Removed workaround where removing adjacent ALOAD-POP instruction pairs, since it was fixed in new version of Janino compiler
- Removed ModifiedUnparser and used DeepCopier for modifying methods instead of modifying it with custom Unparser implementation


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/MethodGrabbingVisitor.java
```diff
@@ -64,9 +67,31 @@ public void traverseClassDeclaration(AbstractClassDeclaration classDeclaration)
     @Override
     public void traverseMethodDeclarator(MethodDeclarator methodDeclarator) {
       if (captureMethods) {
+        // Generate a "labeled statement".
```
arina-ielchiieva: Please add comments inline explaining each action.

vvysotskyi:
@arina-ielchiieva, I have added comments with explanations, could you please take a look again?

arina-ielchiieva:
Looks good, +1

PR #1557, opened 2018-11-28
DRILL-6863: Drop table is not working if path within workspace starts…

KazydubB:
… with "/"

- Made workspace to be honored when table/view name starts with "/" for DROP TABLE, DROP VIEW, CREATE VIEW and SELECT from view queries;
- Made "/{name}" and "{name}" to be equivalent names (the leading "/" is removed) when creating temporary tables so that `SELECT ... FROM "/{name}" ...` and `SELECT ... FROM "{name}" ...` produce the same results and behave as regular tables in the context.


exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTTAS.java
```diff
@@ -446,6 +446,23 @@ public void testJoinTemporaryWithPersistentTable() throws Exception {
         .go();
   }
 
+  @Test
+  public void testCreateTemporaryTableWithAndWithoutLeadingSlashAreTheSame() throws Exception{
+    String tablename = "tmp_table";
+
+    try {
+      test("CREATE TEMPORARY TABLE %s.%s AS SELECT * from cp.`region.json`", DFS_TMP_SCHEMA, tablename);
```
vdiravka: lead all identifiers to the same upper or below case.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTTAS.java
```diff
@@ -446,6 +446,23 @@ public void testJoinTemporaryWithPersistentTable() throws Exception {
         .go();
   }
 
+  @Test
+  public void testCreateTemporaryTableWithAndWithoutLeadingSlashAreTheSame() throws Exception{
```
vdiravka: `Create` can be omitted possibly

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java
```diff
@@ -751,4 +751,53 @@ public void selectFromViewCreatedOnCalcite1_4() throws Exception {
...
+          .baselineColumns("region_id")
+          .baselineValues(0L)
+          .go();
+    } finally {
+      test("DROP VIEW IF EXISTS `%s`.`%s`", DFS_TMP_SCHEMA, viewName);
+    }
+  }
+
+  @Test
+  public void testViewIsFoundWithinWorkspaceWhenNameStartsWithSlash() throws Exception {
```
vdiravka: I think `WithinWorkspace` can be ommited, since there is always some workspace used for tables. Even for default workspace it is still valid. Right?

exec/java-exec/src/test/java/org/apache/drill/TestDropTable.java
```diff
@@ -21,23 +21,28 @@
...
 @Category(SqlTest.class)
 public class TestDropTable extends PlanTestBase {
 
-  private static final String CREATE_SIMPLE_TABLE = "create table %s as select 1 from cp.`employee.json`";
-  private static final String CREATE_SIMPLE_VIEW = "create view %s as select 1 from cp.`employee.json`";
-  private static final String DROP_TABLE = "drop table %s";
-  private static final String DROP_TABLE_IF_EXISTS = "drop table if exists %s";
-  private static final String DROP_VIEW_IF_EXISTS = "drop view if exists %s";
-  private static final String BACK_TICK = "`";
+  private static final String CREATE_SIMPLE_TABLE = "create table `%s` as select 1 from cp.`employee.json`";
```
vdiravka: What is changed in these rows except the row with `BACK_TICK`?

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java
```diff
@@ -751,4 +751,53 @@ public void selectFromViewCreatedOnCalcite1_4() throws Exception {
...
+          .baselineValues(true,
+              String.format("View [%s] deleted successfully from schema [%s].", viewName, DFS_TMP_SCHEMA))
+          .go();
+    } finally {
+      test("DROP VIEW IF EXISTS `%s`.`%s`", DFS_TMP_SCHEMA, viewName);
+    }
+  }
+
+  @Test
+  public void testViewIsCreatedWithinWorkspace() throws Exception {
```
vdiravka: Do we need similar test cases for tables?

exec/java-exec/src/test/java/org/apache/drill/TestDropTable.java
```diff
@@ -220,4 +218,36 @@ public void testDropTableIfExistsWhileItIsAView() throws Exception {
       test(DROP_VIEW_IF_EXISTS, viewName);
     }
   }
+
+  @Test
+  public void testDropTableNameStartsWithSlash() throws Exception {
```
arina-ielchiieva: Please move test for CTTAS to the TestCTTAS class.

exec/java-exec/src/test/java/org/apache/drill/TestDropTable.java
```diff
@@ -220,4 +218,36 @@ public void testDropTableIfExistsWhileItIsAView() throws Exception {
...
+          .baselineColumns("ok", "summary")
+          .baselineValues(true, String.format("Table [%s] dropped", tableName))
+          .go();
+    } finally {
+      test(DROP_TABLE_IF_EXISTS, tableName);
+    }
+  }
+
+  @Test
+  public void testDropTemporaryTableNameStartsWithSlash() throws Exception {
```
arina-ielchiieva: What if user has created table with slash initially?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DropTableHandler.java
```diff
@@ -56,7 +57,7 @@ public DropTableHandler(SqlHandlerConfig config) {
   @Override
   public PhysicalPlan getPlan(SqlNode sqlNode) throws ValidationException, RelConversionException, IOException {
     SqlDropTable dropTableNode = ((SqlDropTable) sqlNode);
-    String originalTableName = dropTableNode.getName();
+    String originalTableName = FileSelection.removeLeadingSlash(dropTableNode.getName());
```
arina-ielchiieva: Maybe instead of removing slashes we should disallow creating such tables?

KazydubB: Actually, this may appear counter-intuitive for users especially when Drill supports `SELECT` with leading slash (see [DRILL-1130](https://issues.apache.org/jira/browse/DRILL-1130)). When a table is created and its name does start with '/' it will be created within workspace, i.e. will be rooted from the workspace.
You could think of it as if you performed `touch` (on Unix[-like] OS) to create a file: `touch /${filename}` and `touch ${filename}` should create a file in the same directory when working directory (workspace) is root ('/').

arina-ielchiieva: Sounds fair, thanks for the explanation.

exec/java-exec/src/test/java/org/apache/drill/TestDropTable.java
vdiravka: I have not noticed it :) Thank you!

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestViewSupport.java
vdiravka: It's fine

PR #1560, opened 2018-12-04
DRILL-6877: NPE when starting Drillbit

dvjyothsna:


dvjyothsna:
@sohami Please review.

vdiravka:
@dvjyothsna Please follow the usual PR and commits naming. It allows to add link for PR to the Jira ticket automatically.
> DRILL-6877: NPE when starting Drill on Windows

dvjyothsna:
@vdiravka Changed it.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -374,17 +374,22 @@ public void run () {
...
     private void pollShutdown(Drillbit drillbit) throws IOException, InterruptedException {
-      final Path drillPidDirPath = FileSystems.getDefault().getPath(System.getenv("DRILL_PID_DIR"));
-      final String gracefulFileName = System.getenv("GRACEFUL_SIGFILE");
+      final String drillHome = System.getenv("DRILL_HOME");
+      final String gracefulFile = System.getenv("GRACEFUL_SIGFILE");
+      if (drillHome == null || gracefulFile == null) {
+        logger.warn("Cannot access graceful file. Graceful shutdown from command line will not be supported.");
+        return;
+      }
+      final Path drillHomePath = FileSystems.getDefault().getPath(drillHome);
```
arina-ielchiieva: Does this guarantee that local file system will be used? For example, on the cluster default file system might be distributed one.

sohami: The defaultFileSystem is chosen either by `java.nio.file.spi.DefaultFileSystemProvider` system property or system-default one which is created [here](https://github.com/frohoff/jdk8u-jdk/blob/master/src/solaris/classes/sun/nio/fs/DefaultFileSystemProvider.java#L57) based on local filesystem
As far as DistributedFileSystem is concerned I don't see it provides any constructor with a provider which NIO Filesystem expects and hence will fail in that case. So the api should only be good for local filesystem.
Do we expect Drill binaries (not data) to be installed on Distributed FileSystem as well ? If yes then this will not work and might throw `RuntimeException`. May be we should catch and log warning in case of error on this thread instead of throwing RuntimeException.

arina-ielchiieva: Can we use the following: `Paths.get(drillHome).getFileSystem()`?

dvjyothsna:
@arina-ielchiieva , @sohami  Made changes and rebased. Please review it once again.

sohami:
+1 LGTM. Please wait for approval from @arina-ielchiieva

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -374,17 +374,22 @@ public void run () {
...
     private void pollShutdown(Drillbit drillbit) throws IOException, InterruptedException {
-      final Path drillPidDirPath = FileSystems.getDefault().getPath(System.getenv("DRILL_PID_DIR"));
-      final String gracefulFileName = System.getenv("GRACEFUL_SIGFILE");
+      final String drillHome = System.getenv("DRILL_HOME");
+      final String gracefulFile = System.getenv("GRACEFUL_SIGFILE");
+      if (drillHome == null || gracefulFile == null) {
+        logger.warn("Cannot access graceful file. Graceful shutdown from command line will not be supported.");
+        return;
+      }
+      final Path drillHomePath = Paths.get(drillHome);
```
arina-ielchiieva: 1. We might check if `drillHomePath` actually exists and reachable and issue a warning rather than fail if something is going wrong.
2. Please update PR and commit error description, it turned out NPE exists on other systems as well (just remove `on Windows part`).

dvjyothsna: @arina-ielchiieva 
1. Added try-catch block while getting the drillHome path. What do you mean by if it is not set correctly? Both script and the code access the same variable and create or look for the graceful file.
2. Updated the PR.
3. Added java doc.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -373,18 +374,33 @@ public void run () {
       }
     }
 
+    /*
+     * Poll for the graceful file, if the file is found cloase the drillbit. Incase if the DRILL_HOME path is not
+     * set, graceful shutdown will not be supported from the command line.
```
arina-ielchiieva: incase -> in case

vdiravka:
@dvjyothsna I have addressed minor comment and squashed the commits.

arina-ielchiieva:
@vdiravka thanks a lot!

PR #1563, opened 2018-12-06
DRILL-6883: Force reload of options after resetting parameter

kkhatua:
If an update is initiated, the webpage loaded changes. Attempting a reset tries to reload this page, but it does not show the updated values from the back end.
This patch forces the reload of updated values by redirecting to /options page

### DIFF
--- a/exec/java-exec/src/main/resources/rest/options.ftl
+++ b/exec/java-exec/src/main/resources/rest/options.ftl
@@ -23,7 +23,7 @@
     <script type="text/javascript" language="javascript" src="/static/js/dataTables.colVis-1.1.0.min.js"></script>
     <script>
         function resetToDefault(optionName, optionValue, optionKind) {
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/resources/rest/options.ftl
```diff
@@ -22,9 +22,29 @@
...
+        $.post("/option/"+optionName, {kind: optionKind, name: optionName, value: optionValue}, function () {
+            location.reload(true);
+        });
+    }
+
+    //Read Values and apply
+    function alterSysOptionUsingId(optionRawName) {
+        //Escaping '.' for id search
+        let optionName = optionRawName.replace(/\./gi, "\\.");
+        let optionKind = $("#"+optionName+" input[name='kind']").attr("value");
```
arina-ielchiieva: Please add comments with explanation above lines 36, 37, 39.

kkhatua: Done.

arina-ielchiieva:

PR #1565, opened 2018-12-07
DRILL-6874: Close input stream after AsyncPageReaderTask is completed

KazydubB:


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/columnreaders/AsyncPageReader.java
```diff
@@ -417,6 +417,9 @@ public Void call() throws IOException {
       if (totalValuesRead >= totalValuesCount) {
         try {
           queue.put(ReadStatus.EMPTY);
+          // Some InputStreams (like S3ObjectInputStream) should be closed
+          // as soon as possible to make the connection reusable.
+          parent.inputStream.close();
```
arina-ielchiieva: 1. Should we handle any exceptions?
2. Please rename commit message.

KazydubB:
@arina-ielchiieva, thank you for the review. I've made changes, please take a look.

arina-ielchiieva:
+1

PR #1566, opened 2018-12-07
DRILL-6887: Fix FunctionInitializerTest.init() failure

vdiravka:
It fixes `FunctionInitializerTest.init()` failure for jdk9 and above.

### DIFF
--- a/exec/java-exec/src/test/resources/drill-udf/pom.xml
+++ b/exec/java-exec/src/test/resources/drill-udf/pom.xml
@@ -28,6 +28,8 @@
   <version>1.0</version>
 
   <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
     <jar.finalName>${project.name}</jar.finalName>
     <custom.buildDirectory>${project.basedir}/target</custom.buildDirectory>
     <drill.version>1.13.0</drill.version>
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/test/resources/drill-udf/pom.xml
```diff
@@ -28,6 +28,8 @@
   <version>1.0</version>
 
   <properties>
+    <maven.compiler.source>1.8</maven.compiler.source>
+    <maven.compiler.target>1.8</maven.compiler.target>
```
arina-ielchiieva: @vdiravka  as far as I understand explicitly setting source and target version is not required. Maybe remove the setting?

vdiravka: After removing these properties the [default values](https://github.com/apache/maven-compiler-plugin/blob/875947802d28bcf8348eb634985c0512a790170a/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java#L169) for them will be 1.6:
> Since `maven-compiler-plugin` 3.8.0 the default value has changed from 1.5 to 1.6 

Even with explicit 1.6 values for `source` and `target` properties the tests pass, but these are not the right versions for current Drill version. 

According to [Maven docs](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html) we need to specify the properties for Java8.

PR #1568, opened 2018-12-08
DRILL-6878: Use DrillPushRowKeyJoinToScan rule on DrillJoin pattern to o account for DrillSemiJoin

gparai:
@amansinha100 can you please review the PR? Thanks!

With the fix, I see the following wrong results which would be fixed in a separate JIRA. I can disable them as well.

IndexPlanTest.testCastTimestampPlan
IndexPlanTest.testRowkeyJoinPushdown_1
IndexPlanTest.testRowkeyJoinPushdown_10
IndexPlanTest.testRowkeyJoinPushdown_13
IndexPlanTest.testRowkeyJoinPushdown_6
IndexPlanTest.testRowkeyJoinPushdown_7
IndexPlanTest.testRowkeyJoinPushdown_9

IndexPlanTest.testNoFilterGroupByHashIndex
IndexPlanTest.testNoFilterOrderByHashIndex

testRowkeyJoinPushdown_12 is disabled because we no longer get the desired pattern after semi-join kicks-in. The join order changes making it illegal to apply the transformation. Hence, we do not generate a rowkeyjoin so I disabled this positive testcase.

### DIFF
--- a/contrib/format-maprdb/src/test/java/com/mapr/drill/maprdb/tests/index/IndexPlanTest.java
+++ b/contrib/format-maprdb/src/test/java/com/mapr/drill/maprdb/tests/index/IndexPlanTest.java
@@ -1130,6 +1130,7 @@ public class IndexPlanTest extends BaseJsonTest {
     );
   }
 
+  @Ignore
   @Test
   public void testCastTimestampPlan() throws Exception {
     String query = "SELECT  t.id.ssn as ssn FROM hbase.`index_test_primary` as t " +
@@ -1698,8 +1699,7 @@ public class IndexPlanTest extends BaseJsonTest {
   public void testRowkeyJoinPushdown_1() throws Exception {
     // _id IN (select col ...)
     String query = "select t1.id.ssn as ssn from hbase.`index_test_primary` t1 where _id in (select t2._id " +
-        " from hbase.`index_test_primary` t2 where cast(t2.activity.irs.firstlogin as timestamp) = " +
-        " to_timestamp('2013-02-04 22:34:38.0', 'YYYY-MM-dd HH:mm:ss.S'))";
+        " from hbase.`index_test_primary` t2 where t2.address.city = 'pfrrs' and t2.address.state = 'pc')";
     try {
       test(incrRowKeyJoinConvSelThreshold + ";" + lowNonCoveringSelectivityThreshold + ";");
       PlanTestBase.testPlanMatchingPatterns(query, new String[] {"RowKeyJoin"}, new String[] {});
@@ -1707,6 +1707,7 @@ public class IndexPlanTest extends BaseJsonTest {
           .sqlQuery(query)
           .ordered()
           .baselineColumns("ssn").baselineValues("100007423")
+          .baselineColumns("ssn").baselineValues("100008861")
           .go();
     } finally {
       test(defaultRowKeyConvSelThreshold + ";" + defaultnonCoveringSelectivityThreshold + ";");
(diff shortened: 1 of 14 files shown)


amansinha100:
@gparai  I am not seeing the value of the PR if most of the RowKey join pushdown tests are going to be marked ignored.  I believe the separate issue you are referring to is related to index filter conditions on timestamp columns.   Can we modify the tests to use a different index column ?  At least it will be testing the rowkey join pushdown functionality.

gparai:
@amansinha100 I have modified the rowkeyjoin pushdown testcases. With these changes I now see only the following unit test failures which are unrelated to this feature.
IndexPlanTest.testNoFilterGroupByHashIndex
IndexPlanTest.testNoFilterOrderByHashIndex
Please take a look. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillJoinRel.java
```diff
@@ -45,6 +46,7 @@
  */
 public class DrillJoinRel extends DrillJoinRelBase implements DrillRel {
   public static final String EQUALITY_CONDITION = "==";
+  boolean isSemiJoin;
```
amansinha100: Since we already have DrillSemiJoinRel as a logical join,  it is better to not add this flag here.

gparai: @amansinha100 done. Please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillJoinRelBase.java
```diff
@@ -209,4 +210,36 @@ public static RelOptCost computeHashJoinCostWithRowCntKeySize(RelOptPlanner plan
     return costFactory.makeCost(buildRowCount + probeRowCount, cpuCost, 0, 0, memCost);
   }
 
+  @Override public boolean isValid(Litmus litmus, Context context) {
```
amansinha100: I think this row type validation could be moved to the `RowKeyJoinRel`.   Doing this in the base class of the joins does not seem right considering the Calcite is already doing the check for SemiJoin in its `isValid()` implementation,

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillJoinRel.java
```diff
@@ -55,53 +53,33 @@
...
     this.leftKeys = leftKeys;
     this.rightKeys = rightKeys;
-    this.isSemiJoin = isSemiJoin;
   }
 
   @Override
-  public DrillJoinRel copy(RelTraitSet traitSet, RexNode condition, RelNode left, RelNode right, JoinRelType joinType,
-      boolean semiJoinDone) {
-    return new DrillJoinRel(getCluster(), traitSet, left, right, condition, joinType, semiJoinDone);
+  public DrillJoinRel copy(RelTraitSet traitSet, RexNode condition, RelNode left, RelNode right, JoinRelType joinType, boolean semiJoinDone) {
```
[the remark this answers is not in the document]
gparai: This overrides the parent method. Technically it does not seem necessary to fix it as of now.

contrib/format-maprdb/src/test/java/com/mapr/drill/maprdb/tests/index/IndexPlanTest.java
```diff
@@ -1130,6 +1130,7 @@ public void testHashIndexNoRemovingSort() throws Exception {
     );
   }
 
+  @Ignore ("Enable after MAPRDB-1375 is fixed")
```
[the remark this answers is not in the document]
gparai: Done

gparai:
@amansinha100 I have made the changes you requested. Please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/RowKeyJoinRel.java
```diff
@@ -50,20 +66,72 @@ public RowKeyJoinRel(RelOptCluster cluster, RelTraitSet traits, RelNode left, Re
...
+
   public static RowKeyJoinRel convert(Join join, ConversionContext context) throws InvalidRelException {
     Pair<RelNode, RelNode> inputs = getJoinInputs(join, context);
     RexNode rexCondition = getJoinCondition(join, context);
     RowKeyJoinRel joinRel = new RowKeyJoinRel(context.getCluster(), context.getLogicalTraits(),
         inputs.left, inputs.right, rexCondition, join.getJoinType());
     return joinRel;
   }
+
+  @Override public boolean isValid(Litmus litmus, Context context) {
```
amansinha100: Pls add a comment here about why `isValid()` has to be overridden.

gparai: Done

gparai:
@amansinha100 thanks for the review and approval. I have addressed your remaining review comments.

gparai:
Thanks so much @vdiravka for merging it in. Yes, I plan to create a JIRA to track the disabled tests - they have some dependency on the index reference implementation so will need to be addressed later.

PR #1569, opened 2018-12-08
DRILL-6888: Move nested classes outside HashAggTemplate to allow for plain java compile option

Ben-Zvi:
Moved two nested classes (HashAggSpilledPartition and HashAggUpdater) from inside HashAggTemplate into independent files (using "refactor--move").

   The only additional change was to pass the allocator to the HashAggUpdater via the constructor.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggUpdater.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggUpdater.java
@@ -1,0 +1,44 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.aggregate;
+
+import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.physical.impl.common.SpilledState;
+
+public class HashAggUpdater implements SpilledState.Updater {
+  private final BufferAllocator allocator;
+  public HashAggUpdater(BufferAllocator allocator) { this.allocator = allocator; }
+
+  @Override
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.physical.impl.common.SpilledState;
+
+public class HashAggUpdater implements SpilledState.Updater {
+  private BufferAllocator allocator;
+  HashAggUpdater(BufferAllocator allocator) { this.allocator = allocator; }
+
+  @Override
+  public void cleanup() {
+    this.cleanup();
```
paul-rogers: Infinite recursion

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.impl.aggregate;
+
+import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.physical.impl.common.SpilledState;
+
+public class HashAggUpdater implements SpilledState.Updater {
+  private BufferAllocator allocator;
```
paul-rogers: private final

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggUpdater.java
Ben-Zvi: Good catch !  This is how the nested class code (in the master branch) was implemented. And then "refactor" was used to "move" this class out, so it kept this "this".
   Maybe the original intention was to call the containing class' _cleanup_, like  `HashAggTemplate.this.cleanup()`

So this method is probably never called .... 
Needs a fix anyway -- should be a no-op (no reason to call the HashAggTemplate's cleanup() ).

Ben-Zvi:
Note that the first nested class is static, and the other is not. After extracting the first static one, the code generation still failed. I'll recreate these and add the generated code to the Jira.

Ben-Zvi:
@paul-rogers please have a look at the Jira. Attached are two codegen files (from a simple hash agg query); one file from the master branch as is, and the other after extracting the nested static class HashAggSpilledPartition, but leaving the other non-static nested class HashAggUpdater.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/HashAggUpdater.java
Ben-Zvi: Fixed - removed this "recursion".

Ben-Zvi:
@paul-rogers thanks for the review; maybe all the additional changes should be done as a separate Jira that would retire the Janino, and switch to plain Java.

PR #1571, opened 2018-12-11
DRILL-6895: Warnings for license headers checks and wrong license hea…

vdiravka:
…ders in some files

To verify it please run:
`mvn license:check -Dlicense.skip=false`

See details in [DRILL-6895](https://issues.apache.org/jira/browse/DRILL-6895)

### DIFF
--- a/distribution/src/resources/drill-conf
+++ b/distribution/src/resources/drill-conf
@@ -1,19 +1,21 @@
 #!/bin/bash
 #
-# Licensed to the Apache Software Foundation (ASF) under one or more
-# contributor license agreements.  See the NOTICE file distributed with
-# this work for additional information regarding copyright ownership.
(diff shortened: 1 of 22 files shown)


distribution/src/resources/drill-conf
```diff
@@ -1,19 +1,21 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
arina-ielchiieva: Not sure if this change is necessary. Please check here and below.

vdiravka: It is good for python projects and for FreeBSD and OpenBSD systems.
None of these points satisfy Drill project. Therefore I will revert that changes and change `/usr/bin/env bash` in all other files.

vdiravka: Done

arina-ielchiieva:
+1

PR #1580, opened 2018-12-20
DRILL-6915: Disable generation of test tables with case-sensitive names for non-Linux systems

vvysotskyi:
- Splited scripts to generate regular tables for tests in one script and tables with a different case in names in another one.
- Introduced "linux" profile to specify a script which generates tables with names in different case only when the Linux system is used. 
- Disabled test which checks table names case sensitivity for non-Linux systems.
- Updated `hibernate-validator` dependency for `jcabi-mysql-maven-plugin` to fix `NoClassDefFoundError` error and specified `jaxb-api` to avoid problems with JDK 9+.
- Changed driver from `com.mysql.jdbc.Driver` to `com.mysql.cj.jdbc.Driver` for tests to avoid warnings (see https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-api-changes.html)

For details please see [DRILL-6915](https://issues.apache.org/jira/browse/DRILL-6915).

### DIFF
--- a/contrib/storage-jdbc/pom.xml
+++ b/contrib/storage-jdbc/pom.xml
@@ -34,6 +34,7 @@
     <mysql.connector.version>8.0.13</mysql.connector.version>
     <derby.database.name>drill_derby_test</derby.database.name>
     <mysql.database.name>drill_mysql_test</mysql.database.name>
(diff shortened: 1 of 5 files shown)


arina-ielchiieva:
Reviewed and tested the fix, looks good, +1.
Once @Ben-Zvi confirms that test fixes the issue on his env, PR is good to go.

PR #1583, opened 2018-12-21
DRILL-6922: Excluded options set on query level from query profile

KazydubB:


arina-ielchiieva:
@KazydubB I don't think that skipping query options completely is the best option. Maybe we should consider having two class variables: one for session options, another one for query options.

arina-ielchiieva:
Closing PR as alternative one was opened  - https://github.com/apache/drill/pull/1584.

PR #1585, opened 2018-12-23
DRILL-6925: Unable to generate Protobuf

vdiravka:
- use ${maven.multiModuleProjectDirectory}/header to find header file from any submodule
- suppress UnresolvedMavenProperty, since IDE expects that property should be set explicitly
- update "kr.motd.maven:os-maven-plugin" github.com/trustin/os-maven-plugin to the latest 1.6.1 version
- correction of ${user.name} propery for "maven-jar-plugin" <Built-By>
- update "apache-rat-plugin" to solve undefined "excludeSubprojects" in IDE

With above changes Drill's root maven POM file hasn't red undefined properties and tags in IDE.


vdiravka:
@arina-ielchiieva please review.

pom.xml
```diff
@@ -320,6 +320,7 @@
       <plugin>
         <groupId>org.apache.rat</groupId>
         <artifactId>apache-rat-plugin</artifactId>
+        <version>0.13</version>
```
arina-ielchiieva: Where rat version came from before?

arina-ielchiieva: Also we have this part in the code:
```
                  <pluginExecutionFilter>
                    <groupId>org.apache.rat</groupId>
                    <artifactId>apache-rat-plugin</artifactId>
                    <versionRange>[0.10,)</versionRange>
```

vdiravka: It came from `apache-21.pom` file
```
<plugin>
    <groupId>org.apache.rat</groupId>
    <artifactId>apache-rat-plugin</artifactId>
     <version>0.12</version>
</plugin>
```

That part relates to `lifecycle-mapping` plugin, which is intended for Eclipse m2e settings only. It has no influence on the Maven build itself. However `versionRange` has not upper bound there.

arina-ielchiieva:
Sounds good, +1.

vdiravka:
@arina-ielchiieva Thanks for the review. Also I've regenerate Java and C++ protobuf files. Please take a look.

arina-ielchiieva:
Looks good, thanks for checking Protobuf files for actuality.

PR #1591, opened 2019-01-02
DRILL-6933: Fix ctrl+enter when Impersonation is disabled

kkhatua:
Without impersonation, the key combination did not work because a conditional block of the FTLs skipped the Javascript functions required for submitting with the key combination.
This commit fixes that by unifying the functions that are common to both approaches of submitting a query (with and without impersoination)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
@@ -130,7 +130,7 @@ public class ProfileWrapper {
     }
     this.options = options;
 
-    this.onlyImpersonationEnabled = WebServer.isImpersonationOnlyEnabled(drillConfig);
+    this.onlyImpersonationEnabled = WebServer.isOnlyImpersonationEnabled(drillConfig);
     this.noProgressWarningThreshold = String.valueOf(drillConfig.getInt(ExecConstants.PROFILE_WARNING_PROGRESS_THRESHOLD));
   }
 
--- a/exec/java-exec/src/main/resources/rest/profile/profile.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/profile.ftl
@@ -23,18 +23,16 @@
 <script src="/static/js/dagre-d3.min.js"></script>
 <script src="/static/js/graph.js"></script>
 <script src="/static/js/jquery.dataTables-1.10.16.min.js"></script>
-<#if model.isOnlyImpersonationEnabled()>
-    <script src="/static/js/jquery.form.js"></script>
-    <script src="/static/js/querySubmission.js"></script>
-</#if>
-  <!-- Ace Libraries for Syntax Formatting -->
-  <script src="/static/js/ace-code-editor/ace.js" type="text/javascript" charset="utf-8"></script>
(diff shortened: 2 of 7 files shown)


exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -506,7 +504,8 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
     document.getElementById('queryForm')
             .addEventListener('keydown', function(e) {
       if (!(e.keyCode == 13 && (e.metaKey || e.ctrlKey))) return;
-      if (e.target.form) doSubmitQueryWithUserName();
+      if (e.target.form) 
+        <#if model.isImpersonationOnlyEnabled()>doSubmitQueryWithUserName()<#else>submitQuery()</#if>;
```
sohami: I believe the root cause of the issue is here and below where for non-impersonation case submit was not actually done. In those cases username will be null inside `doSubmitQueryWithUserName` and will return with an alert.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -344,8 +344,8 @@ public String getOperatorsJSON() {
    * @return true if impersonation is enabled without authentication,
    *         is needed to indicated if user name should be included when re-running the query
    */
-  public boolean isOnlyImpersonationEnabled() {
-    return onlyImpersonationEnabled;
+  public boolean isImpersonationOnlyEnabled() {
+    return impersonationOnlyEnabled;
```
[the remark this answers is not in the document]
kkhatua: I'm reflecting the actual method call's format https://github.com/apache/drill/blob/5ceff0a005b4fbfbb552275039a9af24f7f0f44b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java#L131
Should I change this across the board then?

exec/java-exec/src/main/resources/rest/profile/profile.ftl
kkhatua: Yes, this was the root cause, because we trap the key even (Ctrl + Enter) and submit the query. 
We were trapping the event it for only impersonation scenarios. This takes care of that and abstracts the submit call. (i.e. the AJAX call). This way, we have only one method to do the actual submission, and the non-existent `username` will be handled by not setting header.

Also, PR #1592 will use this method to further improve the user experience by blocking interaction.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -60,7 +60,7 @@
   private final HashMap<String, Long> majorFragmentTallyMap;
   private final long majorFragmentTallyTotal;
   private final OptionList options;
-  private final boolean onlyImpersonationEnabled;
+  private final boolean isOnlyImpersonationEnabled;
```
sohami: usually the convention is that boolean variable name should not start with `"is"`, it's the getter function which should have that. Variable name should only be `onlyImpersonationEnabled`.

kkhatua: Done.

PR #1592, opened 2019-01-02
DRILL-6939: Indicate when a query is submitted and is in progress

kkhatua:
On query submission, a modal popup blocks further interaction until the result of the query are available.
Since the Query ID is not available at this point, the only way to cancel a running query is to navigate to it via the `/profiles` tab. For this, the modal allows a button to pop out to this tab without closing the current window which is waiting for the result-set.

Screenshot (when querying sys.profiles with 5.8K entries)
image


exec/java-exec/src/main/resources/rest/static/js/querySubmission.js
```diff
@@ -11,6 +11,38 @@
...
  */
 
+//Elements for Timer in LoadingModal
+var elapsedTime = 0;
+var delay = 1000; //msec
+var timeTracker = null; //Handle for stopping watch
+
+//Padding for time
+function str_pad_left(string,pad,length) {
+    return (new Array(length+1).join(pad)+string).slice(-length);
```
sohami: How about below:

```
function str_pad_left(timeVal, length) {
  return timeVal.toString().slice(-length);
}
```

exec/java-exec/src/main/resources/rest/query/query.ftl
```diff
@@ -39,6 +39,32 @@
     Sample SQL query: <strong>SELECT * FROM cp.`employee.json` LIMIT 20</strong>
   </div>
 
+  <!-- Loading Modal -->
```
sohami: It looks like profile.ftl and query.ftl has lots of same duplicated code. It would be good if we can somehow refactor it so that both uses same piece of code

kkhatua: I refactored the Javascript to have the two FTLs share it. However, these is HTML element that I didn't think needed a separate shared FTL for embedding.

exec/java-exec/src/main/resources/rest/static/js/querySubmission.js
kkhatua: That actually shows only one digit for values less than 10. Modifying it to 
`let prettyTime = ("0" + minutes).slice(-2)+':'+ ("0" + seconds).slice(-2);`

exec/java-exec/src/main/resources/rest/query/query.ftl
kkhatua: Using Freemarker's `<#include>` , embedded the shared code for the modal.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -147,6 +147,9 @@ table.sortable thead .sorting_desc { background-image: url("/static/img/black-de
             </button>
           </form>
       </p>
+
+<#include "*/runningQuery.ftl">
```
sohami: Thanks for doing this. May be in future we can move some of the other common code in runningQuery.ftl too.

PR #1593, opened 2019-01-02
DRILL-6050: Provide a limit to number of rows fetched for a query in UI

kkhatua:
Currently, the WebServer side needs to process the entire set of results and stream it back to the WebClient, which puts immense pressure on the WebServer when rendering the resultset. 

Since the WebUI does paginate results, we can load a larger set for pagination on the browser client and relieve pressure off the WebServer to host all the data (most of which will never be streamed to the browser).
e.g. Fetching all rows from a 1Billion records table is impractical and can be capped at (say) 1K. Currently, the user has to explicitly specify LIMIT in the submitted query. 

An option is provided in the field to allow for this entry, and can be set to selected by default for the Web UI.
The submitted query indicates that an auto-limiting wrapper was applied.
In addition, the resultset is now configurable to allow for the default number or rows displayed per page to be changed from 10 to anything that the user might want.

Configuration additions in `drill-module.conf` (changes should be made in `drill-override.conf` ):
```
drill.exec.http: {
    web.client.resultset: {
        autolimit {
            enable: true,
            rows: 1000
        },
        defaultRowsPerPage: 10
    }
}
```

Screenshot with default (unselected) set to limit of 23:
image

Screenshot with default rows per page changed to 12.
image

Screenshot of the profile indicates the auto-limit having been applied:
image


exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -155,6 +155,13 @@ drill.exec: {
             reservation: 0,
             maximum: 9223372036854775807
         }
+    },
+    web.client.resultset: {
+        autolimit {
+            enable: true,
+            rows: 1000
+        },
+        defaultRowsPerPage: 10
```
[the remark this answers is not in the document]
kkhatua: Modifying this to allows admins to specify the entire list. i.e. they can change the options from the default `[10,25,50,100]`

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -155,6 +155,13 @@ drill.exec: {
             reservation: 0,
             maximum: 9223372036854775807
         }
+    },
+    web.client.resultset: {
+        autolimit {
```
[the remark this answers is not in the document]
kkhatua: Unchecking the auto-limit by default.

kkhatua:
@sohami  Done the changes.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -225,9 +225,9 @@ private ExecConstants() {
   public static final String HTTP_SPNEGO_PRINCIPAL = "drill.exec.http.auth.spnego.principal";
   public static final String HTTP_SPNEGO_KEYTAB = "drill.exec.http.auth.spnego.keytab";
   //Control Web UI Resultset
-  public static final String HTTP_WEB_CLIENT_RESULTSET_AUTOLIMIT_ENABLE = "drill.exec.http.web.client.resultset.autolimit.enable";
+  public static final String HTTP_WEB_CLIENT_RESULTSET_AUTOLIMIT_CHECKED = "drill.exec.http.web.client.resultset.autolimit.checked";
   public static final String HTTP_WEB_CLIENT_RESULTSET_AUTOLIMIT_ROWS = "drill.exec.http.web.client.resultset.autolimit.rows";
-  public static final String HTTP_WEB_CLIENT_RESULTSET_DEFAULTROWSPERPAGE = "drill.exec.http.web.client.resultset.defaultRowsPerPage";
+  public static final String HTTP_WEB_CLIENT_RESULTSET_ROWS_PER_PAGE = "drill.exec.http.web.client.resultset.rowsPerPage";
```
sohami: This parameter now represents all the options for rows per page not a single value so consider naming it `rowsPerPageOptions` / `rowsPerPageValues`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryResources.java
```diff
@@ -128,9 +133,9 @@ public int getDefaultRowsAutoLimited() {
     private final List<String> columns;
     private final List<List<String>> rows;
     private final String queryId;
-    private final int defaultRowsPerPage;
+    private final String defaultRowsPerPage;
```
sohami: Same here and below. Rename`defaultRowsPerPage` --> `rowsPerPageOptions`

exec/java-exec/src/main/resources/rest/static/js/querySubmission.js
```diff
@@ -72,4 +82,26 @@ function submitQuery() {
...
+    //dBug: console.log("Query Input:" + origQueryText);
+    var mustWrapWithLimit = $('input[name="forceLimit"]:checked').length > 0;
+    if (mustWrapWithLimit) {
+        var semicolonIdx = origQueryText.lastIndexOf(';');
+        //Check and eliminate trailing semicolon
+        if (semicolonIdx  == origQueryText.length-1 ) {
+          origQueryText = origQueryText.substring(0, semicolonIdx)
+        }
+        var qLimit = $('#queryLimit').val();
+        var wrappedQuery = "-- [autoLimit: " + qLimit + " rows]\nselect * from (\n" + origQueryText + "\n) limit " + qLimit;
```
sohami: Should we insert a semicolon in the end of wrapped query ?

kkhatua: Not required. Drill (Foreman, I think) itself removes the query during processing. That is why you don't need a semi-colon when submitting a query as well.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryResources.java
kkhatua: +1

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
kkhatua: +1 Will change it to `rowsPerPageValues`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryResources.java
sohami: Looks like you missed updating the name here.

sohami:
1 minor pending comment other than that LGTM. +1

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/QueryResources.java
kkhatua: Yes.. missed it. Fixed it all around.

ihuzenko:
Hello @kkhatua , this solution may cause submission of queries with incorrect syntax. For example, when user enters 'SHOW DATABASES' and after limitation wrapping 'SELECT * FROM (SHOW DATABASES) LIMIT 10' will be posted. This results into parsing errors, like: 
`Query Failed: An Error Occurred
org.apache.drill.common.exceptions.UserRemoteException: PARSE ERROR: Encountered "( show" at line 2, column 15. Was expecting one of: <IDENTIFIER> ... `. 

Could you please fix the issue and test limit functionality with all non-select kinds of queries supported by Drill ? Also I'd like to kindly ask you for change tooltip for question mark from onclick alert to on hover one, for example see [Tooltips](https://www.w3schools.com/css/css_tooltip.asp). Thank you in advance.

kkhatua:
@ihuzenko thanks for the suggestions. I've opened a JIRA for this issue. [DRILL-6960](https://issues.apache.org/jira/browse/DRILL-6960)

PR #1595, opened 2019-01-04
DRILL-6918: Querying empty topics fails with "NumberFormatException"

aravi5:
If none of the project / filter columns, exist in the records, `ensureAtLeastOneField` (or the Scan operator) adds at least one field as nullable integer (or nullable varchar if `allTextmode` is enabled).

The downstream Filter operator would then go on to fail with `NumberFormatException` because it tries to convert empty fields to integers.

Since `ensureAtLeastOneField` is called after reading all the messages in a batch, it can be skipped if the batch is empty.


kkhatua:
@aravi5  do you have a reviewer in mind?

aravi5:
Yes, @sohami and @HanumathRao, can you please review the change?

sohami:
@aravi5 - The title of the PR and JIRA doesn't match. Can you please fix that and also squash both the commits and rebase on master.

PR #1599, opened 2019-01-05
DRILL-6903: SchemaBuilder code improvements

arina-ielchiieva:
1. ColumnBuilder: setPrecisionAndScale method
2. SchemaContainer: addColumn method parameter AbstractColumnMetadata was changed to ColumnMetadata
3. MapBuilder / RepeatedListBuilder / UnionBuilder: added constructors without parent, made buildColumn method public
4. TupleMetadata: added toMetadataList method
5. Other refactoring

Details in [DRILL-6903](https://issues.apache.org/jira/browse/DRILL-6903).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/model/MetadataProvider.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/model/MetadataProvider.java
@@ -18,7 +18,6 @@
 package org.apache.drill.exec.physical.rowSet.model;
 
 import org.apache.drill.exec.record.MaterializedField;
-import org.apache.drill.exec.record.metadata.AbstractColumnMetadata;
 import org.apache.drill.exec.record.metadata.ColumnMetadata;
 import org.apache.drill.exec.record.metadata.MetadataUtils;
 import org.apache.drill.exec.record.metadata.RepeatedListColumnMetadata;
@@ -38,17 +37,16 @@ public interface MetadataProvider {
   TupleMetadata tuple();
   VariantMetadata variant();
 
-  public static class VectorDescrip {
+  class VectorDescrip {
     public final MetadataProvider parent;
     public final ColumnMetadata metadata;
 
     public VectorDescrip(MetadataProvider provider, ColumnMetadata metadata) {
-      parent = provider;
+      this.parent = provider;
       this.metadata = metadata;
     }
 
-    public VectorDescrip(MetadataProvider provider, int index,
-        MaterializedField field) {
+    public VectorDescrip(MetadataProvider provider, int index, MaterializedField field) {
       this(provider, provider.metadata(index, field));
     }
 
@@ -57,7 +55,7 @@ public interface MetadataProvider {
     }
   }
 
-  public static class MetadataCreator implements MetadataProvider {
+  class MetadataCreator implements MetadataProvider {
 
     private final TupleSchema tuple;
 
@@ -101,7 +99,7 @@ public interface MetadataProvider {
     }
   }
 
-  public static class VariantSchemaCreator implements MetadataProvider {
+  class VariantSchemaCreator implements MetadataProvider {
 
     private final VariantSchema variantSchema;
 
@@ -130,7 +128,7 @@ public interface MetadataProvider {
     }
   }
 
-  public static class ArraySchemaCreator implements MetadataProvider {
+  class ArraySchemaCreator implements MetadataProvider {
 
     private final RepeatedListColumnMetadata arraySchema;
 
@@ -142,7 +140,7 @@ public interface MetadataProvider {
     public ColumnMetadata metadata(int index, MaterializedField field) {
       assert index == 0;
       assert arraySchema.childSchema() == null;
-      AbstractColumnMetadata childSchema = MetadataUtils.fromField(field.cloneEmpty());
+      ColumnMetadata childSchema = MetadataUtils.fromField(field.cloneEmpty());
       arraySchema.childSchema(childSchema);
       return childSchema;
     }
@@ -163,7 +161,7 @@ public interface MetadataProvider {
     }
   }
 
-  public static class MetadataRetrieval implements MetadataProvider {
+  class MetadataRetrieval implements MetadataProvider {
 
     private final TupleMetadata tuple;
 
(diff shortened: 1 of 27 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/model/MetadataProvider.java
```diff
@@ -38,17 +37,16 @@
   TupleMetadata tuple();
   VariantMetadata variant();
 
-  public static class VectorDescrip {
+  class VectorDescrip {
```
paul-rogers: public? Or, if intentionally protected, is handy to declare that so we know that protected is intentional.

Also, the class changed from static to inner: this means that it must be created in the context of its outer class, and carries a pointer to the outer object. Since that was not needed here, the original declaration used "static" to avoid unnecessary complexity.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/AbstractColumnMetadata.java
```diff
@@ -87,7 +86,8 @@ public AbstractColumnMetadata(AbstractColumnMetadata from) {
     expectedElementCount = from.expectedElementCount;
   }
 
-  protected void bind(TupleSchema parentTuple) { }
+  @Override
+  public void bind(TupleMetadata parentTuple) { }
```
paul-rogers: Not sure if an issue here, but the "Schema"/"Metadata" split handles the fact that the column accessors use these interfaces, but are in the vector package. The "Metadata" interfaces thus are designed to be visible to vector. On the other hand, the implementation needed visibility to BatchSchema, which was defined in java-exec.

So, some careful fiddling was needed to ensure that the column accessors could work just with the interfaces, leaving the implementations to be defined in java-exec.

All this can certainly be fixed; just tried to minimize code impact during the long merge process.

Anyway, the change above may not be an issue, but you now have the context to understand if it is.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/MapBuilder.java
```diff
@@ -123,30 +126,36 @@ public RepeatedListBuilder addRepeatedList(String name) {
     return tupleBuilder.addRepeatedList(this, name);
   }
 
-  private MapColumnMetadata buildCol() {
+  public MapColumnMetadata buildColumn() {
```
paul-rogers: This was private because only the top-level schema builder was meant to be public. These intermediate items are used internally to create a map in the middle of building a schema.

Might well work by itself, just didn't test it that way. Looks like you've made the parent optional, which might be OK.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/RepeatedListBuilder.java
```diff
@@ -26,12 +26,15 @@
...
  */
-
 public class RepeatedListBuilder implements SchemaContainer {
 
   private final SchemaContainer parent;
   private final String name;
-  private AbstractColumnMetadata child;
+  private ColumnMetadata child;
+
+  public RepeatedListBuilder(String name) {
```
paul-rogers: Here and elsewhere. Since these classes now work in two modes, might be good to explain that somewhere, especially how to use the class as a standalone builder.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/TupleBuilder.java
```diff
@@ -139,21 +135,17 @@ public MapBuilder addMapArray(SchemaContainer parent, String name) {
   }
 
   public UnionBuilder addUnion(SchemaContainer parent, String name) {
-    return new UnionBuilder(parent, name, MinorType.UNION, DataMode.OPTIONAL);
+    return new UnionBuilder(parent, name, MinorType.UNION);
   }
 
   public UnionBuilder addList(SchemaContainer parent, String name) {
-    return new UnionBuilder(parent, name, MinorType.LIST, DataMode.REPEATED);
+    return new UnionBuilder(parent, name, MinorType.LIST);
```
paul-rogers: Nice simplification, since the type implies the mode.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/TupleSchema.java
```diff
@@ -145,6 +145,11 @@ public boolean isEquivalent(TupleMetadata other) {
     return cols;
   }
 
+  @Override
+  public List<ColumnMetadata> toMetadataList() {
+    return new ArrayList<>(nameSpace.entries());
+  }
```
paul-rogers: How will this be used? A list of metadata is OK for serialization. But, is very awkward for semantic processing. On a large row, an O(n) lookup will turn into an O(n^2) algorithm as each column is looked up. The tuple structure is designed to provide both a name index and an order.

(By the way, if you have not already, the tuple schema class can be simplified by the use of the LinkedHashMap class that I didn't know about at the time...)

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/TupleSchema.java
```diff
@@ -179,18 +184,16 @@ public String fullName() {
...
-        buf.append(", ");
-      }
-      buf.append(md.toString());
-    }
-    buf.append("]");
-    return buf.toString();
+
+    builder.append(nameSpace.entries().stream()
+      .map(ColumnMetadata::toString)
+      .collect(Collectors.joining(", ")));
```
paul-rogers: Show off! That's some fancy functional code for such a simple task... Much cleaner.

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/test/TestSchemaBuilder.java
```diff
@@ -81,7 +79,7 @@ public void testRowPreBuilt() {
 
     MaterializedField aField = MaterializedField.create("a",
         Types.optional(MinorType.VARCHAR));
-    AbstractColumnMetadata bCol = MetadataUtils.newScalar("b",
+    ColumnMetadata bCol = MetadataUtils.newScalar("b",
```
paul-rogers: Consider adding test functions for the new stand-alone versions of the builders.

arina-ielchiieva:
@paul-rogers thanks for the code review. Addressed code review comments.

PR #1603, opened 2019-01-08
DRILL-6955: storage-jdbc tests improvements

vvysotskyi:
- Remove plugins usage for instantiating test databases;
- Replace derby with h2 database.

For details see [DRILL-6955](https://issues.apache.org/jira/browse/DRILL-6955).


contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithMySQLIT.java
```diff
@@ -17,24 +17,101 @@
...
+
+      JdbcStoragePlugin jdbcCaseSensitiveStoragePlugin = new JdbcStoragePlugin(jdbcCaseSensitiveStorageConfig,
+          cluster.drillbit().getContext(), mysqlCaseSensitivePluginName);
+      pluginRegistry.addPluginToPersistentStoreIfAbsent(mysqlCaseSensitivePluginName, jdbcCaseSensitiveStorageConfig, jdbcCaseSensitiveStoragePlugin);
+    }
+  }
+
+  @AfterClass
+  public static void stopMysql() {
+    mysqld.stop();
```
arina-ielchiieva: Maybe we should check for null first?

contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithH2IT.java
```diff
@@ -18,48 +18,98 @@
...
   }
 
   @Test
   public void testCrossSourceMultiFragmentJoin() throws Exception {
-    testNoResult("SET `planner.slice_target` = 1");
-    test("select x.person_id, y.salary from derby.drill_derby_test.person x "
-        + "join %s y on x.person_id = y.person_id ", TABLE_NAME);
+    try {
+      client.alterSession(ExecConstants.SLICE_TARGET, 1);
+      client.queryBuilder()
```
arina-ielchiieva: Not critical but there is `queryBuilder()` method that implicitly performs `client.queryBuilder()`, also there is `run` method for those queries that does not need to return any result (check `ClusterTest` class).

vvysotskyi: Thanks for the advice, used `queryBuilder()` and `run()` where possible.

contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithMySQLIT.java
vvysotskyi: Agree, added a check.

arina-ielchiieva:

PR #1605, opened 2019-01-09
DRILL-6956: Maintain a single entry for Drill Version in the pom file

kkhatua:
Currently, updating the version information for a Drill release involves updating 30+ pom files.
The right way would be to use the Multi Module Setup for Maven CI.
https://maven.apache.org/maven-ci-friendly.html#Multi_Module_Setup

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -24,7 +24,7 @@
   <parent>
     <artifactId>drill-root</artifactId>
     <groupId>org.apache.drill</groupId>
-    <version>1.16.0-SNAPSHOT</version>
+    <version>${revision}</version>
   </parent>
 
   <artifactId>drill-common</artifactId>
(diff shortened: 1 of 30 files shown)


vdiravka:
@kkhatua This is a good change, but there are two concerns with it:

- according to [maven doc](https://maven.apache.org/maven-ci-friendly.html#Multi_Module_Setup) this feature works starting from 3.5.0-beta-1 Maven version. It means that we push all devs to update their Maven to 3.6.0 version. Currently Drill can be built with [Maven 3.3.3 version](https://github.com/apache/drill/blob/master/docs/dev/Environment.md#how-to-build-and-run-apache-drill). Do we really want it? Possibly it can be discussed on mailing list.
- Did you verify `mvn release:prepare` [maven phase](https://github.com/parthchandra/drill/wiki/Drill-Release-Process#manual-release-process)? Did it update version entry only in a single place?

PR #1610, opened 2019-01-11
DRILL-6969: Fix inconsistency of reading MaprDB JSON tables using hive plugin when native reader is enabled

vvysotskyi:
For problem description see [DRILL-6969](https://issues.apache.org/jira/browse/DRILL-6969).

### DIFF
--- a/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
+++ b/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
@@ -35,6 +35,7 @@ import org.apache.drill.common.expression.ValueExpressions.TimeExpression;
 import org.apache.drill.common.expression.ValueExpressions.TimeStampExpression;
 import org.apache.drill.common.expression.ValueExpressions.VarDecimalExpression;
 import org.apache.drill.common.expression.visitors.AbstractExprVisitor;
+import org.apache.drill.exec.expr.fn.impl.DateUtility;
 import org.joda.time.LocalTime;
 import org.ojai.Value;
 import org.ojai.types.ODate;
@@ -42,7 +43,6 @@ import org.ojai.types.OTime;
 
 import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableMap;
 import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableSet;
-import com.mapr.db.rowcol.KeyValueBuilder;
 import com.mapr.db.util.SqlHelper;
 
 import org.ojai.types.OTimestamp;
@@ -51,8 +51,8 @@ class CompareFunctionsProcessor extends AbstractExprVisitor<Boolean, LogicalExpr
 
   private String functionName;
   private Boolean success;
-  private Value value;
-  private SchemaPath path;
+  protected Value value;
+  protected SchemaPath path;
 
   public CompareFunctionsProcessor(String functionName) {
     this.functionName = functionName;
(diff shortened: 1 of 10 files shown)


vvysotskyi:

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
```diff
@@ -69,21 +69,52 @@ public Boolean visitUnknown(LogicalExpression e, LogicalExpression valueArg) thr
...
+      }
+    };
+    return processWithEvaluator(call, processor);
+  }
+
+  private static CompareFunctionsProcessor processWithEvaluator(FunctionCall call, CompareFunctionsProcessor evaluator) {
     String functionName = call.getName();
     LogicalExpression nameArg = call.args.get(0);
-    LogicalExpression valueArg = call.args.size() >= 2? call.args.get(1) : null;
-    CompareFunctionsProcessor evaluator = new CompareFunctionsProcessor(functionName);
```
amansinha100: The new code modifies the evaluator that is supplied by changing the `functionName` and `success` flag and returns   that instance.  Could this cause side effects ? for example if the caller passed the same evaluator to 2 different methods and was expecting the return to be a new instance ?  The 2 called methods may overwrite each other's state.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
```diff
@@ -69,21 +69,52 @@ public Boolean visitUnknown(LogicalExpression e, LogicalExpression valueArg) thr
...
+   * before converting to {@link OTimestamp} instance.
+   *
+   * @param call function call to be pushed
+   * @return CompareFunctionsProcessor instance which contains converted function call
+   */
+  public static CompareFunctionsProcessor processWithTimeZoneOffset(FunctionCall call) {
+    CompareFunctionsProcessor processor = new CompareFunctionsProcessor(call.getName()) {
+      @Override
+      protected boolean visitTimestampExpr(SchemaPath path, TimeStampExpression valueArg) {
+        // converts timestamp value from local time zone to UTC
```
amansinha100: You may want to add that  '....UTC since the record reader reads the timestamp in local timezone if the timezonewithoffset flag is enabled'

amansinha100:
Changes overall LGTM.  One minor and one less minor comment .. pls see if my interpretation is correct.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
vvysotskyi: This method is private and contains common code for existing public `CompareFunctionsProcessor.process()` and new `CompareFunctionsProcessor.processWithTimeZoneOffset()` methods.
These public methods create new `CompareFunctionsProcessor` instances (or its inheritors) which are passed to this method, so the behavior is not changed.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
vvysotskyi: Thanks, added.

vvysotskyi:
@amansinha100, thanks for the review, I have addressed CR comments.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
amansinha100: Ok, sounds good.

PR #1611, opened 2019-01-12
DRILL-6971: Display query state in query result page

sohami:
With this change query state will be displayed in the query result page as below:

Example of completed query:
completed_example

Example of cancelled query:
cancelled_example

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/AbstractDisposableUserClientConnection.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/rpc/AbstractDisposableUserClientConnection.java
@@ -45,6 +45,8 @@ public abstract class AbstractDisposableUserClientConnection implements UserClie
 
   protected volatile UserException exception;
 
+  protected String queryState;
+
   /**
    * Wait until the query has completed or timeout is passed.
    *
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/resources/rest/query/result.ftl
```diff
@@ -34,6 +34,9 @@
     <td align='left'>
       <button type="button"  title="Open in new window" onclick="popOutProfile('${model.getQueryId()}');" class="btn btn-default btn-sm">
       <b>Query Profile:</b> ${model.getQueryId()} <span class="glyphicon glyphicon-new-window"/></button>
+     </td>
+     <td>
+         <span class="input-group-addon" style="font-size:95%"><b>Query State:</b> ${model.getQueryState()}</span>
```
kkhatua: It might be worth putting the Status in a Label with the colour indicative of the status.
Here is a good reference: https://www.w3schools.com/bootstrap/bootstrap_badges_labels.asp

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/AbstractDisposableUserClientConnection.java
```diff
@@ -73,6 +75,7 @@ public void sendResult(RpcOutcomeListener<Ack> listener, QueryResult result) {
 
     // Release the wait latch if the query is terminated.
     final QueryState state = result.getQueryState();
+    queryState = state.toString();
```
kkhatua: Should this be labelled as `queryStateAsStr` since we already have a `state` variable in the same block's context?

PR #1612, opened 2019-01-12
DRILL-6970: fix issue with logregex format plugin where drillbuf was overflowing

jcmcote:


arina-ielchiieva:
@jcmcote could you please add unit test?

arina-ielchiieva:
@jcmcote is there any update?

jcmcote:
I did not have time to add a test case yet. However I know the fix is good,
using it in production.

cgivre:
@jcmcote 
Could you just modify one of the existing unit tests with a long string?

JohnOmernik:
Is there a JIRA for this?. I am getting an IndexOutofBoundsException, and am wondering if it's the same issue.  (Drill 1.14, I am trying to see if this would fix the plugin)

SYSTEM ERROR: IndexOutOfBoundsException: index: 32760, length: 9 (expected: range(0, 32768))

jcmcote:
yes that's the issue. 32k seems to be the buffer size and is never
recycled. I think that's what the safe methods do.

jcmcote:
I think I fixed the build issue but it does not look like it is picking it up...

PR #1614, opened 2019-01-17
DRILL-6962: Function coalesce returns an Error when none of the columns in coalesce exist in a parquet file

KazydubB:
- Updated UntypedNullVector to hold value count when vector is allocated and transfered to another one;
- Updated RecordBatchLoader and DrillCursor to handle case when only UntypedNull values are present in RecordBatch (special case when data buffer is null but actual values are present);
- Added functions to cast UntypedNull value to other types;
- Moved UntypedReader, UntypedHolderReaderImpl and UntypedReaderImpl from org.apache.drill.exec.vector.complex.impl to org.apache.drill.exec.vector package.

### DIFF
--- a/exec/java-exec/src/main/codegen/data/Casts.tdd
+++ b/exec/java-exec/src/main/codegen/data/Casts.tdd
@@ -168,5 +168,22 @@
     {from: "NullableVarBinary", to: "NullableFloat8", major: "EmptyString", javaType:"Double", parse:"Double"},
 
     {from: "NullableVarBinary", to: "NullableVarDecimal", major: "NullableVarCharDecimalComplex"},
+
+    {from: "UntypedNull", to: "Bit", major: "UntypedNull"},
+    {from: "UntypedNull", to: "TinyInt", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Int", major: "UntypedNull"},
+    {from: "UntypedNull", to: "BigInt", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Float4", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Float8", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Date", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Time", major: "UntypedNull"},
+    {from: "UntypedNull", to: "TimeStamp", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Interval", major: "UntypedNull"},
+    {from: "UntypedNull", to: "IntervalDay", major: "UntypedNull"},
+    {from: "UntypedNull", to: "IntervalYear", major: "UntypedNull"},
+    {from: "UntypedNull", to: "VarBinary", major: "UntypedNull"},
+    {from: "UntypedNull", to: "VarChar", major: "UntypedNull"},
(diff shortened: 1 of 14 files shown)


arina-ielchiieva:
@vvysotskyi please review.

KazydubB:
@vvysotskyi I've addressed your comments, please take a look.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
```diff
@@ -752,18 +770,59 @@ public void testCastTimeLiteralInFilter() throws Exception {
...
+          .sqlQuery(q)
           .unOrdered()
           .baselineColumns("coal")
           .baselineValues(new Object[] {null})
           .go();
     }
   }
+
+  private static Map<String, Triple<TypeProtos.MinorType, Integer, Integer>> createCastTypeMap() {
+    Map<String, Triple<TypeProtos.MinorType, Integer, Integer>> typesMap = new HashMap<>();
```
vvysotskyi: It would be better to pass `MaterializedField` instance instead of passing `TypeProtos.MinorType, Integer, Integer`.

exec/java-exec/src/main/codegen/data/Casts.tdd
```diff
@@ -168,5 +168,25 @@
     {from: "NullableVarBinary", to: "NullableFloat8", major: "EmptyString", javaType:"Double", parse:"Double"},
 
     {from: "NullableVarBinary", to: "NullableVarDecimal", major: "NullableVarCharDecimalComplex"},
+
+    {from: "UntypedNull", to: "TinyInt", major: "UntypedNull"},
```
vvysotskyi: Is it a complete list of Drill types? I don't see bit for example

exec/java-exec/src/main/codegen/data/Casts.tdd
```diff
@@ -168,5 +168,25 @@
...
+    {from: "UntypedNull", to: "Time", major: "UntypedNull"},
+    {from: "UntypedNull", to: "TimeStamp", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Interval", major: "UntypedNull"},
+    {from: "UntypedNull", to: "IntervalDay", major: "UntypedNull"},
+    {from: "UntypedNull", to: "IntervalYear", major: "UntypedNull"},
+    {from: "UntypedNull", to: "VarBinary", major: "UntypedNull"},
+    {from: "UntypedNull", to: "VarChar", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Var16Char", major: "UntypedNull"},
+    {from: "UntypedNull", to: "VarDecimal", major: "UntypedNull"},
+    {from: "UntypedNull", to: "Decimal9", major: "UntypedNull"},
```
vvysotskyi: Please remove deprecated decimal types.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestCastFunctions.java
KazydubB: Good point. made `Map<String, TypeProtos.MajorType>` instead.

KazydubB:
@vvysotskyi thanks for you review. Updated.

PR #1617, opened 2019-01-21
DRILL-6533: Allow using literal values in functions which expect FieldReader instead of ValueHolder

vvysotskyi:
For details please see [DRILL-6533](https://issues.apache.org/jira/browse/DRILL-6533).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/interpreter/InterpreterEvaluator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/interpreter/InterpreterEvaluator.java
@@ -24,6 +24,7 @@ import java.util.Objects;
 import javax.annotation.Nullable;
 import javax.inject.Inject;
 
+import org.apache.drill.exec.expr.BasicTypeHelper;
 import org.apache.drill.shaded.guava.com.google.common.base.Function;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.expression.BooleanOperator;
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/interpreter/InterpreterEvaluator.java
```diff
@@ -307,36 +317,47 @@ public ValueHolder visitFunctionHolderExpression(FunctionHolderExpression holder
...
       DrillSimpleFuncHolder holder = (DrillSimpleFuncHolder) holderExpr.getHolder();
 
-      ValueHolder [] args = new ValueHolder [holderExpr.args.size()];
+      // function arguments may have different types:
+      // usually ValueHolder inheritors but sometimes FieldReader ones
+      Object[] args = new Object[holderExpr.args.size()];
       for (int i = 0; i < holderExpr.args.size(); i++) {
-        args[i] = holderExpr.args.get(i).accept(this, inIndex);
+        ValueHolder valueHolder = holderExpr.args.get(i).accept(this, inIndex);
+        args[i] = valueHolder;
```
arina-ielchiieva: Maybe it is better to define argument in the variable and then in the end set it `args`?

vvysotskyi: Agree, thanks, done.

arina-ielchiieva:
+1

PR #1619, opened 2019-01-23
DRILL-6910: Allow applying DrillPushProjectIntoScanRule at physical phase

vvysotskyi:
- Enhance `DrillPushProjectIntoScanRule` to be applied at the physical stage;
- Avoid recreating `TableScan` rel nodes with the same columns during applying `DrillPushProjectIntoScanRule`;
- Add unit test;
- Enable tests for [DRILL-912](https://issues.apache.org/jira/browse/DRILL-912) (the reason why they were disabled was fixed a long time ago).

For problem description please see [DRILL-6910](https://issues.apache.org/jira/browse/DRILL-6910).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushProjectIntoScanRule.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushProjectIntoScanRule.java
@@ -25,10 +25,16 @@ import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.core.TableScan;
 import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.rules.ProjectRemoveRule;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexNode;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
+import org.apache.drill.exec.planner.common.DrillProjectRelBase;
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushProjectIntoScanRule.java
```diff
@@ -43,51 +49,75 @@
...
     try {
-
       if (scan.getRowType().getFieldList().isEmpty()) {
         return;
       }
 
       ProjectPushInfo projectPushInfo = DrillRelOptUtil.getFieldsInformation(scan.getRowType(), project.getProjects());
-      if (!canPushProjectIntoScan(scan.getTable(), projectPushInfo)) {
+      if (!canPushProjectIntoScan(scan.getTable(), projectPushInfo)
+          || skipScanConversion(projectPushInfo.createNewRowType(project.getCluster().getTypeFactory()), scan)) {
```
arina-ielchiieva: If we skip project push down into scan because scan and rows types are the same, project will be still in plan? or it will be removed by some other rule later?

vvysotskyi: For this case project will be removed using `ProjectRemoveRule`.
Thanks for pointing this, added a comment to clarify it.

arina-ielchiieva:
+1

PR #1620, opened 2019-01-24
DRILL-6997: Semijoin is changing the join ordering for some tpcds que…

HanumathRao:
…ries.

After transforming DrillJoin -- (DrillRel, DrillAgg) to DrillSemiJoinRel, join ordering is not considering the Semijoin with other joins in the query. This is resulting in sub-optimal plans during the planning phase. The following code changes are made to resolve the issue.
a) A new rule DrillSemiJoinRule is introduced to transform the above pattern to DrillSemiJoinRel.
b) RowType of the DrillSemiJoinRel is same as that of Drill's join relations. 
c) validity checks are overloaded to check for the appropriate row type.

@amansinha100  Can you please review this PR.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillSemiJoinRel.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillSemiJoinRel.java
@@ -17,13 +17,19 @@
  */
 package org.apache.drill.exec.planner.logical;
 
+import org.apache.calcite.rex.RexChecker;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.util.Litmus;
+import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableList;
 import org.apache.calcite.plan.RelOptCluster;
 import org.apache.calcite.plan.RelTraitSet;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.JoinInfo;
 import org.apache.calcite.rel.core.JoinRelType;
 import org.apache.calcite.rel.core.SemiJoin;
+import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.validate.SqlValidatorUtil;
 import org.apache.calcite.util.ImmutableIntList;
 import org.apache.calcite.util.Pair;
 import org.apache.drill.common.expression.FieldReference;
@@ -78,7 +84,6 @@ public class DrillSemiJoinRel extends SemiJoin implements DrillJoin, DrillRel {
     List<String> fields = new ArrayList<>();
     fields.addAll(getInput(0).getRowType().getFieldNames());
     fields.addAll(getInput(1).getRowType().getFieldNames());
-    Preconditions.checkArgument(DrillJoinRel.isUnique(fields));
     final int leftCount = left.getRowType().getFieldCount();
     final List<String> leftFields = fields.subList(0, leftCount);
     final List<String> rightFields = fields.subList(leftCount, leftCount + right.getRowType().getFieldCount());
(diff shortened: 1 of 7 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -277,6 +277,10 @@ protected DrillRel convertToRawDrel(final RelNode relNode) throws SqlUnsupported
         }
       }
 
+      if (context.getPlannerSettings().isSemiJoinEnabled() &&
```
amansinha100: Pls add a brief explanation as comments for why the semi join conversion is done at this stage of the planning process after the normal join ordering is done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillSemiJoinRule.java
```diff
@@ -0,0 +1,182 @@
...
+      super(joinClass, aggregateClass, relBuilderFactory, description);
+    }
+  }
+
+  /**
+   * Check for the row schema if they aggregate's output rowtype is different
+   * from its input rowtype then do not convert the join to a semi join.
+   */
+  private static boolean isRowTypeSame(RelNode join, RelNode left, RelNode rightInput) {
+    return join.getRowType().getFieldCount() == left.getRowType().getFieldCount() +
```
amansinha100: I am wondering if this should check for field count == 1 or leave it general ? If we leave it general, then should make sure to test correctness. My preference is to restrict it.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillSemiJoinRule.java
```diff
@@ -0,0 +1,182 @@
...
+import org.apache.calcite.tools.RelBuilderFactory;
+import org.apache.calcite.util.ImmutableBitSet;
+import org.apache.calcite.util.ImmutableIntList;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+
+/**
+ * Planner rule that creates a {@code DrillSemiJoinRule} from a
```
amansinha100: creates a DrillSemiJoinRel (not Rule)

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillSemiJoinRel.java
```diff
@@ -99,6 +104,49 @@ public LogicalOperator implement(DrillImplementor implementor) {
...
+                + " failures in condition " + condition);
+      }
+    }
+    return litmus.succeed();
+  }
+
+  @Override public RelDataType deriveRowType() {
+    return SqlValidatorUtil.deriveJoinRowType(
+            left.getRowType(),
+            right.getRowType(),
```
amansinha100: The right side's row type being projected is a key difference compared to the Calcite's row type, so pls add an explicit comment here for it.

HanumathRao:
@amansinha100  Thank you for the review. I have addressed the review comments. Regarding the multi column predicate testing for semi-join this test case is already added in the previous checkin. It produces the semi-join plan. I have not checked for the correctness of the result in the test case because this is having an issue when assertions are enabled in the runtime code. I will talk to @Ben-Zvi  and open an issue. However, I have checked for the correctness manually and it does produce correct rows.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -277,6 +277,12 @@ protected DrillRel convertToRawDrel(final RelNode relNode) throws SqlUnsupported
         }
       }
 
+      /*Ideally this conversion can be handled during logical planning phase itself but currently the join ordering algorithm in calcite
+        is not considering the semi-joins. This can lead to sub optimal plans. Hence converting the joins to semi-joins post join planning.*/
```
amansinha100: For future reference, pls add the CALCITE-xxx JIRA number for this.

HanumathRao:
@amansinha100  Thank you for the review. I have addressed the comments.

PR #1624, opened 2019-01-27
DRILL-7007: Use verify method in row set tests

paul-rogers:
Many of the early RowSet-based tests used the pattern:

```
    new RowSetComparison(expected)
      .verifyAndClearAll(result);
```

Revise this to use the simplified form:

```
    RowSetUtilities.verify(expected, result);
```

The original form is retained when tests use additional functionality, such as the ability to perform multiple verifications on the same expected batch.


arina-ielchiieva:
+1

PR #1625, opened 2019-01-27
DRILL-7008: Drillbits: clear stale shutdown hooks

vlsi:
`ShutdownThread` is no longer required when `Drillbit#close()` is called.

mvn install for Drill project consumed 600MiB (there were 160 shutdown hooks)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -99,6 +99,7 @@ public class Drillbit implements AutoCloseable {
   private boolean quiescentMode;
   private boolean forcefulShutdown = false;
   private GracefulShutdownThread gracefulShutdownThread;
+  private Thread shutdownHook;
   private boolean interruptPollShutdown = true;
 
   public void setQuiescentMode(boolean quiescentMode) {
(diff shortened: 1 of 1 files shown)


vlsi:
By the way, it improves Travis build time from 42 to 33 minutes:

arina-ielchiieva:
@vlsi thanks for the PR. Could you please create Apache Jira and add it's number in commit message and PR title?

vlsi:
@arina-ielchiieva , done

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -258,6 +260,13 @@ public synchronized void close() {
     }
     final Stopwatch w = Stopwatch.createStarted();
     logger.debug("Shutdown begun.");
+    // We don't really want for Drillbits to pile up in memory, so the hook should be removed
+    // It might be better to use PhantomReferences to cleanup as soon as Drillbit becomes
+    // unreachable, however current approach seems to be good enough.
+    Thread shutdownHook = this.shutdownHook;
+    if (shutdownHook != null && Thread.currentThread() != shutdownHook) {
+      Runtime.getRuntime().removeShutdownHook(shutdownHook);
```
arina-ielchiieva: @vlsi `removeShutdownHook` can throw `IllegalStateException` if the virtual machine is already in the process of shutting down. Maybe we should add try / catch block to ignore such exception?

vlsi: \@arina-ielchiieva , it is handled by  `Thread.currentThread() != shutdownHook`, isn't it?

vlsi: @arina-ielchiieva , I've added try-catch

arina-ielchiieva: Thanks.

arina-ielchiieva:
+1, @vlsi thanks for making the changes and the fact that tests run much faster is awesome.

PR #895, opened 2017-08-04
DRILL-5704: Improve error message on client side when queries fail wi…

sohami:
…th "Failed to create schema tree." when Impersonation is enabled and logins are anonymous

PR #1626, opened 2019-01-28
DRILL-6855: Query from non-existent proxy user fails with "No default schema selected" when impersonation is enabled

aravi5:
This PR intends to throw an error immediately when `UserSession.getDefaultSchema` fails with an `IOException`. Similar changes were done as a part of [DRILL-5704](https://github.com/apache/drill/pull/895). However, with [dynamic loading of schema](https://issues.apache.org/jira/browse/DRILL-5089), the previous fix is no more relevant. This PR tries to resolve the issue on similar lines.

### DIFF
--- a/common/src/main/java/org/apache/drill/common/exceptions/UserExceptionUtils.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/UserExceptionUtils.java
@@ -1,0 +1,27 @@
+package org.apache.drill.common.exceptions;
+
+/**
+ * Provides utilities (such as retrieving hints) to add more context to UserExceptions.
+ */
+public class UserExceptionUtils {
+  public static final String USER_DOES_NOT_EXIST =
+      "Username is absent in connection URL or doesn't exist on Drillbit node." +
+          " Please specify a username in connection URL which is present on Drillbit node.";
+
+  private UserExceptionUtils() {
+    //Restrict instantiation
+  }
+
+  private static String decorateHint(final String text) {
+    return String.format("[Hint: %s]", text);
+  }
+  public static String getUserHint(final Throwable ex) {
+    if (ex.getMessage().startsWith("Error getting user info for current user")) {
+      //User does not exist hint
+      return decorateHint(USER_DOES_NOT_EXIST);
+    } else {
+      //No hint can be provided
+      return "";
+    }
+  }
+}
--- a/exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
+++ b/exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
@@ -25,6 +25,8 @@ import org.apache.calcite.schema.SchemaPlus;
 import org.apache.calcite.schema.impl.AbstractSchema;
 import org.apache.calcite.util.BuiltInMethod;
 import org.apache.drill.common.exceptions.ExecutionSetupException;
+import org.apache.drill.common.exceptions.UserException;
+import org.apache.drill.common.exceptions.UserExceptionUtils;
 import org.apache.drill.exec.planner.sql.SchemaUtilites;
 import org.apache.drill.exec.store.SchemaConfig;
 import org.apache.drill.exec.store.StoragePlugin;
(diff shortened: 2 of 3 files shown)


exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
```diff
@@ -115,8 +116,25 @@ public void loadSchemaFactory(String schemaName, boolean caseSensitive) {
...
+
+      // Improve the error message for client side.
+      final String errorMsg = "Error getting user info for current user";
+      if (iex.getMessage().startsWith(errorMsg)) {
+        final String contextString = "[Hint: Username is absent in connection URL or doesn't " +
+                                         "exist on Drillbit node. Please specify a username in connection " +
+                                         "URL which is present on Drillbit node.]";
+        exceptBuilder.addContext(contextString);
+      }
+      throw exceptBuilder.build(logger);
```
sohami: I am not sure if there is any use case because of which the exceptions were originally consumed and not thrown from here. If not and throwing exception is fine then why not throw `UserException` for all the exceptions caught here ? 
Also my preference will be that this errorMsg check and adding hint message should be put as close to the original thrower of the IOException as possible. Like in this case it will be [here](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemSchemaFactory.java#L86). Reason being then if we want to provide Hints for IOException from different storage plugin then it can be done in each storage plugin rather than having that logic here.

aravi5: I agree with your point that `ExecutionSetupException` should be handled similar to `IOException`. I did not make the changes since I did not fully implications and was trying to limit the impact. May be @arina-ielchiieva / @chunhui-shi could weigh in their thoughts here.

I have added a utility that could provide hints, if any. This way we can have a single place for all hints.

exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
```diff
@@ -117,6 +119,14 @@ public void loadSchemaFactory(String schemaName, boolean caseSensitive) {
       }
     } catch(ExecutionSetupException | IOException ex) {
       logger.warn("Failed to load schema for \"" + schemaName + "\"!", ex);
+      // We can't proceed further without a schema, throw a runtime exception.
+      UserException.Builder exceptBuilder =
+          UserException
+              .resourceError(ex)
```
sohami: should be `executionError` instead of `resourceError`.

common/src/main/java/org/apache/drill/common/exceptions/UserExceptionUtils.java
```diff
@@ -0,0 +1,23 @@
...
+  }
+  public static String getUserHint(final Throwable ex) {
+    if (ex.getMessage().startsWith("Error getting user info for current user")) {
+      //User does not exist hint
+      return decorateHint(USER_DOES_NOT_EXIST);
+    } else {
+      //No hint can be provided
+      return "";
+    }
+  }
```
sohami: Since it's a utility class and contains only static methods/variables so better to make the constructor private.

exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
aravi5: The reason I chose `resourceError` is that schema as a resource is unavailable and it requires a user to externally resolve the issue. Could you elaborate on why this should be `executionError`?

common/src/main/java/org/apache/drill/common/exceptions/UserExceptionUtils.java
aravi5: Sure.

exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
sohami: Had an offline discussion on this that IOException from different plugins is overloading different IOException category. ResourceError being most common category we are going with it for now.

aravi5:
Done.

PR #1638, opened 2019-02-10
DRILL-7022: Partition pruning is not happening the first time after the metadata auto-refresh

vvysotskyi:
For problem description please see [DRILL-7022](https://issues.apache.org/jira/browse/DRILL-7022).

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
@@ -902,6 +902,28 @@ public class TestParquetMetadataCache extends PlanTestBase {
             .run();
   }
 
+  @Test
+  public void testAutoRefreshPartitionPruning() throws Exception {
+    test("create table dfs.tmp.`orders` partition by (o_orderstatus) as\n" +
+        "select * from cp.`tpch/orders.parquet`");
+
+    test("refresh table metadata dfs.tmp.`orders`");
+
+    File ordersTable = new File(dirTestWatcher.getDfsTestTmpDir(), "orders");
+
+    // sets last-modified time of directory greater than the time of cache file to force metadata cache file auto-refresh
+    assertTrue("Unable to change the last-modified time of table directory",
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -902,6 +902,27 @@ public void testEmptyDirectoryWithMetadataDirFile() throws Exception {
...
+    // sets last-modified time of directory greater than the time of cache file to force metadata cache file auto-refresh
+    assertTrue("Unable to change the last-modified time of table directory",
+        ordersTable.setLastModified(new File(ordersTable, Metadata.METADATA_FILENAME).lastModified() + 100500));
+
+    String query = "select * from dfs.tmp.`orders`\n" +
+        "where o_orderstatus = 'O' and o_orderdate < '1995-03-10'";
+    PlanTestBase.testPlanOneExpectedPattern(query, "numRowGroups=1");
+
+    int actualRowCount = testSql(query);
+    assertEquals("Row count does not match the expected value", 1, actualRowCount);
```
vdiravka: It will be good to use `checkForMetadataFile()` method to check that metadata is actually refreshed.
Possibly change `void` return to `File` for that method and compare timestamps of initial metadata file and new one.

vvysotskyi: It's a good idea, but tests may run so fast, that modification time won't change (there is a bug in `lastModified` method where milliseconds are lost: https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8177809), so this check wouldn't work. Adding `Thread.sleep(1000)` will help, but it will make the test much slower. So I leave it as it is.

vdiravka: It makes sense. 
Also it may be reasonable to add the comment to add this functionality once Drill will be on Java 10 or above.

vvysotskyi: Done.

PR #1639, opened 2019-02-11
DRILL-6642: Update protocol-buffers version

agozhiy:
1. Updated protobuf to version 3.6.0
2. Added protobuf to the root pom dependency management
3. Added classes BoundedByteString and LiteralByteString for compatibility with HBase
4. Added ProtobufPatcher to provide compatibility with MapR-DB and HBase


contrib/storage-hbase/src/main/java/com/google/protobuf/LiteralByteString.java
```diff
@@ -0,0 +1,141 @@
...
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
+import java.util.List;
+
+/**
+ * This class is needed for compatibility with recent versions of protobuf.
+ * There was a class with the same name that was package local but in version 3.6.0 it became private.
+ * Contains an instance of {@link ByteString.LiteralByteString} and delegates all method calls to it.
+ */
+class LiteralByteString extends ByteString {
```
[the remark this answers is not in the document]
agozhiy: Class LiteralByteString became inner in version 3.6.0, so it cannot be accessed by the old path.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
```diff
@@ -0,0 +1,140 @@
...
+  private static boolean patched;
+
+  public static synchronized void patch() {
+    if (!patched) {
+      try {
+        patchByteString();
+        patchGeneratedMessageLite();
+        patchGeneratedMessageLiteBuilder();
+        patched = true;
+      } catch (Throwable e) {
```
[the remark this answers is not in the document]
agozhiy: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+
+  /**
+   * HBase client overrides methods from {@link com.google.protobuf.ByteString},
+   * that were made final in version 3.6.0 of protobuf.
+   * This method removes the final modifiers.
+   *
+   * @throws Exception
```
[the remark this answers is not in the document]
agozhiy: Changed the declarations to the concrete exception types, added descriptions.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
```diff
@@ -0,0 +1,140 @@
...
+    // MapR-DB client doesn't use it, but it was added in overridden equals() method.
+    // Adding default implementation.
+    CtMethod cm = cc.getDeclaredMethod("dynamicMethod", new CtClass[] {
+        cp.get("com.google.protobuf.GeneratedMessageLite$MethodToInvoke"),
+        cp.get("java.lang.Object"),
+        cp.get("java.lang.Object")});
+    addImplementation(cm, "if ($1.equals(com.google.protobuf.GeneratedMessageLite.MethodToInvoke.GET_DEFAULT_INSTANCE)) {" +
+                                  "  return this;" +
+                                  "} else {" +
+                                  "  return null;" +
```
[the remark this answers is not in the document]
agozhiy: Cannot throw exception here: this method is still used in several places, but in most cases the returned value is not used.

agozhiy:
@vvysotskyi, I addressed your comments. Also, I regenerated protobufs and modified travis.yml to use version 3.6.0. It will generate protobuf deb package and cache it for later reuse.
Note that the protobufs are still using syntax proto2. Updating to proto3 will require deeper investigation and most likely significant changes in code and approach. I'll create a new JIRA ticked for that.
Could you please review?

agozhiy:
Made ProtobufPatcher fully responsible for compatibility with HBase, removed redundant classes.

contrib/storage-hbase/src/main/java/com/google/protobuf/LiteralByteString.java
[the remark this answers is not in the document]
agozhiy: Used javassist to create the same classes in the old path and inherited them from the current ones.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
```diff
@@ -0,0 +1,183 @@
...
+import javassist.Modifier;
+import javassist.NotFoundException;
+
+public class ProtobufPatcher {
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ProtobufPatcher.class);
+
+  private static boolean patched;
+
+  public static synchronized void patch() {
```
vvysotskyi: Could you please add javadoc for this method?

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
```diff
@@ -0,0 +1,183 @@
...
+    if (!patched) {
+      try {
+        patchByteString();
+        patchGeneratedMessageLite();
+        patchGeneratedMessageLiteBuilder();
+        patched = true;
```
vvysotskyi: I think this flag should be set at the beginning of the block. For the case when we have several calls of this method, and the first attempt to patch was unsuccessful, there would be no more other attempts. Or is it possible that after failure, a method call will work?

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
```diff
@@ -0,0 +1,183 @@
...
+      } catch (Exception e) {
+        logger.warn("Unable to patch Protobuf.", e);
+      }
+    }
+  }
+
+  /**
+   * HBase client overrides methods from {@link com.google.protobuf.ByteString},
+   * that were made final in version 3.6.0 of protobuf.
+   * This method removes the final modifiers. It also returns classes
```
vvysotskyi: Could you please replace rephrase about returning the classes to reflect the fact that this method actually creates and loads classes which were made private nested in the newer version.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
agozhiy: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
agozhiy: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ProtobufPatcher.java
agozhiy: Done.

vvysotskyi:
@agozhiy, looks like there are conflicts in protobuf files. Could you please rebase onto the master and resolve them?

vvysotskyi:
@agozhiy, there is a newer version of protobuf: 3.6.1, is it makes sense to update it in the scope of this PR?

agozhiy:
@vvysotskyi, I updated protobufs to 3.6.1 and resolved the conflicts.

HanumathRao:
@agozhiy can you please rebase on the latest master. Looks like there are conflicts with latest changes that went into master. Thanks.

agozhiy:
Conflicts were resolved.

PR #1640, opened 2019-02-13
DRILL-7038: Queries on partitioned columns scan the entire datasets

KazydubB:
Added new optimizer rule which checks if query references directory columns only and has DISTINCT or GROUP BY operation. If the condition holds, instead of scanning full file set the following will be performed:
- if there is cache metadata file, these directories will be read from it,
- otherwise directories will be gathered from selection object (PartitionLocation).

In the end Scan node will be transformed to DrillValuesRel (containing constant literals) with gathered values so no "scan" will be performed.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
@@ -17,19 +17,31 @@
  */
 package org.apache.drill.exec.planner.logical.partition;
 
+import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 
 import org.apache.drill.exec.util.DrillFileSystemUtil;
 import org.apache.drill.shaded.guava.com.google.common.base.Stopwatch;
 import org.apache.calcite.adapter.enumerable.EnumerableTableScan;
+import org.apache.calcite.plan.RelTraitSet;
+import org.apache.calcite.rel.core.Aggregate;
 import org.apache.calcite.rel.core.Filter;
 import org.apache.calcite.rel.core.Project;
 import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.logical.LogicalValues;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rel.type.RelDataTypeFieldImpl;
+import org.apache.calcite.rel.type.RelRecordType;
 import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.util.BitSets;
 import org.apache.drill.common.expression.ErrorCollectorImpl;
 import org.apache.drill.common.expression.LogicalExpression;
@@ -50,15 +62,20 @@ import org.apache.drill.exec.planner.PartitionDescriptor;
 import org.apache.drill.exec.planner.PartitionLocation;
 import org.apache.drill.exec.planner.logical.DrillOptiq;
 import org.apache.drill.exec.planner.logical.DrillParseContext;
+import org.apache.drill.exec.planner.logical.DrillRel;
+import org.apache.drill.exec.planner.logical.DrillRelFactories;
 import org.apache.drill.exec.planner.logical.DrillScanRel;
 import org.apache.drill.exec.planner.logical.DrillTable;
 import org.apache.drill.exec.planner.logical.DrillTranslatableTable;
+import org.apache.drill.exec.planner.logical.DrillValuesRel;
 import org.apache.drill.exec.planner.logical.RelOptHelper;
 import org.apache.drill.exec.planner.physical.PlannerSettings;
 import org.apache.drill.exec.planner.physical.PrelUtil;
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.VectorContainer;
+import org.apache.drill.exec.store.ColumnExplorer;
 import org.apache.drill.exec.store.StoragePluginOptimizerRule;
+import org.apache.drill.exec.store.dfs.FileSelection;
 import org.apache.drill.exec.store.dfs.FormatSelection;
 import org.apache.drill.exec.store.dfs.MetadataContext;
 import org.apache.drill.exec.store.dfs.MetadataContext.PruneStatus;
@@ -70,9 +87,6 @@ import org.apache.calcite.plan.RelOptRuleOperand;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.rex.RexNode;
 import org.apache.commons.lang3.tuple.Pair;
-
-import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
-import org.apache.drill.shaded.guava.com.google.common.collect.Maps;
 
 import org.apache.drill.exec.vector.ValueVector;
 import org.apache.hadoop.fs.Path;
@@ -144,6 +158,10 @@ public abstract class PruneScanRule extends StoragePluginOptimizerRule {
     return new DirPruneScanFilterOnScanRule(optimizerRulesContext);
   }
 
+  public static RelOptRule getConvertAggScanToValuesRule(OptimizerRulesContext optimizerRulesContext) {
+    return new ConvertAggScanToValuesRule(optimizerRulesContext);
+  }
+
   protected void doOnMatch(RelOptRuleCall call, Filter filterRel, Project projectRel, TableScan scanRel) {
 
     final String pruningClassName = getClass().getName();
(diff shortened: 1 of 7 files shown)


priteshm:
@gparai @HanumathRao can you'll review?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
```diff
@@ -550,4 +567,180 @@ private static void setPruneStatus(MetadataContext metaContext, PruneStatus prun
...
+
+    private List<String> getValues(String selectionRoot, List<String> directories, List<Integer> indexes) {
+      List<String> values = new ArrayList<>();
+      for (String dir : directories) {
+        List<String> parts = ColumnExplorer.listPartitionValuesDirectory(dir, selectionRoot);
+        for (int index : indexes) {
+          if (index < parts.size()) {
+            values.add(parts.get(index));
+          } else {
+            values.add(null);
```
gparai: Why do we add null here?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
```diff
@@ -550,4 +567,180 @@ private static void setPruneStatus(MetadataContext metaContext, PruneStatus prun
...
+        if (!dirPattern.matcher(field).matches()) {
+          return false;
+        }
+      }
+
+      return scan.isDistinct() || aggregate.getGroupCount() > 0;
+    }
+
+    @Override
+    public void onMatch(RelOptRuleCall call) {
```
gparai: Please add a comment describing what this transformation does maybe with an example.

KazydubB:
@gparai thank you, added comments.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
```diff
@@ -550,4 +567,211 @@ private static void setPruneStatus(MetadataContext metaContext, PruneStatus prun
...
+      } finally {
+        if (totalPruningTime != null) {
+          logger.debug("Total pruning elapsed time: {} ms", totalPruningTime.elapsed(TimeUnit.MILLISECONDS));
+        }
+      }
+    }
+
+    private List<String> getValues(String selectionRoot, List<String> directories, List<Integer> indexes) {
+      List<String> values = new ArrayList<>();
+      for (String dir : directories) {
```
vdiravka: Consider 
```suggestion
    directories.stream()
        .map(dir -> ColumnExplorer.listPartitionValuesDirectory(dir, selectionRoot))
        .forEachOrdered(parts -> indexes.stream()
            .map(index -> index < parts.size() ? parts.get(index) : null)
            .forEach(values::add));
```
```

File: exec/java-exec/src/main/java/org/apache/drill/exec/store/ColumnExplorer.java [hunk 4]
```diff
@@ -224,6 +224,33 @@ public static boolean isPartitionColumn(String partitionDesignator, String path)
     return Arrays.asList(diffDirectoryNames);
   }
 
+  public static List<String> listPartitionValuesDirectory(String dirPath, String root) {
```
vdiravka: It copies 90% of `listPartitionValues()` method. Could you just pass some boolean flag to the old method, for example   `hasDirsOnly` as [here](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/planner/DFSFilePartitionLocation.java#L29)?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
```diff
@@ -550,4 +567,211 @@ private static void setPruneStatus(MetadataContext metaContext, PruneStatus prun
...
+          }
+        }
+      }
+
+      return values;
+    }
+
+    private List<String> getValues(List<PartitionLocation> partitions, List<Integer> indexes) {
+      List<String> values = new ArrayList<>(partitions.size() * indexes.size());
+      for (PartitionLocation partition : partitions) {
```
vdiravka: Why do not use Stream API?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
KazydubB: Added comment.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/partition/PruneScanRule.java
KazydubB: Thanks for suggestion but I find it harder to read.

exec/java-exec/src/test/java/org/apache/drill/TestPartitionPruning.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @BeforeClass
+  public static void setUp() {
+    dirTestWatcher.copyResourceToRoot(Paths.get("multilevel"));
+    dirTestWatcher.copyResourceToRoot(Paths.get("multilevel/parquet"), Paths.get(TABLE_WITH_METADATA));
+  }
+
+  @Test
+  public void testPartitionPruning() throws Exception {
+    String[] tableNames = {"multilevel/parquet", "multilevel/json", "multilevel/csv"};
```
vdiravka: Interesting that we had not Partition Pruning for the CSV or JSON formats earlier.
Could elaborate how to update Drill doc ["Partition Pruning Introduction"](https://drill.apache.org/docs/partition-pruning-introduction/) in accordance to your improvements?

KazydubB:
@vdiravka currently we can select partition columns for tables consisting of many directories containing JSON/csv as well. There is no need to update documentation as this feature does not add new functionality but only optimizes existing one. I see that the confusion may arise from bad naming - hence I renamed the test class with its methods and also renamed the rule class.

Resolved the conflicts. Please, take a look once more.

gparai:
LGTM +1

exec/java-exec/src/test/java/org/apache/drill/TestPartitionColumnScanPruning.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.apache.drill.categories.PlannerTest;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import java.nio.file.Paths;
+
+@Category({PlannerTest.class})
+public class TestPartitionColumnScanPruning extends PlanTestBase {
```
vdiravka: Please add also the test case for Hive table with partitions, for instance one test case in the `TestHiveStorage`

exec/java-exec/src/test/java/org/apache/drill/TestPartitionColumnScanPruning.java
```diff
@@ -0,0 +1,73 @@
...
+  private static final String TABLE_WITH_METADATA = "parquetTable1";
+
+  @BeforeClass
+  public static void setUp() {
+    dirTestWatcher.copyResourceToRoot(Paths.get("multilevel"));
+    dirTestWatcher.copyResourceToRoot(Paths.get("multilevel/parquet"), Paths.get(TABLE_WITH_METADATA));
+  }
+
+  @Test
+  public void testPartitionColumnScanPruning() throws Exception {
```
vdiravka:
```suggestion
  public void testDirScanToValuesConverting() throws Exception {
```

exec/java-exec/src/test/java/org/apache/drill/TestPartitionColumnScanPruning.java
```diff
@@ -0,0 +1,73 @@
...
+    };
+    for (String tableName : tableNames) {
+      for (String query : queries) {
+        testPlanMatchingPatterns(String.format(query, tableName), new String[]{"Values\\(tuples="}, new String[]{"Scan\\(table="});
+      }
+    }
+  }
+
+  @Test
+  public void testPartitionColumnScanPruningWithMetadataCache() throws Exception {
```
vdiravka:
```suggestion
  public void testDirScanToValuesConvertingWithMetadataCache() throws Exception {
```

exec/java-exec/src/test/java/org/apache/drill/TestPartitionColumnScanPruning.java
```diff
@@ -0,0 +1,73 @@
...
+    test("refresh table metadata dfs.`%s`", TABLE_WITH_METADATA);
+    checkForMetadataFile(TABLE_WITH_METADATA);
+    String query = String.format("select distinct dir0, dir1 from dfs.`%s`", TABLE_WITH_METADATA);
+    PlanTestBase.testPlanMatchingPatterns(query, new String[]{"Values\\(tuples="}, null);
+  }
+
+  @Test
+  public void testPartitionColumnScanPruningIsNotApplied() throws Exception {
```
vdiravka:
```suggestion
  public void testDirScanToValuesConvertingIsNotApplied() throws Exception {
```

exec/java-exec/src/test/java/org/apache/drill/TestPartitionColumnScanPruning.java
vdiravka:
```suggestion
public class TestDirScanToValuesConverting extends PlanTestBase {
```

KazydubB: This change targets file system databases only. Hive tables do not support querying of partition columns (dir0, dir1 etc.).

exec/java-exec/src/test/java/org/apache/drill/TestDirScanToValuesConverting.java
```diff
@@ -0,0 +1,73 @@
...
+
+import org.apache.drill.categories.PlannerTest;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import java.nio.file.Paths;
+
+@Category({PlannerTest.class})
+public class TestDirScanToValuesConverting extends PlanTestBase {
```
vdiravka: Please make class name to be consistent with test names.
```suggestion
public class TestDirScanToValuesConversion extends PlanTestBase {
```

PR #1641, opened 2019-02-14
DRILL-4858: REPEATED_COUNT on an array of maps and an array of arrays is not implemented

KazydubB:
- Implemented 'repeated_count' function for repeated MAP and repeated LIST;
- Updated RepeatedListReader and RepeatedMapReader implementations to return correct value from size() method.


exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedListReaderImpl.java
```diff
@@ -80,7 +79,10 @@ public void reset() {
 
   @Override
   public int size() {
-    return maxOffset - currentOffset;
+    if (currentOffset == NO_VALUES) {
+      return 0;
+    }
+    return maxOffset - currentOffset - 1;
```
vdiravka: Consider:
`return currentOffset == NO_VALUES ? 0 : maxOffset - currentOffset - 1;`

vdiravka: Introduce also `isNull()` method like in `RepeatedMapReaderImpl` class

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
```diff
@@ -17,36 +17,161 @@
...
+          "[1, 2, 3], [4, 5], [6], [7], [8, 9], [2, 3], [2, 3], [2, 3], [2]",
+          "[1, 2], [3], [4], [5]",
+          "",
+          "[1], [2], [3]"};
+      int elementId = 1;
+      for (String value : arrayElements) {
+        String entry = String.format("{\"id\": %d, \"array\": [%s]}\n", elementId++, value);
+        writer.write(entry);
+      }
+    }
```
vdiravka: Why do not use datasource instead of creating file in the test?

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
```diff
@@ -17,36 +17,161 @@
...
-       .baselineValues(false)
-       .build().run();
+    testBuilder()
+        .sqlQuery("select repeated_contains(topping, 'Choc*') from cp.`testRepeatedWrite.json`")
+        .ordered()
+        .baselineColumns("EXPR$0")
+        .baselineValuesForSingleColumn(true, true, true, true, false)
+        .go();
+
+    testBuilder()
```
vdiravka: It is better to divide these big tests into smaller one. Usually one `testBuilder` per test case.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SimpleRepeatedFunctions.java
```diff
@@ -77,8 +79,6 @@ public void eval() {
     }
   }
 
-  /*
-  // TODO - need to confirm that these work   SMP: They do not
```
vdiravka: What about TODO?
> // TODO - replace with a freemarker template and fill out the rest of the types
>   // focused on getting functions defined for JSON types as this is the primary format
>   // users are extracting repeated data out of currently

May it be used for other datatypes?

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
```diff
@@ -17,36 +17,161 @@
...
+      {"id": 5, "array": [[1, 2, 3], [4, 5], [6], [7], [8, 9], [2, 3], [2, 3], [2, 3], [2]]}
+      {"id": 6, "array": [[1, 2], [3], [4], [5]]}
+      {"id": 7, "array": []}
+      {"id": 8, "array": [[1], [2], [3]]}
+    */
+    String fileName = "repeated_count_list.json";
+    try (BufferedWriter writer = new BufferedWriter(new FileWriter(
+        new File(dirTestWatcher.getRootDir(), fileName)))) {
+      String[] arrayElements = {
+          "[1, 2], [1, 3], [2, 3]",
```
vdiravka: Use `TestBuilder.listOf()` to create a `JsonStringArrayList` from the given values.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
```diff
@@ -17,36 +17,161 @@
...
+      {"mapArray": []}
+      {"mapArray": [{"field1": 1, "field2": "val3"}]}
+      {"mapArray": [{"field1": 1, "field2": "val4"}, {"field1": 2}, {"field1": 2}, {"field2": "val1"}, {"field1": 2}]}
+      {"mapArray": [{"field1": 1, "field2": "val3"}, {"field1": 2}]}
+    */
+    String fileName = "repeated_count_map.json";
+    try (BufferedWriter writer = new BufferedWriter(new FileWriter(
+        new File(dirTestWatcher.getRootDir(), fileName)))) {
+      String[] arrayElements = {
+          "{\"field1\": 1, \"field2\": \"val1\"}, {\"field1\": 2}",
```
vdiravka: Use `TestBuilder.mapOf()` to create a `JsonStringHashMap` from the given values.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SimpleRepeatedFunctions.java
KazydubB: Not sure what you mean. Do you mean to create a template instead as the TODO suggests?

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedListReaderImpl.java
ihuzenko: In my opinion usage ternary or full conditional construct doesn't impact code readability here. I'm confused by unclear naming of the fields and necessity to use 'magic number' along with the fields. @KazydubB  could you please rename or provide descriptive javadocs and constant for 1 to make offsets usage look clear ?

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SimpleRepeatedFunctions.java
vdiravka: > Not sure what you mean. Do you mean to create a template instead as the TODO suggests?

Yes. But I am not sure that it will save much more space for current `repeated_count` functions, since there are different implementations of `eval()` method. But if some additional data types can be used with this function, then it makes sense to add them and to consider rewriting this calss with FreeMarker template.
So can other data types be used with `repeated_count` function?

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedListReaderImpl.java
vdiravka: @KazydubB It will be good to have it for consistency with `ReapetedMapReaderImpl`.
You can use parentheses. 
`return isNull() ? 0 : (maxOffset - currentOffset - 1);`
The final decision is yours

@ihuzenko Good note. I think it will be good to have a comment, why `-1` is need.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
```diff
@@ -24,154 +24,72 @@
...
 
-import java.io.BufferedWriter;
-import java.io.File;
-import java.io.FileWriter;
+import java.nio.file.Paths;
 
 @Category(SqlFunctionTest.class)
 public class TestNewSimpleRepeatedFunctions extends ClusterTest {
 
+  private static final String SELECT_REPEATED_CONTAINS = "select repeated_contains(topping, '%s*') from cp.`testRepeatedWrite.json`";
```
vdiravka: It can be used as a local variable for `testRepeatedContainsForWildCards` test

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
```diff
@@ -24,154 +24,72 @@
...
 
-    testBuilder()
-        .sqlQuery(selectQuery + " where repeated_count(array) > 4", fileName)
-        .ordered()
-        .baselineColumns("EXPR$0")
-        .baselineValuesForSingleColumn(5, 9)
-        .go();
+  @Test
+  public void testRepeatedCountRepeatedListInHaving() throws Exception {
+    String query = SELECT_REPEATED_COUNT_LIST + " group by 1 having repeated_count(array) < 4";
```
vdiravka: There is overloaded `sqlQuery(String query, Object... replacements)` method.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
```diff
@@ -24,154 +24,72 @@
...
+    performTest(query, 3, 0, 2);
+  }
 
-    testBuilder()
-        .sqlQuery(selectQuery + " group by 1", fileName)
-        .unOrdered()
-        .baselineColumns("EXPR$0")
-        .baselineValuesForSingleColumn(3, 0, 2, 5, 9, 4)
-        .go();
+  private void performTest(String query, Object... expectedValues) throws Exception {
```
vdiravka: I see why you added it, but tests are worse readable with this method. It is necessary to see the implementation of this method to understand the tests.
I suggest you to use regular `testBuilder` here, but if you want to proceed with this, please add it to the `ClusterTest` class.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/impl/SimpleRepeatedFunctions.java
KazydubB: Moved `repeated_count` functions with uniform implementation to a separate freemarker template file. Thanks for pointing the `TODO` out :+1:

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
KazydubB: Not sure why this should be used - I find the concatenation fine. If you have any objections, please feel free to point it out.
Just to be sure, do you mean to have the base query with `%s` so the additional part should be formatted (well, passed to the method as the second parameter)?

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedListReaderImpl.java
KazydubB: I agree that this is not clear why the `-1` is added, so decided to decrement `maxOffset` value by `1` for the count to be an expected statement `maxOffset - currentOffset`.

Also did add a method to `RepeatedListReaderImpl` but named it `isEmpty()` (instead of `isNull()`) as it appears to fit better.

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
KazydubB: Removed the method.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
```diff
@@ -479,6 +479,14 @@ private HoldingContainer visitValueVectorReadExpression(ValueVectorReadExpressio
 
       if (!hasReadPath && !complex) {
         JBlock eval = new JBlock();
+
+        if (repeated) {
```
vdiravka: Was it the bug?
Am I right, we just didn't reach this part of code with other functions with `Repeated` DataMode?

exec/java-exec/src/test/java/org/apache/drill/exec/fn/impl/TestNewSimpleRepeatedFunctions.java
vdiravka: agree, it will be more complicated with `String.format`

exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
KazydubB: It was kind of a bug. Yes, this code wasn't reached by other functions (at least those I am aware of).

KazydubB:
Thanks, @vdiravka. I've updated Jira description and squashed the commits.

PR #1642, opened 2019-02-17
DRILL-6734: JDBC storage plugin returns null for fields without aliases

vvysotskyi:
- Added changes to pass output column names taken from row type to JdbcRecordReader and use them for storing the results since column names in the result set may differ when aliases aren't specified and it causes problems for Drill operators like Project which rely on field names during the execution.

For problem description please see [DRILL-6734](https://issues.apache.org/jira/browse/DRILL-6734).

### DIFF
--- a/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcRecordReader.java
+++ b/contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcRecordReader.java
@@ -28,12 +28,12 @@ import java.sql.Statement;
 import java.sql.Time;
 import java.sql.Timestamp;
 import java.util.Calendar;
+import java.util.List;
 import java.util.TimeZone;
 
 import javax.sql.DataSource;
 
 import org.apache.drill.common.AutoCloseables;
-import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.types.TypeProtos;
 import org.apache.drill.common.types.TypeProtos.MajorType;
@@ -63,7 +63,7 @@ import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableMap;
 
 @SuppressWarnings("unchecked")
 class JdbcRecordReader extends AbstractRecordReader {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory
       .getLogger(JdbcRecordReader.class);
 
   private static final ImmutableMap<Integer, MinorType> JDBC_TYPE_MAPPINGS;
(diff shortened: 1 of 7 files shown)


contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcRecordReader.java
```diff
@@ -180,22 +182,34 @@ private static String nameFromType(int javaSqlType) {
...
       ImmutableList.Builder<Copier<?>> copierBuilder = ImmutableList.builder();
 
-      for (int i = 1; i <= columns; i++) {
-        final String name = meta.getColumnLabel(i);
-        final int jdbcType = meta.getColumnType(i);
-        final int width = meta.getPrecision(i);
-        final int scale = meta.getScale(i);
+      for (int i = 1; i <= columnsCount; i++) {
+        String name = columns.get(i - 1);
+        int jdbcType = meta.getColumnType(i);
```
vdiravka: Please put the comment here that column index in `ResultSetMetaData` starts from 1

contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcRecordReader.java
```diff
@@ -245,13 +259,11 @@ public void setup(OperatorContext operatorContext, OutputMutator output) throws
   @Override
   public int next() {
     int counter = 0;
-    Boolean b = true;
```
vdiravka: :1st_place_medal:

contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithH2IT.java
```diff
@@ -63,7 +63,9 @@ public static void initH2() throws Exception {
       URL scriptFile = TestJdbcPluginWithH2IT.class.getClassLoader().getResource("h2-test-data.sql");
       Assert.assertNotNull("Script for test tables generation 'h2-test-data.sql' " +
           "cannot be found in test resources", scriptFile);
-      RunScript.execute(connection, new FileReader(scriptFile.getFile()));
+      try (FileReader fileReader = new FileReader(scriptFile.getFile())) {
```
vdiravka: :+1:

contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithMySQLIT.java
```diff
@@ -252,4 +252,46 @@ public void testCaseSensitiveTableNames() throws Exception {
...
+    testBuilder()
+        .sqlQuery(query)
+        .unOrdered()
+        .baselineColumns("person_id", "person_id0")
+        .baselineValues(1, 1)
+        .baselineValues(2, 2)
+        .baselineValues(3, 3)
+        .baselineValues(5, 5)
+        .go();
+  }
```
vdiravka: What do you think about some extra test case with aliases? Something similar to:
 ``select person_id as ID from mysql.`drill_mysql_test`.person limit 2``

contrib/storage-jdbc/src/main/java/org/apache/drill/exec/store/jdbc/JdbcRecordReader.java
vvysotskyi: Thanks, done.

contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithMySQLIT.java
vvysotskyi: Added test with a query which has aliases for column reference and for expressions.

vdiravka:
+1, LGTM

PR #1646, opened 2019-02-20
DRILL-6852: Adapt current Parquet Metadata cache implementation to use Drill Metastore API

vvysotskyi:
The goal for these changes is to allow usage of optimizations produced for parquet for other storage formats with metadata usage.

In this PR new classes for using all metadata were introduced: TableMetadata, PartitionMetadata, FileMetadata, and RowGroupMetadata. These classes have API to provide a schema, statistics for metadata level in general or for specific columns.

AbstractGroupScanWithMetadata was introduced which contains base logic for handling metadata at table, partition or file level. It contains base methods for filtering and applying the limit to the group scan. Existing AbstractParquetGroupScan class was made the inheritor of AbstractGroupScanWithMetadata and reworked to use new metadata classes.

The logic for populating HiveDrillNativeParquetScan and ParquetGroupScan instances was moved into BaseParquetTableMetadataProvider inheritors, so it allows to replace specific TableMetadataProvider by providers which fetches metadata from the metastore or other sources in the future.

Also, methods to convert existing parquet statistics in BaseParquetTableMetadataProvider and in ParquetTableMetadataUtils were added to use new metadata classes.

Classes which contains logic for filtering statistics (ParquetRGFilterEvaluator, ParquetFilterPredicate implementations and other) were reworked to use new metadata classes, so there is no dependency on parquet-specific things there.


vvysotskyi:
@amansinha100, @dvjyothsna, @Ben-Zvi I have created draft PR for adapting Parquet Metadata cache implementation to use Drill Metastore API. Could you please do a cursory review?

Ben-Zvi:
@vvysotskyi - This is quite a sizable PR. Can you write some overview of the key changes and reasons behind them to guide our review ?

exec/java-exec/src/main/java/org/apache/drill/metastore/ColumnStatistics.java
```diff
@@ -0,0 +1,73 @@
...
+package org.apache.drill.metastore;
+
+import java.util.Comparator;
+
+/**
+ * Represents collection of statistics values for specific column.
+ *
+ * @param <T> type of column values
+ */
+public interface ColumnStatistics<T> {
```
amansinha100: Suggest naming this as ExactColumnStatistics and related implementors of this interface accordingly.  The reason is that we want to avoid confusion with **estimated** column statistics. 

cc @gparai

exec/java-exec/src/main/java/org/apache/drill/metastore/Metastore.java
```diff
@@ -0,0 +1,46 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.metastore;
+
+import java.util.List;
+
+public interface Metastore {
```
amansinha100: Pls add javadoc for all the interface APIs.

exec/java-exec/src/main/java/org/apache/drill/metastore/TableMetadata.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.drill.exec.record.metadata.TupleSchema;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Metadata which corresponds to the table level.
+ */
+public class TableMetadata implements BaseMetadata {
```
amansinha100: Is TableMetadata eventually the object that will be published to INFORMATION_SCHEMA ?  Should it have JSON serializers and deserializers ?

exec/java-exec/src/main/java/org/apache/drill/metastore/TableMetadata.java
```diff
@@ -0,0 +1,124 @@
...
+  private final String tableName;
+  private final String location;
+  private final TupleSchema schema;
+  private final Map<SchemaPath, ColumnStatistics> columnsStatistics;
+  private final Map<String, Object> tableStatistics;
+  private final long lastModifiedTime;
+  private final String owner;
+  private final Set<String> partitionKeys;
```
amansinha100: Conceptually, the fact that partitionKeys belongs here makes me think what about other physical attributes of a table ?     Such as primary keys,  indexing columns etc.  However, a distinction is that here TableMetadata is only targeting file system tables, not tables that are native to other storage/format plugins.  For example, MapR-DB tables will have their own version of TableMetadata which may not match with this class definition.   Are you thinking that TableMetadata will be the universal table metadata for **ALL** types of tables regardless of their data source ?

exec/java-exec/src/main/java/org/apache/drill/metastore/expr/FilterPredicate.java
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.metastore.expr;
+
+import org.apache.drill.exec.expr.stat.RowsMatch;
+
+public interface FilterPredicate<T extends Comparable<T>> {
```
amansinha100: When creating these interfaces regarding Filters, let's think of them being shared between both planning time and run-time because we want to allow for run-time pruning for parquet row groups.  During run-time, I would think there is no reason to go through the Metastore.  

cc @Ben-Zvi

exec/java-exec/src/main/java/org/apache/drill/metastore/expr/StatisticsConstants.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.metastore.expr;
+
+public interface StatisticsConstants {
```
amansinha100: Similar to the prior comment about 'Exact' statistics, let's start differentiating between these and the estimated statistics.  @gparai  I believe you have a similar set of constants for the estimated stats such as NDV, NonNullRowCount etc.    It would be better to be explicit by prefixing 'ExactStatistics' vs. 'EstimatedStatistics' .  Otherwise there's going to be confusion.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushProjectIntoScanRule.java
```diff
@@ -61,12 +61,12 @@ protected boolean skipScanConversion(RelDataType projectRelDataType, TableScan s
   public static final RelOptRule DRILL_LOGICAL_INSTANCE =
       new DrillPushProjectIntoScanRule(LogicalProject.class,
           DrillScanRel.class,
-          "DrillPushProjIntoDrillRelScan");
+          "DrillPushProjectIntoScanRule");
```
dvjyothsna: Should this be DrillPushProjectIntoDrillRelScan?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -533,36 +536,27 @@ private ParquetFileMetadata_v3 getParquetFileMetadata_v3(ParquetTableMetadata_v3
...
+      module.addSerializer(ColumnMetadata_v3.class, new ColumnMetadata_v3.Serializer());
+    }
     mapper.registerModule(module);
     OutputStream os = fs.create(p);
     mapper.writerWithDefaultPrettyPrinter().writeValue(os, parquetTableMetadata);
     os.flush();
     os.close();
   }
 
-  private void writeFile(ParquetTableMetadataDirs parquetTableMetadataDirs, Path p, FileSystem fs) throws IOException {
```
dvjyothsna: Is this moved to some other location?

vvysotskyi: This method contains code almost the same as the previous one, so it was removed and the previous one is used.

exec/java-exec/src/main/java/org/apache/drill/metastore/expr/StatisticsConstants.java
vvysotskyi: Agree, thanks, will rename it to `ExactStatisticsConstants`.

exec/java-exec/src/main/java/org/apache/drill/metastore/expr/FilterPredicate.java
vvysotskyi: Agree that this interface may be used at run-time, will move it to another package. Usage of this interface assumes using `ColumnStatistics` classes which may be populated by the statistics from table footers.

exec/java-exec/src/main/java/org/apache/drill/metastore/TableMetadata.java
vvysotskyi: Changes in this PR does not assume that this class will be serialized/deserialized. Currently, information to populate this class is taken from parquet metadata.

exec/java-exec/src/main/java/org/apache/drill/metastore/TableMetadata.java
vvysotskyi: Agree, will create the interface with name `TableMetadata`, move base methods there and rename this class into `FileTableMetadata`.

exec/java-exec/src/main/java/org/apache/drill/metastore/Metastore.java
vvysotskyi: This interface was added at the beginning, but it wasn't used yet. I will remove it from current PR and will add with further changes where it will be used. I will add javadocs to other places where it is absent.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillPushProjectIntoScanRule.java
vvysotskyi: No, it is useful to have a description of the rule which contains the name of the class to find the rule reading the logs for example. So I will change it to `DrillPushProjectIntoScanRule:logical`.

dvjyothsna: Yes adding the word "logical" will help differentiate between logical and physical.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
```diff
@@ -0,0 +1,734 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * Represents table group scan with metadata usage.
+ */
+public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupScan {
+
+  protected TableMetadataProvider metadataProvider;
```
amansinha100: At first, I was thinking that a TableMetadataProvider would provide table metadata for multiple tables (the term 'provider' seems to imply that).  However, here each table has its own provider which generates the corresponding TableMetadata.  One thing we should measure is the impact on the heap usage during planning time based on these  changes by examining a somewhat complex query (e.g a 10 table join from TPC-DS).   Previously, each table would have 1 ParquetGroupScan and within that a list of RowGroupInfo objects plus other related objects.   With this PR, the additional objects would be TableMetadataProvider, TableMetadata, list of FileMetadata and list of PartitionMetadata (if partition is present).  The row group level things are bundled into RowGroupMetadata.  We want to move towards a 'lean planner' where complex query plan generation can be done very quickly, so let's keep that requirement in mind.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Can the name parquetTableMetadata be changed to something generic because this is used even to write the list of directories.

vvysotskyi:
Here is a class diagram for some classes from this PR: https://docs.google.com/presentation/d/1oXw3f77i_CQzdwA0Nu6PW4FX8uqYRds3nw7isBaOyJA/edit?usp=sharing

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
vvysotskyi: Agree, renamed to `parquetMetadata`. Its Javadoc clarifies that it can be table or directory metadata

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
vvysotskyi: @amansinha100, I have compared the heap usage for several TPC-DS queries on sf100 data but I didn't see a significant difference. I used VisualVM to see heap usage.

I agree with you that it may cause degradation for larger tables and currently, I'm reworking `TableMetadataProvider` to store fewer objects.

amansinha100: @vvysotskyi thanks for checking the heap usage.  I will wait for your additional changes.

vvysotskyi:
@amansinha100, @dvjyothsna, @Ben-Zvi, this PR is ready for the review, could you please take a look?

exec/java-exec/src/main/java/org/apache/drill/metastore/expr/FilterPredicate.java
amansinha100: Should this be made independent of metastore ?

exec/java-exec/src/main/java/org/apache/drill/metastore/BaseMetadata.java
```diff
@@ -0,0 +1,80 @@
...
+package org.apache.drill.metastore;
+
+import org.apache.drill.common.expression.SchemaPath;
+import org.apache.drill.exec.record.metadata.ColumnMetadata;
+import org.apache.drill.exec.record.metadata.TupleSchema;
+
+import java.util.Map;
+
+/**
+ * Common statistics provider for table, partition, file or row group.
```
amansinha100: Provider for not just statistics but also Tuple Schema and ColumnMetadata

vvysotskyi:
@amansinha100, thanks for the review, I have addressed CR comments.
Here is the results of 10 consequent runs of `TestTpchPlanning` test (was used `mvn test -Dtest=org.apache.drill.TestTpchPlanning -pl exec/java-exec` command to run the test)

  | With changes | Without changes
-- | -- | --
1 | 42,82 | 25,971
2 | 24,475 | 24,904
3 | 24,621 | 27,618
4 | 24,345 | 26,197
5 | 24,699 | 24,624
6 | 27,288 | 28,193
7 | 24,34 | 25,152
8 | 25,273 | 25,432
9 | 24,367 | 26,141
10 | 24,418 | 25,121

image

In averaege, there is no degradation, the first deviation may be connected with env.

amansinha100:
Thanks for capturing the timings.  These LGTM.  +1

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetGroupScan.java
```diff
@@ -219,229 +200,176 @@ public ScanStats getScanStats() {
...
-          implicitColValues);
+    RowGroupScanFilterer builder = getFilterer().getFiltered(optionManager, filterPredicate, schemaPathsInExpr);
 
-      Map<SchemaPath, ColumnStatistics> columnStatisticsMap = statCollector.collectColStat(schemaPathsInExpr);
-
-      ParquetFilterPredicate.RowsMatch match = ParquetRGFilterEvaluator.matches(filterPredicate,
-          columnStatisticsMap, rowGroup.getRowCount(), parquetTableMetadata, rowGroup.getColumns(), schemaPathsInExpr);
-      if (match == ParquetFilterPredicate.RowsMatch.NONE) {
-        continue; // No row comply to the filter => drop the row group
+    if (getRowGroupsMetadata() != null) {
```
Ben-Zvi: Looks like _getRowGroupsMetadata()_ will never return a null. So are all the "else" clauses below dead code ?

sohami:
@vvysotskyi  - I am removing ready-to-commit label for this PR for now since there is an open question from @Ben-Zvi. Once that is resolved please place the label back.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetGroupScan.java
vvysotskyi: Good question! In the current code, it cannot be null since `ParquetMetadataProvider` cannot return null when calling `getRowGroupsMeta()` method, but in future will be provided other implementations of `TableMetadataProvider` which can return null for example for the case when the number of row groups is pretty large. So all the "else" will be suitable for such a case.

Ben-Zvi: Thanks @vvysotskyi ; at some point we could add your answer in a comment there.

vvysotskyi: I have made some changes, so now for the case when metadata is not available, and it may be returned as a collection, an empty collection will be returned instead of null to reduce calls into `TableMetadataProvider`. Added the corresponding comment into `TableMetadataProvider` and replaced unneeded null checks by `!.isEmpty()` check in the last commit.

vvysotskyi:
@Ben-Zvi, can this PR be merged, or something else should be done?

vvysotskyi:
Rebased onto the master and squashed commits.

amansinha100:
Since this is good to go, I will mark the JIRA as ready-to-commit.

PR #1650, opened 2019-02-22
DRILL-6707: Join Batch Memory Manager to resize current outgoing batch no larger than current batch size

Ben-Zvi:
The source of the bug was retaining a checkArgument, which originally did a sanity check that 64K (the old batch size) was not exceeded.  When batch resizing was implemented for Merge-Join, that checkArgument was kept (with the new target size).
    So when the batch was resized while the outgoing already had more rows than the new target, the next check for a full outgoing hit that checkArgument and caused this failure.

The fix: Remove that checkArgument.  (The other code changes are for debug or cosmetics).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
@@ -125,7 +125,7 @@ public class MergeJoinBatch extends AbstractBinaryRecordBatch<MergeJoinPOP> {
     @Override
     public void update(int inputIndex) {
       super.update(inputIndex, status.getOutPosition());
-      status.setTargetOutputRowCount(super.getCurrentOutgoingTargetOutputRowCount());
+      status.setTargetOutputRowCount(super.getCurrentOutgoingMaxRowCount());
       RecordBatchIOType type = inputIndex == 0 ? RecordBatchIOType.INPUT_LEFT : RecordBatchIOType.INPUT_RIGHT;
       RecordBatchStats.logRecordBatchStats(type, getRecordBatchSizer(inputIndex), getRecordBatchStatsContext());
     }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
@@ -125,7 +125,7 @@ public class MergeJoinBatch extends AbstractBinaryRecordBatch<MergeJoinPOP> {
     @Override
     public void update(int inputIndex) {
       super.update(inputIndex, status.getOutPosition());
-      status.setTargetOutputRowCount(super.getCurrentOutgoingMaxRowCount());
+      status.setTargetOutputRowCount(super.getCurrentOutgoingMaxRowCount()); // calculated by update()
       RecordBatchIOType type = inputIndex == 0 ? RecordBatchIOType.INPUT_LEFT : RecordBatchIOType.INPUT_RIGHT;
       RecordBatchStats.logRecordBatchStats(type, getRecordBatchSizer(inputIndex), getRecordBatchStatsContext());
     }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
@@ -124,7 +124,8 @@ public class MergeJoinBatch extends AbstractBinaryRecordBatch<MergeJoinPOP> {
      */
     @Override
     public void update(int inputIndex) {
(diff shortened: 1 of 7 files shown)


Ben-Zvi:
And here is an example of the debug message (where before the failure happened):
```
2019-02-21 15:58:16,871 [2390c58b-2a69-0416-80cb-0802eca8b730:frag:4:0] DEBUG o.a.d.e.p.impl.join.MergeJoinBatch - Update outgoing batch to size 16383, which is smaller than the current batch size of 16426
```

Ben-Zvi:
@sohami - an updated version was pushed - The original Merge-Join check was restored, and the two changes you proposed are in: `adjustOutputRowCount()` now rounds to the NEXT power of two, and the `updateInternal()` caps the returned result at the current batch size (so no need to reallocate, etc.)
  The other code changes are just debug and cosmetics.
  Note that two unit test files had failures due to the "rounding up": `testUnnestLimitBatchSize()` - where I removed the last comment and matched the prior comment. And in `TestOutputBatchSize` - many similar adjustments to expected (max) batch size and (min) number of batches (the increase in batch size led to fewer batches).

Ben-Zvi:
Pushed another commit, where the changes are:

(1) Renamed all the "TargetOutputRowCount" to "MaxRowCount" ...
(2) Removed the "minus 1" for `setCurrentOutgoingMaxRowCount()` (Though this should be restored; see below) 
(3) In the record batch memory manager, made
          `      MAX_NUM_ROWS = ValueVector.MAX_ROW_COUNT - 1` 
     So all the references start 1 below the 64k.
(4) `setOutputRowCount()` has two implementations; the second just set the given value as is. So changed it to check for a power of two (and greater than 1), and then subtract 1.

The test `testUnnestLimitBatchSize_WithExcludedCols()` still fails now. The reason may have to do with LateralJoinBatch-innerNext(), which calls "update" on then memory manager (twice - left and right), and **AFTER** that calls allocateVectors().   However the update assumed that the outgoing vectors are already allocated.
   Also allocateVectors() is using "outputIndex == 0" to check if the allocation has not yet happened. This may be wrong (e.g., allocated, but batch is still empty).
    In the failed test, the allocated batch was of size 255, but empty, and the update calculated that the remaining size fits 128 entries. So the new allocation size became 127, and the "max rows" was calculated as MIN ( 255, 0 + 128 ), hence 128. ==> Thus undoing (2) above could fix that.

sohami:
@Ben-Zvi - Changes in step (3) and (4) are not required that will mess up the calculation of `RecordBatchMemoryManager` and it's resulting in lots of test failures as well.

For lateral tests as discussed there is a separate change needed in lateral to call `getOutputRowCount()` in case when new output batch is allocated and `getCurrentOutgoingMaxRowCount` in cases when already allocated output batch is used to fill rows from new incoming. After that the failing test will pass. 
Also the `outputIndex` check in `allocateVectors` is there to handle a special case of left join. I have added comments to explain the scenario.
With all above changes there were few `HashJoin Tests `which were failing because of a bug and I have addressed that too in my commit.
Please see all set of changes here: https://github.com/Ben-Zvi/drill/pull/9 . You can add that commit in your PR. Currently I am running tests to see if there are any other issues which may need change in the respective operator.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/MergeJoinBatch.java
```diff
@@ -124,7 +124,8 @@
      */
     @Override
     public void update(int inputIndex) {
-      status.setTargetOutputRowCount(super.update(inputIndex, status.getOutPosition()));
+      super.update(inputIndex, status.getOutPosition());
+      status.setTargetOutputRowCount(super.getCurrentOutgoingMaxRowCount());
```
Ben-Zvi: Like the other joins, add a comment: // calculated by update()

exec/java-exec/src/main/java/org/apache/drill/exec/record/JoinBatchMemoryManager.java
```diff
@@ -40,7 +40,27 @@ public JoinBatchMemoryManager(int outputBatchSize, RecordBatch leftBatch,
...
+   * Notice three (possibly) different "row counts" for the outgoing batches:
+   *
+   *  1. The rowCount that the current outgoing batch was allocated with (always a power of 2; e.g. 8192)
+   *  2. The new rowCount computed based on the newly seen input rows (always a power of 2); may be bigger than (1) if the
+   *     new input rows are much smaller than before (e.g. 16384), or smaller (e.g. 4096) if the new rows are much wider.
+   *     Subsequent outgoing batches would be allocated based on this (2) new rowCount.
+   *  3. The target rowCount for the current outgoing batch. While initially (1), it may be resized down if the new rows
+   *     are getting bigger. In any case it won't be resized above (1) (to avoid IOOB) or below the current number of rows
+   *     in that batch (i.e., outputPosition). (Need not be a power of two; e.g., 7983).
+   *
```
Ben-Zvi: Add a comment about the "- 1" issue (e.g., power of 2, minus 1).

PR #1652, opened 2019-02-23
DRILL-7046: Support for loading and parsing new RM config file

sohami:


sohami:
@HanumathRao  - Can you please help review this PR ?

paul-rogers:
Quick question. Why do we need a separate RM config file? Why can't this just be part of the existing Drill config file? By having a separate file, it just makes yet another detail for users to understand and distribute. I also wonder if this would suggest to future developers that each new feature should have its own config file?

As a counter-example, the earlier "large/small queue" RM system hosted it's config settings within the existing Drill config file?

Perhaps an intro comment for this commit could explain the reasoning?

exec/java-exec/src/main/java/org/apache/drill/exec/rpc/user/UserServer.java
```diff
@@ -197,7 +197,8 @@ protected MessageLite getResponseDefaultInstance(int rpcType) throws RpcExceptio
   }
 
   /**
-   * {@link AbstractRemoteConnection} implementation for user connection. Also implements {@link UserClientConnection}.
+   * {@link org.apache.drill.exec.rpc.AbstractRemoteConnection} implementation for user connection.
```
paul-rogers: Nit: The super class/interface is clear from the generated Javadoc; that need not be explained here. What is helpful is that which cannot be extracted from the code: what is the purpose of this class? What the heck is a "Bit to user connection"? Is this where the Drillbit connects to a neural implant in the user's brain? The Drillbit connects back to the user's client? Something else?

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/NodeResources.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.resourcemgr;
+
+public class NodeResources {
+
+  private long memoryInBytes;
+
+  private int numVirtualCpu;
```
HanumathRao: I was just thinking should we call vcpu explicitly here / just name it as a cpu.?

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/QueueAssignmentResult.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class QueueAssignmentResult {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(QueueAssignmentResult.class);
+
+  private List<ResourcePool> selectedLeafPools = new ArrayList<>();
+
+  private List<ResourcePool> rejectedPools = new ArrayList<>();
+
+  public void addSelectedPool(ResourcePool pool) {
+    Preconditions.checkState(pool.isLeafPool(), "Selected pool %s is not a leaf pool",
```
HanumathRao: Should we warn that some non-leaf pool is added and continue or stop proceeding. If the caller takes care of proceeding then may be a comment would be better here.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -677,6 +678,16 @@ private ExecConstants() {
   public static final LongValidator QUEUE_TIMEOUT = new PositiveLongValidator("exec.queue.timeout_millis", Long.MAX_VALUE,
       new OptionDescription("Indicates how long a query can wait in queue before the query fails. Range: 0-9223372036854775807"));
 
+  // New Smart RM boot time configs
+  public static final String RM_QUERY_TAGS_KEY = "exec.rm.queryTags";
+  public static final StringValidator RM_QUERY_TAGS_VALIDATOR = new StringValidator(RM_QUERY_TAGS_KEY,
+    new OptionDescription("Allows user to set coma separated list of tags for all the queries submitted over a session"));
+
+  public static final String RM_QUEUES_WAIT_FOR_PREFERRED_NODES_KEY = "exec.rm.queues.wait_for_preferred_nodes";
+  public static final StringValidator RM_QUEUES_WAIT_FOR_PREFERRED_NODES_VALIDATOR = new StringValidator
```
HanumathRao: Can this be made as  BooleanValidator instead of StringValidator?

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/selectionpolicy/RandomQueueSelection.java
```diff
@@ -0,0 +1,50 @@
...
+    super("random");
+  }
+
+  @Override
+  public ResourcePool selectQueue(List<ResourcePool> allPools, QueryContext queryContext,
+                                  NodeResources maxResourcePerNode) throws QueueSelectionException {
+    if (allPools.size() == 0) {
+      throw new QueueSelectionException("Input pool list is empty for random queue selection");
+    }
+    Collections.shuffle(allPools);
```
HanumathRao: Shouldn't this policy also select a randomQueue which satisfies the maxResourcePerNode requirement?

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/selectionpolicy/BestFitQueueSelection.java
```diff
@@ -0,0 +1,93 @@
...
+    final long queryMaxNodeMemory = maxResourcePerNode.getMemoryInMB();
+    ResourcePool selectedPool = null;
+    for (ResourcePool pool : allPools) {
+      selectedPool = pool;
+      long poolMaxNodeMem = pool.getQueuryQueue().getMaxQueryMemoryInMBPerNode();
+      if (poolMaxNodeMem == queryMaxNodeMemory) {
+        break;
+      } else if (poolMaxNodeMem > queryMaxNodeMemory) {
+        break;
+      }
```
HanumathRao: Can this condition be changed to poolMaxNodeMem >= queryMaxNodeMemory ?

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/ResourcePoolTreeImpl.java
```diff
@@ -0,0 +1,161 @@
...
+    final QueueAssignmentResult assignmentResult = selectAllQueues(queryContext);
+    final List<ResourcePool> selectedPools = assignmentResult.getSelectedLeafPools();
+    if (selectedPools.size() == 0) {
+      throw new QueueSelectionException(String.format("No resource pools to choose from for the query: %s",
+        queryContext.getQueryId()));
+    } else if (selectedPools.size() == 1) {
+      return selectedPools.get(0).getQueuryQueue();
+    }
+
+    return selectionPolicy.selectQueue(selectedPools, queryContext, queryMaxNodeResource).getQueuryQueue();
```
HanumathRao: getQueryQueue instead of getQueuryQueue?

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/ResourcePoolImpl.java
```diff
@@ -0,0 +1,273 @@
...
+
+  /**
+   * Used to determine if a ResourcePool is selected for a given query or not. It uses the assigned selector of this
+   * ResourcePool which takes in query metadata to determine if a query is allowed in this pool.
+   * @param assignmentResult - Used to keep track of all selected leaf pools and all rejected pools for given query
+   * @param queryContext - Contains query metadata like user, groups, tags, etc used by ResourcePoolSelector
+   * @return - true - if given query is selected by this ResourcePool otherwise false
+   */
+  @Override
+  public boolean visitAndSelectPool(QueueAssignmentResult assignmentResult, QueryContext queryContext) {
```
HanumathRao: Is the boolean return type used anywhere?

exec/java-exec/src/test/java/org/apache/drill/exec/resourcemgr/TestRMConfigLoad.java
```diff
@@ -0,0 +1,114 @@
...
+  public void testDefaultRMOnlyRMDisabled() throws Exception {
+    ClusterFixtureBuilder fixtureBuilder = ClusterFixture.builder(dirTestWatcher)
+      .configProperty(ExecConstants.RM_ENABLED, false)
+      .configProperty(ExecConstants.DRILL_PORT_HUNT, true)
+      .withLocalZk();
+
+    try (ClusterFixture cluster = fixtureBuilder.build()) {
+      ResourceManager resourceManager = cluster.drillbit().getContext().getResourceManager();
+      assertTrue(resourceManager instanceof DefaultResourceManager);
+    }
```
HanumathRao: All the tests are positive tests in that ResourceManager is always instance of DefaultResourceManager. Are you planning to add more tests when the tests are enabled?

sohami:
@paul-rogers - For question related to new configuration file. 
- Usually number of configurations for a new feature in Drill ranges from 5-6 at max. For RM the configuration is extensive to provide lots of flexible options for an admin. As contrast to earlier queue now admin can configure as many queues needed and hierarchy of ResourcePool can be both deep and wide. Each ResourcePool have options to configure selectors, memory resource percentage, queue, etc. Given the configuration is so huge I think it calls for a separation into it's own file. 
- Also when RM is enabled then only these configurations will be needed. If RM related configuration is present in drill-override.conf with feature disabled then still Drill will load all of it in memory which won't happen with this separate file in place.

This PR only add's support for the new RM configuration parsing and creating in-memory structures. There are still lots of components which needs to be developed like planner components, scheduler, etc. All the changes will come in phases. I have already added functional spec in the JIRA which has detailed explanation about the configuration. But I have also tried adding `package-info.java` with necessary information. Also I have moved all the files to a new config package under resourcemgr package to represent this PR contains only config changes. Please let me know if it looks sufficient to you. As far as detailed design is concerned you can find the document explaining it in [DRILL-7026](https://issues.apache.org/jira/browse/DRILL-7026) which is parent JIRA of [DRILL-7046](https://issues.apache.org/jira/browse/DRILL-7046).

sohami:
@paul-rogers - I have rebased the PR on latest master to resolve conflicts. Can you please review the last changes ?

HanumathRao:
@sohami  Thanks for making the changes. Changes look good from my end. LGTM +1.

PR #1654, opened 2019-02-24

kkhatua:
_**Note:** Closed the old PR #755 and #1250 and opening this._

When Drill is displaying profiles stored on the file system (Local or Distributed), it does so by loading the entire list of `.sys.drill` files in the profile directory, sorting and deserializing. This can get expensive, since only a single CPU thread does this.
As an example, a directory of 120K profiles, the time to just fetch the list of files alone is about 6 seconds. After that, based on the number of profiles being rendered, the time varies. An average of 30ms is needed to deserialize a standard profile, which translates to an additional 3sec for the rendering of default 100 profiles.

A user reported issue confirms just that:
[DRILL-5028](https://issues.apache.org/jira/browse/DRILL-5028) Opening profiles page from web ui gets very slow when a lot of history files have been stored in HDFS or Local FS

Additional JIRAs filed ask for managing these profiles
[DRILL-2362](https://issues.apache.org/jira/browse/DRILL-2362) Drill should manage Query Profiling archiving
[DRILL-2861](https://issues.apache.org/jira/browse/DRILL-2861) enhance drill profile file management

This PR brings the following enhancements to achieve that:

1. Improve loading times by pinning the deserialized list in memory (TreeSet; for maintaining a memory-efficient sortedness of the profiles). That way, if we do not detect any new profiles in the profileStore (i.e. profile directory) since the last time a web-request for rendering the profiles was made, we can re-serve the same listing and skip making a trip to the filesystem to re-fetch all the profiles.
2. Leverage Guava Cache to save on deserializing profiles, since the WebServer makes 2 calls to deserialize a file on disk for rendering.

Reload & reconstruction of the profiles in the Tree is done in the event of any of the following states changing:
i. Modification Time of profile dir
ii. Number of profiles in the profile dir
iii. Number of profiles requested exceeds existing the currently available list

Access to the profiles requires a lock to ensure that at any time, only one of the requesting users is able to trigger a reconstruction of the TreeSet. If reconstruction is not required, it uses the existing cache rather than attempting to read the entire profile directory's contents all over again.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -236,6 +236,7 @@ public final class ExecConstants {
   public static final String SYS_STORE_PROVIDER_LOCAL_ENABLE_WRITE = "drill.exec.sys.store.provider.local.write";
   public static final String PROFILES_STORE_INMEMORY = "drill.exec.profiles.store.inmemory";
   public static final String PROFILES_STORE_CAPACITY = "drill.exec.profiles.store.capacity";
+  public static final String PROFILES_STORE_CACHE_SIZE = "drill.exec.profiles.store.cache.size";
   public static final String IMPERSONATION_ENABLED = "drill.exec.impersonation.enabled";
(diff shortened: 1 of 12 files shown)


kkhatua:
image

PR #1655, opened 2019-02-24
DRILL-6927: Avoid double conversion from impala timestamp when hive native parquet reader is used

vvysotskyi:
For problem description please see [DRILL-6927](https://issues.apache.org/jira/browse/DRILL-6927).


vdiravka:
@vvysotskyi  LGTM, it brings Drill closer to make these two options `true` by default. 
Please also add unit test in `TestHiveDrillNativeParquetReader`.

vvysotskyi:
@vdiravka, thanks for the review, added unit test.

PR #1656, opened 2019-02-25
DRILL-7056: Drill fails with NPE when starting in distributed mode & …

kkhatua:
…31010 port is used

This occurs because during the WebServer.close() , one of the tasks for clean up is to delete the temp directory which never existed. This PR's fix does a check for that.


vvysotskyi:
@kkhatua, please add the unit test (you may use `TestGracefulShutdown.testDrillbitWithSamePortContainsShutdownThread` as an example).

Also, in `WebServer` `tmpJavaScriptDir` is marked for deletion (`tmpJavaScriptDir.deleteOnExit()`) and this directory is also deleted in the `close()` method. Please either remove `deleteOnExit()` call or remove `FileUtils.deleteDirectory(tmpJavaScriptDir);`.

If `FileUtils.deleteDirectory(tmpJavaScriptDir);` will be left, I think it makes sence to replace it with `FileUtils.deleteQuietly(tmpJavaScriptDir);` to avoid possible errors in `close()` method.

And finally please order the methods in `WebServer` according to JCC, since `getTmpJavaScriptDir()` method is added before the constructor in your fix for DRILL-5735.

kkhatua:
@vvysotskyi 
I've made the changes and added a unit test. Please review

exec/java-exec/src/test/java/org/apache/drill/test/TestWebServerTempDir.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+import java.io.File;
+
+@Category({UnlikelyTest.class})
+public class TestWebServerTempDir extends BaseTestQuery {
```
vvysotskyi: Could you please use an existing class instead of providing the new one for a single test?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -467,43 +452,64 @@ public void close() throws Exception {
     if (embeddedJetty != null) {
       embeddedJetty.stop();
     }
-    //Deleting temp directory
-    FileUtils.deleteDirectory(getTmpJavaScriptDir());
+    // Deleting temp directory, if it exists
+    if (tmpJavaScriptDir != null) {
+      FileUtils.deleteDirectory(tmpJavaScriptDir);
```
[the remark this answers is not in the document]
kkhatua: 👍

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebServer.java
```diff
@@ -467,43 +452,64 @@ public void close() throws Exception {
...
 
-    //Template source Javascript file
+    // Template source Javascript file
     InputStream optionsDescripTemplateStream = Resource.newClassPathResource(OPTIONS_DESCRIBE_TEMPLATE_JS).getInputStream();
-    //Generated file
-    File optionsDescriptionFile = new File(getTmpJavaScriptDir(), OPTIONS_DESCRIBE_JS);
+    // Generated file
+    File optionsDescriptionFile = new File(getOrCreateTmpJavaScriptDir(), OPTIONS_DESCRIBE_JS);
     final String file_content_footer = "};";
     optionsDescriptionFile.deleteOnExit();
```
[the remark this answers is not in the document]
kkhatua: 👍 
I assumed that for deleting a dir, it had to be empty. I see that FileUtils doesn't restrict that.

exec/java-exec/src/test/java/org/apache/drill/test/TestWebServerTempDir.java
kkhatua: Made it part of TestGracefulShutdown since that class comes closest to dealing with server startup and shutdown.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -561,4 +562,9 @@ private static String stripQuotes(final String s, final String systemProp) {
     // return as-is
     return s;
   }
+
+  // Only for testing purposes (DRILL-7056)
+  public File getWebServerTempDirPath() {
```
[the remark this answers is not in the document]
kkhatua: Done.

kkhatua:
@vvysotskyi  please review the changes. I've merged the test into Graceful Shutdown tests as that class was the closest in terms of what we are testing.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -205,6 +212,47 @@ public void testDrillbitWithSamePortContainsShutdownThread() throws Exception {
...
+      //      2. Second instance startup should fail, because ports are occupied by the first one
+      Drillbit originalDrillbit = fixture.drillbit();
+      assertNotNull("First drillbit instance should be initialized", originalDrillbit);
+      originalDrillbitTempDir = getWebServerTempDirPath(originalDrillbit);
+      assertTrue("First drillbit instance should have a temporary Javascript dir initialized", originalDrillbitTempDir.exists());
+      try {
+        twinDrillbitOnSamePort.run();
+        fail("Invocation of 'twinDrillbitOnSamePort.run()' should throw UserException");
+      } catch (UserException e) {
+        assertNull("Second drillbit instance should NOT have a temporary Javascript dir", getWebServerTempDirPath(twinDrillbitOnSamePort));
```
vvysotskyi: `webServerHandle.getOrCreateTmpJavaScriptDir()` method actually creates or returns created temp dir, is it required to call it on the non-started drillbit?

kkhatua: I test by calling the method because the non-started drillbit is still instantiated, but failed to come up. 
The method will not create a temp dir if the startup failed. I thought it was a valid test.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -205,6 +212,47 @@ public void testDrillbitWithSamePortContainsShutdownThread() throws Exception {
...
+  }
+
+  private static File getWebServerTempDirPath(Drillbit drillbit) {
+    File webServerTempDirPath = null;
+    try {
+      Field webServerField = FieldUtils.getField(drillbit.getClass(), "webServer", true);
+      WebServer webServerHandle = (WebServer) FieldUtils.readField(webServerField, drillbit, true);
+      webServerTempDirPath = webServerHandle.getOrCreateTmpJavaScriptDir();
+    } catch (IllegalAccessException e) {
+        e.printStackTrace();
```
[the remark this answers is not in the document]
kkhatua: Ok

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -205,6 +212,47 @@ public void testDrillbitWithSamePortContainsShutdownThread() throws Exception {
...
+      //      1. First drillbit instance should be started normally
+      //      2. Second instance startup should fail, because ports are occupied by the first one
+      Drillbit originalDrillbit = fixture.drillbit();
+      assertNotNull("First drillbit instance should be initialized", originalDrillbit);
+      originalDrillbitTempDir = getWebServerTempDirPath(originalDrillbit);
+      assertTrue("First drillbit instance should have a temporary Javascript dir initialized", originalDrillbitTempDir.exists());
+      try {
+        twinDrillbitOnSamePort.run();
+        fail("Invocation of 'twinDrillbitOnSamePort.run()' should throw UserException");
+      } catch (UserException e) {
```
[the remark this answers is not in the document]
kkhatua: Strange that my Eclipse IDE didn't catch this.

kkhatua:
@vvysotskyi  Done

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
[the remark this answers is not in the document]
vvysotskyi: I don't see a check for error message. As an example, you may use `TestGracefulShutdown.testDrillbitWithSamePortContainsShutdownThread()` test.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
vvysotskyi: Actually, the check you have added in `getOrCreateTmpJavaScriptDir()` doesn't guarantee that drillbit is started. Context may be non-nullable since exception may happen later, but drill is not started. Please revert that check. Also, I think this method should not be called on closed drillbit, so it should be removed from the test.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
```diff
@@ -205,6 +212,44 @@ public void testDrillbitWithSamePortContainsShutdownThread() throws Exception {
...
+      assertNotNull("First drillbit instance should be initialized", originalDrillbit);
+      originalDrillbitTempDir = getWebServerTempDirPath(originalDrillbit);
+      assertTrue("First drillbit instance should have a temporary Javascript dir initialized", originalDrillbitTempDir.exists());
+      try {
+        twinDrillbitOnSamePort.run();
+        fail("Invocation of 'twinDrillbitOnSamePort.run()' should throw UserException");
+      } catch (UserException userEx) {
+        assertNull("Second drillbit instance should NOT have a temporary Javascript dir", getWebServerTempDirPath(twinDrillbitOnSamePort));
+      } catch (Exception e) {
+        fail("Invocation of 'twinDrillbitOnSamePort.run()' was expected to throw a UserException. Actual exception : "+ e.getMessage());
```
vvysotskyi: This check will hide the stack trace of the error. Please either remove this catch block or rethrow the exception.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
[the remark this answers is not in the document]
kkhatua: Ok. 
I was assuming that the exception thrown is related to the Drillbit not binding to the port, but it could have been some other resource exception as well. Added the check.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
kkhatua: Good point... Failing the test isn't sufficient. Will rethrow.

kkhatua:
Updated the PR.

exec/java-exec/src/test/java/org/apache/drill/test/TestGracefulShutdown.java
vvysotskyi: Oh, looks like I was mistaken when said that rethrowing will be enough. It wouldn't be rethrown since the test will fail at this line. So please remove this block.

PR #1659, opened 2019-02-26
DRILL-7035: Drill C++ Client crashes on multiple SaslAuthenticatorImpl destruction due to communication error

debraj92:
The fix ensures that the check for  null of m_saslAuthenticator and delete of m_saslAuthenticator object are atomic and thread safe.


debraj92:
The patches are pushed wrongly in the master branch. I will create a new PR

PR #1662, opened 2019-02-27
DRILL-6825: Apply different hash algorithms to different data types

weijietong:
This PR introduces those main changes:
* Add hash32(int index) method to the ValueVector interface.
* To Interger data types, introduce https://gist.github.com/badboy/6267743 multiply-shift algorithm which was proved to be good at performance by the paper https://15721.courses.cs.cmu.edu/spring2019/papers/17-hashjoins/richter-vldb2015.pdf. Other data types like double,float,string still use the Murmur3 hash method.
* Since HashJoin needs to solve implicit casting , it treats all numeric data types as double. So this PR only apply to HashAgg operator.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/ChainedHashTable.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/ChainedHashTable.java
@@ -24,9 +24,7 @@ import com.sun.codemodel.JExpression;
 import org.apache.drill.common.expression.ErrorCollector;
 import org.apache.drill.common.expression.ErrorCollectorImpl;
 import org.apache.drill.common.expression.LogicalExpression;
-import org.apache.drill.common.expression.ValueExpressions;
 import org.apache.drill.common.logical.data.NamedExpression;
-import org.apache.drill.common.types.TypeProtos;
 import org.apache.drill.common.types.Types;
 import org.apache.drill.exec.compile.sig.GeneratorMapping;
 import org.apache.drill.exec.compile.sig.MappingSet;
@@ -43,11 +41,9 @@ import org.apache.drill.exec.expr.fn.FunctionGenerationHelper;
 import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.physical.impl.join.JoinUtils;
-import org.apache.drill.exec.planner.physical.HashPrelUtil;
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.RecordBatch;
 import org.apache.drill.exec.record.TypedFieldId;
-import org.apache.drill.exec.record.VectorAccessible;
 import org.apache.drill.exec.record.VectorContainer;
 import org.apache.drill.exec.vector.ValueVector;
 
@@ -110,10 +106,6 @@ public class ChainedHashTable {
     new MappingSet("htRowIdx", null, "htContainer", null, SETUP_INTERIOR_CONSTANT, BOTH_KEYS_NULL);
   private final MappingSet KeyMatchHtableProbeMapping =
       new MappingSet("htRowIdx", null, "htContainer", null, SETUP_INTERIOR_CONSTANT, KEY_MATCH_PROBE);
-  private final MappingSet GetHashIncomingBuildMapping =
-      new MappingSet("incomingRowIdx", null, "incomingBuild", null, DO_SETUP_CONSTANT, GET_HASH_BUILD);
-  private final MappingSet GetHashIncomingProbeMapping =
-      new MappingSet("incomingRowIdx", null, "incomingProbe", null, DO_SETUP_CONSTANT, GET_HASH_PROBE);
   private final MappingSet SetValueMapping =
       new MappingSet("incomingRowIdx" /* read index */, "htRowIdx" /* write index */,
           "incomingBuild" /* read container */, "htContainer" /* write container */, SETUP_INTERIOR_CONSTANT,
@@ -155,7 +147,7 @@ public class ChainedHashTable {
     // Uncomment out this line to debug the generated code.
     // This code is called from generated code, so to step into this code,
     // persist the code generated in HashAggBatch also.
-    // top.saveCodeForDebugging(true);
+    //top.saveCodeForDebugging(true);
     top.preferPlainJava(true); // use a subclass
     ClassGenerator<HashTable> cg = top.getRoot();
     ClassGenerator<HashTable> cgInner = cg.getInnerGenerator("BatchHolder");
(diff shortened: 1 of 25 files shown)


weijietong:
@Ben-Zvi could you review this PR ?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/ChainedHashTable.java
```diff
@@ -43,11 +41,9 @@
 import org.apache.drill.exec.memory.BufferAllocator;
 import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.physical.impl.join.JoinUtils;
-import org.apache.drill.exec.planner.physical.HashPrelUtil;
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.RecordBatch;
 import org.apache.drill.exec.record.TypedFieldId;
-import org.apache.drill.exec.record.VectorAccessible;
 import org.apache.drill.exec.record.VectorContainer;
 import org.apache.drill.exec.vector.ValueVector;
```
Ben-Zvi: (This comment should be at lines 71-77 below; github does not let putting a comment there....)
Need also to remove the code generation for "getHashBuild" and "getHashProbe".

weijietong:
Since HashJoin needs to solve implicit casting,it treats all different numeric data types as double. So this PR only apply to HashAgg.

Ben-Zvi:
@weijietong can you add a detailed description of your work - the hash functions are a sensitive part of the execution, and possibly more people would like to look into the changes.

weijietong:
@Ben-Zvi I have added some description.

Ben-Zvi:
Adding the hash32() method to the ValueVector is useful; however picking up algorithms just based on a paper or being famous may not be good enough.  At my previous employer I evaluated many hash functions by actually running (stand alone) performance and distribution tests. One clear result shown back then is that murmur performed well on long strings, less good on shorter data.

   How do the new hash functions in IntegerHashing compare with the existing one in HashHelper ?

     The Boost implementation of hash_combine looks "fishy" (e.g., some bits get more used than others) -- see some more critique at  https://stackoverflow.com/questions/35985960/c-why-is-boosthash-combine-the-best-way-to-combine-hash-values  

    Why can't the seed be given directly to the hash function instead of being "combined" later ?

   Another good hash function used in the past (don't recall any name) worked with a map of 256 prime numbers, and the code (starting with the seed) was using each  input byte as an index to the map - rotate old value, XOR with new mapped value, continue ....

   Now things may perform differently in Java.  
   Also - do you know of any open source hash functions we can just import instead of writing the code in Drill ?

weijietong:
The IntegerHashing's method was also used in ClickHouse for integer types(see: https://github.com/yandex/ClickHouse/blob/master/dbms/src/Common/HashTable/Hash.h   intHash32 method). CK does a fine hashing method choosing according to the data types and keys width which is valuable for us to learn. As you mentioned Murmur3Hash does not have a good performance at the shorter integer case.So it's better to use the IntegerHash at the integer keys case.

The Boost implementation's discussion you mentioned I had read before. But I think it's reasonable why Boost still keep the current implementation now as a base library. 

The reason to keep seed away from the hash32 function and involve the Boost's hash_combine method is that I want to change the current hashing strategy later. I plan to change the hash32(hash32(hash32)) row iterate model to `hash32() hash_combine hash32() hash_combine hash32()` column combine model at the multi-keys case. The row iterate module has a data dependency and will hurt the cpu pipeline performance.

Other hashing methods I know can be found here: https://github.com/benalexau/hash-bench.  It's a java hashing method collection. The benchmark I run showed that https://github.com/OpenHFT/Zero-Allocation-Hashing/blob/master/src/main/java/net/openhft/hashing/LongHashFunction.java 's city_1_1 has a best performance at 32,64 bytes key width.

I also wonder whether we can do the join keys data type implication at the project node later. So the HashJoin and Exchange node can also benefit from this PR.

weijietong:
@Ben-Zvi  I have done a benchmark about IntegerHash and MurmurHash3_32. The result shows that IntegerHash has nearly 2 times performance than MurmurHash3_32.
```
HashBenchInteger.hashInteger  integerHash       N/A  avgt    5   3.987 ± 0.162  ns/op
HashBenchInteger.hashInteger   Murmur3_32       N/A  avgt    5   6.626 ± 0.085  ns/op
HashBenchInteger.hashLong     integerHash       N/A  avgt    5   4.903 ± 0.320  ns/op
HashBenchInteger.hashLong      Murmur3_32       N/A  avgt    5   8.525 ± 0.649  ns/op
```
I also run two sql queries comparison which are using IntegerHash and MurmurHash separately to the long data types. The queries are hashing hotspot:
Q1:
```
select c.c_custkey, count(*)
from dfs.`/tpch100/customer` c
group by c.c_custkey
```
Q2:
```
select c.c_custkey,c.c_nationkey, count(*)
from dfs.`/tpch100/customer` c
group by c.c_custkey,c.c_nationkey
```
The dataset is tpc-h scale 100.  The query result shows that : 
To Q1:
IntegerHash has a 5% query performance improvement than MurmurHash3_32(IntegerHash: 14.029 sec, MurmurHash3_32: 14.870 sec ).
To Q2:
IntegerHash has a 16% query performance improvement than MurmurHash3_32(IntegerHash: 20.499 sec,MurmurHash3_32: 23.921 sec).

It is clear that the more integer datatype columns grouped by, the more query performance will be gained.

exec/vector/src/main/java/org/apache/drill/exec/hash/MurmurHashing.java
```diff
@@ -0,0 +1,136 @@
...
+package org.apache.drill.exec.hash;
+
+import io.netty.buffer.DrillBuf;
+import io.netty.util.internal.PlatformDependent;
+
+/**
+ * 32bit hashing for string and so on.
+ * murmurhash3
+ */
+public class MurmurHashing
```
Ben-Zvi: BTW, seems that the same code already exists in org.apache.drill.exec.expr.fn.impl.MurmurHash3

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
```diff
@@ -624,6 +639,27 @@ private void retryAfterOOM(boolean batchAdded) throws RetryAfterSpillException {
...
+  /**
+   *   Return the Hash Value for the row in the Build incoming batch at index:
+   *   (Only apply to Hash Aggregate there's no "Build" side -- only one batch - this one)
+   *
+   * @param incomingRowIdx
+   * @return
+   * @throws SchemaChangeException
+   */
+  @Override
+  public int getBuildHashCodeTreatDataTypes(int incomingRowIdx) throws SchemaChangeException {
```
Ben-Zvi: A benefit of this work can be to extract the Hash-Value calculation from the HashTable class(es). Conceptually the two are independent -- the hash-value is a parameter given to the hash table API.

weijietong:
@Ben-Zvi  The hash code computation of HashAgg has been moved to `HashAggTemplate`. Your review advices have been applied. Please check, tks!

Ben-Zvi:
@weijietong - how about postponing this work to 1.17.0 ?  It is a change in a critical part of the engine, and we need to get a better understanding of the impact, and 1.16.0 is too close. (Also I am very busy with some 1.16 work now, and have no time to test myself).  In 1.17.0 the work can be extended to Hash-Join and exchanges to be more complete.

weijietong:
@Ben-Zvi agree with you. I have changed the issue to 1.17.0

weijietong:
@Ben-Zvi Do you have the time to go on this work ?

arina-ielchiieva:
@weijietong that's the status of this PR?

PR #1664, opened 2019-02-27
Fix few issues with the poweredBy page

aashreya:


aashreya:
@kkhatua And @bbevens

PR #1666, opened 2019-02-28
DRILL-7058: Refresh command to support subset of columns

dvjyothsna:
@gparai  Please review

### DIFF
--- a/exec/java-exec/src/main/codegen/data/Parser.tdd
+++ b/exec/java-exec/src/main/codegen/data/Parser.tdd
@@ -37,12 +37,13 @@
     "METADATA",
     "IF",
     "JAR",
-    "PROPERTIES"
+    "PROPERTIES",
     "ANALYZE",
     "COMPUTE",
     "ESTIMATE",
     "STATISTICS",
-    "SAMPLE"
+    "SAMPLE",
+    "COLUMNS"
   ]
 
   # List of methods for parsing custom SQL statements.
@@ -854,6 +855,7 @@
         "YEAR",
         # "YEARS", # not a keyword in Calcite
         "ZONE",
+        "COLUMNS"
       ]
 
   # List of additional join types. Each is a method with no arguments.
--- a/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
+++ b/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
@@ -464,22 +464,30 @@ SqlNode SqlDropSchema(SqlParserPos pos) :
 
 /**
  * Parse refresh table metadata statement.
- * REFRESH TABLE METADATA tblname
+ * REFRESH TABLE METADATA [COLUMNS ((field1, field2,..) | NONE)] tblname
  */
 SqlNode SqlRefreshMetadata() :
 {
     SqlParserPos pos;
     SqlIdentifier tblName;
-    SqlNodeList fieldList;
+    SqlNodeList fieldList = null;
     SqlNode query;
+    boolean allColumns = true;
 }
 {
     <REFRESH> { pos = getPos(); }
     <TABLE>
     <METADATA>
+    [
+        <COLUMNS> { allColumns = false; }
+        (   fieldList = ParseRequiredFieldList("Table")
+            |
+            <NONE>
+        )
+    ]
     tblName = CompoundIdentifier()
     {
-        return new SqlRefreshMetadata(pos, tblName);
+        return new SqlRefreshMetadata(pos, tblName, SqlLiteral.createBoolean(allColumns, getPos()), fieldList);
     }
 }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
@@ -17,6 +17,13 @@
  */
 package org.apache.drill.exec.planner.sql.handlers;
 
+import java.util.HashSet;
+import java.util.Set;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlLiteral;
+import org.apache.calcite.sql.SqlNodeList;
+import org.apache.calcite.sql.parser.SqlParserPos;
+import org.apache.drill.common.expression.SchemaPath;
 import static org.apache.drill.exec.planner.sql.SchemaUtilites.findSchema;
 
 import org.apache.calcite.schema.SchemaPlus;
@@ -69,6 +76,9 @@ public class RefreshMetadataHandler extends DefaultSqlHandler {
       }
 
       final String tableName = refreshTable.getName();
+      final SqlNodeList columnList = getColumnList(refreshTable);
+      final Set<String> columnSet = getColumnRootSegments(columnList);
+      final SqlLiteral allColumns = refreshTable.getAllColumns();
 
       if (tableName.contains("*") || tableName.contains("?")) {
         return direct(false, "Glob path %s not supported for metadata refresh", tableName);
@@ -121,7 +131,7 @@ public class RefreshMetadataHandler extends DefaultSqlHandler {
         .withFormatConfig((ParquetFormatConfig) formatConfig)
         .withOptions(context.getOptions())
         .build();
-      Metadata.createMeta(fs, selectionRoot, readerConfig);
+      Metadata.createMeta(fs, selectionRoot, readerConfig, allColumns.booleanValue(), columnSet);
       return direct(true, "Successfully updated metadata for table %s.", tableName);
 
     } catch(Exception e) {
@@ -130,5 +140,29 @@ public class RefreshMetadataHandler extends DefaultSqlHandler {
     }
   }
 
+  private Set<String> getColumnRootSegments(SqlNodeList columnList) {
+    Set<String> columnSet = new HashSet<>();
+    if (columnList != null) {
+      for (SqlNode column : columnList.getList()) {
+        // Add only the root segment. Collect metadata for all the columns under that root segment
+        columnSet.add(SchemaPath.parseFromString(column.toString()).getRootSegmentPath());
+      }
+    }
+    return columnSet;
+  }
+
+  /**
+   * Generates the column list specified in the Refresh statement
+   * @param sqlrefreshMetadata sql parse node representing refresh statement
+   * @return list of columns specified in the refresh command
+   */
+  private SqlNodeList getColumnList(final SqlRefreshMetadata sqlrefreshMetadata) {
+    SqlNodeList columnList = sqlrefreshMetadata.getFieldList();
+    if (columnList == null || !SqlNodeList.isEmptyList(columnList)) {
+      columnList = new SqlNodeList(SqlParserPos.ZERO);
+      columnList.add(new SqlIdentifier(SchemaPath.STAR_COLUMN.rootName(), SqlParserPos.ZERO));
+    }
+    return columnList;
+  }
 
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
@@ -24,6 +24,7 @@ import org.apache.calcite.sql.SqlIdentifier;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlLiteral;
 import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlNodeList;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.SqlSpecialOperator;
 import org.apache.calcite.sql.SqlWriter;
@@ -43,15 +44,19 @@ public class SqlRefreshMetadata extends DrillSqlCall {
   public static final SqlSpecialOperator OPERATOR = new SqlSpecialOperator("REFRESH_TABLE_METADATA", SqlKind.OTHER_DDL) {
     @Override
     public SqlCall createCall(SqlLiteral functionQualifier, SqlParserPos pos, SqlNode... operands) {
-      return new SqlRefreshMetadata(pos, (SqlIdentifier) operands[0]);
+      return new SqlRefreshMetadata(pos, (SqlIdentifier) operands[0], (SqlLiteral) operands[1], (SqlNodeList) operands[2]);
     }
   };
 
   private SqlIdentifier tblName;
+  private final SqlLiteral allColumns;
+  private final SqlNodeList fieldList;
 
-  public SqlRefreshMetadata(SqlParserPos pos, SqlIdentifier tblName){
+  public SqlRefreshMetadata(SqlParserPos pos, SqlIdentifier tblName, SqlLiteral allColumns, SqlNodeList fieldList){
     super(pos);
     this.tblName = tblName;
+    this.allColumns = allColumns;
+    this.fieldList = fieldList;
   }
 
   @Override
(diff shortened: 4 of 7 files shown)


exec/java-exec/src/main/codegen/data/Parser.tdd
```diff
@@ -38,6 +38,7 @@
     "IF",
     "JAR",
     "PROPERTIES"
+    "COLUMNS",
```
arina-ielchiieva: Looks like comma should be before `COLUMNS`.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -464,22 +464,30 @@ SqlNode SqlDropSchema(SqlParserPos pos) :
...
 }
 {
     <REFRESH> { pos = getPos(); }
     <TABLE>
     <METADATA>
+    [
+        <COLUMNS> { allColumns = false; }
+        (   fieldList = ParseRequiredFieldList("Table")
+            |
+            <NONE>
```
arina-ielchiieva: Do we need `NONE` here? I think if columns keyword is present list of columns is required.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -550,9 +558,9 @@ Pair<SqlNodeList, SqlNodeList> ParenthesizedCompoundIdentifierList() :
     SqlIdentifier id;
 }
 {
-    id = SimpleIdentifier() {list.add(id);}
+    id = CompoundIdentifier() {list.add(id);}
```
arina-ielchiieva: 1. Why this change is needed? 
2. Please updated java-doc to reflect new method behavior: `Parses a comma-separated list of simple identifiers`.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
```diff
@@ -130,5 +140,25 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ForemanSetupException {
...
+        columnSet.add(SchemaPath.parseFromString(column.toString()).getRootSegmentPath());
+      }
+    }
+    return columnSet;
+  }
+
+  /* Generates the column list specified in the Refresh statement */
+  private SqlNodeList getColumnList(final SqlRefreshMetadata sqlrefreshMetadata) {
+    SqlNodeList columnList = sqlrefreshMetadata.getFieldList();
+    if (columnList == null || columnList.size() <= 0) {
```
arina-ielchiieva: `columnList.size() <= 0` -> `!columnsList.isEmpty()`

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
```diff
@@ -130,5 +140,25 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ForemanSetupException {
...
+    if (columnList != null) {
+      for (SqlNode column : columnList.getList()) {
+        // Add only the root segment. Collect metadata for all the columns under that root segment
+        columnSet.add(SchemaPath.parseFromString(column.toString()).getRootSegmentPath());
+      }
+    }
+    return columnSet;
+  }
+
+  /* Generates the column list specified in the Refresh statement */
```
arina-ielchiieva: Please add proper java doc rather than comment.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
```diff
@@ -43,15 +44,27 @@
...
+      return new SqlRefreshMetadata(pos, (SqlIdentifier) operands[0], (SqlLiteral) operands[1], (SqlNodeList) operands[2]);
     }
   };
 
   private SqlIdentifier tblName;
+  private final SqlLiteral allColumns;
+  private final SqlNodeList fieldList;
 
-  public SqlRefreshMetadata(SqlParserPos pos, SqlIdentifier tblName){
+  public SqlNodeList getFieldList() {
```
arina-ielchiieva: Please move new getters after other getters, not before the constructor.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -464,22 +464,30 @@ SqlNode SqlDropSchema(SqlParserPos pos) :
...
+ * REFRESH TABLE METADATA [COLUMNS ((field1, field2,..) | NONE)] tblname
  */
 SqlNode SqlRefreshMetadata() :
 {
     SqlParserPos pos;
     SqlIdentifier tblName;
-    SqlNodeList fieldList;
+    SqlNodeList fieldList = null;
     SqlNode query;
+    boolean allColumns = true;
```
arina-ielchiieva: I don't think we need allColumns state here, if fieldList is empty it means we refresh all columns, if not then part of them. Having `SqlNodeList fieldList` is enough.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
```diff
@@ -43,15 +44,27 @@
...
+    return fieldList;
+  }
+
+  public SqlLiteral getAllColumns() {
+    return allColumns;
```
arina-ielchiieva: As mentioned above this state can be determined from fieldList, no need for second class variable.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
```diff
@@ -63,6 +76,8 @@ public SqlOperator getOperator() {
   public List<SqlNode> getOperandList() {
     List<SqlNode> ops = Lists.newArrayList();
     ops.add(tblName);
+    ops.add(allColumns);
+    ops.add(fieldList);
```
arina-ielchiieva: You also need to amend `public void unparse(SqlWriter writer, int leftPrec, int rightPrec) {` based on existence of `fieldList`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -101,14 +102,15 @@ private Metadata(ParquetReaderConfig readerConfig) {
 
   /**
    * Create the parquet metadata file for the directory at the given path, and for any subdirectories.
-   *
    * @param fs file system
    * @param path path
    * @param readerConfig parquet reader configuration
+   * @param allColumns
```
arina-ielchiieva: Please add description to params to avoid warnings in the IDE.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -203,12 +205,14 @@ private static boolean ignoreReadingMetadata(MetadataContext metaContext, Path p
    *
    * @param path to the directory of the parquet table
    * @param fs file system
+   * @param allColumns
```
arina-ielchiieva: Same here.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
dvjyothsna: Yes, since the columns keyword is optional, metadata will be collected for all the columns if the Columns keyword is not present. So the only way to express not to collect metadata for any column is to have NONE keyword which makes it more explicit.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
dvjyothsna: I don't think isEmpty() method exists for SqlNodeList class.

vdiravka: Looks like you can use the similar static method:
`SqlNodeList.isEmptyList(columnList)`

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
dvjyothsna: We have three scenarios here:
Refresh metadata for all the columns -  allColumns - true, fieldList - []
Refresh metadata for subset of columns - allColumns -false, fieldList - [c1,c2,..]
Refresh metadata and don't collect column metadata - allColumns - false, fieldList - []
Since we have three scenarios we can't represent all three by using just one variable.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Added description

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Added description

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
dvjyothsna: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
dvjyothsna: Please see above comment.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/parser/SqlRefreshMetadata.java
dvjyothsna: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
dvjyothsna: Added java doc.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
dvjyothsna: Changed to use isEmptyList()

dvjyothsna:
@arina-ielchiieva Added examples to the jira and made changes based on your review comments. I discussed about unit tests with Aman and the only way we can unit test is creating cache file and asserting it with a static file.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
```diff
@@ -130,5 +140,29 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ForemanSetupException {
...
+        columnSet.add(SchemaPath.parseFromString(column.toString()).getRootSegmentPath());
+      }
+    }
+    return columnSet;
+  }
+
+  /**
+   * Generates the column list specified in the Refresh statement
+   * @param sqlrefreshMetadata sql parse node representing refresh statement
+   * @return columnlist list of columns specified in the refresh command
```
gparai: You could skip the variable name `columnlist` in the return. See elsewhere in this file.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -413,7 +422,7 @@ private ColTypeInfo getColTypeInfo(MessageType schema, Type type, String[] path,
    * Get the metadata for a single file
    */
   private ParquetFileMetadata_v3 getParquetFileMetadata_v3(ParquetTableMetadata_v3 parquetTableMetadata,
-      final FileStatus file, final FileSystem fs) throws IOException, InterruptedException {
+                                                           final FileStatus file, final FileSystem fs, boolean allColumns, Set<String> columnSet) throws IOException, InterruptedException {
```
gparai: Correct indentation?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -462,25 +470,28 @@ private ParquetFileMetadata_v3 getParquetFileMetadata_v3(ParquetTableMetadata_v3
...
-        boolean statsAvailable = stats != null && !stats.isEmpty();
-        if (statsAvailable) {
-          if (stats.hasNonNullValue()) {
-            minValue = stats.genericGetMin();
-            maxValue = stats.genericGetMax();
-            if (containsCorruptDates == ParquetReaderUtility.DateCorruptionStatus.META_SHOWS_CORRUPTION && columnTypeMetadata.originalType == OriginalType.DATE) {
-              minValue = ParquetReaderUtility.autoCorrectCorruptedDate((Integer) minValue);
-              maxValue = ParquetReaderUtility.autoCorrectCorruptedDate((Integer) maxValue);
+          // Store column metadata only if allColumns is set to true or if the column belongs to the subset of columns specified in the refresh command
+          if (allColumns || columnSet == null || !allColumns && columnSet != null && columnSet.size() > 0 && columnSet.contains(columnSchemaName.getRootSegmentPath())) {
```
gparai: Correct indentation?

dvjyothsna:
@gparai I looked at the unit test in TestAnalyze and realized that .stats.drill is a directory whereas .drill.parquet_metadata is a JSON file. So that file can't be queried. I need to either change the file name to have .json extension or create a directory and put metadata cache file under that. I'm not sure how much impact it will have on backward compatibility.

arina-ielchiieva:
@dvjyothsna there are a couple of options to verify json metadata:
1.  use table function to query from json thus your file does not need to have json extension;
2. modify file system plugin for this test to accept such files;
3. parse json code and verify its content.

amansinha100:
The type of unit tests that depend on the structure of the JSON cache file could be problematic to maintain.  We do expect the nature of the metadata stored in the cache file to change/evolve depending on the requirements.   Hence, wherever appropriate we should use metrics that are invariants such that in future unit tests don't need to be updated frequently.   For example,  in TestParquetMetadataCache.java, we use the flag **usedMetadataFile = true/false**  for testing.  The structure of the cache file was not analyzed there. 
Can we define additional metrics that are likely to remain invariant but are able to exercise the 'subset-of-coulumns' functionality ?

dvjyothsna:
@amansinha100 I agree with you. In case if we are to change the structure of the metadata cache file or even change the format in the next version of Drill, the tests need to be rewritten. Yes, metrics of that sort can be added. The proposal is to store a flag allColumnMetadata(to represent column metadata is stored for all the columns) and the subset of columns in the summary file and use it during auto-refresh to store metadata for only the subset of columns. This flag and list can be used to verify the refresh with a subset of columns functionality. But again this will be done in a different Jira in the future.

arina-ielchiieva:
We can unit test that syntax works fine without checking file metadata content, there are at least three options on how command can be written. It's much better than not writing tests at all.

amansinha100:
@dvjyothsna based on @arina-ielchiieva 's last suggestion, did you get a chance to add unit tests for the command syntax (for the 3 variations) ?

dvjyothsna:
@arina-ielchiieva, @amansinha100 I was held up with some other issue. Added some unit tests just to test the refresh command and I think more rigorous tests will be included in the future jiras where there will be some changes to the metadata cache file structure.

dvjyothsna:
Thank you, Arina for the review!!

PR #1672, opened 2019-03-02
DRILL-7049 return VARBINARY as a string with escaped non printable bytes

jcmcote:
fix bug in string_binary, it was not doing reallocIfNeeded

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/util/ValueVectorElementFormatter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/util/ValueVectorElementFormatter.java
@@ -52,28 +52,47 @@ public class ValueVectorElementFormatter {
    * @return the formatted value, null if failed
    */
   public String format(Object value, TypeProtos.MinorType minorType) {
+    boolean handled = false;
+	String str = null;
     switch (minorType) {
       case TIMESTAMP:
         if (value instanceof LocalDateTime) {
-          return format((LocalDateTime) value,
+          handled = true;
+          str = format((LocalDateTime) value,
                         options.getString(ExecConstants.WEB_DISPLAY_FORMAT_TIMESTAMP),
                         (v, p) -> v.format(getTimestampFormatter(p)));
         }
+        break;
       case DATE:
         if (value instanceof LocalDate) {
-          return format((LocalDate) value,
+          handled = true;
+          str = format((LocalDate) value,
                         options.getString(ExecConstants.WEB_DISPLAY_FORMAT_DATE),
                         (v, p) -> v.format(getDateFormatter(p)));
         }
+        break;
       case TIME:
         if (value instanceof LocalTime) {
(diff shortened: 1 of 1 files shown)


arina-ielchiieva:
@vdiravka please review.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ValueVectorElementFormatter.java
```diff
@@ -52,28 +52,51 @@ public ValueVectorElementFormatter(OptionManager options) {
...
+        }
+        else {
+          str = value.toString();
+        }
+      break;
       default:
-        return value.toString();
+        str = value.toString();
     }
+    return str;
```
vdiravka: Looks like you can remove all `else` blocks and put here `return value.toString();`.
_Note:_ `break` statements should be present (as is in your code), return statements should be preserved as earlier, default block can be omitted.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ValueVectorElementFormatter.java
```diff
@@ -52,28 +52,42 @@ public ValueVectorElementFormatter(OptionManager options) {
...
+        break;
+      case VARBINARY:
+        if (value instanceof byte[]) {
+          byte[] bytes = (byte[]) value;
+          str = org.apache.drill.common.util.DrillStringUtils.toBinaryString(bytes);
+        }
+        break;
+    }
+
+    if(str == null) {
```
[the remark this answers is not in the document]
agozhiy: I agree. It is intended to work as follows: if the format is incorrect, then the exception is being logged and the formatter returns null. Returning values with default formatting can be confusing. And if we return an error message, that would be a column type violation.

exec/java-exec/src/main/java/org/apache/drill/exec/util/ValueVectorElementFormatter.java
```diff
@@ -52,28 +52,47 @@ public ValueVectorElementFormatter(OptionManager options) {
...
+      case VARBINARY:
+        if (value instanceof byte[]) {
+          handled = true;
+          byte[] bytes = (byte[]) value;
+          str = org.apache.drill.common.util.DrillStringUtils.toBinaryString(bytes);
+        }
+        break;
+    }
+
+    if (!handled) {
```
vdiravka: It looks like current code execution is the same as in your PR. But logic from PR is more complex: additional flag `handled`, breaks in switch statements...
I think we can leave current code from Drill master and add  your `case VARBINARY`.
Seems it will be enough.

arina-ielchiieva:
@vdiravka could you please prepare branch with original commit and commit with the fixes on top of it? Looks like @jcmcote is not resounding but we still want to merge the changes.

cc @sohami

PR #1676, opened 2019-03-04
DRILL-7074: Scan framework fixes and enhancements

paul-rogers:
This PR is a roll-up of a number of fixes and enhancements to the new “scan framework” that resulted from the effort to migrate the CSV (compliant text) reader to the new framework. To keep reviews manageable, this PR includes just the scan framework changes. Since this framework is not yet used in any “production” code, this part of the review should be relatively easy. A separate PR will provide the reposted CSV reader.

Changes include:

* More documentation
* Improved support for the “schema only” first batch that (it seems) downstream operators need.
* Support for the original behavior of the wildcard with partitioned columns, and the proposed behavior identified in DRILL-7055. Restores behavior previously removed. An option switches between the two behaviors.
* More permissive combinations of the `columns` column and the wildcard. (Turns out that current behavior ,as show, by unit tests, is pretty varied.)
* Several renames and refactorings
* Fixed a bug in how the projection set is passed to the result set loader.
* Multiple new or revised unit tests.


paul-rogers:
Passed most tests on my box except for this one that fails randomly from time to time:

```
[ERROR] Errors: 
[ERROR]   TestPStoreProviders.verifyZkStore:67 » NoSuchElement
```

arina-ielchiieva:
+1, LGTM.

PR #1681, opened 2019-03-06
DRILL-7051: Upgrade to Jetty 9.3 

vdiravka:
- upgrade Jetty dependencies to 9.3 version
- downgrade Jersey dependencies to 2.25.1 version
- add JavaDocs and code refactoring
- adaptation to the new Jetty API - SessionHandler, LoginService, AbstractLoginService (jetty 9.4 only).
For Jetty and Jersey versions update to the latest versions [DRILL-7135](https://issues.apache.org/jira/browse/DRILL-7135) is created.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/auth/DrillSpnegoAuthenticator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/auth/DrillSpnegoAuthenticator.java
@@ -29,6 +29,7 @@ import org.eclipse.jetty.security.authentication.SessionAuthentication;
 import org.eclipse.jetty.security.authentication.SpnegoAuthenticator;
 import org.eclipse.jetty.server.Authentication;
 import org.eclipse.jetty.server.HttpChannel;
+import org.eclipse.jetty.server.HttpConnection;
 import org.eclipse.jetty.server.Request;
 import org.eclipse.jetty.server.Response;
 import org.eclipse.jetty.server.UserIdentity;
@@ -152,7 +153,7 @@ public class DrillSpnegoAuthenticator extends SpnegoAuthenticator {
         }
 
         response.setContentLength(0);
-        final HttpChannel channel = HttpChannel.getCurrentHttpChannel();
+        final HttpChannel channel = HttpConnection.getCurrentConnection().getHttpChannel();
         final Response base_response = channel.getResponse();
         final Request base_request = channel.getRequest();
         final int redirectCode =
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/auth/DrillSpnegoAuthenticator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/auth/DrillSpnegoAuthenticator.java
@@ -28,10 +28,7 @@ import org.eclipse.jetty.security.authentication.DeferredAuthentication;
 import org.eclipse.jetty.security.authentication.SessionAuthentication;
 import org.eclipse.jetty.security.authentication.SpnegoAuthenticator;
 import org.eclipse.jetty.server.Authentication;
-import org.eclipse.jetty.server.HttpChannel;
-import org.eclipse.jetty.server.HttpConnection;
 import org.eclipse.jetty.server.Request;
-import org.eclipse.jetty.server.Response;
 import org.eclipse.jetty.server.UserIdentity;
 
 import javax.servlet.ServletRequest;
@@ -151,15 +148,12 @@ public class DrillSpnegoAuthenticator extends SpnegoAuthenticator {
             newUri = WebServerConstants.WEBSERVER_ROOT_PATH;
           }
         }
-
         response.setContentLength(0);
-        final HttpChannel channel = HttpConnection.getCurrentConnection().getHttpChannel();
-        final Response base_response = channel.getResponse();
-        final Request base_request = channel.getRequest();
-        final int redirectCode =
-            base_request.getHttpVersion().getVersion() < HttpVersion.HTTP_1_1.getVersion() ? 302 : 303;
+        Request baseRequest = Request.getBaseRequest(req);
+        int redirectCode =
+            baseRequest.getHttpVersion().getVersion() < HttpVersion.HTTP_1_1.getVersion() ? 302 : 303;
         try {
-          base_response.sendRedirect(redirectCode, res.encodeRedirectURL(newUri));
+          baseRequest.getResponse().sendRedirect(redirectCode, res.encodeRedirectURL(newUri));
         } catch (IOException e) {
           logger.error("DrillSpnegoAuthenticator: Failed while using the redirect URL {} from client {}", newUri,
               req.getRemoteAddr(), e);
--- a/pom.xml
+++ b/pom.xml
@@ -2529,33 +2529,6 @@
           </dependency>
           <!--/GlassFish Jersey dependecies-->
 
-          <!--Javax dependecies-->
-          <dependency>
-            <groupId>javax.servlet.jsp</groupId>
-            <artifactId>javax.servlet.jsp-api</artifactId>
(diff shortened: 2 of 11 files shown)


vdiravka:
@sohami Could you please review?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/auth/DrillSpnegoAuthenticator.java
```diff
@@ -152,7 +153,7 @@ private Authentication authenticateSession(ServletRequest request, ServletRespon
         }
 
         response.setContentLength(0);
-        final HttpChannel channel = HttpChannel.getCurrentHttpChannel();
+        final HttpChannel channel = HttpConnection.getCurrentConnection().getHttpChannel();
         final Response base_response = channel.getResponse();
         final Request base_request = channel.getRequest();
```
sohami: Please update this as below:
```
final Request base_request = Request.getBaseRequest(req);
final Response base_response = base_request.getResponse();
```

pom.xml
```diff
@@ -85,6 +85,7 @@
     <reflections.version>0.9.10</reflections.version>
     <avro.version>1.8.2</avro.version>
     <metrics.version>4.0.2</metrics.version>
+    <jetty.version>9.4.15.v20190215</jetty.version>
```
sohami: I think this should still be defined in` java-exec pom.xml` file since it's only used in that package. Root pom should control the common dependencies across multiple packages.

vdiravka: In current PR I have added jetty-dependencies to the `dependencyManagement` block only.
It was done in scope of DRILL-68.. I have added a [comment](https://issues.apache.org/jira/browse/DRILL-6540?focusedCommentId=16606306&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16606306) that it was necessary for the HBase unit tests. So it can't be excluded.
Also I checked `mvn dependency:tree -Dincludes=org.eclipse.jetty` without `dependencyManagement` block and found different versions of Jetty

```
[INFO] org.apache.drill.contrib:drill-format-mapr:jar:1.16.0-SNAPSHOT
[INFO] +- com.mapr.fs:mapr-hbase:jar:6.1.0-mapr:compile
[INFO] |  \- org.apache.hbase:hbase-server:jar:2.1.1:compile
[INFO] |     +- org.apache.hbase:hbase-http:jar:2.1.1:compile
[INFO] |     |  +- org.eclipse.jetty:jetty-util:jar:9.3.19.v20170502:compile
[INFO] |     |  \- org.eclipse.jetty:jetty-util-ajax:jar:9.3.19.v20170502:compile
[INFO] |     \- org.eclipse.jetty:jetty-webapp:jar:9.3.19.v20170502:compile
[INFO] |        \- org.eclipse.jetty:jetty-xml:jar:9.3.19.v20170502:compile
[INFO] \- org.apache.drill.exec:drill-java-exec:jar:1.16.0-SNAPSHOT:compile
[INFO]    +- org.eclipse.jetty:jetty-server:jar:9.4.15.v20190215:compile
[INFO]    |  +- org.eclipse.jetty:jetty-http:jar:9.4.15.v20190215:compile
[INFO]    |  \- org.eclipse.jetty:jetty-io:jar:9.4.15.v20190215:compile
[INFO]    +- org.eclipse.jetty:jetty-servlet:jar:9.4.15.v20190215:compile
[INFO]    |  \- org.eclipse.jetty:jetty-security:jar:9.4.15.v20190215:compile
[INFO]    +- org.eclipse.jetty:jetty-servlets:jar:9.4.15.v20190215:compile
[INFO]    \- org.glassfish.jersey.containers:jersey-container-jetty-servlet:jar:2.8:compile
[INFO]       \- org.glassfish.jersey.containers:jersey-container-jetty-http:jar:2.8:compile
[INFO]          \- org.eclipse.jetty:jetty-continuation:jar:9.1.1.v20140108:compile
[INFO] ------------------------------------------------------------------------
```
Therefore since Jetty dependencies are used not only in `java-exec` it is nesessary to handle their versions in Drill root POM file.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/auth/DrillSpnegoAuthenticator.java
vdiravka: Thanks, it looks better.
Also I have dropped usage of `base_response` local variable.

pom.xml
sohami: Since during Drill's upgrade of Jetty version there was code change needed due to api changes. I am expecting even though if we enforce Jetty version to be `9.4.15` from root pom dependencyManagement block, it can cause issues for drill-format-mapr jar which is dependent upon mapr-hbase which in turn on jetty 9.3.19. There might be some api compatibility mismatch due to Jetty version upgrade.

vdiravka:
@sohami I have resolved conflict with master branch. Could you please review the changes made.

pom.xml
sohami: what is your thought on this ?

vdiravka: I will investigate how hbase and mapr-db jetty dependencies affect Drill.

vdiravka: It looks like HBase was earlier on Jetty 9.4 version, but there is an issue with `MiniDFSCluster` [HBASE-18943](https://issues.apache.org/jira/browse/HBASE-18943), [HBASE-19390](https://issues.apache.org/jira/browse/HBASE-19390), so they reverted to 9.3 Jetty version. It was the same reason why I have updated Jetty in the scope of [DRILL-6540: Upgrade to HADOOP-3.1 libraries](https://issues.apache.org/jira/browse/DRILL-6540).
It also appears that new Jersey version requires Jetty 9.4 version, see [comment](https://github.com/apache/hbase/blob/master/pom.xml#L1533) in HBase.
Possibly we should consider the newest Jetty and Jersey versions only after [HBASE-19256](https://issues.apache.org/jira/browse/HBASE-19256) resolution.

So I will check my changes for Hadoop3.1 libs version in Drill with newest Jetty and Jersey. If there are some issues with them, I will put to Drill the same Jetty and Jersey versions as in HBase project.

sohami: So with newest version of Jetty I am guessing you mean 9.3 since that is what is required by MiniDfsCluster test in Hadoop 3.1 as well ? It won't work with anything beyond that version unless HBASE-19256 is resolved.

vdiravka: Finally jetty 9.3 is chosen for Drill.
Jetty dependencies are used in `java-exec pom.xml`, but I've left versions control in `dependencyManagement` block of root POM to avoid picking invalid jetty version by maven, in case when some libs will have other version.
For instance we can't exclude jetty from `hadoop-common` and `hbase` dependencies. But they have different jetty minor versions: 
[9.3.24.v20180605](https://github.com/apache/hadoop/blob/trunk/hadoop-project/pom.xml#L38)  for Hadoop
[9.3.19.v20170502](https://github.com/apache/hbase/blob/rel/2.1.0/pom.xml#L1352) for HBase 2.1 and [9.3.25.v20180904](https://github.com/apache/hbase/blob/master/pom.xml#L1529) for master HBase version.
I didn't find any API compatibility differences between these jetty minor versions (only 9.4 has it).
Possibly in future we can consider shade Jetty version in Drill, [DRILL-7135](https://issues.apache.org/jira/browse/DRILL-7135). Not sure that is necessary for now.

pom.xml
```diff
@@ -2553,7 +2548,13 @@
             <version>4.0.1</version>
             <scope>provided</scope>
           </dependency>
-          <!--/Javax Servlet dependecies-->
+
+          <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
```
sohami: why we need this and other Javax Servlet dependencies ?

PR #1682, opened 2019-03-06
 DRILL-7081: Upgrade GlassFish Jersey and Javax Servlet dependecies 

vdiravka:
- Update dependencies versions
     * update GlassFish Jersey and Javax Servlet dependecies versions
     * add `org.glassfish.jersey.inject:jersey-hk2`, since hard dependency was removed from Jersey, starting from [Jersey 2.26](https://jersey.github.io/release-notes/2.26.html).
- Exclude dependency from jdbc-all
- Removal redundant "bcpkix-jdk15on" exclusion
     * it is mentioned twice in the code
- Proper exclusion of "jackson-dataformat-hocon" dependency
     * since it is direct dependency of `java-exec` it should be excluded from `java-exec` directly, not in maven-shade-plugin
- Removal redundant "excludeSubprojects" config property for Maven Rat Plugin
    * `excludeSubprojects` name is invalid, the proper one is [`excludeSubProjects`](http://creadur.apache.org/rat/apache-rat-plugin/rat-mojo.html#excludeSubProjects) with capital `P`. Since it is not valid it is ignored by Maven Rat Plugin (true is default value). If it is false, all exclusions from license check should be specified in root POM file. But currently in Drill different exclusions are specified in different modules. So the proper value is `true` (default value). So the property can be removed.
    * `rat.excludeSubprojects` properties are removed too, since `true` is the default value.

### DIFF
--- a/exec/jdbc-all/pom.xml
+++ b/exec/jdbc-all/pom.xml
@@ -81,12 +81,24 @@
           <artifactId>commons-codec</artifactId>
         </exclusion>
         <exclusion>
-          <groupId>org.bouncycastle</groupId>
-          <artifactId>bcpkix-jdk15on</artifactId>
+          <groupId>org.glassfish.jersey.containers</groupId>
+          <artifactId>jersey-container-jetty-servlet</artifactId>
         </exclusion>
         <exclusion>
-          <artifactId>jersey-container-jetty-servlet</artifactId>
-          <groupId>org.glassfish.jersey.containers</groupId>
+          <groupId>org.glassfish.jersey.media</groupId>
+          <artifactId>jersey-media-multipart</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.glassfish.jersey.ext</groupId>
+          <artifactId>jersey-mvc-freemarker</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.glassfish.jersey.media</groupId>
+          <artifactId>jersey-media-json-jackson</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.glassfish.jersey.inject</groupId>
+          <artifactId>jersey-hk2</artifactId>
         </exclusion>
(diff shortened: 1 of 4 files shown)


vdiravka:
@sohami Could you please take a look?

exec/jdbc-all/pom.xml
```diff
@@ -576,7 +576,7 @@
                           This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                         </message>
-                        <maxsize>37500000</maxsize>
+                        <maxsize>38600000</maxsize>
```
sohami: are these due to both `jetty` and `jersey` dependencies being pulled in ? We should run mvn dependency:tree command with and without the PR to see.

vdiravka: I forgot to check jdbc-all exclusions.
So `jetty` was fully excluded before updates. And almost all `Glassfish Jersey` dependencies were excluded. But I had added new one without exclusion - `org.glassfish.jersey.inject:jersey-hk2`.
For now it is excluded and jar file size can be untouched.

vdiravka:
@sohami I have solved the issue with jdbc-all jar size and removed dependency on Update Jetty #1681 commit from this PR.
So you can review it independently from #1681

exec/jdbc-all/pom.xml
```diff
@@ -516,7 +518,7 @@
                   This is likely due to you adding new dependencies to a java-exec and not updating the excludes in this module. This is important as it minimizes the size of the dependency of Drill application users.
 
                   </message>
-                  <maxsize>40000000</maxsize>
+                  <maxsize>41600000</maxsize>
```
sohami: How about this one ?

vdiravka: Sorry, overlooked it. There is no need more in this increase.

PR #1686, opened 2019-03-09
DRILL-6524: Prevent incorrect scalar replacement for the case of assigning references inside if block

vvysotskyi:
This pull request contains two commits:
- The first one contains changes to prevent incorrect scalar replacement for the case of assigning references inside if block to avoid wrong results for some cases
- The second commit contains changes to assign holder fields instead of assigning just object reference for the case when the source object is not used anymore (a common example is the result of UDF execution), so it will allow producing scalar replacement more frequently.

Therefore with these two commits, a scalar replacement will be produced correctly and more frequently despite preventing it in the first commit.

For detailed problem description please see comment in [DRILL-6524](https://issues.apache.org/jira/browse/DRILL-6524)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/compile/bytecode/MethodAnalyzer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/compile/bytecode/MethodAnalyzer.java
@@ -17,16 +17,26 @@
  */
 package org.apache.drill.exec.compile.bytecode;
 
+import org.objectweb.asm.tree.AbstractInsnNode;
+import org.objectweb.asm.tree.InsnList;
+import org.objectweb.asm.tree.LabelNode;
+import org.objectweb.asm.tree.MethodNode;
 import org.objectweb.asm.tree.analysis.Analyzer;
+import org.objectweb.asm.tree.analysis.AnalyzerException;
 import org.objectweb.asm.tree.analysis.Frame;
 import org.objectweb.asm.tree.analysis.Interpreter;
 import org.objectweb.asm.tree.analysis.Value;
+
+import java.util.ArrayDeque;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.Set;
 
 /**
  * Analyzer that allows us to inject additional functionality into ASMs basic analysis.
  *
  * <p>We need to be able to keep track of local variables that are assigned to each other
- * so that we can infer their replacability (for scalar replacement). In order to do that,
+ * so that we can infer their replaceability (for scalar replacement). In order to do that,
  * we need to know when local variables are assigned (with the old value being overwritten)
  * so that we can associate them with the new value, and hence determine whether they can
  * also be replaced, or not.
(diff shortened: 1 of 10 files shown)


arina-ielchiieva:
@vdiravka could you please review?

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestClassTransformation.java
```diff
@@ -79,11 +86,10 @@ public void testCompilationNoDebug() throws CompileException, ClassNotFoundExcep
     ClassSet classSet = new ClassSet(null, cg.getDefinition().getTemplateClassName(), cg.getMaterializedClassName());
     String sourceCode = cg.generateAndGet();
     sessionOptions.setLocalOption(ClassCompilerSelector.JAVA_COMPILER_OPTION, ClassCompilerSelector.CompilerPolicy.JDK.name());
-
     sessionOptions.setLocalOption(ClassCompilerSelector.JAVA_COMPILER_DEBUG_OPTION, false);
-    @SuppressWarnings("resource")
```
vdiravka: Is this warning displayed now?

exec/vector/src/main/java/org/apache/drill/exec/vector/ValueHolderHelper.java
```diff
@@ -229,4 +231,77 @@ public static VarDecimalHolder getVarDecimalHolder(DrillBuf buf, BigDecimal bigD
...
+          case VAR16CHAR:
+          case VARBINARY:
+          case VARCHAR:
+            result.add("start");
+            result.add("end");
+            result.add("buffer");
+            return result;
+          case UNION:
+            result.add("reader");
+            return result;
```
vdiravka: What about additional datatypes: `MAP`, `LIST`, `NULL` ( and others from `Types.getBaseSqlTypeName()`)?
Is `emptyResult` fine for them?

exec/java-exec/src/main/java/org/apache/drill/exec/compile/bytecode/MethodAnalyzer.java
```diff
@@ -36,70 +46,147 @@
  * as factories that will provide our own derivative of Frame<> which we use to detect
  */
 public class MethodAnalyzer<V extends Value> extends Analyzer <V> {
+
+  // list of method instructions which is analyzed
+  private InsnList insnList;
+
+  /**
+   * Constructor.
```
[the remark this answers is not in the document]
vvysotskyi: Thanks, removed.

exec/java-exec/src/main/java/org/apache/drill/exec/compile/bytecode/MethodAnalyzer.java
```diff
@@ -36,70 +46,147 @@
...
-   * that are assigned to the same local variable slot.
+   * that are assigned to the same local variable slot. Also it controls stack to determine whether
+   * object was assigned to the value declared outside of conditional block.
    *
    * <p>Since this is almost a pass-through, the constructors' arguments match
    * those from Frame<>.
    */
   private static class AssignmentTrackingFrame<V extends Value> extends Frame<V> {
+
+    // represents stacks of variable sets declared inside current code block
```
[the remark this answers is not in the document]
vvysotskyi: Thanks, done.

exec/java-exec/src/test/java/org/apache/drill/exec/compile/TestClassTransformation.java
vvysotskyi: No, it is not displayed, after removing this line.

exec/vector/src/main/java/org/apache/drill/exec/vector/ValueHolderHelper.java
vvysotskyi: They are handled separately in the code generation classes, so there is no need to specify them here.

PR #1688, opened 2019-03-11
DRILL-7092: Rename map to struct in schema definition

arina-ielchiieva:
1. Renamed map to struct in schema parser.
2. Updated sqlTypeOf function to return STRUCT instead of MAP, drillTypeOf function will return MAP as before until internal renaming is done.
3. Add is_struct alias to already existing is_map function. Function should be revisited once Drill supports true maps.
4. Updated unit tests.

Details in Jira [DRILL-7092](https://issues.apache.org/jira/browse/DRILL-7092).

### DIFF
--- a/common/src/main/java/org/apache/drill/common/types/Types.java
+++ b/common/src/main/java/org/apache/drill/common/types/Types.java
@@ -185,7 +185,7 @@ public class Types {
       // Composite types and other types that are not atomic types (SQL standard
       // or not) except ARRAY types (handled above):
 
-      case MAP:             return "MAP";
+      case MAP:             return "STRUCT"; // Drill map represents struct and in future will be renamed
       case LATE:            return "ANY";
       case NULL:            return "NULL";
       case UNION:           return "UNION";
@@ -267,7 +267,7 @@ public class Types {
       case "INTERVAL":                      return java.sql.Types.OTHER;  // JDBC (4.1) has nothing for INTERVAL
       case "INTERVAL YEAR TO MONTH":        return java.sql.Types.OTHER;
       case "INTERVAL DAY TO SECOND":        return java.sql.Types.OTHER;
-      case "MAP":                           return java.sql.Types.OTHER; // Drill doesn't support java.sql.Struct
+      case "STRUCT":                        return java.sql.Types.OTHER; // Drill doesn't support java.sql.Struct
       case "NATIONAL CHARACTER VARYING":    return java.sql.Types.NVARCHAR;
       case "NATIONAL CHARACTER":            return java.sql.Types.NCHAR;
       case "NULL":                          return java.sql.Types.NULL;
--- a/exec/java-exec/src/main/antlr4/org/apache/drill/exec/record/metadata/schema/parser/SchemaLexer.g4
+++ b/exec/java-exec/src/main/antlr4/org/apache/drill/exec/record/metadata/schema/parser/SchemaLexer.g4
@@ -55,6 +55,7 @@ MINUTE: 'MINUTE';
 SECOND: 'SECOND';
 
 MAP: 'MAP';
+STRUCT: 'STRUCT';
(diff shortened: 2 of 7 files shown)


arina-ielchiieva:
@vvysotskyi / @paul-rogers could you please take a look?

ihuzenko:
> Yes, indeed there has long been a misconception around the use of the type "MAP" in Drill. It is, indeed, closer to a Hive and Arrow struct. Since it is unclear how would create a true map using vectors, I suppose that, originally, merging the map and struct concepts was considered a harmless simplification.
> 
> While it is easy to rename the items in this new parser code, it will be far harder to rename the many existing vectors, metadata elements and so on. Even harder still to rename to SQL data type itself. Fortunately, it seems that the only place the term "Map" is surfaced to the user in the code is in the "typeof" set of functions. Up until now, Drill did not have many DDL statements.
> 
> So, the change here LGTM, but I wonder how we plan to roll out the changes to other parts of the code. Would that be a breaking change for 2.0?
> 
> Finally, if this rename is done to prepare for introducing a true map type (collection of key/value pairs where the set of values is uncorrelated across rows), then is there a design for that? Would be good to start thinking how that would fit into the new row-set framework which is heavily based on the current "struct" design.
> 
> +1

Hello @paul-rogers , yes both changes (renaming and new MapVector) are in our future plans. Today I've created two related Jira tickets [DRILL-7097](https://issues.apache.org/jira/browse/DRILL-7097) and [DRILL-7096](https://issues.apache.org/jira/browse/DRILL-7096) . For now it's pretty raw ideas, but it would be very good if you can join discussion in jira and share your thoughts with me and @KazydubB .

ihuzenko:
Hello @paul-rogers , yes both changes (renaming and new MapVector) are in our future plans. Today I've created two related Jira tickets [DRILL-7097](https://issues.apache.org/jira/browse/DRILL-7097) and [DRILL-7096](https://issues.apache.org/jira/browse/DRILL-7096) . For now it's pretty raw ideas, but it would be very good if you can join discussion in jira and share your thoughts with me and @KazydubB .

ihuzenko:
> Yes, indeed there has long been a misconception around the use of the type "MAP" in Drill. It is, indeed, closer to a Hive and Arrow struct. Since it is unclear how would create a true map using vectors, I suppose that, originally, merging the map and struct concepts was considered a harmless simplification.
> 
> While it is easy to rename the items in this new parser code, it will be far harder to rename the many existing vectors, metadata elements and so on. Even harder still to rename to SQL data type itself. Fortunately, it seems that the only place the term "Map" is surfaced to the user in the code is in the "typeof" set of functions. Up until now, Drill did not have many DDL statements.
> 
> So, the change here LGTM, but I wonder how we plan to roll out the changes to other parts of the code. Would that be a breaking change for 2.0?
> 
> Finally, if this rename is done to prepare for introducing a true map type (collection of key/value pairs where the set of values is uncorrelated across rows), then is there a design for that? Would be good to start thinking how that would fit into the new row-set framework which is heavily based on the current "struct" design.
> 
> +1

    Hello @paul-rogers , yes both changes (renaming and new MapVector) are in our future plans. Today I've created two related Jira tickets [DRILL-7097](https://issues.apache.org/jira/browse/DRILL-7097) and [DRILL-7096](https://issues.apache.org/jira/browse/DRILL-7096) . For now it's pretty raw ideas, but it would be very good if you can join discussion in jira and share your thoughts with me and @KazydubB .

ihuzenko:
Hello @paul-rogers , yes both changes (renaming and new MapVector) are in our future plans. Today I've created two related Jira tickets [DRILL-7097](https://issues.apache.org/jira/browse/DRILL-7097) and [DRILL-7096](https://issues.apache.org/jira/browse/DRILL-7096) . For now it's pretty raw ideas, but it would be very good if you can join discussion in jira and share your thoughts with me and @KazydubB .

PR #1689, opened 2019-03-12
DRILL-7061: Setting limit results to 1000 on web UI causes parse error

kkhatua:
Freemarker by default introduces a comma in numeric values greater than 999. This corrects that by removing the ',' in the default limit size


kkhatua:
@sohami it looks like the `wrapQuery()` function was accidentally removed in DRILL-7036 (UI Alert Improvement) #1644 . I've restored the functionality. 
DRILL-6960 (#1608) should address the bigger issue separately.

sohami:
@kkhatua - Since client side limit feature doesn't handle all the queries it would be better to disable this feature until DRILL-6960 is ready. If for any reason DRILL-6960 is delayed post 1.16.0 then just having client side change can create confusion on what are supported set of queries with autoLimit feature.

PR #1690, opened 2019-03-12
DRILL-7086: Output schema for row set mechanism

paul-rogers:
Enhances the row set mechanism to take an "output schema" that describes the vectors to create. The "input schema" describes the type that the reader would like to write. A conversion mechanism inserts a conversion shim to convert from the input to output type.

The "output schema" will be the one provided by the new schema mechanism: a later PR will connect the mechanism here with the "output schema" provided in the physical plan.

The output schema is optional. Within an output schema, the columns are optional. If not output schema is provided, or no column matches a given input column, then the input schema determines vector type as was the case before this enhancement.

This PR includes several "starter" conversion classes. These are mostly prototypes, only the string-to-int version has been fully tested. The date, time and date time versions show the use of the new format property added to the column metadata class.

A format or storage plugin can specify its own conversion rules. For example, the plugin for HBase might provide byte-array-to-whatever converters.

An odd aspect of the current implementation is that the type conversion is done at reader open time. If the reader detects a column type which cannot be converted, using known rules, to the desired output type, the query will fail. This is odd because one might expect this error to be caught at plan time. But, Drill, of course, is schema-on-read, so read time is when we'd detect the conflict. May not be a problem in real tables.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/OptionBuilder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/OptionBuilder.java
@@ -25,10 +25,11 @@ import org.apache.drill.exec.physical.rowSet.impl.ResultSetLoaderImpl.ResultSetO
 import org.apache.drill.exec.physical.rowSet.project.RequestedTuple;
 import org.apache.drill.exec.record.metadata.TupleMetadata;
 import org.apache.drill.exec.vector.ValueVector;
+import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
 
 /**
  * Builder for the options for the row set loader. Reasonable defaults
- * are provided for all options; use these options for test code or
+ * are provided for all options; use the default options for test code or
  * for clients that don't need special settings.
  */
 
@@ -40,8 +41,10 @@ public class OptionBuilder {
   protected ResultVectorCache vectorCache;
   protected TupleMetadata schema;
   protected long maxBatchSize;
+  protected SchemaTransformer schemaTransformer;
 
   public OptionBuilder() {
+    // Start with the default option values.
     ResultSetOptions options = new ResultSetOptions();
     vectorSizeLimit = options.vectorSizeLimit;
     rowCountLimit = options.rowCountLimit;
@@ -52,8 +55,7 @@ public class OptionBuilder {
    * Specify the maximum number of rows per batch. Defaults to
    * {@link BaseValueVector#INITIAL_VALUE_ALLOCATION}. Batches end either
    * when this limit is reached, or when a vector overflows, whichever
-   * occurs first. The limit is capped at
-   * {@link ValueVector#MAX_ROW_COUNT}.
+   * occurs first. The limit is capped at {@link ValueVector#MAX_ROW_COUNT}.
    *
    * @param limit the row count limit
    * @return this builder
@@ -129,10 +131,23 @@ public class OptionBuilder {
     return this;
   }
 
+  /**
+   * Provide an optional higher-level schema transformer which can convert
+   * columns from one type to another.
+   *
+   * @param transform the column conversion factory
+   * @return this builder
+   */
+  public OptionBuilder setSchemaTransform(SchemaTransformer transform) {
+    schemaTransformer = transform;
+    return this;
+  }
+
   // TODO: No setter for vector length yet: is hard-coded
   // at present in the value vector.
 
   public ResultSetOptions build() {
+    Preconditions.checkArgument(projection == null || projectionSet == null);
     return new ResultSetOptions(this);
   }
 }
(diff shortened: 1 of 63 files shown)


exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/AbstractWriteConverter.java
```diff
@@ -38,12 +41,38 @@
...
+      } catch (NoSuchMethodException | SecurityException e) {
+        throw new IllegalStateException(e);
+      }
+    }
+
+    @Override
+    public AbstractWriteConverter newWriter(ScalarWriter baseWriter) {
+      try {
+        return ctor.newInstance(baseWriter);
+      } catch (InstantiationException | IllegalAccessException
```
arina-ielchiieva:
```suggestion
      } catch (ReflectiveOperationException e) {
```

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/ConvertStringToTimeStamp.java
```diff
@@ -0,0 +1,55 @@
...
+package org.apache.drill.exec.vector.accessor.convert;
+
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+
+import org.apache.drill.exec.vector.accessor.ScalarWriter;
+
+/**
+ * Convert a VARCHAR column to an TIMESAMP column following the Java rules
```
arina-ielchiieva:
```suggestion
 * Convert a VARCHAR column to an TIMESTAMP column following the Java rules
```

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/package-info.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Defines the type conversion mechanism along with several basic
+ * conversions.
+ */
+package org.apache.drill.exec.vector.accessor.convert;
```
arina-ielchiieva: Please add conversion for decimal and float.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/impl/OptionBuilder.java
```diff
@@ -129,10 +133,60 @@ public OptionBuilder setSchema(TupleMetadata schema) {
     return this;
   }
 
+  /**
+   * Provide an optional column conversion factory. Can create a "shim"
```
arina-ielchiieva: From the description is not quite clear to me why we need to both options: column conversions and schema transform. Could not we have only one that would function as both? Can you please explain a little bit more?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/StandardConversions.java
```diff
@@ -0,0 +1,238 @@
...
+    case VARCHAR:
+      return new ConversionDefn(convertFromVarchar(outputSchema));
+    case TINYINT:
+      switch (outputSchema.type()) {
+      case SMALLINT:
+      case INT:
+      case BIGINT:
+      case FLOAT4:
+      case FLOAT8:
+      case DECIMAL18:
```
arina-ielchiieva: Please exclude all decimal minor types except of `VARDECIMAL`, they are deprecated.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/package-info.java
paul-rogers: Now is "later"... Added those which are type-save (widening conversions.) How do we want to handle overflow in narrowing conversions? SQL says overflow is an error, so do we want to fail the query? Set the value to null? Truncate the high-order bits?

For float-to-integral conversions, SQL says that the behavior is implementation-specific. Do we want to round using Java semantics?

paul-rogers:
Added more commits to add more tests and better support date/time types. See commit message for details.

Have not squashed commits to make it easier to review recent additions to this PR.

paul-rogers:
For narrowing conversions, I ended up with a bit of an ad-hoc solution that follows the Java conventions.

* When converting from int to TINYINT, or SMALLINT, the upper bits are simply truncated. (The code uses a setInt() accessor for all these types. No checking is done to maximize performance.)
* When converting from long to int, an exception is thrown for overflow. (The "natural" method for an int is setInt(), so the setLong() uses the Java Math.toExactInt() method which throws an exception for overflow.)
* When converting from double to BigInt, follows the Java convention of converting to the closest available value.
* When converting from double to other integral types, first follows the double-to-BigInt rules, then the long-to-int rule.

These are the simplest, fastest conversions based on the idea that data should be correct for the target type. We can tinker with the overflow logic over time without affecting the "happy path" for correct values.

arina-ielchiieva:
Looks good, +1

paul-rogers:
@arina-ielchiieva, thanks for the review. Squashed commits. Ready to commit.

PR #1691, opened 2019-03-12
DRILL-7093 Batch Sizing in SingleSender

bitblender:
This PR contains changes for implementing batch sizing in SingleSender. The sizing is done by using the RecordBatchSizer.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
@@ -838,4 +838,13 @@ public class MergingRecordBatch extends AbstractRecordBatch<MergingReceiverPOP> 
         container, outgoingPosition, Arrays.toString(incomingBatches), Arrays.toString(batchOffsets),
         Arrays.toString(tempBatchHolder), Arrays.toString(inputCounts), Arrays.toString(outputCounts));
   }
+
+  @Override
+  public WritableBatch getWritableBatch(int startIndex, int length) {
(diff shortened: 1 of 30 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/mergereceiver/MergingRecordBatch.java
```diff
@@ -838,4 +838,13 @@ public void dump() {
         container, outgoingPosition, Arrays.toString(incomingBatches), Arrays.toString(batchOffsets),
         Arrays.toString(tempBatchHolder), Arrays.toString(inputCounts), Arrays.toString(outputCounts));
   }
+
+  @Override
+  public WritableBatch getWritableBatch(int startIndex, int length) {
+    VectorContainer partialContainer = new VectorContainer(context.getAllocator(), getSchema());
```
HanumathRao: Looks like getWritableBatch in MergingRecordBatch, SpilledRecordBatch, AbstractRecordBatch and RecordBatchLoader is same.  Is there any way to write a common function in the hierarchy and use it in all these scenarios?

exec/java-exec/src/main/java/org/apache/drill/exec/record/SimpleRecordBatch.java
```diff
@@ -88,6 +88,11 @@ public WritableBatch getWritableBatch() {
     throw new UnsupportedOperationException();
   }
 
+  @Override
+  public WritableBatch getWritableBatch(int start, int length) {
+    throw new UnsupportedOperationException();
```
HanumathRao: Should there be a string in similar lines to that of SchemalessBatch exception.

HanumathRao:
Code changes look good to me. I have few minor comments.

PR #1695, opened 2019-03-14
DRILL-7103: Incorrect conversion to integer in BsonRecordReader#writeTimeStamp

oleg-zinovev:
Fix for DRILL-7103:
Incorrect conversion to integer in BsonRecordReader#writeTimeStamp

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
@@ -75,17 +75,19 @@ public class TestBsonRecordReader {
     writer.reset();
     bsonReader.write(writer, new BsonDocumentReader(bsonDoc));
     SingleMapReaderImpl mapReader = (SingleMapReaderImpl) writer.getMapVector().getReader();
-    assertEquals(10l, mapReader.reader("seqNo").readLong().longValue());
+    assertEquals(10L, mapReader.reader("seqNo").readLong().longValue());
   }
 
   @Test
   public void testTimeStampType() throws IOException {
     BsonDocument bsonDoc = new BsonDocument();
-    bsonDoc.append("ts", new BsonTimestamp(1000, 10));
+    bsonDoc.append("ts_small", new BsonTimestamp(1000, 10));
+    bsonDoc.append("ts_large", new BsonTimestamp(1000000000, 10));
     writer.reset();
     bsonReader.write(writer, new BsonDocumentReader(bsonDoc));
     SingleMapReaderImpl mapReader = (SingleMapReaderImpl) writer.getMapVector().getReader();
-    assertEquals(1000000l, mapReader.reader("ts").readLocalDateTime().atZone(ZoneOffset.systemDefault()).toInstant().toEpochMilli());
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -88,6 +88,17 @@ public void testTimeStampType() throws IOException {
     assertEquals(1000000l, mapReader.reader("ts").readLocalDateTime().atZone(ZoneOffset.systemDefault()).toInstant().toEpochMilli());
   }
 
+  @Test
+  public void testTimeStampType_now() throws IOException {
```
vvysotskyi: Looks like this is almost the same test as the test before. Can they both be combined, for example, we can write two records (or more) into `BsonDocument`. Also, please use a literal instead of `System.currentTimeMillis()`.

oleg-zinovev: fixed

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -81,11 +81,13 @@ public void testIntType() throws IOException {
...
   public void testTimeStampType() throws IOException {
     BsonDocument bsonDoc = new BsonDocument();
-    bsonDoc.append("ts", new BsonTimestamp(1000, 10));
+    bsonDoc.append("ts_small", new BsonTimestamp(1000, 10));
+    bsonDoc.append("ts_large", new BsonTimestamp(1000000000, 10));
     writer.reset();
     bsonReader.write(writer, new BsonDocumentReader(bsonDoc));
     SingleMapReaderImpl mapReader = (SingleMapReaderImpl) writer.getMapVector().getReader();
-    assertEquals(1000000l, mapReader.reader("ts").readLocalDateTime().atZone(ZoneOffset.systemDefault()).toInstant().toEpochMilli());
+    assertEquals(1000000l, mapReader.reader("ts_small").readLocalDateTime().atZone(ZoneOffset.systemDefault()).toInstant().toEpochMilli());
```
vvysotskyi: Please change`l` to upper case.
```suggestion
    assertEquals(1000000L, mapReader.reader("ts_small").readLocalDateTime().atZone(ZoneOffset.systemDefault()).toInstant().toEpochMilli());
```

exec/java-exec/src/test/java/org/apache/drill/exec/store/bson/TestBsonRecordReader.java
```diff
@@ -81,11 +81,13 @@ public void testIntType() throws IOException {
...
+    assertEquals(1000000000000l, mapReader.reader("ts_large").readLocalDateTime().atZone(ZoneOffset.systemDefault()).toInstant().toEpochMilli());
```
vvysotskyi:
```suggestion
    assertEquals(1000000000000L, mapReader.reader("ts_large").readLocalDateTime().atZone(ZoneOffset.systemDefault()).toInstant().toEpochMilli());
```

PR #1696, opened 2019-03-14
DRILL-7095: Expose table schema (TupleMetadata) to physical operator (EasySubScan)

arina-ielchiieva:
1. Add system / session option `store.table.use_schema_file` to control if file schema can be used during query execution. False by default.
2. Added methods in StoragePlugin interface which allow to create Group Scan with provided table schema.
3. EasyGroupScan and EasySubScan now contain table schema, also they are able to serialize / deserialize it along with other scan properties.
4. DrillTable which is the main entry point for schema provisioning, has method to store schema and later uses it to create physical scan.
5. WorkspaceSchema when returning Drill table instance will get table schema from table root if available and if `store.table.use_schema_file` is set to true.

This PR is the next step for Schema Provisioning project which currently exposes schema only for text reader.

Jira: [DRILL-7095](https://issues.apache.org/jira/browse/DRILL-7095).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
@@ -63,6 +63,7 @@ import org.apache.drill.exec.planner.logical.DrillViewTable;
 import org.apache.drill.exec.planner.logical.DynamicDrillTable;
 import org.apache.drill.exec.planner.logical.FileSystemCreateTableEntry;
 import org.apache.drill.exec.planner.sql.ExpandingConcurrentMap;
+import org.apache.drill.exec.record.metadata.schema.FsMetastoreSchemaProvider;
 import org.apache.drill.exec.store.AbstractSchema;
 import org.apache.drill.exec.store.PartitionNotFoundException;
 import org.apache.drill.exec.store.SchemaConfig;
@@ -109,7 +110,7 @@ public class WorkspaceSchemaFactory {
       WorkspaceConfig config,
       List<FormatMatcher> formatMatchers,
       LogicalPlanPersistence logicalPlanPersistence,
-      ScanResult scanResult) throws ExecutionSetupException, IOException {
+      ScanResult scanResult) throws ExecutionSetupException {
     this.logicalPlanPersistence = logicalPlanPersistence;
     this.fsConf = plugin.getFsConf();
     this.plugin = plugin;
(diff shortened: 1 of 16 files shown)


arina-ielchiieva:
@paul-rogers please review.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/easy/EasySubScan.java
```diff
@@ -47,18 +48,20 @@
...
+    @JacksonInject StoragePluginRegistry engineRegistry,
+    @JsonProperty("columns") List<SchemaPath> columns,
+    @JsonProperty("selectionRoot") Path selectionRoot,
+    @JsonProperty("partitionDepth") int partitionDepth,
+    @JsonProperty("schema") TupleSchema schema
+    ) throws ExecutionSetupException {
     super(userName);
     this.formatPlugin = (EasyFormatPlugin<?>) engineRegistry.getFormatPlugin(storageConfig, formatConfig);
     Preconditions.checkNotNull(this.formatPlugin);
+    this.formatPlugin.setSchema(schema);
```
paul-rogers: Again, can't store this in the plugin: the plugin is (or should be) shared.

Or, are we serializing the plugin (config) to handle table properties? If so, then the schema can be set via the Drill web console, which I don't think we want as it will encourage people to create plugin configs for table schema, but that will require that each file have a distinct file suffix, which will cause confusion. (Already does for CSV/CSVH).

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/TextFormatPlugin.java
```diff
@@ -219,6 +220,8 @@ protected ColumnsScanFramework buildFramework(
     }
   }
 
+  private TupleMetadata schema;
```
paul-rogers: See earlier notes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/TextFormatPlugin.java
```diff
@@ -343,4 +344,14 @@ protected ScanStats getScanStats(final PlannerSettings settings, final EasyGroup
     final double estRowCount = data / estimatedRowSize;
     return new ScanStats(GroupScanProperty.NO_EXACT_ROW_COUNT, (long) estRowCount, 1, data);
   }
+
+  @Override
+  public void setSchema(TupleMetadata schema) {
+    this.schema = schema;
+  }
```
paul-rogers: Would be best if the schema can be immutable: set in the constructor and not changeable once set. Else, it is very unclear what the semantics will be.

Same is true, by the way, for the scan nodes.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -683,5 +683,6 @@ drill.exec.options: {
     exec.query.return_result_set_for_ddl: true,
     # ========= rm related options ===========
     exec.rm.queryTags: "",
-    exec.rm.queues.wait_for_preferred_nodes: true
+    exec.rm.queues.wait_for_preferred_nodes: true,
+    store.table.use_schema_file: false
```
paul-rogers: Move this closer to tother store-related options?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/WorkspaceSchemaFactory.java
```diff
@@ -572,9 +573,21 @@ public Table getTable(String tableName) {
       }
       final DrillTable table = tables.get(tableKey);
       setMetadataTable(table, tableName);
+      setSchema(table, tableName);
       return table;
     }
 
+    private void setSchema(DrillTable table, String tableName) {
```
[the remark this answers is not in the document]
arina-ielchiieva: Might seem weird but it does not, only contains selection of files.

https://github.com/apache/drill/blob/469be17597e7b7c6bc1de9863dcb6c5604a55f0c/exec/java-exec/src/main/java/org/apache/drill/exec/planner/logical/DrillTable.java

arina-ielchiieva:
@paul-rogers you are right, I should not have modified plugin. I have updated the code and addressed other code review comments, please review.

PR #1697, opened 2019-03-15
DRILL-7105: Error while building the Drill native client

agozhiy:
Added a compiler option in CMakeLists.txt to support the ISO C++ 2011 standard.
Also, changed the CMake min version to 3.1.3 to match the min version specified in protobuf.

### DIFF
--- a/contrib/native/client/CMakeLists.txt
+++ b/contrib/native/client/CMakeLists.txt
@@ -15,7 +15,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-cmake_minimum_required(VERSION 2.6)
+cmake_minimum_required(VERSION 3.1.3)
+
+set (CMAKE_CXX_STANDARD 11)
+set (CMAKE_CXX_STANDARD_REQUIRED ON)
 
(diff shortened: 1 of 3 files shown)


contrib/native/client/CMakeLists.txt
```diff
@@ -22,6 +22,14 @@ cmake_policy(SET CMP0043 NEW)
 cmake_policy(SET CMP0048 NEW)
 enable_testing()
 
+if (CMAKE_VERSION VERSION_LESS "3.1")
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
vvysotskyi: What about `MSVC` and others?

agozhiy: Changed the approach to rising the min version as it required by protobuf.

kkhatua:
EMail Reference:
https://lists.apache.org/thread.html/5563f9ce8085720a37f28b0e266ac3c392ec40e5c91c7572bd726fdf@%3Cdev.drill.apache.org%3E

PR #1702, opened 2019-03-19
DRILL-7107 Unable to connect to Drill 1.15 through ZK

bitblender:
With these changes, the ACLProvider is passed to ZKClusterCoordinator to avoid issues that arise when a client instantiates a ZKClusterCoordinator in a secure setup.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKClusterCoordinator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKClusterCoordinator.java
@@ -18,7 +18,6 @@
 package org.apache.drill.exec.coord.zk;
 
 import static org.apache.drill.shaded.guava.com.google.common.collect.Collections2.transform;
-
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKClusterCoordinator.java
```diff
@@ -81,23 +78,20 @@
...
-    this(config, connect, null);
+  public ZKClusterCoordinator(DrillConfig config, String connect) {
+    this(config, connect, new DefaultACLProvider());
   }
 
-  public ZKClusterCoordinator(DrillConfig config, BootStrapContext context)
-          throws IOException, DrillbitStartupException {
-    this(config, null, context);
+  public ZKClusterCoordinator(DrillConfig config, ACLProvider aclProvider) {
+    this(config, null, aclProvider);
```
sohami: Please add test for this codepath using `ClusterFixture`.

bitblender: I tried writing a test where the Drillbits (inside ClusterFixture) are setup with ZK_APPLY_SECURE_ACL=false (to avoid the need to setup a secure ZK server within the unit test) and the ClientFixture is setup with ZK_APPLY_SECURE_ACL=true (to simulate the failure). Starting a test with different values for the same property turns out to be quite hard because the ClusterFixture internally instantiates a ClientFixure. Changing this behavior might affect other tests.

sohami: Ok. Please open a JIRA to track it for future if needed.

PR #1703, opened 2019-03-19
DRILL-7110: Skip writing profile when an ALTER SESSION is executed

kkhatua:
Allows (by default) for `ALTER SESSION SET <option>=<value>` queries to NOT be writen to the profile store. This would avoid the risk of potentially adding up to a lot of profiles being written unnecessarily, since those changes are also reflected on the queries that follow.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -842,7 +842,10 @@ public final class ExecConstants {
    * for any query.
    */
   public static final String ENABLE_QUERY_PROFILE_OPTION = "exec.query_profile.save";
-  public static final BooleanValidator ENABLE_QUERY_PROFILE_VALIDATOR = new BooleanValidator(ENABLE_QUERY_PROFILE_OPTION, null);
+  public static final BooleanValidator ENABLE_QUERY_PROFILE_VALIDATOR = new BooleanValidator(ENABLE_QUERY_PROFILE_OPTION, new OptionDescription("Save completed profiles to the persistent store"));
+  //Allow to skip writing Alter Session profiles
+  public static final String SKIP_ALTER_SESSION_QUERY_PROFILE = "exec.query_profile.alter_session.skip";
+  public static final BooleanValidator SKIP_SESSION_QUERY_PROFILE_VALIDATOR = new BooleanValidator(SKIP_ALTER_SESSION_QUERY_PROFILE, new OptionDescription("Skip saving ALTER SESSION profiles"));
 
   /**
    * Profiles are normally written after the last client message to reduce latency.
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
@@ -77,6 +77,7 @@ public class QueryContext implements AutoCloseable, OptimizerRulesContext, Schem
   private final QueryContextInformation queryContextInfo;
   private final ViewExpansionContext viewExpansionContext;
   private final SchemaTreeProvider schemaTreeProvider;
+  private boolean skipProfileWrite;
   /** Stores constants and their holders by type */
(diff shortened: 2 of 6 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -92,6 +93,7 @@ public QueryContext(final UserSession session, final DrillbitContext drillbitCon
     this.drillbitContext = drillbitContext;
     this.session = session;
     this.queryId = queryId;
+    skipWritingProfile(false);
```
arina-ielchiieva:
```suggestion
    this.skipProfileWrite = false;
```

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -838,7 +838,10 @@ private ExecConstants() {
    * for any query.
    */
   public static final String ENABLE_QUERY_PROFILE_OPTION = "exec.query_profile.save";
-  public static final BooleanValidator ENABLE_QUERY_PROFILE_VALIDATOR = new BooleanValidator(ENABLE_QUERY_PROFILE_OPTION, null);
+  public static final BooleanValidator ENABLE_QUERY_PROFILE_VALIDATOR = new BooleanValidator(ENABLE_QUERY_PROFILE_OPTION, new OptionDescription("Save completed profiles to the persistent store"));
+  //Allow to skip writing Alter Session profiles
+  public static final String SKIP_ALTER_SESSION_QUERY_PROFILE = "exec.query_profile.alter_session.skip";
+  public static final BooleanValidator SKIP_SESSION_QUERY_PROFILE_VALIDATOR = new BooleanValidator(SKIP_ALTER_SESSION_QUERY_PROFILE, new OptionDescription("Skip saving ALTER SESSION profiles"));
```
arina-ielchiieva: How we handle reset command? Will it be logged?

kkhatua: No. Any `ALTER SESSION ...` query will not be logged as the option[s] is set in SESSION scope. With `RESET` , any subsequent query will log with no Session options showing up in the profile. The proof of the altered session options will always show up in queries' profiles, so logging any `ALTER SESSION ...` query is redundant.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/SetOptionHandler.java
```diff
@@ -113,6 +113,12 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ValidationException, Foreman
       }
     }
 
+    // Skip writing profiles for "ALTER SESSION SET" queries
+    if (optionScope == OptionScope.SESSION && options.getBoolean(ExecConstants.SKIP_ALTER_SESSION_QUERY_PROFILE)) {
+      logger.debug("Will not write profile for ALTER SESSION SET ... ");
+      context.skipWritingProfile(true);
```
sohami: this logic can be moved to line 73. No need to again check for `optionScope` here.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
sohami: no need to explicitly set skipProfileWrite to false in constructor.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -785,8 +785,8 @@ public void close() throws Exception {
 
       // Debug option: write query profile before sending final results so that
       // the client can be certain the profile exists.
-
-      if (profileOption == ProfileOption.SYNC) {
+      boolean skipProfileWrite = queryContext.isSkipProfileWrite();
```
sohami: use `final`

kkhatua:
@sohami / @arina-ielchiieva I've made the minor changes. Please review and let me know if anything else is missing.

sohami:
LGTM

arina-ielchiieva:

PR #1704, opened 2019-03-19
DRILL-7113: Fix creation of filter conditions for IS NULL and IS NOT …

amansinha100:
…NULL for MapR-DB format plugin


amansinha100:
@HanumathRao  could you pls review ?

HanumathRao:
@amansinha100  Thanks for making the changes. Changes look good to me. +1

PR #1707, opened 2019-03-20
DRILL-7125: REFRESH TABLE METADATA fails after upgrade from Drill 1.1…

sohami:
…3.0 to Drill 1.15.0

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
@@ -24,23 +24,21 @@ import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.module.SimpleModule;
 import com.fasterxml.jackson.module.afterburner.AfterburnerModule;
-import org.apache.drill.exec.serialization.PathSerDe;
-import java.util.Set;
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -675,32 +699,28 @@ private boolean tableModified(List<Path> directories, Path metaFilePath, Path pa
...
     }
+    return logAndStopTimer(isModified, timer, numDirs);
+  }
+
+  private boolean logAndStopTimer(boolean isModified, Stopwatch timer, int numDirectories) {
     if (timer != null) {
-      logger.debug("No directories were modified. Took {} ms to check modification time of {} directories",
-          timer.elapsed(TimeUnit.MILLISECONDS), numDirs);
+      logger.debug("Directory {} was modified. Took {} ms to check modification time of {} directories",
+        isModified, timer.elapsed(TimeUnit.MILLISECONDS), numDirectories);
```
amansinha100: It is actually useful to have a log message even for the case when no directories were modified..that way we can account for the elapsed time.  You can preserve the prior log message conditional based on the `isModified` flag.

sohami: Thanks for catching this. Somehow I missed logging was done in both cases. Have made the change.

arina-ielchiieva:
+1

sohami:
Thanks for the review. Squashed and rebased.

PR #1708, opened 2019-03-21
DRILL-7118: Filter not getting pushed down on MapR-DB tables.

HanumathRao:
The code changes here reverts back to the old behavior of pushing the predicate (is null) to the maprdb scan. The code is inadvertently removed during other changes that went in. 
The JIRA which changed the code in this file is DRILL-6969: Fix inconsistency of reading MaprDB JSON tables using hive

@amansinha100  please review this PR.

### DIFF
--- a/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
+++ b/contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
@@ -107,13 +107,13 @@ class CompareFunctionsProcessor extends AbstractExprVisitor<Boolean, LogicalExpr
     LogicalExpression nameArg = call.args.get(0);
     LogicalExpression valueArg = call.args.size() >= 2 ? call.args.get(1) : null;
 
-    if (valueArg != null) {
-      if (VALUE_EXPRESSION_CLASSES.contains(nameArg.getClass())) {
-        LogicalExpression swapArg = valueArg;
-        valueArg = nameArg;
-        nameArg = swapArg;
-        evaluator.functionName = COMPARE_FUNCTIONS_TRANSPOSE_MAP.get(functionName);
-      }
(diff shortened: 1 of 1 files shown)


amansinha100:
@HanumathRao for future reference, can you mention in the description which prior JIRA had made the relevant change that caused change in behavior.

HanumathRao:
> @HanumathRao for future reference, can you mention in the description which prior JIRA had made the relevant change that caused change in behavior.

I have updated the description.

contrib/format-maprdb/src/main/java/org/apache/drill/exec/store/mapr/db/json/CompareFunctionsProcessor.java
```diff
@@ -107,15 +107,13 @@ private static CompareFunctionsProcessor processWithEvaluator(FunctionCall call,
...
-        evaluator.functionName = COMPARE_FUNCTIONS_TRANSPOSE_MAP.get(functionName);
-      }
-      evaluator.success = nameArg.accept(evaluator, valueArg);
+    if (VALUE_EXPRESSION_CLASSES.contains(nameArg.getClass())) {
+      LogicalExpression swapArg = valueArg;
+      valueArg = nameArg;
+      nameArg = swapArg;
+      evaluator.functionName = COMPARE_FUNCTIONS_TRANSPOSE_MAP.get(functionName);
     }
+    evaluator.success = nameArg.accept(evaluator, valueArg);
```
vvysotskyi: This check was added to avoid NPE, but for the case when `VALUE_EXPRESSION_CLASSES.contains(nameArg.getClass())` is false, and `valueArg` is null, NPE will not occur. But should we add the next check?
```suggestion
    if (nameArg != null) {
      evaluator.success = nameArg.accept(evaluator, valueArg);
    }
```

HanumathRao: @vvysotskyi  I have made the changes.

PR #1711, opened 2019-03-21
DRILL-7011: Support schema in scan framework

paul-rogers:
Adds schema support to the row set-based scan framework and to the "V3" text reader based on that framework.

Adding the schema made clear that passing options as a long list of constructor arguments was not sustainable. Refactored code to use a builder pattern instead.

Added support for default values in the "null column loader", which required adding a "setValue" method to the column accessors.

Added unit tests for all new or changed functionality. See TestCsvWithSchema for the overall test of the entire integrated mechanism.


paul-rogers:
@arina-ielchiieva, here is the full integrated schema feature for review. For some reason, the unit TestCsvWithSchema test runs in Eclipse, but not from the command line. I'll investigate that tomorrow. Because of that, I've not done a full unit test run. Still, there is plenty of code to review while I figure out the unit test issue.

paul-rogers:
Turns out that we need to use the Joda date/time classes because the Joda Period class matches the Drill INTERVAL semantics, but Java 8 date/time has no similar concept. The latest commit uses Joda throughout the row set classes.

This version adds tests additional unit tests for CSV format with schema. These tests illustrate problems with a lack of schema, and show how a schema resolves the issues.

Resolved the test issue noted earlier. Full tests still need to be run before this PR can be considered ready to commit.

One open issue is the use of a wildcard query with a schema: should the wildcard expand to just the columns in each file (as in prior versions), to the set of columns in the schema, or a combination of the two? This requires a bit more thought to resolve, especially since the schema, in this PR, is supposed to be partial.

exec/java-exec/src/test/java/org/apache/drill/exec/store/easy/text/compliant/TestCsvWithSchema.java
```diff
@@ -82,6 +167,468 @@ public void testSchema() throws Exception {
...
+          }
+          result.clear();
+        }
+      }
+
+      // Outcome is random (which is the key problem). Don't assert on these
+      // because doing so can lead to a flakey test.
+
+      if (!sawFile1First || ! sawFile2First || !sawFullSchema || !sawPartialSchema || !sawIdAsCol0 || !sawIdAsCol1) {
+        System.out.println("Some variations did not occur");
```
arina-ielchiieva: I know PR is not complete yet, just a note to remove `System.out.println` during final clean up.

exec/java-exec/src/test/java/org/apache/drill/exec/store/easy/text/compliant/TestCsvWithSchema.java
```diff
@@ -82,6 +167,468 @@ public void testSchema() throws Exception {
...
+   * 0: "1", "bob  ", "      "
+   * 1: "2", "    ", "    "
+   * 2: "3", " 
+    * ", ""
+   * 3: "4", "  
+   *    " java.lang.NegativeArraySizeException: null
+   *      at io.netty.buffer.DrillBuf.unsafeGetMemory(DrillBuf.java:852) ~[classes/:4.0.48.Final]
+   * </pre></code>
+   */
+  @Test
```
arina-ielchiieva: Code is not highlighted as code in GitHub, I remember there were similar issues before, if I am not mistaken some hidden symbols were copied from text editor. Could you please check?

arina-ielchiieva:
@paul-rogers 
Actually when I was presenting the schema provisioning design, there were a proposal to add schema property `drill.is_full_schema`. By default it’s `false`, thus we assume that schema is partial.
If user wants to indicate that schema is strict and to ignore all columns except of those indicated in schema, he needs to create schema the following way:
	`create schema (col int) properties ('drill.is_full_schema' = 'true')`

Since most of the `default` property problems are related to star queries, we can state the following:
1. For queries with defined list of columns (aka projection queries: `select id, name from t`), we apply schema consistently.
2. For star queries and when schema property `drill.is_full_schema` is set to `false`, we might get inconsistent results with default values but it's ok since we discover schema on the read.
3. For star queries and when schema property `drill.is_full_schema` is set to `true`, we project only those columns indicated in schema.
What do you think?

exec/java-exec/src/test/java/org/apache/drill/exec/store/easy/text/compliant/TestCsvWithSchema.java
paul-rogers: Actually, this is a poor person's "test". Can't actually fail when some variations don't occur as test will be flaky. Since this is more of a demo of existing problems, it is not really a test. We'll remove this code when we switch to V3 as the default.

exec/java-exec/src/test/java/org/apache/drill/exec/store/easy/text/compliant/TestCsvWithSchema.java
paul-rogers: Looks OK to me. I suspect the problem might be with the volume of changes; that sometimes throws off browsers. Try viewing this again without expanding any other files that are not expanded by default.

paul-rogers:
Added wildcard projection handling for both "lenient" and "strict" schemas. Added tests for the scan framework. Added a basic tests for "lenient" wildcard projection with CSV files. Have not yet added CSV tests for strict schema because I need to know how to pass a table-level property with the new CREATE SCHEMA command.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/file/MetadataColumn.java
```diff
@@ -25,6 +25,8 @@
 
 public abstract class MetadataColumn extends ResolvedColumn implements ConstantColumnSpec {
 
+
+  public static final int ID = 15;
```
arina-ielchiieva: Do we still need ID?

arina-ielchiieva:
@paul-rogers command syntax is the following:
```
CREATE [OR REPLACE] SCHEMA
[LOAD 'file:///path/to/file']
[(column_name data_type nullability,...)]
[FOR TABLE `table_name`]
[PATH 'file:///schema_file_path/schema_file_name'] 
[PROPERTIES ('key1'='value1', 'key2'='value2', ...)]
```
`PROPERTIES` should be provided in parenthesis, in a form of key / value pairs where value follows after the key and equal sign, each enclosed the single quotes. Key / value groups should be separated by commas.
```
create schema
(col1 int, col2 int)
for table t
properties (
'drill.strict' = 'true',
'some_other_prop' = 'val')
```
In `.drill.schema` JSON file this would look the following way:
```
{
  "table" : "dfs.tmp.`t`",
  "schema" : {
    "columns" : [
      {
        "name" : "col1",
        "type" : "INT",
        "mode" : "OPTIONAL"
      },
      {
        "name" : "col2",
        "type" : "INT",
        "mode" : "OPTIONAL"
      }
    ],
    "properties" : {
      "drill.strict" : "true",
      "some_other_prop" : "val"
    }
  },
  "version" : 1
}
```
During deserialization they will be stored in `TupleMetadata` class (use `property(key)`, `properties` methods to extract them).

If you want to add column properties, similar syntax will be used, except instead of parenthesis you need to use curly braces:
```
create schema
(col1 int, col2 int properties {'drill.strict' = 'true'})
for table t
properties (
'drill.strict' = 'true',
'some_other_prop' = 'val')
```
JSON output:
```
{
  "table" : "dfs.tmp.`t`",
  "schema" : {
    "columns" : [
      {
        "name" : "col1",
        "type" : "INT",
        "mode" : "OPTIONAL",
        "properties" : {
          "drill.strict" : "true"
        }
      },
      {
        "name" : "col2",
        "type" : "INT",
        "mode" : "OPTIONAL"
      }
    ],
    "properties" : {
      "drill.is_strict_schema" : "true",
      "some_other_prop" : "val"
    }
  },
  "version" : 1
}
```
Please let me know if there are any other syntax related questions.

paul-rogers:
Addressed review comments. Added more implicit type conversions.

arina-ielchiieva:
It's an impressive PR and it's good that we did code review interactively. Overall, looks good. Please squash the commits and it will be good to go.

paul-rogers:
Final cleanup to ensure all check style warnings are addressed and unit tests pass.
As part of this, removed a bunch of unused "suppress warnings" annotations that are no longer needed in Java 8.

This PR should be ready to go. Since this has gotten rather large (sorry!) we'll tackle enforcing "strict" column schema in the next one.

arina-ielchiieva:
+1, let's finish pending changes in different PR.

PR #1712, opened 2019-03-21
DRILL-7079: Drill can't query views from the S3 storage when plain authentication is enabled

KazydubB:
Please see [DRILL-7079](https://issues.apache.org/jira/browse/DRILL-7079) for more info.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillFile.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillFile.java
@@ -19,6 +19,7 @@ package org.apache.drill.exec.dotdrill;
 
 import org.apache.drill.common.config.LogicalPlanPersistence;
 import org.apache.drill.exec.store.dfs.DrillFileSystem;
+import org.apache.drill.exec.util.ImpersonationUtil;
 import org.apache.hadoop.fs.FileStatus;
 
 import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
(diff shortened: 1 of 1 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/dotdrill/DotDrillFile.java
```diff
@@ -55,6 +56,13 @@ public DotDrillType getType(){
    * @return Return owner of the file in underlying file system.
    */
   public String getOwner() {
+    if (type == DotDrillType.VIEW && status.getOwner().isEmpty()) {
```
vvysotskyi: What about regular tables, parquet metadata cache files or stat files? Is it possible that empty `status.getOwner()` may cause problems for them?

KazydubB: No, the other 'files' work fine.

ihuzenko: If owner is not always empty for .view.drill files on S3, maybe it makes sense to inverse the check a little bit, like: 

```java
  public String getOwner() {
    String owner = status.getOwner();
    if (owner.isEmpty() && type == DotDrillType.VIEW) {
      // Drill view S3AFileStatus is not populated with owner (it has default value of "").
      // This empty String causes IllegalArgumentException to be thrown (if impersonation is enabled) in
      // SchemaTreeProvider#createRootSchema(String, SchemaConfigInfoProvider). To work-around the issue
      // we can return current user as if they were the owner of the file (since they have access to it).
      owner = ImpersonationUtil.getProcessUserName();
    }
    return owner;
  }
```
Also what if owner is empty but file type is not ```DotDrillType.VIEW``` ? Should we throw exception in such case to detect potential future problems early in such case ?

KazydubB: Owner for views on S3 is always empty.
The `getOwner()` method is currently called for views only and for other files the method which throws the `IllegalArgumentException` won't be called with empty `String`. So I suppose we should not throw an `Exception` from the method.

PR #1714, opened 2019-03-23
DRILL-7048: Implement JDBC Statement.setMaxRows() with System Option

kkhatua:
This is a rework of #1608 with complete JDBC support. All the issues in #1608 have been addressed here as this PR is a superset of #1608 and #1593 
This introduces support for JDBC's Statement.setMaxRows(int) API, which can help Drill execute a query much faster if it knows that not ALL the records in the resultset will be consumed upfront.
This PR is broken into 4 commits
1. Introduces the core changes to support the feature within Drill's execution engine.
2. Has corresponding Protobuf changes
3. Support for REST API
4. Support for JDBC API

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
@@ -41,6 +41,7 @@ import org.apache.drill.exec.rpc.user.UserSession;
 import org.apache.drill.exec.server.DrillbitContext;
 import org.apache.drill.exec.server.QueryProfileStoreContext;
 import org.apache.drill.exec.server.options.OptionValue;
+import org.apache.drill.exec.server.options.OptionValue.OptionScope;
 import org.apache.drill.exec.server.options.QueryOptionManager;
 import org.apache.drill.exec.store.PartitionExplorer;
 import org.apache.drill.exec.store.PartitionExplorerImpl;
@@ -60,7 +61,6 @@ import io.netty.buffer.DrillBuf;
 // TODO - consider re-name to PlanningContext, as the query execution context actually appears
 // in fragment contexts
 public class QueryContext implements AutoCloseable, OptimizerRulesContext, SchemaConfigInfoProvider {
-
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(QueryContext.class);
   public enum SqlStatementType {OTHER, ANALYZE, CTAS, EXPLAIN, DESCRIBE_TABLE, DESCRIBE_SCHEMA, REFRESH, SELECT, SETOPTION};
 
@@ -105,6 +105,23 @@ public class QueryContext implements AutoCloseable, OptimizerRulesContext, Schem
       this.table = new DrillOperatorTable(drillbitContext.getFunctionImplementationRegistry(), drillbitContext.getOptionManager());
     } else {
       this.table = drillbitContext.getOperatorTable();
+    }
+
+    // Checking for limit on ResultSet rowcount and if user attempting to override the system value
+    int sessionMaxRowCount = queryOptions.getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue();
+    int defaultMaxRowCount = queryOptions.getOptionManager(OptionScope.SYSTEM).getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue();
+    int autoLimitRowCount = 0;
+    if (sessionMaxRowCount > 0 && defaultMaxRowCount > 0) {
+      autoLimitRowCount = Math.min(sessionMaxRowCount, defaultMaxRowCount);
+    } else {
+      autoLimitRowCount = Math.max(sessionMaxRowCount, defaultMaxRowCount);
+    }
+    if (autoLimitRowCount == defaultMaxRowCount && defaultMaxRowCount != sessionMaxRowCount) {
+      // Required to indicate via OptionScope=QueryLevel that session limit is overridden by system limit
+      queryOptions.setLocalOption(ExecConstants.QUERY_MAX_ROWS, autoLimitRowCount);
+    }
+    if (autoLimitRowCount > 0) {
+      logger.debug("ResultSet size is auto-limited to {} rows [Session: {} / Default: {}]", autoLimitRowCount, sessionMaxRowCount, defaultMaxRowCount);
     }
 
     queryContextInfo = Utilities.createQueryContextInfo(session.getDefaultSchemaPath(), session.getSessionId());
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
@@ -281,10 +281,10 @@ public class SystemOptionManager extends BaseOptionManager implements AutoClosea
       new OptionDefinition(ExecConstants.DETERMINISTIC_SAMPLING_VALIDATOR),
       new OptionDefinition(ExecConstants.NDV_BLOOM_FILTER_ELEMENTS_VALIDATOR),
       new OptionDefinition(ExecConstants.NDV_BLOOM_FILTER_FPOS_PROB_VALIDATOR),
-      new OptionDefinition(ExecConstants.RM_QUERY_TAGS_VALIDATOR,
-        new OptionMetaData(OptionValue.AccessibleScopes.SESSION_AND_QUERY, false, false)),
+      new OptionDefinition(ExecConstants.RM_QUERY_TAGS_VALIDATOR, new OptionMetaData(OptionValue.AccessibleScopes.SESSION_AND_QUERY, false, false)),
       new OptionDefinition(ExecConstants.RM_QUEUES_WAIT_FOR_PREFERRED_NODES_VALIDATOR),
-      new OptionDefinition(ExecConstants.TDIGEST_COMPRESSION_VALIDATOR)
+      new OptionDefinition(ExecConstants.TDIGEST_COMPRESSION_VALIDATOR),
+      new OptionDefinition(ExecConstants.QUERY_MAX_ROWS_VALIDATOR, new OptionMetaData(OptionValue.AccessibleScopes.ALL, true, false))
     };
 
     CaseInsensitiveMap<OptionDefinition> map = Arrays.stream(definitions)
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
@@ -143,6 +143,11 @@ public class Foreman implements Runnable {
     this.closeFuture = initiatingClient.getChannelClosureFuture();
     closeFuture.addListener(closeListener);
 
+    // Apply AutoLimit on resultSet (Usually received via REST APIs)
+    final int autoLimit = queryRequest.getAutolimitRowcount();
+    if (autoLimit > 0) {
+      connection.getSession().getOptions().setLocalOption(ExecConstants.QUERY_MAX_ROWS, autoLimit);
+    }
     this.queryContext = new QueryContext(connection.getSession(), drillbitContext, queryId);
     this.queryManager = new QueryManager(queryId, queryRequest, drillbitContext.getStoreProvider(),
         drillbitContext.getClusterCoordinator(), this);
@@ -239,7 +244,7 @@ public class Foreman implements Runnable {
     try {
       /*
        Check if the foreman is ONLINE. If not don't accept any new queries.
-      */
+       */
       if (!drillbitContext.isForemanOnline()) {
         throw new ForemanException("Query submission failed since Foreman is shutting down.");
       }
@@ -504,7 +509,7 @@ public class Foreman implements Runnable {
     } catch (Exception e) {
       queryStateProcessor.moveToState(QueryState.FAILED, e);
     } finally {
-       /*
+      /*
        * Begin accepting external events.
        *
        * Doing this here in the finally clause will guarantee that it occurs. Otherwise, if there
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
@@ -377,6 +377,12 @@ public class QueryManager implements AutoCloseable {
       profileBuilder.setQuery(queryText);
     }
 
+    int autoLimitRowCount = foreman.getQueryContext().getOptions().getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue();
+    if (autoLimitRowCount > 0) {
+      profileBuilder.setAutoLimit(autoLimitRowCount);
+      logger.debug("The query's resultset was limited to {} rows", autoLimitRowCount);
+    }
+
     fragmentDataMap.forEach(new OuterIter(profileBuilder));
 
     return profileBuilder.build();
--- a/exec/jdbc/src/main/java/org/apache/drill/jdbc/DrillPreparedStatement.java
+++ b/exec/jdbc/src/main/java/org/apache/drill/jdbc/DrillPreparedStatement.java
@@ -31,5 +31,4 @@ import java.sql.PreparedStatement;
  * @see #unwrap
  */
 public interface DrillPreparedStatement extends PreparedStatement {
-
 }
--- a/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillCursor.java
+++ b/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillCursor.java
@@ -361,6 +361,9 @@ public class DrillCursor implements Cursor {
             ExecConstants.JDBC_BATCH_QUEUE_THROTTLING_THRESHOLD );
     resultsListener = new ResultsListener(this, batchQueueThrottlingThreshold);
     currentBatchHolder = new RecordBatchLoader(client.getAllocator());
+
+    // Set Query Timeout
+    logger.debug("Setting timeout as {}", this.statement.getQueryTimeout());
     setTimeout(this.statement.getQueryTimeout());
   }
 
--- a/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
+++ b/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
@@ -21,11 +21,13 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.SQLFeatureNotSupportedException;
 import java.sql.SQLType;
+import java.sql.Statement;
 
 import org.apache.calcite.avatica.AvaticaParameter;
 import org.apache.calcite.avatica.AvaticaPreparedStatement;
 import org.apache.calcite.avatica.Meta;
 import org.apache.calcite.avatica.Meta.StatementHandle;
+import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.proto.UserProtos.PreparedStatement;
 import org.apache.drill.jdbc.AlreadyClosedSqlException;
 import org.apache.drill.jdbc.DrillPreparedStatement;
@@ -258,5 +260,13 @@ abstract class DrillPreparedStatementImpl extends AvaticaPreparedStatement
   public void setObject(int parameterIndex, Object x, SQLType targetSqlType) throws SQLException {
     checkOpen();
     super.setObject(parameterIndex, x, targetSqlType);
+  }
+
+  @Override
+  public void setLargeMaxRows(long maxRowCount) throws SQLException {
+    super.setLargeMaxRows(maxRowCount);
+    try (Statement statement = this.connection.createStatement()) {
+      statement.execute("ALTER SESSION SET `" + ExecConstants.QUERY_MAX_ROWS + "`=" + maxRowCount);
+    }
   }
 }
--- a/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillStatementImpl.java
+++ b/exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillStatementImpl.java
@@ -25,6 +25,7 @@ import org.apache.calcite.avatica.AvaticaResultSet;
 import org.apache.calcite.avatica.AvaticaStatement;
 import org.apache.calcite.avatica.Meta.StatementHandle;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
+import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.jdbc.AlreadyClosedSqlException;
 import org.apache.drill.jdbc.DrillStatement;
 
@@ -269,5 +270,11 @@ public class DrillStatementImpl extends AvaticaStatement implements DrillStateme
   @Override
   public void setUpdateCount(int value) {
     updateCount = value;
+  }
+
+  @Override
+  public void setLargeMaxRows(long maxRowCount) throws SQLException {
+    super.setLargeMaxRows(maxRowCount);
+    execute("ALTER SESSION SET `" + ExecConstants.QUERY_MAX_ROWS + "`=" + maxRowCount);
   }
 }
--- a/exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
+++ b/exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
@@ -1,0 +1,261 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.jdbc;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+import java.util.Random;
+import java.util.concurrent.Semaphore;
+
+import org.apache.drill.exec.ExecConstants;
+import org.apache.drill.exec.planner.physical.PlannerSettings;
+import org.apache.drill.categories.JdbcTest;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import org.junit.experimental.categories.Category;
+
+/**
+ * Test for Drill's implementation of PreparedStatement's get/setMaxRows methods
+ */
+@Category(JdbcTest.class)
+public class PreparedStatementMaxRowsTest extends JdbcTestBase {
+
+  private static final Random RANDOMIZER = new Random(20150304);
+
+  private static final String SYS_OPTIONS_SQL = "SELECT * FROM sys.options";
+  private static final String SYS_OPTIONS_SQL_LIMIT_10 = "SELECT * FROM sys.options LIMIT 12";
+  private static final String ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X = "ALTER SYSTEM SET `" + ExecConstants.QUERY_MAX_ROWS + "`=";
+  // Lock used by all tests to avoid corrupting test scenario
+  private static final Semaphore maxRowsSysOptionLock = new Semaphore(1);
+
+  private static Connection connection;
+
+
+  @BeforeClass
+  public static void setUpConnection() throws SQLException {
+    Driver.load();
+    Properties properties = new Properties();
+    // Increased prepared statement creation timeout to avoid timeout failures
+    properties.setProperty(ExecConstants.bootDefaultFor(ExecConstants.CREATE_PREPARE_STATEMENT_TIMEOUT_MILLIS), "30000");
+    connection = DriverManager.getConnection("jdbc:drill:zk=local", properties);
+    try (Statement stmt = connection.createStatement()) {
+      stmt.execute(String.format("alter session set `%s` = true", PlannerSettings.ENABLE_DECIMAL_DATA_TYPE_KEY));
+    }
+  }
+
+  @AfterClass
+  public static void tearDownConnection() throws SQLException {
+    if (connection != null) {
+      connection.close();
+    }
+  }
+
+  @Before
+  public void getExclusiveLock() {
+    // Acquire lock
+    maxRowsSysOptionLock.acquireUninterruptibly();
+  }
+
+  @After
+  public void releaseExclusiveLock() {
+    // Release lock either way
+    maxRowsSysOptionLock.release();
+  }
+
+  ////////////////////////////////////////
+  // Query maxRows methods:
+
+  /**
+   * Test for reading of default max rows
+   */
+  @Test
+  public void testDefaultGetMaxRows() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      int maxRowsValue = pStmt.getMaxRows();
+      assertEquals(0, maxRowsValue);
+    }
+  }
+
+  /**
+   * Test Invalid parameter by giving negative maxRows value
+   */
+  @Test
+  public void testInvalidSetMaxRows() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      //Setting negative value
+      int valueToSet = -10;
+      int origMaxRows = pStmt.getMaxRows();
+      try {
+        pStmt.setMaxRows(valueToSet);
+      } catch (final SQLException e) {
+        assertThat(e.getMessage(), containsString("illegal maxRows value: " + valueToSet));
+      }
+      // Confirm no change
+      assertEquals(origMaxRows, pStmt.getMaxRows());
+    }
+  }
+
+  /**
+   * Test setting a valid maxRows value
+   */
+  @Test
+  public void testValidSetMaxRows() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      //Setting positive value
+      int valueToSet = RANDOMIZER.nextInt(59) + 1;
+      pStmt.setMaxRows(valueToSet);
+      assertEquals(valueToSet, pStmt.getMaxRows());
+    }
+  }
+
+  /**
+   * Test setting maxSize as zero (i.e. no Limit)
+   */
+  @Test
+  public void testSetMaxRowsAsZero() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      pStmt.setMaxRows(0);
+      pStmt.execute();
+      ResultSet rs = pStmt.getResultSet();
+      int rowCount = 0;
+      while (rs.next()) {
+        rs.getBytes(1);
+        rowCount++;
+      }
+      rs.close();
+      assertTrue(rowCount > 0);
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value lower than existing query limit
+   */
+  @Test
+  public void testSetMaxRowsLowerThanQueryLimit() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL_LIMIT_10)) {
+      int valueToSet = RANDOMIZER.nextInt(9) + 1; // range: [1-9]
+      pStmt.setMaxRows(valueToSet);
+      pStmt.executeQuery();
+      ResultSet rs = pStmt.getResultSet();
+      int rowCount = 0;
+      while (rs.next()) {
+        rs.getBytes(1);
+        rowCount++;
+      }
+      rs.close();
+      assertEquals(valueToSet, rowCount);
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value higher than existing query limit
+   */
+  @Test
+  public void testSetMaxRowsHigherThanQueryLimit() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL_LIMIT_10)) {
+      int valueToSet = RANDOMIZER.nextInt(10) + 11; // range: [11-20]
+      pStmt.setMaxRows(valueToSet);
+      pStmt.executeQuery();
+      ResultSet rs = pStmt.getResultSet();
+      int rowCount = 0;
+      while (rs.next()) {
+        rs.getBytes(1);
+        rowCount++;
+      }
+      rs.close();
+      assertTrue(valueToSet > rowCount);
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value lower than existing SYSTEM limit
+   */
+  @Test
+  public void testSetMaxRowsLowerThanSystemLimit() throws SQLException {
+    int sysValueToSet = RANDOMIZER.nextInt(5) + 6; // range: [6-10]
+    setSystemMaxRows(sysValueToSet);
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      int valueToSet = RANDOMIZER.nextInt(5) + 1; // range: [1-5]
+      pStmt.setMaxRows(valueToSet);
+      pStmt.executeQuery();
+      ResultSet rs = pStmt.getResultSet();
+      int rowCount = 0;
+      while (rs.next()) {
+        rs.getBytes(1);
+        rowCount++;
+      }
+      rs.close();
+      assertEquals(valueToSet, rowCount);
+    }
+    setSystemMaxRows(0); //RESET
+  }
+
+  /**
+   * Test setting maxSize at a value higher than existing SYSTEM limit
+   */
+  @Test
+  public void testSetMaxRowsHigherThanSystemLimit() throws SQLException {
+    int sysValueToSet = RANDOMIZER.nextInt(5) + 6; // range: [6-10]
+    setSystemMaxRows(sysValueToSet);
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      int valueToSet = RANDOMIZER.nextInt(5) + 11; // range: [11-15]
+      pStmt.setMaxRows(valueToSet);
+      pStmt.executeQuery();
+      ResultSet rs = pStmt.getResultSet();
+      int rowCount = 0;
+      while (rs.next()) {
+        rs.getBytes(1);
+        rowCount++;
+      }
+      rs.close();
+      assertEquals(sysValueToSet, rowCount);
+    }
+    setSystemMaxRows(0); //RESET
+  }
+
+
+  //////////
+  // Parameters-not-implemented tests:
+
+  // Sets the SystemMaxRows option
+  private void setSystemMaxRows(int sysValueToSet) throws SQLException {
+    // Setting the value
+    try (Statement stmt = connection.createStatement()) {
+      stmt.executeQuery(ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X + sysValueToSet);
+      ResultSet rs = stmt.getResultSet();
+      while (rs.next()) { /*Do Nothing*/ }
+      rs.close();
+    } catch (SQLException e) {
+      throw e;
+    }
+  }
+}
--- a/exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
+++ b/exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
@@ -72,6 +72,7 @@ import org.junit.experimental.categories.Category;
 public class PreparedStatementTest extends JdbcTestBase {
 
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PreparedStatementTest.class);
+  private static final Random RANDOMIZER = new Random(20150304);
 
   private static final String SYS_VERSION_SQL = "select * from sys.version";
   private static final String SYS_RANDOM_SQL =
@@ -272,8 +273,8 @@ public class PreparedStatementTest extends JdbcTestBase {
       int valueToSet = -10;
       try {
         stmt.setQueryTimeout(valueToSet);
-      } catch ( final SQLException e) {
-        assertThat( e.getMessage(), containsString( "illegal timeout value") );
+      } catch (final SQLException e) {
+        assertThat(e.getMessage(), containsString( "illegal timeout value") );
       }
     }
   }
@@ -285,7 +286,7 @@ public class PreparedStatementTest extends JdbcTestBase {
   public void testValidSetQueryTimeout() throws SQLException {
     try (PreparedStatement stmt = connection.prepareStatement(SYS_VERSION_SQL)) {
       //Setting positive value
-      int valueToSet = new Random(20150304).nextInt(59)+1;
+      int valueToSet = RANDOMIZER.nextInt(59) + 1;
       logger.info("Setting timeout as {} seconds", valueToSet);
       stmt.setQueryTimeout(valueToSet);
       assertEquals(valueToSet, stmt.getQueryTimeout());
@@ -461,5 +462,4 @@ public class PreparedStatementTest extends JdbcTestBase {
       }
     }
   }
-
 }
--- a/exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementMaxRowsTest.java
+++ b/exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementMaxRowsTest.java
@@ -1,0 +1,248 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.jdbc;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertEquals;
+
+import org.apache.drill.categories.JdbcTest;
+import org.apache.drill.exec.ExecConstants;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.Statement;
+import java.util.Random;
+import java.util.concurrent.Semaphore;
+import java.sql.SQLException;
+
+/**
+ * Test for Drill's implementation of Statement's get/setMaxRows methods
+ */
+@Category(JdbcTest.class)
+public class StatementMaxRowsTest extends JdbcTestBase {
+
+  private static final Random RANDOMIZER = new Random(20150304);
+
+  private static final String SYS_OPTIONS_SQL = "SELECT * FROM sys.options";
+  private static final String SYS_OPTIONS_SQL_LIMIT_10 = "SELECT * FROM sys.options LIMIT 12";
+  private static final String ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X = "ALTER SYSTEM SET `" + ExecConstants.QUERY_MAX_ROWS + "`=";
+  // Lock used by all tests to avoid corrupting test scenario
+  private static final Semaphore maxRowsSysOptionLock = new Semaphore(1);
+
+  private static Connection connection;
+
+  @BeforeClass
+  public static void setUpStatement() throws SQLException {
+    // (Note: Can't use JdbcTest's connect(...) because JdbcTest closes
+    // Connection--and other JDBC objects--on test method failure, but this test
+    // class uses some objects across methods.)
+    connection = new Driver().connect("jdbc:drill:zk=local", null);
+  }
+
+  @AfterClass
+  public static void tearDownStatement() throws SQLException {
+    if (connection != null) {
+      connection.close();
+    }
+  }
+
+  @Before
+  public void getExclusiveLock() {
+    // Acquire lock
+    maxRowsSysOptionLock.acquireUninterruptibly();
+  }
+
+  @After
+  public void releaseExclusiveLock() {
+    // Release lock either way
+    maxRowsSysOptionLock.release();
+  }
+
+  ////////////////////////////////////////
+  // Query maxRows methods:
+
+  /**
+   * Test for reading of default max rows
+   */
+  @Test
+  public void testDefaultGetMaxRows() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      int maxRowsValue = stmt.getMaxRows();
+      assertEquals(0, maxRowsValue);
+    }
+  }
+
+  /**
+   * Test Invalid parameter by giving negative maxRows value
+   */
+  @Test
+  public void testInvalidSetMaxRows() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      // Setting negative value
+      int valueToSet = -10;
+      int origMaxRows = stmt.getMaxRows();
+      try {
+        stmt.setMaxRows(valueToSet);
+      } catch (final SQLException e) {
+        assertThat(e.getMessage(), containsString("illegal maxRows value: " + valueToSet));
+      }
+      // Confirm no change
+      assertEquals(origMaxRows, stmt.getMaxRows());
+    }
+  }
+
+  /**
+   * Test setting a valid maxRows value
+   */
+  @Test
+  public void testValidSetMaxRows() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      // Setting positive value
+      int valueToSet = RANDOMIZER.nextInt(59) + 1;
+      stmt.setMaxRows(valueToSet);
+      assertEquals(valueToSet, stmt.getMaxRows());
+    }
+  }
+
+  /**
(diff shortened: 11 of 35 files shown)


exec/jdbc/src/main/java/org/apache/drill/jdbc/DrillPreparedStatement.java
```diff
@@ -32,4 +33,25 @@
  */
 public interface DrillPreparedStatement extends PreparedStatement {
 
+  /**
+   * @throws  SQLException
+   *            Any SQL exception
+   */
+  @Override
+  int getMaxRows() throws SQLException;
```
vvysotskyi: What was the reason for declaring this method and the method below?They both are already available in `Statement` interface.

exec/jdbc/src/main/java/org/apache/drill/jdbc/DrillStatement.java
```diff
@@ -53,6 +53,28 @@ void setQueryTimeout( int seconds )
              JdbcApiSqlException,
              SQLException;
 
+  /**
+   * @throws  SQLException
+   *            Any SQL exception
+   */
+  @Override
+  int getMaxRows() throws SQLException;
```
vvysotskyi: The same question as above.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillCursor.java
```diff
@@ -361,6 +361,9 @@ void close() {
             ExecConstants.JDBC_BATCH_QUEUE_THROTTLING_THRESHOLD );
     resultsListener = new ResultsListener(this, batchQueueThrottlingThreshold);
     currentBatchHolder = new RecordBatchLoader(client.getAllocator());
+
+    // Set Query Timeout and MaxRows
```
vvysotskyi: Looks like `MaxRows` wasn't set here.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
```diff
@@ -259,4 +261,17 @@ public void setObject(int parameterIndex, Object x, SQLType targetSqlType) throw
     checkOpen();
     super.setObject(parameterIndex, x, targetSqlType);
   }
+
+  @Override
+  public void setLargeMaxRows(long maxRowCount) throws SQLException {
+    Statement setMaxStmt = this.connection.createStatement();
+    setMaxStmt.execute("ALTER SESSION SET `" + ExecConstants.QUERY_MAX_ROWS + "`="+maxRowCount);
```
vvysotskyi:
```suggestion
    setMaxStmt.execute("ALTER SESSION SET `" + ExecConstants.QUERY_MAX_ROWS + "`=" + maxRowCount);
```

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillStatementImpl.java
```diff
@@ -270,4 +271,10 @@ public void setResultSet(AvaticaResultSet resultSet) {
   public void setUpdateCount(int value) {
     updateCount = value;
   }
+
+  @Override
+  public void setLargeMaxRows(long maxRowCount) throws SQLException {
+    execute("ALTER SESSION SET `" + ExecConstants.QUERY_MAX_ROWS + "`="+maxRowCount);
```
vvysotskyi: And please fix spaces here

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillStatementImpl.java
```diff
@@ -270,4 +271,10 @@ public void setResultSet(AvaticaResultSet resultSet) {
...
+    this.maxRowCount = maxRowCount;
```
vvysotskyi: Looks like `maxRowCount` is taken from avatica. Does it provide similar functionality we want to implement?

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -72,12 +73,18 @@
...
+  private static final Random RANDOMIZER = new Random(20150304);
 
   private static final String SYS_VERSION_SQL = "select * from sys.version";
   private static final String SYS_RANDOM_SQL =
       "SELECT cast(random() as varchar) as myStr FROM (VALUES(1)) " +
       "union SELECT cast(random() as varchar) as myStr FROM (VALUES(1)) " +
       "union SELECT cast(random() as varchar) as myStr FROM (VALUES(1)) ";
+  private static final String SYS_OPTIONS_SQL = "SELECT * FROM sys.options";
+  private static final String SYS_OPTIONS_SQL_LIMIT_10 = "SELECT * FROM sys.options LIMIT 12";
+  private static final String ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X = "ALTER SYSTEM SET `"+ExecConstants.QUERY_MAX_ROWS+"`=";
```
vvysotskyi:
```suggestion
  private static final String ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X = "ALTER SYSTEM SET `" + ExecConstants.QUERY_MAX_ROWS + "`=";
```

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillCursor.java
kkhatua: Actually, we don't need it here any more. I'll fix the comment.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
kkhatua: 👍

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillStatementImpl.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: 👍  Will fix for `StatementTest` as well

exec/java-exec/src/main/java/org/apache/drill/exec/server/options/SystemOptionManager.java
```diff
@@ -283,7 +283,9 @@
       new OptionDefinition(ExecConstants.NDV_BLOOM_FILTER_FPOS_PROB_VALIDATOR),
       new OptionDefinition(ExecConstants.RM_QUERY_TAGS_VALIDATOR,
         new OptionMetaData(OptionValue.AccessibleScopes.SESSION_AND_QUERY, false, false)),
-      new OptionDefinition(ExecConstants.RM_QUEUES_WAIT_FOR_PREFERRED_NODES_VALIDATOR)
+      new OptionDefinition(ExecConstants.RM_QUEUES_WAIT_FOR_PREFERRED_NODES_VALIDATOR),
+      new OptionDefinition(ExecConstants.RETURN_RESULT_SET_FOR_DDL_VALIDATOR),
```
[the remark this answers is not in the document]
kkhatua: Looks like this slipped through when resolving merge conflicts with the latest master branch.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -283,6 +298,21 @@ public RemoteFunctionRegistry getRemoteFunctionRegistry() {
     return drillbitContext.getRemoteFunctionRegistry();
   }
 
+  /**
+   * Returns the maximum size of auto-limited resultset
+   * @return Maximum size of auto-limited resultSet
+   */
+  public int getAutoLimitRowCount() {
```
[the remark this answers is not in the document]
kkhatua: Ok. Since I was using it at 4 other places, It seemed cleaner to have a single API instead of resolving it through `getOptions().getOption...`

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -107,6 +108,20 @@ public QueryContext(final UserSession session, final DrillbitContext drillbitCon
...
+    if (sessionMaxRowCount > 0 && defaultMaxRowCount > 0) {
+      this.autoLimitRowCount = Math.min(sessionMaxRowCount, defaultMaxRowCount);
+    } else {
+      this.autoLimitRowCount = Math.max(sessionMaxRowCount, defaultMaxRowCount);
+    }
+    if (autoLimitRowCount == defaultMaxRowCount && defaultMaxRowCount != sessionMaxRowCount) {
+      // Required to indicate via OptionScope=QueryLevel that session limit is overridden by system limit
+      queryOptions.setLocalOption(ExecConstants.QUERY_MAX_ROWS, autoLimitRowCount);
+    }
+    logger.debug("ResultSet size is auto-limited to {} rows [Session: {} / Default: {}]", this.autoLimitRowCount, sessionMaxRowCount, defaultMaxRowCount);
```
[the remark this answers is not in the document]
kkhatua: Good point. Will mark it for when autoLimit is non-zero.

exec/jdbc/src/main/java/org/apache/drill/jdbc/DrillPreparedStatement.java
kkhatua: I think I was declaring it because it seemed Avatica preferred using `setLargeMaxRows() / getLargeMaxRows()` . I, anyway, cannot really override this (all calls to `setMaxRows() / getMaxRows()` are redirected to the `...Large...` methods), so I'll remove it.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillStatementImpl.java
kkhatua: We need this here to ensure that when `getLargeMaxRows()` is called, we are reading it back from the value that was set using `setLargeMaxRows()`. Avatica only holds the value that has been set.

kkhatua:
@vvysotskyi  made changes and verified that tests passed.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -143,13 +143,18 @@ public Foreman(final WorkerBee bee, final DrillbitContext drillbitContext,
...
+      connection.getSession().getOptions().setLocalOption(ExecConstants.QUERY_MAX_ROWS, autoLimit);
+    }
     this.queryContext = new QueryContext(connection.getSession(), drillbitContext, queryId);
     this.queryManager = new QueryManager(queryId, queryRequest, drillbitContext.getStoreProvider(),
         drillbitContext.getClusterCoordinator(), this);
     this.queryRM = drillbitContext.getResourceManager().newQueryRM(this);
     this.fragmentsRunner = new FragmentsRunner(bee, initiatingClient, drillbitContext, this);
-    this.queryStateProcessor = new QueryStateProcessor(queryIdString, queryManager, drillbitContext, new ForemanResult());
     this.profileOption = setProfileOption(queryContext.getOptions());
+    this.queryStateProcessor = new QueryStateProcessor(queryIdString, queryManager, drillbitContext, new ForemanResult());
```
vvysotskyi: Could you please explain the reason for this change?

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
```diff
@@ -541,9 +546,9 @@ private void runPreparedStatement(final PreparedStatementHandle preparedStatemen
           ServerPreparedStatementState.PARSER.parseFrom(preparedStatementHandle.getServerInfo());
     } catch (final InvalidProtocolBufferException ex) {
       throw UserException.parseError(ex)
-          .message("Failed to parse the prepared statement handle. " +
-              "Make sure the handle is same as one returned from create prepared statement call.")
-          .build(logger);
+      .message("Failed to parse the prepared statement handle. " +
```
vvysotskyi: The initial indentation was better, please revert this change.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/QueryManager.java
```diff
@@ -379,6 +379,12 @@ private QueryProfile getQueryProfile(UserException ex) {
       profileBuilder.setQuery(queryText);
     }
 
+    final int autoLimitRowCount = foreman.getQueryContext().getOptions().getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue();
```
vvysotskyi: Please remove `final` modifier here and in other places where it is not required.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
```diff
@@ -259,4 +261,17 @@ public void setObject(int parameterIndex, Object x, SQLType targetSqlType) throw
     checkOpen();
     super.setObject(parameterIndex, x, targetSqlType);
   }
+
+  @Override
+  public void setLargeMaxRows(long maxRowCount) throws SQLException {
+    Statement setMaxStmt = this.connection.createStatement();
+    setMaxStmt.execute("ALTER SESSION SET `" + ExecConstants.QUERY_MAX_ROWS + "`=" + maxRowCount);
+    setMaxStmt.close();
```
vvysotskyi: Please use `try with resources` for `Statement` instead of manually closing it. Also, please rename `setMaxStmt` into `statement`.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
```diff
@@ -259,4 +261,17 @@ public void setObject(int parameterIndex, Object x, SQLType targetSqlType) throw
...
+    this.maxRowCount = maxRowCount;
```
vvysotskyi: Please remove this line and put into the beginning of the method call `super.setLargeMaxRows(maxRowCount)`, so it will check that connection is open and validate `maxRowCount` value.

The same for changes in `DrillStatementImpl`.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
```diff
@@ -259,4 +261,17 @@ public void setObject(int parameterIndex, Object x, SQLType targetSqlType) throw
...
+  }
+
+  @Override
+  public long getLargeMaxRows() {
```
vvysotskyi: Please remove this method since Avatica provides its implementation.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
 
+  ////////////////////////////////////////
+  // Query maxRows methods:
+
+  /**
+   * Test for reading of default max rows
+   */
+  @Test
+  public void   testDefaultGetMaxRows() throws SQLException {
+    try(PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
```
vvysotskyi: Please add spaces here and in other places where needed (and remove where they are excessive, like in the signature of this test).
```suggestion
    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+      int valueToSet = RANDOMIZER.nextInt(9)+1; //range:[1-9]
+      pStmt.setMaxRows(valueToSet);
+      pStmt.executeQuery();
+      ResultSet rs = pStmt.getResultSet();
+      int rowCount = 0;
+      while (rs.next()) {
+        rs.getBytes(1);
+        rowCount++;
+      }
+      logger.info("MaxRows was set as {} and query limited to 10 returned a resultSet of {} rows", valueToSet, rowCount);
```
vvysotskyi: Please remove loggers from unit tests, it is redundant.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -462,4 +618,25 @@ public void testParamSettingWhenUnsupportedTypeSaysUnsupported() throws SQLExcep
...
+  // Sets the SystemMaxRows option
+  private void setSystemMaxRows(int sysValueToSet) throws SQLException {
+    // Acquire lock if Non-Zero Value (i.e. a test is in progress)
+    if (sysValueToSet != 0) {
+      maxRowsSysOptionLock.acquireUninterruptibly();
+    }
+    // Setting the value
+    try (Statement stmt = connection.createStatement()) {
+      stmt.executeQuery(ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X + sysValueToSet);
+      ResultSet rs = stmt.getResultSet();
```
vvysotskyi: Is it required to obtain a result set and iterate thru it?

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -72,12 +73,18 @@
...
   private static final String SYS_VERSION_SQL = "select * from sys.version";
   private static final String SYS_RANDOM_SQL =
       "SELECT cast(random() as varchar) as myStr FROM (VALUES(1)) " +
       "union SELECT cast(random() as varchar) as myStr FROM (VALUES(1)) " +
       "union SELECT cast(random() as varchar) as myStr FROM (VALUES(1)) ";
+  private static final String SYS_OPTIONS_SQL = "SELECT * FROM sys.options";
+  private static final String SYS_OPTIONS_SQL_LIMIT_10 = "SELECT * FROM sys.options LIMIT 12";
+  private static final String ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X = "ALTER SYSTEM SET `" + ExecConstants.QUERY_MAX_ROWS + "`=";
+  //Locks used across StatementTest and PreparedStatementTest
+  private static Semaphore maxRowsSysOptionLock = StatementTest.maxRowsSysOptionLock;
```
vvysotskyi: `connection` is created in this class, but the same `Semaphore` is used from `StatementTest` class for this test class and another one. Should they use their own instances of `Semaphore`?

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -462,4 +618,25 @@ public void testParamSettingWhenUnsupportedTypeSaysUnsupported() throws SQLExcep
     }
   }
 
+
+  // Sets the SystemMaxRows option
+  private void setSystemMaxRows(int sysValueToSet) throws SQLException {
```
vvysotskyi: I'm not sure that the approach used for blocking the connection should be implemented in this method. I think unit tests should acquire and release the lock.

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,173 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+
+  /**
+   * Test setting maxSize as zero (i.e. no Limit)
+   */
+  @Test
+  public void testSetMaxRowsAsZero() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      stmt.setMaxRows(0);
+      stmt.executeQuery(SYS_OPTIONS_SQL);
+      ResultSet rs = stmt.getResultSet();
```
vvysotskyi: `ResultSet` should be also closed.

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/Drill2489CallsAfterCloseThrowExceptionsTest.java
```diff
@@ -537,7 +547,15 @@ public void testclosedPreparedStmtOfOpenConnMethodsThrowRight() {
         new ClosedPreparedStatementChecker(PreparedStatement.class,
                                            closedPreparedStmtOfOpenConn);
 
-    checker.testAllMethods();
+    //List of methods now supported
+    @SuppressWarnings("serial")
```
vvysotskyi: Please revert these changes in this class once calls to Avatica methods are used.

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -107,6 +107,23 @@ public QueryContext(final UserSession session, final DrillbitContext drillbitCon
       this.table = drillbitContext.getOperatorTable();
     }
 
+    // Checking for limit on ResultSet rowcount and if user attempting to override the system value
+    final int sessionMaxRowCount = queryOptions.getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue();
```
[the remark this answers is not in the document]
kkhatua: In previous reviews, I've been asked to keep immutable variables as `final`.

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
kkhatua: Since I was originally changing the ForemanResult, I had reordered to ensure that the `queryContext` was set with the right values before the `queryStateProcessor` is initialized. I'll reorder to the original way,

exec/java-exec/src/main/java/org/apache/drill/exec/work/foreman/Foreman.java
kkhatua: Ok.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: No. If the query fails, you won't get the resultset object.

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
kkhatua: Ok, but `Statement.close()` also closes the `ResultSet`

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: The unit tests use this method to set and release the lock. It reduces the lock code being repeated across unit tests.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: I tried that, but the unit tests attempt to test with setting up specific SYSTEM values. Without this, there is a race condition where both `StatementTest` and `PreparedStatementTest` attempt to set SYSTEM values... and get different results (resulting in test failures).

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
```diff
@@ -283,6 +300,13 @@ public RemoteFunctionRegistry getRemoteFunctionRegistry() {
     return drillbitContext.getRemoteFunctionRegistry();
   }
 
+  /**
+   * Allows to disable autolimit in case it is not applicable
+   */
+  public void disableAutoLimit() {
```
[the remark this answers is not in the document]
kkhatua: 👍

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
kkhatua: 👍

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
kkhatua: Ok, but I think it should be done post-close() because in case of a failure, we dont want to accidentally set through `super` and then worry about rolling back.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: Same as previous comment regarding `final`

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/Drill2489CallsAfterCloseThrowExceptionsTest.java
kkhatua: Not sure what you meant by the change. Should I remove the SuppressWarnings?

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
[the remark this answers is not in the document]
vvysotskyi: I'm not sure that it was exactly the same case, but in this case `final` is redundant. The scope of variable usage is small, and there is no risk of accidental changing its value.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
vvysotskyi: It is important to check that connection is opened at first, and it is done in the parent method.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
vvysotskyi: What is the purpose of this message? It will be newer read by dev except for the case when a test failed and logging is enabled. But the real purpose may be when this message is moved into an assertion check, so in the case of test failure, it is printed and in other cases, logging does not happen.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
kkhatua: 👍

exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java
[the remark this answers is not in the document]
kkhatua: 👍

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: I had this when i was writing test cases. So, you're right... these are redundant.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
```diff
@@ -259,4 +261,13 @@ public void setObject(int parameterIndex, Object x, SQLType targetSqlType) throw
     checkOpen();
     super.setObject(parameterIndex, x, targetSqlType);
   }
+
+  @Override
+  public void setLargeMaxRows(long maxRowCount) throws SQLException {
+    super.setLargeMaxRows(maxRowCount);
+    try (Statement statement = this.connection.createStatement()) {
+      statement.execute("ALTER SESSION SET `" + ExecConstants.QUERY_MAX_ROWS + "`=" + maxRowCount);
+      statement.close();
```
vvysotskyi: Please remove this line, `statement` will be closed automatically since it is used in `try with resources`.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
vvysotskyi: Looks like these tests are connected to the same drillbit. In this case, the same `Semaphore` instance should be used. Thanks for the clarification.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -285,7 +292,7 @@ public void testInvalidSetQueryTimeout() throws SQLException {
   public void testValidSetQueryTimeout() throws SQLException {
     try (PreparedStatement stmt = connection.prepareStatement(SYS_VERSION_SQL)) {
       //Setting positive value
-      int valueToSet = new Random(20150304).nextInt(59)+1;
+      int valueToSet = RANDOMIZER.nextInt(59)+1;
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(59) + 1;
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
   }
 
+  ////////////////////////////////////////
+  // Query maxRows methods:
+
+  /**
+   * Test for reading of default max rows
+   */
+  @Test
+  public void   testDefaultGetMaxRows() throws SQLException {
```
vvysotskyi:
```suggestion
  public void testDefaultGetMaxRows() throws SQLException {
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+   * Test Invalid parameter by giving negative maxRows value
+   */
+  @Test
+  public void testInvalidSetMaxRows() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      //Setting negative value
+      int valueToSet = -10;
+      try {
+        pStmt.setMaxRows(valueToSet);
+      } catch ( final SQLException e) {
```
vvysotskyi:
```suggestion
      } catch (final SQLException e) {
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+        assertThat( e.getMessage(), containsString("VALIDATION ERROR: Option exec.query.max_rows must be between 0 and ") );
```
vvysotskyi:
```suggestion
        assertThat(e.getMessage(), containsString("VALIDATION ERROR: Option exec.query.max_rows must be between 0 and "));
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+  }
+
+  /**
+   * Test setting a valid maxRows value
+   */
+  @Test
+  public void testValidSetMaxRows() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      //Setting positive value
+      int valueToSet = RANDOMIZER.nextInt(59)+1;
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(59) + 1;
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+      pStmt.setMaxRows(valueToSet);
+      assertEquals( valueToSet, pStmt.getMaxRows() );
```
vvysotskyi:
```suggestion
      assertEquals(valueToSet, pStmt.getMaxRows());
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value lower than existing query limit
+   */
+  @Test
+  public void testSetMaxRowsLowerThanQueryLimit() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL_LIMIT_10)) {
+      int valueToSet = RANDOMIZER.nextInt(9)+1; //range:[1-9]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(9) + 1; // range: [1-9]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value higher than existing query limit
+   */
+  @Test
+  public void testSetMaxRowsHigherThanQueryLimit() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL_LIMIT_10)) {
+      int valueToSet = RANDOMIZER.nextInt(10)+11; //range:[11-20]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(10) + 11; // range: [11-20]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+      assertTrue(valueToSet > rowCount);
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value lower than existing SYSTEM limit
+   */
+  @Test
+  public void testSetMaxRowsLowerThanSystemLimit() throws SQLException {
+    int sysValueToSet = RANDOMIZER.nextInt(5)+6; //range:[6-10]
```
vvysotskyi:
```suggestion
    int sysValueToSet = RANDOMIZER.nextInt(5) + 6; // range: [6-10]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    setSystemMaxRows(sysValueToSet);
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      int valueToSet = RANDOMIZER.nextInt(5)+1; //range:[1-5]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(5) + 1; // range: [1-5]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    }
+    setSystemMaxRows(0); //RESET
+  }
+
+  /**
+   * Test setting maxSize at a value higher than existing SYSTEM limit
+   */
+  @Test
+  public void testSetMaxRowsHigherThanSystemLimit() throws SQLException {
+    int sysValueToSet = RANDOMIZER.nextInt(5)+6; //range:[6-10]
```
vvysotskyi:
```suggestion
    int sysValueToSet = RANDOMIZER.nextInt(5) + 6; // range: [6-10]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    setSystemMaxRows(sysValueToSet);
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      int valueToSet = RANDOMIZER.nextInt(5)+11; //range:[11-15]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(5) + 11; // range: [11-15]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
vvysotskyi: It looks like a hack to release the lock for one value and acquire for others. Except adding locks into tests directly, the more clear way may be to add a boolean flag into method argument whether to acquire or release the lock.

Also, I'm wondering whether these tests which change autoLimit does not affect other tests, i.e. when one test sets autoLimit to 3, but another test (which does not contain `setSystemMaxRows()` call, so it is not locked) expects result with 5 rows.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
vvysotskyi: Ok, thanks for the clarification.

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+   * Test Invalid parameter by giving negative maxRows value
+   */
+  @Test
+  public void testInvalidSetMaxRows() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      // Setting negative value
+      int valueToSet = -10;
+      try {
+        stmt.setMaxRows(valueToSet);
+      } catch ( final SQLException e) {
```
vvysotskyi:
```suggestion
      } catch (final SQLException e) {
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+        assertThat( e.getMessage(), containsString("VALIDATION ERROR: Option exec.query.max_rows must be between 0 and ") );
```
vvysotskyi:
```suggestion
        assertThat(e.getMessage(), containsString("VALIDATION ERROR: Option exec.query.max_rows must be between 0 and "));
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+  }
+
+  /**
+   * Test setting a valid maxRows value
+   */
+  @Test
+  public void testValidSetMaxRows() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      // Setting positive value
+      int valueToSet = RANDOMIZER.nextInt(59)+1;
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(59) + 1;
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+      stmt.setMaxRows(valueToSet);
+      assertEquals( valueToSet, stmt.getMaxRows() );
```
vvysotskyi:
```suggestion
      assertEquals(valueToSet, stmt.getMaxRows());
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value lower than existing query limit
+   */
+  @Test
+  public void testSetMaxRowsLowerThanQueryLimit() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      int valueToSet = RANDOMIZER.nextInt(9)+1; // range:[1-9]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(9) + 1; // range: [1-9]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value higher than existing query limit
+   */
+  @Test
+  public void testSetMaxRowsHigherThanQueryLimit() throws SQLException {
+    try (Statement stmt = connection.createStatement()) {
+      int valueToSet = RANDOMIZER.nextInt(10)+11; // range:[11-20]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(10) + 11; // range: [11-20]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+      assertTrue(valueToSet > rowCount);
+    }
+  }
+
+  /**
+   * Test setting maxSize at a value lower than existing SYSTEM limit
+   */
+  @Test
+  public void testSetMaxRowsLowerThanSystemLimit() throws SQLException {
+    int sysValueToSet = RANDOMIZER.nextInt(5)+6; // range:[6-10]
```
vvysotskyi:
```suggestion
    int sysValueToSet = RANDOMIZER.nextInt(5) + 6; // range: [6-10]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    setSystemMaxRows(sysValueToSet);
+    try (Statement stmt = connection.createStatement()) {
+      int valueToSet = RANDOMIZER.nextInt(5)+1; // range:[1-5]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(5) + 1; // range: [1-5]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    }
+    setSystemMaxRows(0); // Reset value
+  }
+
+  /**
+   * Test setting maxSize at a value higher than existing SYSTEM limit
+   */
+  @Test
+  public void testSetMaxRowsHigherThanSystemLimit() throws SQLException {
+    int sysValueToSet = RANDOMIZER.nextInt(5)+6; // range:[6-10]
```
vvysotskyi:
```suggestion
    int sysValueToSet = RANDOMIZER.nextInt(5) + 6; // range: [6-10]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -239,4 +246,174 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+    setSystemMaxRows(sysValueToSet);
+    try (Statement stmt = connection.createStatement()) {
+      int valueToSet = RANDOMIZER.nextInt(5)+11; // range:[11-15]
```
vvysotskyi:
```suggestion
      int valueToSet = RANDOMIZER.nextInt(5) + 11; // range: [11-15]
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/Drill2489CallsAfterCloseThrowExceptionsTest.java
```diff
@@ -537,7 +547,14 @@ public void testclosedPreparedStmtOfOpenConnMethodsThrowRight() {
         new ClosedPreparedStatementChecker(PreparedStatement.class,
                                            closedPreparedStmtOfOpenConn);
 
-    checker.testAllMethods();
+    //List of methods now supported
+    Set<String> methodsToSkip = new LinkedHashSet<String>() {{
```
vvysotskyi: I meant to revert all these changes, they are not needed for now.

exec/jdbc/src/main/java/org/apache/drill/jdbc/impl/DrillPreparedStatementImpl.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: 👍 
Will check across all the tests.

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/Drill2489CallsAfterCloseThrowExceptionsTest.java
kkhatua: Ok. These tests were failing because it we had implemented the `setMaxRows()` and `getMaxRows()` methods. These tests, it seems, expected them to throw exceptions.
With the `super.setLargeMaxRows()`, this should not be necessary anymore.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: There are _getter-setter_ method tests that don't rely on what the `SYSTEM` value is, so we don't need to acquire explicit locks. Only when a query is being executed, do we need to ensure that changes to the `SYSTEM` value are synchronized because that is when `QueryContext` will apply a `SESSION` or `SYSTEM` value.  Using the lock to synchronize the change, indirectly synchronizes these specific tests as well.
We don't have any other control on the parallelism or order of the tests within the 2 suites (`PreparedStatementTest` and `StatementTest`)... so using a common lock across both tests is the safest way to run such tests with an element of randomness (using the `RANDOMIZER` object).

vvysotskyi: `PreparedStatementTest` contains `testSetQueryTimeoutAsZero()` test which verifies that the result should have 3 rows. This test will be executed without considering the `Semaphore` lock. So for the case when one of your tests changed the value of autoLimit, and before resetting the value this test is executed, it will return fewer rows than it should.

So adding these tests may introduce random failures of other tests.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
```diff
@@ -410,6 +417,155 @@ public void testNonTriggeredQueryTimeout() throws SQLException {
...
+   */
+  @Test
+  public void testInvalidSetMaxRows() throws SQLException {
+    try (PreparedStatement pStmt = connection.prepareStatement(SYS_OPTIONS_SQL)) {
+      //Setting negative value
+      int valueToSet = -10;
+      try {
+        pStmt.setMaxRows(valueToSet);
+      } catch (final SQLException e) {
+        assertThat(e.getMessage(), containsString("illegal maxRows value: " + valueToSet));
```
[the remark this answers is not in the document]
kkhatua: I would think that a thrown exception is sufficient, but I'll add the check to compare against the value before the `setMaxRows()` is called.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: Strangely enough, these tests have never failed. I can set the `assert` to check for a non-zero value. Alternatively, I could disable the tests:
`testSetMaxRowsHigherThanSystemLimit()` and `testSetMaxRowsLowerThanSystemLimit()`

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/Drill2489CallsAfterCloseThrowExceptionsTest.java
```diff
@@ -537,8 +537,6 @@ public void testclosedPreparedStmtOfOpenConnMethodsThrowRight() {
         new ClosedPreparedStatementChecker(PreparedStatement.class,
                                            closedPreparedStmtOfOpenConn);
 
-    checker.testAllMethods();
```
ihuzenko: Does deletion of the line means that test now checks nothing ? Maybe it's accidental change and needs revert or test method may be safely removed ?

ihuzenko:
Hello @kkhatua, generally the PR looks good, only one comment related to test. And interesting thing with UI is that setting auto limit value to "100 " is prohibited because of space sign after number, should we trim all accepted values ?

exec/jdbc/src/test/java/org/apache/drill/jdbc/test/Drill2489CallsAfterCloseThrowExceptionsTest.java
kkhatua: You're right. It was an accidental change, which is why it went through the tests. I'll restore it. Thanks for catching it.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementTest.java
kkhatua: As per our chat, I've introduced `@Before` and `@After` methods for synchronizing the `system` level modifications to the options.

kkhatua:
@vvysotskyi , @ihuzenko 
I've done the changes and verified the tests. If everything is fine, I'll rebase on the latest master (there are small conflicts due to new commits on master introducing additional system options)

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementTest.java
```diff
@@ -69,6 +78,17 @@ public static void tearDownStatement() throws SQLException {
     connection.close();
   }
 
+  @Before
```
vvysotskyi: This will slow down all the tests in this class besides your tests.
Also, since connection string is `jdbc:drill:zk=local`, so separate embedded drill instance will be used for every test class and there is no need to lock other test classes.

Could you please move all your tests into a separate class and add lock there?

kkhatua: Ok. I'll move only the `get|setMaxRows` related tests.

kkhatua:
Done.
This is the reference branch before the squash: https://github.com/kkhatua/drill/commits/optionMaxRows into commit ID `a17fddb27bef647024748ef53aa1b336bc17da79`

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
```diff
@@ -0,0 +1,265 @@
...
+
+import org.junit.experimental.categories.Category;
+
+/**
+ * Test for Drill's implementation of PreparedStatement's get/setMaxRows methods
+ */
+@Category(JdbcTest.class)
+public class PreparedStatementMaxRowsTest extends JdbcTestBase {
+
+//  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(PreparedStatementMaxRowsTest.class);
```
vvysotskyi: Please remove this line.

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
```diff
@@ -0,0 +1,265 @@
...
+  private static final Random RANDOMIZER = new Random(20150304);
+
+  private static final String SYS_OPTIONS_SQL = "SELECT * FROM sys.options";
+  private static final String SYS_OPTIONS_SQL_LIMIT_10 = "SELECT * FROM sys.options LIMIT 12";
+  private static final String ALTER_SYS_OPTIONS_MAX_ROWS_LIMIT_X = "ALTER SYSTEM SET `" + ExecConstants.QUERY_MAX_ROWS + "`=";
+  // Lock used by all tests to avoid corrupting test scenario
+  static final Semaphore maxRowsSysOptionLock = new Semaphore(1);
```
vvysotskyi: please make it private

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
```diff
@@ -0,0 +1,265 @@
...
+  private static Connection connection;
+
+
+  @BeforeClass
+  public static void setUpConnection() throws SQLException {
+    Driver.load();
+    Properties properties = new Properties();
+    // Increased prepared statement creation timeout so test doesn't timeout on my laptop
+    properties.setProperty(ExecConstants.bootDefaultFor(ExecConstants.CREATE_PREPARE_STATEMENT_TIMEOUT_MILLIS), "30000");
+    connection = DriverManager.getConnection( "jdbc:drill:zk=local", properties);
```
vvysotskyi:
```suggestion
    connection = DriverManager.getConnection("jdbc:drill:zk=local", properties);
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
```diff
@@ -0,0 +1,265 @@
...
+  static final Semaphore maxRowsSysOptionLock = new Semaphore(1);
+
+  private static Connection connection;
+
+
+  @BeforeClass
+  public static void setUpConnection() throws SQLException {
+    Driver.load();
+    Properties properties = new Properties();
+    // Increased prepared statement creation timeout so test doesn't timeout on my laptop
```
vvysotskyi: Please rephrase it to:
```suggestion
    // Increased prepared statement creation timeout to avoid timeout failures 
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
```diff
@@ -0,0 +1,265 @@
...
+    properties.setProperty(ExecConstants.bootDefaultFor(ExecConstants.CREATE_PREPARE_STATEMENT_TIMEOUT_MILLIS), "30000");
+    connection = DriverManager.getConnection( "jdbc:drill:zk=local", properties);
+    try(Statement stmt = connection.createStatement()) {
```
vvysotskyi:
```suggestion
    try (Statement stmt = connection.createStatement()) {
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  @AfterClass
+  public static void tearDownConnection() throws SQLException {
+    if (connection != null) {
+      try (Statement stmt = connection.createStatement()) {
+        stmt.execute(String.format("alter session set `%s` = false", PlannerSettings.ENABLE_DECIMAL_DATA_TYPE_KEY));
+      }
+    }
+    connection.close();
```
vvysotskyi: Please move this call inside if statement. Also, since the connection is closed, there is no need to restore session options.

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementMaxRowsTest.java
```diff
@@ -0,0 +1,247 @@
...
+import java.sql.Statement;
+import java.util.Random;
+import java.util.concurrent.Semaphore;
+import java.sql.SQLException;
+
+/**
+ * Test for Drill's implementation of Statement's get/setMaxRows methods
+ */
+@Category(JdbcTest.class)
+public class StatementMaxRowsTest extends JdbcTestBase {
```
vvysotskyi: The same as for `PreparedStatementMaxRowsTest.java`

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementMaxRowsTest.java
```diff
@@ -0,0 +1,247 @@
...
+  static final Semaphore maxRowsSysOptionLock = new Semaphore(1);
+
+  private static Connection connection;
+
+  @BeforeClass
+  public static void setUpStatement() throws SQLException {
+    // (Note: Can't use JdbcTest's connect(...) because JdbcTest closes
+    // Connection--and other JDBC objects--on test method failure, but this test
+    // class uses some objects across methods.)
+    connection = new Driver().connect( "jdbc:drill:zk=local", null );
```
vvysotskyi:
```suggestion
    connection = new Driver().connect("jdbc:drill:zk=local", null);
```

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/PreparedStatementMaxRowsTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementMaxRowsTest.java
kkhatua: 👍

exec/jdbc/src/test/java/org/apache/drill/jdbc/StatementMaxRowsTest.java
kkhatua: 👍

kkhatua:
@vvysotskyi made the changes in the `optionsMaxRows` branch and squashed in the `DRILL-7048` branch.

PR #1720, opened 2019-03-29
DRILL-7140: RM: Drillbits fail with "No enum constant org.apache.dril…

sohami:
…l.exec.resourcemgr.config.selectionpolicy.QueueSelectionPolicy.SelectionPolicy.bestfit"


HanumathRao:
Changes look good to me. +1

vvysotskyi:
@HanumathRao, in future, when approving the PR, please check that Jira is in Reviewable state and please add `ready-to-commit` label, so the Jira will be present in the list of Jiras, ready to be merged.

PR #1723, opened 2019-03-31
DRILL-7063: Seperate metadata cache file into summary, file metadata

dvjyothsna:
DRILL-7065: Added backward compatibility to be able to fall back to read previous versions of metadata cache files if current version is not available.

Drill-7066: Auto-refresh to pick up interesting columns from the existing columns and create metadata only for the existing interesting columns

Ignored few tests in testParquetRowGroupFilterEval which have filters on non-existent columns. These will be enabled once run-time row group pruning is enabled.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
@@ -161,7 +161,7 @@ public class RefreshMetadataHandler extends DefaultSqlHandler {
    */
   private SqlNodeList getColumnList(final SqlRefreshMetadata sqlrefreshMetadata) {
     SqlNodeList columnList = sqlrefreshMetadata.getFieldList();
-    if (columnList == null || !SqlNodeList.isEmptyList(columnList)) {
+    if (SqlNodeList.isEmptyList(columnList)) {
       columnList = new SqlNodeList(SqlParserPos.ZERO);
       columnList.add(new SqlIdentifier(SchemaPath.STAR_COLUMN.rootName(), SqlParserPos.ZERO));
     }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
@@ -154,7 +154,7 @@ public abstract class BaseParquetMetadataProvider implements ParquetMetadataProv
           metadata = getRowGroupsMeta();
         }
         tableStatistics.put(TableStatisticsKind.ROW_COUNT.getName(), TableStatisticsKind.ROW_COUNT.mergeStatistics(metadata));
-        columnsStatistics = ParquetTableMetadataUtils.mergeColumnsStatistics(metadata, fields.keySet(), ParquetTableMetadataUtils.PARQUET_STATISTICS);
+        columnsStatistics = ParquetTableMetadataUtils.mergeColumnsStatistics(metadata, fields.keySet(), ParquetTableMetadataUtils.PARQUET_STATISTICS, parquetTableMetadata);
       } else {
         columnsStatistics = new HashMap<>();
         tableStatistics.put(TableStatisticsKind.ROW_COUNT.getName(), getParquetGroupScanStatistics().getRowCount());
@@ -166,6 +166,7 @@ public abstract class BaseParquetMetadataProvider implements ParquetMetadataProv
               ColumnStatisticsKind.NULLS_COUNT.getName(), getParquetGroupScanStatistics().getRowCount() - columnValueCount);
           columnsStatistics.put(partitionColumn, new ColumnStatisticsImpl(stats, ParquetTableMetadataUtils.getNaturalNullsFirstComparator()));
         }
+        columnsStatistics.putAll(ParquetTableMetadataUtils.populateNonInterestingColumnsStats(columnsStatistics.keySet(), parquetTableMetadata));
       }
       tableMetadata = new FileTableMetadata(tableName, tableLocation, schema, columnsStatistics, tableStatistics,
           -1, "root", partitionKeys);
@@ -232,7 +233,7 @@ public abstract class BaseParquetMetadataProvider implements ParquetMetadataProv
           }
 
           for (Map.Entry<Object, Collection<Path>> valueLocationsEntry : partitionsForValue.asMap().entrySet()) {
-            HashMap<SchemaPath, ColumnStatistics> columnsStatistics = new HashMap<>();
+            Map<SchemaPath, ColumnStatistics> columnsStatistics = new HashMap<>();
 
             Map<String, Object> statistics = new HashMap<>();
             Object partitionKey = valueLocationsEntry.getKey();
@@ -244,8 +245,7 @@ public abstract class BaseParquetMetadataProvider implements ParquetMetadataProv
             statistics.put(TableStatisticsKind.ROW_COUNT.getName(), GroupScan.NO_COLUMN_STATS);
             columnsStatistics.put(partitionColumn,
                 new ColumnStatisticsImpl<>(statistics,
-                    ParquetTableMetadataUtils.getComparator(getParquetGroupScanStatistics().getTypeForColumn(partitionColumn).getMinorType())));
-
+                        ParquetTableMetadataUtils.getComparator(getParquetGroupScanStatistics().getTypeForColumn(partitionColumn).getMinorType())));
             partitions.add(new PartitionMetadata(partitionColumn, getTableMetadata().getSchema(),
                 columnsStatistics, statistics, (Set<Path>) valueLocationsEntry.getValue(), tableName, -1));
           }
@@ -313,7 +313,7 @@ public abstract class BaseParquetMetadataProvider implements ParquetMetadataProv
           }
         }
 
-        FileMetadata fileMetadata = ParquetTableMetadataUtils.getFileMetadata(fileRowGroups, tableName);
+        FileMetadata fileMetadata = ParquetTableMetadataUtils.getFileMetadata(fileRowGroups, tableName, parquetTableMetadata);
         files.put(fileMetadata.getLocation(), fileMetadata);
       }
     }
@@ -344,4 +344,5 @@ public abstract class BaseParquetMetadataProvider implements ParquetMetadataProv
   }
 
   protected abstract void initInternal() throws IOException;
+
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetFormatPlugin.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetFormatPlugin.java
@@ -297,12 +297,34 @@ public class ParquetFormatPlugin implements FormatPlugin {
       return super.isReadable(fs, selection, fsPlugin, storageEngineName, schemaConfig);
     }
 
-    private Path getMetadataPath(FileStatus dir) {
-      return new Path(dir.getPath(), Metadata.METADATA_FILENAME);
+    private Path getOldMetadataPath(FileStatus dir) {
+      return new Path(dir.getPath(), Metadata.OLD_METADATA_FILENAME);
     }
 
+    /**
+     * Check if the metadata cache files exist
+     * @param dir the path of the directory
+     * @param fs
+     * @return true if both file metadata and summary cache file exist
+     * @throws IOException in case of problems during accessing files
+     */
     private boolean metaDataFileExists(FileSystem fs, FileStatus dir) throws IOException {
-      return fs.exists(getMetadataPath(dir));
+      boolean fileExists = true;
+      for (String metaFileName : Metadata.CURRENT_METADATA_FILENAMES) {
+        Path path = new Path(dir.getPath(), metaFileName);
+        if (!fs.exists(path)) {
+          fileExists = false;
+        }
+      }
+      if (fileExists) {
+        return true;
+      } else {
+        // Check if the older version of metadata file exists
+        if (fs.exists(getOldMetadataPath(dir))) {
+          return true;
+        }
+      }
+      return false;
     }
 
     boolean isDirReadable(DrillFileSystem fs, FileStatus dir) {
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
@@ -24,6 +24,7 @@ import org.apache.drill.common.types.TypeProtos;
 import org.apache.drill.common.types.Types;
 import org.apache.drill.exec.planner.physical.PlannerSettings;
 import org.apache.drill.exec.server.options.OptionManager;
+import org.apache.drill.exec.store.parquet.metadata.MetadataBase;
 import org.apache.drill.exec.store.parquet.metadata.MetadataVersion;
 import org.apache.drill.exec.util.Utilities;
 import org.apache.drill.exec.work.ExecErrorConstants;
@@ -62,8 +63,6 @@ import java.util.concurrent.TimeUnit;
 
 import static org.apache.drill.exec.store.parquet.metadata.Metadata_V2.ColumnTypeMetadata_v2;
 import static org.apache.drill.exec.store.parquet.metadata.Metadata_V2.ParquetTableMetadata_v2;
-import static org.apache.drill.exec.store.parquet.metadata.Metadata_V3.ColumnTypeMetadata_v3;
-import static org.apache.drill.exec.store.parquet.metadata.Metadata_V3.ParquetTableMetadata_v3;
 import static org.apache.drill.exec.store.parquet.metadata.MetadataBase.ColumnMetadata;
 import static org.apache.drill.exec.store.parquet.metadata.MetadataBase.ParquetTableMetadataBase;
 import static org.apache.drill.exec.store.parquet.metadata.MetadataBase.ParquetFileMetadata;
@@ -318,7 +317,7 @@ public class ParquetReaderUtility {
     int minRowGroups = Integer.MAX_VALUE;
     int maxNumColumns = 0;
 
-    // Setting Min / Max values for V2 and V3 versions; for versions V3_3 and above need to do decoding
+    // Setting Min / Max values for V2, V3 and V4 versions; for versions V3_3 and above need to do decoding
     boolean needDecoding = new MetadataVersion(parquetTableMetadata.getMetadataVersion()).compareTo(new MetadataVersion(3, 3)) >= 0;
     for (ParquetFileMetadata file : parquetTableMetadata.getFiles()) {
       if ( timer != null ) { // for debugging only
@@ -371,20 +370,12 @@ public class ParquetReaderUtility {
    */
   private static Set<List<String>> getBinaryColumnsNames(ParquetTableMetadataBase parquetTableMetadata) {
     Set<List<String>> names = new HashSet<>();
-    if (parquetTableMetadata instanceof ParquetTableMetadata_v2) {
-      for (ColumnTypeMetadata_v2 columnTypeMetadata :
-        ((ParquetTableMetadata_v2) parquetTableMetadata).columnTypeInfo.values()) {
-        if (columnTypeMetadata.primitiveType == PrimitiveTypeName.BINARY
-            || columnTypeMetadata.primitiveType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
-          names.add(Arrays.asList(columnTypeMetadata.name));
-        }
-      }
-    } else if (parquetTableMetadata instanceof ParquetTableMetadata_v3) {
-      for (ColumnTypeMetadata_v3 columnTypeMetadata :
-        ((ParquetTableMetadata_v3) parquetTableMetadata).columnTypeInfo.values()) {
-        if (columnTypeMetadata.primitiveType == PrimitiveTypeName.BINARY
-            || columnTypeMetadata.primitiveType == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
-          names.add(Arrays.asList(columnTypeMetadata.name));
+    List<? extends MetadataBase.ColumnTypeMetadata> columnTypeMetadataList = parquetTableMetadata.getColumnTypeInfoList();
+    if (columnTypeMetadataList != null) {
+      for (MetadataBase.ColumnTypeMetadata columnTypeMetadata : columnTypeMetadataList) {
+        if (columnTypeMetadata.getPrimitiveType() == PrimitiveTypeName.BINARY
+                || columnTypeMetadata.getPrimitiveType() == PrimitiveTypeName.FIXED_LEN_BYTE_ARRAY) {
+          names.add(Arrays.asList(columnTypeMetadata.getName()));
         }
       }
     }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
@@ -94,7 +94,6 @@ public class ParquetTableMetadataProviderImpl extends BaseParquetMetadataProvide
           entries.add(new ReadEntryWithPath(fileName));
         }
       }
-
       init();
     }
   }
@@ -109,19 +108,54 @@ public class ParquetTableMetadataProviderImpl extends BaseParquetMetadataProvide
     return selectionRoot;
   }
 
+  /**
+   * Returns list of metadata cache files
+   * @param p directory path of the cache file
+   * @param fs filesystem object
+   * @return list of cache files found in the given directory path
+   */
+  public List<Path> populateMetaPaths(Path p, DrillFileSystem fs) throws IOException {
+    List<Path> metaFilepaths = new ArrayList<>();
+    for (String filename : Metadata.CURRENT_METADATA_FILENAMES) {
+      metaFilepaths.add(new Path(p, filename));
+    }
+    for (String filename : Metadata.OLD_METADATA_FILENAMES) {
+      // Read the older version of metadata file if the current version of metadata cache files donot exist.
+      if (fileExists(fs, metaFilepaths)) {
+        return metaFilepaths;
+      }
+      metaFilepaths.clear();
+      metaFilepaths.add(new Path(p, filename));
+    }
+    if (fileExists(fs, metaFilepaths)) {
+      return metaFilepaths;
+    }
+    return new ArrayList<>();
+  }
+
+  public boolean fileExists(DrillFileSystem fs, List<Path> paths) throws IOException {
+    for (Path path : paths) {
+      if (!fs.exists(path)) {
+        return false;
+      }
+    }
+    return true;
+  }
+
   @Override
   protected void initInternal() throws IOException {
     try (FileSystem processUserFileSystem = ImpersonationUtil.createFileSystem(ImpersonationUtil.getProcessUserName(), fs.getConf())) {
-      Path metaPath = null;
+      // Depending on the version of metadata this may represent more than 1 metadata file paths.
+      List<Path> metaPaths = new ArrayList<>();
       if (entries.size() == 1 && parquetTableMetadata == null) {
         Path p = Path.getPathWithoutSchemeAndAuthority(entries.get(0).getPath());
         if (fs.isDirectory(p)) {
           // Using the metadata file makes sense when querying a directory; otherwise
           // if querying a single file we can look up the metadata directly from the file
-          metaPath = new Path(p, Metadata.METADATA_FILENAME);
+          metaPaths = populateMetaPaths(p, fs);
         }
-        if (!metaContext.isMetadataCacheCorrupted() && metaPath != null && fs.exists(metaPath)) {
-          parquetTableMetadata = Metadata.readBlockMeta(processUserFileSystem, metaPath, metaContext, readerConfig);
+        if (!metaContext.isMetadataCacheCorrupted() && !metaPaths.isEmpty()) {
+          parquetTableMetadata = Metadata.readBlockMeta(processUserFileSystem, metaPaths, metaContext, readerConfig);
           if (parquetTableMetadata != null) {
             usedMetadataCache = true;
           }
@@ -131,11 +165,10 @@ public class ParquetTableMetadataProviderImpl extends BaseParquetMetadataProvide
         }
       } else {
         Path p = Path.getPathWithoutSchemeAndAuthority(selectionRoot);
-        metaPath = new Path(p, Metadata.METADATA_FILENAME);
-        if (!metaContext.isMetadataCacheCorrupted() && fs.isDirectory(selectionRoot)
-            && fs.exists(metaPath)) {
+        metaPaths = populateMetaPaths(p, fs);
+        if (!metaContext.isMetadataCacheCorrupted() && fs.isDirectory(selectionRoot) && !metaPaths.isEmpty()) {
           if (parquetTableMetadata == null) {
-            parquetTableMetadata = Metadata.readBlockMeta(processUserFileSystem, metaPath, metaContext, readerConfig);
+            parquetTableMetadata = Metadata.readBlockMeta(processUserFileSystem, metaPaths, metaContext, readerConfig);
           }
           if (parquetTableMetadata != null) {
             usedMetadataCache = true;
@@ -179,17 +212,19 @@ public class ParquetTableMetadataProviderImpl extends BaseParquetMetadataProvide
     }
 
     // use the cacheFileRoot if provided (e.g after partition pruning)
-    Path metaFilePath = new Path(cacheFileRoot != null ? cacheFileRoot : selectionRoot, Metadata.METADATA_FILENAME);
-    if (!fs.exists(metaFilePath)) { // no metadata cache
+    Path path = cacheFileRoot != null ? cacheFileRoot : selectionRoot;
+    // Depending on the version of metadata this may represent more than 1 metadata file paths.
+    List<Path> metaPaths = populateMetaPaths(path, fs);
+    if (metaPaths.isEmpty()) { // no metadata cache
       if (selection.isExpandedPartial()) {
-        logger.error("'{}' metadata file does not exist, but metadata directories cache file is present", metaFilePath);
+        logger.error("'{}' metadata file/files does not exist, but metadata directories cache file is present", metaPaths.toString());
         metaContext.setMetadataCacheCorrupted(true);
       }
 
       return selection;
     }
 
-    return expandSelectionFromMetadataCache(selection, metaFilePath);
+    return expandSelectionFromMetadataCache(selection, metaPaths);
   }
 
   /**
@@ -213,19 +248,19 @@ public class ParquetTableMetadataProviderImpl extends BaseParquetMetadataProvide
    * This function also initializes a few of ParquetGroupScan's fields as appropriate.
    *
    * @param selection initial file selection
-   * @param metaFilePath metadata cache file path
+   * @param metaFilePaths metadata cache file path
    * @return file selection read from cache
    *
    * @throws org.apache.drill.common.exceptions.UserException when the updated selection is empty, this happens if the user selects an empty folder.
    */
-  private FileSelection expandSelectionFromMetadataCache(FileSelection selection, Path metaFilePath) throws IOException {
+  private FileSelection expandSelectionFromMetadataCache(FileSelection selection, List<Path> metaFilePaths) throws IOException {
     // get the metadata for the root directory by reading the metadata file
     // parquetTableMetadata contains the metadata for all files in the selection root folder, but we need to make sure
     // we only select the files that are part of selection (by setting fileSet appropriately)
 
     // get (and set internal field) the metadata for the directory by reading the metadata file
     FileSystem processUserFileSystem = ImpersonationUtil.createFileSystem(ImpersonationUtil.getProcessUserName(), fs.getConf());
-    parquetTableMetadata = Metadata.readBlockMeta(processUserFileSystem, metaFilePath, metaContext, readerConfig);
+    parquetTableMetadata = Metadata.readBlockMeta(processUserFileSystem, metaFilePaths, metaContext, readerConfig);
     if (ignoreExpandingSelection(parquetTableMetadata)) {
       return selection;
     }
@@ -264,9 +299,10 @@ public class ParquetTableMetadataProviderImpl extends BaseParquetMetadataProvide
       for (FileStatus status : fileStatuses) {
         Path currentCacheFileRoot = status.getPath();
         if (status.isDirectory()) {
-          //TODO [DRILL-4496] read the metadata cache files in parallel
-          Path metaPath = new Path(currentCacheFileRoot, Metadata.METADATA_FILENAME);
-          MetadataBase.ParquetTableMetadataBase metadata = Metadata.readBlockMeta(processUserFileSystem, metaPath, metaContext, readerConfig);
+          // TODO [DRILL-4496] read the metadata cache files in parallel
+          // Depending on the version of metadata this may represent more than 1 metadata file paths.
+          List<Path> metaPaths = populateMetaPaths(currentCacheFileRoot, fs);
+          MetadataBase.ParquetTableMetadataBase metadata = Metadata.readBlockMeta(processUserFileSystem, metaPaths, metaContext, readerConfig);
           if (ignoreExpandingSelection(metadata)) {
             return selection;
           }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
@@ -26,7 +26,8 @@ import org.apache.drill.exec.resolver.TypeCastRules;
 import org.apache.drill.exec.server.options.OptionManager;
 import org.apache.drill.exec.store.ColumnExplorer;
 import org.apache.drill.exec.store.parquet.metadata.MetadataBase;
-import org.apache.drill.exec.store.parquet.metadata.Metadata_V3;
+import org.apache.drill.exec.store.parquet.metadata.MetadataVersion;
+import org.apache.drill.exec.store.parquet.metadata.Metadata_V4;
 import org.apache.drill.metastore.BaseMetadata;
 import org.apache.drill.metastore.CollectableColumnStatisticsKind;
 import org.apache.drill.metastore.ColumnStatistics;
@@ -162,15 +163,16 @@ public class ParquetTableMetadataUtils {
   /**
    * Merges list of specified metadata into the map of {@link ColumnStatistics} with columns as keys.
    *
+   * @param <T>                 type of metadata to collect
    * @param metadataList        list of metadata to be merged
    * @param columns             set of columns whose statistics should be merged
    * @param statisticsToCollect kinds of statistics that should be collected
-   * @param <T>                 type of metadata to collect
+   * @param parquetTableMetadata ParquetTableMetadata object to fetch the non-interesting columns
    * @return list of merged metadata
    */
   @SuppressWarnings("unchecked")
   public static <T extends BaseMetadata> Map<SchemaPath, ColumnStatistics> mergeColumnsStatistics(
-      Collection<T> metadataList, Set<SchemaPath> columns, List<CollectableColumnStatisticsKind> statisticsToCollect) {
+          Collection<T> metadataList, Set<SchemaPath> columns, List<CollectableColumnStatisticsKind> statisticsToCollect, MetadataBase.ParquetTableMetadataBase parquetTableMetadata) {
     Map<SchemaPath, ColumnStatistics> columnsStatistics = new HashMap<>();
 
     for (SchemaPath column : columns) {
@@ -192,7 +194,7 @@ public class ParquetTableMetadataUtils {
       }
       columnsStatistics.put(column, new ColumnStatisticsImpl(statisticsMap, statisticsList.iterator().next().getValueComparator()));
     }
-
+    columnsStatistics.putAll(populateNonInterestingColumnsStats(columnsStatistics.keySet(), parquetTableMetadata));
     return columnsStatistics;
   }
 
@@ -201,9 +203,10 @@ public class ParquetTableMetadataUtils {
    *
    * @param rowGroups list of {@link RowGroupMetadata} to be merged
    * @param tableName name of the table
+   * @param parquetTableMetadata
    * @return {@link FileMetadata} instance
    */
-  public static FileMetadata getFileMetadata(List<RowGroupMetadata> rowGroups, String tableName) {
+  public static FileMetadata getFileMetadata(List<RowGroupMetadata> rowGroups, String tableName, MetadataBase.ParquetTableMetadataBase parquetTableMetadata) {
     if (rowGroups.isEmpty()) {
       return null;
     }
@@ -213,7 +216,7 @@ public class ParquetTableMetadataUtils {
     TupleSchema schema = rowGroups.iterator().next().getSchema();
 
     return new FileMetadata(rowGroups.iterator().next().getLocation(), schema,
-      mergeColumnsStatistics(rowGroups, rowGroups.iterator().next().getColumnsStatistics().keySet(), PARQUET_STATISTICS),
+      mergeColumnsStatistics(rowGroups, rowGroups.iterator().next().getColumnsStatistics().keySet(), PARQUET_STATISTICS, parquetTableMetadata),
       fileStatistics, tableName, -1);
   }
 
@@ -238,7 +241,7 @@ public class ParquetTableMetadataUtils {
     partStatistics.put(TableStatisticsKind.ROW_COUNT.getName(), TableStatisticsKind.ROW_COUNT.mergeStatistics(files));
 
     return new PartitionMetadata(partitionColumn, files.iterator().next().getSchema(),
-        mergeColumnsStatistics(files, columns, PARQUET_STATISTICS), partStatistics, locations, tableName, -1);
+        mergeColumnsStatistics(files, columns, PARQUET_STATISTICS, null), partStatistics, locations, tableName, -1);
   }
 
   /**
@@ -260,7 +263,7 @@ public class ParquetTableMetadataUtils {
    * @return map with converted row group metadata
    */
   @SuppressWarnings("unchecked")
-  public static Map<SchemaPath, ColumnStatistics> getRowGroupColumnStatistics(
+  private static Map<SchemaPath, ColumnStatistics> getRowGroupColumnStatistics(
       MetadataBase.ParquetTableMetadataBase tableMetadata, MetadataBase.RowGroupMetadata rowGroupMetadata) {
 
     Map<SchemaPath, ColumnStatistics> columnsStatistics = new HashMap<>();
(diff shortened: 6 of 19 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.store.parquet;
 
+import static java.lang.Thread.sleep;
```
vdiravka: No need to import static `sleep` method. There are a lot of similar usages in Drill code and class usage is better for visibility.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -91,9 +95,14 @@
 public class Metadata {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Metadata.class);
 
-  public static final String[] OLD_METADATA_FILENAMES = {".drill.parquet_metadata.v2"};
+  public static final String[] OLD_METADATA_FILENAMES = {".drill.parquet_metadata", ".drill.parquet_metadata.v2"};
   public static final String METADATA_FILENAME = ".drill.parquet_metadata";
   public static final String METADATA_DIRECTORIES_FILENAME = ".drill.parquet_metadata_directories";
+  public static final String FILE_METADATA_FILENAME = ".drill.parquet_file_metadata.v4";
+  public static final String METADATA_SUMMARY_FILENAME = ".drill.parquet_summary_metadata.v4";
+  public static final String[] CURRENT_METADATA_FILENAMES = {METADATA_SUMMARY_FILENAME, FILE_METADATA_FILENAME};
```
vdiravka: What about `METADATA_DIRECTORIES_FILENAME`? Is it current metadata fileName?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -91,9 +95,14 @@
 public class Metadata {
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(Metadata.class);
 
-  public static final String[] OLD_METADATA_FILENAMES = {".drill.parquet_metadata.v2"};
+  public static final String[] OLD_METADATA_FILENAMES = {".drill.parquet_metadata", ".drill.parquet_metadata.v2"};
   public static final String METADATA_FILENAME = ".drill.parquet_metadata";
   public static final String METADATA_DIRECTORIES_FILENAME = ".drill.parquet_metadata_directories";
+  public static final String FILE_METADATA_FILENAME = ".drill.parquet_file_metadata.v4";
```
vdiravka: Looks like `METADATA_FILENAME` should be some sort of `OLD_METADATA_FILENAME`
but `FILE_METADATA_FILENAME` -> `METADATA_FILENAME`

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -633,43 +713,120 @@ private void readBlockMeta(Path path, boolean dirsOnly, MetadataContext metaCont
...
+        module.setUseOptimizedBeanDeserializer(true);
+        mapper.registerModule(serialModule);
+        mapper.registerModule(module);
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+        InputStream is = fs.open(summaryFile);
+        Metadata_V4.MetadataSummary metadataSummary = mapper.readValue(is, Metadata_V4.MetadataSummary.class);
+        return metadataSummary;
+        }
+    } catch (IOException e) {
+      return null;
```
vdiravka: How the case with null is handled?
It is necessary to print the info from e to logs possibly

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -633,43 +713,120 @@ private void readBlockMeta(Path path, boolean dirsOnly, MetadataContext metaCont
...
+    return metadataSummary.isAllColumnsInteresting();
+  }
+
+  public static Metadata_V4.MetadataSummary getSummary(FileSystem fs, Path metadataParentDir, boolean autoRefreshTriggered, ParquetReaderConfig readerConfig) {
+    Path summaryFile = new Path(metadataParentDir, METADATA_SUMMARY_FILENAME);
+    Path metadataDirFile = new Path(metadataParentDir, METADATA_DIRECTORIES_FILENAME);
+    MetadataContext metaContext = new MetadataContext();
+    try {
+      if (!fs.exists(summaryFile)) {
+        return null;
```
vdiravka: logger

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -712,5 +869,4 @@ private boolean logAndStopTimer(boolean isModified, String directoryName,
     return isModified;
   }
 
-}
-
+}
```
vdiravka:
```suggestion
}

```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataPathUtils.java
```diff
@@ -60,6 +61,22 @@
    * @param baseDir base parent directory
    * @return list of files with absolute paths
    */
+  public static List<ParquetFileMetadata_v4> convertToFilesWithAbsolutePathsForV4(
```
vdiravka: Please modify existing method `convertToFilesWithAbsolutePaths` for using with `ParquetFileMetadata_v4` and `ParquetFileMetadata_v3` to avoid duplicating the code.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V1.java
```diff
@@ -90,6 +90,16 @@ public Integer getDefinitionLevel(String[] columnName) {
       return null;
     }
 
+    @Override
```
vdiravka: `@JsonIgnore` ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -0,0 +1,648 @@
...
+import static org.apache.drill.exec.store.parquet.metadata.MetadataBase.ParquetFileMetadata;
+import static org.apache.drill.exec.store.parquet.metadata.MetadataBase.ParquetTableMetadataBase;
+import static org.apache.drill.exec.store.parquet.metadata.MetadataBase.RowGroupMetadata;
+import static org.apache.drill.exec.store.parquet.metadata.MetadataVersion.Constants.V4;
+import org.apache.hadoop.fs.Path;
+import org.apache.parquet.io.api.Binary;
+import org.apache.parquet.schema.OriginalType;
+import org.apache.parquet.schema.PrimitiveType;
+
+public class Metadata_V4 {
```
vdiravka: Please check all methods whether they are serializable or not.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -956,4 +957,302 @@ public void testRefreshNone() throws Exception {
...
+    String tableName = "orders_ctas_refresh_not_null";
+    test("use dfs");
+    test("create table `%s/t4` as select * from cp.`tpch/orders.parquet`", tableName);
+    test("refresh table metadata COLUMNS (o_orderdate) %s", tableName);
+    String query = String.format("Select count(*) as cnt from `%s` where o_orderpriority is not null", tableName);
+
+    checkForMetadataFile(tableName);
+
+    testBuilder()
+            .sqlQuery(query)
```
vdiravka: 4 spaces indent is used in Drill code for builder's chain.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -956,4 +957,302 @@ public void testRefreshNone() throws Exception {
...
+    createMetadataDir(tableName);
+    testBuilder()
+            .sqlQuery(rowCountQuery)
+            .unOrdered()
+            .baselineColumns("rowCount")
+            .baselineValues(50L)
+            .go();
+  }
+
+  @Ignore
```
vdiravka: Why it is disabled?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/RefreshMetadataHandler.java
```diff
@@ -161,7 +161,7 @@ public PhysicalPlan getPlan(SqlNode sqlNode) throws ForemanSetupException {
    */
   private SqlNodeList getColumnList(final SqlRefreshMetadata sqlrefreshMetadata) {
     SqlNodeList columnList = sqlrefreshMetadata.getFieldList();
-    if (columnList == null || !SqlNodeList.isEmptyList(columnList)) {
```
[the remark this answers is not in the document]
dvjyothsna: Removed the extra check.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -387,6 +389,14 @@ private static boolean allowBinaryMetadata(String drillVersion, ParquetReaderCon
           names.add(Arrays.asList(columnTypeMetadata.name));
         }
       }
+    } else if (parquetTableMetadata instanceof ParquetTableMetadata_v4) {
```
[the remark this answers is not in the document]
dvjyothsna: Changed to get columnTypeInfo from MetadataBase

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -0,0 +1,648 @@
...
+    public void setTotalRowCount(Long totalRowCount) {
+      this.totalRowCount = totalRowCount;
+    }
+
+    public Long getTotalRowCount() {
+      return this.totalRowCount;
+    }
+  }
+
+  //  @JsonTypeName (V4)
```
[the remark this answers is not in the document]
dvjyothsna: Added JavaDoc

exec/java-exec/src/test/java/org/apache/drill/PlanTestBase.java
```diff
@@ -465,4 +469,24 @@ private static String getPrefixJoinOrderFromPlan(String plan, String joinKeyWord
...
+    try {
+      tmpDir = dirTestWatcher.getRootDir().getCanonicalPath();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+    File metadataDir = dirTestWatcher.makeRootSubDir(Paths.get(tmpDir+"/"+table+"/metadataDir"));
+    File metaFile, newFile;
+    metaFile = table.startsWith(tmpDir) ? FileUtils.getFile(table, Metadata.METADATA_SUMMARY_FILENAME)
+            : FileUtils.getFile(tmpDir, table, Metadata.METADATA_SUMMARY_FILENAME);
+    newFile = new File(tmpDir+"/"+table+"/summary_meta.json");
```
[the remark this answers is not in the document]
dvjyothsna: Changed

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
dvjyothsna: Changed to use Thread.sleep()

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: METADATA_DIRECTORIES_FILENAME is same across all the versions. CURRENT_METADATA_FILENAMES contains the newer version cache files. In this case summary and file metadata cache file names

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Added loggers.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Added logging.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Changed

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
dvjyothsna: Changed.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
dvjyothsna: Enabled the test case

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataVersion.java
```diff
@@ -145,6 +145,11 @@ public int compareTo(MetadataVersion o) {
      */
     public static final String V3_3 = "3.3";
 
+  /**
+   *  Version 4: Split the metadata cache file into summary and file metadata
+   */
+  public static final String V4 = "4";
```
amansinha100: It occurs to me that for consistency we should follow a convention of `MajorVersion.MinorVersion`  i.e 4.0 instead of just 4.  In Version 3 we started adding the minor versions later but now that we are following that model, it would make sense to have a 2 digit scheme.  Let's discuss..

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -0,0 +1,672 @@
...
+    @Override
+    public ParquetTableMetadataBase clone() {
+      return new ParquetTableMetadata_v4(metadataSummary, fileMetadata);
+    }
+
+    @Override
+    public String getDrillVersion() {
+      return metadataSummary.drillVersion;
+    }
+
```
amansinha100: Code style..please remove extra blank lines throughout this file

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -230,68 +254,102 @@ private static boolean ignoreReadingMetadata(MetadataContext metaContext, Path p
...
-    ParquetTableMetadata_v3 parquetTableMetadata = new ParquetTableMetadata_v3(SUPPORTED_VERSIONS.last().toString(),
-                                                                                DrillVersionInfo.getVersion());
+    Metadata_V4.MetadataSummary metadataSummary = new Metadata_V4.MetadataSummary(SUPPORTED_VERSIONS.last().toString(), DrillVersionInfo.getVersion());
+    ParquetTableMetadata_v4 parquetTableMetadata = new ParquetTableMetadata_v4(metadataSummary);
     if (childFiles.size() > 0) {
-      List<ParquetFileMetadata_v3 > childFilesMetadata = getParquetFileMetadata_v3(parquetTableMetadata, childFiles, allColumns, columnSet);
-      metaDataList.addAll(childFilesMetadata);
-      // Note that we do not need to merge the columnInfo at this point. The columnInfo is already added
-      // to the parquetTableMetadata.
+      List<ParquetFileAndRowCountMetadata> ChildFileAndRowCountMetadata = getParquetFileMetadata_v4(parquetTableMetadata, childFiles, allColumnsInteresting, columnSet);
```
amansinha100: By convention, variable names start with lower case.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -230,68 +254,102 @@ private static boolean ignoreReadingMetadata(MetadataContext metaContext, Path p
...
+      for (ParquetFileAndRowCountMetadata parquetFileAndRowCountMetadata : ChildFileAndRowCountMetadata) {
+        metaDataList.add(parquetFileAndRowCountMetadata.getFileMetadata());
+        dirTotalRowCount = dirTotalRowCount + parquetFileAndRowCountMetadata.getFileRowCount();
+        Map<ColumnTypeMetadata_v4.Key, Long> totalNullCountMap = parquetFileAndRowCountMetadata.getTotalNullCountMap();
+        if (columnTypeInfoSet.isEmpty()) {
+          columnTypeInfoSet.putAll(parquetTableMetadata.getColumnTypeInfoMap());
```
amansinha100: Shouldn't this set be populated before entering the loop ? There does not seem to be a dependency on anything computed in the loop.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -230,68 +254,102 @@ private static boolean ignoreReadingMetadata(MetadataContext metaContext, Path p
...
+        }
+        for (ColumnTypeMetadata_v4.Key columnName: totalNullCountMap.keySet()) {
```
amansinha100: 'columnName' indicates it is a string but it is actually a struct.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -230,68 +254,102 @@ private static boolean ignoreReadingMetadata(MetadataContext metaContext, Path p
...
+          ColumnTypeMetadata_v4 columnTypeMetadata_v4 = columnTypeInfoSet.get(columnName);
+          if ( columnTypeMetadata_v4.totalNullCount < 0 || totalNullCountMap.get(columnName) < 0) {
+            columnTypeMetadata_v4.totalNullCount = NULL_COUNT_NOT_EXISTS;
+          } else {
+            columnTypeMetadata_v4.totalNullCount = columnTypeMetadata_v4.totalNullCount + totalNullCountMap.get(columnName);
```
amansinha100: Suggest using +=

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -230,68 +254,102 @@ private static boolean ignoreReadingMetadata(MetadataContext metaContext, Path p
...
+      for (ParquetFileAndRowCountMetadata parquetFileAndRowCountMetadata : ChildFileAndRowCountMetadata) {
+        metaDataList.add(parquetFileAndRowCountMetadata.getFileMetadata());
+        dirTotalRowCount = dirTotalRowCount + parquetFileAndRowCountMetadata.getFileRowCount();
+        Map<ColumnTypeMetadata_v4.Key, Long> totalNullCountMap = parquetFileAndRowCountMetadata.getTotalNullCountMap();
+        if (columnTypeInfoSet.isEmpty()) {
+          columnTypeInfoSet.putAll(parquetTableMetadata.getColumnTypeInfoMap());
+        }
+        for (ColumnTypeMetadata_v4.Key columnName: totalNullCountMap.keySet()) {
+          ColumnTypeMetadata_v4 columnTypeMetadata_v4 = columnTypeInfoSet.get(columnName);
+          if ( columnTypeMetadata_v4.totalNullCount < 0 || totalNullCountMap.get(columnName) < 0) {
```
amansinha100: If the first part of the || is true, it could mean that this is the first time we are initializing the `totalNullCount` .. so is that even a valid check ?  If the second part was >= 0,  we should not be setting the `NULL_COUNT_NOT_EXISTS`.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -230,68 +254,102 @@ private static boolean ignoreReadingMetadata(MetadataContext metaContext, Path p
...
-        columnTypeInfoSet.putAll(subTableMetadata.columnTypeInfo);
+        if (columnTypeInfoSet.isEmpty()) {
+          columnTypeInfoSet.putAll(subTableColumnTypeInfo);
+        } else {
+          for (ColumnTypeMetadata_v4.Key key : subTableColumnTypeInfo.keySet()) {
+            ColumnTypeMetadata_v4 columnTypeMetadata_v4 = columnTypeInfoSet.get(key);
+            if (columnTypeMetadata_v4 == null) {
+              columnTypeMetadata_v4 = subTableColumnTypeInfo.get(key);
+            } else {
+              if (subTableColumnTypeInfo.get(key).totalNullCount < 0 || columnTypeMetadata_v4.totalNullCount < 0) {
```
amansinha100: If at least one of these conditions is true, then shouldn't we *not* be setting the `NULL_COUNT_NOT_EXISTS` ?  A similar comment is further below.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -149,20 +157,25 @@ public static ParquetTableMetadata_v3 getParquetTableMetadata(Map<FileStatus, Fi
...
-      return null;
-    }
     Metadata metadata = new Metadata(readerConfig);
-    metadata.readBlockMeta(path, false, metaContext, fs);
+    if (paths.isEmpty()) {
+      metaContext.setMetadataCacheCorrupted(true);
+    }
+    for (Path path: paths) {
+      if (ignoreReadingMetadata(metaContext, path)) {
+        return null;
```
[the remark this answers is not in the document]
dvjyothsna: Paths contain paths for Summary and File metadata file. If we pull out the ignoreReadingMetadata out of the for loop, we won't be able to ignore reading the Paths[1] file if the Paths[0] one fails. We would like to ignore reading of the metadata if the reading of the Paths[0] file fails.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -0,0 +1,648 @@
...
+        }
+      }
+    }
+  }
+
+
+  /**
+   * A struct that contains the metadata for a column in a parquet file
+   */
+  public static class ColumnMetadata_v4 extends ColumnMetadata {
```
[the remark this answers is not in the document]
dvjyothsna: Changed ColumnMetadata_v4 to extend ColumnMetadata_v3.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataPathUtils.java
dvjyothsna: Changed the method.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
```diff
@@ -17,8 +17,11 @@
  */
 package org.apache.drill.exec.store.parquet;
 
+import java.util.Comparator;
 import org.apache.drill.exec.physical.base.GroupScan;
 import org.apache.drill.exec.physical.base.ParquetMetadataProvider;
+import static org.apache.drill.exec.store.parquet.ParquetTableMetadataUtils.getComparator;
```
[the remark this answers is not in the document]
dvjyothsna: Imported ParquetTableMetadataUtils and called comparator

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
```diff
@@ -167,6 +172,7 @@ public TableMetadata getTableMetadata() {
           columnsStatistics.put(partitionColumn, new ColumnStatisticsImpl(stats, ParquetTableMetadataUtils.getNaturalNullsFirstComparator()));
         }
       }
+      columnsStatistics = populateColumnStats(columnsStatistics);
```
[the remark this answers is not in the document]
dvjyothsna: Added javadoc for populateColumnStats().

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
```diff
@@ -244,8 +250,8 @@ public TableMetadata getTableMetadata() {
             statistics.put(TableStatisticsKind.ROW_COUNT.getName(), GroupScan.NO_COLUMN_STATS);
             columnsStatistics.put(partitionColumn,
                 new ColumnStatisticsImpl<>(statistics,
-                    ParquetTableMetadataUtils.getComparator(getParquetGroupScanStatistics().getTypeForColumn(partitionColumn).getMinorType())));
-
+                    getComparator(getParquetGroupScanStatistics().getTypeForColumn(partitionColumn).getMinorType())));
+//            columnsStatistics = populateColumnStats(columnsStatistics);
```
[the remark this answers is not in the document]
dvjyothsna: Removed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetFormatPlugin.java
```diff
@@ -298,11 +298,28 @@ public DrillTable isReadable(DrillFileSystem fs, FileSelection selection,
     }
 
     private Path getMetadataPath(FileStatus dir) {
-      return new Path(dir.getPath(), Metadata.METADATA_FILENAME);
+      return new Path(dir.getPath(), Metadata.OLD_METADATA_FILENAME);
+    }
+
+    /**
+     * Check if the metadata version 4 files exist
```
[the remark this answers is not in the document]
dvjyothsna: Changed the javadoc

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetFormatPlugin.java
```diff
@@ -298,11 +298,28 @@ public DrillTable isReadable(DrillFileSystem fs, FileSelection selection,
...
+        if (!fs.exists(path)) {
+          return false;
+        }
+      }
+      return true;
     }
 
     private boolean metaDataFileExists(FileSystem fs, FileStatus dir) throws IOException {
-      return fs.exists(getMetadataPath(dir));
+      return fs.exists(getMetadataPath(dir)) || currentMetadataFileExists(dir, fs);
```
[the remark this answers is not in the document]
dvjyothsna: Extracted both methods into one.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
```diff
@@ -109,19 +109,59 @@ public Path getSelectionRoot() {
...
+      }
+    }
+    return true;
+  }
+
   @Override
   protected void initInternal() throws IOException {
     try (FileSystem processUserFileSystem = ImpersonationUtil.createFileSystem(ImpersonationUtil.getProcessUserName(), fs.getConf())) {
-      Path metaPath = null;
+      List<Path> metaPaths = new ArrayList<>();
```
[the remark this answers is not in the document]
dvjyothsna: Added a comment.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
```diff
@@ -109,19 +109,59 @@ public Path getSelectionRoot() {
...
+   * @return list of cache files found in the given directory path
+   */
+  public List<Path> populateMetaPaths(Path p, DrillFileSystem fs) throws IOException {
+    List<Path> metaFilepaths = new ArrayList<>();
+    for (String filename : Metadata.CURRENT_METADATA_FILENAMES) {
+      metaFilepaths.add(new Path(p, filename));
+    }
+
+    //Read the older version of metadata file if the current version of metadata cache files donot exist.
+    for (String filename : Metadata.OLD_METADATA_FILENAMES) {
```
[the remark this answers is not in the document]
dvjyothsna: Changed

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
```diff
@@ -109,19 +109,59 @@ public Path getSelectionRoot() {
...
+      if (fileExists(fs, metaFilepaths)) {
+        return metaFilepaths;
+      }
+      metaFilepaths.clear();
+      metaFilepaths.add(new Path(p, filename));
+    }
+
```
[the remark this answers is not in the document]
dvjyothsna: Removed the extra lines

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
```diff
@@ -282,7 +286,24 @@ public static PartitionMetadata getPartitionMetadata(SchemaPath partitionColumn,
       statistics.put(ColumnStatisticsKind.NULLS_COUNT.getName(), nulls);
       columnsStatistics.put(colPath, new ColumnStatisticsImpl(statistics, comparator));
     }
+    columnsStatistics = populateColumnStats(columnsStatistics, tableMetadata);
+    return columnsStatistics;
+  }
 
+  private static Map<SchemaPath, ColumnStatistics> populateColumnStats(Map<SchemaPath, ColumnStatistics> columnsStatistics, MetadataBase.ParquetTableMetadataBase parquetTableMetadata) {
```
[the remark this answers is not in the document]
dvjyothsna: Extracted both methods into one and added javadoc.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetMetadataCache.java
```diff
@@ -182,31 +182,31 @@ public void testFix4449() throws Exception {
...
     testBuilder()
-      .sqlQuery("SELECT COUNT(*) cnt FROM (" +
-        "SELECT l_orderkey FROM dfs.tmp.`4449` WHERE l_discount < 0.05" +
-        " UNION ALL" +
-        " SELECT l_orderkey FROM dfs.tmp.`4449` WHERE l_discount > 0.02)")
-      .unOrdered()
-      .baselineColumns("cnt")
-      .baselineValues(71159L)
-      .go();
+            .sqlQuery("SELECT COUNT(*) cnt FROM (" +
```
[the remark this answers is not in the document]
dvjyothsna: Removed the extraneous white spaces.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataVersion.java
dvjyothsna: Changed to 4.0

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Changed

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
dvjyothsna: Removed blank lines.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Moved the if case from the for loop.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Changed

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Added a comment to explain this.

dvjyothsna:
@vdiravka @amansinha100 I have addressed the review comments. Can you please review it once again.

vvysotskyi:
@dvjyothsna, have you addressed my comment about handling non-existing columns?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -0,0 +1,540 @@
...
+      }
+
+      public static class DeSerializer extends KeyDeserializer {
+
+        public DeSerializer() {
+        }
+
+        @Override
+        public Object deserializeKey(String key, com.fasterxml.jackson.databind.DeserializationContext ctxt) {
+          // key string should contain '`' char if the field was serialized as SchemaPath object
```
vvysotskyi: This code was added for v3 because in one of the versions was changed the format of columns serialization. For now, it is enough to specify
```
.addDeserializer(SchemaPath.class, new SchemaPath.De());
```
instead of this class.

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetFilterPushDown.java
```diff
@@ -138,18 +138,25 @@ public void testIntPredicateWithEval() throws Exception {
     testParquetRowGroupFilterEval(footer, "intCol = 50 or intCol = 160", RowsMatch.SOME);
 
     //"nonExistCol" does not exist in the table. "AND" with a filter on exist column
+
+    //Ignored till run-time row group pruning is added.
```
vvysotskyi: See my comment about handling non-existent columns.

arina-ielchiieva: @dvjyothsna / @amansinha100 not sure that we should allow ignoring tests. Does it mean that we are breaking existing functionality?

amansinha100: I believe these tests should be enabled now .. it might be an omission.  there are couple of new unit tests for non-existing columns in TestParquetMetadataCache.

dvjyothsna: I'm re-enabling these tests back. All the handling of non-existent columns works fine now.

dvjyothsna:
@vvysotskyi Yes, I have addressed the non-existent columns case and added few unit tests.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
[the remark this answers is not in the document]
amansinha100: Yeah, looks like an omission ...

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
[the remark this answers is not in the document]
amansinha100: I see that `ignoreReadingMetadata()` method does not actually examine the `path`.  It just checks if the metadata cache is corrupted.  From a logical standpoint, I don't see why one would ignore the Summary file but not the File metadata (or vice-versa).   From a performance reason also, suppose paths[0] was not ignored (hypothetically), and we execute the next statement `readBlockMeta()` and read the entire File metadata and subsequently find out that paths[1] is ignored, it would be wasted effort.  What do you think ?

amansinha100:
@dvjyothsna could you see if @vvysotskyi 's comments have also been addressed and just mark 'done' for those.  Thanks for you patience in making these changes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
amansinha100: @dvjyothsna to follow up on this..it is not sufficient to do the logging.  One option is to return the `IOException` and let the caller handle it.  Another is to do something similar to `readBlockMeta` in the current code..if encounters an IOException, it marks the cache as corrupted in the metadata context.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -0,0 +1,648 @@
...
+        if (value.nulls != null) {
+          jgen.writeObjectField("nulls", value.nulls);
+        }
+        jgen.writeEndObject();
+      }
+    }
+  }
+
+  @JsonTypeName(V4)
+  public static class MetadataSummary {
```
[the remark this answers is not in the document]
amansinha100: @vdiravka the Appendix section of the design doc has an example.  @dvjyothsna can you make sure that is updated to reflect the as-implemented version ?

exec/java-exec/src/test/java/org/apache/drill/PlanTestBase.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.drill;
 
+import java.nio.file.Paths;
```
amansinha100: We use the Hadoop library's  `Path` class everywhere .. is that not sufficient for your use ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
```diff
@@ -165,6 +165,7 @@ public TableMetadata getTableMetadata() {
               TableStatisticsKind.ROW_COUNT.getName(), columnValueCount,
               ColumnStatisticsKind.NULLS_COUNT.getName(), getParquetGroupScanStatistics().getRowCount() - columnValueCount);
           columnsStatistics.put(partitionColumn, new ColumnStatisticsImpl(stats, ParquetTableMetadataUtils.getNaturalNullsFirstComparator()));
+          columnsStatistics = ParquetTableMetadataUtils.populateNonInterestingColumnsStats(columnsStatistics, parquetTableMetadata);
```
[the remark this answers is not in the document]
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
```diff
@@ -282,7 +286,32 @@ public static PartitionMetadata getPartitionMetadata(SchemaPath partitionColumn,
...
+    return columnsStatistics;
+  }
 
+  /**
+   * Populates the non-interesting column's statistics
+   * @param columnsStatistics
+   * @param parquetTableMetadata
+   * @return returns new column statistics map
+   */
+  public static Map<SchemaPath, ColumnStatistics> populateNonInterestingColumnsStats(
```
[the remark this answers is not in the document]
dvjyothsna: Changed to pass schemapaths to the method.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V3.java
```diff
@@ -163,6 +174,15 @@ public String getDrillVersion() {
...
     }
 
+    @JsonIgnore
+    public ConcurrentHashMap<Metadata_V3.ColumnTypeMetadata_v3.Key, Metadata_V3.ColumnTypeMetadata_v3> getColumnTypeInfoMap() {
+      return this.columnTypeInfo;
+    }
+
+    @Override
+    public List<? extends MetadataBase.ColumnTypeMetadata> getColumnTypeInfoList() {
+      return new ArrayList(this.columnTypeInfo.values());
```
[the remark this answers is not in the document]
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
```diff
@@ -109,19 +108,57 @@ public Path getSelectionRoot() {
     return selectionRoot;
   }
 
+  /**
+   * Returns list of metadata cache files
+   * @param p directory path of the cache file
+   * @param fs
```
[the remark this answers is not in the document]
dvjyothsna: Added description.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
```diff
@@ -109,19 +108,57 @@ public Path getSelectionRoot() {
...
+   * @return list of cache files found in the given directory path
+   */
+  public List<Path> populateMetaPaths(Path p, DrillFileSystem fs) throws IOException {
+    List<Path> metaFilepaths = new ArrayList<>();
+    for (String filename : Metadata.CURRENT_METADATA_FILENAMES) {
+      metaFilepaths.add(new Path(p, filename));
+    }
+    for (String filename : Metadata.OLD_METADATA_FILENAMES) {
+      //Read the older version of metadata file if the current version of metadata cache files donot exist.
```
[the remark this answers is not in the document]
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
```diff
@@ -109,19 +108,57 @@ public Path getSelectionRoot() {
...
+    }
+    if (fileExists(fs, metaFilepaths)) {
+      return metaFilepaths;
+    }
+    return new ArrayList<>();
+  }
+
+  public boolean fileExists(DrillFileSystem fs, List<Path> paths) throws IOException {
+    if (paths.isEmpty()) {
+      return false;
```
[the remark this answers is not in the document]
dvjyothsna: Changed to return true for this case.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
```diff
@@ -162,15 +163,16 @@ public static RowGroupMetadata getRowGroupMetadata(MetadataBase.ParquetTableMeta
   /**
    * Merges list of specified metadata into the map of {@link ColumnStatistics} with columns as keys.
    *
+   * @param <T>                 type of metadata to collect
    * @param metadataList        list of metadata to be merged
    * @param columns             set of columns whose statistics should be merged
    * @param statisticsToCollect kinds of statistics that should be collected
-   * @param <T>                 type of metadata to collect
+   * @param parquetTableMetadata
```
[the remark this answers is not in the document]
dvjyothsna: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
dvjyothsna: Looks like just adding SchemaPath deserializer is not working

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetFilterPushDown.java
```diff
@@ -138,18 +138,22 @@ public void testIntPredicateWithEval() throws Exception {
...
+
     testParquetRowGroupFilterEval(footer, "intCol > 100 and nonExistCol = 100", RowsMatch.NONE);
     testParquetRowGroupFilterEval(footer, "intCol > 50 and nonExistCol = 100", RowsMatch.NONE); // since nonExistCol = 100 -> Unknown -> could drop.
     testParquetRowGroupFilterEval(footer, "nonExistCol = 100 and intCol > 50", RowsMatch.NONE); // since nonExistCol = 100 -> Unknown -> could drop.
     testParquetRowGroupFilterEval(footer, "intCol > 100 and nonExistCol < 'abc'", RowsMatch.NONE);
     testParquetRowGroupFilterEval(footer, "nonExistCol < 'abc' and intCol > 100", RowsMatch.NONE); // nonExistCol < 'abc' hit NumberException and is ignored, but intCol >100 will
+
     // say "drop".
-    testParquetRowGroupFilterEval(footer, "intCol > 50 and nonExistCol < 'abc'", RowsMatch.SOME); // because nonExistCol < 'abc' hit NumberException and
+//    testParquetRowGroupFilterEval(footer, "intCol > 50 and nonExistCol < 'abc'", RowsMatch.SOME); // because nonExistCol < 'abc' hit NumberException and
```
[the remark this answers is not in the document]
dvjyothsna: Unignored this test

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V3.java
```diff
@@ -149,6 +150,16 @@ public Integer getDefinitionLevel(String[] columnName) {
       return getColumnTypeInfo(columnName).definitionLevel;
     }
 
+    @JsonIgnore @Override
```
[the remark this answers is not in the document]
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataBase.java
```diff
@@ -53,7 +54,9 @@
       @JsonSubTypes.Type(value = Metadata_V3.ParquetTableMetadata_v3.class, name = V3),
       @JsonSubTypes.Type(value = Metadata_V3.ParquetTableMetadata_v3.class, name = V3_1),
       @JsonSubTypes.Type(value = Metadata_V3.ParquetTableMetadata_v3.class, name = V3_2),
-      @JsonSubTypes.Type(value = Metadata_V3.ParquetTableMetadata_v3.class, name = V3_3)
+      @JsonSubTypes.Type(value = Metadata_V3.ParquetTableMetadata_v3.class, name = V3_3),
+      @JsonSubTypes.Type(value = Metadata_V3.ParquetTableMetadata_v3.class, name = V4),
```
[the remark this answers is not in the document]
dvjyothsna: Changed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataVersion.java
```diff
@@ -145,6 +145,11 @@ public int compareTo(MetadataVersion o) {
      */
     public static final String V3_3 = "3.3";
 
+  /**
+   *  Version 4: Split the metadata cache file into summary and file metadata
```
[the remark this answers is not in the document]
dvjyothsna: Changed

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V2.java
```diff
@@ -128,6 +129,16 @@ public Integer getDefinitionLevel(String[] columnName) {
       return null;
     }
 
+    @JsonIgnore @Override
```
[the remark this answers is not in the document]
dvjyothsna: Changed.

exec/java-exec/src/test/java/org/apache/drill/PlanTestBase.java
dvjyothsna: Looks like we can't use Hadoop library path here and moreover this test imports java.nio.file.Files and there are a lot of tests that import java.nio.file.Paths.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
dvjyothsna: We will not be serializing the ParquetTableMetadata_v4 from V4 since we always serialize the metadatasummary and fileMetadata objects which is why we need not add JsonIgnore annotations for the methods in ParquetTableMetadata_v4 class.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Addressed this based on autoRefreshTriggered flag

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
```diff
@@ -109,19 +108,57 @@ public Path getSelectionRoot() {
...
+    if (fileExists(fs, metaFilepaths)) {
+      return metaFilepaths;
+    }
+    return new ArrayList<>();
+  }
+
+  public boolean fileExists(DrillFileSystem fs, List<Path> paths) throws IOException {
+    if (paths.isEmpty()) {
+      return true;
+    }
```
Ben-Zvi: very minor - if "paths" is empty, then the following for-loop would not iterate and "true" would be returned anyway. So this "if" statement is redundant.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -633,43 +716,169 @@ private void readBlockMeta(Path path, boolean dirsOnly, MetadataContext metaCont
...
+   * @param readerConfig
+   * @return returns metadata summary
+   */
+  public static Metadata_V4.MetadataSummary getSummary(FileSystem fs, Path metadataParentDir, boolean autoRefreshTriggered, ParquetReaderConfig readerConfig) {
+    Path summaryFile = getSummaryFileName(metadataParentDir);
+    Path metadataDirFile = getDirFileName(metadataParentDir);
+    MetadataContext metaContext = new MetadataContext();
+    try {
+      // If autoRefresh is not triggered and none of the metadata files exist
+      if (!autoRefreshTriggered && metadataExists(fs, metadataParentDir)) {
```
amansinha100: This condition check does not look right..shouldn't it check metadataDirFile ? and should it be based on autoRefreshTriggered being true ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
dvjyothsna: Changed

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: So this is the case where getSummary is called to get the row count and null count while applying convertcounttodirectscan. This check looks if any of the metadata cache files exist at all in the directory. If none of the cache files are found implies that metadata doesn't exist for that directory so null is returned in that case. The autorefreshtriggered is set to true only while getting interesting columns from the previous summary file.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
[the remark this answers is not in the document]
dvjyothsna: Lets take this scenario where the summary file is corrupted but the file metadata file is intact. If we pull out the ignoreReadingMetadata (), we do this check of ignoring only before reading the summary file. But after reading the corrupted summary file, metadatacachecorrupted will be set to true. Reading of filemetadata will not be skipped if we don't check metadatacachecorrupted status. Regarding the performance  if the Paths[1] is corrupt, we will get to know that it is corrupt only in the readBlockMeta(). So even pulling out ignoreReadingMetadata() doesn't help this case. And always Paths[0] has summary and since summary is not very big it wouldn't impact the performance a lot.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataProviderImpl.java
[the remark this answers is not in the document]
dvjyothsna: I have moved this comment just before the if statement.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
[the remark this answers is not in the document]
amansinha100: Pls add a comment indicating that the ignore flag is set as part of `readBlockMeta()` which is called later in this loop, so each iteration of the loop needs to check if a previous call to readBlockMeta found a corrupted metadata.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
amansinha100: hmm..so shouldn't this be `! metadataExists()` ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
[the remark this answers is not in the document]
dvjyothsna: Added this comment.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
dvjyothsna: Yes, Changed it.

dvjyothsna:
@amansinha100 I have addressed the review comments. Can you please take a look at it.

dvjyothsna:
Thank you @amansinha100. Squashed the commits into one commit.

PR #1724, opened 2019-04-01
DRILL-7147: Source order of drill-env.sh and distrib-env.sh should be swapped

Agirish:
Drill environment properties in drill-env.sh should override corresponding properties set in distrib-env.sh


Agirish:
+ @paul-rogers (not able to tag you as a reviewer).

paul-rogers:
Please see note in the JIRA ticket. Current behavior is correct. I suspect that the issue is with one specific variable. Which one cause the issue that this is trying to fix?

I believe that there are DoY unit tests that attempt to verify correct behavior (I remember writing them.) This fix would break those tests.

Agirish:
@kkhatua , I had the same reaction at first and hence the PR :) But Paul pointed out that the current behavior is right.
@paul-rogers agree with your explanation! Thanks for taking a look. I'll close this PR.

PR #1725, opened 2019-04-01
DRILL-7146: Query failing with NPE when ZK queue is enabled.

HanumathRao:
This PR fixes the issue to not look for physical operator in memory adjusted operators of the QueueQueryParallelizer when planHasMemory is set to true.

@sohami  can you please review this fix.


exec/java-exec/src/test/java/org/apache/drill/exec/planner/rm/TestMemoryCalculator.java
```diff
@@ -59,6 +59,7 @@
 
   private static final long DEFAULT_SLICE_TARGET = 100000L;
   private static final long DEFAULT_BATCH_SIZE = 16*1024*1024;
+  private static final String ENABLE_QUEUE = "drill.exec.queue.embedded.enable";
```
sohami: Minor comment. You can use: `EmbeddedQueryQueue.ENABLED` instead of again defining it here.

HanumathRao: I have updated the test case.

PR #1726, opened 2019-04-02
DRILL-7143: Support default value for empty columns

paul-rogers:
Modifies the prior work to add default values for columns. The prior work added defaults
when the entire column is missing from a reader (the old Nullable Int column). The Row
Set mechanism now will also "fill empty" slots with the default value.

Added default support for the column writers. The writers automatically obtain the
default value from the column schema. The default can also be set explicitly on
the column writer.

Updated the null column mechanism to use this feature rather than the ad-hoc
implemention in the prior commit.

Semantics changed a bit. Only Required columns take a default. The default value
is ignored or nullable columns since nullable columns already have a file default: NULL.

Updated the CSV-with-schema tests to illustrate the new behavior.

### DIFF
--- a/common/src/main/java/org/apache/drill/common/types/BooleanType.java
+++ b/common/src/main/java/org/apache/drill/common/types/BooleanType.java
@@ -15,13 +15,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.drill.exec.expr;
+package org.apache.drill.common.types;
+
+import java.util.Map;
 
 import org.apache.drill.common.map.CaseInsensitiveMap;
-
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
 
 /**
  * Enum that contains two boolean types: TRUE and FALSE.
@@ -30,30 +28,26 @@ import java.util.Map;
  * {@link <a href="https://www.postgresql.org/docs/9.6/static/datatype-boolean.html">Postgre Documentation</a>}
  */
 public enum BooleanType {
-  TRUE(1, Arrays.asList("t", "true", "y", "yes", "on", "1")),
-  FALSE(0, Arrays.asList("f", "false", "n", "no", "off", "0"));
+  TRUE(1, new String [] {"true", "1", "t", "y", "yes", "on",}),
+  FALSE(0, new String [] {"false", "0", "f", "n", "no", "off"});
 
   private final int numericValue;
-  private final List<String> literals;
+  private final String[] literals;
 
-  BooleanType(int numericValue, List<String> literals) {
+  BooleanType(int numericValue, String[] literals) {
     this.numericValue = numericValue;
     this.literals = literals;
   }
 
-  public int getNumericValue() {
-    return numericValue;
-  }
+  public int getNumericValue() { return numericValue; }
 
-  public List<String> getLiterals() {
-    return literals;
-  }
+  public String[] getLiterals() { return literals; }
 
   /** Contains all literals that are allowed to represent boolean type. */
   private static final Map<String, BooleanType> allLiterals = CaseInsensitiveMap.newHashMap();
   static {
-    for (BooleanType booleanType : BooleanType.values()) {
-      for (String literal : booleanType.getLiterals()) {
+    for (final BooleanType booleanType : BooleanType.values()) {
+      for (final String literal : booleanType.getLiterals()) {
         allLiterals.put(literal, booleanType);
       }
     }
@@ -76,4 +70,23 @@ public enum BooleanType {
     return booleanType;
   }
 
+  /**
+   * Runtime form of Boolean conversion: allows any of the valid "true" values;
+   * assumes all other values are false. Does case-insensitive comparisons.
+   * If the string must be trimmed, the caller should do it.
+   *
+   * @param value non-null string value
+   * @return true (if one of the TRUE literals), else false
+   */
+
+  public static boolean fromString(final String value) {
+    // Optimized for runtime speed
+    final String lower = value.toLowerCase();
+    for (int i = 0; i < TRUE.literals.length; i++) {
+      if (TRUE.literals[i].equals(lower)) {
+        return true;
+      }
+    }
+    return false;
+  }
 }
--- a/common/src/main/java/org/apache/drill/common/types/Types.java
+++ b/common/src/main/java/org/apache/drill/common/types/Types.java
@@ -30,7 +30,6 @@ import org.apache.drill.common.types.TypeProtos.MinorType;
 
 import com.google.protobuf.TextFormat;
 
-@SuppressWarnings("WeakerAccess")
 public class Types {
 
   public static final int MAX_VARCHAR_LENGTH = 65535;
@@ -64,8 +63,11 @@ public class Types {
     if (type.getMode() == REPEATED) {
       return false;
     }
+    return isNumericType(type.getMinorType());
+  }
 
-    switch(type.getMinorType()) {
+  public static boolean isNumericType(final MinorType type) {
+    switch (type) {
     case BIGINT:
     case VARDECIMAL:
     case DECIMAL38SPARSE:
@@ -463,22 +465,28 @@ public class Types {
     default:
       return true;
     }
-
   }
 
   public static boolean isFixedWidthType(final MajorType type) {
-    switch(type.getMinorType()) {
+    return isFixedWidthType(type.getMinorType());
+  }
+
+  public static boolean isFixedWidthType(final MinorType type) {
+    return ! isVarWidthType(type);
+  }
+
+  public static boolean isVarWidthType(final MinorType type) {
+    switch (type) {
     case VARBINARY:
     case VAR16CHAR:
     case VARCHAR:
     case UNION:
     case VARDECIMAL:
-      return false;
-    default:
       return true;
+    default:
+      return false;
     }
   }
-
 
   /**
    * Checks if given major type is string scalar type.
@@ -490,7 +498,7 @@ public class Types {
     if (type.getMode() == REPEATED) {
       return false;
     }
-    switch(type.getMinorType()) {
+    switch (type.getMinorType()) {
     case FIXEDCHAR:
     case FIXED16CHAR:
     case VARCHAR:
@@ -736,8 +744,8 @@ public class Types {
    */
   public static MajorType.Builder calculateTypePrecisionAndScale(MajorType leftType, MajorType rightType, MajorType.Builder typeBuilder) {
     if (leftType.getMinorType().equals(rightType.getMinorType())) {
-      boolean isScalarString = Types.isScalarStringType(leftType) && Types.isScalarStringType(rightType);
-      boolean isDecimal = isDecimalType(leftType);
+      final boolean isScalarString = Types.isScalarStringType(leftType) && Types.isScalarStringType(rightType);
+      final boolean isDecimal = isDecimalType(leftType);
 
       if ((isScalarString || isDecimal) && leftType.hasPrecision() && rightType.hasPrecision()) {
         typeBuilder.setPrecision(Math.max(leftType.getPrecision(), rightType.getPrecision()));
@@ -769,8 +777,8 @@ public class Types {
       return true;
     }
 
-    List<MinorType> subtypes1 = type1.getSubTypeList();
-    List<MinorType> subtypes2 = type2.getSubTypeList();
+    final List<MinorType> subtypes1 = type1.getSubTypeList();
+    final List<MinorType> subtypes2 = type2.getSubTypeList();
     if (subtypes1 == subtypes2) { // Only occurs if both are null
       return true;
     }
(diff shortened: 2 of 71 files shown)


paul-rogers:
@arina-ielchiieva, here is a first-cut at the improved default values. Have tested selected mechanisms and CSV with schema. Have not yet run the full set of unit tests. Consider this a "preview" to begin the code review in parallel with the remaining busy-work needed to complete the PR.

common/src/main/java/org/apache/drill/common/types/Types.java
```diff
@@ -463,23 +462,29 @@ public static boolean usesHolderForGet(final MajorType type) {
...
-    switch(type.getMinorType()) {
+    return isFixedWidthType(type.getMinorType());
+  }
+
+  public static boolean isFixedWidthType(final MinorType type) {
+    return ! isVarWidthType(type);
+  }
+
+  public static boolean isVarWidthType(final MinorType type) {
+    switch(type) {
```
arina-ielchiieva:
```suggestion
    switch (type) {
```

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/ScalarReader.java
```diff
@@ -86,4 +87,10 @@
   LocalDate getDate();
   LocalTime getTime();
   Instant getTimestamp();
+
+  /**
+   * Return the value of the object using the extended type.
+   * @return
```
arina-ielchiieva: Please move add description to the return to avoid warnings in the IDE (just move upper line to `@return`).

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/AbstractWriteConverter.java
```diff
@@ -68,6 +68,11 @@ public ColumnMetadata schema() {
     return baseWriter.schema();
   }
 
+  @Override
+  public void setDefaultValue(Object value) {
+    throw new IllegalStateException("Cannot set a default value through a shim; types conflict.");
```
arina-ielchiieva: Should we include `value` in the error message?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/impl/VectorPrinter.java
```diff
@@ -33,7 +32,10 @@
   public static void printOffsets(UInt4Vector vector, int start, int length) {
     header(vector, start, length);
     for (int i = start, j = 0; j < length; i++, j++) {
-      if (j > 0) {
+      if (j % 40 == 0) {
```
arina-ielchiieva: How this will look like after the change?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractFixedWidthWriter.java
```diff
@@ -93,17 +112,62 @@ protected final int prepareWrite(int writeIndex) {
...
+      try {
+        // Catches int overflow. Does not catch overflow for smaller types.
+        setInt(Math.toIntExact(value));
+      } catch (final ArithmeticException e) {
+        throw InvalidConversionError.writeError(schema(), value, e);
+      }
+    }
+
+    @Override
+    public final void setDouble(final double value) {
```
arina-ielchiieva: Double covers Float as well?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/NullableScalarWriter.java
```diff
@@ -278,4 +278,9 @@ public void dump(HierarchicalFormatter format) {
     baseWriter.dump(format);
     format.endObject();
   }
+
+  @Override
+  public void setDefaultValue(Object value) {
+    throw new UnsupportedOperationException("Default values not supported for nullable types");
```
arina-ielchiieva: Maybe include `value` into the error message?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/OffsetVectorWriterImpl.java
```diff
@@ -302,4 +312,9 @@ public void dump(HierarchicalFormatter format) {
       .attribute("nextOffset", nextOffset)
       .endObject();
   }
+
+  @Override
+  public void setDefaultValue(Object value) {
+    throw new UnsupportedOperationException("Encoding not supported for offset vectors");
```
arina-ielchiieva: Same here.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/impl/VectorPrinter.java
paul-rogers: Before this change, I had a vector of 1000 items all on one line. After this change, the output is 40 elements per line. Note that this code is used only during debugging.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractFixedWidthWriter.java
paul-rogers: Yes, just as setInt() covers TInyInt, SmallInt, Int, UInt1, and UInt2.

paul-rogers:
Additional fixes: for Boolean and Decimal issues. Turns out the column writers for these types had some issues, now fixed.

exec/vector/src/main/codegen/templates/ColumnAccessors.java
```diff
@@ -302,97 +350,70 @@ public final Instant getTimestamp() {
...
+    <#if varWidth>
       super(((${drillType}Vector) vector).getOffsetVector());
-      <#else>
-      </#if>
-      <#if drillType == "VarDecimal">
+    </#if>
+    <#if drillType == "VarDecimal">
       // VarDecimal requires a scale. If not set, assume 0
       MajorType type = vector.getField().getType();
+      precision = type.hasPrecision() ? type.getPrecision() : 9;
```
arina-ielchiieva: Replace 9 with 38, Drill current behavior is to set default 38 when no precision is available.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/ConvertStringToBoolean.java
```diff
@@ -37,7 +37,11 @@ public void setString(final String value) {
       baseWriter.setNull();
     } else {
       try {
-        baseWriter.setInt(Boolean.parseBoolean(value) ? 1 : 0);
+        int intValue = Boolean.parseBoolean(value) ? 1 : 0;
```
arina-ielchiieva: Consider using `BooleanType.get` instead, it will cover all cases with numeric and literal values. Though we might need to move `BooleanType` to common module so it will be visible in vector one.

https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/expr/BooleanType.java#L70

common/src/main/java/org/apache/drill/common/types/BooleanType.java
```diff
@@ -15,45 +15,39 @@
...
-  FALSE(0, Arrays.asList("f", "false", "n", "no", "off", "0"));
+  TRUE(1, new String [] {"true", "1", "t", "y", "yes", "on",}),
+  FALSE(0, new String [] {"false", "0", "f", "n", "no", "off"});
 
   private final int numericValue;
-  private final List<String> literals;
+  private final String[] literals;
 
-  BooleanType(int numericValue, List<String> literals) {
+  BooleanType(int numericValue, String literals[]) {
```
arina-ielchiieva: `String literals[]` -> `String[] literals`

paul-rogers:
Fixed a number of issues relating to Boolean and Decimal types. Provided support for blank columns in CSV. Added a bunch more unit tests.

At this point, we should have everything we wanted for the alpha version of the schema feature. There is much more we can do. But, the most important next step is to allow others to try the feature and provide feedback.

This PR passes the full set of Drill unit tests. After a final code review, it should be ready to go.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/ColumnMetadata.java
```diff
@@ -57,6 +57,13 @@
    */
   public static final String PROJECTED_PROP = DRILL_PROP_PREFIX + "projected";
 
+  /**
+   * Indicates how to handle blanks. Must be one of the valid values defined
+   * in AbstractConvertFromString. Normally set on the converter by the plugin
+   * rather than by the user in the schema.
+   */
+  public static final String BLANK_AS_PROP = DRILL_PROP_PREFIX + "blank-as";
```
arina-ielchiieva: Since this is an interface, not need to `public static final`.

common/src/main/java/org/apache/drill/common/types/BooleanType.java
arina-ielchiieva: Please fix array declaration.

arina-ielchiieva:
@paul-rogers let me summarize new blank handling strategy. 
We introduce new column property `drill.blank-as`. It can have four values: `blank-as`, `null`, `0`, `simple`.  User can indicate which strategy to use in column schema. For example: 
`create or replace schema (id int, amount int properties { 'drill.blank-as' = 'simple' }) for table dfs.tmp.text_table`
If no blank strategy is indicated, behavior is the following:
1. for columns which are not indicated in schema, `blank-as` strategy will be used. I suggest, we rename `blank-as` value to `blank`. For example: `property { 'drill.blank-as' = 'blank' }. Why do you think?
2. for nullable columns indicated in schema, `null` strategy will be used.
3. for non-nullable columns indicated in schema with or without default value, which strategy will be used? `0` or `simple`? So far I don't see much difference between both, maybe we don't need one of them?


I have left some other comments in the PR, once we clarify this issue and address code review comments, PR will be good to go.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/SchemaHandler.java
```diff
@@ -139,6 +139,11 @@ public PhysicalPlan getPlan(SqlNode sqlNode) {
           .message(e.getMessage())
           .addContext("Error while preparing / creating schema for [%s]", schemaSource)
           .build(logger);
+      } catch (IllegalArgumentException e) {
```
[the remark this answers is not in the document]
paul-rogers: I did, in fact, use that approach at first. But, there are two distinct error cases here.

1) Could not read your file: there is something wrong at that level.
2) Could read the file just fine, the file parsed, but there is something wrong with the semantics (at present of the VarDecimal type.)

We could have conditional logic to create the right kind of error. Or, we can just do it the easy way as the code here does.

In fact, I think we could change the message for the IOException case to say that there was a problem reading the file.

By the way, were are parse exceptions reported? Should we be catching parse exceptions here?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/AbstractConvertFromString.java
```diff
@@ -26,9 +31,214 @@
  */
 public abstract class AbstractConvertFromString extends AbstractWriteConverter {
 
+  /**
+   * Property to control how the conversion handles blanks. Blanks are
+   * zero-length text fields (after triming whitespace.)
+   */
+  public static final String BLANK_ACTION_PROP = "blank-as";
```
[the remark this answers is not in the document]
paul-rogers: I did consider an enum. However, since these are just strings, the extra mechanism of an enum did not seem to add value: we would never actually use the enum.

paul-rogers:
@arina-ielchiieva, your summary is mostly correct. Just to refine a bit.

By default, if code using the row set framework asks to convert strings to other types, blanks have no special meaning. A blank will be parsed as any other string, which typically produces an error.

Any client of the row set framework can specify a blank-handling policy. Using an internal property set. The name of this internal property is `blank-as`. There are four choices:

* Unset: use the default policy described above.
* `null`: If the column is nullable, treat the blank as null. If non-nullable, leave the blank unchanged.
* `0`: Replace blanks with the value "0" for numeric types.
* `skip`: Skip blank values. This will set the column to its default value: `NULL` for nullable columns, the default value for non-nullable columns. If no default is set, then the "default default" of all-zero bytes is used.

(Note that I renamed "simple" to "skip".)

Normally, the blank policy is set by the reader. For example for CSV, it seemed to make sense to use the `skip` policy.

But, to provide maximum flexibility (and because there are many different requirements), the user can also optionally set the `drill.blank-as` property on a column. If set, that property overrides anything the reader may have set. For example, suppose I want to use -1 for missing columns, but 0 for blank columns. I could set the column default value to -1, then set the `drill.blank-as` property to `0`.

The bottom line for users of the CSV file format, with a schema, is that, by default, blanks are skipped and either become `NULL` or the default value.

Note also that this change strips leading and trailing white space from columns prior to type conversion. So a value of "  " is trimmed to "" and treated as a blank string. Trimming is *not* done for values stored as VARCHAR. In this case, if the value is "  ", that is what will be stored in the vector.

This is all a first draft. Let's get this into the hands of users as an "alpha", get some feedback, and adjust the code based on what we learn.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/SchemaHandler.java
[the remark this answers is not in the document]
arina-ielchiieva: Good point, I'll check.

paul-rogers:
Squashed commits. Should be ready to go.

PR #1727, opened 2019-04-02
DRILL-7145: Exceptions happened during retrieving values from ValueVe…

agozhiy:
…ctor are not being displayed at the Drill Web UI

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.server.rest;
 
+import org.apache.drill.exec.ExecConstants;
 import org.apache.drill.exec.util.ValueVectorElementFormatter;
 import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
 import org.apache.drill.shaded.guava.com.google.common.collect.Maps;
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/WebUserConnection.java
```diff
@@ -151,7 +151,7 @@ public void sendData(RpcOutcomeListener<Ack> listener, QueryWritableBatch result
         loader.clear();
       }
     } catch (Exception e) {
-      exception = UserException.systemError(e).build(logger);
+      throw UserException.systemError(e).build(logger);
```
aielchiieva: I don't think we should throw an exception here. We should stick to original approach and store it but just add method `getException()` in`AbstractDisposableUserClientConnection` class, similar to `getError()`. And then do proper handling of both.

arina-ielchiieva: I don't think we should throw an exception here. We should stick to original approach and store it but just add method getException() inAbstractDisposableUserClientConnection class, similar to getError(). And then do proper handling of both.

agozhiy: Changed the approach to calling listener.failed() instead of throwing the exception.

arina-ielchiieva:
+1, LGTM.

arina-ielchiieva:
Revoking +1 since more changes are needed.

arina-ielchiieva:
+1

PR #1728, opened 2019-04-02
DRILL-7089: Implement caching for TableMetadataProvider at query level and adapt statistics to use Drill metastore API

vvysotskyi:
In the scope of this PR introduced caching of table metadata (schema and statistics) at the query level.
Introduced `MetadataProviderManager` which holds both `SchemaProvider` and `DrillStatsTable` and `TableMetadataProvider` if it was already created.
`MetadataProviderManager` instance will be cached and used for every `DrillTable` which corresponds to the same table.
Such an approach was used to preserve lazy initialization of group scan and `TableMetadataProvider` instances, so once the first instance of `TableMetadataProvider` is created, it will be stored in the `MetadataProviderManager` and its metadata will be reused for all further `TableMetadataProvider` instances.

Another part of this PR is connected with the adoption of statistics to use Drill Metastore API. Enhanced logic to distinguish exact and estimated metadata, and used `TableMetadata` for obtaining statistics.

Will create and attach a class diagram later.

Also, tests should be run for this PR, so for now, I'll leave it in draft state.

### DIFF
--- a/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveParquetTableMetadataProvider.java
+++ b/contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveParquetTableMetadataProvider.java
@@ -49,17 +49,17 @@ public class HiveParquetTableMetadataProvider extends BaseParquetMetadataProvide
                                          HivePartitionHolder hivePartitionHolder,
                                          HiveStoragePlugin hiveStoragePlugin,
                                          ParquetReaderConfig readerConfig) throws IOException {
-    super(entries, readerConfig, null, null);
+    super(entries, readerConfig);
     this.hiveStoragePlugin = hiveStoragePlugin;
     this.hivePartitionHolder = hivePartitionHolder;
 
-    init();
+    init(null);
   }
 
   public HiveParquetTableMetadataProvider(HiveStoragePlugin hiveStoragePlugin,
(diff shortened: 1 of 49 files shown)


vvysotskyi:
Diagrams of the classes introduced in this PR: https://docs.google.com/presentation/d/1XG_xgR4okzXaJ3Z7HFHfzCwlM5VkNfre8GFEAd2Zo8k/edit?usp=sharing

vvysotskyi:

vvysotskyi:
Regression tests are passed for this PR, so it is ready to be reviewed.

amansinha100:
@vvysotskyi one other thought is that since this PR touches several of the Statistics related classes,  considering the Drill 1.16.0 release is imminent and the core statistics testing is ongoing, it would be disruptive to introduce these changes at this stage.  Even though functional regression might pass, we have had performance issues even with statistics disabled, so I feel hesitant for 1.16.0.  Let me know what you think.

amansinha100:
@vvysotskyi thanks for the explanations.  I am mostly good with the PR..so its a +1 from me but it would be good if @gparai can take a quick look at the statistics related changes.

vvysotskyi:
Rebased onto the master and resolved merge conflicts.

amansinha100:
@vvysotskyi if the regression tests pass you can update the read-to-commit label in the JIRA.

vvysotskyi:
@amansinha100, thanks for pointing this, they passed, so I have added `ready-to-commit` label.

PR #1730, opened 2019-04-02
DRILL-7152: During histogram creation handle the case when all values…

amansinha100:
… of a column are NULLs.

Please see [DRILL-7152](https://issues.apache.org/jira/browse/DRILL-7152) for a description of the issue.  It was caused because all the column's values are NULLs and the t-digest code-gen functions tried to generate an output for an empty t-digest since it does not store any NULL values.  The fix is to check the t-digest size() first before trying to create the output.


amansinha100:
@gparai could you please review ?  Thanks.

gparai:
@amansinha100 please take a look at the Travis failure. Otherwise, changes LGTM.

PR #1733, opened 2019-04-04
DRILL-7119: Compute range predicate selectivity using histograms.

amansinha100:


amansinha100:

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/NumericEquiDepthHistogram.java
```diff
@@ -69,27 +79,177 @@ public void setNumRowsPerBucket(long numRows) {
...
+
+            // if value is less than or equal to the first bucket's start point then all rows qualify
+            int result = value.compareTo(buckets[first]);
+            if (result <= 0) {
+              return LARGE_SELECTIVITY;
+            }
+            // if value is greater than the end point of the last bucket, then none of the rows qualify
+            result = value.compareTo(buckets[last]);
+            if (result > 0) {
+              return SMALL_SELECTIVITY;
```
gparai: For posterity, please add an explanation for SMALL_SELECTIVITY > 0 and the meaning of buckets[n], if not added elsewhere.

gparai:
@amansinha100 except the minor comment the code looks good. I do not see a testcase illustrating the use of histograms.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/NumericEquiDepthHistogram.java
amansinha100: Done.  I added a comment when we declare the class variable.

amansinha100:
@gparai I added a unit test for histogram usage and addressed your comment.  Could you pls take another look ?

PR #1738, opened 2019-04-07
DRILL-7062: Initial implementation of run-time row-group pruning

Ben-Zvi:
This PR is the initial implementation of the **Run-Time row-group pruning** (RTP); i.e., as oppose to Plan-Time pruning which usually works serially over the metadata - the run-time work applies the query predicate in parallel, at each minor fragment that scans several Parquet row-groups, using their files' footers.  Note that the new RTP implementation currently is independent from the plan-time pruning; e.g., the same predicate applied for plan-time may be applied again at run-time (future work would remove or reduce this predicate).

  Overview of the key changes:

(1) The RTP is enabled by passing the relevant filter expression to the runtime via the method *setFilterForRuntime* (in AbstractGroupScanWithMetadata). A value of null or a TRUE boolean (the default initial value) for the filter in that class means "disable RTP".
    The RTP can also be avoided/skipped using a new "skip RTP" option (default false). Due to some issue in the Hive Native scanner, the RTP is also currently disabled for that scanner (will create later a Jira to fix that).
  The *setFilterForRuntime* method is called in *doOnMatch* (in ParquetPushDownFilter), depending on the results of the plan-time pruning (e.g., a new filter expression may be composed).

(2) Changes for the Plan-Time: Changed the "Threshold" option from Positive Long to any Long, so that plan-time can be disabled with a value like zero. Also rewrote the documentation string for that option.
    Also added a special check-and-return for this option (in AbstractParquetGroupScan) , else the prior code "over-pruned" in that case.

(3) The selection-root Path field was added to the AbstractParquetRowGroupScan.

(4) A new constructor added for ParquetGroupScan (so it can be used without a format plugin).

(5) Two new query profile Metrics were added to the ParquetReaderStats: NUM_ROWGROUPS (i.e., the number assigned to this fragment (after plan time pruning)), and ROWGROUPS_PRUNED (i.e. how many of those rowgroups were eliminated/pruned). When all the rowgroups are pruned, the two numbers are identical.

(6) A test was added which disables plan-time pruning, selects from a multi-rowgroup file, and then verifies the result metrics.

(7) The "filter=(...)" now shows in the Scan operator in the physical plan. So needed to fix some unit tests that did not expect this part (basically added '.*' in the expected expression). 

(8) The actual Run Time Pruning work is taking place in the method *getBatch* (in AbstractParquetScanBatchCreator). There is a "for" loop iterating over all the assigned rowgroups, creating a reader for each (to be given to the scan batch later). The pruning code was added after the original code that gets the rowgroup footer.
     The pruning is done by applying the filter predicate (similar to how the plan-time works) to each rowgroup's footer, and skipping that footer when the predicate fails (i.e., no reader would be created for that rowgroup).
    In case **all** the rowgroups are pruned, the first rowgroup is anyway selected so that the schema would be returned. Note: In this case, our initial implementation assigned zero rows to read for that reader, but something broke at the scanner (also broke when 1 row was set). So currently *_all_* the rows of that rowgroup are read in this case. (Future: fix that to avoid waste).

   In order to use the pruning logic at run-time, some changes were made:
 (_a_) Several methods and definitions in Metadata.java were changed to become public and static.
       Note that they refer explicitly to metadata V3 (later to become V4), even though the run-time is independent of the metadata versioning (there was no cleaner abstraction to use).
 (_b_) The prior code (in the for loop) was already reading the footer; to avoid reading the footer again, some of the APIs in Metadata.java were changed to pass along the footer, and avoid reading it again (if it is not null).
   Some testing we did show that the savings (by not reading the footer again) are minor (probably due to file system caching). So if people think that the APIs change are not good, this item (c) can be undone.
 (_c_) The pruning code initialized few "per file" objects once per file (that's where (3) above is needed), and then builds the per-rowgroup filter-predicate and columns-statistics (where (4) is needed), and performs the match.
(_d_) The rest of the code in the original for loop (which created the reader, etc) was moved into a new private method - *createReaderAndImplicitColumns* (that code was moved by Intellij, so no need to review every line :-).  The reason is that this method needs to be called once outside the for loop (in the case where all rowgroups were pruned).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
@@ -727,6 +727,10 @@ public final class ExecConstants {
 
   public static final String BOOTSTRAP_STORAGE_PLUGINS_FILE = "bootstrap-storage-plugins.json";
 
+  public static final String SKIP_RUNTIME_ROWGROUP_PRUNING_KEY = "exec.storage.skip_runtime_rowgroup_pruning";
+  public static final OptionValidator SKIP_RUNTIME_ROWGROUP_PRUNING = new BooleanValidator(SKIP_RUNTIME_ROWGROUP_PRUNING_KEY,
+    new OptionDescription("Enables skipping the runtime pruning of the rowgroups"));
+
   public static final String DRILL_SYS_FILE_SUFFIX = ".sys.drill";
 
   public static final String ENABLE_WINDOW_FUNCTIONS = "window.enable";
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
@@ -32,6 +32,7 @@ import org.apache.drill.exec.compile.sig.ConstantExpressionIdentifier;
 import org.apache.drill.exec.expr.ExpressionTreeMaterializer;
 import org.apache.drill.exec.expr.fn.FunctionImplementationRegistry;
 import org.apache.drill.exec.expr.stat.RowsMatch;
+import org.apache.drill.exec.ops.OptimizerRulesContext;
 import org.apache.drill.exec.ops.UdfUtilities;
 import org.apache.drill.exec.planner.physical.PlannerSettings;
 import org.apache.drill.exec.record.MaterializedField;
@@ -67,6 +68,9 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+
+// import static org.apache.drill.exec.ExecConstants.HIVE_OPTIMIZE_SCAN_WITH_NATIVE_READERS;
+import static org.apache.drill.exec.ExecConstants.SKIP_RUNTIME_ROWGROUP_PRUNING_KEY;
 
 /**
  * Represents table group scan with metadata usage.
@@ -191,12 +195,25 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
     this.filter = filter;
   }
 
+  /**
+   *  Set the filter - thus enabling runtime rowgroup pruning
+   *  The runtime pruning can be disabled with an option.
+   * @param filterExpr The filter to be used at runtime to match with rowgroups' footers
+   * @param optimizerContext The context for the options
+   */
+  public void setFilterForRuntime(LogicalExpression filterExpr, OptimizerRulesContext optimizerContext) {
+    OptionManager options = optimizerContext.getPlannerSettings().getOptions();
+    boolean skipRuntimePruning = options.getBoolean(SKIP_RUNTIME_ROWGROUP_PRUNING_KEY); // if option is set to disable runtime pruning
+    if ( ! skipRuntimePruning ) { setFilter(filterExpr); }
+  }
+
   @Override
   public AbstractGroupScanWithMetadata applyFilter(LogicalExpression filterExpr, UdfUtilities udfUtilities,
       FunctionImplementationRegistry functionImplementationRegistry, OptionManager optionManager) {
 
     // Builds filter for pruning. If filter cannot be built, null should be returned.
-    FilterPredicate filterPredicate = getFilterPredicate(filterExpr, udfUtilities, functionImplementationRegistry, optionManager, true);
+    FilterPredicate filterPredicate =
+            getFilterPredicate(filterExpr, udfUtilities, functionImplementationRegistry, optionManager, true);
     if (filterPredicate == null) {
       logger.debug("FilterPredicate cannot be built.");
       return null;
@@ -271,6 +288,14 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
    */
   protected abstract GroupScanWithMetadataFilterer getFilterer();
 
+  public FilterPredicate getFilterPredicate(LogicalExpression filterExpr,
+                                                   UdfUtilities udfUtilities,
+                                                   FunctionImplementationRegistry functionImplementationRegistry,
+                                                   OptionManager optionManager,
+                                                   boolean omitUnsupportedExprs) {
+    return getFilterPredicate(filterExpr, udfUtilities, functionImplementationRegistry, optionManager,
+            omitUnsupportedExprs, supportsFileImplicitColumns(), (TupleSchema) getTableMetadata().getSchema());
+  }
   /**
    * Returns parquet filter predicate built from specified {@code filterExpr}.
    *
@@ -281,20 +306,19 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
    *                                       may be omitted from the resulting expression
    * @return parquet filter predicate
    */
-  public FilterPredicate getFilterPredicate(LogicalExpression filterExpr,
+  public static FilterPredicate getFilterPredicate(LogicalExpression filterExpr,
                                             UdfUtilities udfUtilities,
                                             FunctionImplementationRegistry functionImplementationRegistry,
                                             OptionManager optionManager,
-                                            boolean omitUnsupportedExprs) {
-    TupleMetadata types = getSchema();
-    if (types == null) {
-      throw new UnsupportedOperationException("At least one schema source should be available.");
-    }
+                                            boolean omitUnsupportedExprs,
+                                            boolean supportsFileImplicitColumns,
+                                            TupleSchema tupleSchema) {
+    TupleMetadata types = tupleSchema.copy();
 
     Set<SchemaPath> schemaPathsInExpr = filterExpr.accept(new FilterEvaluatorUtils.FieldReferenceFinder(), null);
 
     // adds implicit or partition columns if they weren't added before.
-    if (supportsFileImplicitColumns()) {
+    if (supportsFileImplicitColumns) {
       for (SchemaPath schemaPath : schemaPathsInExpr) {
         if (isImplicitOrPartCol(schemaPath, optionManager) && SchemaPathUtils.getColumnMetadata(schemaPath, types) == null) {
           types.add(MaterializedField.create(schemaPath.getRootSegmentPath(), Types.required(TypeProtos.MinorType.VARCHAR)));
@@ -467,7 +491,7 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
   protected abstract boolean supportsFileImplicitColumns();
   protected abstract List<String> getPartitionValues(LocationProvider locationProvider);
 
-  protected boolean isImplicitOrPartCol(SchemaPath schemaPath, OptionManager optionManager) {
+  public static boolean isImplicitOrPartCol(SchemaPath schemaPath, OptionManager optionManager) {
     Set<String> implicitColNames = ColumnExplorer.initImplicitFileColumns(optionManager).keySet();
     return ColumnExplorer.isPartitionColumn(optionManager, schemaPath) || implicitColNames.contains(schemaPath.getRootSegmentPath());
   }
@@ -628,7 +652,6 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
             matchAllMetadata = true;
             partitions = filterAndGetMetadata(schemaPathsInExpr, source.getPartitionsMetadata(), filterPredicate, optionManager);
           } else {
-            matchAllMetadata = false;
             overflowLevel = MetadataLevel.PARTITION;
           }
         }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
@@ -17,6 +17,19 @@
  */
 package org.apache.drill.exec.store.parquet;
 
+import org.apache.drill.common.expression.LogicalExpression;
+import org.apache.drill.common.expression.SchemaPath;
+import org.apache.drill.common.expression.ValueExpressions;
+import org.apache.drill.exec.expr.FilterPredicate;
+import org.apache.drill.exec.expr.fn.FunctionImplementationRegistry;
+import org.apache.drill.exec.expr.stat.RowsMatch;
+import org.apache.drill.exec.physical.base.AbstractGroupScanWithMetadata;
+import org.apache.drill.exec.record.metadata.TupleSchema;
+import org.apache.drill.exec.store.CommonParquetRecordReader;
+import org.apache.drill.exec.store.parquet.metadata.Metadata;
+import org.apache.drill.exec.store.parquet.metadata.MetadataBase;
+import org.apache.drill.exec.store.parquet.metadata.Metadata_V4;
+import org.apache.drill.metastore.ColumnStatistics;
 import org.apache.drill.shaded.guava.com.google.common.base.Functions;
 import org.apache.drill.shaded.guava.com.google.common.base.Stopwatch;
 import org.apache.drill.shaded.guava.com.google.common.collect.Maps;
@@ -28,11 +41,11 @@ import org.apache.drill.exec.physical.impl.ScanBatch;
 import org.apache.drill.exec.proto.helper.QueryIdHelper;
 import org.apache.drill.exec.server.options.OptionManager;
 import org.apache.drill.exec.store.ColumnExplorer;
-import org.apache.drill.exec.store.RecordReader;
 import org.apache.drill.exec.store.dfs.DrillFileSystem;
 import org.apache.drill.exec.store.parquet.columnreaders.ParquetRecordReader;
 import org.apache.drill.exec.store.parquet2.DrillParquetReader;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.Path;
 import org.apache.parquet.hadoop.CodecFactory;
 import org.apache.parquet.hadoop.ParquetFileReader;
@@ -42,10 +55,12 @@ import org.apache.parquet.hadoop.util.HadoopInputFile;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 public abstract class AbstractParquetScanBatchCreator {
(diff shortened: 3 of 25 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -149,6 +219,77 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
+          footer,
+          rowGroup,
+          columnExplorer.getTableColumns(),
+          fs,
+          containsCorruptDates);
+    } else {
+      reader = new ParquetRecordReader(context,
+          rowGroup.getPath(),
+          rowGroup.getRowGroupIndex(),
+          rowGroup.getNumRecordsToRead(), // if readSchemaOnly - then set to zero rows to read (currently breaks the ScanBatch)
```
amansinha100: The `readSchemaOnly` is not getting used, so why even pass it to this function ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -68,76 +83,131 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
+              prevRowGroupPath = rowGroup.getPath(); // for next time
+            }
+
+            MetadataBase.RowGroupMetadata rowGroupMetadata = fileMetadataV3.getRowGroups().get(rowGroup.getRowGroupIndex());
+
+            Map<SchemaPath, ColumnStatistics> columnsStatistics = ParquetTableMetadataUtils.getRowGroupColumnStatistics(tableMetadataV3, rowGroupMetadata);
+
+            List<SchemaPath> columns = columnsStatistics.keySet().stream().collect(Collectors.toList());
+
+            ParquetGroupScan parquetGroupScan = new ParquetGroupScan( context.getQueryUserName(), metadataProvider, fileSelection, columns, readerConfig, filterExpr);
```
amansinha100: Creating a new instance of the ParquetGroupScan seems too much overhead since this is happening for each row group.   I understand how you may have arrived at this since the run-time does not have access to the planner objects such as the original ParquetGroupScan.   However, the only reason you need this is to create a FilterPredicate, which is accessible through the ParquetGroupScan.  Could we just create a single (dummy) instance of ParquetGroupScan using 1 file (or row group) outside the main For loop and just create the FilterPredicate right in the beginning before even iterating over all the row groups ?  The construction of the FilterPredicate in `getFilterPredicate()`  itself does not seem to have any direct dependence on any internal state of the ParquetGroupScan as far as I can tell.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
```diff
@@ -52,6 +52,7 @@
 
   private final ParquetFormatPlugin formatPlugin;
   private final ParquetFormatConfig formatConfig;
+  private final Collection<DrillbitEndpoint> drillbitEndpoints;
```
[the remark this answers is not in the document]
Ben-Zvi: How can we get to the formatPlugin (or its context) at the runtime ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -68,76 +83,131 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
+              logger.trace("ParquetTrace,Read Footer,{},{},{},{},{},{},{}", "", rowGroup.getPath(), "", 0, 0, 0, timeToRead);
+            }
+            footers.put(rowGroup.getPath(), footer);
           }
+          ParquetMetadata footer = footers.get(rowGroup.getPath());
+
+          //
+          //   If a filter is given (and it is not just "TRUE") - then use it to perform run-time pruning
+          //
+          if ( filterExpr != null && ! (filterExpr instanceof ValueExpressions.BooleanExpression)  ) { // skip when no filter or filter is TRUE
```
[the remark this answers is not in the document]
Ben-Zvi: Can the planner ever create a FALSE filter expression ? It is equivalent to eliminating the Scan operator ....

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -68,76 +83,131 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
-      } catch (IOException e) {
-        throw new ExecutionSetupException(e);
+      // Update stats (same in every reader - the others would just overwrite the stats)
+      for (RecordReader rr : readers ) {
+        if ( rr instanceof ParquetRecordReader ) {
+          ((ParquetRecordReader) rr).updateRowgroupsStats(rowGroupScan.getRowGroupReadEntries().size(), rowgroupsPruned);
+        }
       }
+
+    } catch (IOException|InterruptedException e) {
```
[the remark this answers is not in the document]
Ben-Zvi: The *Metadata.getParquetFileMetadata_v3(...)* on line 155.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
Ben-Zvi: This was a last minute change (before it used this flag to set the num-rows-to-read to zero).
Soon (another Jira) we should retry implementing this option, so left that flag in the code as a reminder (maybe should be just a TODO comment)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
Ben-Zvi: 1. The total "per rowgroup" overhead (including the ParquetGroupScan construction) is usually less than a millisecond (on my Mac). (That does not include the "per file/footer" overhead).
2. In case the input list of rowgroups spans multiple files, we'd need to recreate a ParquetGroupScan for each file (could there be changes in the columns or schema between files ? Seem there is some dependence on those.)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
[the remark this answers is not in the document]
amansinha100: We do want to get the schema even for FALSE filter.  Currently, what actually happens is that a FALSE filter in the logical plan gets converted to a LIMIT 0, so typically the scanner would see limit 0; however, if the FALSE filter was created during some intermediate stage of planning due to other transformations that occurred after the Limit_0 transformation, then it is possible that the runtime might see a FALSE filter.  I haven't confirmed it but I believe it is possible.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
amansinha100: I am actually less concerned about the elapsed time and more about the number of objects created on the Java heap.  The `ParquetGroupScan` is at least 100 bytes each and since we are targeting use cases with about million row groups per minor fragment, it means 100 MB per minor fragment (assuming the garbage collector is not doing the cleanup while the Foo loop is executing).  If there are 24 minor frags, this would add up to 2.4GB heap per node which would cause quite a bit of GC activity subsequently.  

I have made some changes on top of your branch which avoids this by directly calling a static method to create a FilterPredicate.   The pruning unit tests are passing with these changes.  I will send my changes to you offline and we can discuss.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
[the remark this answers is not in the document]
amansinha100: The runtime has access to the `ExecutorFragmentContext` which contains the `getBits()` method that returns the list of DrillBits.

exec/java-exec/src/main/java/org/apache/drill/exec/ExecConstants.java
```diff
@@ -727,6 +727,10 @@ private ExecConstants() {
 
   public static final String BOOTSTRAP_STORAGE_PLUGINS_FILE = "bootstrap-storage-plugins.json";
 
+  public static final String SKIP_RUNTIME_ROWGROUP_PRUNING_KEY = "exec.storage.skip_runtime_rowgroup_pruning";
```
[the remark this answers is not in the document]
Ben-Zvi: The namespace 'exec.storage.parquet' is not yet used for any option, and adding "parquet_" to the option name makes it longer and more typo prone (unless you are French :-) . Any user who understands when/why to use this option probably knows that "rowgroups" means Parquet.
      Are we going to introduce several other new options under this namespace ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -127,9 +127,21 @@ public static void createMeta(FileSystem fs, Path path, ParquetReaderConfig read
...
    */
-  public static ParquetTableMetadata_v3 getParquetTableMetadata(FileSystem fs, String path, ParquetReaderConfig readerConfig) throws IOException {
+  public static ParquetTableMetadata_v3 getParquetTableMetadata(ParquetMetadata footer, FileSystem fs, String path, ParquetReaderConfig readerConfig) throws IOException {
     Metadata metadata = new Metadata(readerConfig);
-    return metadata.getParquetTableMetadata(path, fs);
+    return metadata.getParquetTableMetadata(path, fs, footer);
+  }
+
+  /**
+   *  When the footer is not yet available (it would be read)
```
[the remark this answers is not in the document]
Ben-Zvi: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetGroupScan.java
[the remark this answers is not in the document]
Ben-Zvi: With the latest changes, no need for a special constructor for the ParquetGroupScan.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -149,6 +219,77 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
+   * @param oContext
+   * @param columnExplorer
+   * @param readers
+   * @param implicitColumns
+   * @param mapWithMaxColumns
+   * @param rowGroup
+   * @param fs
+   * @param footer
+   * @param readSchemaOnly - if true sets the number of rows to read to be zero
+   * @return
```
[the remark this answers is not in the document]
Ben-Zvi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -149,6 +219,77 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
+  /**
+   *  Create a reader and add it to the list of readers.
+   *
+   * @param context
+   * @param rowGroupScan
+   * @param oContext
+   * @param columnExplorer
+   * @param readers
+   * @param implicitColumns
+   * @param mapWithMaxColumns
```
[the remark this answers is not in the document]
Ben-Zvi: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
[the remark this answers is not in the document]
Ben-Zvi: Added a check for true - getBoolean() .

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
Ben-Zvi: Commented this out, and added TODO comments.

Ben-Zvi:
The code has been modified, simplified and passes all the tests. The changes from the first PR:

(a) Cancelled item (4) above (new ctor for ParquetGroupScan).
(b) Pass the TupleSchema thru the plan API, and use it to construct the FilterPredicate outside the runtime FOR loop.
(c) Reduction of any unneeded new object construction and other work inside the FOR loop. E.g., only allocate metadata for the "interesting" columns.
(d) Undid most of the footer passing thru the Metadata API (item (8-b) above).
(e) Solved the issue with the Hive Drill Native Scan, thus RTP is no longer disabled in that case (item (1) above)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata.java
```diff
@@ -448,23 +476,23 @@ private ColTypeInfo getColTypeInfo(MessageType schema, Type type, String[] path,
...
-      metadata = processUserUgi.doAs((PrivilegedExceptionAction<ParquetMetadata>)() -> {
-        try (ParquetFileReader parquetFileReader = ParquetFileReader.open(HadoopInputFile.fromStatus(file, conf), readerConfig.toReadOptions())) {
-          return parquetFileReader.getFooter();
-        }
-      });
-    } catch(Exception e) {
-      logger.error("Exception while reading footer of parquet file [Details - path: {}, owner: {}] as process user {}",
-        file.getPath(), file.getOwner(), processUserUgi.getShortUserName(), e);
-      throw e;
+  public static ParquetFileMetadata_v3 getParquetFileMetadata_v3(ParquetTableMetadata_v3 parquetTableMetadata, ParquetMetadata footer,
```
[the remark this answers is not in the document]
Ben-Zvi: Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -68,76 +84,144 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
+      if ( readers.size() == 0 && firstRowGroup != null ) {
+        logger.trace("All row groups were pruned out. Returning the first: {} (row count {}) for its schema", firstRowGroup.getPath(), firstRowGroup.getNumRecordsToRead());
+        DrillFileSystem fs = fsManager.get(rowGroupScan.getFsConf(firstRowGroup), firstRowGroup.getPath());
+        mapWithMaxColumns = createReaderAndImplicitColumns(context, rowGroupScan, oContext, columnExplorer, readers, implicitColumns, mapWithMaxColumns, firstRowGroup, fs,
+          firstFooter /*, true */);
+      }
+
+      // Update stats (same in every reader - the others would just overwrite the stats)
+      for (RecordReader rr : readers ) {
+        if ( rr instanceof ParquetRecordReader ) {
```
amansinha100: It would be useful to have this stats for the `DrillParquetReader` as well.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
```diff
@@ -68,76 +84,144 @@ protected ScanBatch getBatch(ExecutorFragmentContext context, AbstractParquetRow
...
 
-      } catch (IOException e) {
-        throw new ExecutionSetupException(e);
+            //
+            // Perform the Run-Time Pruning - i.e. Skip this rowgroup if the match fails
+            //
+            RowsMatch match = FilterEvaluatorUtils.matches(filterPredicate, columnsStatistics, footerRowCount);
+            if (timer != null) { // if tracing
+              long timeToRead = timer.elapsed(TimeUnit.MICROSECONDS);
+              logger.trace("Run-time pruning: {} row-group {} (RG index: {} row count: {}), took {} usec", match == RowsMatch.NONE ? "Excluded" : "Included", rowGroup.getPath(),
```
amansinha100: Ok to log this at the TRACE level but could you add an INFO level log ***outside*** the for loop indicating how many row groups were evaluated and how many were pruned out.  Even though the operator profile has the metrics,  the log would be quite useful.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
```diff
@@ -43,6 +44,7 @@
 import org.apache.drill.exec.store.ColumnExplorer;
 import org.apache.drill.exec.store.dfs.FileSelection;
 import org.apache.drill.exec.store.parquet.FilterEvaluatorUtils;
+// import org.apache.drill.exec.store.parquet.ParquetGroupScan;
```
amansinha100: Remove.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
rhou1: Please also add a keyword such as PARQUET_RUNTIME_PRUNING so that the QA test framework can find this log message and verify it.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
Ben-Zvi: Done, in a followup commit

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
Ben-Zvi: Done, in a followup commit

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
Ben-Zvi: Done in a followup commit

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/AbstractParquetScanBatchCreator.java
amansinha100: It does not look like any INFO level logging was added..I still see only TRACE and DEBUG levels. 
Also, @rhou1's suggestion for annotating the info message with a prefix seems reasonable for testing purposes.

PR #1741, opened 2019-04-08
DRILL-7159: Fix typeString method to return correct name for MAP (aka STRUCT)

arina-ielchiieva:
Jira [DRILL-7159](https://issues.apache.org/jira/browse/DRILL-7159).


arina-ielchiieva:
@vvysotskyi please review.

PR #1742, opened 2019-04-09
DRILL-7160: e.q.max_rows QUERY-level option shown even if not set

kkhatua:
The fix is to force setting to zero IFF autoLimit was intended to be set originally but is inapplicable; such as 'SHOW DATABASES'. If autolimit was not intended to be applied, setting the value to zero is not required.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
@@ -216,27 +216,31 @@ public class DrillSqlWorker {
     return handler.getPlan(sqlNode);
   }
 
-  private static boolean isAutoLimitShouldBeApplied(QueryContext context, SqlNode sqlNode) {
-    return (context.getOptions().getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue() > 0) && sqlNode.getKind().belongsTo(SqlKind.QUERY)
-        && (sqlNode.getKind() != SqlKind.ORDER_BY || isAutoLimitLessThanOrderByFetch((SqlOrderBy) sqlNode, context));
+  private static boolean isAutoLimitShouldBeApplied(SqlNode sqlNode, int queryMaxRows) {
+    return (queryMaxRows > 0) && sqlNode.getKind().belongsTo(SqlKind.QUERY)
+        && (sqlNode.getKind() != SqlKind.ORDER_BY || isAutoLimitLessThanOrderByFetch((SqlOrderBy) sqlNode, queryMaxRows));
   }
 
   private static SqlNode checkAndApplyAutoLimit(SqlConverter parser, QueryContext context, String sql) {
     SqlNode sqlNode = parser.parse(sql);
-    if (isAutoLimitShouldBeApplied(context, sqlNode)) {
-      sqlNode = wrapWithAutoLimit(sqlNode, context);
+    int queryMaxRows = context.getOptions().getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue();
+    if (isAutoLimitShouldBeApplied(sqlNode, queryMaxRows)) {
+      sqlNode = wrapWithAutoLimit(sqlNode, queryMaxRows);
     } else {
-      context.getOptions().setLocalOption(ExecConstants.QUERY_MAX_ROWS, 0);
+      //Force setting to zero IFF autoLimit was intended to be set originally but is inapplicable
+      if (queryMaxRows > 0) {
+        context.getOptions().setLocalOption(ExecConstants.QUERY_MAX_ROWS, 0);
+      }
     }
     return sqlNode;
   }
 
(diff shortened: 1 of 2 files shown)


vvysotskyi:
@kkhatua, `exec.query.max_rows` query level option should be present in query profile only for the case when auto-limit is applied or can be applied, but currently, it is present even for the case when the non-select query is submitted and `exec.query.max_rows` is set to non-zero value. Please fix this case and verify that other corner cases are handled correctly.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
```diff
@@ -336,7 +336,8 @@ public String getOperatorsJSON() {
   }
 
   public Map<String, String> getQueryOptions() {
-    return getOptions(o -> OptionValue.OptionScope.QUERY == o.getScope());
+    // Skip reporting QUERY_MAX_ROWS if it is inapplicable and set to zero (e.g. query -> SHOW FILES)
+    return getOptions(o -> OptionValue.OptionScope.QUERY == o.getScope() && !(ExecConstants.QUERY_MAX_ROWS.equals(o.getName()) && String.valueOf(o.getValue()).equals("0")));
```
vvysotskyi: Please revert this change.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
```diff
@@ -226,7 +226,10 @@ private static SqlNode checkAndApplyAutoLimit(SqlConverter parser, QueryContext
     if (isAutoLimitShouldBeApplied(context, sqlNode)) {
       sqlNode = wrapWithAutoLimit(sqlNode, context);
     } else {
-      context.getOptions().setLocalOption(ExecConstants.QUERY_MAX_ROWS, 0);
+        //Force setting to zero IFF autoLimit was intended to be set originally but is inapplicable
```
vvysotskyi: Please fix indentation and please refactor methods added in the previous commit to pass the value of `context.getOptions().getOption(ExecConstants.QUERY_MAX_ROWS).num_val.intValue()` instead of `context`.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
kkhatua: This change actually helps to not show the value of Zero being applied, when the query is inapplicable. Are you sure we want to remove this?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
kkhatua: 👍

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
arina-ielchiieva: I think we should not set max rows count to query context if it is the same as default, thus this count won't appear in Query Profile. Current solution just fixes the symptom not the real problem.

kkhatua: Agreed. I am not sure, but I think I did this here 
https://github.com/apache/drill/blob/c3ee7949656fb4c1b144e1633f97002c159ec8f3/exec/java-exec/src/main/java/org/apache/drill/exec/ops/QueryContext.java#L120
because of the complexity in the logic  and (I think) because without this, the session value (0) gets precedence when the SYSTEM default exists.
I can revisit this post release to avoid blocking the release for now.

vvysotskyi: Yes, as I wrote in the previous comment, we can skip the issue I mentioned in one of the comments for now.

kkhatua:
@vvysotskyi , @arina-ielchiieva 

With the latest update, for non-applicable queries, there are no options shown that indicate that the `max_rows` has been set.
For applicable queries, based on the combination of what the SESSION and SYSTEM (default) values are, you get the following outcome with the scope of the option also indicated as shown in the profile:

|	SYSTEM	|	SESSION	|	Final	| ScopeSet	|
|----------------|--------------|-------------|----------------|
|	0		|	0		|	0		|	N/A		|
|	15		|	0		|	15		|	N/A		|
|	0		|	10		|	10		|	SESSION	|
|	15		|	10		|	10		|	SESSION	|
|	15		|	20		|	15		|	QUERY	|

The last one is required because there is no way for me to remove the SESSION level value and let only the SYSTEM value persist.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/profile/ProfileWrapper.java
vvysotskyi: @kkhatua, you haven't reverted this change.

PR #1743, opened 2019-04-10
DRILL-7165: Redundant Checksum calculating for ASC files

vdiravka:
- change 'checksum-maven-plugin' 'goal' - 'artifacts' -> 'files'
- specify 'includes' in 'fileSet' for 'checksum-maven-plugin'

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1792,7 +1792,7 @@
               <execution>
                 <id>source-release-checksum</id>
                 <goals>
-                  <goal>artifacts</goal>
+                  <goal>files</goal>
                 </goals>
               </execution>
             </executions>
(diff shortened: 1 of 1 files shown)


vdiravka:
@sohami Couold you please review?
Here is one not mandatory change: `drill-root` -> `apache-drill` project `artifactId`. 
It is more convenient to use it as a parameter `${project.artifactId}` or `${project.parent.artifactId}` instead of hardcoding `apache-drill` everywhere. 
The question is it safe to change Drill project `artifactId`? Looks like [`drill-root`](https://mvnrepository.com/artifact/org.apache.drill/drill-root) isn't used by external tools. 
Not sure there are other risks to rename it. If they are, please let me know.

sohami:
@vdiravka - I would recommend handling the rename change in separate PR rather than doing now if there is even slight risk of it breaking anything.

vdiravka:
@sohami PR for project root POM ArtifactId rename is created: #1746

PR #1744, opened 2019-04-11
Drill 7148 - Join order, multi-col ndv and aggregate rowcount fixes for TPCH queries

gparai:
@amansinha100 can you please do an initial review of the proposed fixes. @rhou1 can you please run perf tests with this branch. Here is a description of each commit:

DRILL-7148: Use different join cardinality and ndv estimation
Enable using the new formula which had the limitation of less accurate ndv computation. That has been addressed as part of this fix. Also, earlier multi-col ndv computation does not seem as accurate which has been fixed.

DRILL-7148: Additional performance fixes
HashAgg rowcount was underestimated for the first phase of a two-phase hash agg. This has been addressed.

DRILL-7148: Fix statistics row count
There was a bug which did not result in using correct rowcount for EnumerableTableScan which was affecting initial join ordering.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
@@ -667,4 +667,66 @@ public abstract class DrillRelOptUtil {
     }
     return drillTable;
   }
+
+  public static List<Pair<Integer, Integer>> analyzeSimpleEquiJoin(Join join) {
+    List<Pair<Integer, Integer>> joinConditions = new ArrayList<>();
+    try {
+      RexVisitor<Void> visitor =
+          new RexVisitorImpl<Void>(true) {
+            public Void visitCall(RexCall call) {
+              if (call.getKind() == SqlKind.AND || call.getKind() == SqlKind.OR) {
+                super.visitCall(call);
+              } else {
+                if (call.getKind() == SqlKind.EQUALS) {
+                  int leftFieldCount = join.getLeft().getRowType().getFieldCount();
+                  int rightFieldCount = join.getRight().getRowType().getFieldCount();
+                  RexNode leftComparand = call.operands.get(0);
+                  RexNode rightComparand = call.operands.get(1);
+                  RexInputRef leftFieldAccess = (RexInputRef) leftComparand;
+                  RexInputRef rightFieldAccess = (RexInputRef) rightComparand;
+                  if (leftFieldAccess.getIndex() >= leftFieldCount + rightFieldCount ||
+                      rightFieldAccess.getIndex() >= leftFieldCount + rightFieldCount) {
+                    joinConditions.clear();
+                    throw new Util.FoundOne(call);
+                  }
+                  /* Both columns reference same table */
+                  if ((leftFieldAccess.getIndex() >= leftFieldCount &&
+                      rightFieldAccess.getIndex() >= leftFieldCount) ||
+                          (leftFieldAccess.getIndex() < leftFieldCount &&
+                              rightFieldAccess.getIndex() < leftFieldCount)) {
+                    joinConditions.clear();
+                    throw new Util.FoundOne(call);
+                  } else {
+                    if (leftFieldAccess.getIndex() < leftFieldCount) {
+                      joinConditions.add(Pair.of(leftFieldAccess.getIndex(),
+                          rightFieldAccess.getIndex() - leftFieldCount));
+                    } else {
+                      joinConditions.add(Pair.of(rightFieldAccess.getIndex(),
+                          leftFieldAccess.getIndex() - leftFieldCount));
+                    }
+                  }
+                }
+              }
+              return null;
+            }
+          };
+      join.getCondition().accept(visitor);
+    } catch (Util.FoundOne ex) {
+      Util.swallow(ex, null);
+    }
+    return joinConditions;
+  }
+
+  public static List<RexInputRef> findAllRexInputRefs(final RexNode node) {
+    List<RexInputRef> rexRefs = new ArrayList<>();
+    RexVisitor<Void> visitor =
+            new RexVisitorImpl<Void>(true) {
+              public Void visitInputRef(RexInputRef inputRef) {
+                rexRefs.add(inputRef);
+                return super.visitInputRef(inputRef);
+              }
+            };
+    node.accept(visitor);
+    return rexRefs;
+  }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
@@ -17,14 +17,18 @@
  */
 package org.apache.drill.exec.planner.cost;
 
+import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.plan.volcano.RelSubset;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.SingleRel;
 import org.apache.calcite.rel.core.Join;
 import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.core.TableScan;
 import org.apache.calcite.rel.core.Window;
 import org.apache.calcite.rel.metadata.ReflectiveRelMetadataProvider;
 import org.apache.calcite.rel.metadata.RelMdDistinctRowCount;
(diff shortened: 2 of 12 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdRowCount.java
```diff
@@ -53,7 +53,21 @@ public Double getRowCount(Aggregate rel, RelMetadataQuery mq) {
 
     if (groupKey.isEmpty()) {
       return 1.0;
-    } else {
+    } else if (rel instanceof AggPrelBase &&
+            ((AggPrelBase) rel).getOperatorPhase() == AggPrelBase.OperatorPhase.PHASE_1of2) {
+      // Phase 1 Aggregate would return rows in the range [NDV, input_rows]. Hence, use the
+      // existing estimate of 1/10 * input_rows
+        Double distinctRowCount = mq.getRowCount(rel.getInput()) / 10;
```
amansinha100: Can we not get this default value from an existing Calcite utility ?  This will allow keeping in sync in the future if the ratio changes.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
```diff
@@ -667,4 +667,66 @@ public static DrillTable getDrillTable(final TableScan scan) {
...
+                super.visitCall(call);
+              } else {
+                if (call.getKind() == SqlKind.EQUALS) {
+                  int leftFieldCount = join.getLeft().getRowType().getFieldCount();
+                  int rightFieldCount = join.getRight().getRowType().getFieldCount();
+                  RexNode leftComparand = call.operands.get(0);
+                  RexNode rightComparand = call.operands.get(1);
+                  RexInputRef leftFieldAccess = (RexInputRef) leftComparand;
+                  RexInputRef rightFieldAccess = (RexInputRef) rightComparand;
+                  if (leftFieldAccess.getIndex() >= leftFieldCount + rightFieldCount ||
```
amansinha100: When will this condition happen ?  i.e the join column's index is out of range ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -80,10 +87,10 @@ public Double getDistinctRowCount(Join rel, RelMetadataQuery mq,
 
   @Override
   public Double getDistinctRowCount(RelNode rel, RelMetadataQuery mq, ImmutableBitSet groupKey, RexNode predicate) {
-    if (rel instanceof DrillScanRelBase) {                  // Applies to both Drill Logical and Physical Rels
+    if (rel instanceof TableScan) {                         // Applies to both Drill Logical and Physical Rels
```
amansinha100: Comment should indicate that it applies to Calcite logical table scan also since that seems to be the purpose of this change.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -208,8 +244,64 @@ public Double getDistinctRowCount(DrillJoinRelBase joinRel, RelMetadataQuery mq,
...
+        final RelOptUtil.InputFinder inputFinder = RelOptUtil.InputFinder.analyze(filter);
+        joinFiltersSet.add(inputFinder.inputBitSet.build());
+      }
+      for (int idx = 0; idx < groupKey.length(); idx++) {
+        if (groupKey.get(idx)) {
+          // GBY key is present in some filter - now try options A) and B) as described above
+          double ndvSGby = 0;
+          ImmutableBitSet sGby = getSingleGbyKey(groupKey, idx);
+          if (sGby != null) {
+            for (ImmutableBitSet jFilter : joinFiltersSet) {
```
amansinha100: Suppose the query is  ' SELECT  distinct a1 FROM t1, t2, t3  where t1.a1 = t2.a2 and t1.a1 = t2.a3 '
In this case, the grouping key is involved in 2 joins .. can we do an offline walk-through to see whether the top join's NDV is being used.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -229,4 +321,42 @@ public Double getDistinctRowCount(DrillJoinRelBase joinRel, RelMetadataQuery mq,
...
+  }
+
+  private double getPredSelectivityContainingInputRef(RexNode predicate, int inputRef,
+      RelMetadataQuery mq, TableScan scan) {
+    if (predicate instanceof RexCall) {
+      if (predicate.getKind() == SqlKind.AND) {
+        double sel, andSel = 1.0;
+        for (RexNode op : ((RexCall) predicate).getOperands()) {
+          sel = getPredSelectivityContainingInputRef(op, inputRef, mq, scan);
+          andSel *= sel > 0 ? sel : 1.0;
```
amansinha100: `sel > 0` already includes the value 1.0 .. can you clarify this setting ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -148,29 +154,59 @@ private Double getDistinctRowCount(DrillScanRelBase scan, RelMetadataQuery mq, D
...
+    if (((int)PrelUtil.getPlannerSettings(
+        scan.getCluster().getPlanner()).getStatisticsJoinCardinalityMode() & 2) == 2) {
+      for (int i = 0; i < groupKey.length(); i++) {
+        final String colName = type.getFieldNames().get(i);
+        // Skip NDV, if not available
+        if (!groupKey.get(i)) {
+          continue;
+        }
+        ColumnStatistics columnStatistics = tableMetadata != null ? tableMetadata.getColumnStatistics(SchemaPath.getSimplePath(colName)) : null;
+        Double ndv = columnStatistics != null ? (Double) columnStatistics.getStatistic(ColumnStatisticsKind.NVD) : null;
```
amansinha100: `ColumnStatisticsKind.NVD` ==> NDV

amansinha100:
@gparai the phase 1 aggregate related changes and the `TableScan` changes LGTM.  For the join cardinality, I would like to step back and look at the overall set of changes not just from this PR but prior ones to get a better understanding.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
amansinha100: If this is a sanity check borrowed from Calcite, maybe you can add a comment.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
```diff
@@ -667,4 +667,66 @@ public static DrillTable getDrillTable(final TableScan scan) {
...
+                      rightFieldAccess.getIndex() >= leftFieldCount + rightFieldCount) {
+                    joinConditions.clear();
+                    throw new Util.FoundOne(call);
+                  }
+                  /* Both columns reference same table */
+                  if ((leftFieldAccess.getIndex() >= leftFieldCount &&
```
amansinha100: If you have a condition such as t1.a1 = t1.b1, in theory this should not cause problems since the NDV of a1 and b1 is known and we can use the same join cardinality formula.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -208,8 +244,64 @@ public Double getDistinctRowCount(DrillJoinRelBase joinRel, RelMetadataQuery mq,
...
+                if (sidx < left.getRowType().getFieldCount()) {
+                  ndvSGby = mq.getDistinctRowCount(left, ImmutableBitSet.of(sidx), leftPred);
+                } else {
+                  ndvSGby = mq.getDistinctRowCount(right, ImmutableBitSet.of(sidx-left.getRowType().getFieldCount()), rightPred);
+                }
+              }
+            }
+            ++gbyCols;
+            // Multiply NDV(s) of different GBY cols to determine the overall NDV
+            distRowCount *= ndvSGby * plannerSettings.getStatisticsMultiColNdvAdjustmentFactor();
```
amansinha100: The multiplication with the multicolumn factor could be done outside the loop..

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -148,29 +154,59 @@ private Double getDistinctRowCount(DrillScanRelBase scan, RelMetadataQuery mq, D
...
+        if (!groupKey.get(i)) {
+          continue;
+        }
+        ColumnStatistics columnStatistics = tableMetadata != null ? tableMetadata.getColumnStatistics(SchemaPath.getSimplePath(colName)) : null;
+        Double ndv = columnStatistics != null ? (Double) columnStatistics.getStatistic(ColumnStatisticsKind.NVD) : null;
+        if (ndv == null) {
+          continue;
+        }
+        s *= ndv;
+        selectivity = getPredSelectivityContainingInputRef(predicate, i, mq, scan);
```
amansinha100: Add a comment here regarding the fact that combined NDV that was computed previously is getting scaled down by the predicate selectivity.  Also would be useful to consider the example:  'select a, b from t where a = 10 group by a, b';   Here NDV(a) is affected by the predicate, but NDV(b) is not.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdRowCount.java
gparai: Done

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
gparai: It is a sanity check.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
gparai: Simplified it to avoid confusion

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/DrillRelOptUtil.java
gparai: For such conditions we generate a filter condition. For a self-join, it would be two different tables so that would still go fine.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java

gparai:
@amansinha100 I have addressed your review comments. Please take a look.

amansinha100:
@gparai pls address the travis failures.

sohami:
@gparai - With this commit I am seeing plan verification failures for 2 queries in [Functional Tests](http://10.10.104.91:8080/job/Apache_Drill_Functional_Tests/1823/console). After removing this commit from my merge branch the tests are fine. So for now I am removing the ready-to-commit tag from this JIRA. Can you please look into the failures meanwhile ?

PR #1745, opened 2019-04-11
DRILL-7166: Count query with wildcard should skip reading of metadata summary file

dvjyothsna:
Count(*) or Count(column) queries use the aggregated row count and null count from the metadata summary file without reading the large file metadata. When the directory filter has a wildcard, count cannot be fetched from the metadata summary file since the summary file contains count of all the children underneath that and there is no way to filter using wild card. The ConvertCountToDirectScan physical rule will be applied to these cases.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestConvertCountToDirectScan.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestConvertCountToDirectScan.java
@@ -238,4 +238,40 @@ public class TestConvertCountToDirectScan extends PlanTestBase {
       test("drop table if exists %s", tableName);
     }
   }
+
+  @Test
+  public void testCountsWithWildCard() throws Exception {
+    test("use dfs.tmp");
+    String tableName = "parquet_table_counts";
+
+    try {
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/planner/logical/TestConvertCountToDirectScan.java
```diff
@@ -238,4 +238,40 @@ public void testCountsWithMetadataCacheSummaryAndDirPruning() throws Exception {
...
   }
+
+  @Test
+  public void testCountsWithWildCard() throws Exception {
+    test("use dfs.tmp");
+    String tableName = "parquet_table_counts";
+
+    try {
+      for (int i = 0; i < 10; i++) {
+        test(String.format("create table `%s/12/%s` as select * from cp.`tpch/orders.parquet`", tableName, i));
```
amansinha100: Since the test does not actually need a larger table, could you perhaps use a smaller table (like nation) here since there are 13 CTAS statements and doing a CTAS on 'orders' adds extra time to the test.

dvjyothsna:
@amansinha100 I have changed the unit test. Can you please take a look at it

PR #1747, opened 2019-04-12
DRILL-7167: Implemented DESCRIBE TABLE statement

dgrinchenko:
JIRA [DRILL-7167](https://issues.apache.org/jira/browse/DRILL-7167).

- altered parser implementation to honor DESCRIBE TABLE syntax
- extended test coverage to check the new statement

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestInfoSchema.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestInfoSchema.java
@@ -184,6 +184,16 @@ public class TestInfoSchema extends BaseTestQuery {
   }
 
   @Test
+  public void describeTableWithTableKeyword() throws Exception {
+    test("USE INFORMATION_SCHEMA");
+    testBuilder()
+        .sqlQuery("DESCRIBE TABLE CATALOGS")
+        .unOrdered()
+        .sqlBaselineQuery("DESCRIBE CATALOGS")
+        .go();
+  }
+
+  @Test
   public void describeTableWithSchema() throws Exception{
     testBuilder()
         .sqlQuery("DESCRIBE INFORMATION_SCHEMA.`TABLES`")
@@ -193,6 +203,15 @@ public class TestInfoSchema extends BaseTestQuery {
         .baselineValues("TABLE_SCHEMA", "CHARACTER VARYING", "NO")
         .baselineValues("TABLE_NAME", "CHARACTER VARYING", "NO")
         .baselineValues("TABLE_TYPE", "CHARACTER VARYING", "NO")
+        .go();
+  }
+
+  @Test
+  public void describeTableWithSchemaAndTableKeyword() throws Exception {
+    testBuilder()
+        .sqlQuery("DESCRIBE TABLE INFORMATION_SCHEMA.`TABLES`")
+        .unOrdered()
+        .sqlBaselineQuery("DESCRIBE INFORMATION_SCHEMA.`TABLES`")
         .go();
   }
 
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestInfoSchema.java
```diff
@@ -183,6 +183,16 @@ public void describeTable() throws Exception{
         .go();
   }
 
+  @Test
+  public void describeTableWithTableKeyword() throws Exception {
+    testBuilder()
+        .sqlQuery("DESCRIBE TABLE CATALOGS")
+        .unOrdered()
+        .optionSettingQueriesForTestQuery("USE INFORMATION_SCHEMA")
```
arina-ielchiieva: This applies for both sqlQuery and baselineQuery? Correct?

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestInfoSchema.java
```diff
@@ -220,7 +239,29 @@ public void describeWhenSameTableNameExistsInMultipleSchemas() throws Exception{
...
-      test("DROP VIEW dfs.tmp.`TABLES`");
+      test("DROP VIEW IF EXISTS dfs.tmp.`TABLES`");
+    }
+  }
+
+  @Test
+  public void describeWhenSameTableNameExistsInMultipleSchemasWithTableKeyword() throws Exception{
+    try {
+      test("USE dfs.tmp");
+      test("CREATE OR REPLACE VIEW `TABLES` AS SELECT full_name FROM cp.`employee.json`");
```
arina-ielchiieva: Use select from values instead of cp (ex: `select 'A' as c from (values(1))`), this would load only one row which is faster than all rows from `employee.json`.

arina-ielchiieva: Just realized that it's a view creation. You can leave it as is.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestInfoSchema.java
dgrinchenko: right, will add same option for the baseline query.

arina-ielchiieva: Use can add before calling testBuilder: `test("USE INFORMATION_SCHEMA")`

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestInfoSchema.java
```diff
@@ -235,6 +276,16 @@ public void describeTableWithColumnName() throws Exception{
         .go();
   }
 
+  @Test
+  public void describeTableWithColumnNameAndTableKeyword() throws Exception{
+    testBuilder()
+        .sqlQuery("DESCRIBE TABLE `TABLES` TABLE_CATALOG")
+        .unOrdered()
+        .optionSettingQueriesForTestQuery("USE INFORMATION_SCHEMA")
```
arina-ielchiieva: Same here.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestInfoSchema.java
dgrinchenko: gotcha

arina-ielchiieva:
+1, LGTM. Please squash the commits.
@dgrinchenko thanks for making the changes and congratulations with he first PR :)

dgrinchenko:
> @dgrinchenko thanks for making the changes and congratulations with the first PR :)

Thanks!

PR #1749, opened 2019-04-15
DRILL-7177: Format Plugin for Excel Files

cgivre:
This pull request adds the functionality to enable Drill to query Excel files. 

# Excel Format Plugin
This plugin enables Drill to read Microsoft Excel files.  This format is best used with Excel files that do not have extensive formatting, however it will work with formatted files, by allowing you to define a region within the file where the data is.  

The plugin will automatically evaluate cells which contain formulae.  

## Plugin Configuration 
This plugin has several configuration variables which must be set in order to read Excel files effectively.  Since Excel files often contain other elements besides data, you can use the configuration variables to define a region within your spreadsheet in which Drill should extract data.  This is potentially useful if your spreadsheet contains a lot of formatting or other complications. 

* `headerRow`:  Set to -1 if there are no column headers.  
* `lastRow`:  This defines the last row of your data.  The default is an arbitrary large number.  You only will need to set this if you want Drill to stop reading at a specific location.
* `sheetName`:  This is the name of the sheet you want to query.  This will default to the first sheet in the file if left undefined. 
* `firstColumn`:  If you want to define a region within a spreadsheet, this is the left-most column index.  This is indexed from one.  If set to `0` Drill will start at the left most column.
* `lastColumn`:  If you want to define a region within a spreadsheet, this is the right-most column index.  This is indexed from one.  If set to `0` Drill will read all available columns.  This is not inclusive, so if you ask for columns 2-5 you will get columns 2,3 and 4. 

## Usage
You can specify the configuration at runtime via the `table()` method or in the storage plugin configuration.  For instance, if you just want to query an Excel file, you could execute the query as follows:

```
SELECT <fields> 
FROM dfs.`somefile.xlsx`
```

If you wanted to query a different sheet other than the default, use the `table()` method as shown below:
```
SELECT <fields> 
FROM table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'secondSheet'))
```
Theoretically, you could join data together from different sheets as follows:
```
SELECT <fields> 
FROM table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'secondSheet')) AS t1
INNER JOIN table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'thirdSheet')) AS t2 
ON t1.id = t2.id
```


kkhatua:
@cgivre This would be a nice add-on.
Can you rebase your commit and resolve the conflicts?
Also, are there limitations, such as versions, data type detection, etc?

cgivre:
@kkhatua 
This plugin will only work with Excel files from Excel 2007 or greater.  (The XML Excel files).  This does do data type detection, however with limits.  All numeric columns are interpreted as doubles.  
If a column is recognized as a time, the plugin will map those to a Drill date/time column.  
The plugin will execute formulas.

arina-ielchiieva:
@cgivre I suggest new storage or format plugins to be implemented using new row set-based framework implemented by @paul-rogers. Besides better memory management, usage of this framework will allow new plugin / format to use schema provisioning feature. Hopefully, existing plugins will be re-written to use this framework but new should be already implemented using this framework.
Also please note that you need to add `bootstrap-format-plugins.json` as per changes in DRILL-7030.

oleg-zinovev:
Since I have same plugin for internal use (https://github.com/idvp-project/drill-storage-excel):
1) XSSFWorkbook has a awful memory usage. http://apache-poi.1045710.n5.nabble.com/HSSF-and-XSSF-memory-usage-some-numbers-td4312784.html. Reading a 10-15 mb Xlsx file can easily lead to OutOfMemory. 
2) Excel does not guarantee that the column value type remains static. IMHO, it is better to read everything as lines.

cgivre:
Working... @paul-rogers  If you have time, I'd love a code review...

cgivre:
@oleg-zinovev 
In response to your comments, I added a config option to allow the user to have the plugin read everything as `VARCHAR` or they can set it to infer data types.  IMHO, if Drill can do it, it should and if not, there should be a graceful way to fall back.

cgivre:
@vdiravka, @arina-ielchiieva this PR is ready for review with the EVF.   Could someone please take a look?  Thx

arina-ielchiieva:
@cgivre since @paul-rogers is EVF creator, it would be nice if he could his opinion on the changes...

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,287 @@
...
+    TupleMetadata expectedSchema = new SchemaBuilder()
+            .add("id", TypeProtos.MinorType.FLOAT8, TypeProtos.DataMode.OPTIONAL)
+            .add("first_name", TypeProtos.MinorType.VARCHAR, TypeProtos.DataMode.OPTIONAL)
+            .add("last_name", TypeProtos.MinorType.VARCHAR, TypeProtos.DataMode.OPTIONAL)
+            .add("email", TypeProtos.MinorType.VARCHAR, TypeProtos.DataMode.OPTIONAL)
+            .add("gender", TypeProtos.MinorType.VARCHAR, TypeProtos.DataMode.OPTIONAL)
+            .add("birthdate", TypeProtos.MinorType.VARCHAR, TypeProtos.DataMode.OPTIONAL)
+            .add("balance", TypeProtos.MinorType.FLOAT8, TypeProtos.DataMode.OPTIONAL)
+            .add("order_count", TypeProtos.MinorType.FLOAT8, TypeProtos.DataMode.OPTIONAL)
+            .add("average_order", TypeProtos.MinorType.FLOAT8, TypeProtos.DataMode.OPTIONAL)
```
paul-rogers: Simpler: `addNullable(...)` Here and below.

protocol/src/main/java/org/apache/drill/exec/proto/UserBitShared.java
```diff
@@ -980,6 +988,7 @@ public static CoreOperatorType forNumber(int value) {
         case 60: return UNPIVOT_MAPS;
         case 61: return STATISTICS_MERGE;
         case 62: return LTSV_SUB_SCAN;
+        case 64: return EXCEL_SUB_SCAN;
```
paul-rogers: Please double-check if this number has since been taken.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+  private FileSplit split;
+
+  private ResultSetLoader loader;
+
+  private int recordCount;
+
+  public static class ExcelReaderConfig {
+    protected final ExcelFormatPlugin plugin;
+
+    protected int headerRow;
```
paul-rogers: Probably all `final`?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+
+    protected boolean readAllFieldsAsVarChar;
+
+    protected boolean evaluateFormulae;
+
+    protected TupleMetadata schema;
+
+    protected String sheetName;
+
+    public ExcelReaderConfig(ExcelFormatPlugin plugin, int headerRow, int lastRow, int firstColumn, int lastColumn, boolean readAllFieldsAsVarChar, boolean evaluateFormulae,
```
paul-rogers: Since we are passing in the format plugin, we can get the corresponding config and simply set the values from there.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    }
+  }
+
+  public ExcelBatchReader(ExcelReaderConfig readerConfig) {
+    this.readerConfig = readerConfig;
+  }
+
+  @Override
+  public boolean open(FileSchemaNegotiator negotiator) {
+    verifyConfigOptions();
```
paul-rogers: Can this be done once in the plugin rather than each (of possibly many) readers?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    split = negotiator.split();
+    openFile(negotiator);
+    TupleMetadata schema = defineSchema();
```
paul-rogers: Schema not passed to the schema negotiator as an early schema?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    this.loader = negotiator.build();
+    return true;
+  }
+
+  private void openFile(FileScanFramework.FileSchemaNegotiator negotiator) {
+    try {
+      String filePath = split.getPath().toString();
+      this.fsStream = negotiator.fileSystem().open(new Path(filePath));
```
paul-rogers: Do we need to convert from a path to a string to a path? Can we just pass `split.getPath()` directly to `open()`?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    } catch (Exception e) {
+      throw UserException.dataReadError(e).message("Failed to open open input file: %s", split.getPath().toString()).message(e.getMessage()).build(logger);
+    }
+
+    // Evaluate formulae
+    if (readerConfig.evaluateFormulae) {
+      this.evaluator = workbook.getCreationHelper().createFormulaEvaluator();
+    }
+    this.workbook.setMissingCellPolicy(Row.MissingCellPolicy.CREATE_NULL_AS_BLANK);
+    this.sheet = getSheet();
```
paul-rogers: The `this.` prefix not needed unless field has same name as a local variable.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    return getColumnHeaders(builder);
+  }
+
+  protected TupleMetadata getColumnHeaders(SchemaBuilder builder) {
+    //Get the field names
+    int columnCount = 0;
+    if (readerConfig.headerRow >= 0) {
+      columnCount = sheet.getRow(readerConfig.headerRow).getPhysicalNumberOfCells();
+    } else {
+      columnCount = sheet.getRow(0).getPhysicalNumberOfCells();
```
paul-rogers: Nit:

```
    int targetRow = readerConfig.headerRow >= 0 ? readerConfig.headerRow : 0;
    int columnCount = sheet.getRow(targetRow).getPhysicalNumberOfCells();
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    }
+    this.excelFieldNames = new ArrayList<>(columnCount);
+    this.rowIterator = sheet.iterator();
+
+    //If there are no headers, create columns names of field_n
+    if (readerConfig.headerRow == -1) {
+      String missingFieldName;
+      for (int i = 0; i < columnCount; i++) {
+        missingFieldName = MISSING_FIELD_NAME_HEADER + (i + 1);
+        ExcelFormatPlugin.makeColumn(builder, missingFieldName, TypeProtos.MinorType.VARCHAR);
```
paul-rogers: Since `makeColumn()` is only used here, move it to this class?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+            tempColumnName = cell.getStringCellValue()
+              //.replaceAll("_", "__")  // Not sure if we need this
+              .replace(PARSER_WILDCARD, SAFE_WILDCARD).replaceAll("\\.", SAFE_SEPARATOR);
+            ExcelFormatPlugin.makeColumn(builder, tempColumnName, TypeProtos.MinorType.VARCHAR);
+            excelFieldNames.add(colPosition, tempColumnName);
+            break;
+          case NUMERIC:
+            tempColumnName = String.valueOf(cell.getNumericCellValue());
+            ExcelFormatPlugin.makeColumn(builder, tempColumnName, TypeProtos.MinorType.FLOAT8);
+            excelFieldNames.add(colPosition, tempColumnName);
```
paul-rogers: The bove two lines appear the same, save type, in both cases. Pull out as a function?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+      //Read the header row
+      Iterator<Cell> cellIterator = row.cellIterator();
+      int colPosition = 0;
+      String tempColumnName = "";
+
+      while (cellIterator.hasNext()) {
+        Cell cell = cellIterator.next();
+        // TODO Potential NPE if cell evaluation is turned off
+        CellValue cellValue = evaluator.evaluate(cell);
+        switch (cellValue.getCellTypeEnum()) {
```
paul-rogers: Are `STRING` and `NUMERIC` the only two cases? If not, what do we do for the other cases?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+   */
+  private XSSFSheet getSheet() {
+    int sheetIndex = 0;
+    if (!readerConfig.sheetName.isEmpty()) {
+      sheetIndex = workbook.getSheetIndex(readerConfig.sheetName);
+    }
+
+    //If the sheet name is not valid, throw user exception
+    if (sheetIndex == -1) {
+      throw UserException.dataReadError().message("Could not open sheet " + readerConfig.sheetName).build(logger);
```
paul-rogers: `validationError`?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+      }
+      rowWriter.start();
+      for (; colPosition < finalColumn; colPosition++) {
+        Cell cell = row.getCell(colPosition);
+
+        CellValue cellValue = evaluator.evaluate(cell);
+        fieldName = excelFieldNames.get(colPosition);
+
+        // Remove newlines from field names
+        fieldName = fieldName.replaceAll("\\n", SAFE_NEWLINE_REPLACEMENT);
```
paul-rogers: Similar to a comment made in another PR... We have a fixed set of columns in a spreadsheet. Column names must, presumably, be the same for a column in each row in which it appears:

|   | A | B |
 -- | -- | -- |
| 1 | name | balance |
| 2 | fred | 123.45 |
| 3 | barney | 98.76 |

Given this, you can create an array of `ScalarWriter` objects, then index into this array as you read each column. This saves doing he column name edits, and name lookup, on every cell.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+        }
+      }
+      rowWriter.save();
+      recordCount++;
+    }
+    return true;
+  }
+
+  private void writeStringColumn(TupleWriter rowWriter, String name, String value) {
+    int index = rowWriter.tupleSchema().index(name);
```
paul-rogers: Again, no need for the dynamic lookup: the schema can be computed once.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    if (index == -1) {
+      ColumnMetadata colSchema = MetadataUtils.newScalar(name, TypeProtos.MinorType.FLOAT8, TypeProtos.DataMode.OPTIONAL);
+      index = rowWriter.addColumn(colSchema);
+    }
+    ScalarWriter colWriter = rowWriter.scalar(index);
+    colWriter.setDouble(value);
+  }
+
+  private void writeTimestampColumn(TupleWriter rowWriter, String name, Instant ts) {
```
paul-rogers: Never called? No entry for this in the type switch?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,140 @@
...
+
+  public int firstColumn = 0;
+
+  public int lastColumn = 0;
+
+  public boolean readAllFieldsAsVarChar = false;
+
+  public boolean evaluateFormulae = true;
+
+  public String sheetName = "";
```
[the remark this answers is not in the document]
cgivre: I don't think either will pose a problem.  In the case of the empty string, the reader won't find a sheet with that name and throw an error.    If the sheetname is undefined, (`null`) the reader will default to the first sheet in the file. 

https://github.com/apache/drill/blob/0a86d930a13e763e093fad52f3423d22cc4635b2/contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java#L227

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: There are 2.5 data types:  `String`, `Numeric` .  The 0.5 data type are timestamps, which are treated as numeric.

contrib/format-excel/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.apache.poi</groupId>
+      <artifactId>poi-ooxml</artifactId>
+      <version>3.17</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.poi</groupId>
+      <artifactId>poi-ooxml-schemas</artifactId>
+      <version>3.17</version>
```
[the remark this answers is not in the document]
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,140 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import org.apache.drill.shaded.guava.com.google.common.base.Objects;
+import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableList;
+import org.apache.drill.common.logical.FormatPluginConfig;
+import org.apache.drill.exec.store.excel.ExcelBatchReader.ExcelReaderConfig;
+
+import java.util.Arrays;
+import java.util.List;
+
+@JsonTypeName("excel")
```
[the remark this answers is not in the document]
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,140 @@
...
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ExcelFormatConfig implements FormatPluginConfig {
+  //This is the theoretical maximum number of rows in an Excel spreadsheet
+  private final int MAX_ROWS = 1048576;
+
+  public List<String> extensions;
+
+  public int headerRow = 0;
```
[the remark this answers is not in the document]
cgivre: Fixed all...

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatPlugin.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.drill.exec.server.options.OptionManager;
+import org.apache.drill.exec.store.dfs.easy.EasyFormatPlugin;
+import org.apache.drill.exec.store.dfs.easy.EasySubScan;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.drill.exec.store.excel.ExcelBatchReader.ExcelReaderConfig;
+
+
+public class ExcelFormatPlugin extends EasyFormatPlugin<ExcelFormatConfig> {
+
+  private static final String DEFAULT_NAME = "excel";
```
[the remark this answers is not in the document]
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: I now see the issue here... Basically, if Drill is attempting to read the headers, they will be one of the two options: numeric or string.  For a column header, this needs to be coerced into a String.  Excel treats the `timestamp` data type as a `numeric` so for the purpose of getting the column header name, it can be rolled up into the `numeric` data type.

contrib/format-excel/README.md
```diff
@@ -0,0 +1,36 @@
+# Excel Format Plugin
+This plugin enables Drill to read Microsoft Excel files.  This format is best used with Excel files that do not have extensive formatting, however it will work with formatted files, by allowing you to define a region within the file where the data is.
```
arina-ielchiieva: Please remove additional spaces after dots... Not sure what formatting was used but they appear everywhere in .md file.

contrib/format-excel/README.md
```diff
@@ -0,0 +1,36 @@
+# Excel Format Plugin
+This plugin enables Drill to read Microsoft Excel files.  This format is best used with Excel files that do not have extensive formatting, however it will work with formatted files, by allowing you to define a region within the file where the data is.  
+
+The plugin will automatically evaluate cells which contain formulae.  
+
+## Plugin Configuration 
+This plugin has several configuration variables which must be set in order to read Excel files effectively.  Since Excel files often contain other elements besides data, you can use the configuration variables to define a region within your spreadsheet in which Drill should extract data.  This is potentially useful if your spreadsheet contains a lot of formatting or other complications. 
+
+* `headerRow`:  Set to -1 if there are no column headers.  
+* `lastRow`:  This defines the last row of your data.  The default is an arbitrary large number.  You only will need to set this if you want Drill to stop reading at a specific location.
```
arina-ielchiieva: Can you please specify exact number?

arina-ielchiieva: I ma not sure that referring as `you ` is common documentation language style...

contrib/format-excel/README.md
```diff
@@ -0,0 +1,36 @@
...
+* `sheetName`:  This is the name of the sheet you want to query.  This will default to the first sheet in the file if left undefined. 
+* `firstColumn`:  If you want to define a region within a spreadsheet, this is the left-most column index.  This is indexed from one.  If set to `0` Drill will start at the left most column.
+* `lastColumn`:  If you want to define a region within a spreadsheet, this is the right-most column index.  This is indexed from one.  If set to `0` Drill will read all available columns.  This is not inclusive, so if you ask for columns 2-5 you will get columns 2,3 and 4. 
+
+## Usage
+You can specify the configuration at runtime via the `table()` method or in the storage plugin configuration.  For instance, if you just want to query an Excel file, you could execute the query as follows:
```
arina-ielchiieva:
```suggestion
You can specify the configuration at runtime via the `table()` function or in the storage plugin configuration.  For instance, if you just want to query an Excel file, you could execute the query as follows:
```

contrib/format-excel/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+```
+SELECT <fields> 
+FROM dfs.`somefile.xlsx`
+```
+
+If you wanted to query a different sheet other than the default, use the `table()` method as shown below:
```
arina-ielchiieva:
```suggestion
If you wanted to query a different sheet other than the default, use the `table()` function as shown below:
```

contrib/format-excel/README.md
```diff
@@ -0,0 +1,36 @@
...
+```
+SELECT <fields> 
+FROM table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'secondSheet'))
+```
+Theoretically, you could join data together from different sheets as follows:
```
arina-ielchiieva: Why theoretically? Did you try to this? If yes and it works, than please remove theoretically :) If it does not better remove the example.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+import org.joda.time.Instant;
+
+import java.util.Iterator;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class ExcelBatchReader implements ManagedReader<FileSchemaNegotiator> {
+  private ExcelReaderConfig readerConfig;
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(ExcelBatchReader.class);
```
arina-ielchiieva: Imports...

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    protected boolean readAllFieldsAsVarChar;
+
+    protected boolean evaluateFormulae;
+
+    protected TupleMetadata schema;
+
+    protected String sheetName;
+
+    public ExcelReaderConfig(ExcelFormatPlugin plugin, int headerRow, int lastRow, int firstColumn, int lastColumn, boolean readAllFieldsAsVarChar, boolean evaluateFormulae,
+                             //TupleMetadata schema,
```
arina-ielchiieva: Remove comment.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    return true;
+  }
+
+  private void openFile(FileScanFramework.FileSchemaNegotiator negotiator) {
+    try {
+      String filePath = split.getPath().toString();
+      this.fsStream = negotiator.fileSystem().open(new Path(filePath));
+      this.workbook = new XSSFWorkbook(this.fsStream.getWrappedStream());
+    } catch (Exception e) {
+      throw UserException.dataReadError(e).message("Failed to open open input file: %s", split.getPath().toString()).message(e.getMessage()).build(logger);
```
arina-ielchiieva: Please write chaining calls from the new line.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    this.workbook.setMissingCellPolicy(Row.MissingCellPolicy.CREATE_NULL_AS_BLANK);
+    this.sheet = getSheet();
+  }
+
+  /**
+   * This helper function gets the column headers.
+   * If the user specified no columns headers it populates an array with field names of field_n
+   */
+  private TupleMetadata defineSchema() {
+    // TODO Use case if there are no column headers
```
arina-ielchiieva: Please remove todo-s in production code.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+
+    // Evaluate formulae
+    if (readerConfig.evaluateFormulae) {
+      this.evaluator = workbook.getCreationHelper().createFormulaEvaluator();
+    }
+    this.workbook.setMissingCellPolicy(Row.MissingCellPolicy.CREATE_NULL_AS_BLANK);
+    this.sheet = getSheet();
+  }
+
+  /**
```
arina-ielchiieva: Use proper java-doc with `@return`

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
arina-ielchiieva: Can this read compressed file? For example, `test.xlsx.gz`?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+
+      Row row = rowIterator.next();
+      // If the user specified that there are no headers, get the column count
+      if (readerConfig.headerRow == -1 && recordCount == 0) {
+        this.totalColumnCount = row.getLastCellNum();
+      }
+
+      String fieldName;
+      if (row.getLastCellNum() < totalColumnCount) {
+        throw UserException.dataReadError().message("Wrong number of columns in row: %d", row.getLastCellNum()).build(logger);
```
arina-ielchiieva: Same here

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+
+  private static final String SAFE_SEPARATOR = "_";
+
+  private static final String PARSER_WILDCARD = ".*";
+
+  private static final String MISSING_FIELD_NAME_HEADER = "field_";
+
+  private static final String SAFE_NEWLINE_REPLACEMENT = " ";
+
+  private XSSFSheet sheet;
```
arina-ielchiieva: can some of the fields be final?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+    }
+  }
+
+  @Override
+  public void close() {
+    if (workbook != null) {
+      try {
+        workbook.close();
+      } catch (IOException e) {
+        logger.warn(e.getMessage());
```
arina-ielchiieva: Please add context to the erorr message: `logger.warn("Error when closing XSSFWorkbook resource: {}, e.getMessage())"`

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,398 @@
...
+      }
+      workbook = null;
+    }
+
+    if (fsStream != null) {
+      try {
+        fsStream.close();
+      } catch (IOException e) {
+        logger.warn(e.getMessage());
```
arina-ielchiieva: Same here.

paul-rogers:
One additional thought: how are column types handled? Do we require all fields to have the same types within a column?

Name | Balance
----| -----
Fred | 123.45
Barney | 556.78

If so, then it might make sense to hold an array of column handler objects, like in the earlier version of the Regex plugin, that holds the column accessor and performs any required type conversions. This would be cleaner/faster than doing a `switch` per column.

Can a column type vary?

Name | Balance
----| -----
Fred | 123.45
Barney | "Not sure"

If so, then the code has to handle conversions. The above would generate an error, but would we allow:

Name | Balance
----| -----
Fred | 123.45
Barney | "556.78"

Can a column start as empty (null) so that we have to defer type selection?

Name | Balance
----| -----
Fred | 
Barney | 556.78

If so, then we need a way to defer column type selection until we see the first non-null value. The (not yet merged) new JSON reader hands the deferred-type case; I can share that logic if you need it.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: See comment below..... The long and the short of it is that we don't really know the schema until we read the first row of data.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: If the sheet has no column header, Drill assigns names of `field_n` for each column.  This is verified with this test:
https://github.com/apache/drill/blob/b9d502108eb8b5e199c8855c793ea34f7537aa1c/contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java#L234-252

cgivre: If the sheet has no column header, Drill assigns names of `field_n` for each column.  This is verified with this test:
https://github.com/apache/drill/blob/b9d502108eb8b5e199c8855c793ea34f7537aa1c/contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java#L234-L252

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,444 @@
...
+    loader = negotiator.build();
+    rowWriter = loader.writer();
+    openFile(negotiator);
+    defineSchema();
+    return true;
+  }
+
+  private void openFile(FileScanFramework.FileSchemaNegotiator negotiator) {
+    try {
+      fsStream = negotiator.fileSystem().open(split.getPath());
```
arina-ielchiieva: Since format is compressible `openPossiblyCompressedStream` should be used instead of `open`.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,444 @@
...
+
+  private static final String HEADER_NEW_LINE_REPLACEMENT = "__";
+
+  private static final String MISSING_FIELD_NAME_HEADER = "field_";
+
+  private XSSFSheet sheet;
+
+  private XSSFWorkbook workbook;
+
+  private FSDataInputStream fsStream;
```
arina-ielchiieva:
```suggestion
  private InputStream fsStream;
```

cgivre: Fixed

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,274 @@
...
+    try {
+      run(sql);
+      fail();
+    } catch (DrillRuntimeException e) {
+      assertTrue(e.getMessage().contains("Could not open sheet "));
+    }
+  }
+
+  @Test
+  public void testDefineColumnsQuery() throws RpcException {
```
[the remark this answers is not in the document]
cgivre: I think I got all these cases.

contrib/format-excel/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+SELECT <fields> 
+FROM table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'secondSheet')) AS t1
+INNER JOIN table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'thirdSheet')) AS t2 
+ON t1.id = t2.id
+```
+
+### Known Limitations:
+At present, Drill requires that all columns be of the same data type. If they are not, Drill will throw an exception upon trying to read a column of mixed data type. If you are
+ trying to query data with heterogenoeus columns, it will be necessary to set `readAllColumnsAsVarChar` to `true`.
```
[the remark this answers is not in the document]
cgivre: Ok...

contrib/format-excel/README.md
```diff
@@ -0,0 +1,54 @@
...
+SELECT <fields> 
+FROM table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'secondSheet')) AS t1
+INNER JOIN table( dfs.`test_data.xlsx` (type => 'excel', sheetName => 'thirdSheet')) AS t2 
+ON t1.id = t2.id
+```
+
+### Known Limitations:
+At present, Drill requires that all columns be of the same data type. If they are not, Drill will throw an exception upon trying to read a column of mixed data type. If you are
+ trying to query data with heterogenoeus columns, it will be necessary to set `readAllColumnsAsVarChar` to `true`.  
+
```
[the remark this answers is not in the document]
cgivre: No..   Column headers are not required and if they are not present, Drill will assign names of `field_n`.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,444 @@
...
+        CellValue cellValue = evaluator.evaluate(cell);
+        if (cellValue == null) {
+          String fieldValue = "";
+          if (firstLine) {
+            addColumnToArray(rowWriter, excelFieldNames.get(colPosition), MinorType.VARCHAR);
+          }
+          columnWriters.get(colWriterIndex).setString(fieldValue);
+
+        } else {
+          switch (cellValue.getCellTypeEnum()) {
```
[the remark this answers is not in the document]
cgivre: Wouldn't this still require some switch statement to load the data?  I see where you're going with this, but you still need to have a switch statement to even get the cell value as the ways of retrieving it are different.

cgivre:
@paul-rogers @arina-ielchiieva I addressed the review comments with the exception of one which I had a question about.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: Fixed and added a unit test

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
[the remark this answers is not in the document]
cgivre: Hi @paul-rogers 
I see where you're going with this, however, there are a few issues and I'm not sure that it's going to be possible to remove the switch statement.  The main issue is that the column names are created by reading the header line.  This is done here:
https://github.com/apache/drill/blob/a23ca66bc5da2e89b3957ad0b3cabb8c4ebdfcc4/contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java#L166-L237.

However, the actual data types aren't assigned until the reader actually reads the first line of data which occurs here:
https://github.com/apache/drill/blob/a23ca66bc5da2e89b3957ad0b3cabb8c4ebdfcc4/contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java#L320-#L330

There has to be additional logic in the second part so that the reader can add the columns to the schema (see below) and that requires the `switch` statement because the ways of accessing the different columns from the Excel sheet are different.

https://github.com/apache/drill/blob/a23ca66bc5da2e89b3957ad0b3cabb8c4ebdfcc4/contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java#L332-L356

If the columns were already defined, what you are describing makes sense, but I'm not seeing how to do it with getting rid of the `switch` statement.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+import org.apache.drill.exec.store.dfs.easy.EasySubScan;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.drill.exec.store.excel.ExcelBatchReader.ExcelReaderConfig;
+import org.slf4j.Logger;
+
+
+public class ExcelFormatPlugin extends EasyFormatPlugin<ExcelFormatConfig> {
+
+  protected static final String DEFAULT_NAME = "excel";
+  private static final Logger logger = org.slf4j.LoggerFactory.getLogger(ExcelFormatPlugin.class);
```
arina-ielchiieva: Please use import org.slf4j.LoggerFactory

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+      throw UserException
+        .validationError()
+        .message("Invalid column configuration.  The first column index is greater than the last column index.")
+        .build(logger);
+    }
+
+    if (readerConfig.firstColumn < 0) {
+      throw UserException
+        .validationError()
+        .message("Invalid value for first column.  Index must be greater than zero.")
```
arina-ielchiieva:
```suggestion
        .message("Invalid value for first column. Index must be greater than zero.")
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+   * </ul>
+   *
+   * @param readerConfig
+   */
+  private void verifyConfigOptions(ExcelReaderConfig readerConfig) {
+    // Validate the config variables
+    if ((readerConfig.lastColumn < readerConfig.firstColumn) && readerConfig.lastColumn != 0) {
+      throw UserException
+        .validationError()
+        .message("Invalid column configuration.  The first column index is greater than the last column index.")
```
arina-ielchiieva:
```suggestion
        .message("Invalid column configuration. The first column index is greater than the last column index.")
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+      throw UserException
+        .validationError()
+        .message("Invalid value for first column.  Index must be greater than zero.")
+        .build(logger);
+    }
+
+    if (readerConfig.lastColumn < 0) {
+      throw UserException
+        .validationError()
+        .message("Invalid value for last column.  Index must be greater than zero.")
```
arina-ielchiieva:
```suggestion
        .message("Invalid value for last column. Index must be greater than zero.")
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+        .build(logger);
+    }
+
+    if (readerConfig.headerRow > readerConfig.lastRow) {
+      throw UserException
+        .validationError()
+        .message("Invalid value for headerRow.  Header row must be less than last row.")
```
arina-ielchiieva:
```suggestion
        .message("Invalid value for headerRow. Header row must be less than last row.")
```

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,383 @@
...
+  public static void setup() throws Exception {
+    ClusterTest.startCluster(ClusterFixture.builder(dirTestWatcher));
+    definePlugin();
+  }
+
+  private static void definePlugin() throws ExecutionSetupException {
+    ExcelFormatConfig sampleConfig = new ExcelFormatConfig();
+
+    // Define a temporary plugin for the "cp" storage plugin.
+    Drillbit drillbit = cluster.drillbit();
```
arina-ielchiieva: There is a handy method to define format out-of-box, please see https://github.com/apache/drill/blob/master/exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestTextWriter.java#L250 as example.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,383 @@
...
+    final StoragePluginRegistry pluginRegistry = drillbit.getContext().getStorage();
+    final FileSystemPlugin plugin = (FileSystemPlugin) pluginRegistry.getPlugin("cp");
+    final FileSystemConfig pluginConfig = (FileSystemConfig) plugin.getConfig();
+    pluginConfig.getFormats().put("sample", sampleConfig);
+    pluginRegistry.createOrUpdate("cp", pluginConfig, false);
+  }
+
+  @Test
+  public void testStarQuery() throws Exception {
+    String sql = "SELECT * FROM cp.`excel/test_data.xlsx` LIMIT 5";
```
arina-ielchiieva: If you are using limit how do you know that first 5 rows will ones you expect? Limit does not guarantee it would be first 5 rows. It's better to test all data set or reduce file size to exactly 5 rows.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,383 @@
...
+    testBuilder()
+      .sqlQuery(sql)
+      .ordered().baselineColumns("col1", "col2")
+      .baselineValues("1.0", "Bob")
+      .baselineValues("2.0", "Steve")
+      .baselineValues("3.0", "Anne")
+      .baselineValues("Bob", "3.0")
+      .go();
+  }
+
```
arina-ielchiieva: Please remove extra new lines.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,383 @@
...
+import org.junit.experimental.categories.Category;
+import org.apache.drill.categories.RowSetTests;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@Category(RowSetTests.class)
+public class TestExcelFormat extends ClusterTest {
+
+  @ClassRule
```
arina-ielchiieva: No need to define `dirTestWatcher`, it already exists.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatPlugin.java
```diff
@@ -0,0 +1,141 @@
...
+   * This function verifies that the user entered valid user configuration options.  Specifically it verifies that:
+   * <ul>
+   * <li>the lastColumn is greater than the first column</li>
+   * <li>The lastColumn is not zero</li>
+   * <li>firstColumn is greater than zero</li>
+   * <li>lastColumn is greater than zero</li>
+   * <li>The headerRow index is less than the lastRow index</li>
+   * </ul>
+   *
+   * @param readerConfig
```
arina-ielchiieva: please add description to avoid warning in the IDE.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public int getLastColumn() {
+    return lastColumn;
+  }
+
+  public boolean getReadAllFieldsAsVarChar() {
+    return allTextMode;
+  }
+
+  public void setHeaderRow(int row) {
```
arina-ielchiieva: Do you need setters?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,137 @@
...
+import org.apache.drill.shaded.guava.com.google.common.collect.ImmutableList;
+import org.apache.drill.common.logical.FormatPluginConfig;
+import org.apache.drill.exec.store.excel.ExcelBatchReader.ExcelReaderConfig;
+
+import java.util.Arrays;
+import java.util.List;
+
+@JsonTypeName(ExcelFormatPlugin.DEFAULT_NAME)
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ExcelFormatConfig implements FormatPluginConfig {
```
arina-ielchiieva: Please add test to check serialization and deserialization. First you need to generate physical plan and then submit it. Example:

```
  @Test
  public void testSerDe() throws Exception {
    String sql = "select count(*) as cnt from cp.`tpch/nation.parquet`";
    String plan = queryBuilder().sql(sql).explainJson();
    long cnt = queryBuilder().physical(plan).singletonLong();
    assertEquals("Counts should match",25L, cnt);
  }
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,526 @@
...
+    }
+  }
+
+  @Override
+  public void close() {
+    if (workbook != null) {
+      try {
+        workbook.close();
+      } catch (IOException e) {
+        logger.warn("Error when closing XSSFWorkbook resource: {}, e.getMessage())");
```
arina-ielchiieva:
```suggestion
        logger.warn("Error when closing XSSFWorkbook resource: {}", e.getMessage()));
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,526 @@
...
+      }
+      workbook = null;
+    }
+
+    if (fsStream != null) {
+      try {
+        fsStream.close();
+      } catch (IOException e) {
+        logger.warn("Error when closing XSSFWorkbook resource: {}, e.getMessage())");
```
arina-ielchiieva:
```suggestion
        logger.warn("Error when closing XSSFWorkbook resource: {}", e.getMessage()));
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,526 @@
...
+    } else if (type == MinorType.FLOAT8) {
+      cellWriterArray.add(new NumericCellWriter(columnWriters.get(index)));
+    } else if (type == MinorType.TIMESTAMP) {
+      cellWriterArray.add(new TimestampCellWriter(columnWriters.get(index)));
+    }
+  }
+
+  /**
+   * Returns the index of the first row of actual data.  This function is to be used to find the header row as the POI skips blank rows.
+   * @return:  The headerRow index, corrected for blank rows
```
arina-ielchiieva:
```suggestion
   * @return The headerRow index, corrected for blank rows
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,526 @@
...
+    } else {
+      CellType cellType = cellValue.getCellTypeEnum();
+      if (cellType == CellType.STRING || readerConfig.readAllFieldsAsVarChar) {
+        addColumnToArray(rowWriter, excelFieldNames.get(colPosition), MinorType.VARCHAR);
+      } else if (cellType == CellType.NUMERIC && DateUtil.isCellDateFormatted(cell)) {
+        addColumnToArray(rowWriter, excelFieldNames.get(colPosition), MinorType.TIMESTAMP);
+      } else if (cellType == CellType.NUMERIC) {
+        addColumnToArray(rowWriter, excelFieldNames.get(colPosition), MinorType.FLOAT8);
+      } else {
+        logger.warn("Unknown data type.");
```
arina-ielchiieva: Include cell type into want message...
If type is not supported which type it would have?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,526 @@
...
+  private boolean firstLine;
+
+  private FileSplit split;
+
+  private ResultSetLoader loader;
+
+  private int recordCount;
+
+  public static class ExcelReaderConfig {
+    final ExcelFormatPlugin plugin;
```
arina-ielchiieva: Can files be private? Or they used elsewhere?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -0,0 +1,526 @@
...
+import org.joda.time.Instant;
+import org.slf4j.Logger;
+
+import java.io.InputStream;
+import java.util.Iterator;
+import java.io.IOException;
+import java.util.ArrayList;
+
+public class ExcelBatchReader implements ManagedReader<FileSchemaNegotiator> {
+  private ExcelReaderConfig readerConfig;
```
arina-ielchiieva: final?

contrib/format-excel/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+      <version>${poi.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.poi</groupId>
+      <artifactId>ooxml-schemas</artifactId>
+      <version>1.4</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-compress</artifactId>
```
arina-ielchiieva: What for this library is used?

cgivre: Good question... removed

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: Shortened file to 5 rows

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: There are really only two data types, `NUMERIC` and `STRING`.   I don't think this case will ever happen.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: That's good to know.  Will use next time.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,360 @@
...
+      .build();
+
+    new RowSetComparison(expected).verifyAndClearAll(results);
+  }
+
+  /**
+   * This tests to what happens if you attempt to query a completely empty sheet
+   * The result is an empty data set.
+   *
+   * @throws RpcException
```
arina-ielchiieva: add description to avoid warning in IDE

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
arina-ielchiieva: Why not use it now? It's just one line of code....

```
cluster.defineFormat("cp", "sample", sampleConfig);
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,117 @@
...
+  public int firstColumn;
+
+  public int lastColumn;
+
+  public boolean allTextMode;
+
+  public String sheetName = "";
+
+
+  public ExcelFormatConfig() {
```
arina-ielchiieva: There are not other constructors thus no need to add default one...

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,360 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+@Category(RowSetTests.class)
+public class TestExcelFormat extends ClusterTest {
+
+  private static FileSystem fs;
+
+  private static CompressionCodecFactory factory;
```
arina-ielchiieva: Looks like test for compression is missing. Plus there is no need to define codec factory on class level if it will be used in one test...

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: I believe that I made all fields which can be final final

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: Done..

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: Added test.

cgivre:
Thanks @paul-rogers and @arina-ielchiieva for your review.  Commits squashed.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,406 @@
...
+      .build();
+
+    new RowSetComparison(expected).verifyAndClearAll(results);
+  }
+
+  private void generateCompressedFile(String fileName, String codecName, String outFileName) throws IOException {
+    CompressionCodec codec = factory.getCodecByName(codecName);
+    assertNotNull(codecName + " is not found", codec);
+
+    System.out.println();
```
arina-ielchiieva: Please remove.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,406 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
+
+@Category(RowSetTests.class)
+public class TestExcelFormat extends ClusterTest {
+
+  private static FileSystem fs;
```
arina-ielchiieva: Please remove fs and factory from class level, since they are used only in one test.

contrib/format-excel/README.md
```diff
@@ -0,0 +1,59 @@
...
+      "type": "excel"
+    }
+```
+The plugin has many other configuration options listed below:
+
+* `headerRow`:  Set to `-1` if there are no column headers.  Defaults to `0`. If the data does not have a header row, Drill will assign column names of `field_n` for each column
+. If the sheet starts with a series of empty rows, Drill will ignore these empty rows, so there is no need to set the `headerRow` in that case.
+* `lastRow`:  This defines the last row of data. The default is 1048576 which is the theoretical row limit for Excel files. It is only necessary to set this if you want Drill to
+ stop reading at a specific location.
+* `sheetName`:  This is the name of the sheet which Drill will query. This will default to the first sheet in the file if left undefined.  Drill will throw an exception if the
```
arina-ielchiieva: @cgivre I still see two spaces between sentences, please fix.

arina-ielchiieva:
@cgivre also please create Jira for `more efficient way to read each cell` which @paul-rogers has mentioned and add details on what should be done. Thanks.

cgivre:
@arina-ielchiieva I created https://issues.apache.org/jira/browse/DRILL-7423 to address future improvements.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: Oops.  Done

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: Done

contrib/format-excel/README.md
cgivre: Sorry about that.  This is a long standing habit I have when writing.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,403 @@
...
+
+  @BeforeClass
+  public static void setup() throws Exception {
+    ClusterTest.startCluster(ClusterFixture.builder(dirTestWatcher));
+
+    ExcelFormatConfig formatConfig  = new ExcelFormatConfig();
+    cluster.defineFormat("cp", "excel", formatConfig);
+
+    // Needed for compressed file unit test
+    dirTestWatcher.copyResourceToRoot( Paths.get("excel/"));
```
arina-ielchiieva:
```suggestion
    dirTestWatcher.copyResourceToRoot(Paths.get("excel/"));
```

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -0,0 +1,403 @@
...
+import static org.junit.Assert.fail;
+
+@Category(RowSetTests.class)
+public class TestExcelFormat extends ClusterTest {
+
+  @BeforeClass
+  public static void setup() throws Exception {
+    ClusterTest.startCluster(ClusterFixture.builder(dirTestWatcher));
+
+    ExcelFormatConfig formatConfig  = new ExcelFormatConfig();
```
arina-ielchiieva:
```suggestion
    ExcelFormatConfig formatConfig = new ExcelFormatConfig();
```

protocol/src/main/java/org/apache/drill/exec/proto/UserBitShared.java
```diff
@@ -901,6 +905,10 @@ private FragmentState(int value) {
      * <code>LTSV_SUB_SCAN = 62;</code>
      */
     public static final int LTSV_SUB_SCAN_VALUE = 62;
+    /**
+     * <code>EXCEL_SUB_SCAN = 64;</code>
+     */
+    public static final int EXCEL_SUB_SCAN_VALUE = 64;
```
arina-ielchiieva: @cgivre did you generate C++ code, protobuf generating instruction, please check step 3:
```
If changes are made to the DrillClient's protobuf, you would need to regenerate the sources for the C++ client as well.
Steps for regenerating the sources are available https://github.com/apache/drill/blob/master/contrib/native/client/
```
https://github.com/apache/drill/tree/master/protocol

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,113 @@
...
+@JsonTypeName(ExcelFormatPlugin.DEFAULT_NAME)
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ExcelFormatConfig implements FormatPluginConfig {
+
+  private static final List<String> DEFAULT_EXTS = ImmutableList.of("xlsx");
+
+  // This is the theoretical maximum number of rows in an Excel spreadsheet
+  private final int MAX_ROWS = 1048576;
+
+  public List<String> extensions;
```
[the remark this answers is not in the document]
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,113 @@
...
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
+    if (obj == null || getClass() != obj.getClass()) {
+      return false;
+    }
+    ExcelFormatConfig other = (ExcelFormatConfig) obj;
+    return Objects.equal(headerRow, other.headerRow)
```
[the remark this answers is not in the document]
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.drill.exec.store.excel.ExcelBatchReader.ExcelReaderConfig;
+
+import java.util.Arrays;
+import java.util.List;
+
+@JsonTypeName(ExcelFormatPlugin.DEFAULT_NAME)
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ExcelFormatConfig implements FormatPluginConfig {
+
+  private static final List<String> DEFAULT_EXTS = ImmutableList.of("xlsx");
```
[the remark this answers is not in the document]
cgivre: Fixed.  Is `ImmutableList` ok to leave?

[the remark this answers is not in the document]
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+@JsonTypeName(ExcelFormatPlugin.DEFAULT_NAME)
+@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+public class ExcelFormatConfig implements FormatPluginConfig {
+
+  private static final List<String> DEFAULT_EXTS = Arrays.asList("xlsx");
```
arina-ielchiieva: Please remove.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
```diff
@@ -0,0 +1,109 @@
...
+
+  // This is the theoretical maximum number of rows in an Excel spreadsheet
+  private final int MAX_ROWS = 1048576;
+
+  public List<String> extensions = DEFAULT_EXTS;
```
arina-ielchiieva:
```suggestion
  public List<String> extensions = Collections.singletonList("xlsx");
```

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
cgivre: Done

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelFormatConfig.java
cgivre: Done

arina-ielchiieva:
+1, LGTM

arina-ielchiieva:
I'll run full test suit and merge the PR soon.

cgivre:
Thanks @arina-ielchiieva and @paul-rogers for your review and help with this plugin!

PR #1750, opened 2019-04-15
DRILL-2362: Profile Mgmt

kkhatua:
This PR is a WIP for managing a large number of profiles. It involves the following features.

1. Write profiles to indexed partitions (created on the fly, and default being organized in nested directories by year, month and date).
2. Read chronologically from the above partitioned dirs. This improves performance by scanning and retrieving only from the most recent profiles
3. Leverage Guava Cache by saving on cost of deserializing a profile multiple times from the disk. (Even 1 attempt at rendering a profile leads to atleast 2 times deserialization).
4. Infer which partitioned dir has a profile based on queryId alone. This means that rather than scanning all the directories, we reverse engineer the query ID to figure out the approximate start time of the query to narrow down on the profile's location.
5. Trace Exception [qId: 259432dc-7f8e-8fc5-af69-16a1ca817689 ] -> This is a sample bad profile and make the UI more robust in handling bad profiles that cant be deserialized
6. Auto Index for 1st time (In batches of 10000) from root dir (sync if Distributed). Using ZK, synchronization is maintained when multiple Drillbits are sharing the same profile location


kkhatua:
@arina-ielchiieva I'm looking for a suggestion on how to manage existing profiles. 
Currently, during startup, we automatically index any profiles in the root of the `profiles` dir into their respective location. As a default, I have set it at 1000, and the Drillbits syncrhonize if the profile dir is on a DFS.
However, I am not sure if we should automatically index any new profile that has been copy-pasted into the root dir. For e.g., we might get a profile from a JIRA and would like to view it. Should we leave it there (and try to render it) or should we index it ASAP ?

kkhatua:
@arina-ielchiieva  could you please review this PR ?

jnturton:
Hi Kunal! Thank you for this contribution, I'd like to help to move it forward. 

> However, I am not sure if we should automatically index any new profile that has been copy-pasted into the root dir. For e.g., we might get a profile from a JIRA and would like to view it. Should we leave it there (and try to render it) or should we index it ASAP ?

I share your concern here.  I think we should consider having Drill only write _new_ profiles to partitioned directories. Any partitioning of historical profiles can be done externally by admins, in my opinion, and we can add examples of "housekeeping" scripts for doing that to the Drill documentation.

Would you like to do any of the following?

- Resume the work on this PR, I volunteer to be a reviewer.
- Receive a PR from me to your fork here that's rebased and has some changes I think we want.
- You're too busy now, so I'll pull your commits here into a new branch of my own and open a new PR.

Thanks
James

jnturton:
Or, if we do want to keep this built in ability to partition existing profiles, perhaps we should have it launched from a button on the Profiles page in the web UI instead of on Drillbit startup? That would remove the complication of which Drillbit does the work and the worries of slowing down startup or partitioning profiles that nobody wanted partitioned.

PR #1751, opened 2019-04-16
DRILL-7164: KafkaFilterPushdownTest is sometimes failing to pattern match correctly

aravi5:
- The structure of "cost" in JSON plan changed as a part of #1677 and cost validation was accidentally.
- While correcting the cost validation, the substring validation is changed to regex validation to make it more robust.

### DIFF
--- a/contrib/storage-kafka/src/test/java/org/apache/drill/exec/store/kafka/KafkaFilterPushdownTest.java
+++ b/contrib/storage-kafka/src/test/java/org/apache/drill/exec/store/kafka/KafkaFilterPushdownTest.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.store.kafka;
 
+import org.apache.drill.PlanTestBase;
 import org.apache.drill.categories.KafkaStorageTest;
 import org.apache.drill.categories.SlowTest;
 import org.apache.kafka.common.serialization.StringSerializer;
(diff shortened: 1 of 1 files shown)


aravi5:
@HanumathRao, @sohami - Can you review the PR?

contrib/storage-kafka/src/test/java/org/apache/drill/exec/store/kafka/KafkaFilterPushdownTest.java
```diff
@@ -63,7 +62,9 @@ public void testPushdownOnOffset() throws Exception {
         TestQueryConstants.JSON_PUSHDOWN_TOPIC, predicate1, predicate2);
 
     runKafkaSQLVerifyCount(queryString, expectedRowCount);
-    testPhysicalPlan(queryString, String.format(expectedSubStr, expectedRowCount));
+    PlanTestBase.testPlanWithAttributesMatchingPatterns(queryString,
```
sohami: also since you moved to using `testPlanWithAttributesMatchingPatterns` from `testPhysicalPlan` now the `rowCount` value is different than `expectedRowCount`. Is that intended ?

aravi5:
@sohami has the proposed changes ready that addresses his review comments and will open another PR for it

PR #1752, opened 2019-04-17
DRILL-7182: Incorrect access specifier for Join DrillRelMdDistinctRowcount

gparai:
@amansinha100 can you please review this PR? Thanks!

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
@@ -83,7 +83,7 @@ public class DrillRelMdDistinctRowCount extends RelMdDistinctRowCount{
     if (rel instanceof DrillScanRelBase) {                  // Applies to both Drill Logical and Physical Rels
       if (!DrillRelOptUtil.guessRows(rel)) {
         DrillTable table = Utilities.getDrillTable(rel.getTable());
-        return getDistinctRowCount(((DrillScanRelBase) rel), mq, table, groupKey, rel.getRowType(), predicate);
+        return getDistinctRowCountInternal(((DrillScanRelBase) rel), mq, table, groupKey, rel.getRowType(), predicate);
       } else {
         /* If we are not using statistics OR there is no table or metadata (stats) table associated with scan,
          * estimate the distinct row count. Consistent with the estimation of Aggregate row count in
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -174,7 +174,7 @@ private Double getDistinctRowCount(DrillScanRelBase scan, RelMetadataQuery mq, D
     }
   }
 
-  public Double getDistinctRowCount(DrillJoinRelBase joinRel, RelMetadataQuery mq, ImmutableBitSet groupKey,
+  private Double getDistinctRowCount(DrillJoinRelBase joinRel, RelMetadataQuery mq, ImmutableBitSet groupKey,
```
amansinha100: Wouldn't it be better to rename this (and the other private methods here) with an 'internal' suffix to avoid potential issues in the future ?  Only the Override methods remain unchanged.   What do you think ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -92,7 +92,7 @@ public Double getDistinctRowCount(RelNode rel, RelMetadataQuery mq, ImmutableBit
         if (rel instanceof DrillScanRel) {
           // The existing Drill behavior is to only use this estimation for DrillScanRel and not ScanPrel.
           // TODO: We may potentially do it for ScanPrel (outside the scope of statistics)
-          return rel.estimateRowCount(mq) * 0.1;
+          return ((DrillScanRel)rel).estimateRowCount(mq) * 0.1;
```
vvysotskyi: This cast is not needed, please revert this change.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
gparai: Done.

gparai:
@amansinha100 / @vvysotskyi I have addressed your review comments. Please take a look.

vvysotskyi:
+1

PR #1757, opened 2019-04-19
DRILL-7186: Add back storage.json REST endpoint (regression after DRILL-6562)

arina-ielchiieva:
Details - [DRILL-7186](https://issues.apache.org/jira/browse/DRILL-7186).


arina-ielchiieva:
@vdiravka please review.

PR #1762, opened 2019-04-22
DRILL-7191 Distributed state persistence and Integration of Distributed queue configuration with Planner

HanumathRao:
This PR contains changes for the support of RM Framework both on execution and planning side.
1) Refactoring existing ZK based queue to accommodate new Distributed queue for RM. Moved QueryResourceAllocators memory allocation code to utility classes like ZKQueueMemoryAllocationUtilities and DefaultMemoryAllocationUtilities. Refactored the Parallelizer code to accommodate the memory adjustment for the operators during parallelization phase. There are 3 different implementation of SimpleParallelizer such as ZKQueueParallelizer, DistributedQueueParallelizer and DefaultParallelizer which will be used by ZK based RM, Distributed RM and Non RM configuration.

2) Planner integration with RM to select queue and reduce query level memory to be within queue limits. Changes to handle scenarios where buffered operator are at least getting minimum required memory allocation. Based on the calculated memory for each operator within each fragment it’s initial and maximum memory allocation is set which is later consumed by execution layer to enforce memory limits.

3) Introduced new DrillNode class to deal with issues when DrillbitEndpoint is searched in a map using some of it’s field.

4) Changes to support storing UUID for each Drillbit Service Instance locally to be used by planner and execution layer. This UUID is used to uniquely identify a Drillbit and register Drillbit information in the RM StateBlobs. Introduced a PersistentStore named ZookeeperTransactionalPersistenceStore with Transactional capabilities using Zookeeper Transactional API’s. This is used for updating RM State blobs as all the updates need to happen in transactional manner. Added RMStateBlobs definition and support for serde to Zookeeper. Implementation for DistributedRM and its corresponding QueryRM apis.

5) Updated the state management of Query in Foreman so that same Foreman object can be submitted multiple times. Also introduced concept of 2 maps keeping track of waiting and running queries. These were done to support for async admit protocol which will be needed with Distributed RM.

6) Support for serde of optimalMemoryAllocation for each operator in each minor fragment in QueryProfile. This is needed to verify the optimalMemory calculated by planner is correct.


exec/java-exec/src/main/java/org/apache/drill/exec/planner/fragment/ZKQueueParallelizer.java
```diff
@@ -0,0 +1,120 @@
...
+    public Map<String, Collection<PhysicalOperator>> getNodeMap() {
+      Map<DrillbitEndpoint, Collection<PhysicalOperator>> endpointCollectionMap = bufferedOperators.asMap();
+      Map<String, Collection<PhysicalOperator>> nodeMap = new HashMap<>();
+      for (Map.Entry<DrillbitEndpoint, Collection<PhysicalOperator>> entry : endpointCollectionMap.entrySet()) {
+        nodeMap.put(entry.getKey().getAddress(), entry.getValue());
+      }
+
+      return nodeMap;
+    }
+  }
```
sohami: Couple of things:
1) Why not construct `Map<String, Collection<PhysicalOperator>>` to start with instead of `Multimap<DrillbitEndpoint, PhysicalOperator> `and then converting it.
2) `bufferedOperators.asMap()` is not guaranteed to work in the same way as intended since DrillbitEndpoint `equals` and `hashCode` methods are not reliable which is why we created `DrillNode`.

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZKClusterCoordinator.java
```diff
@@ -249,14 +253,19 @@ public RegistrationHandle update(RegistrationHandle handle, State state) {
...
   public Collection<DrillbitEndpoint> getOnlineEndPoints() {
-    Collection<DrillbitEndpoint> runningEndPoints = new ArrayList<>();
-    for (DrillbitEndpoint endpoint: endpoints){
-      if(isDrillbitInState(endpoint, State.ONLINE)) {
-        runningEndPoints.add(endpoint);
+    return getOnlineEndpointsUUID().keySet();
+  }
+
+  @Override
+  public Map<DrillbitEndpoint, String> getOnlineEndpointsUUID() {
```
HanumathRao: Looks like this is duplicate code similar to that of the LocalClusterCoordinator. Can you move this function into a common place and use it in both the places?

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZookeeperClient.java
```diff
@@ -227,13 +230,33 @@ public void create(final String path) {
...
-      curator.create().withMode(mode).forPath(target);
+      curator.create().creatingParentsIfNeeded().withMode(mode).forPath(target);
       getCache().rebuildNode(target);
     } catch (final Exception e) {
       throw new DrillRuntimeException("unable to put ", e);
     }
   }
 
+  public void createAsTransaction(List<String> paths) {
+    Preconditions.checkNotNull(paths, "no paths provided to create");
```
HanumathRao: Is it also better to check for empty list of paths? Also it might be good to add a comment for this function.

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZookeeperClient.java
```diff
@@ -301,6 +324,54 @@ public void put(final String path, final byte[] data, DataChangeVersion version)
...
+    putAsTransaction(pathsWithData, null);
+  }
+
+  /**
+   * Puts the given sets of blob and their data's in a transactional manner. It expects all the blob path to exist
+   * before calling this api.
+   * @param pathsWithData - map of blob paths to update and the final data
+   * @param version - version holder
+   */
+  public void putAsTransaction(Map<String, byte[]> pathsWithData, DataChangeVersion version) {
```
HanumathRao: Do we currently use non null version. If not then can you please mention it in the comment that this is needed for future use.

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/rmblobmgr/RMConsistentBlobStoreManager.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+  }
+
+  @Override
+  public void reserveResources(Map<String, NodeResources> queryResourceAssignment,
+                               QueryQueueConfig selectedQueue, String leaderId,
+                               String queryId, String foremanNode) throws Exception {
+    // Looks like leader hasn't changed yet so let's try to reserve the resources
+    // See if the call is to reserve or free up resources
+    Map<String, NodeResources> resourcesMap = queryResourceAssignment;
```
HanumathRao: Can this be changed to the following code.
    Map<String, NodeResources> resourcesMap =  queryResourceAssignment.entrySet().stream()
      .collect(Collectors.toMap(Map.Entry::getKey,
        (x) -> new NodeResources(x.getValue().getVersion(),
                                 -x.getValue().getMemoryInBytes(),
                                 -x.getValue().getNumVirtualCpu())));

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/rmblobmgr/RMConsistentBlobStoreManager.java
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * RM state blobs manager which does all the update to the blobs under a global lock and in transactional manner.
+ * Since the blobs are updated by multiple Drillbit at same time to maintain the strongly consistent information in
+ * these blobs it uses a global lock shared across all the Drillbits.
+ */
+public class RMConsistentBlobStoreManager implements RMBlobStoreManager {
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(RMConsistentBlobStoreManager.class);
+
+  private static final String RM_BLOBS_ROOT = "rm/blobs";
```
HanumathRao: do we need to start this path with "/" ?

exec/java-exec/src/main/java/org/apache/drill/exec/coord/zk/ZookeeperClient.java
```diff
@@ -301,6 +324,54 @@ public void put(final String path, final byte[] data, DataChangeVersion version)
...
+        if (version != null) {
+          transaction = transaction.setData().withVersion(version.getVersion()).forPath(target, entry.getValue()).and();
+        } else {
+          transaction = transaction.setData().forPath(target, entry.getValue()).and();
+        }
+        targetPaths.add(target);
+        totalDataBytes += entry.getValue().length;
+      }
+
+      // If total set operator payload is greater than 1MB then curator set operation will fail
```
HanumathRao: greater than or equal to?

exec/java-exec/src/main/java/org/apache/drill/exec/resourcemgr/rmblobmgr/RMConsistentBlobStoreManager.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    try {
+      updateBlobs(queryResourceAssignment, selectedQueue, leaderId, queryId, foremanNode);
+    } catch (Exception ex) {
+      logger.error("Failed to update the blobs", ex);
+      throw ex;
+    } finally {
+      // Check if the caller has acquired the mutex
+      if (globalBlobMutex.isAcquiredInThisProcess()) {
```
HanumathRao: why is this check required. Shouldn't be the case that we should acquire the lock if we are here?

PR #1764, opened 2019-04-24
DRILL-7206: using primitive int list at right or full join case

weijietong:
This change is to avoid the unneeded box ,unbox operation.


arina-ielchiieva:
There is no link to Jira (PR name and commit message are invalid).

weijietong:
@arina-ielchiieva thanks for the tip.

arina-ielchiieva:
@Ben-Zvi could you please review?

PR #1765, opened 2019-04-24
DRILL-7202: Failed query shows warning that fragments made no progress

kkhatua:
This bug arises because of the `toggleWarning()` javascript function that shows the warning if the number of slow major fragments matches the number of major fragments.
For failed queries, the number of major fragments = 0; and that matches the number of tags (which is also zero for a non-running query). Hence the warning shows.


arina-ielchiieva:
+1

PR #1770, opened 2019-04-26
DRILL-7050: RexNode convert exception in sub-query

vvysotskyi:
- Bumped up Calcite version to include a fix for CALCITE-2954;
- Added a call to validate query when `inferUnknownTypes()` is called to replace temporary table name in sub-queries from the project list;
- Added unit tests for both issues.

For problem description please see [DRILL-7050](https://issues.apache.org/jira/browse/DRILL-7050).


arina-ielchiieva:

PR #1771, opened 2019-04-26
DRILL-7199: Optimize population of metadata for non-interesting columns

dvjyothsna:
Currently the non-interesting column metadata is populated for all types of metadata including rowgroup metadata. It's a huge overkill if there are large number of row groups. With this PR, non-interesting column metadata is populated only once when all the other types of metadata is populated in the BaseParquetMetadataProvider.java. This optimization reduced the planning time from 17 sec to 5 sec when there are 35000 row groups.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
@@ -51,6 +51,7 @@ import org.apache.drill.metastore.ColumnStatistics;
 import org.apache.drill.metastore.ColumnStatisticsKind;
 import org.apache.drill.metastore.FileMetadata;
 import org.apache.drill.metastore.LocationProvider;
+import org.apache.drill.metastore.NonInterestingColumnsMetadata;
 import org.apache.drill.metastore.PartitionMetadata;
 import org.apache.drill.metastore.TableMetadata;
 import org.apache.drill.metastore.TableStatisticsKind;
@@ -86,6 +87,7 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
   // partition metadata info: mixed partition values for all partition keys in the same list
   protected List<PartitionMetadata> partitions;
 
+  protected NonInterestingColumnsMetadata nonInterestingColumnsMetadata;
   protected List<SchemaPath> partitionColumns;
   protected LogicalExpression filter;
   protected List<SchemaPath> columns;
@@ -115,7 +117,7 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
     this.partitionColumns = that.partitionColumns;
     this.partitions = that.partitions;
     this.files = that.files;
-
+    this.nonInterestingColumnsMetadata = that.nonInterestingColumnsMetadata;
     this.fileSet = that.fileSet == null ? null : new HashSet<>(that.fileSet);
   }
 
@@ -151,18 +153,29 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
    */
   @Override
   public long getColumnValueCount(SchemaPath column) {
-    long tableRowCount = (long) TableStatisticsKind.ROW_COUNT.getValue(getTableMetadata());
+    long tableRowCount, colNulls;
+    Long nulls;
     ColumnStatistics columnStats = getTableMetadata().getColumnStatistics(column);
-    long colNulls;
-    if (columnStats != null) {
-      Long nulls = (Long) columnStats.getStatistic(ColumnStatisticsKind.NULLS_COUNT);
-      colNulls = nulls != null ? nulls : Statistic.NO_COLUMN_STATS;
-    } else {
-      return 0;
+    ColumnStatistics nonInterestingColStats = null;
+    if (columnStats == null) {
+      nonInterestingColStats = getNonInterestingColumnsMetadata().getColumnStatistics(column);
     }
+
+    if (columnStats != null) {
+      tableRowCount = (long) TableStatisticsKind.ROW_COUNT.getValue(getTableMetadata());
+    } else if (nonInterestingColStats != null) {
+      tableRowCount = (long) TableStatisticsKind.ROW_COUNT.getValue(getNonInterestingColumnsMetadata());
+    } else {
+      return 0; // returns 0 if the column doesn't exist in the table.
+    }
+
+    columnStats = columnStats != null ? columnStats : nonInterestingColStats;
+    nulls = (Long) columnStats.getStatistic(ColumnStatisticsKind.NULLS_COUNT);
+    colNulls = nulls != null ? nulls : Statistic.NO_COLUMN_STATS;
+
     return Statistic.NO_COLUMN_STATS == tableRowCount
-        || Statistic.NO_COLUMN_STATS == colNulls
-        ? Statistic.NO_COLUMN_STATS : tableRowCount - colNulls;
+            || Statistic.NO_COLUMN_STATS == colNulls
+            ? Statistic.NO_COLUMN_STATS : tableRowCount - colNulls;
   }
 
   @Override
@@ -266,6 +279,7 @@ public abstract class AbstractGroupScanWithMetadata extends AbstractFileGroupSca
       filteredMetadata.withTable(getTableMetadata())
           .withPartitions(getNextOrEmpty(getPartitionsMetadata()))
           .withFiles(filesMap)
+          .withNonInterestingColumns(getNonInterestingColumnsMetadata())
           .withMatching(false);
     }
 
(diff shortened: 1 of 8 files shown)


dvjyothsna:
@amansinha100 , @vvysotskyi Please review.

dvjyothsna:
@amansinha100 Can you please take a look at this

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
```diff
@@ -286,25 +287,27 @@ public static PartitionMetadata getPartitionMetadata(SchemaPath partitionColumn,
...
-      for (Metadata_V4.ColumnTypeMetadata_v4 columnTypeMetadata :
-          ((Metadata_V4.ParquetTableMetadata_v4) parquetTableMetadata).getColumnTypeInfoMap().values()) {
-        SchemaPath schemaPath = SchemaPath.getCompoundPath(columnTypeMetadata.name);
-        if (!schemaPaths.contains(schemaPath)) {
+      ConcurrentHashMap<Metadata_V4.ColumnTypeMetadata_v4.Key, Metadata_V4.ColumnTypeMetadata_v4> columnTypeInfoMap =
+              ((Metadata_V4.ParquetTableMetadata_v4) parquetTableMetadata).getColumnTypeInfoMap();
+
+      if (columnTypeInfoMap == null) {
+        return new NonInterestingColumnsMetadata(columnsStatistics);
+      } // in some cases for runtime pruning
```
amansinha100: It's not clear what this comment means ..

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
```diff
@@ -87,6 +88,7 @@
   private TableMetadata tableMetadata;
   private List<PartitionMetadata> partitions;
   private Map<Path, FileMetadata> files;
+  private NonInterestingColumnsMetadata nonInterestingColumnsMetadata;
```
amansinha100: For older metadata versions (V3.3 and prior) will this be initialized to null  ?  Have you tried with prior versions ?

exec/java-exec/src/main/java/org/apache/drill/metastore/NonInterestingColumnsMetadata.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.drill.exec.record.metadata.ColumnMetadata;
+import org.apache.drill.exec.record.metadata.TupleMetadata;
+import java.util.Map;
+
+/**
+ * Represents a metadata for the non-interesting columns. Since the refresh command doesn't store the non-interesting
+ * columns stats in teh cache file, there is a need to mark column statistics of non-interesting as unknown to
+ * differentiate the non-interesting columns from non-existent columns.
+ */
+public class NonInterestingColumnsMetadata implements BaseMetadata {
```
amansinha100: This class currently stores ColumnStatistics but for other things it returns NULL because it is implementing `BaseMetadata` which is a much broader interface.  If the other types of metadata are not applicable, you should clarify in the comments.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
```diff
@@ -146,15 +148,23 @@ public boolean isMatchAllMetadata() {
...
+    Long nulls;
     ColumnStatistics columnStats = getTableMetadata().getColumnStatistics(column);
-    long colNulls;
+    ColumnStatistics nonInterestingColStats = getNonInterestingColumnsMetadata().getColumnStatistics(column);
+
     if (columnStats != null) {
-      Long nulls = (Long) columnStats.getStatistic(ColumnStatisticsKind.NULLS_COUNT);
-      colNulls = nulls != null ? nulls : GroupScan.NO_COLUMN_STATS;
+      tableRowCount = (long) TableStatisticsKind.ROW_COUNT.getValue(getTableMetadata());
+    } else if (nonInterestingColStats != null) {
```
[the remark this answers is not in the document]
dvjyothsna: Yes, tablemetadata holds only the interesting column stats. ColumnStats will be non-null only for the interesting columns. There are three scenarios here:
 
- Column is interesting. Then the columnStats is not null and it has the correct row count.
- Column is non-interesting. Then the columnStats will be null and nonInterestingColStats is not null and the row count is marked as unknown.
- Column doesn't exist in the table. So both columnStats and nonInterestingColStats will be null. Since the column doesn't exist, the row count is set to 0.

Here we know that a column will only belong to one of the above cases. Thats the reason there is an else-if instead of if.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
dvjyothsna: @Ben-Zvi  Added this comment. I'm assuming columnTypeInfoMap will be null while populating metadata during run-time pruning and non-interesting columns don't make sense.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
dvjyothsna: Column statistics will be an empty map in case of older metadata versions. This is handled by checking if parquetTableMetadata is an instance of Metadata_V4.ParquetTableMetadata_v4.

dvjyothsna: Some of the unit tests actually use the metadata cache files from older versions.

exec/java-exec/src/main/java/org/apache/drill/metastore/NonInterestingColumnsMetadata.java
```diff
@@ -0,0 +1,72 @@
...
+package org.apache.drill.metastore;
+
+import org.apache.drill.common.expression.SchemaPath;
+import org.apache.drill.exec.record.metadata.ColumnMetadata;
+import org.apache.drill.exec.record.metadata.TupleMetadata;
+import java.util.Map;
+
+/**
+ * Represents a metadata for the non-interesting columns. Since the refresh command doesn't store the non-interesting
+ * columns stats in teh cache file, there is a need to mark column statistics of non-interesting as unknown to
```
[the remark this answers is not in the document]
dvjyothsna: Changed the spelling.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/FilterEvaluatorUtils.java
```diff
@@ -62,7 +63,14 @@ public static RowsMatch evalFilter(LogicalExpression expr, MetadataBase.ParquetT
             expr.<Set<SchemaPath>, Void, RuntimeException>accept(new FieldReferenceFinder(), null));
 
     RowGroupMetadata rowGroupMetadata = new ArrayList<>(ParquetTableMetadataUtils.getRowGroupsMetadata(footer).values()).get(rowGroupIndex);
+    NonInterestingColumnsMetadata nonInterestingColumnsMetadata = ParquetTableMetadataUtils.getNonInterestingColumnsMeta(footer);
     Map<SchemaPath, ColumnStatistics> columnsStatistics = rowGroupMetadata.getColumnsStatistics();
+
+    // Add column statistis of non-interesting columns if there are any
```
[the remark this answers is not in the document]
dvjyothsna: Changed the spelling.

exec/java-exec/src/main/java/org/apache/drill/metastore/NonInterestingColumnsMetadata.java
dvjyothsna: Added a comment to clarify that NonInterestingColumnsMetadata is not applicable to some of the methods.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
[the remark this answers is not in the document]
amansinha100: Since only one of `columnStats` (btw, let's call it `interestingColStats` to be consistent) or `nonInterestingColStats` will be non-null, we should ideally invoke `getNonInterestingColumnsMetadata()` only if the `columnStats` is null...otherwise the second call is redundant.  So, something like: 
  `ColumnStatistics interestingColStats = getTableMetadata().getColumnStatistics(column);`
   `ColumnStatistics nonInterestingColStats = null;`
    `if (interestingColStats == null) {`
        `nonInterestingColStats = getNonInterestingColumnsMetadata().getColumnStatistics(column);`
      `}`
Also, for the last `else` case, add a comment that the `return 0` is for non-existing columns.

amansinha100:
Pls squash before final commit.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/AbstractGroupScanWithMetadata.java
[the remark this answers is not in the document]
dvjyothsna: Changed and added a comment

dvjyothsna:
@amansinha100 I have addressed your comments. Can you please take a look

dvjyothsna:
Rebased on master

PR #1772, opened 2019-04-29
DRILL-7187: Improve selectivity estimation of BETWEEN predicates and …

amansinha100:
…arbitrary combination of range predicates.

- Also, propagate the totalRowCount to the histogram selectivity estimation and use it instead of the nonNullCount. 
- Before and after estimates for the following predicate: 
    `where o.o_orderdate >= date '1996-10-01' and o.o_orderdate < date '1996-10-01' + interval '3' month`
BEFORE this PR:  estimated filter row count = **3206**
AFTER this PR:     estimated filter row count = **601**
ACTUAL row count                                           = **561**

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/NumericEquiDepthHistogram.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/NumericEquiDepthHistogram.java
@@ -21,15 +21,17 @@ package org.apache.drill.exec.planner.common;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 
+import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexLiteral;
 import com.tdunning.math.stats.MergingDigest;
-import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
 import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
+import org.apache.drill.shaded.guava.com.google.common.collect.Range;
 
 /**
  * A column specific equi-depth histogram which is meant for numeric data types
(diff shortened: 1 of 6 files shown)


amansinha100:

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdSelectivity.java
```diff
@@ -356,8 +426,8 @@ private boolean isMultiColumnPredicate(final RexNode node) {
     return findAllRexInputRefs(node).size() > 1;
   }
 
-  private static List<RexInputRef> findAllRexInputRefs(final RexNode node) {
-      List<RexInputRef> rexRefs = new ArrayList<>();
+  private static Set<RexInputRef> findAllRexInputRefs(final RexNode node) {
```
gparai: Would this not break the existing logic? For a predicate like $0=$0 using the `Set` would cause the `isMultiColumnPredicate` function to return false.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestAnalyze.java
```diff
@@ -480,6 +490,24 @@ public void testHistogramWithColumnsWithAllNulls() throws Exception {
...
+  @Test
+  public void testHistogramWithBetweenPredicate() throws Exception {
+    try {
+      test("ALTER SESSION SET `planner.slice_target` = 1");
+      test("ALTER SESSION SET `store.format` = 'parquet'");
+      test("create table dfs.tmp.orders2 as select * from cp.`tpch/orders.parquet`");
+      test("analyze table dfs.tmp.orders2 compute statistics");
+      test("alter session set `planner.statistics.use` = true");
+
+      String query = "select 1 from dfs.tmp.orders2 o where o.o_orderdate >= date '1996-10-01' and o.o_orderdate < date '1996-10-01' + interval '3' month";
```
gparai: I do not see the `BETWEEN` clause in this query.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/NumericEquiDepthHistogram.java
```diff
@@ -76,129 +73,188 @@ public NumericEquiDepthHistogram(int numBuckets) {
...
+    long numRows = 0;
+    int result;
+    Double lowValue = null;
+    Double highValue = null;
+    final int first = 0;
+    final int last = buckets.length - 1;
+    int startBucket = first;
+    int endBucket = last;
+
+    if (range.hasLowerBound()) {
```
gparai: There is code duplication for lowValue and highValue except the inverted comparisons. Maybe we can refactor the common portion out?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdSelectivity.java
amansinha100: Yes, thanks for pointing that out, even though that predicate $0=$0 is unexpected (something to investigate in future).  I have reverted this change and it returns a List as before.  Instead, now where the original call to `isMultiColumnPredicate()` happens I added a second condition (line 182) that ensures that conditions of type `$1 > 10 AND $1 < 20` which are created after calling `preProcessRangeConditions()` are not considered the same as multicolumn predicates.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestAnalyze.java
amansinha100: I changed the test name to use 'Interval' instead of 'Between' since that's the main purpose of this test.  There are 2 other tests I added which do use `BETWEEN` clause in `testHistogramWithDataTypes1`.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/common/NumericEquiDepthHistogram.java
amansinha100: There are some key differences in the comparisons and the the bucket ranges that make it difficult to consolidate and even if I was to do it it would look somewhat contrived.  Let me know if you have something specific in mind.  For now I have left it as is.

amansinha100:
@gparai I have addressed your review comments.  Pls take another look.

PR #1775, opened 2019-04-30
DRILL-7227: Fix predicate check in DrillRelOptUtil.analyzeSimpleEquiJoin

gparai:
@amansinha100 can you please review the changes in commit f893ffb? Thanks!

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
@@ -55,8 +55,12 @@ import org.apache.drill.exec.util.Utilities;
 import org.apache.drill.metastore.ColumnStatistics;
 import org.apache.drill.metastore.ColumnStatisticsKind;
 import org.apache.drill.metastore.TableMetadata;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class DrillRelMdDistinctRowCount extends RelMdDistinctRowCount{
+  private static final Logger logger = LoggerFactory.getLogger(DrillRelMdDistinctRowCount.class);
+
   private static final DrillRelMdDistinctRowCount INSTANCE =
       new DrillRelMdDistinctRowCount();
 
@@ -142,10 +146,7 @@ public class DrillRelMdDistinctRowCount extends RelMdDistinctRowCount{
     if (groupKey.length() == 0) {
       return selectivity * rowCount;
     }
-    /* If predicate is present, determine its selectivity to estimate filtered rows. Thereafter,
-     * compute the number of distinct rows
-     */
-    selectivity = mq.getSelectivity(scan, predicate);
+
     TableMetadata tableMetadata;
     try {
       tableMetadata = table.getGroupScan().getTableMetadata();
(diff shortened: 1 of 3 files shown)


gparai:
@amansinha100 could you please review the changes for DRILL-7231. I have added those changes as part of this PR to minimize open PRs.

amansinha100:
@gparai could you pls add some details in the DRILL-7231 about the nature of the issue ?  That will give a context for the proposed fix.

gparai:
@amansinha100 I have added the description of the issue in the JIRA. Please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -155,18 +155,18 @@ private Double getDistinctRowCountInternal(TableScan scan, RelMetadataQuery mq,
...
-        allCols = false;
-        break;
+        continue;
       }
       ColumnStatistics columnStatistics = tableMetadata != null ?
           tableMetadata.getColumnStatistics(SchemaPath.getSimplePath(colName)) : null;
       Double ndv = columnStatistics != null ? (Double) columnStatistics.getStatistic(ColumnStatisticsKind.NDV) : null;
+      // Skip NDV, if not available
       if (ndv == null) {
+        allColsHaveNDV = false;
```
amansinha100: If any one column out of potentially many group-by columns does not have NDV stats, the logic here is that it will continue checking all the columns and compute the selectivity but then eventually on line 185 we return a default rowcount of 10%.  Why not break here if we cannot make use of the other columns's NDV ?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -180,7 +180,7 @@ private Double getDistinctRowCountInternal(TableScan scan, RelMetadataQuery mq,
       }
     }
     s = Math.min(s, rowCount);
-    if (!allCols) {
+    if (s == 1.0 && !allColsHaveNDV) {
```
amansinha100: Its not clear why the check for `s == 1.0` is needed.  
Also, I see that the `else` part is returning the selectivity value whereas the `if` part is returning a row count.  The method is expected to return a rowcount, not selectivity...unless I am missing something.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
```diff
@@ -155,18 +155,18 @@ private Double getDistinctRowCountInternal(TableScan scan, RelMetadataQuery mq,
     }
 
     double s = 1.0;
-    boolean allCols = true;
+    boolean allColsHaveNDV = true;
```
amansinha100: Unrelated to this line but on previous lines 145-148 seem to be duplicates of lines 136-139.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
gparai: `s` is rowcount not selectivity.

amansinha100: Ah,  yes..I somehow mistook that for selectivity.  What about the check for `s == 1.0` ? Even if that value is higher but at least 1 column does not have NDV stats, I would think we want to return the 10% default.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
gparai: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
gparai: Renamed the variable to avoid confusion

gparai:
@amansinha100 I have addressed your review comments. Please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/cost/DrillRelMdDistinctRowCount.java
amansinha100: can you mark this done if this comment is addressed .. seems like the duplicate lines are there.

PR #1776, opened 2019-04-30
DRILL-7223: Create an option to control timeout for REFRESH METADATA

Ben-Zvi:
The new option is passed to the TimedCallable code via the existing ParquetReaderConfig parameter, by adding a Long field there - timeoutPerRunnableInMsec .

The timeout (TIMEOUT_PER_RUNNABLE_IN_MSECS) is used there for two purposes:
(1) Issue a warning when exceeded.
(2) Time-out (i.e. throw TimeoutException ?) the running thread.
Seems that the second one does not work; maybe the last change (DRILL-6281) is the reason.


exec/java-exec/src/main/java/org/apache/drill/exec/store/TimedCallable.java
```diff
@@ -188,6 +187,23 @@ private long getExecutionTime(TimeUnit unit) {
...
+   * @param logger The logger to use to report results.
+   * @param tasks List of callable that should be executed and timed.  If this list has one item, task will be
+   *                  completed in-thread. Each callable must handle {@link InterruptedException}s.
+   * @param parallelism  The number of threads that should be run to complete this task.
+   * @param timeout if bigger than zero, set the timeout per runnable (in msec)
+   * @return The list of outcome objects.
+   * @throws IOException All exceptions are coerced to IOException since this was build for storage system tasks initially.
+   */
+  public static <V> List<V> run(final String activity, final Logger logger, final List<TimedCallable<V>> tasks, int parallelism, long timeout) throws IOException {
+    TIMEOUT_PER_RUNNABLE_IN_MSECS = timeout;
```
amansinha100: By convention we treat these as static final constants (although I see that the `final` is missing).  Also assigning to this static variable is not thread safe. Actually, now that this is a config option, why do we need a static constant ?  Why not just pass it as a parameter wherever needed including passing it to the constructor of classes that are extending the abstract class and then we can set the timeout as a local class variable ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderConfig.java
```diff
@@ -50,6 +50,7 @@
   private boolean enableTimeReadCounter = false;
   private boolean autoCorrectCorruptedDates = true;
   private boolean enableStringsSignedMinMax = false;
+  private long timeoutPerRunnableInMsec = 15_000;
```
amansinha100: Since the 'owner' of the default should be ExecConstants, we should avoid this default value.

exec/java-exec/src/main/resources/drill-module.conf
```diff
@@ -629,6 +629,7 @@ drill.exec.options: {
     store.parquet.reader.columnreader.async: false,
     store.parquet.reader.int96_as_timestamp: false,
     store.parquet.reader.strings_signed_min_max: "",
+    store.parquet.refresh_timeout_per_runnable_in_msec: 15000,
```
amansinha100: See prior comment about omitting 'refresh' .  Also I think that we should not put this in the 'parquet' namespace even though currently it is mainly used for Parquet metadata read.  The reason is the runnable timeout could in theory be used for other formats also.  How about `store.timeout_per_runnable_in_msec`  ?

PR #1777, opened 2019-05-01
DRILL-6965: Implement schema table function parameter

arina-ielchiieva:
1. Added common schema table function parameter with can be used as single unit or with format plugin table function parameters.
2. Allowed creating schema without columns, in case if user needs only to indicate table properties.
3. Added unit tests.

Details in Jira [DRILL-6965](https://issues.apache.org/jira/browse/DRILL-6965).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/schema/SchemaProviderFactory.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/schema/SchemaProviderFactory.java
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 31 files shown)


arina-ielchiieva:
@vvysotskyi please review.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/schema/SchemaProviderFactory.java
```diff
@@ -0,0 +1,87 @@
...
+    ProviderType providerType = ProviderType.valueOf(split[0].trim().toUpperCase());
+    String value = split[1].trim();
+    switch (providerType) {
+      case INLINE:
+        return new InlineSchemaProvider(value);
+      case PATH:
+        char c = value.charAt(0);
+        // if path starts with any type of quotes, strip them and remove escape characters if any
+        if (c == '\'' || c == '"' || c == '`') {
+          value = value.substring(1, value.length() - 1).replaceAll("\\\\(.)", "$1");
```
vvysotskyi: I'm not sure that `replaceAll("\\\\(.)", "$1")` is needed. For the case, if `value` is not escaped and windows path is specified, it may cause problems.

exec/java-exec/src/main/java/org/apache/drill/exec/store/table/function/TableSignature.java
```diff
@@ -0,0 +1,98 @@
...
+  private final String name;
+  private final List<TableParamDef> commonParams;
+  private final List<TableParamDef> specificParams;
+  private final List<TableParamDef> params;
+
+  public static TableSignature of(String name) {
+    return new TableSignature(name, Collections.emptyList(), Collections.emptyList());
+  }
+
+  public static TableSignature of(String name, List<TableParamDef> primaryParams) {
```
vvysotskyi: Could you please rename arguments here and in the method below?

arina-ielchiieva:
@vvysotskyi thanks for the code review. Besides requested changes, converted `SchemaParsingException` into `IOException` to ensure it's always handled in the code. Also for schema parameter added schema validation before executing the query.

PR #1779, opened 2019-05-02
DRILL-7222: Visualize estimated and actual row counts for a query

kkhatua:
With statistics in place, it is useful to have the estimated rowcount along side the actual rowcount query profile's operator overview. 
We extract this from the Physical Plan section of the profile and show it in parenthesis with a toggle button to have the estimates hidden by default.
image

### DIFF
--- a/exec/java-exec/src/main/resources/rest/profile/profile.ftl
+++ b/exec/java-exec/src/main/resources/rest/profile/profile.ftl
@@ -42,7 +42,7 @@
     };
 
     $(document).ready(function() {
-      $(".sortable").DataTable( {
+      $(".sortable").DataTable({
         "searching": false,
         "lengthChange": false,
         "paging": false,
@@ -84,6 +84,29 @@
         document.getElementById(warningElemId).style.display="none";
     }
 
+    //Injects Estimated Rows
+    function injectEstimatedRows() {
+      Object.keys(opRowCountMap).forEach(key => {
+        var tgtElem = $("td.estRowsAnchor[key='" + key + "']"); 
+        var status = tgtElem.append("<div class='estRows' title='Estimated'>(" + opRowCountMap[key] + ")</div>");
+      });
+    }
+
+    //Toggle Estimates' visibility
+    function toggleEstimates(tgtColumn) {
+      var colClass = '.est' + tgtColumn;
+      var estimates = $(colClass);
+      if (estimates.filter(":visible").length > 0) {
+        $(colClass).each(function () {
+          $(this).attr("style", "display:none");
+        });
+      } else {
+        $(colClass).each(function () {
+          $(this).attr("style", "display:block");
+        });
+      }
+    }
+
     //Close the cancellation status popup
     function refreshStatus() {
       //Close PopUp Modal
@@ -94,7 +117,7 @@
     //Cancel query & show cancellation status
     function cancelQuery() {
       document.getElementById("cancelTitle").innerHTML = "Drillbit on " + location.hostname + " says";
-      $.get("/profiles/cancel/"+globalconfig.queryid, function(data, status){/*Not Tracking Response*/});
+      $.get("/profiles/cancel/" + globalconfig.queryid, function(data, status){/*Not Tracking Response*/});
       //Show PopUp Modal
       $("#queryCancelModal").modal("show");
     };
@@ -389,8 +412,17 @@
   </div>
 
   <div class="page-header"></div>
-  <h3>Operator Profiles</h3>
+  <h3>Operator Profiles
+ <button onclick="toggleEstimates('Rows')" class="btn" style="font-size:60%; float:right">Show/Hide Estimated Rows</button></h3>
 
+ <style>
+  .estRows {
+    color:navy;
+    font-style:italic;
+    font-size: 80%;
+    display:<#if model.showEstimatedRows()>block<#else>none</#if>;
+  }
+</style>
   <div class="panel-group" id="operator-accordion">
     <div class="panel panel-default">
       <div class="panel-heading">
@@ -476,6 +508,9 @@
       injectIconByClass("spill-tag","glyphicon-download-alt");
       injectIconByClass("time-skew-tag","glyphicon-time");
       injectSlowScanIcon();
+      //Building RowCount
+      buildRowCountMap();
+      injectEstimatedRows();
     });
 
     //Inject Glyphicon by Class tag
@@ -485,7 +520,7 @@
         var i;
         for (i = 0; i < tagElemList.length; i++) {
             var content = tagElemList[i].innerHTML;
-            tagElemList[i].innerHTML = "<span class=\"glyphicon "+tagIcon+"\">&nbsp;</span>"+content;
+            tagElemList[i].innerHTML = "<span class=\"glyphicon " + tagIcon + "\">&nbsp;</span>" + content;
         }
     }
 
@@ -496,7 +531,7 @@
         var i;
         for (i = 0; i < tagElemList.length; i++) {
             var content = tagElemList[i].innerHTML;
-            tagElemList[i].innerHTML = "<img src='/static/img/turtle.png' alt='slow'> "+content;
+            tagElemList[i].innerHTML = "<img src='/static/img/turtle.png' alt='slow'> " + content;
         }
     }
 
@@ -567,7 +602,7 @@
     var popUpAndPrintPlan = function() {
       var srcSvg = $('#query-visual-canvas');
       var screenRatio=0.9;
-      let printWindow = window.open('', 'PlanPrint', 'width=' + (screenRatio*screen.width) + ',height=' + (screenRatio*screen.height) );
+      let printWindow = window.open('', 'PlanPrint', 'width=' + (screenRatio*screen.width) + ',height=' + (screenRatio*screen.height));
       printWindow.document.writeln($(srcSvg).parent().html());
       printWindow.print();
     };
@@ -587,6 +622,21 @@
       if (e.target.form) 
         <#if model.isOnlyImpersonationEnabled()>doSubmitQueryWithUserName()<#else>doSubmitQueryWithAutoLimit()</#if>;
     });
+
+    // Extract estimated rowcount map
+    var opRowCountMap = {};
+    // Get OpId-Rowocunt Map
+    function buildRowCountMap() {
+      var phyText = $('#query-physical').find('pre').text();
+      var opLines = phyText.split("\n");
+      opLines.forEach(line => {
+        if (line.trim().length > 0) {
+          var opId = line.match(/\d+-\d+/g)[0];
+          var opRowCount = line.match(/rowcount = ([^,]+)/)[1];
+          opRowCountMap[opId] = Number(opRowCount).toLocaleString('en');
+        }
+      });
+    }
     </script>
 
 </#macro>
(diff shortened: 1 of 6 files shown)


khurram-faraaz:
@kkhatua  Will this increase the total query execution time in any way ?

kkhatua:
No. Why do you think it'll increase query execution time? It's just a visual aid to understand if the stats were way off. Users who rely on using sampling to generate stats can see if their sample size is too small.

arina-ielchiieva:
@kkhatua why it's false by default? Also can we just show estimated row count without having separate option for this? If estimated row count is available, we show the info, if not, we don't. Or maybe make decision based on statistics availability, if if available show, otherwise don't.

kkhatua:
@arina-ielchiieva 

The motivation for this PR comes from the need for engineers to analyze queries as plans change due to introduction of statistics. An initial thought was to add an additional column, but, I think, we already have a lot of columns. I've tried to figure which columns to trim, but almost all seem relevant. I know we might come back to doing similar things with Resource Management as well, where we'll again need to work on estimates vs actual. So adding additional columns is not practical.

Showing the estimates based on whether a planning decision was made using statistics is not possible unless the profile JSON itself carries some hint that statistics were used.

Also, I added the toggle button to provide a mechanism to hide the estimates by default (another reason why not an additional column). I'm worried that users will get the impression that there are issues with Drill because of estimates being wildly off. Even if they are sufficiently accurate (like NDV-based estimates vs actual), most users don't have the insight into how the stats are being used.

Users who have insight into such things can make use of the estimates to tune parameters (e.g. broadcast or selectivity thresholds) to force changes in plans that are sub-optimal. Based on this, I thought we should go with the parenthesis option for showing the estimated row counts.

arina-ielchiieva:
Well, I am ok if we leave estimates in parenthesis but I am still unclear why we need config param, if estimates are hidden by default.

kkhatua:
Because engineers are, by nature, lazy and would like to get as much insight with minimal clicks. 
 😇

Changing the default on a development environment helps.  ( @amansinha100 , agree? )

arina-ielchiieva:
Oh, now I see where it comes from :). Just not sure if this is worth adding new config param. Won't be devs lazy enough to update drill config and restart the drillbits? Is adding it as session / system option more convenient for devs?

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -587,6 +622,49 @@
       if (e.target.form) 
         <#if model.isOnlyImpersonationEnabled()>doSubmitQueryWithUserName()<#else>doSubmitQueryWithAutoLimit()</#if>;
     });
+
+    // Convert scientific to Decimal [Ref: https://gist.github.com/jiggzson/b5f489af9ad931e3d186]
+    function scientificToDecimal(num) {
```
arina-ielchiieva: Won't this be are convenient to do in Java and just pass expected string representation to UI?

kkhatua: Ideally, yes.

But the way we construct the `Operators Overview` table is by only injecting values (which then get converted to human readable formats based on data types). That makes it difficult for me to inject these `<div>` elements during construction.

Also, with Javascript, I reduce the processing and construction overhead on the WebServer and have the client browsers do this formatting.

In theory, the entire profile can be constructed by Javascript from the JSON profile, but basic summaries are easier to implement (and faster to calculate) in Java.

kkhatua:
@arina-ielchiieva I havent seen any system/session options for WebServer UI settings. All of them are done using `drill-override.conf`, probably because it makes it easier to implement features for the WebUI. Making it a session option doesn't provide any benefit if there are users who might be able to access Drill's SQL capabilities only via the WebUI. Hence, providing this as a Javascript based feature made more sense.

kkhatua:
@amansinha100  could you also do a review of this? As of now, I don't anticipate any additional work and the PR looks ready to commit.

arina-ielchiieva:
@kkhatua could you please request someone to review JS part. I am not a JS expert but the changes does not look good to me. I am not sure why we need such advanced number parsing so feedback from someone who knows what is going on will be very helpful.

kkhatua:
@arina-ielchiieva the feature allows an advanced user to compare estimated rows with the actual rowcounts for the query. Since the estimates are part of the Logical Plan Text , but in a scientific format, we need to parse it to represent it as a whole number with proper formatting (i.e. use of comma as a 1000s separator). I used a Github snippet since that had the most references and usage, rather than look for a (possible?) Javascript library for this feature. 
@sohami could you take a look as well?

kkhatua:
@agozhiy 
Could you take a look at this PR?

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -587,6 +622,49 @@
...
+      return num;
+    }
+
+    // Extract estimated rowcount map
+    var opRowCountMap = {};
+    // Get OpId-Rowocunt Map
+    function buildRowCountMap() {
+      var phyText = $('#query-physical').find('pre').text();
+      var opLines = phyText.split("\n");
+      for (var l in opLines) {
```
agozhiy:
```suggestion
      opLines.forEach(line => {
```

exec/java-exec/src/main/resources/rest/profile/profile.ftl
agozhiy: Why do you need this at all? Javascript parseInt() function which you use can parse scientific format without a problem.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -84,6 +84,29 @@
         document.getElementById(warningElemId).style.display="none";
     }
 
+    //Injects Estimated Rows
+    function injectEstimatedRows() {
+      Object.keys(opRowCountMap).forEach(key => {
+        var tgtElem = $("td.estRowsAnchor[key='"+key+"']");
```
agozhiy:
```suggestion
        var tgtElem = $("td.estRowsAnchor[key='" + key + "']");
```

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -84,6 +84,29 @@
         document.getElementById(warningElemId).style.display="none";
     }
 
+    //Injects Estimated Rows
+    function injectEstimatedRows() {
+      Object.keys(opRowCountMap).forEach(key => {
+        var tgtElem = $("td.estRowsAnchor[key='"+key+"']"); 
+        var status = tgtElem.append( "<div class='estRows' title='Estimated'>("+opRowCountMap[key]+")</div>" );
```
agozhiy:
```suggestion
        var status = tgtElem.append("<div class='estRows' title='Estimated'>(" + opRowCountMap[key] + ")</div>");
```

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -84,6 +84,29 @@
...
+      });
+    }
+
+    //Toggle Estimates' visibility
+    function toggleEstimates(tgtColumn) {
+      var colClass = '.est'+tgtColumn
```
agozhiy:
```suggestion
      var colClass = '.est' + tgtColumn
```

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -84,6 +84,29 @@
...
+      var estimates = $(colClass);
+      if (estimates.filter(":visible").length > 0) {
+        $(colClass).each(function(){
```
agozhiy:
```suggestion
        $(colClass).each(function () {
```

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -84,6 +84,29 @@
...
+          $(this).attr("style", "display:none");
+        });
+      } else {
+        $(colClass).each(function(){
```
agozhiy:
```suggestion
        $(colClass).each(function () {
```

exec/java-exec/src/main/resources/rest/profile/profile.ftl
kkhatua: I don't believe that works. The parseInt function will parse until the 'E' symbol.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -587,6 +622,49 @@
...
+    var opRowCountMap = {};
+    // Get OpId-Rowocunt Map
+    function buildRowCountMap() {
+      var phyText = $('#query-physical').find('pre').text();
+      var opLines = phyText.split("\n");
+      for (var l in opLines) {
+        var line = opLines[l];
+        if (line.trim().length > 0) {
+          var opId = line.match(/\d+-\d+/g)[0];
+          var opRowCount = line.match(/rowcount = \S+/g)[0].split(' ')[2].replace(',','').trim();
```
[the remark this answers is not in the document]
kkhatua: I can't use this because the `rowcount` might be in the scientific format, resulting in a partial extraction.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -587,6 +622,49 @@
...
+          if (opRowCount.includes("E")) {
+            opRowCountMap[opId] = parseInt(scientificToDecimal(opRowCount)).toLocaleString('en');
```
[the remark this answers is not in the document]
kkhatua: Good point. I'll check for larger values and apply the parseLong() method instead.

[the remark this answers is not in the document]
kkhatua: Tested with ##E18  using the Number() function.

kkhatua:
@agozhiy done most of the changes you've suggested and commented on the rest.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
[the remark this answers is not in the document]
agozhiy: Right. Still, it can be simplified. This should work with both formats:
`line.match(/rowcount = ([^,]+)/)[1]`

exec/java-exec/src/main/resources/rest/profile/profile.ftl
agozhiy: From the Chrome web console:
`> parseInt(1.6E7)`
`< 16000000`
So it works it seems.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
agozhiy: Spaces are still present around the expression inside parentheses. This is inconsistent with the overall code style.

agozhiy:
@kkhatua, thank you for the changes, please answer just a few more comments I left in conversations.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
kkhatua: Yep. Verified it. Thanks for pointing it out.

kkhatua:
@agozhiy I've done the changes requested. Please review.

agozhiy:
@kkhatua, please address one more comment and squash the commits.

exec/java-exec/src/main/resources/rest/profile/profile.ftl
```diff
@@ -587,6 +622,25 @@
...
+    var opRowCountMap = {};
+    // Get OpId-Rowocunt Map
+    function buildRowCountMap() {
+      var phyText = $('#query-physical').find('pre').text();
+      var opLines = phyText.split("\n");
+      opLines.forEach(line => {
+        if (line.trim().length > 0) {
+          var opId = line.match(/\d+-\d+/g)[0];
+          var opRowCount = line.match(/rowcount = ([^,]+)/)[1];
+          if (opRowCount.includes("E")) {
```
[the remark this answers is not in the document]
kkhatua: Sorry... missed this in the email notification. Done now.

agozhiy:
+1. 
Thanks for the changes.

PR #1787, opened 2019-05-11
DRILL-7203: Back button not working 

kkhatua:
The root cause of this bug is that when a form is submitted, Drill uses the POST mechanism via an AJAX call. The response is then used to swap the existing `/query` page's DOM. This means that the browser history does not capture the change because the URL has never changed. As a result, when the user hits the `[BACK]` button, the call is made to the browser which redirects to the last page visited just before `/query`. 
Changing the AJAX call mechanism is risky, so the workaround was to leverage the `sessionStorage` feature supported by most modern (and secure) browsers. When a query is submitted, `sessionStorage`  is used to cache the submitted query (like a session/cookie). If the back button on the page is hit, an additional flag is set, telling the Javascript to revisit the submission page and use the _cached_ query. If the submission page is revisited via any other mechanism, it is assumed that the `[BACK]` button was not pressed (since the additional flag is not set), and the cached query is cleared from the browser memory.

### DIFF
--- a/exec/java-exec/src/main/resources/rest/errorMessage.ftl
+++ b/exec/java-exec/src/main/resources/rest/errorMessage.ftl
@@ -20,12 +20,10 @@
 <#include "*/generic.ftl">
 <#macro page_head>
 <script>
+  //Hack to mimic browser back button
   function goToPreviousPage() {
-    console.log("Going back from "+location.pathname + " to " +  document.referrer + "( "+window.history.state+" )");
-    //window.history.go(-1);
-	sessionStorage.useCachedQuery = true;
-	console.log("Reloading with query: " + sessionStorage.cachedQuery)
-	location.reload();
+    sessionStorage.useCachedQuery = true;
+    location.reload();
   }
 </script>
 </#macro>
@@ -40,7 +38,7 @@
     </div>
     <div class="panel-body"><span style="font-family:courier,monospace;white-space:pre-wrap">${model}</span></div>
   </div>
-  <button type="reset" class="btn btn-default" id="backBtn" style="display:inline" Xonclick="window.history.back()" onclick="goToPreviousPage()"><span class="glyphicon glyphicon-step-backward"></span> Back</button>
+  <button type="reset" class="btn btn-default" id="backBtn" style="display:inline" onclick="goToPreviousPage()"><span class="glyphicon glyphicon-step-backward"></span> Back</button>
 </#macro>
 
 <@page_html/>
--- a/exec/java-exec/src/main/resources/rest/errorMessage.ftl
+++ b/exec/java-exec/src/main/resources/rest/errorMessage.ftl
@@ -19,6 +19,15 @@
 -->
 <#include "*/generic.ftl">
 <#macro page_head>
+<script>
+  function goToPreviousPage() {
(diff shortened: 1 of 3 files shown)


arina-ielchiieva:
@sohami could you please review?

sohami:
@kkhatua : I am not sure if this is the right way to fix this issue. The issue as you mentioned is because we are loading the response from server in same page. I guess the reason for that is because any error in query on server side is also send as a HTTP success response. So we just chose to update the /query page with either actual result or exception from server side. Please consider below 2 options:

1) Create a single `queryResult.ftl` page or repurpose `result.ftl` page, now on server side based on success or failure we populate it with either result.ftl or errorMessage.ftl template. In this way the form on /query page can always be directed to load /queryResult.ftl page (I guess using the `action` property)
2) Somehow in success handler of ajax post request get the response page URL and load that url with the response data. Or may be use redirection mechanism to load the result page. Something like here: https://stackoverflow.com/questions/199099/how-to-manage-a-redirect-request-after-a-jquery-ajax-call

kkhatua:
@sohami 
The issue is not so much with a new FTL page as much as it is with populating the (previous) submitting page's fields with the values just prior to submission. Since the results are loaded into the DOM without a change in the address, there is no way to mimic the `[back]` button's behaviour without this.

dobesv:
It seems like this problem came about because of the need to include the `User-Name` header with the form submission sometimes.  Someone introduced a sort of trick using XHR to submit the form and then replace the whole page body with the resulting page using `document.open` and `document.write`.  However, this method of replacing the page content does not add an element to the history.

Is there a good reason why the username couldn't be included in the form data (instead of the custom header) to avoid this workaround?

Alternatively, perhaps we could skip this workaround at least for people who are not using the `User-name` user impersonation feature - in this case a regular FORM submission would add to the history and reload the page in the "normal" way, avoiding this problem.

paul-rogers:
@dobesv thanks much for digging into this issue. Your knowledge of web front-end mechanics is greatly appreciated. My guess is that we've got an issue where someone confused a REST API with web page display. As far as I can tell, the original design of the Web console was as a set of REST calls, with the HTML UI added later. So, not surprising we ended up with this mess.

Seems reasonable to me that, for REST calls, the user name is in a header. But, for web pages, we put the data in the normal places: a cookie, a page parameter (for GET) or a form field (for POST).

Do you want to try a fix to see if you can make your solution work?

dobesv:
Yeah, I'll see if I have time for that.  If the strict REST API approach is preferred, then the page should use the REST API and operate more like a single-page application that sends and receives JSON payloads.

For now that is probably too big of a project - changing this to submit the username as a form field should be fairly straightforward.

On 2/19/2020 11:20:59 AM, Paul Rogers <notifications@github.com> wrote:
@dobesv [https://github.com/dobesv] thanks much for digging into this issue. Your knowledge of web front-end mechanics is greatly appreciated. My guess is that we've got an issue where someone confused a REST API with web page display. As far as I can tell, the original design of the Web console was as a set of REST calls, with the HTML UI added later. So, not surprising we ended up with this mess.
Seems reasonable to me that, for REST calls, the user name is in a header. But, for web pages, we put the data in the normal places: a cookie, a page parameter (for GET) or a form field (for POST).
Do you want to try a fix to see if you can make your solution work?

arina-ielchiieva:
Closing in favor of #1994.

PR #1788, opened 2019-05-15
DRILL-7204: Add proper validation when creating plugin

agozhiy:


exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/StorageResources.java
```diff
@@ -204,6 +204,13 @@ public JsonResult createOrUpdatePluginJSON(PluginConfigWrapper plugin) {
   @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
   @Produces(MediaType.APPLICATION_JSON)
   public JsonResult createOrUpdatePlugin(@FormParam("name") String name, @FormParam("config") String storagePluginConfig) {
+    name = name.trim();
+    if (name.isEmpty()) {
+      return message("Error (a storage name cannot be empty)");
+    }
+    if (!name.matches("^[a-zA-Z0-9._-]+$")) {
```
arina-ielchiieva: Since we don't enforce this on the backend side, I don't think we should enforce this on the frontend side as well.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/StorageResources.java
```diff
@@ -204,6 +204,13 @@ public JsonResult createOrUpdatePluginJSON(PluginConfigWrapper plugin) {
   @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
   @Produces(MediaType.APPLICATION_JSON)
   public JsonResult createOrUpdatePlugin(@FormParam("name") String name, @FormParam("config") String storagePluginConfig) {
+    name = name.trim();
+    if (name.isEmpty()) {
+      return message("Error (a storage name cannot be empty)");
```
arina-ielchiieva: Could you please replace alerts with more user friendly message window?

vvysotskyi:
@agozhiy, is it possible instead of prohibiting plugins with specific symbols in names, escape them before sending and unescape later?

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/StorageResources.java
agozhiy: Changed the approach to escape special symbols instead of block them.

agozhiy:
@vvysotskyi, thank you for suggestion, did as you said.

Also, added a dialog window template to use it instead of alerts.
Screenshot from 2019-05-17 16-28-45
Screenshot from 2019-05-17 16-28-01

agozhiy:
@arina-ielchiieva, I addressed your comments, would you please review?

exec/java-exec/src/main/resources/rest/storage/update.ftl
```diff
@@ -126,9 +128,9 @@
     }
 
     function deleteFunction() {
-      if (confirm("Are you sure?")) {
-        $.get("/storage/${model.getName()}/delete", serverMessage);
-      }
+      showConfirmationDialog("Are you sure?", function() {
```
arina-ielchiieva: Maybe we should improve message, something like: `Are you sure you want to delete ....?`

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/StorageResources.java
```diff
@@ -204,6 +204,10 @@ public JsonResult createOrUpdatePluginJSON(PluginConfigWrapper plugin) {
   @Consumes(MediaType.APPLICATION_FORM_URLENCODED)
   @Produces(MediaType.APPLICATION_JSON)
   public JsonResult createOrUpdatePlugin(@FormParam("name") String name, @FormParam("config") String storagePluginConfig) {
+    name = name.trim();
```
arina-ielchiieva: Can name be null? Just checking to sure we won't get NPE here.

arina-ielchiieva:
@kkhatua could you please also review JS part?

exec/java-exec/src/main/resources/rest/storage/update.ftl
agozhiy: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/server/rest/StorageResources.java
agozhiy: Under normal circumstances - no. We use a form with two fields to send a POST request, you can leave some of them empty, but they will be initialized anyway. 
You can brake it intentionally of course, for example by deletion of a form filed element via Developer tools or using Postman to send a malformed request.

agozhiy:
@arina-ielchiieva I addressed your comments.
@kkhatua do you have some remarks?

kkhatua:
Sorry. Took me a while to get around to trying this out. I tried breaking it and it looks robust enough. Any malformed calls via REST APIs would need server-side checks, but that would involve much more work and I don't think it is critical.
Nice work, @agozhiy 
+1

arina-ielchiieva:
+1

arina-ielchiieva:
@agozhiy please squash the commits.

agozhiy:
Done.

PR #1789, opened 2019-05-16
DRILL-7181: Improve V3 text reader (row set) error messages

paul-rogers:
Adds an error context to the scan framework to add plugin-specific error
context.

Refines how the `columns` column can be used with the text reader. If
headers are used, then `columns` is just another column. An error is
raised, however, if `columns[x]` is used when headers are enabled.

Added additional unit tests to fully define the `columns` column
behavior.

Added another builder abstraction where a constructor argument list
became too long.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/file/FileScanFramework.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/file/FileScanFramework.java
@@ -22,7 +22,10 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 
+import org.apache.drill.common.exceptions.ChildErrorContext;
+import org.apache.drill.common.exceptions.CustomErrorContext;
 import org.apache.drill.common.exceptions.UserException;
+import org.apache.drill.common.exceptions.UserException.Builder;
 import org.apache.drill.exec.physical.impl.scan.file.FileMetadataManager.FileMetadataOptions;
 import org.apache.drill.exec.physical.impl.scan.framework.ManagedReader;
 import org.apache.drill.exec.physical.impl.scan.framework.ManagedScanFramework;
@@ -73,6 +76,20 @@ public class FileScanFramework extends ManagedScanFramework {
    */
 
   public interface FileSchemaNegotiator extends SchemaNegotiator {
+
+    /**
+     * Gives the Drill file system for this operator.
+     */
+
+    DrillFileSystem fileSystem();
+
+    /**
+     * Describes the file split (path and block offset) for this scan.
+     *
+     * @return Hadoop file split object with the file path, block
+     * offset, and length.
+     */
+    FileSplit split();
   }
 
   /**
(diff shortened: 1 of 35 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/file/FileScanFramework.java
```diff
@@ -206,7 +206,13 @@ protected FileSplit nextSplit() {
 
   @Override
   protected SchemaNegotiatorImpl newNegotiator() {
-    return new FileSchemaNegotiatorImpl(this);
+    SchemaNegotiatorImpl nego = new FileSchemaNegotiatorImpl(this);
+    String tableName = Path.getPathWithoutSchemeAndAuthority(currentSplit.getPath()).toString();
+    if (currentSplit.getStart() != 0) {
+      tableName += String.format(", offset %,d", currentSplit.getStart());
+    }
+    nego.setTableName(tableName);
```
arina-ielchiieva: Not quite sure what table name this will be, why you need offset etc. Can you please give an example?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/framework/SchemaNegotiatorImpl.java
```diff
@@ -73,6 +76,25 @@ public OperatorContext context() {
     return framework.context();
   }
 
+  @Override
+  public RowSetContext parentErrorContext() {
```
arina-ielchiieva: Why we need parent error context?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/ResultSetLoader.java
```diff
@@ -41,6 +42,12 @@
 
   public static final int DEFAULT_ROW_COUNT = BaseValueVector.INITIAL_VALUE_ALLOCATION;
 
+  /**
+   * Context for error messages.
+   */
+
+  RowSetContext context();
```
arina-ielchiieva: Can this context be used for some this else rather than errors?
If not, maybe its better to name it `errorContext` everywhere?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetContext.java
```diff
@@ -0,0 +1,31 @@
...
+package org.apache.drill.exec.physical.rowSet;
+
+import org.apache.drill.common.exceptions.UserException;
+
+/**
+ * Provides context information for errors raised in the result set
+ * loader or related classes.
+ */
+
+public interface RowSetContext {
```
arina-ielchiieva: Same here: should we consider naming `RowSetErrorContext`?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/ResultSetLoader.java
paul-rogers: Renamed the class. Then, restructured the class to be more generic and moved it into common, alongside UserException. Doing so allowed simplifying usage of the context.

paul-rogers:
Revised the error context to be more general; thanks Arina for the suggestions.

Revised the way that the Drill file system and file split is passed to readers to make the code simpler and more consistent.

arina-ielchiieva:
@paul-rogers, new changes look really good. +1
Please squash the commits and fix build errors.

arina-ielchiieva:
Functional / Advanced tests have passed, all unit tests as well.

PR #1790, opened 2019-05-16
DRILL-7257: Set nullable var-width vector lastSet value

paul-rogers:
Turns out this is due to a subtle issue with variable-width nullable
vectors. Such vectors have a lastSet attribute in the Mutator class.
When using "transfer pairs" to copy values, the code somehow decides
to zero-fill from the lastSet value to the record count. The row set
framework did not set this value, meaning that the RemovingRecordBatch
zero-filled the dir0 column when it chose to use transfer pairs rather
than copying values. The use of transfer pairs occurs when all rows in
a batch pass the filter prior to the removing record batch.

Modified the nullable vector writer to properly set the lastSet value at
the end of each batch. Added a unit test to verify the value is set
correctly.

Includes a bit of code clean-up.


arina-ielchiieva:
+1, LGTM.

arina-ielchiieva:
Functional and Advanced tests have passed, all unit tests as well.

PR #1792, opened 2019-05-20
DRILL-7250: Query with CTE fails when its name matches to the table name without access

vvysotskyi:
- Removed `catalogReader.getTable(tempNode.names)` call since it does not resolve table names correctly using provided scope.
- Reworked `catalogReader.isValidSchema()` method to follow logic on how Calcite determines whether the table is present: it uses schemas from `CatalogReader` and resolves specified schema among them.
- Updated Calcite version to use the fix for `CALCITE-3061`.
- Added unit test.

For problem description please see [DRILL-7250](https://issues.apache.org/jira/browse/DRILL-7250).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
+++ b/exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
@@ -61,6 +61,8 @@ public class DynamicRootSchema extends DynamicSchema {
   @Override
   protected CalciteSchema getImplicitSubSchema(String schemaName,
                                                boolean caseSensitive) {
+    // Drill registers schemas in lower case, see AbstractSchema constructor
+    schemaName = schemaName == null ? null : schemaName.toLowerCase();
     CalciteSchema retSchema = getSubSchemaMap().get(schemaName);
     if (retSchema != null) {
       return retSchema;
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
@@ -26,6 +26,7 @@ import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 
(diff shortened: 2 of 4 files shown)


exec/java-exec/src/main/java/org/apache/calcite/jdbc/DynamicRootSchema.java
```diff
@@ -61,6 +61,8 @@ public StoragePluginRegistry getSchemaFactories() {
   @Override
   protected CalciteSchema getImplicitSubSchema(String schemaName,
                                                boolean caseSensitive) {
+    // Drill registers schemas in lower case, see AbstractSchema constructor
+    schemaName = schemaName != null ? schemaName.toLowerCase() : null;
```
arina-ielchiieva: Not critical, but maybe it's better to invert: `schemaName == null ? ...`

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/SqlConverter.java
```diff
@@ -280,12 +280,10 @@ protected void validateFrom(
             changeNamesIfTableIsTemporary(tempNode);
 
             // Check the schema and throw a valid SchemaNotFound exception instead of TableNotFound exception.
-            if (catalogReader.getTable(tempNode.names) == null) {
-              catalogReader.isValidSchema(tempNode.names);
-            }
+            catalogReader.isValidSchema(tempNode.names);
```
arina-ielchiieva: Names cannot be null?

vvysotskyi: Good question, when it is set in the constructor or using a setter of `SqlIdentifier` class, it cannot be null, since there is used `ImmutableList.copyOf()` method for incoming lists. But this field is public and may be changed in some other places (I didn't find any place where it is set to null in the current Drill and Calcite code).

Since Calcite's `SqlIdentifier` code does not assume that `names` may be null (there is a lot of code where this field is used without checks), I think we should also do not check it.

arina-ielchiieva:
+1, LGTM.

PR #1795, opened 2019-05-24
DRILL-7276: Fixed an XSS vulnerability in Drill Web-UI query profile …

agozhiy:
…page

### DIFF
--- a/exec/java-exec/src/main/resources/rest/alertModals.ftl
+++ b/exec/java-exec/src/main/resources/rest/alertModals.ftl
@@ -62,12 +62,21 @@
         var inputValuesKeys = Object.keys(inputValues);
         for (i=0; i<inputValuesKeys.length; ++i) {
             let currKey=inputValuesKeys[i];
-            updatedHtml=updatedHtml.replace(currKey, inputValues[currKey]);
+            updatedHtml=updatedHtml.replace(currKey, escapeHtml(inputValues[currKey]));
         }
         modalBody.innerHTML=updatedHtml;
       }
       //Show Alert
       $('#errorModal').modal('show');
+    }
+
+    function escapeHtml(str) {
+        return str.replace(/&/g,'&amp;')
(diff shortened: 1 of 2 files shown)


vvysotskyi:
@agozhiy, Change looks good, thanks for fixing this issue.

Could you please revise other JS scripts and check the usage of `html(...)` method and other possibly dangerous methods which may cause similar issues?

vvysotskyi:
@agozhiy, I have noticed another issue: when the script is inserted into the `autoLimit` filed, it breaks the warning window with the error message, could you please take a look at this issue?
image

agozhiy:
@vvysotskyi thanks for finding this case, fixed it now. 
Cannot find any other dangerous usages of html() function, in most of the cases we use FreeMarker to substitute values.
Please, review.

exec/java-exec/src/main/resources/rest/alertModals.ftl
```diff
@@ -62,14 +62,18 @@
...
+            updatedHtml=updatedHtml.replace(currKey, escapeHtml(inputValues[currKey]));
         }
         modalBody.innerHTML=updatedHtml;
       }
       //Show Alert
       $('#errorModal').modal('show');
     }
 
+    function escapeHtml(str) {
+        return str.replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;') ;
```
vvysotskyi: Could you please escape other characters as it is described in [Cross Site Scripting Prevention Cheat Sheet](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md#rule-1---html-escape-before-inserting-untrusted-data-into-html-element-content)?

agozhiy: Done.

agozhiy:
Done.

PR #1796, opened 2019-05-26
DRILL-7261: Simplify Easy framework config

paul-rogers:
Most format plugins are created using the Easy format plugin. A recent
change added support for the "row set" scan framework. After converting
the text and log reader plugins, it became clear that the setup code
could be made simpler.

* Add the user name to the "file scan" framework.
* Pass the file system, split and user name to the batch reader via
 the "schema negotiator" rather than via the constructor.
* Create the traditional "scan batch" scan or the new row-set scan via
functions instead of classes.
* Add Easy config option and method to choose the kind of scan
framework.
* Add Easy config options for some newer options such as whether the
plugin supports statistics.
* Batch reader defined via a plugin method in the simplest case.
* Default error context provided if the plugin does not provide one.

Tested by running all unit tests for the CSV reader which is based on
the new framework, and by testing the converted log reader (that reader
is not part of this commit.)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/file/FileScanFramework.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/file/FileScanFramework.java
@@ -62,7 +62,8 @@ import org.apache.hadoop.mapred.FileSplit;
 
 public class FileScanFramework extends ManagedScanFramework {
 
-  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FileScanFramework.class);
+  private static final org.slf4j.Logger logger =
+      org.slf4j.LoggerFactory.getLogger(FileScanFramework.class);
 
   /**
    * The file schema negotiator adds no behavior at present, but is
@@ -80,7 +81,6 @@ public class FileScanFramework extends ManagedScanFramework {
     /**
      * Gives the Drill file system for this operator.
      */
-
     DrillFileSystem fileSystem();
 
     /**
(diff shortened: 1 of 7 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/TextFormatPlugin.java
```diff
@@ -336,6 +268,53 @@ public RecordReader getRecordReader(FragmentContext context,
...
+    // Provide custom error context
+    builder.setContext(
+        new CustomErrorContext() {
+          @Override
+          public void addContext(UserException.Builder builder) {
+            builder.addContext("Format plugin:", PLUGIN_NAME);
+            builder.addContext("Plugin config name:", getName());
+            builder.addContext("Extract headers:",
+                Boolean.toString(getConfig().isHeaderExtractionEnabled()));
+            builder.addContext("Skip headers:",
```
arina-ielchiieva: Skip lines?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/scan/file/FileScanFramework.java
```diff
@@ -89,6 +89,7 @@
      * @return Hadoop file split object with the file path, block
      * offset, and length.
      */
+
```
arina-ielchiieva: Nit: remove new line.

arina-ielchiieva:

paul-rogers:
Squashed commits

PR #1797, opened 2019-05-26
DRILL-7278: Refactor result set loader projection mechanism

paul-rogers:
Drill 1.16 added a enhanced scan framework based on the row set mechanisms, and a "provisioned schema" feature build on top of that framework.

The next step is to convert the log reader as an example of how to create (or revise) a format plugin to use the new framework. The log reader has two "special" columns that can be projected, but do expand as part of the wildcard. Added a column schema property to enable this feature.

Doing this work made clear a bit of refactoring is necessary now that the framework supports:

* Wildcard or explicit projection lists
* Special columns (no wildcard expansion) in either the "reader" or "provided" schema.
* Strict provided schema (reader columns may not be projected in a wildcard if they do not appear in the provided schema.)
* Type conversions: implicit, built-in and custom.
* Recursively apply all of the above to columns in a may or map array.  

All of the above is gathered into a new "projection set" abstraction presented to the result set loader. Each time a column is added (either up-front or as the read proceeds), the loader asks "is this column projected?" The projection set uses the above logic to answer the question, and to optionally insert a type conversion.

Other changes:

* Unifies the type conversion mechanism added as part of provisioned schemas.
* Replaced the previous "schema transformer" mechanism with a new "type  conversion" mechanism that unifies type conversion, provided schemas and an optional custom type conversion mechanism.
* Column writers can report if they are projected. Moved this query
from metadata to the column writer itself.
* Extended and clarified documentation of the feature.
* Revised and/or added unit tests.

Once these changes are in, a separate PR will convert the log reader. That work, in turn, will be written up as a tutorial for the developers who want to create or convert plugins to use the new framework.


exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/ProjectionSet.java
```diff
@@ -0,0 +1,105 @@
...
+ * set of columns defined within the schema. When used with columns
+ * added dynamically, a column may be excluded from the projection
+ * set if it is not part of the defined external schema.</ul>
+ * <li>Metadata filtering: A reader may offer a special column which
+ * is available only in explicit projection, and behaves like Drill's
+ * implicit file columns. Such columns are not included in a "project
+ * all" projection.</li>
+ * <p>
+ * At present, only the top-level row supports these additional filtering
+ * options; they are not supported on mays (though could be with additional
```
arina-ielchiieva:
```suggestion
 * options; they are not supported on maps (though could be with additional
```

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/project/ProjectionType.java
```diff
@@ -15,19 +15,70 @@
...
   WILDCARD,     // *
-  UNSPECIFIED,  // x
+
+  /**
+   * Projection is by simple name. "General" means that
+   * we have no hints about the type of the column from
+   * the projection.
+   */
+
+  GENERAL,      // x
```
arina-ielchiieva: Not sure if `GENERAL` is better than `UNSPECIFIED` but let it be.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/project/ProjectionType.java
```diff
@@ -53,13 +109,26 @@ public static ProjectionType typeFor(MajorType majorType) {
...
+    }
     return SCALAR;
   }
 
-  public boolean isCompatible(ProjectionType other) {
-    switch (other) {
+  /**
+   * Reports if this type (representing an item in a projection list)
+   * is compatible with the projection type representing an actual
+   * column produced by an operator. The check is not symmetric.
```
arina-ielchiieva: `The check is not symmetric.` could you please explain this part?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/project/RequestedTupleImpl.java
```diff
@@ -74,6 +75,7 @@
 public class RequestedTupleImpl implements RequestedTuple {
 
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(RequestedTupleImpl.class);
+  private static final Collection<SchemaPath> PROJECT_ALL = ImmutableList.of(SchemaPath.STAR_COLUMN);
```
arina-ielchiieva: Consider using build-in Java methods instead of guava when possible: `Collections.sigletonList`.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/AbstractColumnMetadata.java
```diff
@@ -30,10 +30,11 @@
 import org.apache.drill.exec.record.metadata.schema.parser.SchemaExprParser;
 import org.joda.time.format.DateTimeFormatter;
 
-import java.io.IOException;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.stream.Collectors;
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
```
arina-ielchiieva: Is this correct import order?

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/ColumnMetadata.java
```diff
@@ -51,19 +51,20 @@
...
    * rather than by the user in the schema.
    */
   String BLANK_AS_PROP = DRILL_PROP_PREFIX + "blank-as";
 
+  /**
+   * Indicates whether to project the column in a wildcard (*) query.
+   * Special columns may be excluded from projection. Certain "special"
+   * columns may be available only when explicitly requested.
+   */
+  String EXCLUDE_FROM_WILDCARD = DRILL_PROP_PREFIX + "special";
```
arina-ielchiieva: Could you please provide an example form the log reader? I did not quite understand from the description.

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/AbstractColumnMetadata.java
paul-rogers: It is the order into which Eclipse sorts the imports automatically on save. If Drill has a preferred order, I can see if I can get Eclipse to follow that order; or I can turn off the auto-sort and organize them manually.

arina-ielchiieva:
+1, LGTM. Please squash the commits.

paul-rogers:
Commits squashed

PR #1798, opened 2019-05-27
DRILL-7279: Enable provided schema for text files without headers

paul-rogers:
Drill 1.16 allowed a "provided schema" for text files with headers. This PR extends the feature to text files without headers. Also adds support to use provided-schema table properties to override plugin config properties.

* Allows a provided schema for text files without headers. The provided schema columns replace the `columns` column that is normally used.
* Allows customizing four of the most commonly used text format properties using table properties. The table properties "override" properties set in the plugin config.
* Added unit tests for the newly supported use cases.

See DRILL-7279 for additional details.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/easy/EasyFormatPlugin.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/easy/EasyFormatPlugin.java
@@ -23,9 +23,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import org.apache.drill.exec.physical.base.MetadataProviderManager;
-import org.apache.drill.shaded.guava.com.google.common.base.Functions;
-import org.apache.drill.shaded.guava.com.google.common.collect.Maps;
 import org.apache.drill.common.exceptions.ExecutionSetupException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.expression.SchemaPath;
@@ -35,12 +32,13 @@ import org.apache.drill.exec.ops.FragmentContext;
 import org.apache.drill.exec.ops.OperatorContext;
 import org.apache.drill.exec.physical.base.AbstractGroupScan;
 import org.apache.drill.exec.physical.base.AbstractWriter;
+import org.apache.drill.exec.physical.base.MetadataProviderManager;
 import org.apache.drill.exec.physical.base.PhysicalOperator;
 import org.apache.drill.exec.physical.base.ScanStats;
 import org.apache.drill.exec.physical.base.ScanStats.GroupScanProperty;
 import org.apache.drill.exec.physical.impl.ScanBatch;
-import org.apache.drill.exec.physical.impl.WriterRecordBatch;
 import org.apache.drill.exec.physical.impl.StatisticsWriterRecordBatch;
+import org.apache.drill.exec.physical.impl.WriterRecordBatch;
 import org.apache.drill.exec.physical.impl.protocol.OperatorRecordBatch;
 import org.apache.drill.exec.physical.impl.scan.ScanOperatorExec;
 import org.apache.drill.exec.physical.impl.scan.file.FileScanFramework;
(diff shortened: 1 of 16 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/easy/EasyFormatPlugin.java
```diff
@@ -278,6 +277,7 @@ public CloseableRecordBatch buildScan(
      * rules for projection. It handles "early" or "late" schema readers. A typical
      * framework builds on standardized frameworks for files in general or text
      * files in particular.
+     * @param options
```
arina-ielchiieva: Please add description to avoid IDE warning.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/v3/CompliantTextBatchReader.java
```diff
@@ -122,10 +122,12 @@ public boolean open(ColumnsSchemaNegotiator schemaNegotiator) {
 
   /**
    * Extract header and use that to define the reader schema.
+   * @param strings
```
arina-ielchiieva: Same here.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/TupleMetadata.java
```diff
@@ -46,6 +46,10 @@
 public interface TupleMetadata extends Propertied, Iterable<ColumnMetadata> {
 
   public static final String IS_STRICT_SCHEMA_PROP = DRILL_PROP_PREFIX + "strict";
+  public static final String HAS_HEADERS_PROP = DRILL_PROP_PREFIX + "headers";
```
arina-ielchiieva: I am not sure this should be placed here, these properties are mostly related to text readers unlike `IS_STRICT_SCHEMA_PROP` or `TupleMetadata` class itself. Maybe its better to move these properties to separate class which will clearly indicate that they are applicable for the text readers only.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/TupleMetadata.java
```diff
@@ -46,6 +46,10 @@
...
+  public static final String SKIP_FIRST_LINE_PROP = DRILL_PROP_PREFIX + "skipFirstLine";
+  public static final String DELIMITER_PROP = DRILL_PROP_PREFIX + "delimiter";
```
arina-ielchiieva: Can we pass special symbols in delimiter prop? How for example new line can be passed?
`'drill.delimiter' = '\n'` will work? Or we need to escape it?

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/TupleMetadata.java
```diff
@@ -46,6 +46,10 @@
...
+  public static final String COMMENT_CHAR_PROP = DRILL_PROP_PREFIX + "commentChar";
```
arina-ielchiieva: Why we only have four properties since we adding such support it would be fair to add all text properties, there are not that many of them. Since I believe fo the the user it would be confusing why only some of them are supported.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/compliant/v3/TextParsingSettingsV3.java
```diff
@@ -17,45 +17,123 @@
...
+   * <li>The defaults in the plugin config (if properties not defined
+   * in the config JSON.</li>
+   * <li>The config values from the config JSON as stored in the
+   * plugin config.</li>
+   * <li>Table function settings expressed in the query (and passed
+   * in as part of the plugin config.</li>
+   * <li>Table properties.</li>
+   * </ul>
+   * <p>
+   * The implementation does not use system/session properties, but
```
arina-ielchiieva: Not sure of value if this comment.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/TupleMetadata.java
paul-rogers: Added unit tests. This revealed that quote delimiters **never worked** in the V2 text reader. That is, the following never worked:

```
quote="@"
escape="~"
```

Data:

```
@foo~@bar@
```

Reworked the parsing code for quoted strings to fix this issue, the output is now `foo@bar`. Also fixed:

```
@foo~~bar@
```
An escape without a quote should be a normal char. The output is now `foo~~bar`.

Verified that V3 works for the quote and delimiter as the same character (did not check V2):

```
quote="@"
escape="@"
```

Data:

```
@foo@@bar@
```

Produces output `foo@bar`.

arina-ielchiieva:
Thanks for adding all text properties. +1, LGTM.
Please squash the commits.

paul-rogers:
Commits squashed

PR #1806, opened 2019-06-13
DRILL-7292: Remove V1 and V2 text readers

paul-rogers:
Drill 1.16 introduced the "V2" text reader based on the row set
and provided schema mechanisms. V3 was available by system/session
option as the functionality was considered experimental.

The functionality has now undergone thorough testing. This commit makes
the V3 text reader available by default, and removes the code for the
original "V1" and the "new" (compliant, "V2") text reader.

The system/session options that controlled reader selection are retained
for backward compatibility, but they no longer do anything.

Specific changes:

* Removed the V2 "compliant" text reader.
* Moved the "V3" to replace the "compliant" version.
* Renamed the "complaint" package to "reader."
* Removed the V1 text reader.
* Moved the V1 text writer (still used with the V2 and V3 readers)
  into a new "writer" package adjacent to the reader.
* Removed the CSV tests for the V2 reader, including those that
  demonstrated bugs in V2.
* V2 did not properly handle the quote escape character. One or two unit
  tests depended on the broken behavior. Fixed them for the correct
  behavior.
* Behavior of "messy quotes" (those that appear in a non-quoted field)
  was undefined for the text reader. Added a test to clearly demonstrate
  the (somewhat odd) behavior. The behavior itself was not changed.

Reran all unit tests to ensure that they work with the now-default V3
text reader.


paul-rogers:
Travis build failed due to protobuf changes. However, this PR changed no protobufs: neither a .proto file or the generated .java files. Anyone know why the Travis build is complaining?

arina-ielchiieva:
Ran the unit tests, Functional and Advanced, there are four expected failures on Functional tests (asked @agozhiy to take a look).
Regarding protobuf job failures, @vvysotskyi will take a look. Mostly likely they are not related to your changes.

arina-ielchiieva:
Issue with protobufs is a bug, will be fixed in https://issues.apache.org/jira/browse/DRILL-7294.

paul-rogers:
@arina-ielchiieva, thank you for the review. Please check if the revised option messages are clear.

The test failures may be due to bugs fixed in the text reader.  @vvysotskyi, check if the files in question try to use the back-slash as the escape character. Drill's text readers are configured to use the quote character as an escape by default, but the code did not correctly handle this case. If the file contains the pattern `\"` or `""`, then the text reader will produce a different result than previously.

Otherwise, please do point me to the errors and I'll track down the issues.

arina-ielchiieva:
@paul-rogers test files are related due to partition column placement change, so they are expected.

+1, LGTM.

PR #1807, opened 2019-06-13
DRILL-7293: Convert the regex ("log") plugin to use EVF

paul-rogers:
Converts the log format plugin (which uses a regex for parsing) to work
with the Extended Vector Format.

This commit provides the basic conversion:

* Use the plugin config object to pass config to the Easy framework.
* Use the EVF scan mechanism in place of the legacy "ScanBatch"
mechanism.
* Minor code and README cleanup.

This commit corresponds to the Basic Tutorial steps in the EVF tutorial.


exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogBatchReader.java
```diff
@@ -0,0 +1,210 @@
...
+import org.apache.drill.exec.physical.rowSet.ResultSetLoader;
+import org.apache.drill.exec.physical.rowSet.RowSetLoader;
+import org.apache.drill.exec.record.metadata.TupleMetadata;
+import org.apache.drill.exec.vector.accessor.ScalarWriter;
+import org.apache.drill.shaded.guava.com.google.common.base.Charsets;
+import org.apache.hadoop.mapred.FileSplit;
+
+public class LogBatchReader implements ManagedReader<FileSchemaNegotiator> {
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LogBatchReader.class);
```
arina-ielchiieva: No need to use full imports.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -18,86 +18,224 @@
...
+import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
 import org.apache.hadoop.conf.Configuration;
 
-import java.util.List;
-import org.apache.hadoop.fs.FileSystem;
-import org.apache.hadoop.fs.Path;
-
 public class LogFormatPlugin extends EasyFormatPlugin<LogFormatConfig> {
+  public static final String PLUGIN_NAME = "logRegex";
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(LogFormatPlugin.class);
```
arina-ielchiieva: Same here.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -18,86 +18,224 @@
...
-        columns, userName, formatConfig);
+  private static EasyFormatConfig easyConfig(Configuration fsConf, LogFormatConfig pluginConfig) {
+    EasyFormatConfig config = new EasyFormatConfig();
+    config.readable = true;
+    config.writable = false;
+    // Should be block splitable, but logic not yet implemented.
+    config.blockSplittable = false;
+    config.compressible = true;
+    config.supportsProjectPushdown = true;
+    config.extensions = Lists.newArrayList(pluginConfig.getExtension());
```
arina-ielchiieva: Is it possible not to use guava list?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -18,86 +18,224 @@
...
-    throw new UnsupportedOperationException("unimplemented");
-  }
+    // Pattern and schema identical across readers; define
+    // up front.
 
-  @Override
-  public int getReaderOperatorType() {
-    return UserBitShared.CoreOperatorType.REGEX_SUB_SCAN_VALUE;
+    Pattern pattern = setupPattern();
+    Matcher m = pattern.matcher("test");
```
arina-ielchiieva: Why we are matching `test` here?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
```diff
@@ -1,8 +1,15 @@
 # Drill Regex/Logfile Plugin
-Plugin for Apache Drill that allows Drill to read and query arbitrary files where the schema can be defined by a regex.  The original intent was for this to be used for log files, however, it can be used for any structured data.
+
+Plugin for Apache Drill that allows Drill to read and query arbitrary files
+where the schema can be defined by a regex.  The original intent was for this
```
arina-ielchiieva:
```suggestion
where the schema can be defined by a regex. The original intent was for this
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
```diff
@@ -11,26 +18,50 @@ If you wanted to analyze log files such as the MySQL log sample shown below usin
...
-* **`regex`**:  This is the regular expression which defines how the log file lines will be split.  You must enclose the parts of the regex in grouping parentheses that you wish to extract.  Note that this plugin uses Java regular expressions and requires that shortcuts such as `\d` have an additional slash:  ie `\\d`.  This field is mandatory.
-* **`extension`**:  This option tells Drill which file extensions should be mapped to this configuration.  Note that you can have multiple configurations of this plugin to allow you to query various log files.  This field is mandatory.
-* **`maxErrors`**:  Log files can be inconsistent and messy.  The `maxErrors` variable allows you to set how many errors the reader will ignore before halting execution and throwing an error.  Defaults to 10.
-* **`schema`**:  The `schema` field is where you define the structure of the log file.  This section is optional.  If you do not define a schema, all fields will be assigned a column name of `field_n` where `n` is the index of the field. The undefined fields will be assigned a default data type of `VARCHAR`.
+
+* **`type`**:  This tells Drill which extension to use.  In this case, it must
```
arina-ielchiieva:
```suggestion
* **`type`**: This tells Drill which extension to use. In this case, it must
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
```diff
@@ -11,26 +18,50 @@ If you wanted to analyze log files such as the MySQL log sample shown below usin
...
+be `logRegex`.  This field is mandatory.
```
arina-ielchiieva:
```suggestion
be `logRegex`. This field is mandatory.
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
```diff
@@ -11,26 +18,50 @@ If you wanted to analyze log files such as the MySQL log sample shown below usin
...
+* **`regex`**:  This is the regular expression which defines how the log file
+lines will be split.  You must enclose the parts of the regex in grouping
```
arina-ielchiieva: Looks like everywhere in the doc there are two spaces before sentences instead of one. Could you please check and fix?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogBatchReader.java
```diff
@@ -0,0 +1,210 @@
...
+        rawColWriter.setString(line);
+        loadVectors(lineMatcher, rowWriter);
+        rowWriter.save();
+      }
+      return true;
+    }
+
+    errorCount++;
+    if (errorCount < maxErrors) {
+      logger.warn("Unmatached line: {}", line);
```
arina-ielchiieva:
```suggestion
      logger.warn("Unmatched line: {}", line);
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogBatchReader.java
```diff
@@ -0,0 +1,210 @@
...
+    } else {
+      throw UserException.parseError()
+          .message("Too many errors.  Max error threshold exceeded.")
```
arina-ielchiieva:
```suggestion
          .message("Too many errors. Max error threshold exceeded.")
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogBatchReader.java
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  private boolean nextLine(RowSetLoader rowWriter) {
+    String line;
+    try {
+      line = reader.readLine();
+    } catch (IOException e) {
+      throw UserException
+          .dataReadError(e)
+          .message("Error reading file:")
```
arina-ielchiieva:
```suggestion
          .message("Error reading file")
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogBatchReader.java
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  private void openFile(FileSchemaNegotiator negotiator) {
+    InputStream in;
+    try {
+      in = negotiator.fileSystem().open(split.getPath());
+    } catch (Exception e) {
+      throw UserException
+          .dataReadError(e)
+          .message("Failed to open open input file: %s", split.getPath())
```
arina-ielchiieva:
```suggestion
          .message("Failed to open input file: %s", split.getPath())
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -18,86 +18,224 @@
...
+    return builder;
   }
 
-  @Override
-  public int getWriterOperatorType() {
-    throw new UnsupportedOperationException("unimplemented");
+  /**
+   * Define the output schema: the schema after type conversions.
+   * Does not include the special columns as those are added only when
+   * requested, and are always VARCHAR.
```
arina-ielchiieva: Add `@param` and `@return` to javadoc

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -18,86 +18,224 @@
...
 
-  @Override
-  public boolean supportsStatistics() {
-    return false;
+  /**
+   * Define the simplified reader schema: this is the format that the reader
+   * understands. All columns are VARCHAR, and the reader can offer the
+   * two special columns.
+   */
+  private TupleMetadata defineReaderSchema(TupleMetadata outputSchema) {
```
arina-ielchiieva: Same here.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
```diff
@@ -389,4 +416,56 @@ public void testRawUMNoSchema() throws RpcException {
 
     RowSetUtilities.verify(expected, results);
   }
+
+  @Test
+  public void testProvidedSchema() throws Exception {
```
arina-ielchiieva: Could you please add unit tests to check how this format plugin works with schema parameter in table function?
Example: `org.apache.drill.TestSchemaWithTableFunction`
We might need to check two cases:
`select * from table(t(schema=>'inline=(col1 varchar)'))`
`select * from table(t(type=>'logRegex', schema=>'inline=(col1 varchar)'))`

paul-rogers: Short answer: it does not seem to work. I tried this in the past and found that table functions take only simple values (numbers, strings), not lists. Since this plugin uses a list, I never could figure out how to use it with table functions. In particular, how would the table function know how to create the instance of `LogFormatField` within the list? Am I missing something?

This plugin, in particular, would very much benefit from the use of a table function so that the user does not have to define a new plugin config for each new file type.

If there is a way to make this work, we can add the test and describe the answer in the README file.

paul-rogers:
This PR is now failing due to the Protobuf errors. Thanks @vvysotskyi for fixing them. I'll rebase onto that fix once it is committed and the reviewers have approved the commits.

paul-rogers:
Added the ability to specify the regex (and column schema) in the provided schema. Defined a table property for the regex. Although we can't (yet) use table properties to define the schema, we can now use `CREATE SCHEMA` to define both the regex and the schema.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
arina-ielchiieva: `select * from table(t(schema=>'inline=(col1 varchar)'))` should work disregarding format properties. But log format has schema property so I am wondering if there will be a clash or schema parameter will be correctly resolved, since log format has it as list and schema provisioning in string.
Since now log format supports schema provisioning, the above query should apply schema for log files, could you please check this query?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatField.java
```diff
@@ -18,35 +18,31 @@
...
+
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 
+
+/**
+ * The three configuration options for a field are:
+ * <ol>
+ * <li>The field name</li>
+ * <li>The data type (fieldType).  Field type defaults to VARCHAR
```
arina-ielchiieva: Extra space before `Field`

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
```diff
@@ -143,6 +143,14 @@ cardinality.
 You may find it helpful to specify the regex and column names via the plugin
 config, types via the `CREATE SCHEMA` command.
 
+## Table Functions
+
+Log files come in many forms. It would be very convenient to use Drill table
```
arina-ielchiieva: Table function will work for all log format properties, except of list. Knowing that list is not supported, does it makes sense to replace list schema parameter with String and rename it to avoid clash with schema parameter for schema provisioning.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
```diff
@@ -129,19 +129,62 @@ Drill 1.16 introduced the `CREATE SCHEMA` command to allow you to define the
...
-config, types via the `CREATE SCHEMA` command.
+### Provided Schema Provides The Regex
+
+Another way to use the provided schema is to define an empty plugin config; don't
+even provide the regex. Use table properties to define the regex (and the maximum
+error count, if desired.)
+
+In this case:
+
+* Set the table property `drill.regex.regex` to the desired pattern.
```
arina-ielchiieva: I think we should use different naming, `drill.regex.regex` look awkward. Maybe `drill.regex.pattern` or something like this?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogFormatPlugin.java
```diff
@@ -47,23 +47,30 @@
 import org.slf4j.LoggerFactory;
 
 public class LogFormatPlugin extends EasyFormatPlugin<LogFormatConfig> {
-  public static final String PLUGIN_NAME = "logRegex";
   private static final Logger logger = LoggerFactory.getLogger(LogFormatPlugin.class);
 
+  public static final String PLUGIN_NAME = "logRegex";
+  public static final String PROP_PREFIX = TupleMetadata.DRILL_PROP_PREFIX + "regex.";
```
arina-ielchiieva: Since properties are `log` specific should we add `log` in the properties naming as well as we did for `text` properties?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
paul-rogers: As I recall, Drill does not have a good way to deal with changes to the schema of a storage plugin. Some time back, I remember struggling to understand why my server would not start, only to eventually learn that some plugin or other changed its config and so Drill failed when trying to load the existing config from ZK. Has this been fixed?

If we change schema to a string, we'd need to run code to convert old configs. Also, we'd have the problem of what to do with the type property. We could not easily convert an existing config into a table schema.

Given these uncertainties, my thought was to leave the config alone and try to fit in the provided schema as best we can on top of the existing config.

Can you think of a better approach?

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
paul-rogers: Agree, it is pretty awkward. The saving grace is that I did, I believe, change "regex" to "logRegex" as you suggested. That is, the second item is the plugin "type" name.

When we worked on the text reader, I had first tried to choose good names for the third item. You rightly pointed out that it might be easier to remember if we simply use the existing config field names, which is what I did here.

So, even if the names are awkward, the pattern we've evolved is:

```
drill.<plugin type name>.<config field name>
```

That said, I'm open to suggestions if there is a better way to handle these names; now is the time to make improvements before folks deploy schema files with the names.

arina-ielchiieva: I think using `drill.logRegex.regex` will be fine.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/README.md
arina-ielchiieva: I guess initial choice of list property did not take into account that it does not work with table function. I don't think you can fix this backward compatibility in ZK but since this plugin is a role model for others I think it should have proper configuration, so changing `schema` property to be `String` instead `List` might be reasonable. Or can we have both properties one in String, another in `List<String>`? We can indicate in release notes that log plugin has been changed and config in ZK must be updated.

paul-rogers:
Rebased on latest master.

arina-ielchiieva:
@paul-rogers I am still unclear if you have tried the following query for log plugin data: `select * from table(t(schema=>'inline=(col1 varchar)'))` where `t` is table with log plugin data. Did you try it? I suppose it should work.

paul-rogers:
@arina-ielchiieva, I was able to get the plugin to work for this query:

```
SELECT * FROM table(dfs.tf.table1(
  type => 'logRegex',
  regex => '(\\d\\d\\d\\d)-(\\d\\d)-(\\d\\d) .*',
  maxErrors => 10))
```

To do this, I had to fix some of the issues described in DRILL-7298. In particular, DRILL-6672 notes that table functions are not able to call {{setFoo()}} methods as Jackson can, so table functions only work if the format plugin config fields are {{public}}. The were not public for the log format plugin, so I changed them to {{public}} to get the above query to work.

If we look at the code in [`FormatPluginOptionsDescriptor.createConfigForTable()`](https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FormatPluginOptionsDescriptor.java#L123), we'll see that there is nothing that would handle the `values` syntax suggested in your note. The only supported types are Java primitives.

When I tried this query:

```
SELECT * FROM table(dfs.tf.noGroups(
  type => 'logRegex',
  regex => '(\\d\\d\\d\\d)-(\\d\\d)-(\\d\\d) .*',
  `schema`=>values('month', 'VARCHAR')))
```

I got this result:

```
PARSE ERROR: Encountered "values" at line 1, column 115.

SQL Query: SELECT * FROM table(dfs.tf.noGroups(type => 'logRegex', regex => '(\\d\\d\\d\\d)-(\\d\\d)-(\\d\\d) .*', `schema`=>values('month', 'VARCHAR')))
                                                                                                                             ^
```

So, looks like the {{values}} trick does not work. Even if it did, the code to produce the values argument would use some kind of Java collection which would not match the {{List<LogFormatField>}} of the {{schema}} field.

paul-rogers:
One solution to the schema issue for table functions is to use the `columns` trick from the text reader. If no schema is provided, then instead of creating a set of `field_n` columns, create a single `columns` array column. Specifically, if there is no schema defined for the table, and no schema in the plugin config (perhaps because the plugin config was created via a table function), then just use `columns`.

If I get some time, I'll try this out. With the EVF, this might actually be pretty simple. Might be best to add such a feature via another PR.

arina-ielchiieva:
@paul-rogers thanks for making the changes. Regarding List parameter in table function, yes, it does not work and its a known issue, format plugins should not use list.
Though in my previous comments I have asked to try schema parameter in String not in List.
Adding example of the queries I expect should work once again:
1. `SELECT * FROM table(dfs.tf.noGroups(
  type => 'logRegex',
  regex => '(\\d\\d\\d\\d)-(\\d\\d)-(\\d\\d) .*',
  `schema`=>inline(`month` varchar)))
2. select * from table(t(schema=>'inline=(col1 varchar)'))

paul-rogers:
@arina-ielchiieva, thanks, I didn't realize we'd extended the table functions. Out of curiosity, how does the schema form know which plugin config to use? Or, does this form create a schema object and use the normal path for the plugin config?

Might we have this documented somewhere?

arina-ielchiieva:
@paul-rogers this was introduced in https://issues.apache.org/jira/browse/DRILL-6965 as part of schema provisioning project, Jira has the description and has doc-impacting label, hopefully it would be documented some day.
Yes, you are right plugin configs are initialized the usual way and once `DrillTable` instance is created, it is enriched with schema from schema parameter. All magic is done in `org.apache.drill.exec.store.AbstractSchema` class.

paul-rogers:
@arina-ielchiieva, added a unit test to show that the schema-only table function works.

Tried to create a test that combined a "plugin" table function with the "schema" attribute. This failed due to the unfortunate use of "schema" as plugin property name. You've pointed out this issue all along, I finally understood why it was a problem. Still, I'm reluctant to change the config property name for fear of breaking compatibility.

As it turns out, this limitation is only a minor nuisance since the only reason to combine the two kinds of table functions is to specify the regex property. A unit test shows that the regex can be specified as a table property instead.

Also, went ahead and added support for the `columns` column. If no schema is provided (not in the plugin config, not in a table function, not in a provided schema), then rather than creating a set of dummy fields `field_0`, `field_1`, etc., the plugin how follows the text format plugin and puts the fields into the `columns` array. The dummy fields are still used if the user specifies at least one column schema, but the regex has more groups than specified columns.

This means that, if the user uses a table function to specify just the regex, the user gets a reasonable result: the fields come back in the `columns` array.

Unit tests show the new `columns` array support.

paul-rogers:
Rebased on latest master. One unrelated test fails:

```
[ERROR] Errors: 
[ERROR]   TestDynamicUDFSupport.testDropFunction » UserRemote VALIDATION ERROR: From lin...
```

This test fails about 50% of the time, so it is probably not related to this change. The failure prevents running subsequent tests. But, mock data source support is likely not used by the other packages.

exec/java-exec/src/main/java/org/apache/drill/exec/store/log/LogBatchReader.java
```diff
@@ -43,60 +44,135 @@
...
     unmatchedColWriter = writer.scalar(UNMATCHED_LINE_COL_NAME);
+    saveMatchedRows = rawColWriter.isProjected();
 
     // If no match-case columns are projected, and the unmatched
     // columns is unprojected, then we want to count (matched)
     // rows.
 
     saveMatchedRows |= !unmatchedColWriter.isProjected();
+
+    // This reader is unusual: it can save only unmatched rows,
```
arina-ielchiieva: Not quite sure I understand meaning of such reader.

arina-ielchiieva:

paul-rogers:
Rebased on master. Addressed review comments. Squashed commits.

arina-ielchiieva:
+1

PR #1811, opened 2019-06-20
DRILL-7297: Query hangs in planning stage when Error is thrown

vvysotskyi:
Please see [DRILL-7297](https://issues.apache.org/jira/browse/DRILL-7297) for problem description.


amansinha100:
+1

PR #1817, opened 2019-06-29
DRILL-6225: Add support for boost 1.68 with openSSL 1.1.0/1.1.1 support

debraj92:
The change is to upgrade boost to the version 1.68. This ensures support for openssl 1.1.1 and 1.1.0.


arina-ielchiieva:
@debraj92 please squash the commits and address protobuf job failures - https://travis-ci.org/apache/drill/jobs/552040788.

arina-ielchiieva:
@debraj92 there are some failures on protobuf job. Could you please check?
https://travis-ci.org/apache/drill/jobs/552669860

dgrinchenko:
@debraj92 for me the change is unclear:
- the summary states about update to the libboost version 1.68, while PR updating it to 1.65
- travis-ci currently using Ubuntu Xenial as an LTS distribution, and libboost 1.65 strictly require Ubuntu Bionic (https://packages.ubuntu.com/search?keywords=libboost1.65-all-dev). Also, Ubuntu Xenial EOL is roughly in 2022 and i don't see, why we should remove support of this version of the Linux. Requirements for CentOS version and RedHat linux need to be verified too.
- why we changed expected function argument to be passed by reference?

debraj92:
Hi @dgrinchenko ,

Actually the change is incomplete because I ran into a problem with the travis building. Before explaining the problem, let me answer your questions and make the idea clear:

1. The purpose of this change is to enable support for openssl 1.1.0/1.1.1. This requires updating boost (to at least 1.64). So, I updated boost to the latest - 1.68. But, then I ran into problems with building using travis (as explained later) and I started experimenting with other versions of boost to see if the issue can be resolved. Thats why you see 1.65 there in the change. Ideally it should be 1.68.

2. Agreed, the changes in travis.yml are only experiments to get the build pass. Will explain the problem shortly.

3. Boost API has changed. The argument now needs to be passed by reference. This is the main purpose of the change. Without the change drill client code would not compile with boost 1.68.

Now, the problem. To get travis to build drill client code I would have to specify the location for downloading the boost 1.68 static libs. I could not find a repo from where I can do this and hence the change in the travis.yml file is not accurate. One option is to build boost from source but that would unnecessarily slow down the build. I also tried other boost versions (1.65, 1.67) but travis build did not succeed. Not sure how to proceed from here now.

dgrinchenko:
@debraj92 please read carefully point 2 of my comment, it describes why travis failing for you. [Here ](https://packages.ubuntu.com/xenial/libboost-all-dev)is the max version of libboost available for the travis. 

The another problem - if you decide to build own version of libboost, then it would need to be linked statically to the client, to not interference with already installed system version of libboost, otherwise your patch would change the minimal system version requirements for building and using drill client

debraj92:
Hi ,
I am closing this PR now because it looks like upgrading boost at this point is not feasible. I will leave the JIRA open and keep the patch locally so that we can apply it in the future.

PR #1822, opened 2019-07-05
DRILL-7317: Close ClassLoaders used for udf jars uploading when closing FunctionImplementationRegistry

vvysotskyi:
Please see [DRILL-7317](https://issues.apache.org/jira/browse/DRILL-7317) for problem description.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
@@ -29,6 +29,8 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -83,7 +85,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
  * <li><b>function holder for upper(VARCHAR-REQUIRED)</b> is {@link DrillFuncHolder} initiated for each function.</li>
  *
  */
-public class FunctionRegistryHolder {
+public class FunctionRegistryHolder implements AutoCloseable {
 
   private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FunctionRegistryHolder.class);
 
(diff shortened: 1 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -151,6 +151,20 @@ public void removeJar(String jarName) {
     }
   }
 
+  /**
+   * Removes all jars from {@link #jars} map and all associated with jars functions
+   * from {@link #functions}.
+   * This is write operation, so one user at a time can call perform such action,
+   * others will wait till first user completes his action.
+   */
+  public void removeAllJars() {
```
arina-ielchiieva: To improve performance please consider adding close method that will just close appropriate classloaders and then nullify map holders.

vvysotskyi: Good idea!

I have made this change, could you please review it again?

vvysotskyi:
The latest change caused some tests failures, I'll inform when they are resolved...

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -376,28 +377,78 @@ private void removeAllByJar(String jarName) {
...
+          for (Map.Entry<String, Queue<String>> functionEntry : jar.entrySet()) {
+            if (closeClassLoader(functionEntry.getKey(), functionEntry.getValue())) {
+              // class loader is closed, iterates to another jar
+              break;
+            }
+          }
+        }
+      });
+
+      jars.clear();
```
arina-ielchiieva: Giving a second thought we might consider not clearing the maps (better decision would be to nullify them but I assume we won't to remove final flag), then I suggest we leave them as is, since we are closing the Drillbit anyway.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -376,28 +377,78 @@ private void removeAllByJar(String jarName) {
...
+      functions.clear();
+    }
+  }
+
+  /**
+   * Produces search of {@link DrillFuncHolder} which corresponds to specified {@code String functionName}
+   * with signature from {@code Queue<String> functionSignatures},
+   * closes its class loader if {@link DrillFuncHolder} is found and returns true. Otherwise faalse is returned.
```
arina-ielchiieva:
```suggestion
   * closes its class loader if {@link DrillFuncHolder} is found and returns true. Otherwise false is returned.
```

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -376,28 +377,78 @@ private void removeAllByJar(String jarName) {
...
+   *
+   * @param functionName       name of the function
+   * @param functionSignatures function signatures
+   * @return returns first found {@link DrillFuncHolder} instance
+   */
+  private DrillFuncHolder getDrillFuncHolder(String functionName, Queue<String> functionSignatures) {
+    Map<String, DrillFuncHolder> functionHolders = functions.get(functionName);
+    Iterator<String> signaturesIterator = functionSignatures.iterator();
+    DrillFuncHolder drillFuncHolder = null;
+    while (signaturesIterator.hasNext() && drillFuncHolder == null) {
```
arina-ielchiieva: Consider using streams `findAny`.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -376,28 +377,78 @@ private void removeAllByJar(String jarName) {
...
+  }
+
+  /**
+   * Produces search of {@link DrillFuncHolder} which corresponds to specified {@code String functionName}
+   * with signature from {@code Queue<String> functionSignatures},
+   * closes its class loader if {@link DrillFuncHolder} is found and returns true. Otherwise faalse is returned.
+   *
+   * @param functionName       name of the function
+   * @param functionSignatures function signatures
+   * @return {@code true} if {@link DrillFuncHolder} is found and its class loader is closed
```
arina-ielchiieva: Not exactly, you return true if even there was a failure when closing class loader, more correct would be to describe as `return true if func holder was found and attempted to close class loader disregarding the result`.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
```diff
@@ -376,28 +377,78 @@ private void removeAllByJar(String jarName) {
...
-              ((AutoCloseable) classLoader).close();
-            } catch (Exception e) {
-              logger.warn("Problem during closing class loader", e);
-            }
-          }
-          break;
-        }
-      }
+      // closes class loader only one time
+      isClosed = !isClosed && closeClassLoader(function, functionSignatures);
```
arina-ielchiieva: There is only once class loader per jar, we can consider using `findAny` as I have suggested below.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
vvysotskyi: Agree, reverted this change.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
vvysotskyi: Tanks, fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
vvysotskyi: Thanks, replaced it with the sentence you have proposed.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
vvysotskyi: Thanks, it looks better with `findAny`.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/registry/FunctionRegistryHolder.java
vvysotskyi: That's correct that there is only one classloader per jar, but except closing it, we should also remove function signatures from `functions` map, so we need to iterate further.
With the current code, `closeClassLoader` method is called only once per jar.

arina-ielchiieva:

PR #1827, opened 2019-07-21
DRILL-7327: Log Regex Plugin Won't Recognize Schema

paul-rogers:
The previous commit revised the plugin config classes to work
with table functions. That caused Jackson to stop working for
the classess. Fixed those issues and added unit tests.


cgivre:
I built this branch and manually tested it. This fixes the issue.  @paul-rogers can you remove the banned dependency so we can commit?
Thanks for fixing this!

paul-rogers:
@cgivre, fixed the incorrect import.

PR #1829, opened 2019-07-27
DRILL-7096: Develop vector for canonical Map<K,V>

KazydubB:
- Added new type `TRUEMAP`
- Created value vectors for the type for single and repeated cases;
- Implemented corresponding `FieldReader`s and `FieldWriters`;
- Made changes in `EvaluationVisitor` to be able to read values from the map by key;
- Made changes to `DrillParquetGroupConverter` to be aware of Parquet's `MAP` type.


paul-rogers:
@KazydubB, when last this was discussed on the Drill dev mail list, there was a suggestion for the type name from Ted (can't recall the suggested name.) That name did seem a bit better than "true map". "True map" suggests that the other "map" is a "false map."

paul-rogers:
Looks like lots of work went into adapting the old-school complex readers and writers for this new type, but the newer column readers and writers don't have support. As a result, folks who want to use this feature need to use the old-style readers; the newer style "EVF" won't be usable for true maps.

While undoubtedly this implementation is exactly what one particular use case needs; I do wonder at the wisdom of creating pools of partially supported features that reflect which sponsor pays for the work.

Said another way, it would be very helpful, to encourage folks to create new readers, if the true map PR included EVF support.

paul-rogers:
here seems to be little documentation of the design or implementation: I find myself having to reverse engineer a design from the code. Would it be helpful for other reviewers if we included a bit of explanation?

In particular, what is the structure of the new vector? The true map vector seems to have a user-defined key and value type. Does this mean keys can be integers or maps? Do the proposed map functions handle non-Varchar keys? Do they handle repeated or map keys? If not, should the implementation restrict key types?

How do we handle varying value types? (I have a map with, say, {name: "Fred", balance: 123}. Would the user specify a Union? Have we fixed the many existing problems with the Union type? Or, is the true map meant to be like a Java map: Map<K,V>, and not a Python or JSON map with string keys and values of any type?

A bit of documentation (a README file, a package-info file or just a Javadoc explanation in the vector class) would help answer these questions.

paul-rogers:
This PR has no unit tests. While I am sure that this particular developer got everything 100% correct, I do wonder if future developers might accidentally introduce bugs that will only be caught by users in production. Without unit tests, we are performing without a safety net. Reviews must execute tests in our heads since there is no code to do it for us.

I realize that the sponsor probably really wants this code in, and will worry about tests later. But, as an open source project, I do thing it is a good practice to require unit tests for all new and changed code.

arina-ielchiieva:
@paul-rogers 
1. agree regarding naming  (it a least must have underscore `true_map`) but I agree that we need some better naming. canonical_map? Any other suggestions?
2. Support was added for older framework but I agree that for EVF it should be added as well. Since you knowledge keeper for EVF could you please provide some guidelines where changes need to be done? Its' still crucial to add support for older framework since currently only two readers use it (text and log format) and none of them support maps. This work was done primary for parquet reader (I am not sure when it will migrate to EVF). So to have maps in both is reasonable.
3. I see there are unit tests in this PR, maybe not the ones are expected, could you please expand and give examples which tests do you expect.
4. I believe there are some documentation in the Jiras but I agree that better java-doc or readme.md should have been provided.

paul-rogers:
@arina-ielchiieva, checked the dev archives: looks the previously suggested name was DICT. We could also allow DICTIONARY (similar to how we allow both VARCHAR and CHARACTER VARYING). The DICT name might be better since a MAP<K,V> is not like a generalized Python or JSON map (as we discussed.) We are actually defining a dictionary of keys and values in which the values must all be of the same type. I don't feel strongly about the name, but I did think it was a good idea when Ted suggested it.

As we can see, a huge amount of work went into this feature. (@KazydubB clearly did a fantastic job understanding all the bits and pieces.) I was kind of hoping we could handle the DICT type as just a bit of metadata sugar on top of the existing REPEATED MAP vector. This PR does do that, of course. Can we make the change with less impact?

The rough idea might be to create a new base class for both `RepeatedMap` and DICT. (Maybe we call it `AbstractRepeatedMap`). The structure would be identical for both, only the metadata bits would vary. A repeated map would allow adding any number of vectors, the DICT would take a `MaterializedField` that has the required key and value children in place, then create those two child vectors.

With this, any code that currently references `RepeatedMapVector` could change to reference `AbstractRepeatedMap` instead. Immediately, all existing code that works with repeated maps (including EVF) would work with `DICT`, especially when reading the vector. That is, if code does not really care about the meaning of the children, then it should not care if the vector is a `RepeatedMap`, or the more restricted form, a `DICT`. Flatten, for example, should work just as well for either case.

 The one part we'd want to add is the step to create the `DICT` vector and its `MaterializedField`. Presumably that code is in this PR for some cases.  For the RowSet and EVF stuff, we'd want to specify the `DICT<K,V>` info in metadata. We'd prohibit adding new children to a `DICT`. `RepeatedMap` would continue to allow adding new children. It may be that this is really the only difference.

Suppose we used the old-style complex writer or the newer column writer. The client should not care if it is writing a MAP or DICT: it will just work with fields. Here's the RowSet version treating a `DICT` as a repeated `MAP`:

```
TupleMetadata schema = ...
   // Convenience short-hand to assume the key is a VARCHAR, we specify the value type.
   .addDict("contacts", MinorType.VARCHAR) ...
// Create row set and writer here...
ArrayWriter dictWriter = rowSetWriter.array("contacts");
TupleWriter kvWriter = dictWriter.tuple();
kvWriter.scalar("key").setString("fred");
kvWriter.scalar("value").setString("555-1212");
dictWriter.save()
```

Similarly, for read:

```
RowSetReader reader = rowSetWithDict.reader();
ArrayReader dictReader = reader.array("contacts");
TupleReader kvReader = dictReader.tuple();
...
// On each row
while (dictReader.next()) {
  System.println(dictReader.scalar("key").getString() + ", " +
   dictReader.scalar("value").getString());
```

The above cleanly handles the case where the key and value types can be anything. The above works for values that are unions, maps, etc.

Also, the `RowSetBuilder` should just work to build a dictionary, and the verifier should also work to compare two row sets.

Let's compare the above idea with the current PR. In this PR, we do create a new vector, but we add a bit too much specialized semantics. Doing so then forces us to create new complex readers and writers. For the RowSet framework, we'd have to add new column readers and writers. And, since we are creating all this new code, we need unit tests for it.

That is, we are not exploiting the fact that a DICT really is a repeated MAP, just one with some restrictions. If, instead, we could add DICT as just a new way to create a kind of repeated MAP, and treat it as a repeated map for reads and writes, then we can leverage all the existing repeated map tests; we'd just add tests to ensure that the DICT creation code works.

To summarize, can we implement a DICT as mostly new metadata, and new mechanisms for vector creation, but leave vector writing and reading pretty much the same as a generic repeated map?

These are all just suggestions and brainstorming; perhaps @KazydubB already looked into this approach and found issues. Would be helpful to know what issue came up.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/TrueMapVector.java
```diff
@@ -0,0 +1,248 @@
...
+  private MajorType keyType;
+  private MajorType valueType;
+
+  public TrueMapVector(MaterializedField field, BufferAllocator allocator, CallBack callBack) {
+    super(field.clone(), allocator, callBack);
+    reader = new SingleTrueMapReaderImpl(TrueMapVector.this);
+  }
+
+  public TrueMapVector(MaterializedField field, BufferAllocator allocator, CallBack callBack, MajorType keyType, MajorType valueType) {
+    this(field.clone(), allocator, callBack);
```
ihuzenko: double clone ?

exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetComplex.java
```diff
@@ -220,4 +221,510 @@ public void testReadRepeatedDecimals() throws Exception {
         .baselineValues(ints, longs, fixedLen, fixedLen)
         .go();
   }
+
+  @Test
```
arina-ielchiieva: @KazydubB I see only unit tests for Parquet. Please add tests for  other formats in Drill that support maps. For example Avro: https://avro.apache.org/docs/current/spec.html#Maps

Also please adds tests to check that maps are written correctly for Parquet and Avro.

KazydubB:
Hi @paul-rogers, thank you for your elaborate answer.
This map is intended to be a map in Java's sense, with strict key and map types. The key type was indeed planned to be primitive only while the value could have been of any type either primitive or complex, but I accidentally removed the check when finalizing the changes :(. Just to be clear, key type is not limited to `VARCHAR`, but can be of any other, like `INT`, `BIGINT`, `FLOAT4` etc.

This new type is intended to be used for sources which support this type of maps, i.e. Parquet files, Hive tables etc.
Support for EVF was not considered because I thought it was not supporting readers which support such map type, but thank you for guidance, I'll look into it. Of course, it should be added.

The vector `extends` `RepeatedMapVector` and is essentially a repeated map vector (as you've noted) with constraints regarding its children and type of one of the child, 'key' (it `@Override`s `RepeatedMapVector.Accessor#getObject(int)` as well). This constraint is implemented in `TrueMapVector#putChild(String, ValueVector)` which dissalows the vector to have other children fields. Additionally, new reader and writer was introduced which also `extend` (inherit from) repeated map's reader and writer respectively. The new writer differs from its parent by how it handles children's offsets and introduces two essential methods, `startKeyValuePair()` and `endKeyValuePair()`, to separate entries. And new type's reader adds methods to find index for a given key and read a value for a given key into passed `ValueHolder`.
With this approach for writer, writing data into new map vector is not any different than writing data into repeated map (of course, if user (developer) is not trying to access "unknown" field, other than defined `"key"` and `"value"` on the writer for the new type). Flatten, for example, does behave the same for the two vectors. In places these two vectors (`TrueMapVector` and `RepeatedMapVector`) have different behavior, this is done in code; in other case the common behavior (logic) is used.
That being said, I do not see a reason to create another `AbstractRepeatedMap` as I think extending `RepeatedMapVector` suffices (unless we really need to separate these two vectors).

I am going to add documentation to code to avoid reverse engineering for developers (sorry for making you go through this!) and add unit tests for the new `ValueVector` and its writer and slightly refactor the code to implement aforementioned constraints. Also, I'll rename the type to `DICT`.

contrib/storage-hive/core/src/main/java/org/apache/drill/exec/store/hive/HiveUtilities.java
```diff
@@ -368,8 +368,13 @@ public static MajorType getMajorTypeFromHiveTypeInfo(final TypeInfo typeInfo, fi
         return typeBuilder.build();
       }
 
-      case LIST:
       case MAP:
+        return MajorType.newBuilder()
```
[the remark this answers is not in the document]
KazydubB: Yes, this change is not needed, didn't remove it. Thanks!

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchSizer.java
```diff
@@ -327,6 +327,7 @@ public boolean isVariableWidth() {
 
     public boolean isComplex() {
       return metadata.getType().getMinorType() == MinorType.MAP ||
```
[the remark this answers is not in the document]
KazydubB: :+1:

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
```diff
@@ -175,6 +183,38 @@ private static MessageType getProjection(MessageType schema,
...
+          if (field.getName().equalsIgnoreCase(fieldName)) {
+            fieldName = field.getName();
+            found = true;
+            break;
+          }
+        }
+        result = found ? groupType.getType(fieldName) : null;
+      } else {
+        // the segment is array index
+        if (groupType.getOriginalType() == OriginalType.LIST) {
```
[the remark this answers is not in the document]
KazydubB: It was made intentionally, but I see your point. Will combine.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/TrueMapVector.java
KazydubB: Good catch, thanks!

paul-rogers:
@KazydubB, thanks for the explanations. Probably not productive for us to continue to debate the design now that the code is done. Probably best to get this in so folks can use it; leaving refinements for later.

You correctly point out that EVF is used by only two readers, neither of which support a TRUEMAP. The remaining EVF reader, JSON, uses true maps (not a TRUEMAP) that can map string keys to any kind of value, so it will continue to need to use the (non-true) MAP type. So, we won't need EVF TRUEMAP support for that reader either. Instead, we can let a volunteer can add EVF support later when it might be needed.

The thought about a new base class simply recognized that a TRUEMAP does not have the same creation semantics as a Repeated MAP, so having TRUEMAP inherit from the leaf Repeated MAP type introduces very odd code hacks. (Like saying a Dog is a subclass of Cat, but without the purr and much more social.) But, again, I suppose we can fix that later if we have the need.

Still, I would suggest that, if you can spare a bit of time, that we add tests so we're not using the Parquet reader as our primary testing mechanism. The Parquet reader is quite complex and debugging this code in the context of the Parquet reader will be difficult.

KazydubB:
@ihuzenko I've addressed review comments, please take a look

exec/java-exec/src/main/codegen/templates/AbstractRecordWriter.java
```diff
@@ -71,6 +71,16 @@ public FieldConverter getNewRepeatedListConverter(int fieldId, String fieldName,
...
   }
 
+  @Override
+  public FieldConverter getNewDictConverter(int fieldId, String fieldName, FieldReader reader) {
+    throw new UnsupportedOperationException("Doesn't support writing TrueMap");
+  }
+
+  @Override
+  public FieldConverter getNewRepeatedDictConverter(int fieldId, String fieldName, FieldReader reader) {
+    throw new UnsupportedOperationException("Doesn't support writing RepeatedTrueMap");
```
arina-ielchiieva: Repeated Dict?

exec/java-exec/src/main/codegen/templates/AbstractRecordWriter.java
```diff
@@ -71,6 +71,16 @@ public FieldConverter getNewRepeatedListConverter(int fieldId, String fieldName,
     throw new UnsupportedOperationException("Doesn't support writing RepeatedList");
   }
 
+  @Override
+  public FieldConverter getNewDictConverter(int fieldId, String fieldName, FieldReader reader) {
+    throw new UnsupportedOperationException("Doesn't support writing TrueMap");
```
arina-ielchiieva: Dict?

exec/java-exec/src/main/java/org/apache/drill/exec/record/RecordBatchLoader.java
```diff
@@ -116,7 +116,7 @@ public boolean load(RecordBatchDef def, DrillBuf buf) throws SchemaChangeExcepti
 
         // If the field is a map, check if the map schema changed.
```
arina-ielchiieva: Update comment to reflect new logic.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -713,4 +715,63 @@ public static boolean containsComplexColumn(ParquetMetadata footer, List<SchemaP
...
+   * Converts list of {@link OriginalType}s to list of {@link org.apache.drill.common.types.TypeProtos.MajorType}s.
+   * <b>NOTE</b>: current implementation cares about {@link OriginalType#MAP} only
+   * converting it to {@link org.apache.drill.common.types.TypeProtos.MinorType#DICT}.
+   * Other original types are converted to {@code null}.
+   *
+   * @param originalTypes list of Parquet's types
+   * @return list containing either {@code null} or type with minor
+   *         type {@link org.apache.drill.common.types.TypeProtos.MinorType#DICT} values
+   */
+  public static List<TypeProtos.MajorType> getComplexTypes(List<OriginalType> originalTypes) {
```
arina-ielchiieva: It would be nice to explain why others converted to null...

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetReaderUtility.java
```diff
@@ -713,4 +715,63 @@ public static boolean containsComplexColumn(ParquetMetadata footer, List<SchemaP
...
+      }
+    }
+
+    return result;
+  }
+
+  /**
+   * Checks whether group field matches pattern for Logical Map type:
+   *
+   * <map-repetition> group <name> (MAP) {
```
arina-ielchiieva: Use proper html formatting.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -85,6 +84,22 @@
...
+  /**
+   * Name of nested group for Parquet's {@code MAP} type.
+   * @see <a href="https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps">MAP logical type</a>
+   */
+  private static final String GROUP_KEY_VALUE_NAME = "key_value";
+  /**
+   * Name of nested group for Parquet's {@code LIST} type.
+   * @see <a href="https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists">LIST logical type</a>
+   */
+  private static final String GROUP_LIST_NAME = "list";
```
arina-ielchiieva: Recently Igor was adding similar constants in https://github.com/apache/drill/pull/1844, please revisit.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
```diff
@@ -516,13 +495,46 @@ private static long convertToDrillDateValue(int dateValue) {
...
-        TypeProtos.MinorType leastRestrictiveType = TypeCastRules.getLeastRestrictiveType(Arrays.asList(majorType.getMinorType(), columnType.getMinorType()));
-        if (leastRestrictiveType != majorType.getMinorType()) {
-          columns.put(columnPath, columnType);
+      putType(columns, columnPath, columnType);
+    }
+    return columns;
+  }
+
+  /**
+   * Returns map of column names with their drill types for every {@code NameSegment} in {@code SchemaPath}
```
arina-ielchiieva:
```suggestion
   * Returns map of column names with their Drill types for every {@code NameSegment} in {@code SchemaPath}
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetTableMetadataUtils.java
```diff
@@ -573,19 +585,30 @@ public static OriginalType getOriginalType(MetadataBase.ParquetTableMetadataBase
...
-          }
-        }
-      });
+      fileColumns.forEach((columnPath, type) -> putType(columns, columnPath, type));
+    }
+    return columns;
+  }
+
+  static Map<SchemaPath, TypeProtos.MajorType> resolveIntermediateFields(MetadataBase.ParquetTableMetadataBase parquetTableMetadata) {
+    LinkedHashMap<SchemaPath, TypeProtos.MajorType> columns = new LinkedHashMap<>();
```
arina-ielchiieva: Can we use `Map<SchemaPath, TypeProtos.MajorType> columns = ...`?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/MetadataVersion.java
```diff
@@ -145,10 +145,16 @@ public int compareTo(MetadataVersion o) {
...
-   *  Version 4.0: Split the metadata cache file into summary and file metadata
-   */
-  public static final String V4 = "4.0";
+    /**
+     *  Version 4.0: Split the metadata cache file into summary and file metadata
+     */
+    public static final String V4 = "4.0";
+
+    /**
+     *  Version 4.1: Added parents' original types in {@link Metadata_V4.ColumnTypeMetadata_v4}
```
arina-ielchiieva: Please provide an example...

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -400,11 +405,23 @@ public Object deserializeKey(String key, com.fasterxml.jackson.databind.Deserial
...
    */
   public static class ColumnMetadata_v4 extends Metadata_V3.ColumnMetadata_v3 {
+
+    List<OriginalType> originalTypes;
+
     public ColumnMetadata_v4() {
     }
 
     public ColumnMetadata_v4(String[] name, PrimitiveType.PrimitiveTypeName primitiveType, Object minValue, Object maxValue, Long nulls) {
+      this(name, primitiveType, minValue, maxValue, nulls, new ArrayList<>());
```
arina-ielchiieva: Collections.emptyList()?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -400,11 +405,23 @@ public Object deserializeKey(String key, com.fasterxml.jackson.databind.Deserial
...
+    }
+
+    public ColumnMetadata_v4(String[] name, PrimitiveType.PrimitiveTypeName primitiveType, Object minValue, Object maxValue, Long nulls, List<OriginalType> originalTypes) {
       super(name, primitiveType, minValue, maxValue, nulls);
+      this.originalTypes = Collections.unmodifiableList(originalTypes);
```
arina-ielchiieva: `this.parentTypes = new ArrayList<>(parentTypes);` you wrap into list, here you wrap in unmodifiable... looks like different approach for similar cases...

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillMapGroupConverter.java
```diff
@@ -0,0 +1,111 @@
...
+    private static final int INDEX_KEY = 0;
+    private static final int INDEX_VALUE = 1;
+
+    KeyValueGroupConverter(OutputMutator mutator, GroupType schema, OptionManager options,
+                           ParquetReaderUtility.DateCorruptionStatus containsCorruptedDates) {
+      super(mutator, writer, options, containsCorruptedDates);
+
+      Type keyType = schema.getType(INDEX_KEY);
+      if (!keyType.isPrimitive()) {
+        throw new DrillRuntimeException("Map supports primitive key only. Found: " + keyType);
```
arina-ielchiieva: Map or Dict?

exec/java-exec/src/main/java/org/apache/drill/exec/vector/complex/FieldIdUtil.java
```diff
@@ -196,18 +206,48 @@ public static TypedFieldId getFieldId(ValueVector vector, int id, SchemaPath exp
...
+      MajorType vectorType = vector.getField().getType();
+      builder.intermediateType(vectorType);
+      builder.addId(id);
+      if (seg.isLastPath()) {
+        builder.finalType(vectorType);
+        return builder.build();
+      } else {
+        PathSegment child = seg.getChild();
+        if (!child.isArray()) {
+          // repeated map is accessed not by index, ignore?
```
arina-ielchiieva: ?

exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestDictVector.java
```diff
@@ -0,0 +1,460 @@
...
+
+      BaseReader.DictReader mapReader = mapVector.getReader();
+      index = 0;
+      for (Map<Object, Object> map : maps) {
+        mapReader.setPosition(index++);
+        for (Map.Entry<Object, Object> entry : map.entrySet()) {
+          mapReader.next();
+          Float actualKey = mapReader.reader(DictVector.FIELD_KEY_NAME).readFloat();
+          Long actualValue = mapReader.reader(DictVector.FIELD_VALUE_NAME).readLong();
+          Assert.assertEquals(entry.getKey(), actualKey);
```
arina-ielchiieva: Use static imports...

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/AbstractMapColumnMetadata.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.util.stream.Collectors;
+
+/**
+ * Describes a base column type for map, dict, repeated map and repeated dict. All are tuples that have a tuple
+ * schema as part of the column definition.
+ */
+public abstract class AbstractMapColumnMetadata extends AbstractColumnMetadata {
+
+  TupleMetadata parentTuple;
```
arina-ielchiieva: Why not protected?

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/AbstractRepeatedMapVector.java
```diff
@@ -0,0 +1,524 @@
...
+  }
+
+  public void allocateNew(int groupCount, int innerValueCount) {
+    clear();
+    try {
+      allocateOffsetsNew(groupCount);
+      for (ValueVector v : getChildren()) {
+        AllocationHelper.allocatePrecomputedChildCount(v, groupCount, 50, innerValueCount);
+      }
+    } catch (OutOfMemoryException e){
```
arina-ielchiieva:
```suggestion
    } catch (OutOfMemoryException e) {
```

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/DictVector.java
```diff
@@ -0,0 +1,321 @@
...
+ */
+public final class DictVector extends AbstractRepeatedMapVector {
+
+  public final static MajorType TYPE = Types.optional(MinorType.DICT);
+
+  public static final String FIELD_KEY_NAME = "key";
+  public static final String FIELD_VALUE_NAME = "value";
+  public static final List<String> fieldNames = Collections.unmodifiableList(Arrays.asList(FIELD_KEY_NAME, FIELD_VALUE_NAME));
+
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(DictVector.class);
```
arina-ielchiieva: Imports...

logical/src/main/java/org/apache/drill/common/expression/SchemaPath.java
```diff
@@ -74,6 +74,16 @@ public static SchemaPath getCompoundPath(String... strings) {
     return new SchemaPath(s);
   }
 
+  public static SchemaPath getCompoundPath(int to, String... strings) {
```
arina-ielchiieva: Please add java doc with example.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
```diff
@@ -179,6 +189,31 @@ private static MessageType getProjection(MessageType schema,
...
+   * @param segment segment which type will be returned
+   * @return type corresponding to the {@code segment} or {@code null} if there is no field found in {@code type}.
+   */
+  private static Type getType(Type type, PathSegment segment) {
+    Type result = null;
+    if (type != null && !type.isPrimitive()) {
+      GroupType groupType = type.asGroupType();
+      if (segment.isNamed()) {
+        String fieldName = segment.getNameSegment().getPath();
+        Optional<Type> foundType = groupType.getFields().stream()
```
vvysotskyi: Is it required to declare `foundType` here, since methods calls chain may be continued?

exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestDictVector.java
```diff
@@ -0,0 +1,460 @@
...
+  }
+
+  @Test
+  public void testVectorCreation() {
+    MaterializedField field = MaterializedField.create("map", DictVector.TYPE);
+    try (DictVector mapVector = new DictVector(field, allocator, null)) {
+      mapVector.allocateNew();
+
+      List<Map<Object, Object>> maps = Arrays.asList(
+          TestBuilder.mapOfObject(4f, 1L, 5.3f, 2L, 0.3f, 3L, -0.2f, 4L, 102.07f, 5L),
```
vvysotskyi: Please rework these tests to use generics if possible.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/AbstractRepeatedMapVector.java
```diff
@@ -0,0 +1,524 @@
...
+
+    public AbstractRepeatedMapTransferPair(T to) {
+      this(to, true);
+    }
+
+    public AbstractRepeatedMapTransferPair(T to, boolean allocate) {
+      this.from = (T) AbstractRepeatedMapVector.this;
+      this.to = to;
+      this.pairs = new TransferPair[from.size()];
+      // todo: this.to.ephPair
```
vvysotskyi: Please either address this `TODO` or remove it.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/AbstractRepeatedMapVector.java
```diff
@@ -0,0 +1,524 @@
...
+    public void exchange(ValueVector.Mutator other) {
+    }
+  }
+
+  @Override
+  public void clear() {
+    getMutator().reset();
+
+    offsets.clear();
+    for(final ValueVector vector : getChildren()) {
```
vvysotskyi:
```suggestion
    for (ValueVector vector : getChildren()) {
```

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/BaseRepeatedValueVector.java
```diff
@@ -250,6 +251,54 @@ public void exchange(ValueVector other) {
...
+      target.allocateNew();
+      for (int i = 0; i < length; i++) {
+        copyValueSafe(startIndex + i, i);
+      }
+    }
+
+    protected void copyValueSafe(int destIndex, int start, int end) {
+      final TransferPair vectorTransfer = children[1];
+      int newIndex = target.getOffsetVector().getAccessor().get(destIndex);
+      //todo: make this a bulk copy.
```
vvysotskyi:
```suggestion
      // TODO: make this a bulk copy.
```

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/DictVector.java
```diff
@@ -0,0 +1,321 @@
...
+ * @see SingleDictReaderImpl reader corresponding to the vector
+ * @see org.apache.drill.exec.vector.complex.impl.SingleDictWriter writer corresponding to the vector
+ */
+public final class DictVector extends AbstractRepeatedMapVector {
+
+  public final static MajorType TYPE = Types.optional(MinorType.DICT);
+
+  public static final String FIELD_KEY_NAME = "key";
+  public static final String FIELD_VALUE_NAME = "value";
+  public static final List<String> fieldNames = Collections.unmodifiableList(Arrays.asList(FIELD_KEY_NAME, FIELD_VALUE_NAME));
```
vvysotskyi: There is no need to wrap `Arrays.asList`

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedDictReaderImpl.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.exec.expr.holders.RepeatedDictHolder;
+import org.apache.drill.exec.vector.ValueVector;
+import org.apache.drill.exec.vector.complex.RepeatedDictVector;
+import org.apache.drill.exec.vector.complex.reader.FieldReader;
+import org.apache.drill.exec.vector.complex.writer.BaseWriter;
+
+public class RepeatedDictReaderImpl extends AbstractFieldReader {
+
+  private static final int NO_VALUES = Integer.MAX_VALUE - 1;
```
vvysotskyi: I think it would be better to use some negative value there.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedMapWriter.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override
+  public void start() {
+    // update the repeated vector to state that there is current+1 objects.
+    final RepeatedMapHolder h = new RepeatedMapHolder();
+    final RepeatedMapVector map = container;
+    final RepeatedMapVector.Mutator mutator = map.getMutator();
+
+    // Make sure that the current vector can support the end position of this list.
+    if(container.getValueCapacity() <= idx()) {
```
vvysotskyi:
```suggestion
    if (container.getValueCapacity() <= idx()) {
```

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedMapWriter.java
```diff
@@ -0,0 +1,60 @@
...
+      mutator.setValueCount(idx()+1);
+    }
+
+    map.getAccessor().get(idx(), h);
+    if (h.start >= h.end) {
+      container.getMutator().startNewValue(idx());
+    }
+    currentChildIndex = container.getMutator().add(idx());
+    for(final FieldWriter w : fields.values()) {
```
vvysotskyi:
```suggestion
    for (FieldWriter w : fields.values()) {
```

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/impl/RepeatedMapWriter.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  public void start() {
+    // update the repeated vector to state that there is current+1 objects.
+    final RepeatedMapHolder h = new RepeatedMapHolder();
+    final RepeatedMapVector map = container;
+    final RepeatedMapVector.Mutator mutator = map.getMutator();
+
+    // Make sure that the current vector can support the end position of this list.
+    if(container.getValueCapacity() <= idx()) {
+      mutator.setValueCount(idx()+1);
```
vvysotskyi:
```suggestion
      mutator.setValueCount(idx() + 1);
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/FileMetadataCollector.java
```diff
@@ -240,20 +240,23 @@ private void addColumnMetadata(String[] columnName,
...
       this.originalType = originalType;
+      this.parentTypes = parentTypes;
       this.precision = precision;
       this.scale = scale;
       this.repetitionLevel = repetitionLevel;
       this.definitionLevel = definitionLevel;
     }
 
-    static ColTypeInfo of(MessageType schema, Type type, String[] path, int depth) {
+    static ColTypeInfo of(MessageType schema, Type type, String[] path, int depth, List<OriginalType> originalTypes) {
```
ihuzenko: Please use consistent naming. Field and constructor param was named ```parentTypes``` but here  ```originalTypes``` .

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/Metadata_V4.java
```diff
@@ -311,7 +314,8 @@ public Long getRowCount() {
     public ColumnTypeMetadata_v4() {
     }
 
-    public ColumnTypeMetadata_v4(String[] name, PrimitiveType.PrimitiveTypeName primitiveType, OriginalType originalType, int precision, int scale, int repetitionLevel, int definitionLevel, long totalNullCount, boolean isInteresting) {
+    public ColumnTypeMetadata_v4(String[] name, PrimitiveType.PrimitiveTypeName primitiveType, OriginalType originalType, int precision, int scale, int repetitionLevel,
```
ihuzenko: 10 arguments, isn't it's time for builder ?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillMapGroupConverter.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.drill.exec.vector.complex.DictVector;
+import org.apache.drill.exec.vector.complex.writer.BaseWriter;
+import org.apache.drill.exec.vector.complex.writer.BaseWriter.DictWriter;
+import org.apache.parquet.io.api.Converter;
+import org.apache.parquet.schema.GroupType;
+import org.apache.parquet.schema.Type;
+
+import java.util.Collections;
+
+class DrillMapGroupConverter extends DrillParquetGroupConverter {
```
ihuzenko: Usually parquet primitive and group converter classes defined as a static nested classes inside ```DrillParquetGroupConverter```. I think we have two options to preserve consistency : 1) make this class static nested too or 2) make separate package and extract all static nested classes to it + move this one too.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillMapGroupConverter.java
```diff
@@ -0,0 +1,111 @@
...
+
+  private class KeyValueGroupConverter extends DrillParquetGroupConverter {
+
+    private static final int INDEX_KEY = 0;
+    private static final int INDEX_VALUE = 1;
+
+    KeyValueGroupConverter(OutputMutator mutator, GroupType schema, OptionManager options,
+                           ParquetReaderUtility.DateCorruptionStatus containsCorruptedDates) {
+      super(mutator, writer, options, containsCorruptedDates);
+
```
ihuzenko: I think it makes sense to extract creation of key and value converters to 2 private methods, then here code will be simplified to something like: 
```java

converters.add(createKeyConverter(...)); 
converters.add(createValueConverter(...));

```

agozhiy:
I think we have to support backward compatibility by adding an option. Users could already use the old data representation style in their queries, and now they will need to be rewritten.
For example, this won't work anymore:
```select t.c.`value` from (select flatten(x.fields.map) c from x```

File: exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java [hunk 38]
```diff
@@ -123,14 +126,22 @@ private static MessageType getProjection(MessageType schema,
...
     // get a list of modified columns which have the array elements removed from the schema path since parquet schema doesn't include array elements
+    // or if field is MAP then array/name segments are removed from the schema as well as obtaining elements by key is handled in EvaluationVisitor.
     List<SchemaPath> modifiedColumns = new LinkedList<>();
     for (SchemaPath path : columns) {
 
       List<String> segments = new ArrayList<>();
+      Type segmentType = schema;
       for (PathSegment seg = path.getRootSegment(); seg != null; seg = seg.getChild()) {
+
+        segmentType = getType(segmentType, seg);
```
ihuzenko: Please move down this and next line to be just before the ```if (isMap) {``` also add one line comment inside ```if``` describing why ```break;``` was called.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
```diff
@@ -179,6 +189,31 @@ private static MessageType getProjection(MessageType schema,
...
   }
 
+  /**
+   * Get type from the supplied {@code type} corresponding to given {@code segment}.
+   * @param type type to extract field corresponding to segment
+   * @param segment segment which type will be returned
+   * @return type corresponding to the {@code segment} or {@code null} if there is no field found in {@code type}.
+   */
+  private static Type getType(Type type, PathSegment segment) {
+    Type result = null;
```
ihuzenko: please rename local variable: 
```suggestion
    Type segmentType = null;
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
```diff
@@ -179,6 +189,31 @@ private static MessageType getProjection(MessageType schema,
     return projection;
   }
 
+  /**
+   * Get type from the supplied {@code type} corresponding to given {@code segment}.
+   * @param type type to extract field corresponding to segment
+   * @param segment segment which type will be returned
+   * @return type corresponding to the {@code segment} or {@code null} if there is no field found in {@code type}.
+   */
+  private static Type getType(Type type, PathSegment segment) {
```
ihuzenko: please rename method and parameter : 
```suggestion
  private static Type getSegmentType(Type parentSegmentType, PathSegment segment) {
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet2/DrillParquetReader.java
```diff
@@ -179,6 +189,31 @@ private static MessageType getProjection(MessageType schema,
...
+      if (segment.isNamed()) {
+        String fieldName = segment.getNameSegment().getPath();
+        Optional<Type> foundType = groupType.getFields().stream()
+            .filter(f -> f.getName().equalsIgnoreCase(fieldName))
+            .findAny();
+        result = foundType.map(field -> groupType.getType(field.getName()))
+            .orElse(null);
+      } else if (type.getOriginalType() == OriginalType.LIST) { // the segment is array index
+        // get element type of the list
+        result = groupType.getType(0).asGroupType().getType(0);
```
ihuzenko: Unsafe when list's element type isn't group. Such schema does not meet the logical list definition, but the case is possible.

exec/java-exec/src/main/java/org/apache/drill/exec/vector/complex/FieldIdUtil.java
```diff
@@ -58,8 +58,17 @@ public static TypedFieldId getFieldIdIfMatchesUnion(UnionVector unionVector, Typ
   }
 
 
-  public static TypedFieldId getFieldIdIfMatches(ValueVector vector, TypedFieldId.Builder builder, boolean addToBreadCrumb, PathSegment seg) {
-    if (vector instanceof RepeatedMapVector && seg != null && seg.isArray() && !seg.isLastPath()) {
+  public static TypedFieldId getFieldIdIfMatches(ValueVector vector, TypedFieldId.Builder builder,
```
ihuzenko: please add javadoc

exec/java-exec/src/test/java/org/apache/drill/exec/record/vector/TestDictVector.java
```diff
@@ -0,0 +1,460 @@
...
+          TestBuilder.mapOfObject(45f, 6L, 9.2f, 7L),
+          TestBuilder.mapOfObject(4.01f, 8L, 4f, 9L, -2.3f, 10L),
+          TestBuilder.mapOfObject(-2.5f, 11L),
+          TestBuilder.mapOfObject(),
+          TestBuilder.mapOfObject(11f, 12L, 9.73f, 13L, 4f, 14L)
+      );
+
+      BaseWriter.DictWriter mapWriter = new SingleDictWriter(mapVector, null);
```
ihuzenko: This code : 
```java
      BaseWriter.DictWriter mapWriter = new SingleDictWriter(mapVector, null);
      int index = 0;
      for (Map<Object, Object> map : maps) {
        mapWriter.setPosition(index++);
        mapWriter.start();
        for (Map.Entry<Object, Object> entry : map.entrySet()) {
          mapWriter.startKeyValuePair();
          mapWriter.float4(DictVector.FIELD_KEY_NAME).writeFloat4((float) entry.getKey());
          mapWriter.bigInt(DictVector.FIELD_VALUE_NAME).writeBigInt((long) entry.getValue());
          mapWriter.endKeyValuePair();
        }
        mapWriter.end();
      }
``` 
is repeated few times in the class. Please extract it to helper method named like ```writeMapsToVector```.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -357,6 +358,62 @@ public void testTableIsFoundWithinWorkspaceWhenNameStartsWithSlash() throws Exce
     }
   }
 
+  @Test
+  public void testCTASWithTrueMapInSelect() throws Exception {
```
ihuzenko: I believe @arina-ielchiieva's [comment](https://github.com/apache/drill/pull/1844#discussion_r315588844) is also applicable to this situation. In addition, please replace ```TrueMap``` with ```Dict``` in test method names.

exec/java-exec/src/test/java/org/apache/drill/exec/store/avro/AvroFormatTest.java
```diff
@@ -589,7 +602,169 @@ public void testCountStar() throws Exception {
         .sqlQuery(sql, file)
         .ordered()
         .baselineColumns("row_count")
-        .baselineValues((long)AvroTestUtil.RECORD_COUNT)
+        .baselineValues((long) RECORD_COUNT)
         .go();
   }
+
```
ihuzenko: could you please add one more test with getting by key using dot notation ?

exec/vector/src/main/codegen/templates/BaseReader.java
```diff
@@ -48,6 +49,14 @@
     int size();
     void copyAsValue(MapWriter writer);
   }
+
+  public interface DictReader extends RepeatedMapReader {
```
ihuzenko: Please provide javadoc for interface and methods. This is really very interesting to know difference between 
```java
    int find(String key);
    int find(int key);
```
and when to use them.

exec/vector/src/main/codegen/templates/NullReader.java
```diff
@@ -50,6 +50,10 @@ private NullReader(MajorType type) {
   public MajorType getType() {
     return type;
   }
+
+  public void read(ValueHolder holder) {
+    throw new UnsupportedOperationException("NullReader cannot write into non-nullable holder");
```
ihuzenko: I'm a little bit confused here, because nullable holders also implement the ```ValueHolder``` interface.

exec/vector/src/main/java/org/apache/drill/exec/expr/holders/RepeatedValueHolder.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.expr.holders;
+
+import org.apache.drill.exec.vector.complex.reader.FieldReader;
+
+public abstract class RepeatedValueHolder implements ValueHolder {
+
+  /** The first index (inclusive) into the Vector. **/
```
ihuzenko: Javadoc says about ```the Vector```, but vector is present in child classes. Maybe we can also move here a vector field too using generics ? 
```java
public abstract class RepeatedValueHolder<VECTOR> implements ValueHolder {
  public VECTOR vector; 
}
```

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/AbstractRepeatedMapVector.java
```diff
@@ -0,0 +1,524 @@
...
+  protected AbstractRepeatedMapVector(MaterializedField field, BufferAllocator allocator, CallBack callBack) {
+    this(field, new UInt4Vector(BaseRepeatedValueVector.OFFSETS_FIELD, allocator), callBack);
+  }
+
+  protected AbstractRepeatedMapVector(MaterializedField field, UInt4Vector offsets, CallBack callBack) {
+    super(field, offsets.getAllocator(), callBack);
+    this.offsets = offsets;
+    this.emptyPopulator = new EmptyValuePopulator(offsets);
+  }
+
```
ihuzenko: In my opinion it would be nice to group abstract methods together, place them before others and provide basic javadoc to help implementors in future.

exec/java-exec/src/main/java/org/apache/drill/exec/record/TypedFieldId.java
```diff
@@ -149,16 +156,44 @@ public static Builder newBuilder() {
...
+          .build();
+    }
+
+    public static TypedFieldId build(MajorType type, boolean isHyper, int... fieldIds) {
+      Builder builder = new Builder().intermediateType(type)
+          .finalType(type)
+          .secondaryFinal(type)
+          .addIds(fieldIds);
+      if (isHyper) {
+        builder.hyper();
```
ihuzenko: The ```hyper()``` could simply accept boolean.

exec/java-exec/src/main/java/org/apache/drill/exec/record/TypedFieldId.java
```diff
@@ -149,16 +156,44 @@ public static Builder newBuilder() {
...
-    boolean hyperReader = false;
-    boolean withIndex = false;
-    boolean isListVector = false;
-    Map<Integer, MajorType> types = new HashMap<>();
+    boolean hyperReader;
+    boolean withIndex;
+    boolean isListVector;
+    BitSet dictBitSet = new BitSet();
+
+    public static TypedFieldId build(MajorType type, int... fieldIds) {
```
ihuzenko: is it possible to get rid of overloaded ```build(...)``` methods completely ?

exec/java-exec/src/main/java/org/apache/drill/exec/record/TypedFieldId.java
```diff
@@ -38,47 +36,51 @@
...
+        .finalType(finalType)
+        .secondaryFinal(secondaryFinal)
+        .addIds(fieldIds)
+        .remainder(remainder)
+        .copyDictBitSet(dictBitSet);
+    if (isHyperReader) {
+      builder.hyper();
+    }
+    if (isListVector) {
+      builder.listVector();
```
ihuzenko: The listVector() could simply accept boolean.

exec/vector/src/main/java/org/apache/drill/exec/expr/holders/RepeatedValueHolder.java
```diff
@@ -0,0 +1,35 @@
...
+public abstract class RepeatedValueHolder<V extends ValueVector> implements ValueHolder {
+
+  /** The first index (inclusive) into the Vector. **/
+  public int start;
+
+  /** The last index (exclusive) into the Vector. **/
+  public int end;
+
+  /** The Vector holding the actual values. **/
+  public V vector;
```
vvysotskyi: Generics usage in `ValueHolder` class may be risky. Could you please confirm that there are no issues for bot JDK and Janino compilers for the code with generics?

KazydubB:
@ihuzenko I've updated the PR. @vvysotskyi yes, you're right. Reverted generics.

PR #1830, opened 2019-07-28
DRILL-7333: Set container record count where missing

paul-rogers:
Many operators fail to set the container record count. See DRILL-7333
for details. This commit fixes a large number of operators. Replaces
blocks of copy/paste code to set the container vector value count with
a call to a single function that does the work. That function also
sets the count for the container as a whole, something that many
operators failed to do, causing hard-to-find errors downstream.

Tests: reran all unit tests. A new unit test, to verify that the
count is set, will come in a later commit once all fixes are made.

Also contains code cleanup for items noticed while finding the container count issues.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/StreamingAggBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/StreamingAggBatch.java
@@ -75,12 +75,12 @@ import static org.apache.drill.exec.record.RecordBatch.IterOutcome.OK_NEW_SCHEMA
 import static org.apache.drill.exec.record.RecordBatch.IterOutcome.STOP;
 
 public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(StreamingAggBatch.class);
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(StreamingAggBatch.class);
 
   protected StreamingAggregator aggregator;
   protected final RecordBatch incoming;
   private List<BaseWriter.ComplexWriter> complexWriters;
-  //
+
   // Streaming agg can be in (a) a normal pipeline or (b) it may be in a pipeline that is part of a subquery involving
   // lateral and unnest. In case(a), the aggregator proceeds normally until it sees a group change or a NONE. If a
   // group has changed, the aggregated data is sent downstream and the aggregation continues with the next group. If
@@ -100,17 +100,17 @@ public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
   //   Schema change within a Data Set is not supported.
   //
   //   We will define some states for internal management
-  //
-  private boolean done = false;  // END of all data
+
+  private boolean done;          // END of all data
   private boolean first = true;  // Beginning of new data set. True during the build schema phase. False once the first
                                  // call to inner next is made.
-  private boolean sendEmit = false; // In the case where we see an OK_NEW_SCHEMA along with the end of a data set
-                                    // we send out a batch with OK_NEW_SCHEMA first, then in the next iteration,
-                                    // we send out an empty batch with EMIT.
+  private boolean sendEmit;      // In the case where we see an OK_NEW_SCHEMA along with the end of a data set
+                                 // we send out a batch with OK_NEW_SCHEMA first, then in the next iteration,
+                                 // we send out an empty batch with EMIT.
   private IterOutcome lastKnownOutcome = OK; // keep track of the outcome from the previous call to incoming.next
   private boolean firstBatchForSchema = true; // true if the current batch came in with an OK_NEW_SCHEMA
   private boolean firstBatchForDataSet = true; // true if the current batch is the first batch in a data set
-  private int recordCount = 0;  // number of records output in the current data set
+  private int recordCount;         // number of records output in the current data set
 
   private BatchSchema incomingSchema;
 
@@ -126,7 +126,7 @@ public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
    *
    * We maintain some state to remember that we have done such special handling.
    */
-  private boolean specialBatchSent = false;
+  private boolean specialBatchSent;
   private static final int SPECIAL_BATCH_COUNT = 1;
 
   // TODO: Needs to adapt to batch sizing rather than hardcoded constant value
@@ -177,7 +177,7 @@ public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
         break;
     }
 
-    this.incomingSchema = incoming.getSchema();
+    incomingSchema = incoming.getSchema();
     if (!createAggregator()) {
       state = BatchState.DONE;
     }
@@ -188,6 +188,7 @@ public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
     if (complexWriters != null) {
       container.buildSchema(SelectionVectorMode.NONE);
     }
+    container.setRecordCount(0);
   }
 
   @Override
@@ -207,6 +208,7 @@ public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
       firstBatchForDataSet = true;
       firstBatchForSchema = false;
       recordCount = 0;
+      container.setRecordCount(0);
       specialBatchSent = false;
       return EMIT;
     }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/StreamingAggBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/StreamingAggBatch.java
@@ -20,8 +20,6 @@ package org.apache.drill.exec.physical.impl.aggregate;
 import java.io.IOException;
 import java.util.List;
 
-import org.apache.drill.shaded.guava.com.google.common.annotations.VisibleForTesting;
-import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
 import org.apache.drill.common.exceptions.DrillRuntimeException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.expression.ErrorCollector;
@@ -63,10 +61,14 @@ import org.apache.drill.exec.vector.FixedWidthVector;
 import org.apache.drill.exec.vector.UntypedNullHolder;
 import org.apache.drill.exec.vector.UntypedNullVector;
 import org.apache.drill.exec.vector.ValueVector;
+import org.apache.drill.exec.vector.complex.writer.BaseWriter;
+import org.apache.drill.shaded.guava.com.google.common.annotations.VisibleForTesting;
+import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.sun.codemodel.JExpr;
 import com.sun.codemodel.JVar;
-import org.apache.drill.exec.vector.complex.writer.BaseWriter;
 
 import static org.apache.drill.exec.record.RecordBatch.IterOutcome.EMIT;
 import static org.apache.drill.exec.record.RecordBatch.IterOutcome.NONE;
@@ -75,7 +77,7 @@ import static org.apache.drill.exec.record.RecordBatch.IterOutcome.OK_NEW_SCHEMA
 import static org.apache.drill.exec.record.RecordBatch.IterOutcome.STOP;
 
 public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
-  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(StreamingAggBatch.class);
+  private static final Logger logger = LoggerFactory.getLogger(StreamingAggBatch.class);
 
   protected StreamingAggregator aggregator;
   protected final RecordBatch incoming;
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/FilterRecordBatch.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/FilterRecordBatch.java
@@ -76,51 +76,54 @@ public class FilterRecordBatch extends AbstractSingleRecordBatch<Filter> {
 
   @Override
   protected IterOutcome doWork() {
-    container.zeroVectors();
-    int recordCount = incoming.getRecordCount();
     try {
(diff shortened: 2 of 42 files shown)


paul-rogers:
Several tests fail due a race condition in Drillbit shutdown:

```
[ERROR]  TestUnionAll.testUnionAllEmptySides:348->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR] TestUnionDistinct.testUnionDistinctEmptySides:347->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR]  TestFastComplexSchema.test5:96->BaseTestQuery.testRunAndReturn:340 » Rpc org.a...
[ERROR] TestFlatten.testFlattenAfterJoin1:526->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR] TestFlatten.testFlattenAfterJoin2:537->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR]  TestFlatten.testFlattenAfterSort:514->BaseTestQuery.testRunAndReturn:340 » Rpc
```

A typical error stack:

```
[INFO] Running org.apache.drill.exec.store.parquet.ParquetInternalsTest
16:45:18.456 [BitClient-1] ERROR o.a.d.exec.rpc.RpcExceptionHandler - Exception in RPC communication.  Connection: /172.16.10.3:35130 <--> paul-linux/172.16.10.3:31075 (data client).  Closing connection.
java.lang.IllegalStateException: Attempting operation on allocator when allocator is closed.
Allocator(rpc:bit-data) 0/0/8256/9223372036854775807 (res/actual/peak/limit)
  child allocators: 0
  ledgers: 0
  reservations: 0

        at org.apache.drill.exec.memory.BaseAllocator.assertOpen(BaseAllocator.java:130) ~[drill-memory-base-1.17.0-SNAPSHOT.jar:1.17.0-SNAPSHOT]
        at org.apache.drill.exec.memory.BaseAllocator.buffer(BaseAllocator.java:209) ~[drill-memory-base-1.17.0-SNAPSHOT.jar:1.17.0-SNAPSHOT]
        at org.apache.drill.exec.memory.DrillByteBufAllocator.buffer(DrillByteBufAllocator.java:49) ~[drill-memory-base-1.17.0-SNAPSHOT.jar:1.17.0-SNAPSHOT]
        at org.apache.drill.exec.memory.DrillByteBufAllocator.ioBuffer(DrillByteBufAllocator.java:64) ~[drill-memory-base-1.17.0-SNAPSHOT.jar:1.17.0-SNAPSHOT]
        at io.netty.channel.AdaptiveRecvByteBufAllocator$HandleImpl.allocate(AdaptiveRecvByteBufAllocator.java:104) ~[netty-transport-4.0.48.Final.jar:4.0.48.Final]
        at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:117) ~[netty-transport-4.0.48.Final.jar:4.0.48.Final]
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:645) [netty-transport-4.0.48.Final.jar:4.0.48.Final]
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:580) [netty-transport-4.0.48.Final.jar:4.0.48.Final]
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:497) [netty-transport-4.0.48.Final.jar:4.0.48.Final]
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:459) [netty-transport-4.0.48.Final.jar:4.0.48.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:131) [netty-common-4.0.48.Final.jar:4.0.48.Final]
        at java.lang.Thread.run(Thread.java:748) [na:1.8.0_212]
```

This error is similar to the one that causes a UDF test to randomly fail. I'll try to track it down, but since the error is pre-existing, I'll not hold up this PR for that fix.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/aggregate/StreamingAggBatch.java
```diff
@@ -75,12 +75,12 @@
 import static org.apache.drill.exec.record.RecordBatch.IterOutcome.STOP;
 
 public class StreamingAggBatch extends AbstractRecordBatch<StreamingAggregate> {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(StreamingAggBatch.class);
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(StreamingAggBatch.class);
```
arina-ielchiieva: Please import `org.slf4j.Logger` and `org.slf4j.LoggerFactory`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/FilterRecordBatch.java
```diff
@@ -164,15 +167,14 @@ protected Filterer generateSV4Filterer() throws SchemaChangeException {
...
-
   }
 
   protected Filterer generateSV2Filterer() throws SchemaChangeException {
     final ErrorCollector collector = new ErrorCollectorImpl();
     final List<TransferPair> transfers = Lists.newArrayList();
     final ClassGenerator<Filterer> cg = CodeGenerator.getRoot(Filterer.TEMPLATE_DEFINITION2, context.getOptions());
-    // Uncomment below lines to enable saving generated code file for debugging
-    // cg.getCodeGenerator().plainJavaCapable(true);
+    cg.getCodeGenerator().plainJavaCapable(true);
```
arina-ielchiieva: Is this change needed? Maybe it was left by mistake...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
```diff
@@ -53,27 +53,27 @@
  * the Query's performance ,but just do a memory transfer by the later RemovingRecordBatch op.
  */
 public class RuntimeFilterRecordBatch extends AbstractSingleRecordBatch<RuntimeFilterPOP> {
-  private SelectionVector2 sv2;
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(RuntimeFilterRecordBatch.class);
```
arina-ielchiieva: Imports...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -1048,6 +1059,8 @@ public IterOutcome executeBuildPhase() throws SchemaChangeException {
           read_right_HV_vector = null;
         }
         break;
+      default:
+        throw new IllegalStateException(rightUpstream.name());
```
arina-ielchiieva: Maybe add more descriptive error message.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
```diff
@@ -1250,8 +1263,10 @@ public HashJoinBatch(HashJoinPOP popConfig, FragmentContext context,
     partitions = new HashPartition[0];
 
     // get the output batch size from config.
-    final int configuredBatchSize = (int) context.getOptions().getOption(ExecConstants.OUTPUT_BATCH_SIZE_VALIDATOR);
-    final double avail_mem_factor = (double) context.getOptions().getOption(ExecConstants.OUTPUT_BATCH_SIZE_AVAIL_MEM_FACTOR_VALIDATOR);
+    final int configuredBatchSize = (int) context.getOptions().getOption(
```
arina-ielchiieva: If option is long why we use int?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/project/ProjectorTemplate.java
```diff
@@ -97,6 +97,10 @@ public final void setup(FragmentContext context, RecordBatch incoming, RecordBat
     case TWO_BYTE:
       this.vector2 = incoming.getSelectionVector2();
       break;
+    case NONE:
+      break;
+    default:
+      throw new UnsupportedOperationException();
```
arina-ielchiieva: Please add error message.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/unorderedreceiver/UnorderedReceiverBatch.java
```diff
@@ -230,7 +230,9 @@ public void close() {
 
   @Override
   public VectorContainer getOutgoingContainer() {
-    throw new UnsupportedOperationException(String.format(" You should not call getOutgoingContainer() for class %s", this.getClass().getCanonicalName()));
+    throw new UnsupportedOperationException(
+        String.format("You should not call getOutgoingContainer() for class %s",
```
arina-ielchiieva: I think referring as `you` is not common standard for error messages, please re-write.

exec/java-exec/src/main/java/org/apache/drill/exec/record/AbstractUnaryRecordBatch.java
```diff
@@ -34,9 +34,9 @@
  * @param <T>
  */
 public abstract class AbstractUnaryRecordBatch<T extends PhysicalOperator> extends AbstractRecordBatch<T> {
-  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(new Object() {}.getClass().getEnclosingClass());
+  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(AbstractUnaryRecordBatch.class);
```
arina-ielchiieva: Imports.

arina-ielchiieva:
@paul-rogers 
1. travis job fail because there are some unused imports, could you please them and address code review comment so I can run full test suit for you branch.
2. 
> This commit fixes a large number of operators.

Are operators that are not fixed?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/FilterRecordBatch.java
paul-rogers: The rule is supposed to be that all operators that support "plain Java" declare that they can do so. This allows us, at some point, to flip a switch and use the Java compiler rather than Janino and byte code fixup. So, the "plainJavaCapable" is supposed to be uncommented. Only the "saveCodeForDebugging" should be commented. So, this was fixing an earlier error.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/filter/RuntimeFilterRecordBatch.java
paul-rogers: I wonder, rather than fixing these each time I mark a logger as private, should someone run a script to fix up all the files?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/HashJoinBatch.java
paul-rogers: This is original code, I just split the line for legibility. One could argue that a patch size larger than 2 GB would be excessive, I suppose.

paul-rogers:
@arina-ielchiieva, thank your for your review on a weekend! Addressed the minor comments.

This PR tackles only missing container and record count values.

These errors were found by enabling the "batch validator" and adding many rigorous consistency checks. The validator also checks vectors: there are hundreds of errors in which counts are wrong, offset vectors are invalid and so on. Those will be done as a separate PR to avoid overwhelming the reviewer (or myself.) This is also why the batch validator itself is not part of this PR: it would complain about the invalid vectors.

In starting on the vector fixes, many of these same operators are fixed again, along with a few others.

An earlier comment mentioned that I saw a number of RPC errors due to what seems like a race condition in closing allocators. Oddly, it turns out that Travis did not see those errors, but instead found an out-of-bounds error.

When tests run under Maven, we turn on checking of vector bounds. I discovered that bounds checks are turned off in the IDE, which is why I didn't catch the issue. I'll follow up with another commit for that fix. (That commit will also turn on bounds checking if assertions are enabled so that bounds checking occurs in IDE test runs.)

paul-rogers:
Addressed review comments. Still getting failures when running tests in Maven, but not Eclipse. I guess we should hold off committing this until I track down the issues:

```
[ERROR]   TestUnionAll.testUnionAllEmptySides:348->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR]   TestUnionDistinct.testUnionDistinctEmptySides:347->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR]   TestFastComplexSchema.test5:96->BaseTestQuery.testRunAndReturn:340 » Rpc org.a...
[ERROR]   TestFlatten.testFlattenAfterJoin1:526->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR]   TestFlatten.testFlattenAfterJoin2:537->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR]   TestFlatten.testFlattenAfterSort:514->BaseTestQuery.testRunAndReturn:340 » Rpc
[ERROR]   TestDynamicUDFSupport.testReRegisterTheSameJarWithDifferentContent:600->BaseTestQuery.testRunAndReturn:340 » Rpc
```

arina-ielchiieva:
@paul-rogers mostly rpc errors are caused by other tests failures.
Currently if we look in travis there are still IOB exceptions (https://travis-ci.org/apache/drill/jobs/564781376).

paul-rogers:
@arina-ielchiieva, thanks much for the reviews and the suggestions. I would tend to doubt that the RPC errors I'm seeing are due to either lack of resources or other test failures. For example, the UDF tests fail about half the time when I do builds, even clean builds on master. Maven controls resources, the Dynamic UDF tests are the only ones that fail, and my machine has 32 GB of memory; far more than needed. Something odd is going on; I just don't know what.

In any event, I'll take another crack at tracking down the issue. If that does not work, I'll back out the Flatten and selected other changes until I get a clean build. I'll add Flatten to a later commit once I figure out what's what. Will try to complete this work by next Monday.

paul-rogers:
Going to abandon this. Turns out that Drill has many, many interlocking vector count and size errors. I've not found a good way to work on them in chunks, and the problem is too large to do in one go. (I now understand what Jinfeng was facing a couple of years ago in his "empty batches" project!) Since I'm running out of time to work on this, we'll just leave well enough alone.

Started down this route because the revised JSON reader hit a number of these problems as it used the "right" values which broke code that expected the wrong values. Since there is no time to work on fixing this, it means that the JSON reader can't be offered. Will revisit later if time becomes available.

PR #1834, opened 2019-08-01
DRILL-7335: Fix error when reading csv file with headers only

arina-ielchiieva:
Jira - [DRILL-7335](https://issues.apache.org/jira/browse/DRILL-7335).


arina-ielchiieva:
@paul-rogers could you please review?

PR #1835, opened 2019-08-03
DRILL-7337: Add vararg UDFs support

vvysotskyi:
For details please see [DRILL-7337](https://issues.apache.org/jira/browse/DRILL-7337).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillFuncHolder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillFuncHolder.java
@@ -17,6 +17,12 @@
  */
 package org.apache.drill.exec.expr.fn;
 
+import com.sun.codemodel.JAssignmentTarget;
+import com.sun.codemodel.JCodeModel;
+import com.sun.codemodel.JExpression;
+import com.sun.codemodel.JInvocation;
+import org.apache.drill.exec.expr.holders.RepeatedListHolder;
+import org.apache.drill.exec.expr.holders.ValueHolder;
 import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
 import org.apache.drill.shaded.guava.com.google.common.base.Strings;
 import com.sun.codemodel.JBlock;
@@ -119,9 +125,12 @@ public abstract class DrillFuncHolder extends AbstractFuncHolder {
     return attributes.isNiladic();
   }
 
-
   public boolean isInternal() {
     return attributes.isInternal();
+  }
+
+  public boolean isVarArg() {
+    return attributes.isVarArg();
   }
 
   /**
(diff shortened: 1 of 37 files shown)


cgivre:
This is a really useful feature!

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillFuncHolder.java
```diff
@@ -262,6 +260,78 @@ protected void addProtectedBlock(ClassGenerator<?> g, JBlock sub, String body, H
...
 
+  /**
+   * Checks whether current function is var arg one and declares array for storing vararg function arguments.
+   *
+   * @param model          code model to generate the code
+   * @param jBlock         block of code to be populated
+   * @param inputVariables array of input variables for current function
+   */
+  protected void declareVarArgArray(JCodeModel model, JBlock jBlock, HoldingContainer[] inputVariables) {
+    if (isVarArg()) {
```
arina-ielchiieva: Not sure if this check should be done here, I would assume its caller responsibility first check `isVarArg()` and then call `declareVarArgArray`.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/DrillFuncHolder.java
```diff
@@ -298,11 +368,19 @@ public int getParamCount() {
   }
 
   public boolean isConstant(int i) {
-    return attributes.getParameters()[i].isConstant();
+    return getAttributeParameter(i).isConstant();
+  }
+
+  public ValueReference getAttributeParameter(int i) {
```
arina-ielchiieva: Please add java doc thus there will be no need in the comment in the code.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/fn/FunctionConverter.java
```diff
@@ -93,18 +96,30 @@
       // TODO(Julien): verify there are a few of those and we can load them
       Class<?> fieldClass = field.getFieldClass();
       if (param != null || output != null) {
+        if (Object[].class.isAssignableFrom(fieldClass)) {
+          fieldClass = fieldClass.getComponentType();
+          varArgsCount++;
+        } else if (varArgsCount > 0 && param != null) {
+          return failure("Vararg should be the last argument of function.", func, field);
```
arina-ielchiieva:
```suggestion
          return failure("Vararg should be the last argument in the function.", func, field);
```

arina-ielchiieva:
Also could you please show how varargs functions will be described in `sys.functions` table. Do we need new flag to filter them out?
Example:
```
apache drill (sys)> select * from sys.functions limit 2;
+------+---------------------------------+------------+----------+----------+
| name |            signature            | returnType |  source  | internal |
+------+---------------------------------+------------+----------+----------+
| !=   | BIGINT-REQUIRED,BIGINT-REQUIRED | BIT        | built-in | false    |
| !=   | BIT-REQUIRED,BIT-REQUIRED       | BIT        | built-in | false    |
+------+---------------------------------+------------+----------+----------+
```

arina-ielchiieva:

PR #1836, opened 2019-08-05
DRILL-7156: Support empty Parquet files creation

oleg-zinovev:
PR for Drill empty parquet  files read and write support.

Known limitations:
1) Not working for hive parquet for now
2) Ignores all schemas except last while writing empty parquet file
3) Not support empty schemas (e.g. create table `..` as select * from `empty.json, e.g. {}`)

Short changes description:
1) Parquet footer metadata added
2) Parquet writer checks that at least 1 row has been written. If not - creates a empty parquet file with footer.
3) EmptyParquetRowGroupScan and EmptyParquetScanBatchCreator added

Questions:
1) TestParquetWriterEmptyFiles#testMultipleWriters now creates several empty files, but not fails, since reading of empty parquet is supported. Should I rewrite comment or remove the test?

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
@@ -122,6 +122,9 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
   private PrimitiveTypeName logicalTypeForDecimals;
   private boolean usePrimitiveTypesForDecimals;
 
+  /** Is used to ensure that empty Parquet file will be written if no rows were provided. */
+  private boolean empty = true;
+
   public ParquetRecordWriter(FragmentContext context, ParquetWriter writer) throws OutOfMemoryException {
     this.oContext = context.newOperatorContext(writer);
     this.codecFactory = CodecFactory.createDirectCodecFactory(writer.getFormatPlugin().getFsConf(),
@@ -205,7 +208,7 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
   public void updateSchema(VectorAccessible batch) throws IOException {
     if (this.batchSchema == null || !this.batchSchema.equals(batch.getSchema()) || containsComplexVectors(this.batchSchema)) {
       if (this.batchSchema != null) {
-        flush();
+        flush(false);
       }
       this.batchSchema = batch.getSchema();
       newSchema();
@@ -310,7 +313,7 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
     try {
       boolean newPartition = newPartition(index);
       if (newPartition) {
-        flush();
+        flush(false);
         newSchema();
       }
     } catch (Exception e) {
@@ -318,19 +321,18 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
     }
   }
 
-  private void flush() throws IOException {
+  private void flush(boolean cleanUp) throws IOException {
     try {
       if (recordCount > 0) {
-        parquetFileWriter.startBlock(recordCount);
-        consumer.flush();
-        store.flush();
-        pageStore.flushToFileWriter(parquetFileWriter);
-        recordCount = 0;
-        parquetFileWriter.endBlock();
-
-        // we are writing one single block per file
-        parquetFileWriter.end(extraMetaData);
-        parquetFileWriter = null;
+        flushParquetFileWriter();
+      } else if (cleanUp && empty && schema != null && schema.getFieldCount() > 0) {
+        // Write empty parquet if:
+        // 1) This is a cleanup - no any additional records can be written
+        // 2) No file was written until this moment
+        // 3) Schema is set
+        // 4) Schema is not empty
+        createParquetFileWriter();
+        flushParquetFileWriter();
       }
     } finally {
       store.close();
@@ -347,7 +349,7 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
       long memSize = store.getBufferedSize();
       if (memSize > blockSize) {
         logger.debug("Reached block size " + blockSize);
-        flush();
+        flush(false);
         newSchema();
         recordCountForNextMemCheck = min(max(MINIMUM_RECORD_COUNT_FOR_CHECK, recordCount / 2), MAXIMUM_RECORD_COUNT_FOR_CHECK);
       } else {
@@ -435,29 +437,10 @@ public class ParquetRecordWriter extends ParquetOutputRecordWriter {
 
     // we wait until there is at least one record before creating the parquet file
     if (parquetFileWriter == null) {
-      Path path = new Path(location, prefix + "_" + index + ".parquet");
-      // to ensure that our writer was the first to create output file, we create empty file first and fail if file exists
-      Path firstCreatedPath = storageStrategy.createFileAndApply(fs, path);
-
-      // since parquet reader supports partitions, it means that several output files may be created
-      // if this writer was the one to create table folder, we store only folder and delete it with its content in case of abort
-      // if table location was created before, we store only files created by this writer and delete them in case of abort
-      addCleanUpLocation(fs, firstCreatedPath);
-
-      // since ParquetFileWriter will overwrite empty output file (append is not supported)
-      // we need to re-apply file permission
-      if (useSingleFSBlock) {
-        // Passing blockSize creates files with this blockSize instead of filesystem default blockSize.
-        // Currently, this is supported only by filesystems included in
-        // BLOCK_FS_SCHEMES (ParquetFileWriter.java in parquet-mr), which includes HDFS.
-        // For other filesystems, it uses default blockSize configured for the file system.
-        parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE, blockSize, 0);
-      } else {
-        parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE);
-      }
-      storageStrategy.applyToFile(fs, path);
-      parquetFileWriter.start();
+      createParquetFileWriter();
     }
+
+    empty = false;
     recordCount++;
     checkBlockSizeReached();
   }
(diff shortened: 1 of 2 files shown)


oleg-zinovev:
@arina-ielchiieva, could you please review?

arina-ielchiieva:
@oleg-zinovev thanks for making the changes, though the situation is little bit awkward, since I was working on similar changes and did not know you intend to do them as well (https://issues.apache.org/jira/browse/DRILL-4517). Though I was working on reading empty parquet files but not writing them. I suggest you separate out writing empty parquet files into separate PR as for reading it might be better if my changes will be used instead: first you change metadata cache files and this would affect backward compatibility as well as will have to store more information than needed, secondly your changes does not seem to optimize reading complex types. What do you think?

oleg-zinovev:
Ok, I'll try to rewrite the commit within a week.

arina-ielchiieva:
@oleg-zinovev thanks, I have assigned https://issues.apache.org/jira/browse/DRILL-7156 to you.

arina-ielchiieva:
@oleg-zinovev support empty parquet files reading is already merged into master (https://github.com/apache/drill/commit/4f4e1af53c9abccd1996f3b6841731e68768b48e). Do you plan on working on adding support for writing empty parquet files? We plan to include it in next Drill release (end of August / beginning of September). If yes, please factor out writing empty parquet and update the PR.

oleg-zinovev:
> @oleg-zinovev support empty parquet files reading is already merged into master ([4f4e1af](https://github.com/apache/drill/commit/4f4e1af53c9abccd1996f3b6841731e68768b48e)). Do you plan on working on adding support for writing empty parquet files? We plan to include it in next Drill release (end of August / beginning of September). If yes, please factor out writing empty parquet and update the PR.

Yes, I think I’ll do it today.

arina-ielchiieva:
Regarding this comment

> Questions:
TestParquetWriterEmptyFiles#testMultipleWriters now creates several empty files, but not fails, since reading of empty parquet is supported. Should I rewrite comment or remove the test?

I guess you can remove these tests and add new tests into `org.apache.drill.exec.store.parquet.TestEmptyParquet`.

arina-ielchiieva:
> Ignores all schemas except last while writing empty parquet file
Please provide example...
> Not support empty schemas (e.g. create table .. as select * from empty.json, e.g. {})
What behavior will be in this case? Failure? No-op?

oleg-zinovev:
> What behavior will be in this case? Failure? No-op?
Remained unchanged (No-op)

oleg-zinovev:
> Please provide example...

I added a test TestParquetWriterEmptyFiles#testWriteEmptySchemaChange. As you can see, there is no "a" field in the written schema. 

Probably, it would be correct to write schema for all empty scans, but it will lead to writing "garbage" empty parquet files, if the scan with data is at the end of batch.

arina-ielchiieva:
Is there a way to write combined schema in this case?

oleg-zinovev:
> Please provide example...



> Is there a way to write combined schema in this case?

@arina-ielchiieva , thank you for your comment. 

I can try to make a combined scheme, but:
- What type of field should be written if the first BatchSchema contains field "A" with type "bigint", and the second - field "A" with type "varchar"? The last one wins?
- What about union vectors?

arina-ielchiieva:
Good questions, you can investigate how now union types are handled. Regarding who wins, maybe you can look into `org.apache.drill.exec.physical.impl.union.UnionAllRecordBatch` to see how it Cretes combined schema using precedence rules.

oleg-zinovev:
> Good questions, you can investigate how now union types are handled. Regarding who wins, maybe you can look into `org.apache.drill.exec.physical.impl.union.UnionAllRecordBatch` to see how it Cretes combined schema using precedence rules.

@arina-ielchiieva , thanks for your advice. 
I will try to make a combined scheme by the end of the week.

oleg-zinovev:
@arina-ielchiieva 
Some time spent debugging the test showed that the last schema contains all fields. The field is added in ProjectRecordBatch#setupNewSchemaFromInput. 
In the original version of the test, field A was not added due to plan optimization - condition `1=0` was replaced by` limit 0`

I can still provide a solution with combining schema if required.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -486,11 +467,54 @@ public void abort() throws IOException {
...
+      parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE, blockSize, 0);
+    } else {
+      parquetFileWriter = new ParquetFileWriter(conf, schema, path, ParquetFileWriter.Mode.OVERWRITE);
+    }
+    storageStrategy.applyToFile(fs, path);
+    parquetFileWriter.start();
+  }
+
+  private void flushParquetFileWriter() throws IOException {
+    assert parquetFileWriter != null;
```
arina-ielchiieva: Same here.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -486,11 +467,54 @@ public void abort() throws IOException {
...
   @Override
   public void cleanup() throws IOException {
-    flush();
+    flush(true);
 
     codecFactory.release();
   }
 
+  private void createParquetFileWriter() throws IOException {
+    assert parquetFileWriter == null;
```
arina-ielchiieva: Please remove, there is no need for this check.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -310,27 +312,26 @@ public void checkForNewPartition(int index) {
...
+        flush(false);
         newSchema();
       }
     } catch (Exception e) {
       throw new DrillRuntimeException(e);
     }
   }
 
-  private void flush() throws IOException {
+  private void flush(final boolean cleanUp) throws IOException {
```
arina-ielchiieva:
```suggestion
  private void flush(boolean cleanUp) throws IOException {
```

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriterEmptyFiles.java
```diff
@@ -43,47 +49,99 @@ public void testWriteEmptyFile() throws Exception {
...
       // this specific value will force a flush just after the final row is written
       // this may cause the creation of a new "empty" parquet file
       test("ALTER SESSION SET `store.parquet.block-size` = 19926");
 
       final String query = "SELECT * FROM cp.`employee.json` LIMIT 100";
-      test("CREATE TABLE dfs.tmp.%s AS %s", outputFile, query);
+      test("CREATE TABLE dfs.tmp.%s AS %s", outputFileName, query);
+
+      // Make sure that only 1 parquet file was created
+      Assert.assertEquals(1, outputFile.list((dir, name) -> name.endsWith("parquet")).length);
```
arina-ielchiieva: Please use static import.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriterEmptyFiles.java
```diff
@@ -43,47 +49,99 @@ public void testWriteEmptyFile() throws Exception {
...
+            .build();
+
+    testBuilder()
+            .unOrdered()
+            .sqlQuery("select * from dfs.tmp.%s", outputFileName)
+            .schemaBaseLine(expectedSchema)
+            .go();
   }
 
   @Test // see DRILL-2408
```
arina-ielchiieva: Please remove `see DRILL-2408` references in the class, now the make no sense.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriterEmptyFiles.java
```diff
@@ -43,47 +49,99 @@ public void testWriteEmptyFile() throws Exception {
...
-      testBuilder()
-        .unOrdered()
-        .sqlQuery("SELECT * FROM dfs.tmp.%s", outputFile)
-        .sqlBaselineQuery(query)
-        .go();
-    } finally {
-      runSQL("alter session set `planner.slice_target` = " + ExecConstants.SLICE_TARGET_DEFAULT);
-    }
+    // Make sure that only 1 parquet file was created
+    Assert.assertEquals(1, outputFile.list((dir, name) -> name.endsWith("parquet")).length);
```
arina-ielchiieva: Static import.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriterEmptyFiles.java
```diff
@@ -43,47 +49,99 @@ public void testWriteEmptyFile() throws Exception {
...
+  }
+
+  @Test
+  public void testEmptyFileSchema() throws Exception {
+    final String outputFileName = "testparquetwriteremptyfiles_testemptyfileschema";
+
+    test("CREATE TABLE dfs.tmp.%s AS SELECT * FROM cp.`employee.json` WHERE 1=0", outputFileName);
```
arina-ielchiieva: Please replace select from JSON file with select from Parquet, we need to test that schema is created correctly from already known schema.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -122,6 +122,8 @@
   private PrimitiveTypeName logicalTypeForDecimals;
   private boolean usePrimitiveTypesForDecimals;
 
+  private boolean empty = true;
```
arina-ielchiieva: Please add comment describing the purpose of this flag.

arina-ielchiieva:
> @arina-ielchiieva
> Some time spent debugging the test showed that the last schema contains all fields. The field is added in ProjectRecordBatch#setupNewSchemaFromInput.
> In the original version of the test, field A was not added due to plan optimization - condition `1=0` was replaced by` limit 0`
> 
> I can still provide a solution with combining schema if required.

This case we don't need schema merge,

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriterEmptyFiles.java
oleg-zinovev: If needed, I can also rewrite texts using RowSet.

oleg-zinovev:
@arina-ielchiieva , thanks for review. Fixed.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriterEmptyFiles.java
```diff
@@ -18,23 +18,33 @@
...
 public class TestParquetWriterEmptyFiles extends BaseTestQuery {
 
   @BeforeClass
   public static void initFs() throws Exception {
     updateTestCluster(3, null);
+    dirTestWatcher.copyResourceToRoot(Paths.get("schemachange"));
+    dirTestWatcher.copyResourceToRoot(Paths.get("parquet", "empty"));
   }
 
   @Test // see DRILL-2408
```
arina-ielchiieva: Please remove `see DRILL-2408`

oleg-zinovev: My bad. Done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -122,6 +122,9 @@
   private PrimitiveTypeName logicalTypeForDecimals;
   private boolean usePrimitiveTypesForDecimals;
 
+  /** Whether no rows was written. */
```
arina-ielchiieva: Is used to ensure that empty Parquet file will be written if no rows were provided.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/impl/writer/TestParquetWriterEmptyFiles.java
```diff
@@ -18,72 +18,142 @@
...
+      .schemaBaseLine(expectedSchema)
+      .go();
+
+    // Make sure that only 1 parquet file was created
+    assertEquals(1, outputFile.list((dir, name) -> name.endsWith("parquet")).length);
   }
 
-  @Test // see DRILL-2408
+  @Test
+  public void testSimpleEmptyFileSchema() throws Exception {
```
arina-ielchiieva: This test is redundant since `testComplexEmptyFileSchema` both cases.

arina-ielchiieva: Also we need to add test where we select from non-empty Parquet file but filter condition eliminates all rows, similar as you have for JSON.

arina-ielchiieva:
@oleg-zinovev thanks for making the changes, a couple of minor comments are left...

oleg-zinovev:
@arina-ielchiieva , thanks for review. Done

arina-ielchiieva:
Ran all tests on the test cluster, all passed. LGTM, +1
@oleg-zinovev thanks for making the changes.

PR #1838, opened 2019-08-07
DRILL-7341: Vector reAlloc may fails after exchange

oleg-zinovev:
Fixes a relocate issues after direct  assignment of Vector data field.

### DIFF
--- a/exec/vector/src/main/codegen/templates/FixedValueVectors.java
+++ b/exec/vector/src/main/codegen/templates/FixedValueVectors.java
@@ -210,9 +210,16 @@ public final class ${minor.class}Vector extends BaseDataValueVector implements F
     // a zero-length buffer. Instead, just allocate a 256 byte
     // buffer if we start at 0.
 
-    final long newAllocationSize = allocationSizeInBytes == 0
+    long newAllocationSize = allocationSizeInBytes == 0
         ? 256
         : allocationSizeInBytes * 2L;
+
+    final int currentCapacity = data.capacity();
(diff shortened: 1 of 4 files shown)


exec/vector/src/main/codegen/templates/FixedValueVectors.java
```diff
@@ -210,10 +210,16 @@ public void reAlloc() {
...
     // buffer if we start at 0.
 
-    final long newAllocationSize = allocationSizeInBytes == 0
+    long newAllocationSize = allocationSizeInBytes == 0
         ? 256
         : allocationSizeInBytes * 2L;
 
+    // Some operations, such as Value Vector#exchange, can be change DrillBuf data field without corresponding allocation size changes.
+    // Check that the size of the allocation is sufficient to copy the old buffer.
+    while (newAllocationSize < data.capacity()) {
```
weijietong: please also change the following 232 line: `data.setZero(halfNewCapacity, halfNewCapacity)` logic which initializes the allocated fresh buffer content to zero.

oleg-zinovev: Done.

Could you clarify why DrillBug#setZero performed only for FixedValueVectors?

weijietong: The change is to keep backward compatible. The reason I guess is that most of the fixed VV is number type. Initializing to zero is to keep the java default value convention.

weijietong:
LGTM +1

PR #1839, opened 2019-08-09
DRILL-4517: Support reading empty Parquet files

arina-ielchiieva:
1. Modified flat and complex parquet readers to output schema only when requested number of records to read is 0. In this case readers are not initialized to improve performance.
2. Allowed reading requested number of rows instead of all rows in the row group (DRILL-6528).
3. Fixed issue with nulls number determination in the row group (fixed IsPredicate#isAllNulls method).
4. Allowed reading empty parquet files via adding empty / fake row group.
5. General refactoring and unit tests.

Jira - [DRILL-4517](https://issues.apache.org/jira/browse/DRILL-4517).


exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/metadata/FileMetadataCollector.java
```diff
@@ -0,0 +1,278 @@
...
+      long nullCount = totalNullCountMap.get(columnTypeMetadataKey) + totalNullCount;
+      totalNullCountMap.put(columnTypeMetadataKey, nullCount);
+    }
+    if (thisColumnIsInteresting) {
+      // Save the column schema info. We'll merge it into one list
+      Object minValue = null;
+      Object maxValue = null;
+      boolean statsAvailable = !stats.isEmpty();
+      if (statsAvailable) {
+        if (stats.hasNonNullValue()) {
```
vvysotskyi: Could you please combine these two if statements?

P.S. Since this code was moved from another place, it is optional to address this comment, but in either way, I've written it :)

arina-ielchiieva:
@vvysotskyi thanks for the code review, addressed code review comments.

PR #1843, opened 2019-08-15
DRILL-7350: Move RowSet related classes from test folder

vvysotskyi:
- Moved row-set related classes from test folder into main and changed package from `org.apache.drill.test.rowSet` to `org.apache.drill.exec.physical.rowSet` where other row-set related classes were placed.
- Updated package info for both packages.

For details please see [DRILL-7350](https://issues.apache.org/jira/browse/DRILL-7350).


exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetReader.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.drill.test.rowSet;
+package org.apache.drill.exec.physical.rowSet;
```
arina-ielchiieva: Please rename method: `void setPosn(int index);`

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractRowSet.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.drill.test.rowSet;
+package org.apache.drill.exec.physical.rowSet;
```
arina-ielchiieva: 1. Remove `protected SchemaChangeCallBack callBack = new SchemaChangeCallBack();`
2.Better error message: `public long size() {
    throw new UnsupportedOperationException("getSize");
  }`

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractSingleRowSet.java
```diff
@@ -15,15 +15,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.drill.test.rowSet;
+package org.apache.drill.exec.physical.rowSet;
```
arina-ielchiieva: Constructors can be protected.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/DirectRowSet.java
```diff
@@ -33,9 +32,10 @@
 import org.apache.drill.exec.record.VectorContainer;
 import org.apache.drill.exec.record.selection.SelectionVector2;
 import org.apache.drill.exec.vector.accessor.convert.ColumnConversionFactory;
-import org.apache.drill.test.rowSet.RowSet.ExtendableRowSet;
-import org.apache.drill.test.rowSet.RowSetWriterImpl.WriterIndexImpl;
+import org.apache.drill.exec.physical.rowSet.RowSet.ExtendableRowSet;
+import org.apache.drill.exec.physical.rowSet.RowSetWriterImpl.WriterIndexImpl;
 
+import java.util.HashSet;
```
arina-ielchiieva: `public RowSetWriter writer() {
    return writer(10);
  }`

Create a constant instead with Java-doc.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/DirectRowSet.java
```diff
@@ -132,7 +131,7 @@ public RowSetReader reader() {
 
   @Override
   public SingleRowSet toIndirect() {
-    return new IndirectRowSet(this, Sets.<Integer>newHashSet());
+    return new IndirectRowSet(this, new HashSet<>());
```
arina-ielchiieva: Should be immutable?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/IndirectRowSet.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.drill.test.rowSet;
+package org.apache.drill.exec.physical.rowSet;
```
arina-ielchiieva: Replace usage of `Sets.<Integer>newHashSet()`

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSet.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.drill.test.rowSet;
+package org.apache.drill.exec.physical.rowSet;
```
arina-ielchiieva: 1. Remove public: `public interface SingleRowSet extends RowSet {`
2. Please try to fix all java doc errors.

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/RowSetBuilder.java
```diff
@@ -20,12 +20,15 @@
...
+import org.apache.drill.exec.physical.rowSet.DirectRowSet;
+import org.apache.drill.exec.physical.rowSet.RowSet;
+import org.apache.drill.exec.physical.rowSet.RowSetWriter;
 import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.metadata.MetadataUtils;
 import org.apache.drill.exec.record.metadata.TupleMetadata;
 import org.apache.drill.exec.vector.accessor.convert.ColumnConversionFactory;
 import org.apache.drill.shaded.guava.com.google.common.collect.Sets;
-import org.apache.drill.test.rowSet.RowSet.SingleRowSet;
+import org.apache.drill.exec.physical.rowSet.RowSet.SingleRowSet;
```
arina-ielchiieva: 1. Capacity here can use constant we introduce for `10`.
2. Also can be considered to be moved.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractRowSet.java
vvysotskyi: Thanks, done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractSingleRowSet.java
vvysotskyi: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/DirectRowSet.java
vvysotskyi: Thanks, done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/DirectRowSet.java
vvysotskyi: Agree, replaced with `Collections.emptySet()`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/IndirectRowSet.java
vvysotskyi: Thanks, replaced with `Collections.emptySet()`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSet.java
vvysotskyi: Thanks, done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetPrinter.java
```diff
@@ -15,7 +15,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.drill.test.rowSet;
+package org.apache.drill.exec.physical.rowSet;
```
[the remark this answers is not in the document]
vvysotskyi: Thanks, replaced `PrintStream` usage with `StringBuilder`, renamed class to `RowSetStringBuilder` and renamed its methods.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetReader.java
vvysotskyi: Thanks, renamed to `setPosition()`.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetReader.java
```diff
@@ -46,8 +46,7 @@
    * The index of the underlying row which may be indexed by an
    * SV2 or SV4.
    *
-   * @return
+   * @return index of the underlying row
    */
-
   int offset();
 }
```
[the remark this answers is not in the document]
vvysotskyi: Done.

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/RowSetBuilder.java
vvysotskyi: Thanks, done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+ * Helper class to obtain string representation of RowSet.
+ */
+public class RowSetStringBuilder {
+  private RowSet rowSet;
+
+  public RowSetStringBuilder(RowSet rowSet) {
+    this.rowSet = rowSet;
+  }
+
+  public static void appendTupleSchema(StringBuilder stringBuilder, TupleMetadata schema) {
```
arina-ielchiieva: Why this is public and static?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  @Override
+  public String toString() {
+    StringBuilder result = new StringBuilder();
+    appendTo(result);
+    return result.toString();
+  }
+
+  public void appendTo(StringBuilder stringBuilder) {
```
arina-ielchiieva: Same here?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+ * limitations under the License.
+ */
+package org.apache.drill.exec.physical.rowSet;
+
+import org.apache.drill.exec.record.BatchSchema.SelectionVectorMode;
+import org.apache.drill.exec.record.metadata.ColumnMetadata;
+import org.apache.drill.exec.record.metadata.TupleMetadata;
+
+/**
+ * Helper class to obtain string representation of RowSet.
```
arina-ielchiieva: It would be nice to add an example of the output.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+public class RowSetStringBuilder {
+  private RowSet rowSet;
```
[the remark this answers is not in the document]
vvysotskyi: Thanks, done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
vvysotskyi: Made private and non-static.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
vvysotskyi: Thanks, done.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+        stringBuilder.append(", ");
+        stringBuilder.append(reader.offset());
+        stringBuilder.append(")");
+        break;
+      case TWO_BYTE:
+        stringBuilder.append(" (");
+        stringBuilder.append(reader.offset());
+        stringBuilder.append(")");
+        break;
+      default:
```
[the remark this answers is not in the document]
vvysotskyi: No, it is not required.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
vvysotskyi: Agree, added an example.

paul-rogers:
@vvysotskyi, thanks for making this change. It will allow certain other tasks to be much easier.

The first observation is that it would be best not to put the RowSet and ResultSetLoader stuff in the same package: they represent two independent systems and our package naming should make that clear.

Perhaps

1) Rename the existing physical.rowSet to physical.resultSet.
2) Move the RowSet classes to a new physical.rowSet.

The difference, so we're clear, is that RowSet works on a single batch, while the ResultSet works across a stream of batches. If operators, say, want to use the row set stuff to read batches, then we need a ResultSetReader that automagically handles things like schema changes across batches, etc.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractRowSet.java
```diff
@@ -15,23 +15,20 @@
...
 import org.apache.drill.exec.record.BatchSchema;
 import org.apache.drill.exec.record.VectorAccessible;
 import org.apache.drill.exec.record.VectorContainer;
 import org.apache.drill.exec.record.metadata.TupleMetadata;
-import org.apache.drill.exec.vector.SchemaChangeCallBack;
 
 /**
  * Basic implementation of a row set for both the single and multiple
- * (hyper) varieties, both the fixed and extendible varieties.
+ * (hyper) varieties, both the fixed and extendable varieties.
```
paul-rogers: extensible

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractRowSet.java
```diff
@@ -41,13 +38,19 @@ public AbstractRowSet(VectorContainer container, TupleMetadata schema) {
   }
 
   @Override
-  public VectorAccessible vectorAccessible() { return container(); }
+  public VectorAccessible vectorAccessible() {
```
paul-rogers: Why change formatting? Drill has long used single-line methods for conciseness. Is this an artifact of some code formatter? I didn't see discussion on the dev list about changing our coding standards...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractSingleRowSet.java
```diff
@@ -15,37 +15,27 @@
...
-    public RowSetReader buildReader(AbstractSingleRowSet rowSet, ReaderIndex rowIndex) {
-      TupleMetadata schema = rowSet.schema();
-      return new RowSetReaderImpl(schema, rowIndex,
-          buildContainerChildren(rowSet.container(),
-          new MetadataRetrieval(schema)));
-    }
-  }
-
-  public AbstractSingleRowSet(AbstractSingleRowSet rowSet) {
+  protected AbstractSingleRowSet(AbstractSingleRowSet rowSet) {
```
paul-rogers: Why remove nested classes? These are perfectly normal part of Java that allows small, single-purpose classes to be associated with the one and only class that uses them. Again, if we want to change the Drill style guidelines, let's have the discussion on dev list first.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractSingleRowSet.java
```diff
@@ -66,4 +57,14 @@ public long size() {
   protected RowSetReader buildReader(ReaderIndex rowIndex) {
     return new RowSetReaderBuilder().buildReader(this, rowIndex);
   }
+
+  public static class RowSetReaderBuilder extends BaseReaderBuilder {
```
paul-rogers: See note above.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/DirectRowSet.java
```diff
@@ -45,21 +45,10 @@
...
-      RowSetWriterImpl writer = new RowSetWriterImpl(rowSet, schema, index,
-          buildContainerChildren(rowSet.container(),
-          new MetadataRetrieval(schema)));
-      return writer;
-    }
-  }
+  /**
+   * Initial row count, used for preliminary memory allocation.
+   */
+  public static final int INITIAL_ROW_COUNT = 10;
```
paul-rogers: See note above about removal of nested class.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/DirectRowSet.java
```diff
@@ -141,5 +136,22 @@ public SingleRowSet toIndirect(Set<Integer> skipIndices) {
   }
 
   @Override
-  public SelectionVector2 getSv2() { return null; }
+  public SelectionVector2 getSv2() {
+    return null;
+  }
+
+  public static class RowSetWriterBuilder extends BaseWriterBuilder {
```
paul-rogers: See note above about cluttering the project with many small classes used in only one location.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
```diff
@@ -0,0 +1,112 @@
...
+ * Helper class to obtain string representation of RowSet.
+ * Example of the output:
+ * <pre>
+ *   #: year, month, day
+ *   0: 2017, 12, 17
+ *   1: 2017, 12, 18
+ *   2: 2017, 12, 19
+ * </pre>
+ */
+public class RowSetStringBuilder {
```
paul-rogers: The problem with this approach is that row sets can be large: creating a string to print causes memory pressure.

Consider creating a row set builder that takes a stream (or Writer). The Writer can be a String writer. That way, when used with a large data set, the writer can be to stdout, or whatever.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetWriterImpl.java
```diff
@@ -129,7 +129,7 @@ public RowSetWriter addRow(Object...values) {
 
   @Override
   public RowSetWriter addSingleCol(Object value) {
-    return addRow(new Object[] {value});
+    return addRow(value);
```
paul-rogers: Won't work. Must be an array. The idea is that addRow() takes an array of columns. But, if the row has a single column, and that one column is repeated, the Java code is ambiguous. The original (correct) implementation handles this case.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/package-info.java
```diff
@@ -32,6 +32,22 @@
...
+ * <dd>A read-only, single row set with an SV2. Note that the SV2 itself is
+ * writable (such as for sorting.)</dd>
+ * <dt>ExtendableRowSet</dt>
+ * <dd>A write-only, single row set used to create a new row set. Because of
+ * the way Drill sets row counts, an extendable row set cannot be read; instead
+ * at the completion of the write the extendable row set becomes a direct or
+ * indirect row set.</dd>
+ * <dt>HyperRowSet</dt>
+ * <dd>A read-only row set made up of a collection of record batches, indexed via an
+ * SV4. As with the SV2, the SV4 itself is writable.</dt>
```
paul-rogers: The SV is not writable within a RowSet. Changing the SV will mess up indexing.

The way to use that is to have a DirectRowSet on the original batch. Create the SV. Then, use the DirectRowSet and the SV to create an indirect row set.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/package-info.java
```diff
@@ -32,6 +32,22 @@
  * batch" or "record batch." (But, in Drill, the term "record batch" also
  * usually means an operator on that set of records. Here, a row set is
  * just the rows &nash; separate from operations on that data.</dd>
+ * <dt>SingleRowSet (abstract)</dt>
```
paul-rogers: See general comment. Please do not combine the RowSet and ResultSetLoader classes into a single package: they present very different models used for different purposes.

Yes, it is confusing because there is no ResultSetReader at present, but there should be to handle schema changes, etc. across batches. RowSet handles only a single row set such as in tests. It was deliberately designed to NOT be ResultSetReader solution.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/package-info.java
```diff
@@ -117,12 +134,14 @@
...
- * Each batch is delimited by a call to {@link #startBatch()} and a call to
- * {@link #harvestWithLookAhead()} to obtain the completed batch. Note that readers do not
+ * Each batch is delimited by a call to {@link org.apache.drill.exec.physical.rowSet.ResultSetLoader#startBatch()}
+ * and a call to {@link org.apache.drill.exec.physical.rowSet.impl.VectorState#harvestWithLookAhead()}
+ * to obtain the completed batch. Note that readers do not
  * call these methods; the scan operator does this work.
  * <p>
- * Each row is delimited by a call to {@link #startValue()} and a call to
- * {@link #saveRow()}. <tt>startRow()</tt> performs initialization necessary
+ * Each row is delimited by a call to {@link org.apache.drill.exec.vector.accessor.writer.WriterEvents#startRow()}
```
paul-rogers: WriterEvents is meant to be entirely internal; it should not be part of the API. WriterEvents is the internal implementation of the public API described here.

exec/java-exec/src/test/java/org/apache/drill/test/QueryRowSetIterator.java
```diff
@@ -99,13 +99,6 @@ public DirectRowSet next() {
     }
   }
 
-  public void printAll() {
```
paul-rogers: Please either leave this, or move it to a utility class. This turns out to be a very handy way to validate or debug a unit test: add a call while working on the test to visualize the results, remove the call when done.

paul-rogers:
Two general comments. First, not sure why we needed to convert compact single-line methods (a long-time Drill standard) into four-line methods. Perhaps this is the result of an overly-helpful code formatter?

Second, looks like a bunch of nested helper classes were moved elsewhere. Not sure this is helpful as it clutters the code with small files used in exactly one place. The nested classes (a perfectly normal Java practice) makes clear that these are, in fact, helpers.

vvysotskyi:
@paul-rogers, I have reverted debatable formatting changes. Let's postpone this discussion until we decide to add some additional checkstyle rules.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractSingleRowSet.java
arina-ielchiieva: Agree on moving nested classes to the bottom since it makes code mode readable, first you see all code that belongs to the class and then nested classes.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractRowSet.java
arina-ielchiieva: I think we should enforce such checkstyle. Taking into account that most of IDE can compress such methods automatically plus many other projects follow the same pattern. Personally, I was following such code style as well. @vvysotskyi could you please create Jira?

exec/java-exec/src/test/java/org/apache/drill/test/QueryRowSetIterator.java
```diff
@@ -99,6 +101,13 @@ public DirectRowSet next() {
     }
   }
 
+  public void printAll() {
```
arina-ielchiieva: No sure we should allow printing into stout in the code. When code was for tests, it was ok but since we moved it, I think we should consider passing the stream instead.

arina-ielchiieva:
Overall, LGTM. Let's wait for @paul-rogers approval.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/AbstractRowSet.java
vvysotskyi: I have created https://issues.apache.org/jira/browse/DRILL-7352 for this. I'll populate it with more rules soon.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetWriterImpl.java
paul-rogers: While this would seem to be true, it won't work for the case of a map:

addRow(new Object[]{"foo", 10})

Is the above two column values, or a single column that happens to be an object?

There is actually no need to debate this, however. Without this feature, some unit tests will fail. (I know that because I added this method directly in response to those test failures...)

Still, if you can get those tests to pass, even with the map value ambiguity, then my concern will be addressed.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetStringBuilder.java
paul-rogers: The class name is still "RowSetStringBuilder" Better to rename it to, say, "RowSetFormatter". The constructor should take both a row set and a writer. We can then provide static methods for the two most common cases: print to stdout (which I use all the time when creating tests) and to string (which you seem to prefer):

```
public static void print(RowSet rowSet) {
  new RowSetFormatter(rowSet, System.out).write();
}

public static String toString(RowSet rowSet) {
  StringWriter out = new StringWriter();
  new RowSetFormatter(rowSet, out).write();
  return out.toString();
}
```

The constructor allows other options, such as writing to a file. Sadly, it won't work to write to a log; we'd have to materialize to a string...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/package-info.java
paul-rogers: Thanks!

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
```diff
@@ -781,9 +781,19 @@ public void testPluginSerialization() throws IOException {
 
   @Test
   public void testFirewallSchema() throws RpcException {
-    String sql = "SELECT * FROM cp.`regex/firewall.ssdlog`";
+    String sql = "SELECT * FROM cp.`regex/firewall.ssdlog` limit 0";
     RowSet result = client.queryBuilder().sql(sql).rowSet();
-    result.print();
+    TupleMetadata expectedSchema = new SchemaBuilder()
+        .add("eventDate", TypeProtos.MinorType.TIMESTAMP, TypeProtos.DataMode.OPTIONAL)
```
paul-rogers: Simpler: `addNullable("eventDate", MinorType.TIMESTAMP)`

Import `MinorType`, use the `addNullable` method rather than the more verbose form of using the `DataMode`. That is, express intent, not implementation.

exec/java-exec/src/test/java/org/apache/drill/test/QueryRowSetIterator.java
```diff
@@ -101,7 +103,7 @@ public DirectRowSet next() {
 
   public void printAll() {
     for (DirectRowSet rowSet : this) {
-      rowSet.print();
+      new RowSetStringBuilder(rowSet).writeTo(new OutputStreamWriter(System.out));
```
paul-rogers: This is *not* a simplification; it exposes too much detail. Better:

```
RowSetFormatter.print(rowSet);
```

Of course, if we go this route, we might as well leave the `print()` method in `RowSet` and have it call the above code.

Just to be clear, the purpose of `print()` is to help us developers visualize what's going on in Drill; it is not for production use. This is important because, if you can't visualize a batch, you'll make many mistakes because you can't tell what's what.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
```diff
@@ -781,9 +781,19 @@ public void testPluginSerialization() throws IOException {
 
   @Test
   public void testFirewallSchema() throws RpcException {
-    String sql = "SELECT * FROM cp.`regex/firewall.ssdlog`";
+    String sql = "SELECT * FROM cp.`regex/firewall.ssdlog` limit 0";
     RowSet result = client.queryBuilder().sql(sql).rowSet();
-    result.print();
+    TupleMetadata expectedSchema = new SchemaBuilder()
```
paul-rogers: Another change you might consider is renaming either this, or the other `SchemaBuilder`. Since this is the one we want to use moving forward, maybe rename the other to `BatchSchemaBuilder` so that future folks don't have to figure out the correct import when the reference the class.

exec/java-exec/src/test/java/org/apache/drill/test/QueryRowSetIterator.java
```diff
@@ -17,24 +17,26 @@
...
 import org.apache.drill.test.BufferingQueryEventListener.QueryEvent;
-import org.apache.drill.test.rowSet.DirectRowSet;
+import org.apache.drill.exec.physical.rowSet.DirectRowSet;
 
 public class QueryRowSetIterator implements Iterator<DirectRowSet>, Iterable<DirectRowSet> {
   private final BufferingQueryEventListener listener;
   private int recordCount = 0;
   private int batchCount = 0;
-  QueryId queryId = null;
+  private QueryId queryId = null;
```
paul-rogers: As long as you are touching this code, please remove the do-nothing initializers:

```
private int recordCount;
private int batchCount;
private QueryId queryId;
```

arina-ielchiieva:
@paul-rogers 

> I also recall that Arina once suggested removing an anonymous inner class in favor of a declared class. 

I doubt this is true :) Maybe thrown out of the context though...

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetWriterImpl.java
vvysotskyi: Tests passed with this change. I'm afraid that we are talking about different cases.
`addRow(new Object[]{"foo", 10})` will work as you expected - it will handle incoming array as two column values, but the code in the method which was changed is different a little bit, it is similar to this call:
`addRow((Object) new Object[]{"foo", 10})` - in this case, it will be handled as a single column with array value.

paul-rogers:
@arina-ielchiieva, could be my misunderstanding, though I do recall making the change... Point is, both forms are perfectly valid Java; it is just a preference. To avoid misunderstandings we might want to add to our formatting rules.

Again, just to be clear, I have no issue with whatever rules we decide on; but let's decide on a common set to avoid ping-pong format changes as each of us touches the code.

As it turns out, Charles started a discussion on the dev list. Perhaps you can offer there the format rules you've found to work, such as `int[] foo;` instead of `int foo[];`, classes-at-bottom, no single-line method declarations, preferred import sort order and so on. Might also include preferences for Java over Guava methods for common operations. That way, we can all follow the best practices you've discovered.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetWriterImpl.java
paul-rogers: Ah! But I don't want to do the cast; that is the purpose of this method. I want:

```
  ...addRow(mapValue(10, "foo"), 20); // convenience
  ...addRow(new Object[] {10, "foo"}, 20); // low-level
...
  ...addSingleCol(mapValue(10, "foo"); // convenience
 ... addSingleCol(new Object[] {10, "foo"}); // low-level
```

Unless we wrap that single column in another array, the second form is the same as:

```
 ... addRow(new Object[] {10, "foo"});
```

Which will be interpreted as a set of two columns, not as a single column with a map-value.

I could be wrong, and I'd have to test actual code to be sure. In fact, I'll do that test this weekend.

vvysotskyi: No need to cast, it was an example. In the method where the value is used reference type is already `Object`, so we don't need to do a cast.

> Unless we wrap that single column in another array, the second form is the same as:
> ```
>  ... addRow(new Object[] {10, "foo"});
> ```

Actually not, it will be the same as
```
addRow((Object) new Object[] {10, "foo"});
```
because method argument in `addSingleCol` has `Object` type.

vvysotskyi:
The latest change caused tests failures. It is appeared because with new BatchSchemaBuilder TupleMetadata was used more frequently and some code does not set zero scale and precision, but PrimitiveColumnMetadata does. I'll let you know when this issue is fixed.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
vvysotskyi: I have reverted this change since it caused a lot of tests failures. I'll create a Jira to replace usage of another `SchemaBuilder` with `BatchSchemaBuilder`.

vvysotskyi:
I have reverted breaking change, so for now, all tests passed.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
paul-rogers: Thanks for trying to fix this. Doing the change as another PR sounds good.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetWriterImpl.java
paul-rogers: Thanks for the explanation, makes sense.

exec/java-exec/src/test/java/org/apache/drill/exec/store/log/TestLogReader.java
vvysotskyi: I have created https://issues.apache.org/jira/browse/DRILL-7354 to address this change.

PR #1844, opened 2019-08-20
DRILL-7326: Support repeated lists for CTAS parquet format

ihuzenko:
PR for [DRILL-7326](https://issues.apache.org/jira/browse/DRILL-7326).


exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
```diff
@@ -118,6 +118,18 @@ public void endField() throws IOException {
     }
 
     public abstract void writeField() throws IOException;
+
+    /**
+     * Used by repeated converters for writing parquet logical lists. For schema details see
```
arina-ielchiieva:
```suggestion
     * Used by repeated converters for writing Parquet logical lists. For schema details see
```

exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
```diff
@@ -118,6 +118,18 @@ public void endField() throws IOException {
...
+     * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists.
+     *
+     * @throws IOException
+     */
+    public void writeListField() throws IOException {
+      throw new UnsupportedOperationException(String.format(
+          "Converter '%s' doesn't support writing of list fields.",
```
arina-ielchiieva:
```suggestion
          "Converter '%s' doesn't support writing list fields.",
```

exec/java-exec/src/main/codegen/templates/ParquetOutputRecordWriter.java
```diff
@@ -268,10 +268,112 @@ public void writeField() throws IOException {
     consumer.endField(fieldName, fieldId);
   </#if>
     }
+
+    <#if mode.prefix == "Repeated">
+    /**
+     * Writing list field according to logical type rules
```
arina-ielchiieva:
```suggestion
     * Writelist field according to logical type rules
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -302,6 +312,32 @@ private Type getType(MaterializedField field) {
...
 
+  private void addElementType(ListBuilder<GroupType> listBuilder, MaterializedField elementField) {
+    if (elementField.getDataMode() == DataMode.REPEATED) {
+      MaterializedField child2 = elementField.getChildren().iterator().next();
+      ListBuilder<GroupType> inner = org.apache.parquet.schema.Types.requiredList();
+      addElementType(inner, child2);
+      listBuilder.setElementType(inner.named("element"));
+    } else {
+      Type element = getType(elementField);
+      // rename element type before adding to listBuilder
```
arina-ielchiieva: Why? Add explanation to the comment.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -302,6 +312,32 @@ private Type getType(MaterializedField field) {
     }
   }
 
+  private void addElementType(ListBuilder<GroupType> listBuilder, MaterializedField elementField) {
```
arina-ielchiieva: Please add java doc.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -423,6 +459,34 @@ public void writeField() throws IOException {
...
+
+    RepeatedListParquetConverter(int fieldId, String fieldName, FieldReader reader) {
+      super(fieldId, fieldName, reader);
+      converter = EventBasedRecordWriter.getConverter(ParquetRecordWriter.this, 0, "", reader.reader());
+    }
+
+    @Override
+    public void writeField() throws IOException {
+      consumer.startField(fieldName, fieldId);
+      consumer.startField("list", 0);
```
arina-ielchiieva: Should we use constant?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -423,6 +459,34 @@ public void writeField() throws IOException {
...
+      while (reader.next()) {
+        consumer.startGroup();
+        consumer.startField("element", 0);
```
arina-ielchiieva: Same here...

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -357,6 +358,39 @@ public void testTableIsFoundWithinWorkspaceWhenNameStartsWithSlash() throws Exce
     }
   }
 
+  @Test
```
arina-ielchiieva: These tests test Parquet Writer not something specific to CTAS, so I believe they should be moved to one of the classes in the `org.apache.drill.exec.physical.impl.writer` test package.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -357,6 +358,39 @@ public void testTableIsFoundWithinWorkspaceWhenNameStartsWithSlash() throws Exce
...
+  public void testCtasForList() throws Exception {
+    String tableName = "testCtasForList";
+    try {
+      test("CREATE TABLE `%s`.`%s` AS SELECT l FROM cp.`jsoninput/input2.json`", DFS_TMP_SCHEMA, tableName);
+      testBuilder()
+          .sqlQuery("SELECT * FROM `%s`.`%s` LIMIT 1", DFS_TMP_SCHEMA, "/" + tableName)
```
arina-ielchiieva: Since we already implicitly using string builder, having concatenation for the second parameter looks weird.

exec/java-exec/src/test/java/org/apache/drill/exec/sql/TestCTAS.java
```diff
@@ -357,6 +358,39 @@ public void testTableIsFoundWithinWorkspaceWhenNameStartsWithSlash() throws Exce
...
+    }
+  }
+
+  @Test
+  public void testCtasForRepeatedList() throws Exception {
+    String tableName = "testCtasForRepeatedList";
+    try {
+      test("CREATE TABLE `%s`.`%s` AS SELECT * FROM cp.`jsoninput/repeated_list_bug.json`", DFS_TMP_SCHEMA, tableName);
+      testBuilder()
+          .sqlQuery("SELECT rl FROM `%s`.`%s`", DFS_TMP_SCHEMA, "/" + tableName)
```
arina-ielchiieva: Same here.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -292,7 +294,15 @@ private Type getType(MaterializedField field) {
         }
         return new GroupType(dataMode == DataMode.REPEATED ? Repetition.REPEATED : Repetition.OPTIONAL, field.getName(), types);
       case LIST:
-        throw new UnsupportedOperationException("Unsupported type " + minorType);
+        Preconditions.checkArgument(!field.getChildren().isEmpty(),
```
arina-ielchiieva: Is there a case when this check is needed? If occurrence of such situation is unlikely, it's better to remove if to avoid unnecessary checks during write.

exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
```diff
@@ -118,6 +118,18 @@ public void endField() throws IOException {
     }
 
     public abstract void writeField() throws IOException;
+
+    /**
+     * Used by repeated converters for writing parquet logical lists. For schema details see
+     * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists.
+     *
+     * @throws IOException
```
arina-ielchiieva: add description to avoid warning in the IDE.

exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
```diff
@@ -118,6 +118,18 @@ public void endField() throws IOException {
     }
 
     public abstract void writeField() throws IOException;
+
+    /**
+     * Used by repeated converters for writing parquet logical lists. For schema details see
+     * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists.
```
arina-ielchiieva: Use proper html link for java doc.

exec/java-exec/src/main/codegen/templates/ParquetOutputRecordWriter.java
ihuzenko: I'll drop the javadoc, intent is explained in parent one.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
ihuzenko: I'm also not sure whether constant is necessary here:))) It's just passing index of field within group of one element. I'm afraid having improperly named constant may create only confusion here.

arina-ielchiieva: I don't have strong opinion here, you may leave as it, the main reason why I have written this comment is that I saw the same string repeating several times.

ihuzenko: ok, I'll add few constants to ```FieldConverter```.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -292,7 +293,12 @@ private Type getType(MaterializedField field) {
         }
         return new GroupType(dataMode == DataMode.REPEATED ? Repetition.REPEATED : Repetition.OPTIONAL, field.getName(), types);
       case LIST:
-        throw new UnsupportedOperationException("Unsupported type " + minorType);
+        MaterializedField elementField = field.getChildren().iterator().next();
```
vvysotskyi: I assume you are expecting here `$data$` field, but in some cases, list filed may also have `$offsets$` child. Should we obtain a child with the required name here (`BaseRepeatedValueVector.DATA_VECTOR_NAME`)?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -302,6 +308,40 @@ private Type getType(MaterializedField field) {
...
+    if (elementField.getDataMode() == DataMode.REPEATED) {
+      MaterializedField child2 = elementField.getChildren().iterator().next();
+      ListBuilder<GroupType> inner = org.apache.parquet.schema.Types.requiredList();
+      addElementType(inner, child2);
+      listBuilder.setElementType(inner.named("element"));
+    } else {
+      Type element = getType(elementField);
+      // element may have internal name '$data$',
+      // rename it to 'element' according to Parquet list schema
+      if (element.isPrimitive()) {
```
vvysotskyi: Is it possible when the `element` is not primitive? Or if we currently don't assume such cases, should we throw an exception for non-primitive elements?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
ihuzenko: done

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -302,6 +308,40 @@ private Type getType(MaterializedField field) {
...
+        PrimitiveType primitiveElement = element.asPrimitiveType();
+        element = new PrimitiveType(
+            primitiveElement.getRepetition(),
+            primitiveElement.getPrimitiveTypeName(),
+            "element",
```
vvysotskyi: And this literal may be also replaced.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -302,6 +308,40 @@ private Type getType(MaterializedField field) {
...
+            primitiveElement.getOriginalType()
+        );
+      } else {
+        GroupType groupElement = element.asGroupType();
+        element = new GroupType(groupElement.getRepetition(),
+            "element", groupElement.getFields());
```
vvysotskyi: And this one.

ihuzenko:
@arina-ielchiieva , I've updated the PR. Please take a look again.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
ihuzenko: yes, good idea, thanks:)

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
ihuzenko: yes, it's possible. Normally,  it's necessary to add another tests to check list of struct and repeated list of struct elements.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
```diff
@@ -287,40 +290,70 @@ private Type getType(MaterializedField field) {
...
+  /**
+   * Finds data child field of list or repeated type.
+   *
+   * @param field parent repeated field
+   * @return child data field
+   */
+  private MaterializedField getDataField(MaterializedField field) {
+    return field.getChildren().stream()
+        .filter(child -> BaseRepeatedValueVector.DATA_VECTOR_NAME.equals(child.getName()))
+        .findFirst()
```
arina-ielchiieva: Do we need exactly first or any? If any, better to use `.findAny()`. Why it does not matter if there are several? Please add better java-doc to answer these questions?

exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
```diff
@@ -96,7 +96,7 @@ private void initFieldWriters() throws IOException {
 
   public static abstract class FieldConverter {
     protected static final String LIST = "list";
-    protected static final String ELEMENT = "element";
+    public static final String ELEMENT = "element";
```
arina-ielchiieva: Please other two constants public for consistency.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
ihuzenko: What's wrong with ```findFirst``` if this is simply sequential stream with one data and possibly one offset field ?

exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
ihuzenko: why ? They are not used outside of converter.

arina-ielchiieva: Just for consistency, since `ELEMENT` is already used outside, others might be also used with time. You can leave as is, it was just a suggestion. Anyway, if you leave as is, please order constants according to access modifies: having protected / public / protected is incorrect order.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetRecordWriter.java
arina-ielchiieva: For parallel streams, its a big difference, for sequential stream it does not matter. I understand that you are using sequential stream but using `findAny()` will clearly indicate that order for you does not matter.

ihuzenko: ok I'll add the change, but still I don't understand such argumentation for code analogous to linear search. It's like explaining why linear search returns immediately on match.

exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
ihuzenko: I agree with ordering remark. Regarding visibility I'd like to follow YAGNI principle here.

arina-ielchiieva:

ihuzenko:
@vvysotskyi , @arina-ielchiieva thanks for review :)

exec/java-exec/src/main/codegen/templates/EventBasedRecordWriter.java
KazydubB: I think it is better to move the definition of `ELEMENT` and `LIST` to `ParquetRecordWriter` because `FieldConverter` is used for other (csv, json etc.) formats as well and the group names are needed for Parquet only and it's up to writer to "decide" how to write a file. What do you think?

ihuzenko: ok.

ihuzenko: Only thing here that the constants should be part of parent ```ParquetOutputRecordWriter```.

ihuzenko: @KazydubB , done.

PR #1845, opened 2019-08-20
DRILL-7353: Wrong driver class is written to the java.sql.Driver

agozhiy:
https://issues.apache.org/jira/browse/DRILL-7353

### DIFF
--- a/exec/jdbc-all/pom.xml
+++ b/exec/jdbc-all/pom.xml
@@ -496,6 +496,14 @@
                <exclude>webapps/**</exclude>
              </excludes>
            </filter>
+           <!-- This file is used to automatically load given jdbc driver without calling Class.forName().
+                Excluding the Avatica service file which is conflicting with the Drill one. -->
+           <filter>
+             <artifact>org.apache.calcite.avatica:*</artifact>
+             <excludes>
+               <exclude>META-INF/services/java.sql.Driver</exclude>
+             </excludes>
+           </filter>
          </filters>
         </configuration>
       </plugin>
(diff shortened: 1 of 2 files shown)


arina-ielchiieva:
@agozhiy could you provide more context to this issue. Why this approach was chosen?
Why calcite `java.sql.Driver` is chose over Drill's one? What if other `java.sql.Driver` will be present in other lib? Which one will be chosen? From the first glance, current approach looks brittle, taking into account that some other file might be chosen over Drill's one. Is there a way to merge all `java.sql.Driver` into one or excluding all unnecessary? Since there are no unit tests to ensure the proper `java.sql.Driver` is chosen, we need to make sure such issue won't pop again.

agozhiy:
@arina-ielchiieva, There are two artifacts that provide java.sql.Driver to the jar. Apparently the wrong one overwrites the right. So I excluded the file from Avatica to resolve the conflict.
Initially I tried to combine the files into one so both Avatica and Drill jdbc drivers were contained in java.sql.Driver. But as Avatica driver itself is not present in the final jar, error occurred during initialization and so no drivers were registered automatically at all. It can be fixed if swap driver names in java.sql.Driver, but this is unacceptable.
The best solution would be to include the file from a specific artifact and exclude all the others but I couldn't find a way to do so. But I agree that the current approach is not safe. I'll need to look into it a bit more.

agozhiy:
From the [shade plugin documentation:](http://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html)
> From a logical perspective, includes are processed before excludes, thus it's possible to use an include to collect a set of files from the archive then use excludes to further reduce the set.

So we cannot just exclude everything and then include a specific resource. Instead, we should exclude all interfering ones, that is the approach used in the PR.

exec/jdbc-all/pom.xml
```diff
@@ -496,6 +496,12 @@
                <exclude>webapps/**</exclude>
              </excludes>
            </filter>
+           <filter>
```
arina-ielchiieva: Please add comment explaining why we need to exclude this file.

exec/jdbc-all/pom.xml
```diff
@@ -799,6 +805,12 @@
                       <exclude>webapps/**</exclude>
                     </excludes>
                   </filter>
+                  <filter>
```
arina-ielchiieva: Same here.

arina-ielchiieva:

PR #1848, opened 2019-08-23
DRILL-7360: Refactor WatchService in Drillbit class and fix concurrency issues

arina-ielchiieva:
See [DRILL-7360](https://issues.apache.org/jira/browse/DRILL-7360) for details.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
@@ -52,10 +52,13 @@ import org.apache.drill.exec.work.WorkManager;
 import org.apache.drill.shaded.guava.com.google.common.annotations.VisibleForTesting;
 import org.apache.drill.shaded.guava.com.google.common.base.Stopwatch;
 import org.apache.zookeeper.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.slf4j.bridge.SLF4JBridgeHandler;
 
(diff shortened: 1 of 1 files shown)


arina-ielchiieva:
@vvysotskyi please review.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
```diff
@@ -371,65 +376,91 @@ private void javaPropertiesToSystemOptions() {
...
             }
           }
-        }
-      } finally {
-        if (wk != null) {
-          wk.cancel();
+          if (!watchKey.reset()) {
+            drillbit.interruptPollShutdown = false;
+            return;
+          }
```
sohami: Thanks for adding the call to reset. But I think there is still some more changes needed here. If `watchkey.reset()` returns false, in our case that means we have not seen any event related to graceful file and still observing some failure while resetting the watchKey. In this case we should try to register again because if we returned from here that means GracefulShutdown thread will exit with Drillbit still up and running. Any future attempt for graceful shutdown will not work.

arina-ielchiieva:
@sohami addressed code review comment, please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/server/Drillbit.java
arina-ielchiieva: Done.

arina-ielchiieva:
@sohami thanks for the review. Merging...

PR #1849, opened 2019-08-29
DRILL-7362: COUNT(*) on JSON with outer list results in JsonParse error

oleg-zinovev:
Fix for CountingJsonReader to support json file with list in content.

P.S. At the moment, it contains a lot of C&P from JsonReader class. I can provide another solution that adds a counting mode to the JsonReader itself.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/store/json/TestJsonRecordReader.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/store/json/TestJsonRecordReader.java
@@ -261,4 +261,17 @@ public class TestJsonRecordReader extends BaseTestQuery {
     }
     throw new Exception("testNotCountingQueryNotSkippingInvalidJSONRecords");
   }
+
+  @Test
+  @Category(UnlikelyTest.class)
+  // See DRILL-7362
+  /* Test for CountingJSONReader */
+  public void testContainingArrayCount() throws Exception {
+    testBuilder()
+      .sqlQuery("select count(*) as cnt from cp.`store/json/listdoc.json`")
+      .unOrdered()
+      .baselineColumns("cnt")
+      .baselineValues(2L)
+      .go();
+  }
 }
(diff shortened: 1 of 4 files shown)


vvysotskyi:
@oleg-zinovev, thanks for fixing this issue! 
Could you please reuse code from `JsonReader` class - for example, move required code to the parent class and extend classes which share the code. Also, you have added some additional parameters to the constructor, which may be omitted since they always have the same value.
Please update the test case to check the returned result.

One more thing - please assign Jira to yourself and mark it as "In progress" when you are starting working on it, and change its state to "Reviewable" when PR is opened (Thanks to @arina-ielchiieva  for doing these things for you).

For more details, connected with the development process, please refer to https://drill.apache.org/docs/apache-drill-contribution-guidelines/.

oleg-zinovev:
@vvysotskyi 
Can you review, please?  I tried to fix your comments.

Also, thank you for the information on task management in jira - it looks like the last time I didn’t read the guidelines to the end (Also thanks to the @arina-ielchiieva for fixing jira task for me)

exec/java-exec/src/test/java/org/apache/drill/exec/store/json/TestJsonRecordReader.java
```diff
@@ -261,4 +261,18 @@ public void testNotCountingQueryNotSkippingInvalidJSONRecords()
...
+  @Category(UnlikelyTest.class)
+  // See DRILL-7362
+  /* Test for CountingJSONReader */
+  public void testContainingArrayCount() throws Exception {
+    testBuilder()
+      .sqlQuery("select count(*) as cnt from cp.`store/json/listdoc.json`")
+      .unOrdered()
+      .baselineColumns("cnt")
+      .baselineValues(2L)
+    .build()
```
vvysotskyi: Please fix formatting and use `.go()` instead of `.build().run();`.

PR #1850, opened 2019-09-03
DRILL-7168: Implement ALTER SCHEMA ADD / REMOVE commands

arina-ielchiieva:
Jira - [DRILL-7168](https://issues.apache.org/jira/browse/DRILL-7168).

### DIFF
--- a/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
+++ b/exec/java-exec/src/main/codegen/includes/parserImpls.ftl
@@ -314,12 +314,12 @@ SqlNode SqlCreateSchema(SqlParserPos pos, String createType) :
             token_source.SwitchTo(SCH);
     }
     (
-        <LOAD>
+        <SCH_LOAD>
         {
             load = StringLiteral();
         }
     |
-        <PAREN_STRING>
+        <SCH_PAREN_STRING>
         {
             schema = SqlLiteral.createCharString(token.image, getPos());
         }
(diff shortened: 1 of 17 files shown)


exec/java-exec/src/main/codegen/includes/parserImpls.ftl
```diff
@@ -534,6 +534,110 @@ SqlNode SqlDescribeSchema() :
...
+   ]
+   {
+        if (schema == null && properties == null) {
+             throw new ParseException("ALTER SCHEMA ADD command must have at least <COLUMNS> or <PROPERTIES> keyword indicated.");
+        }
+        return new SqlSchema.Add(pos, table, path, SqlLiteral.createBoolean(replace, getPos()), schema, properties);
+   }
+}
+
+SqlCharStringLiteral ParseSchema() :
```
vvysotskyi: Is it possible to reuse this code in `SqlCreateSchema`?

exec/java-exec/src/main/java/org/apache/drill/exec/record/metadata/schema/StorageProperties.java
```diff
@@ -0,0 +1,71 @@
...
+package org.apache.drill.exec.record.metadata.schema;
+
+import org.apache.drill.exec.store.StorageStrategy;
+
+/**
+ * Holds storage properties used when writing schema container.
+ */
+public class StorageProperties {
+
+  private StorageStrategy storageStrategy;
```
vvysotskyi: final

arina-ielchiieva: Done. Thanks for noticing.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
arina-ielchiieva: No, all because of context switch, for create schema we use load and paren_string where each returns some specific result and one of push returns to default context. For alter schema add we use only paren_string and only this one can return to default context. The reason it is factored out into separate method is to prevent lookahead. Since first we need context switch and only then token search. With this method inline lookahead was happening before context switch and token from different context was never found.

arina-ielchiieva:
@vvysotskyi thanks for the review, addressed code review comments.

exec/java-exec/src/main/codegen/includes/parserImpls.ftl
vvysotskyi: Ok, thanks a lot for a detailed explanation.

PR #1852, opened 2019-09-06
DRILL-7369: Schema for MaprDB tables is not used for the case when several fields are queried

vvysotskyi:
For problem description please refer to [DRILL-7369](https://issues.apache.org/jira/browse/DRILL-7369).


arina-ielchiieva:
+1

PR #1853, opened 2019-09-10
DRILL-7368: Fix Iceberg Metastore failure when filter column contains nulls

arina-ielchiieva:
Jira [DRILL-7368](https://issues.apache.org/jira/browse/DRILL-7368).


arina-ielchiieva:
@vvysotskyi please review.

PR #1854, opened 2019-09-13
DRILL-7373: Fix problems involving reading from DICT type

KazydubB:
- Fixed `FieldIdUtil` to resolve reading from `DICT` for some complex cases;
- optimized reading from `DICT` given a key by passing an appropriate `Object` type to `DictReader#find(...)` and `DictReader#read(...)` methods when schema is known (e.g. when reading from Hive tables) instead of generating it on fly based on int or String path and key type;
- fixed error when accessing value by not existing key value in Avro table.

Jira - [DRILL-7373](https://issues.apache.org/jira/browse/DRILL-7373).

### DIFF
--- a/logical/src/main/java/org/apache/drill/common/expression/PathSegment.java
+++ b/logical/src/main/java/org/apache/drill/common/expression/PathSegment.java
@@ -17,13 +17,29 @@
  */
 package org.apache.drill.common.expression;
 
+import org.apache.drill.common.types.TypeProtos;
+
 public abstract class PathSegment {
+
+  /**
+   * Holds original value associated with the path segment.
+   * Used when reading data from DICT.
+   */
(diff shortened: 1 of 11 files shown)


logical/src/main/java/org/apache/drill/common/expression/PathSegment.java
```diff
@@ -17,13 +17,29 @@
...
+   * Holds original value associated with the path segment.
+   * Used when reading data from DICT.
+   */
+  protected final Object originalValue;
+
+  /**
+   * Indicates the type of original value.
+   * @see #originalValue
+   */
+  protected final TypeProtos.MajorType valueType;
```
ihuzenko:
```suggestion
  protected final TypeProtos.MajorType originalValueType;
```

PR #1859, opened 2019-09-24
DRILL-7170: Ignore uninitialized vector containers for OOM error messages

Ben-Zvi:
A Hash-Table OOM exception may happen while a vector container was allocated but not yet initialized (with _setRecordCount()_ ). This exception is caught and a detailed error message is generated, which involves calling _getRecordCount()_ on existing vector containers, hence could trigger an **IllegalStateException** on such uninitialized container.

   The fix: Ignore such uninitialized vector container (i.e., like record-count == 0) in the code that is used for generating that error message (currently _getActualSize()_ is only used for this purpose).

   Another code change: While testing on the Mac, the spilling code for regular file systems failed to delete leftover spilled directories (which were not empty). The code was changed to delete all the spill files first in a spill directory, and last the directory itself.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/common/HashTableTemplate.java
@@ -459,7 +459,11 @@ public abstract class HashTableTemplate implements HashTable {
         size += ledger.getAccountedSize();
       }
 
-      size += new RecordBatchSizer(htContainer).getActualSize();
+      // In some rare cases (e.g., making a detailed debug msg after an OOM) the container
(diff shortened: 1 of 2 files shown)


Ben-Zvi:
This code is called for cleanup from the SpillSet _close()_ method, and I somehow thought we could have a "delete race" situation (normally _delete_ is done by the operator), though obviously all this should happen sequentially (unless some external user deletes the spill files ....). 
   I'll change this to _delete()_ and retest.

Ben-Zvi:
Thanks for the comment @sohami ; I changed the code to use _delete()_ and ran some rigorous tests (on the Mac) and all ran OK (no leftover spill files/dirs).   The change was force pushed.

PR #1860, opened 2019-09-24
DRILL-7357: Expose Drill Metastore data through information_schema

arina-ielchiieva:
1. Add additional columns to TABLES and COLUMNS tables.
2. Add PARTITIONS table.
3. General refactoring to adjust information_schema data retrieval from multiple sources.

Jira - [DRILL-7357](https://issues.apache.org/jira/browse/DRILL-7357).


arina-ielchiieva:
@vvysotskyi please review.

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/Records.java
```diff
@@ -503,12 +568,97 @@ public Column(String catalog, String schemaName, String tableName, RelDataTypeFi
...
       }
+      this.COLUMN_FORMAT = null;
+      this.NUM_NULLS = null;
+      this.MIN_VAL = null;
+      this.MAX_VAL = null;
+      this.NDV = null;
+      this.EST_NUM_NON_NULLS = null;
+    }
 
+    public Column(String catalog, String schemaName, String tableName, String columnName,
```
vvysotskyi: Looks like this constructor contains well-refactored code with some logic similar to the logic in the constructor above. Is it possible to combine these constructors?

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
```diff
@@ -112,8 +108,7 @@ public String getDigest() {
 
   @Override
   public GroupScan clone(List<SchemaPath> columns) {
-    InfoSchemaGroupScan  newScan = new InfoSchemaGroupScan (this);
-    return newScan;
+    return new InfoSchemaGroupScan (this);
```
vvysotskyi:
```suggestion
    return new InfoSchemaGroupScan(this);
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
```diff
@@ -101,7 +97,7 @@ public int getMaxParallelizationWidth() {
   }
 
   @Override
-  public PhysicalOperator getNewWithChildren(List<PhysicalOperator> children) throws ExecutionSetupException {
+  public PhysicalOperator getNewWithChildren(List<PhysicalOperator> children) {
     return new InfoSchemaGroupScan (this);
```
vvysotskyi:
```suggestion
    return new InfoSchemaGroupScan(this);
```

cgivre:
Hello All, 
This looks really great.  QQ:  Is there any `README` or other documentation that would help users understand the capabilities here?  I'd volunteer to put one together if I could get some help.

arina-ielchiieva:
@cgivre there is no readme in this PR since it's just a part of Drill Metastore work and has no effect until core Metastore work will be committed. I believe once @vvysotskyi opens final PR for Metastore it will include all necessary documentation.

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
arina-ielchiieva: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
arina-ielchiieva: Fixed.

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/Records.java
arina-ielchiieva: Both constructors cannot be combined since they determine `Record.Column` properties based on different sources. One uses `RelDataTypeField`, another `ColumnMetadata`, `ColumnStatistics` and additional properties, mostly relying on `TypeProtos.MajorType`. Logic go obtaining `Record.Column` is different, so I suggest we leave both constructors.

Personally, I would prefer to remove some of the properties from `Record.Column`, like `NUMERIC_PRECISION_RADIX` etc. since I am not sure how much value they bring but I had to leave them for backward compatibility.

arina-ielchiieva:
@vvysotskyi thanks for the code review, addressed comments.

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/Records.java
vvysotskyi: Ok, thanks for the explanation.

common/src/main/java/org/apache/drill/common/types/Types.java
```diff
@@ -93,6 +93,44 @@ public static boolean isNumericType(final MinorType type) {
...
+      case TIMETZ:
+      case DATE:
+      case TIMESTAMP:
+      case TIMESTAMPTZ:
+        return true;
+      default:
+        return false;
+    }
+  }
+
```
arina-ielchiieva: Check.

PR #1863, opened 2019-09-25
DRILL-7387: Failed to get value by int key from map nested into struct

ihuzenko:
Covers case from [DRILL-7387](https://issues.apache.org/jira/browse/DRILL-7387).


ihuzenko:
@KazydubB please review.

PR #1867, opened 2019-10-06
DRILL-7358: Fix COUNT(*) for empty text files

paul-rogers:
Fixes a subtle error when a text file has a header (and so has a
schema), but is in a COUNT(*) query, so that no columns are
projected. Ensures that, in this case, an empty schema is
treated as a valid result set.

Tests: updated CSV tests to include this case.


paul-rogers:
@arina-ielchiieva, please review.

arina-ielchiieva:
+1, @paul-rogers thanks for making the changes.

PR #1868, opened 2019-10-07
DRILL-7377: Nested schemas for dynamic EVF columns

paul-rogers:
The Result Set Loader (part of EVF) allows adding columns up-front
before reading rows (so-called "early schema.") Such schemas allow
nested columns (maps with members, repeated lists with a type, etc.)

The Result Set Loader also allows adding columns dynamically
while loading data (so-called "late schema".) Previously, the code
assumed that columns would be added top-down: first the map, then
the map's contents, etc.

Charles found a need to allow adding a nested column (a repeated
list with a declared list type.)

This patch revises the code to use the same mechanism in both the
early- and late-schema cases, allowing adding nested columns at
any time.

Testing: Added a new unit test case for the repeated list late
schema with content case.

### DIFF
--- a/exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
+++ b/exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
@@ -20,6 +20,7 @@ package org.apache.drill.exec.vector.accessor.writer;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.record.metadata.ColumnMetadata;
 import org.apache.drill.exec.record.metadata.TupleMetadata;
@@ -32,6 +33,8 @@ import org.apache.drill.exec.vector.accessor.ScalarWriter;
 import org.apache.drill.exec.vector.accessor.TupleWriter;
 import org.apache.drill.exec.vector.accessor.VariantWriter;
 import org.apache.drill.exec.vector.accessor.impl.HierarchicalFormatter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Implementation for a writer for a tuple (a row or a map.) Provides access to each
@@ -143,6 +146,8 @@ public abstract class AbstractTupleWriter implements TupleWriter, WriterEvents {
     ObjectWriter addColumn(TupleWriter tuple, MaterializedField field);
   }
 
+  protected static final Logger logger = LoggerFactory.getLogger(AbstractTupleWriter.class);
+
   protected final TupleMetadata tupleSchema;
   protected final List<AbstractObjectWriter> writers;
   protected ColumnWriterIndex vectorIndex;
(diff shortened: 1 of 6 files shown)


paul-rogers:
@arina-ielchiieva, please review. @cgivre, this fixes the issue you found.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
```diff
@@ -143,6 +144,8 @@ public void dump(HierarchicalFormatter format) {
     ObjectWriter addColumn(TupleWriter tuple, MaterializedField field);
   }
 
+  protected static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(AbstractTupleWriter.class);
```
arina-ielchiieva: Please use full imports.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
```diff
@@ -205,20 +208,29 @@ public int addColumnWriter(AbstractObjectWriter colWriter) {
...
-    }
+    verifyAddColumn(column.name());
     final AbstractObjectWriter colWriter = (AbstractObjectWriter) listener.addColumn(this, column);
     return addColumnWriter(colWriter);
   }
 
   @Override
   public int addColumn(MaterializedField field) {
+    verifyAddColumn(field.getName());
+    final AbstractObjectWriter colWriter = (AbstractObjectWriter) listener.addColumn(this, field);
```
arina-ielchiieva: It would be nice if we could ged rid of casting to abstract version of object writer. We can add necessary methods to `ObjectWriter` interface or return `AbstractObjectWriter`. What do you think?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
paul-rogers: Done. But note that, when the code was converted from Logback to slf4j, all of the files ended up with the fully qualified class names. Perhaps someone should write a script to change all of these if we wish to use qualified names.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
paul-rogers: Agreed this is awkward; but it is less awkward than the alternative. The idea is that the interfaces provide a clean interface for clients. I tried making the change, but then the implementation class, `AbstractObjectWriter` starts leaking through in many places. For example, the `getColumn()` method in the `TupleWriter` must return `AbstractObjectWriter`.

So, my thought was that the least intrusive solution is to simply cast to the implementation type the few places that it is needed rather than expose the implementation classes to clients.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
arina-ielchiieva: Perhaps, maybe someone would like to do it :)

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/AbstractTupleWriter.java
arina-ielchiieva: Well, since you used this approach before, I guess it ok to leave it as is but its definitely not Java-way, you kinda expect everybody who wants to implement `ObjectWriter` to extend `AbstractObjectWriter` instead and since this is not mandatory, this could lead to cast class exception in future.

arina-ielchiieva:
+1

paul-rogers:
@arina-ielchiieva, commits squashed

arina-ielchiieva:
Thanks, all tests passed, merging...

PR #1870, opened 2019-10-12
DRILL-7359: Add support for DICT type in RowSet Framework

KazydubB:
Jira - [DRILL-7359](https://issues.apache.org/jira/browse/DRILL-7359).


KazydubB:
@paul-rogers can you please review?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/NullReader.java
```diff
@@ -0,0 +1,327 @@
...
+
+  private static final NullReader INSTANCE = new NullReader();
+
+  public static NullReader instance() {
+    return INSTANCE;
+  }
+
+  @Override
+  public int size() {
+    notSupported();
```
arina-ielchiieva: Please pass message into `notSupported()` method to show that exactly is not supported and why.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.record.metadata;
+
+import org.apache.drill.common.types.TypeProtos;
+import org.apache.drill.exec.record.MaterializedField;
+import org.apache.drill.exec.vector.complex.DictVector;
+
+public class DictBuilder implements SchemaContainer {
```
paul-rogers: Would be great to add a few comments to explain the DICT structure. Is it of the form DICT<KEY, VALUE>, where the key is of one type, the value of another? Or, can keys and values be of different types?

Since a MAP is a collection of (key, value) pairs in which the keys are declared, the keys are strings, and the values can each be of a distinct type, it is not likely that DICT does the same.

I will assume the DICT<KEY_TYPE, VALUE_TYPE> format in my comments here.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  public DictBuilder(SchemaContainer parent, String memberName, TypeProtos.DataMode mode) {
+    this.parent = parent;
+    this.memberName = memberName;
+    this.mode = mode;
+  }
+
+  @Override
+  public void addColumn(ColumnMetadata column) {
```
paul-rogers: If we agree DICT is DICT<KEY_TYPE, VALUE_TYPE> then we might want the building of a DICT to be something like:

```
    TupleMetadata = new SchemaBuilder()
      .addDict("myDict")
        .key(MinorType.VARCHAR)
        .valueArray(MinorType.DOUBLE)
        .resumeSchema()
      ...
```

That is, we need to add one key. Can keys only be scalars? Only non-nullable? If so, then the single `key()` method is all that is needed. This returns the same builder.

Can the value be of any type? Then we need the full set of type builders: scalar, array, map and so on.

The current form does not make this clear: what does it mean to add a column? This isn't a map...

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    assert DictVector.fieldNames.contains(column.name());
+    tupleBuilder.addColumn(column);
+  }
+
+  public DictBuilder addKey(TypeProtos.MajorType type) {
```
paul-rogers: Good, but we allow only one key, do we not? If so, then rather than `addKey()` (which suggests I can add multiple), a simple `key()` would be simpler. Also, we should check if the key type was already set and not allow setting it a second time.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    return add(MaterializedField.create(DictVector.FIELD_KEY_NAME, type));
+  }
+
+  public DictBuilder addValue(TypeProtos.MajorType type) {
+    return add(MaterializedField.create(DictVector.FIELD_VALUE_NAME, type));
+  }
+
+  public DictBuilder add(MaterializedField col) {
```
paul-rogers: Probably no longer need this form. It is in the code to ease conversion from old code, but does not actually seem to have been used.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    assert DictVector.fieldNames.contains(col.getName());
+    tupleBuilder.add(col);
+    return this;
+  }
+
+  public DictBuilder addKey(TypeProtos.MinorType type, TypeProtos.DataMode mode) {
```
paul-rogers: Does it make sense for a key to be repeated or nullable? Isn't non-nullable the only valid key mode?

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    tupleBuilder.add(DictVector.FIELD_KEY_NAME, type, width);
+    return this;
+  }
+
+  public DictBuilder addValue(TypeProtos.MinorType type, int width) {
+    tupleBuilder.add(DictVector.FIELD_VALUE_NAME, type, width);
+    return this;
+  }
+
+  public DictBuilder addKey(TypeProtos.MinorType type, int precision, int scale) {
```
paul-rogers: This form is used for `DECIMAL`. Does it make sense to allow a `DECIMAL` for a key? Maybe here, and in the width case above, just allow a `MajorType`.

Maybe, in general, the `SchemaBuilder` should have relied on the `Types` methods to build up such types rather than having methods for each...

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    return addDecimal(DictVector.FIELD_KEY_NAME, type, TypeProtos.DataMode.REQUIRED, precision, scale);
+  }
+
+  public DictBuilder addValue(TypeProtos.MinorType type, int precision, int scale) {
+    return addDecimal(DictVector.FIELD_VALUE_NAME, type, TypeProtos.DataMode.REQUIRED, precision, scale);
+  }
+
+  public DictBuilder addNullable(String name, TypeProtos.MinorType type) {
```
paul-rogers: Yeah, all these forms make sense when only working with the top-level row, but become tedious when repeated or all these specialized builders...

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+
+  public DictBuilder addDictValue() {
+    return tupleBuilder.addDict(this, DictVector.FIELD_VALUE_NAME);
+  }
+
+  public UnionBuilder addUnionValue() {
+    return tupleBuilder.addUnion(this, DictVector.FIELD_VALUE_NAME);
+  }
+
+  public UnionBuilder addListValue() {
```
paul-rogers: Again, if we can only have one value type, maybe use `listValue()`, etc. instead of `addListValue()`.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
```diff
@@ -0,0 +1,198 @@
...
+    return tupleBuilder.addList(this, DictVector.FIELD_VALUE_NAME);
+  }
+
+  public RepeatedListBuilder addRepeatedListValue() {
+    return tupleBuilder.addRepeatedList(this, DictVector.FIELD_VALUE_NAME);
+  }
+
+  public DictColumnMetadata buildColumn() {
+    return new DictColumnMetadata(memberName, mode, tupleBuilder.schema());
```
paul-rogers: Sorry, I'm confused. A `DICT` is essentially a tuple? I thought it was an array of (key, value) pairs? Essentially (key: ARRAY<KEY_TYPE>, value: ARRAY<VALUE_TYPE>)?

Let's clarify this before I comment more as I'll need to understand that to make useful comments elsewhere.

It may be that this class was patterned after the Tuple classes. But, if so, then those classes may have been poor examples because a MAP solves a different problem than a DICT (I think...)

paul-rogers:
Going to review this in multiple stages. First, let's pin down the desired semantics of the DICT type. See code comments.

paul-rogers:
Let's continue to assume that `DICT` is, essentially, `DICT<KEY_TYPE, VALUE_TYPE>` and that we can think of the `DICT`, when writing, as a pair of arrays: one for keys, one for values. (Because vectors are write-once, we have to add (key, value) pairs one by one. If so, then we need a new form of writer, a `DictWriter` that has semantics such as:

```
  ObjectWriter key();
  ObjectWriter value();
  void save();
```

The `key()` gives us an object writer that lets us access the key writer. If we restrict keys to be scalars, then we can just do:

```
  ScalarWriter key();
```

Values can, I imagine, be of any type. So, the `ObjectWriter` lets us work with them generically. Suppose we had a `DICT<VARCHAR,DOUBLE>`, we could do:

```
   DictWriter dictWriter = rowWriter.dict("myDict");
   dictWriter.key().setString("fred");
   dictWriter.value().scalar().setDouble(123.45)
   dictWriter.save();
   dictWriter.key().setString("barney");
   dictWriter.value().scalar().setDouble(98.76)
   dictWriter.save();
```

This means that, in the `ObjectWriter()`, we need to add a new method, `dict()`, which will return a `DictWriter`, and we need to define the `DictWriter` interface.

There is quite a bit of commentary in the column accessor package that (tries) to explain the structure behind these writers. Perhaps that might help explain the ideas here.

paul-rogers:
On, read, we can access the `DICT` as a dictionary and as an array. This means we need to be a bit clever about designing the `DictReader()` interface.

We can get our `DictReader` similar to the write case:

```
   DictReader dictReader = rowReader.dict("myDict");
```

Now, we have two ways to proceed. Either we want key/value lookups, or array access.

For array access, we want an iterator, much like the `ArrayReader`, but that gives us access to both the key and value. Something like:

```
interface DictReader ...
   ScalarReader key();
   ObjectReader value();
   boolean next();
```

For key/value, we have a challenge. We need a way to specify the key that is type-specific. Then, we need to access the key in a type-specific way. This is the first time we've had such an issue.

Maybe we can build this on the array iterator. In addition to the `next()` method, we can provide a `findFoo()` method, with one for each supported key type.

```
   int findString(String key);
   int findInt(int key);
   ...
```

The semantics might be that the array index is set to the (first) index of the requested value, and that index is returned.

If not found, the return value is -1 and the array index is one past the end of the array.

```
   DictReader dictReader = ...
   if (dictReader.findString("barney")) {
     value = dictReader.value().scalar().getDouble();
   }
```

There are probably other variations. Think how you expect to use the reader (for testing, yes, but what other operations)?

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
KazydubB: The `key` can be of any `REQUIRED` primitive type and `value` can be of any primitive or complex types. Dict's key is not limited to be a string.
Yes, your assumption is correct.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
KazydubB: Yes, naming of the method can be confusing in this case (the method comes from `SchemaContainer`). Added validation to make sure only `key` and `value` fields may be added if they were not previously added.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
KazydubB: Changed the name to the `key()` and added validations, thanks!

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
KazydubB: Yes, `key` can be a `DECIMAL`. Added `key(MajorType)` method to allow creation for the case (the methods validates allowed `key` types).

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
KazydubB: Indeed, the column metadata was patterned after `TUPLE` classes. This may seem weird, but I see it perfectly fit: `dict` is a "tuple" with two fields (`key`, which has type restrictions, and `value`). This "tuple" is then handled by a special writer, `DictWriter`, and a special reader `DictReader`.

KazydubB:
@paul-rogers thank you for review. I've made some changes, please take a look.

There is a (interface) `DictWriter` similar to what you proposed with the exception it `extends` `ArrayWriter`, i.e. when writing a `dict` is assumed to be an array of tuples (with `key` and `value` fields) other than a pair of arrays. (`DictWriter`'s implementation, `ObjectDictWriter`, actually `extends ObjectArrayWriter`).

For reading, there is a (interface) `DictReader` (`extends ColumnReader`) introduces 5 new methods:
```
ObjectReader getValueReader(Object key);
Object get(Object key);
int size();
ValueType keyColumnType();
ObjectType valueColumnType();
```

For key/value lookup `getValueReader(Object key)` is used. When a `key` is found, `value`'s reader is returned with its position set correspondingly to the `key`. In a case there is no such `key` - a special type of `ObjectReader` - `org.apache.drill.exec.vector.accessor.reader.NullReader` will be returned.

I am not sure that array access is needed - can't see any use case yet.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/DictWriter.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.vector.accessor;
+
+/**
+ * Dict writer is represented as an array of Maps (Array with Tuple entry).
```
paul-rogers: Not sure I follow. A Dict is an array of key/value pairs, correct? It is a Map turned sideways: the key/value pairs are rows in an array, not columns in a map (as in the Map) type. So, a Dict is not "an array of Maps" is it?

If we want a Dict to allow heterogeneous value members, then we'd have an array of key/Union pairs, not key/Map pairs. Or, am I missing something?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/DictWriter.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+public interface DictWriter extends ArrayWriter {
+  ValueType keyType();
+  ObjectType valueType();
+  ScalarWriter keyWriter();
+  ObjectWriter valueWriter();
```
paul-rogers: The above is getting close, but it does introduce an ambiguity. The array writer already has the following members:

```
 ObjectType entryType();
 ObjectWriter entry();
```

Along with some other helpers. In the `DictWriter` case, it is not clear that I could use this object a `ArrayWriter`. Which of above does the entry correspond to?

Might it be better to have `DictWriter` inherit directly from `ColumnWriter` so you can define clean semantics?

If it is worth having common functionality (only `save()` would be common), we could define a `SequenceWriter` which is the base of `ArrayWriter` and `DictWriter`.

Or, you could change `ArrayWriter` to add a new element writer; a `DictEntry` which would provide the methods above. That is:

```
  ArrayWriter aw = reader.get('myDict').array();
  DictWriter dw = aw.dict();
  for (int i = 0; i < kvCount; i++) {
    dw.keyWriter().setString("key + " i);
    dw.valueWriter().union().setInt(i);
     aw.save();
  }
```

That is, a Dict is an array. The entries in the array are key/value pairs. The key is a scalar, the value is of some type (above I suggested a Union -- with all the problems that Unions bring.)

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/NullReader.java
```diff
@@ -0,0 +1,369 @@
...
+import org.joda.time.LocalDate;
+import org.joda.time.LocalTime;
+import org.joda.time.Period;
+
+import java.math.BigDecimal;
+
+/**
+ * Dummy reader which returns {@code null} for scalar types and itself for complex types.
+ */
+public class NullReader implements ScalarReader, ArrayReader, TupleReader, VariantReader, ObjectReader, DictReader {
```
paul-rogers: Good idea. But, is this too general? Does it follow the actual metadata structure?

The writer has a `NullScalarWriter` and a few other components that try to allow only structurally-legal operations. I suppose, however, that is less important when reading: being lenient won't break anything.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
paul-rogers: Follow-up question. For the value, is a Dict defined as, say, `DICT<STRING, INT>` so that all values are of the same type? If so, that is the easy case.

Or, can I define a Dict as `DICT<STRING, UNION<INT, STRING, DOUBLE>>`, meaning that key values can be any one of the three types? Note that, in Drill, we use the (barely-working) Union vector to handle this situation.

Or, is the Dict trying, itself, to be a Union for the value type?

Just read the (very nice) comments in `DictVector`, so it looks like, if I want my Dict to hold three types, I make it a `DICT<STRING, UNION>`. Correct?

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
paul-rogers: As a side comment, I will take the blame for the mess we are creating in the schema builder. When I first wrote it, I was mostly concerned with the top-level row and with maps. Then came repeated lists, unions and now Dict.

Is there a cleaner way to handle the nesting? Something clever with parameterized types to avoid all the 'resumeFoo' methods? Feel free to be creative here.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/MetadataUtils.java
```diff
@@ -205,4 +205,10 @@ private static ColumnMetadata newDecimal(String name, MinorType type, DataMode m
     return new PrimitiveColumnMetadata(field);
   }
 
+  private static boolean isScalar(MinorType type) {
+    return type != MinorType.MAP
+        && type != MinorType.UNION
+        && type != MinorType.LIST
+        && type != MinorType.DICT;
+  }
```
paul-rogers: Maybe change this to `isComplex`, and define `isScalar()` as `! isComplex()`.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/MapBuilder.java
```diff
@@ -144,6 +144,14 @@ public RepeatedListBuilder addRepeatedList(String name) {
     return tupleBuilder.addRepeatedList(this, name);
   }
 
+  public DictBuilder addDict(String name) {
+    return tupleBuilder.addDict(this, name);
+  }
```
paul-rogers: Maybe `addDict(String name, MinorType keyType)`. Works because the key must be a scalar and is required. Then the `DictBuilder` simply builds the value type:

```
scalarValue(MinorType type);
MapBuilder mapValue();
MapBuilder mapArrayValue();
UnionBuilder unionValue();
Foo resumeFoo();
```
Can the value be nullable? (The comment in `DictVector` doesn't say yes or no.) If nullable is allowed, then only the scalar can be nullable, so maybe add:

```
nullableValue(MinorType type)
```

paul-rogers:
@KazydubB, thanks much for the changes. This is a complex area. It looks like you are getting a good understanding.

My focus on this review is to help you define the simplest implementation for the schema builder. Maybe we can get that part finished up.

Then, I made a suggestion about how to structure the Dict writer. The new design here is much improved. Still, I think we can improve it a bit more.

Everything said for the writer applies to the reader, so I didn't repeat the comments there.

Once we get the API correct, then I'll dive into the implementation details. That should be a quick review if the we agree on the API, and if you have complete unit tests.

paul-rogers:
The schema builder issue has been nagging at me. I observed above that the current way of handling nested types is getting too complex, with all the special cases and "resumeFoo" methods.

The current design was originally designed to build a `BatchSchema` with `MaterializedField`s,. Those classes require a top-down construction. (That is also why the final `build()` method returns a `BatchSchema` and why we must call `buildSchema()` to get a `TupleMetadata`.)

It turns out that the `SchemaMetadata` classes *do not* have the same constraints. They can actually be built bottom-up. It never occurred to me to clean up the old-style of building when we repurposed `SchemaBuilder` to build metadata objects.

So, since we can change the structure, we could borrow a technique from, I think, certain JSON builders. Rather than having the schema builder create the nested type builder, ask the user to do so explicitly:

```
TupleMetadata schema = new SchemaBuilder()
  .addMap(new MapBuilder("myMap")
    .add("name", MinorType.VARCHAR)
    .addMap(new MapBuilder("nested")
      .add("id", MinorType.INT)
      .build())
    .build())
  .add("last", MinorType.INT)
  .build();
```

Adopting this model as additional benefits. It is currently a bit awkward to create a bare `ColumnMetadata`. We could create helper methods to create each kind of column schema. Something like:

```
foo = SchemaBuilder.column("foo", MinorType.INT, DataMode.REQUIRED);
bar = SchemaBuilder.dict("bar", MinorType.VARCHAR)
  .value(SchemaBuilder.map(...))
  .build();
```

Lots of things we can play with.

Given all of this, my advice for the `DICT` type is to figure out something that works with the structure we already have. Then, as a separate PR, we can clean up the `SchemaBuilder` to make it easier to use for the many complex types we now support. That way, you don't have to solve the schema builder problem in this PR.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/MapBuilder.java
KazydubB: Good point. Yes, `value` can be nullable (added explicit comment to `DictVector`).

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/MetadataUtils.java
KazydubB: Yeah, much negation... Changed the body of the method - didn't want to introduce the new `isComplex()`.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/NullReader.java
KazydubB: It does not follow the actual metadata structure, I suppose general case will do.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/DictWriter.java
KazydubB: My initial intention was to make it something like you proposed in your second option, though was done so implicitly. Made it the way you described with the exception that `DictWriter` actually encapsulates '`DictEntry`' (no need to perform `DictWriter dw = aw.dict()` rather use `DictWriter` itself to write the pairs; `aw` in this case is also `DictWriter`).

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/DictWriter.java
KazydubB: I wanted to say the dict writer is an array of dict `entries` (you were talking about below), just used the `map` somewhat in another context, like some type of tuple.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
KazydubB: I suppose, this will do for now (at least I see it OK). Changed the validation to check if the added field is `value` as `key` can't be complex and the method does not make sense for the case.

exec/vector/src/main/java/org/apache/drill/exec/record/metadata/DictBuilder.java
KazydubB: Yes, `Dict` is defined as `DICT<STRING, INT>` to represent mapping (like in Java's Map<String, Integer>) from `String` to `int`. 
`Dict` does not try to be a `Union` for the value type. If one wants the `Dict` to have `Union` value, one must define it as `DICT<STRING, UNION<...>>` (the `key` type is going to be `STRING` and the `value` is going to be a union of some types; like union with all problems it has in Drill :) )

KazydubB:
@paul-rogers I've updated the PR, please take a look again.
Thanks for your insight!

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/model/single/BaseReaderBuilder.java
```diff
@@ -77,22 +81,48 @@ protected AbstractObjectReader buildVectorReader(ValueVector vector, VectorDescr
...
     }
   }
 
+  private AbstractObjectReader buildDict(ValueVector vector, VectorAccessor va, VectorDescrip descrip) {
+
+    boolean isArray = descrip.metadata.isArray();
+
+    DictVector dictVector;
+    VectorAccessor dictAccessor;
+    if (isArray) {
```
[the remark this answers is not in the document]
KazydubB: To clarify, this builds reader for `REPEATED DICT` and `DICT`. Its fields (`key` and `value`) are handled in `buildMapMembers(AbstractMapVector, MetadataProvider)` which was written for handling fields for `MAP` and `REPEATED MAP` and now for both cases of `DICT`. (Yeah, the method probably should've been renamed taking into account its usage for `DICT`).

Is the above which concerns you or have I unterstood you incorrectly?

exec/java-exec/src/test/java/org/apache/drill/test/rowSet/test/TestRowSet.java
```diff
@@ -586,6 +595,490 @@ public void testRepeatedMapStructure() {
...
+    RowSetUtilities.verify(expected, actual);
+  }
+
+  /**
+   * Utility method to bootstrap a map object easily.
+   *
+   * @param entry key-value sequence
+   * @return map containing key-value pairs from passed sequence
+   */
+  private Map<Object, Object> map(Object... entry) {
```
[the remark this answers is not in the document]
KazydubB: I've tried, but there was an issue with entry ordering (used `HashMap`). Should work with `LinkedHashMap`, I'll try (oops, haven't thought of it back then).

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/NullReader.java
```diff
@@ -0,0 +1,377 @@
...
+
+  @Override
+  public ObjectType valueColumnType() {
+    notSupported("valueColumnType()");
+    return null;
+  }
+
+  private void notSupported(String methodName) {
+    throw new UnsupportedOperationException(methodName + " is not supported in NullReader");
+  }
```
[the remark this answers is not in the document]
KazydubB: I know this, but there is a need to `throw` and current form contains `throw` part. On the other hand, the form you are proposing proposes (somewhat) better code readability and there is no need to write `return`s all the time. Will change.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/NullReader.java
```diff
@@ -0,0 +1,377 @@
...
+  @Override
+  public ScalarReader scalar(TypeProtos.MinorType type) {
+    notSupported("scalar(TypeProtos.MinorType)");
+    return null;
+  }
+
+  @Override
+  public TypeProtos.MinorType dataType() {
+    notSupported("dataType()");
+    return null;
```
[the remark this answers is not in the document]
KazydubB: Yes, you are definetely right. Will add the actual schema to `NullReader`.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/ObjectDictWriter.java
```diff
@@ -0,0 +1,132 @@
...
+  public void setObject(Object object) {
+    if (object instanceof Map) {
+      Map<Object, Object> map = (Map<Object, Object>) object;
+      for (Map.Entry<Object, Object> entry : map.entrySet()) {
+        entryObjectWriter.setObject(0, entry.getKey());
+        entryObjectWriter.setObject(1, entry.getValue());
+        save();
+      }
+    } else {
+      if (object == null) {
```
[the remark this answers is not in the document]
KazydubB: Well, I'm abusing that `null` is not an `instanceof` any type. Will move the null check to the start, to show my intentions better.

Yes, stepping through the array with step 2 will be much better :)

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/NullReader.java
KazydubB: I agree with your thoughts: will use existing `AbstractScalarReader.NullReader` in case of scalar values and will change the `NullReader` for complex case in a similar (to `AbstractScalarReader.NullReader`) maner.

KazydubB:
@paul-rogers, I've updated the PR, please look again

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/UnionReaderImpl.java
```diff
@@ -268,4 +274,51 @@ public String getAsString() {
...
+    AbstractObjectReader[] nullVariants = new AbstractObjectReader[variants.length];
+    for (int i = 0; i < variants.length; i++) {
+      nullVariants[i] = variants[i].createNullReader();
+    }
+    return new NullUnionReader(schema(), unionAccessor, nullVariants);
+  }
+
+  private static class NullUnionReader extends UnionReaderImpl {
+
+    private NullUnionReader(ColumnMetadata metadata, VectorAccessor va, AbstractObjectReader[] variants) {
```
paul-rogers: This approach is a bit better than the big null reader we had before. I wonder, however, why we need such a reader? Null (AKA "dummy") readers exist, but generally for specialized cases (such as a List that has no type.) I wonder, is this for the case where the value of a key/value pair is null?

If so, then two thoughts. First, the right way to handle that is to:

* Use the `isNull()` method on the `ColumnReader` class:

```
if (myDictReader.value().isNull()) {
  // Handle null value
} else {
  int x = myDictReader.value().scalar().getInt();
}
```

The idea is that null values are null: they have no structure. The approach here seems to say that null values have structure (such as a null union), so we can ask what types the null has and so on. But, this is, perhaps, too much of a fantasy.

The other suggestion would be to suggest that values cannot be null, especially if they are of structured types.

Also, note that, for unions, we already have multiple levels of nulls: The union itself can be null. When used in a list, the list entry can be null. The union can have nullable types, so that the union might have an int, but the int is null. This is already a mess, we probably don't want to make it even more complex.

KazydubB: Yes, this is for the case "where the `value` of a `key`/`value` pair is null". This `value` can be `null` for a given `key` if the `key` is not present in the dict for a given row (e.g. if you have a `Map<String, Integer>` with entry `"a" : 1` and you try to `Map#get("b")`, with `Map` being a Java map in the example). That is, the value may be non-nullable, but when you fetch a `value` by `key` and there is no entry for the given key (in a row) then this `Null*Reader` is used. The `Null*Reader`s are created in https://github.com/apache/drill/pull/1870/files#diff-da401fee5976aec49192b5fb6e22de3eR83. This structures are handled outside (of particular `DictReader`) so the structure is retained for uniformity. 
The "dummy" readers are not projected (merely a placeholder, aren't they?), so this is not the way to use them(?).

arina-ielchiieva:
@KazydubB looks like `org.apache.drill.exec.physical.resultSet.project.ProjectionType` does not support Dict type, could you please check?

paul-rogers:
@arina-ielchiieva, @KazydubB, it turns out that `ProjectionType` does support `DICT`. This is a subtle topic. The projection type encodes information we can glean from the `SELECT` list, as encoded in the `SchemaPath`: `a`, `a[3]`, `a.b`. As a result, `DICT` is already covered: a reference will be either to the whole dict: `d`, or to a member: `d.m`.

We would change the `ProjectionType` only if we enhance the `SchemaPath` to, say, allow Python-like syntax: `d[`foo`]. Did we make any such changes?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/UnionReaderImpl.java
paul-rogers: We may want to define our semantics here. If I understand your description, we have:

* The entire `DICT` is `NULL` (is this allowed)?
* The `DICT` has an entry for key `x`, but the value is null.
* The `DICT` has an entry for the `NULL` key.
* The `DICT` has no entry for the key `x`, meaning that the value is implicitly `NULL`.

We have similar situations with the `UNION` type, perhaps we can learn from that:

* The entire `UNION` can be `NULL`.
* The `UNION` can have an entry for, say, `VARCHAR`, but the `VARCHAR` is `NULL`.
* The `UNION` does not have an entry for a given type (say, `INT`), so the value is implicitly `NULL`. (Although we don't allow access to such a type.)

My suggestion would be that we use the `isNull()` method for all the `NULL` cases:

* If the `DICT` is `NULL` for a row, then `isNull()` returns true for the `DICT` reader.
* If the `DICT` has a `NULL` key, then the key `isNull()` for the key returns true.
* If the `DICT` has a key, but the value is `NULL`, then `isNull()` for the value returns true.

Not the tricky case:

* If the `DICT` does not have a key, then we may want a `hasKey()` method, and the value reader should be a Java `null` because it does not exist.

Or:

* If the `DICT` does not have a key, then the value reader's `isNull()` returns true. (This means we cannot differentiate the has-key-with-null-value vs. does-not-have-a-key cases.)

KazydubB: Regarding points of your suggestion:
- `DictReaderImpl` has `NullStateReaders.REQUIRED_STATE_READER` `NullStateReader`, i.e. it can't be `NULL` (like `MapReader`).
- `DICT` can't have `NULL` key as it is `REQUIRED`.
- If the `DICT` has a key, but the value is `NULL`, then `isNull()` for the value returns `true` - yes.
- in the (tricky) case of a missing key from the `DICT`, the latter approach was taken - there is no differentiation between missing key and `NULL` value. (The new `Null*Reader`s are introduced to be used in the case of a missing key - their `isNull()` always returns `true`; note, that in case when a key is present and the value is `NULL`, current (non-null) readers' `isNull()` will return `true` for that particular position, as corresponding value reader keeps track of nullability). The former approach was discarded to ensure uniform behavior without a chance of getting `NPE` or making `null`-checks.

KazydubB:
@paul-rogers, `ProjectionType` now has no notion of `MinorType#DICT` and treats it as `ProjectionType#SCALAR`. And yes, we do need to allow Python-like syntax for the type. (My bad, haven't made changes to this).

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/UnionReaderImpl.java
paul-rogers: @KazydubB, thanks for the explanation. It clarifies something I wondered about: the role of the null state readers. Let's talk a bit about the intended semantics for the readers.

I should be able to write code that caches the readers:

```
DictReader dr = rowReader.dict("myDict");
ScalarWriter kw = dr.key().scalar();
MapWriter vr = dr.value.tuple();
```

That is, I should be able to use the same key reader (`kr`) and value reader (`vr`) for all members. Then, I should either be able to iterate over the values:

```
while (dr.next()) {
  System.out.println(kr.getString());
  System.out.println(vr.getAsString());
}
```

If things work that way, then we don't need a special reader for the null state: the value reader (here, `vr`) would simply return `true` from `isNull()`:

```
while (dr.next()) {
  System.out.println(kr.getString());
  if (vr.isNull()) {
    System.out.println("NULL");
  } else {
    System.out.println(vr.getAsString());
  }
}
```

(Ignoring, for the moment, that `getAsString()` already does a null check.)

Note that, if we read past the end, both the key and value writers will point to nothing. I think that, for the array reader, accessing the value one past the end is an error. For `DICT`, we could simply declare that the value is NULL.

Then, I was going to sketch out the same behavior if we search for a key and found...nothing. Seems that the `DictReader` has no way to search for a key. If it did, it would not be clear what type to use for the key. So, that is a bit of a mess.

Let's assume we did have something to find a specific value, like we do for the arrays where we have `setPosn(int index)`. Maybe we have `findStringKey()` etc. For testing, we could just have `findKey(Object key)` and do a switch on type as we do for the other readers. So:

```
  dr.findStringKey("foo");
  if (vr.isNull()) {
    System.out.println("NULL");
  } else {
    System.out.println(vr.getAsString());
  }
```

Or, even:

```
  if (dr.findStringKey("foo")) {
    System.out.println(vr.getAsString());
  }
```

The only way to find the value is to do a linear search. If we do a linear search, and no value exists, the index will point one past the end of the dict. And, above, we suggested that the key and value readers could return `true` for `isNull()` in those cases.

Given this, unless I'm missing something, I don't see that we need any kind of null value reader at all. In fact, I'd even say that we don't want such a reader if we'd have to do:

```
ObjectReader kr = dr.findStringKey("foo");
if (kr.isNull) {
  // kr is a special null reader
} else {
  // kr is a Map reader
}
```

The above clutters the code and makes it impossible to cache the value reader, which will slow inner-most loops that want to use the reader abstractions.

What do you think?

paul-rogers:
On `ProjectionType`, if the `SchemaPath` is, say, `foo`, then the `ProjectionType` will be `SCALAR`. I can see that this name is misleading. Perhaps `SIMPLE` is better. We don't know that the column **is** scalar, we just know that it is consistent with a scalar. It could be a map, array or dict as well.

But, if you do add a `SchemaPath` representation for `foo['bar']`, then, yes, that can only be a `DICT`, so we can introduce a `DICT` projection type.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/UnionReaderImpl.java
KazydubB: @paul-rogers, first of all, thank you for your thoughts - they are very useful.

Regarding the first case you described (I suppose you meant `ScalarReader kw` and `MapReader vr` in the first code snippet):
one is able to write code with cached readers currently, if needed for such iterations over key-value pairs in the form you wrote it
```
while (dr.next()) {
  System.out.println(kr.getString());
  System.out.println(vr.getAsString());
}
```
with the behaviour you expect (without a special `Null*Reader`s). But there is no need for the case (here and below) to add a special `NULL`-`NULL` entry past the end, as iteration should be over _true_ collection of entries, which were added to the `Dict` (through `DictWriter`).

Regarding finding specific values:
there is a `DictReader#getValueReader(Object key)` method (with return type `ObjectReader`), which allows to obtain `value`'s reader with its position set to the one corresponding to needed `key`. If there is no one for the `key` - the special `Null*Reader` (depends on the type of `value`) is returned. So, this approach enforces non-caching of the value reader (as you have pointed out). The idea was to make `null` `value`s to be handled the same, indepently whether the `null` is from the actual value (i.e., from some entry with some `key` and `null` `value` `"someKey" -> null`) or from absent `key`, e.g. your example:
```
ObjectReader kr = dr.findStringKey("foo");
if (kr.isNull) {
  // kr is a special null reader
} else {
  // kr is a Map reader
}
```
is likely should be done, because of the nature of complex types in Drill - they are generally non-nullable (at least the ones in standard, pre-EVF, Drill's framework) leading to no distinction between, say, empty or `null` `MAP`.

But good point about poorer perfomance (with the current approach).

I like your idea with `boolean findKey(...)` - something similar is done in standard framework as well. Perhaps, I will stick to it. (In this case, your suggestion about treating entries by the end as if they were `null` does make sense).

About designated method for each key type - I assume it is redundant, as `key` type validation is better to be performed once prior to obtaining the data, and general case of `Object key` should do, shouldn't it? 

(Just want to note, that array-like behaviour is useful in testing primarily, the "real" use-case is to obtain elements by key).

P.s. Sorry, missed the comment yesterday.

KazydubB: Yes, there is a need to have multiple methods for each key type to avoid invocation of generalized `keyReader.getObject()` which contains `switch` checking each type (really bad in production) - will fix it.

paul-rogers: @KazydubB, first, thanks for understanding my hastily-written example; I've edited it to be accurate.

We are defining an API. APIs exist to support use cases, so it is very helpful to think about those. (My bad for not doing that earlier; I was focused more on the plumbing aspects early on. This is, in fact, why I still like design docs, even in the era of Agile, so we can visualize these issues early on.)

In my view, there are two key use cases: tests and production.

In tests, we simply want to verify that, say, a reader correctly created (or filtered, or altered) a `DICT`. If a key exists, we want to verify its value. If a key does not exist, we simply need to verify that fact. So, having a generic `boolean findKey(Object key)` is perfectly fine. The overhead of doing a type-based switch statement on every call is fine. (This is why tests often use the `addRow()` methods which, internally, have a big, slow `if` statement based on the type of the value.)

The second use case is production. Here, we can assume our goal is maximum performance via generated code.  I would argue that we do not want any extra code in the access path. Instead, the code should be specific to each type. This is why the readers are designed to go directly from, say, `getString()` into the code that grabs the bytes from the underlying value vector.

For production code, we'd want a type-aware "find" operation. There are a couple of ways to do that. One is by adding a "key accessor":

```
class DictReader ...
   KeyAccessor keyAccessor();
  ...
}

interface KeyAccessor {
  boolean findInt(int key);
  boolean findString(String key);
  ...
}
```

The accessor code gen could then generate type-specific search code by creating a key accessor class for each supported key type. This would be done in the build-time code gen, using Freemarker, as part of the existing `columnAccessors.java` template and code gen.

Now things get interesting. We need to generate code for each SQL operation that uses a `DICT`. Let's make up an example:

```
SELECT ... FROM myTableWithDict WHERE d['foo'] = 20
```

Here, the story splits into two paths. Today, code gen works directly against value vectors: you will need to generate the find code inline for each operator. Longer term, it is my hope that we can shift to using the column readers. (Doing so will simplify code gen, and puts us in a position to use multiple value vector format, such as Arrow, if we wanted.)

So, let's imagine we are using your `DICT` readers in runtime code generation. We want to generate code for the above example. 

In either case (old-school or reader-based), we don't know the type until the first row appears. At that point, we generate run-time code that, in our example, uses a `VARCHAR` key to lookup an `INT` value and send it to the equality function. Very crudely:

```
DictReader dr;
KeyAccessor ka;
ScalarReader vr;
String constKey;
String constArg;

void doSetup() {
  dr = rowReader.dict("d");
  ka = dr.keyAccessor();
  vr = dr.value().scalar();
  constKey = "foo";
  constArg = "bar";
}

void doEval() {
  boolean isEquals;
  boolean found = ka.findString(constKey);
  if (found) {
    isEquals = strEquals(vr.getString(), constArg);
  } else {
    isEquals = false;
  }
  // do something with isEquals...
}
```

The above is written in "code gen style", but greatly simplified.

So, we've looked at two use cases: test and production. In neither did we need to work with a missing value. Instead, we observed the rules that a non-existent key has no value (rather than a non-existent key having a value which is structurally the same as a real value, just null.)

Think of Java. A `null` in Java is the absence of a value, it is not, say a `Foo` object, with all the same fields and methods, that somehow has special `null` semantics. (Yes, this is a huge source of debate in CS. Some languages, like Scala, try to use objects, like `Option`, instead of `null`. But, `null` is still a higher-performance solution.) Further, in SQL, `NULL` is defined as the absence of a value, not a value (say a map) with special `NULL` semantics.

So, I don't think you need to try to define null readers: a `NULL` is missing value: there is nothing there, no structure to explore, no value to get.

Does this make sense?

In summary:

* Define the use cases you want to support.
* Define how those use cases will look up keys and how they will handle non-existent keys.
* Refine the reader API to work well for those use cases.

The good news is that the core functionality is now looking pretty good; we're just fine tuning the functionality to best serve the potential users of your new code.

arina-ielchiieva:
Please fix bug in `org.apache.drill.exec.record.metadata.DictBuilder`
```
  public DictBuilder resumeMap() {
    build();
    return (DictBuilder) parent;
  }
```
Resume map should return map builder.

paul-rogers:
Would be great if we could push this over the finish line. 

There will be a conflict with DRILL-7486.

I made some suggestions about finding DICT values and handling missing values. Let me know if I can help with that work. If we're not sure what we want to do in this area, then, as noted earlier, might be good to commit what we have; then we can refine it over time as we try the code.

KazydubB:
@paul-rogers, I'm sorry for taking so long. I'm planning to finish this in a few days.

arina-ielchiieva:
@KazydubB I see you have pushed new commit, is PR ready for review? Please tag Paul if yes.
Also please check if `org.apache.drill.exec.record.metadata.DictBuilder` method `public DictBuilder resumeMap()` was fixed.

KazydubB:
@paul-rogers, I've updated the PR, please have a look again.
@arina-ielchiieva this is now fixed, thanks for pointing this out!

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ContainerState.java
```diff
@@ -172,4 +172,8 @@ public void close() {
       colState.close();
     }
   }
+
+  public boolean isDict() {
+    return false;
+  }
```
paul-rogers: Should clients of this class care what kind of container this is? Shouldn't all DICT-specific code be in the DICT related classes?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/TupleState.java
```diff
@@ -567,4 +574,187 @@ public void dump(HierarchicalFormatter format) {
...
+
+    @Override
+    public AbstractTupleWriter writer() {
+      return (AbstractTupleWriter) parentColumn.writer().array().dict().tuple();
+    }
+  }
+
+  public static class DictVectorState implements VectorState {
+
+    private final ValueVector vector;
```
paul-rogers: Part of the goal of this class is to provide the actual vector class. Should this be `DictVector`? Same with `vector()`?

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/AbstractKeyAccessor.java
```diff
@@ -0,0 +1,91 @@
...
+  protected final ScalarReader keyReader;
+
+  protected AbstractKeyAccessor(DictReader dictReader, ScalarReader keyReader) {
+    this.dictReader = dictReader;
+    this.keyReader = keyReader;
+  }
+
+  @Override
+  public boolean find(boolean key) {
+    throw new UnsupportedOperationException(this.getClass().getName() + " does not support boolean key.");
```
paul-rogers: Nit:

```
  throw unsupported("boolean");
...

private UnsupportedOperationException unsupported(String type) {
  return new UnsupportedOperationException(String.format(
      "%s does not support a key of type %s.",
      getClass().getName(), type));
}
```

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/TupleState.java
KazydubB: This class is used for both `DictVector` and `RepeatedDictVector`. Unlike for `MAP`, they do not have a common parent class (other than `RepeatedValueVector` interface). Should this be changed to `RepeatedValueVector` or should I create another `VectorState` for `RepeatedDictVector`?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ContainerState.java
KazydubB: It is better, this was removed, yes. I'll try to eliminate this.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ColumnBuilder.java
```diff
@@ -179,6 +189,19 @@ private ColumnState buildPrimitive(ContainerState parent, ColumnReadProjection c
         vectorState);
   }
 
+  /**
+   * Check if this is a special case when vector, writer and column state should be
+   * created for a primitive field though the field itself is not projected. This is
+   * needed because {@code DICT}'s {@code keys} field is not projected but is needed
+   * to be initialized to ensure the dict vector is constructed properly.
```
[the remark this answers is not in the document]
KazydubB: Oops, my bad - hasn't mentioned (just a kind of implicit hint in the `@return`) in the javadoc body (do not know what is the text before `@param`, `@return` etc. is called) that this affects creation of `DICT`'s `key` when `value` is accessed by key only, i.e. `col['someKey']` with `col` being a `DICT` column with `String` (`VARCHAR`) `key`; `value` in this case will be projected.
If the `DICT` is not projected, this won't create any vector: neither `DictVector` nor vector for `key` (or `value`) - special "dummy" vector for `DICT` will be used, as is done for other vectors.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/TupleState.java
paul-rogers: I see the challenge. Might be better to use two distinct classes to model the different implementations. The two "state" classes can have the same API to hide the differences from other code.

This class is responsible for translating Result Set Loader events into "do the right thing" actions on the underlying vectors.

For example, I'd imagine that rollover will be different between the two: the DICT has one offset vector, the Repeated DICT has two. Same is true for vector allocation.

This is tricky; I can't work it out in the abstract. Maybe play with the code to see what works.

exec/java-exec/src/test/java/org/apache/drill/exec/physical/resultSet/impl/TestResultSetLoaderDictArray.java
```diff
@@ -0,0 +1,263 @@
...
+import org.apache.drill.exec.physical.rowSet.RowSet.SingleRowSet;
+import org.apache.drill.test.rowSet.RowSetUtilities;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+/**
+ * Test dict array support in the result set loader.
+ */
+@Category(RowSetTests.class)
+public class TestResultSetLoaderDictArray extends SubOperatorTest {
```
paul-rogers: Thanks for the excellent and complete tests. The one addition I might suggest is to add an overflow test for both the array and non-array cases. Basically, you want to make the vector exceed 16 GB. I do this by using big strings (~ 512 bytes in size.) Use an odd size: 523 or some such, to that overflow occurs in an other than a even row count.

Then, use big keys to trigger overflow of the key vector. Check the resulting output batches. Then, repeat for the value vector.

Overflow is the most complex part of the result set loader. If you look at examples for VarChar fields and arrays, you might see what was done before. (Avoid the "Torture" test: that one is very complex because it tries to set up near-random, yet predictable, overflow patterns.)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ColumnBuilder.java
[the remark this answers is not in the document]
KazydubB: Dict works the same as arrays regarding projection: if the user refers to any `DICT` entry, the whole `DICT` is projected and finding the needed entry is to be handled by some other operator. This entry is not filtered in the Scan.

KazydubB:
@paul-rogers, I've updated the PR: added tests for `key` and `value` overflow for single and `DICT` array, introduced separate `VectorState`s for single `DICT` and `DICT` array, performed minor refacttoring. Please take a look.

PR #1872, opened 2019-10-13
DRILL-7402: Suppress batch dumps for expected failures in tests

paul-rogers:
Drill provides a way to dump the last few batches when an error
occurs. However, in tests, we often deliberately cause something
to fail. In this case, the batch dump is unnecessary.

This enhancement adds a config property, disabled in tests, that
controls the dump activity. The option is enabled in the one test
that needs it enabled.

Jira - [DRILL-7402](https://issues.apache.org/jira/browse/DRILL-7402).

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/TestOperatorDump.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/TestOperatorDump.java
@@ -21,6 +21,7 @@ import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.ConsoleAppender;
 import org.apache.drill.common.exceptions.UserRemoteException;
 import org.apache.drill.exec.exception.OutOfMemoryException;
+import org.apache.drill.exec.physical.impl.BaseRootExec;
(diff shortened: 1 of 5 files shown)


paul-rogers:
@arina-ielchiieva, this is a nice-to-have that makes running unit tests just a bit cleaner.

exec/java-exec/src/test/java/org/apache/drill/TestOperatorDump.java
```diff
@@ -34,16 +41,11 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Set;
-import java.util.stream.Collectors;
-
 import static org.junit.Assert.assertTrue;
 
+import ch.qos.logback.classic.spi.ILoggingEvent;
```
arina-ielchiieva: I thought static imports should be the last ones...

paul-rogers:
@arina-ielchiieva, fixed the import issue. To save time, went ahead and squashed commits.

arina-ielchiieva:
+1, @paul-rogers thanks for making the changes.

PR #1873, opened 2019-10-15
DRILL-6096: Provide mechanism to configure text writer configuration

arina-ielchiieva:
1. Usage of format plugin configuration allows to specify line and field delimiters, quotes and escape characters.
2. Usage of system / session options allows to specify if writer should add headers, force quotes.

Jira - [DRILL-6096](https://issues.apache.org/jira/browse/DRILL-6096).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/writer/TextRecordWriter.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/writer/TextRecordWriter.java
@@ -17,44 +17,49 @@
  */
 package org.apache.drill.exec.store.easy.text.writer;
 
-import java.io.DataOutputStream;
-import java.io.IOException;
-import java.io.PrintStream;
-import java.util.List;
-import java.util.Map;
-
+import com.univocity.parsers.csv.CsvFormat;
+import com.univocity.parsers.csv.CsvWriter;
+import com.univocity.parsers.csv.CsvWriterSettings;
 import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.record.BatchSchema;
+import org.apache.drill.exec.record.MaterializedField;
 import org.apache.drill.exec.store.EventBasedRecordWriter.FieldConverter;
 import org.apache.drill.exec.store.StorageStrategy;
 import org.apache.drill.exec.store.StringOutputRecordWriter;
+import org.apache.drill.exec.store.easy.text.TextFormatPlugin;
 import org.apache.drill.exec.vector.complex.reader.FieldReader;
-import org.apache.drill.shaded.guava.com.google.common.base.Joiner;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 public class TextRecordWriter extends StringOutputRecordWriter {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(TextRecordWriter.class);
+
+  private static final Logger logger = LoggerFactory.getLogger(TextRecordWriter.class);
 
   private final StorageStrategy storageStrategy;
+  private final Configuration fsConf;
 
+  private FileSystem fs;
   private Path cleanUpLocation;
-
   private String location;
   private String prefix;
-
-  private String fieldDelimiter;
   private String extension;
+  // indicates number of a file created by this writer: 0_0_{fileNumberIndex}.csv (ex: 0_0_0.csv)
+  private int fileNumberIndex;
 
-  private int index;
-  private PrintStream stream = null;
-  private FileSystem fs = null;
+  private CsvWriterSettings writerSettings;
+  private CsvWriter writer;
 
-  // Record write status
-  private boolean fRecordStarted = false; // true once the startRecord() is called until endRecord() is called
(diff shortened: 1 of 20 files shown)


arina-ielchiieva:
@vvysotskyi please review.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/writer/TextRecordWriter.java
```diff
@@ -93,29 +112,35 @@ public void startNewSchema(List<String> columnNames) throws IOException {
...
+      throw new IOException(String.format("Unable to create file: %s.", fileName), e);
     }
-    index++;
 
-    stream.println(Joiner.on(fieldDelimiter).join(columnNames));
+    fileNumberIndex++;
+
+    if (writerSettings.isHeaderWritingEnabled()) {
+      writer.writeHeaders(
+        IntStream.range(0, schema.getFieldCount())
```
vvysotskyi: Nit: it can be a little bit optimized to avoid looking up elements from the list by index:
```suggestion
        StreamSupport.stream(schema.spliterator(), false)
```

arina-ielchiieva: Good point, thanks for noticing. Fixed.

paul-rogers:
General question: the idea of giving the user more control is a good one. I'm seeing this need in multiple of recent plugins. I wonder, however, if we should offer a more scalable solution?

We currently have plugin, session, config and table function options all which control some aspect of reading and writing. The plugin and config options really want to be "bulk" or "slow moving" settings: set them once for a system or file format.

Session options have two problems. First, they introduce client state: they must be set and reset around each query that needs them. This prevents load balancing and fail-over. (By contrast, Impala keeps options in the client and sends them to the server on each query, allowing unlimited load balancing via round-robin and allows transparent fail over.)

Further, session options are awkward for many tools: in addition to running a query, the client must send `ALTER SESSION` statements which are somehow associated with that query. Easy for our unit tests, not so easy for most BI tools.

Table options look like what we want, but they have a number of issues, including that, when used, they "erase" any options set in the format plugin.

Is there some other solution we could consider? Better table options? Some syntax that allows setting of options in the query? Some kind of view or schema that encapsulates these options so that they don't have to be provided on every query?

I don't have a solid suggestion; I'm simply pointing out the ease-of-use issue.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/writer/TextRecordWriter.java
```diff
@@ -165,10 +182,12 @@ public void writeField() throws IOException {
...
   public void cleanup() throws IOException {
     super.cleanup();
-    if (stream != null) {
-      stream.close();
-      stream = null;
-      logger.debug("closing file");
+
+    fRecordStarted = false;
+    if (writer != null) {
+      writer.close();
```
paul-rogers: This is a writer. The close can fail with an exception. If it does, the entire operation should fail. Should we throw a UserException` with the details here rather than let the `IOException` bubble up?`

exec/java-exec/src/test/java/org/apache/drill/test/ClusterFixture.java
```diff
@@ -57,10 +40,27 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Properties;
```
paul-rogers: Can we define a Drill standard for import order? The Apache settings I imported into Eclipse are `java`, `javax`, `org`, `com`. Statics are at the top.

In this change, looks like `java` moves the to bottom.

arina-ielchiieva: I use standard IntelliJ IDEA imports order:

image

Regarding updating the check styles, there is a Jira (https://issues.apache.org/jira/browse/DRILL-7352) where we can post comments.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/text/writer/TextRecordWriter.java
arina-ielchiieva: Good point, will update the code.

arina-ielchiieva:
@paul-rogers thanks for the code review, addressed code review comments, force-pushed since there were minor changes in the code. 

Regarding design, the aim of this Jira was just to fix text writer to write proper text files: before if column contained field separator, field was not enclosed in the quotes, thus we were writing text files which Drill could not read. Now when user indicates write format using session option (this is common approach for all formats), Drill produces text files, it can read back. Basically, if user has configured format plugin:
```
  "formats": {
    "csvh": {
      "type": "text",
      "extensions": [
        "csvh"
      ],
      "lineDelimiter": "\n",
      "fieldDelimiter": ",",
      "extractHeader": true
    }
   },
```
Drill will be able to read and write such text files correctly. Same approach is used for `parquet`, `json`. All user needs to do is to indicate write format using session option: `alter session set `store.format` = 'csvh';` (`parquet`, `json`). I am not saying this is ideal and we might need to reconsider such writing approach but I guess not in the scope of Jira since such re-design would touch all file writers.

paul-rogers:
To answer my other comment; about the lack of ease-of-use with the current session options related to file formats: the answer is the schema mechanism you defined a while back. Even if no schema is provided, we should allow the user to override formatting properties using the table options encoded in that schema.

Then, I can write a file using, say pipe delimiters, have that be recorded, and read the file automatically using those delimiters. Probably work to be done, but would we a nice solution that does not require HMS.

arina-ielchiieva:
Sounds interesting, basically, you are suggesting to create implicit schema file when writing a text files (i.e. creating table), thus format options will be picked up directly from schema file when reading table back.

PR #1874, opened 2019-10-18
DRILL-7405: Avoiding download of TPC-H data

Agirish:


Agirish:
Unit tests as well as Functional & Advanced Regression tests from [1] are successful.

[1] https://github.com/mapr/drill-test-framework

vvysotskyi:
I disagree that we should add so large files to the project sources.

I have experimented some time ago with alternative solutions for this problem and one of the ideas was to create a new GitHub repository with these files, use JitPack for publishing archive and use `maven-dependency-plugin` in Drill to obtain and unpack files when the project is built.

Here is a link to the repo with files: https://github.com/vvysotskyi/tpch-parquet and commit with changes in Drill: https://github.com/vvysotskyi/drill/commit/0635133bbd22945e7648791cfd6e2d146730b219.
What do you think about this? You may use these changes if we will choose this approach.

Agirish:
@vvysotskyi , I understand but I am not for using personal GitHub repositories. I don't think it's the best approach for an Apache project. The files are small enough ~3 MB. I'm not sure I agree that this would constitute as large. I think what we have here would be more straightforward.

sohami:
Looks like these files were packaged as jar in Drill class path as an example data for users to run some exploratory queries. I think putting these files as part of source repo should be fine.

@vvysotskyi : I think your main concern is related to the unit tests data files which are merged with the source files. I guess that was done to keep the test execution time lower otherwise ideally unit tests should use the in-memory data generator for it's use. May be we should come up with some policies which can dictate when is it fine to check in the test data file and when one should use in-memory data generator. 
Also how does moving data files to a separate git repo will help here ?

paul-rogers:
In general, it is a good idea to have all the files either in the project, or via a Maven dependency. We have another set of data, the "FoodMart" files, which we pull as a jar file from a repo maintained by Julian Hyde.

So, one possibility is to create another Drill repo that just handles supplementary files. In fact, already have one: the test framework. We could add another.

But, I agree with @sohami, the files are small and simple enough that we might as well add them to Drill directly.

paul-rogers:
How are these files used within Drill? If they are used for unit tests, then we have to consider the module hierarchy. If the files are in contrib, then only modules "later" than contrib can access them. Is this a problem?

arina-ielchiieva:
+1

PR #1876, opened 2019-10-20
DRILL-7412: Minor unit test improvements

paul-rogers:
Many tests intentionally trigger errors. A debug-only log setting
sent those errors to stdout. The resulting stack dumps simply cluttered
the test output, so disabled error output to the console.

Drill can apply bounds checks to vectors. Tests run via Maven
enable bounds checking. Now, bounds checking is also enabled in
"debug mode" (when assertions are enabled, as in an IDE.)

Drill contains two test frameworks. The older BaseTestQuery was
marked as deprecated, but many tests still use it and are unlikely
to be changed soon. So, removed the deprecated marker to reduce the
number of spurious warnings.

Also includes a number of minor clean-ups.

### DIFF
--- a/common/src/main/java/org/apache/drill/common/exceptions/DrillException.java
+++ b/common/src/main/java/org/apache/drill/common/exceptions/DrillException.java
@@ -17,10 +17,10 @@
  */
 package org.apache.drill.common.exceptions;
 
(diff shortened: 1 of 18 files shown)


arina-ielchiieva:
+1

PR #1878, opened 2019-10-20
DRILL-7414: EVF incorrectly sets buffer writer index after rollover

paul-rogers:
Enabling the vector validator on the "new" scan operator, in cases in which overflow occurs, identified that the DrillBuf writer index was not properly set for repeated vectors.
    
Enables such checking, adds unit tests, and fixes the writer index issue.

### DIFF
--- a/exec/java-exec/src/test/java/org/apache/drill/exec/physical/resultSet/impl/TestResultSetLoaderOverflow.java
+++ b/exec/java-exec/src/test/java/org/apache/drill/exec/physical/resultSet/impl/TestResultSetLoaderOverflow.java
@@ -28,10 +28,14 @@ import org.apache.drill.categories.RowSetTests;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.common.types.TypeProtos.DataMode;
 import org.apache.drill.common.types.TypeProtos.MinorType;
+import org.apache.drill.exec.physical.impl.validate.BatchValidator;
 import org.apache.drill.exec.physical.resultSet.ResultSetLoader;
 import org.apache.drill.exec.physical.resultSet.RowSetLoader;
 import org.apache.drill.exec.physical.resultSet.impl.ResultSetLoaderImpl.ResultSetOptions;
+import org.apache.drill.exec.physical.rowSet.RowSet;
+import org.apache.drill.exec.physical.rowSet.RowSetReader;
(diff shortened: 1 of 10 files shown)


paul-rogers:
Another in a series of operator fixes for issues uncovered by thorough auditing of each batch of vectors.

This PR builds n DRILL-7403, which is not yet merged. Once DRILL-7403 is merged, I will rebase this PR so that only the last commit is retained.

Please review just the last commit (for DRILL-7414); the others are covered by the DRILL-7403 PR.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/OffsetVectorWriterImpl.java
```diff
@@ -290,7 +290,7 @@ public void preRollover() {
     // rows. But, this being an offset vector, we add one to account
     // for the extra 0 value at the start.
 
-    setValueCount(vectorIndex.rowStartIndex() + 1);
+    setValueCount(vectorIndex.rowStartIndex());
```
arina-ielchiieva: ` But, this being an offset vector, we add one to account for the extra 0 value at the start.` - Should this comment be updated after the change?

exec/java-exec/src/test/java/org/apache/drill/exec/physical/resultSet/impl/TestResultSetLoaderOverflow.java
```diff
@@ -59,55 +61,64 @@
 
   @Test
   public void testVectorSizeLimit() {
-    TupleMetadata schema = new SchemaBuilder()
+    final TupleMetadata schema = new SchemaBuilder()
```
arina-ielchiieva: Since Java supports effective final there is no need to excessive use of final keyword unless s you want to explicitly indicate that variable is final. I am not going to request the change during code review, I guess this mostly relates to the developers code writing style but since you are making lots of refactoring adding final keyword, just wanted to highlight that this is might be unnecessary.

arina-ielchiieva:
@paul-rogers mostly looks good, one minor concern about if need to update the comment after code change.

arina-ielchiieva:
+1

arina-ielchiieva:
Cherry-picked last tow commits and squashed them, since there were no conflicts, changes were merged into master.

PR #1879, opened 2019-10-21
DRILL-5674: Support ZIP compression

arina-ielchiieva:
1. Added ZipCodec implementation which can read / write single file.
2. Revisited Drill plugin formats to ensure 'openPossiblyCompressedStream' method is used in those which support compression.
3. Added unit tests.
4. General refactoring.

Jira - [DRILL-5674](https://issues.apache.org/jira/browse/DRILL-5674).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
@@ -17,6 +17,16 @@
  */
 package org.apache.drill.exec.store.dfs;
 
+import org.apache.drill.common.exceptions.DrillRuntimeException;
+import org.apache.drill.exec.util.DrillFileSystemUtil;
+import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
+import org.apache.drill.shaded.guava.com.google.common.base.Stopwatch;
+import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
@@ -26,20 +36,12 @@ import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
-import org.apache.drill.shaded.guava.com.google.common.base.Preconditions;
-import org.apache.drill.shaded.guava.com.google.common.base.Stopwatch;
-import org.apache.drill.shaded.guava.com.google.common.collect.Lists;
-
-import org.apache.drill.common.exceptions.DrillRuntimeException;
-import org.apache.drill.exec.util.DrillFileSystemUtil;
-import org.apache.hadoop.fs.FileStatus;
-import org.apache.hadoop.fs.Path;
-
 /**
  * Jackson serializable description of a file selection.
  */
 public class FileSelection {
-  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FileSelection.class);
+
+  private static final Logger logger = LoggerFactory.getLogger(FileSelection.class);
   private static final String WILD_CARD = "*";
 
   private List<FileStatus> statuses;
@@ -73,10 +75,10 @@ public class FileSelection {
   }
 
   private StatusType dirStatus;
-  // whether this selection previously had a wildcard
-  private boolean hadWildcard = false;
-  // whether all partitions were previously pruned for this selection
-  private boolean wasAllPartitionsPruned = false;
+  // whether this selection previously had a wildcard, false by default
+  private boolean hadWildcard;
+  // whether all partitions were previously pruned for this selection, false by default
+  private boolean wasAllPartitionsPruned;
 
   /**
    * Creates a {@link FileSelection selection} out of given file statuses/files and selection root.
@@ -108,7 +110,7 @@ public class FileSelection {
    * Copy constructor for convenience.
    */
   protected FileSelection(FileSelection selection) {
-    Preconditions.checkNotNull(selection, "selection cannot be null");
+    Preconditions.checkNotNull(selection, "File selection cannot be null");
     this.statuses = selection.statuses;
     this.files = selection.files;
     this.selectionRoot = selection.selectionRoot;
@@ -117,6 +119,7 @@ public class FileSelection {
     this.metaContext = selection.metaContext;
     this.hadWildcard = selection.hadWildcard;
     this.wasAllPartitionsPruned = selection.wasAllPartitionsPruned;
+    this.emptyDirectory = selection.emptyDirectory;
   }
 
   public Path getSelectionRoot() {
@@ -127,8 +130,8 @@ public class FileSelection {
     Stopwatch timer = logger.isDebugEnabled() ? Stopwatch.createStarted() : null;
 
     if (statuses == null)  {
-      List<FileStatus> newStatuses = Lists.newArrayList();
-      for (Path pathStr : files) {
+      List<FileStatus> newStatuses = new ArrayList<>();
+      for (Path pathStr : Objects.requireNonNull(files, "Files can not be null if statuses are null")) {
         newStatuses.add(fs.getFileStatus(pathStr));
       }
       statuses = newStatuses;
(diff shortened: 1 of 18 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSelection.java
```diff
@@ -386,17 +387,16 @@ public static void checkBackPaths(String parent, String combinedPath, String sub
...
+        String.format("Invalid path [%s] takes you outside the workspace.", subpath));
     }
   }
 
   public List<FileStatus> getFileStatuses() {
     return statuses;
   }
 
-  public boolean supportDirPrunig() {
+  public boolean supportDirPruning() {
```
paul-rogers: Good catch. `suppportsDirPruning` (with an s)?

The `support` form is imperative, it tells this object to support dir pruning. The `supports` form asks if this object does or does not support dir pruning.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
```diff
@@ -57,7 +61,9 @@
  */
 public class FileSystemPlugin extends AbstractStoragePlugin {
 
-  private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(FileSystemPlugin.class);
+  private static final Logger logger = LoggerFactory.getLogger(FileSystemPlugin.class);
+
+  private static final List<String> BUILT_IN_CODECS = Collections.singletonList(ZipCodec.class.getCanonicalName());
```
paul-rogers: Are no other codecs provided "out of the box"? For others, I need to provide a jar and set a config option? Or, should we move the other built-in ones here and out of the config file?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FormatSelection.java
```diff
@@ -63,6 +60,6 @@ public FileSelection getSelection(){
 
   @JsonIgnore
   public boolean supportDirPruning() {
```
paul-rogers: As above `support` --> `supports`. Is safe because this value is not serialized.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/ZipCodec.java
```diff
@@ -0,0 +1,141 @@
...
+
+/**
+ * ZIP codec implementation which cna read or create single entry.
+ * <p/>
+ * Note: Do not rename this class. Class naming must be 'ZipCodec' so it can be mapped by
+ * {@link org.apache.hadoop.io.compress.CompressionCodecFactory} to the 'zip' extension.
+ */
+public class ZipCodec extends DefaultCodec {
+
+  private static final String EXTENSION = ".zip";
```
paul-rogers: Any need to support G-zip? (`.gz`) or Tar/g-zip (`.tar.gz`)?

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/ZipCodec.java
```diff
@@ -0,0 +1,141 @@
...
+      def.reset();
+    }
+  }
+
+  /**
+   * Writes given data into ZIP archive by placing all data in one entry with default naming.
+   */
+  private static class ZipCompressionOutputStream extends CompressionOutputStream {
+
+    private static final String DEFAULT_ENTRY_NAME = "entry.out";
```
paul-rogers: Should the entry name be the same as the file name so it is sensible if someone unzips the file?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/ParquetFormatPlugin.java
```diff
@@ -59,33 +51,44 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
```
paul-rogers: Maybe change your IDE import order to put java above org? That way, there won't be constant import shuffling each time your IDE touches a file. (Yes, we should decide on preferred order and document it somewhere...)

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcapng/PcapngFormatPlugin.java
```diff
@@ -47,7 +47,7 @@ public PcapngFormatPlugin(String name, DrillbitContext context, Configuration fs
 
   public PcapngFormatPlugin(String name, DrillbitContext context, Configuration fsConf, StoragePluginConfig config, PcapngFormatConfig formatPluginConfig) {
     super(name, context, fsConf, config, formatPluginConfig, true,
-        false, true, false,
+        false, true, true,
```
paul-rogers: Isn't the middle `true` wrong? It is for `blockSplittable`. That means we'll start reading at an arbitrary block boundary. Since this is a binary format, it is not clear that we can scan forward to the beginning of the next record as can be done in Sequence File and (restricted) CSV.

Also, if the file is zip-encoded, then it is never block splittable since Zip files cannot be read at an arbitrary offset.

This creates an issue: the block-splittable attribute right now is a constant. But, if any file is zip-encoded, then it is never block splittable. Any way to handle this fact?

And, any way to test this behaviour?

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcapng/package-info.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 /**
- * For comments on realization of this format plugin look at :
+ * For comments on implementation of this format plugin look at:
```
paul-rogers: "look at" --> "see"

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/ZipCodec.java
arina-ielchiieva: This stream is called by compression codec and we have no control to pass file name etc.
Anyway, Drill is not using output stream, only input stream, I have added implementation only for testing purposes.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/ZipCodec.java
arina-ielchiieva: `org.apache.hadoop.io.compress` supports gzip, bzip2 out of box. We are adding only zip codec implementation since it is missing in `org.apache.hadoop.io.compress` library.

Regarding `.tar.gz`, I am not sure how to support since it basically two compressions one over another and mostly is used for folders. Since Drill allows only to read compressed files not folders, I think for now we don't need it.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FormatSelection.java
arina-ielchiieva: Done.

exec/java-exec/src/main/java/org/apache/drill/exec/store/dfs/FileSystemPlugin.java
arina-ielchiieva: `org.apache.hadoop.io.compress` library supports gzip / bzip2 out of box. Here we only need to add codecs that are missing in this library. I have updated parameter name and added comment to avoid the confusion. `TestCompressedFiles` contains tests for all supported formats.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcapng/PcapngFormatPlugin.java
arina-ielchiieva: 1. Drill uses `BlockMapBuilder` to split file into blocks if possible. According to its code, it tries to split the file if `blockSplittable` is set to true and file IS NOT compressed. So even if format is block splittable but came as compressed file, it won't be split.

https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/store/schedule/BlockMapBuilder.java#L115

Looks like most of compressed formats are not splittable, that's why Drill considers any compressed file not splittable: https://i.stack.imgur.com/jpprr.jpg

2. Regarding blockSplittable for Pcang format, you are right such format is not splittable, as well as Pcap, I have updated the value of `blockSplittable` to `false` for both formats.

https://blog.marouni.fr/pcap2seq/

arina-ielchiieva:
@paul-rogers addressed code review comments.

paul-rogers:
Thanks for the answers to my questions. LGTM.
+1

PR #1886, opened 2019-10-29
DRILL-7273: Introduce operators for handling metadata

vvysotskyi:
Jira: [DRILL-7273](https://issues.apache.org/jira/browse/DRILL-7273)

This pull request introduces commands and operators for collecting table metadata and storing it to the metastore.

Entry point for ANALYZE command is `MetastoreAnalyzeTableHandler` class. It creates plan which includes some metastore-specific operators for collecting metadata.

New operators are the following:
`MetadataAggBatch` - operator which adds aggregate calls for all incoming table columns to calculate required metadata and produces aggregations. If aggregation is performed on top of another aggregation, required aggregate calls for merging metadata will be added.

`MetadataHandlerBatch` - operator responsible for handling metadata returned by incoming aggregate operators and fetching required metadata form the metastore to produce further aggregations.

`MetadataControllerBatch` - responsible for converting obtained metadata, fetching absent metadata from the metastore and storing resulting metadata into the metastore.

`MetastoreAnalyzeTableHandler` has 2 classes which depending on the table type, provides the information required for building a suitable plan for collecting metadata: `AnalyzeInfoProvider` and `MetadataInfoCollector`.

`MetastoreAnalyzeTableHandler` based on segments count, forms plan in the following form:

```
MetadataControllerBatch
	...
		MetadataHandlerBatch
			MetadataAggBatch
				MetadataHandlerBatch
					MetadataAggBatch
						Scan
```
The lowest `MetadataAggBatch` creates required aggregate calls for every (or interesting only) table columns and produces aggregations with grouping by segment columns that correspond to specific table level.
`MetadataHandlerBatch` above it populates batch with additional information about metadata type and other info.
`MetadataAggBatch` above merges metadata calculated before to obtain metadata for parent metadata levels and also stores incoming data to populate it to the metastore later.

`MetadataControllerBatch` obtains all calculated metadata, converts it to the suitable form and sends it to the metastore.

For the case of incremental analyze, `MetastoreAnalyzeTableHandler` creates `Scan` with updated files only and provides `MetadataHandlerBatch` with information about metadata which should be fetched from the metastore, so existing actual metadata wouldn't be recalculated.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
@@ -31,6 +31,7 @@ import org.apache.calcite.tools.RelConversionException;
 import org.apache.calcite.tools.ValidationException;
 import org.apache.drill.common.exceptions.UserException;
 import org.apache.drill.exec.ExecConstants;
+import org.apache.drill.exec.exception.MetadataException;
 import org.apache.drill.exec.ops.QueryContext;
 import org.apache.drill.exec.ops.QueryContext.SqlStatementType;
 import org.apache.drill.exec.physical.PhysicalPlan;
@@ -40,6 +41,7 @@ import org.apache.drill.exec.planner.sql.handlers.DefaultSqlHandler;
 import org.apache.drill.exec.planner.sql.handlers.DescribeSchemaHandler;
 import org.apache.drill.exec.planner.sql.handlers.DescribeTableHandler;
 import org.apache.drill.exec.planner.sql.handlers.ExplainHandler;
+import org.apache.drill.exec.planner.sql.handlers.MetastoreAnalyzeTableHandler;
 import org.apache.drill.exec.planner.sql.handlers.RefreshMetadataHandler;
 import org.apache.drill.exec.planner.sql.handlers.ResetOptionHandler;
 import org.apache.drill.exec.planner.sql.handlers.SchemaHandler;
@@ -49,13 +51,13 @@ import org.apache.drill.exec.planner.sql.parser.DrillSqlCall;
 import org.apache.drill.exec.planner.sql.parser.DrillSqlDescribeTable;
 import org.apache.drill.exec.planner.sql.parser.DrillSqlResetOption;
 import org.apache.drill.exec.planner.sql.parser.DrillSqlSetOption;
-import org.apache.drill.exec.planner.sql.parser.SqlCreateTable;
 import org.apache.drill.exec.planner.sql.parser.SqlSchema;
 import org.apache.drill.exec.testing.ControlsInjector;
 import org.apache.drill.exec.testing.ControlsInjectorFactory;
 import org.apache.drill.exec.util.Pointer;
 import org.apache.drill.exec.work.foreman.ForemanSetupException;
 import org.apache.drill.exec.work.foreman.SqlUnsupportedException;
+import org.apache.drill.shaded.guava.com.google.common.base.Throwables;
 import org.apache.hadoop.security.AccessControlException;
 
 public class DrillSqlWorker {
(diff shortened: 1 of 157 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeInfoProviderRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Returns {@link AnalyzeInfoProvider} instance suitable for specified {@link GroupScan}.
+   *
+   * @param groupScan group scan
+   * @return {@link AnalyzeInfoProvider} instance
+   */
+  public AnalyzeInfoProvider get(GroupScan groupScan) {
+    if (analyzeInfoProviders == null) {
```
arina-ielchiieva: Move this if block to the separate get method: `getAnalyzeInfoProviders()`

exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeInfoProviderRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+      }
+    }
+
+    List<AnalyzeInfoProvider> suitableProviders = analyzeInfoProviders.stream()
+        .filter(analyzeInfoProvider -> analyzeInfoProvider.supportsGroupScan(groupScan))
+        .collect(Collectors.toList());
+
+    if (suitableProviders.isEmpty()) {
+      throw new MetastoreException(String.format("Unable to find AnalyzeInfoProvider implementation for GroupScan class [%s]",
```
arina-ielchiieva: Metastore exception is only for Metastore operations and should not be thrown outside Metastore module...

exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeInfoProviderRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+    } catch (NoSuchMethodException | IllegalAccessException e) {
+      logger.error("AnalyzeInfoProvider implementation class [%s] must have constructor without arguments.");
+      return null;
+    }
+
+    Object instance;
+    try {
+      instance = constructor.invoke();
+    } catch (Throwable e) {
+      logger.error("Unable to init AnalyzeInfoProvider class [{}}]", className, e);
```
arina-ielchiieva: I think warn is more suitable and does not require stackstace. Can only have error message from throwable. Full stackstace can be added in debug level.

exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeInfoProviderRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+      return null;
+    }
+
+    if (!(instance instanceof AnalyzeInfoProvider)) {
+      logger.error("Created instance of [{}] does not implement [{}] interface",
```
arina-ielchiieva: Same here.

exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeInfoProviderTemplate.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package org.apache.drill.exec.metastore.analyze;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation used to mark all {@link AnalyzeInfoProvider} implementations which
```
arina-ielchiieva: Marks all ...

exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeInfoProviderTemplate.java
```diff
@@ -0,0 +1,32 @@
...
+ * should be used for ANALYZE for specific {@link org.apache.drill.exec.physical.base.GroupScan} implementation.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE})
+public @interface AnalyzeInfoProviderTemplate {
```
arina-ielchiieva: Why we call it template? I think `AnalyzeInfoProvider` is more suitable.

exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeInfoProviderRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+              "to prevent using arbitrary implementation.",
+          groupScan.getClass().getSimpleName(), AnalyzeInfoProviderTemplate.class.getName());
+    }
+
+    return suitableProviders.get(0);
+  }
+
+  private List<AnalyzeInfoProvider> init() {
+    List<AnalyzeInfoProvider> providers;
+    List<AnnotatedClassDescriptor> annotatedClasses = classpathScan.getAnnotatedClasses(AnalyzeInfoProviderTemplate.class.getName());
```
arina-ielchiieva: Why we annotate classes if we can simply look for classes which implement `AnalyzeInfoProvider` interface thus annotation can be removed.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
```diff
@@ -131,7 +130,9 @@ private static PhysicalPlan convertPlan(QueryContext context, String sql, Pointe
           "Will sync remote and local function registries if needed and retry " +
           "in case if issue was due to missing function implementation.", e);
       if (context.getFunctionRegistry().syncWithRemoteRegistry(
-          context.getDrillOperatorTable().getFunctionRegistryVersion())) {
+              context.getDrillOperatorTable().getFunctionRegistryVersion())
```
arina-ielchiieva: Please explain? What about CTAS for example?

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/MetastoreAnalyzeTableHandler.java
```diff
@@ -109,7 +108,8 @@ public PhysicalPlan getPlan(SqlNode sqlNode)
         config.getConverter().getDefaultSchema(), sqlAnalyzeTable.getSchemaPath());
     DrillTable table = getDrillTable(drillSchema, sqlAnalyzeTable.getName());
 
-    AnalyzeInfoProvider analyzeInfoProvider = AnalyzeInfoProvider.getAnalyzeInfoProvider(TableType.getTableType(table.getGroupScan()));
+    AnalyzeInfoProvider analyzeInfoProvider = context.getAnalyzeInfoProviderRegistry()
```
arina-ielchiieva: I think GroupScan shoulder return analyzeInfoProvider instance.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/DrillSqlWorker.java
vvysotskyi: Thanks, `CTAS` with `EXPLAIN` also should be present in this check. The goal was to avoid rerunning queries that cannot have UDFs but failed with an exception. For example, for Metastore ANALYZE, Metastore is being disabled using a query-level option before plan constructing but after the check that it was enabled. So after rerunning such a query, it will fail with another exception that Metastore is disabled.

arina-ielchiieva:

PR #1887, opened 2019-10-31
DRILL-7372: MethodAnalyzer consumes too much memory

vvysotskyi:
Jira: [DRILL-7372](https://issues.apache.org/jira/browse/DRILL-7372).

- Replaced `Set<Integer>` with `BitSet` - it allowed to reduce heap usage significantly.
- Avoided copying `localVariablesSet` and `labelsStack` from the source frame when a new frame is created since the branch-sensitive analysis was already done for a source frame.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/compile/bytecode/MethodAnalyzer.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/compile/bytecode/MethodAnalyzer.java
@@ -28,9 +28,8 @@ import org.objectweb.asm.tree.analysis.Interpreter;
 import org.objectweb.asm.tree.analysis.Value;
 
 import java.util.ArrayDeque;
+import java.util.BitSet;
 import java.util.Deque;
-import java.util.HashSet;
-import java.util.Set;
 
 /**
  * Analyzer that allows us to inject additional functionality into ASMs basic analysis.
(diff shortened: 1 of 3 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/compile/bytecode/MethodAnalyzer.java
```diff
@@ -174,14 +171,22 @@ public void initJumpTarget(int opcode, LabelNode target) {
...
-            localVariablesSet.push(new HashSet<>());
+            localVariablesSet.push(new BitSet());
             labelsStack.push(target);
         }
       }
     }
+
+    @Override
+    public void clearStack() {
+      super.clearStack();
```
arina-ielchiieva: Consider nullification or empty collections re-assignement

vvysotskyi: Thanks, done.

arina-ielchiieva:
+1, LGTM.

paul-rogers:
Just a thought: we proved a couple of years back that the modern JVM does a perfectly fine job of scalar replacement. We also found performance benefits from using "plain Java" without class merging. As a result, our byte code analysis is no longer needed. Just flip one option and we can completely remove this code instead of continuing to maintain it.

vvysotskyi:
@paul-rogers, unfortunately, I didn't see the results of these performance measurements, but I believe that there are some corner cases where our byte code analysis will bring some benefit, for example, when the generated method is too large for JVM optimizations and general time for query execution is much larger compared to the time required for producing scalar replacement and classes merging.

PR #1889, opened 2019-11-04
DRILL-7436: Fix record count, vector structure issues in several operators

paul-rogers:
Jira - [DRILL-7436](https://issues.apache.org/jira/browse/DRILL-7436).

Adds additional vector checks to the `BatchValidator`.

Enables checking for the following operators:

* `FilterRecordBatch`
* `PartitionLimitRecordBatch`
* `UnnestRecordBatch`
* `HashAggBatch`
* `RemovingRecordBatch`

Fixes vector count issues for each of these.

Fixes empty-batch (record count = 0) handling in several of the
above operators. Added a method to `VectorContainer` to correctly
create an empty batch. (An empty batch, counter-intuitively,
needs vectors allocated to hold the 0 value in the first
position of each offset vector.)

Disables verbose logging for MongoDB tests. Details are written to
the log rather than the console.

Disables two invalid Mongo tests. See DRILL-7428.

Adjusts the expression tree materializer to not add the LATE type
to Union vectors. (See DRILL-7435.)

Ensures that Union vectors contain valid vectors for each subtype.
The present fix is a work-around, see DRILL-7434 for a better
long-term fix.

Cleans up code formatting and other minor issues in each file touched
during the fixes in this PR.

### DIFF
--- a/contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestConstants.java
+++ b/contrib/storage-mongo/src/test/java/org/apache/drill/exec/store/mongo/MongoTestConstants.java
@@ -77,6 +77,9 @@ public interface MongoTestConstants {
   String TEST_BOOLEAN_FILTER_QUERY_TEMPLATE4 = "select `employee_id` from mongo.%s.`%s` where (position_id = 16 and isFTE = true) or last_name = 'Yonce'";
 
   String TEST_STAR_QUERY_UNSHARDED_DB = "select * from mongo.%s.`%s`";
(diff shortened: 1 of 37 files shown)


paul-rogers:
Sorry for the large number of files changed: the batch count issues are complex and affect many bits of code.

It may be hard to fully understand why certain things changed. This is why the `BathValidator` exists: it runs in every unit test to validate each batch, container and vector. You should assume that the changes all work toward getting each of the target operators to pass these checks.

Note that many operators are not yet tested. This PR was supposed to just be the filter operator. But, it turned out that other operators were creating bad structures for which Filter got the blame. So, the scope had to expand to fix those upstream problems as well.

Many of the changes, however, are code cleanup and should be quick to review.

arina-ielchiieva:
+1, LGTM. Thanks for making the changes.
Ran all tests for branch - all successful.

PR #540, opened 2016-07-05
Fix for DRILL-4759: Drill throwing array index out of bound exception…

ppadma:
… when reading a parquet file written by map reduce program

PR #1891, opened 2019-11-05
DRILL-7409: Moving test with big test data to the drill-test-framework.

denysord88:
Jira  - [DRILL-7409](https://issues.apache.org/jira/browse/DRILL-7409).


denysord88:
@vvysotskyi please review the PR.

cgivre:
@denysord88 Can you please rename with JIRA issue in title?

paul-rogers:
After this change, is this functionality still tested in the Drill unit tests? Or, is the only test of this particular functionality now in the test framework?

denysord88:
@paul-rogers it was the only test for BigInt data in dictionary encoding that I found in Drill unit tests. I'm moving it to the drill-test-framework - PR [601](https://github.com/mapr/drill-test-framework/pull/601)

progers-sumo:
@denysord88, thanks for the explanation. Developers often run the Drill unit tests to verify that a change works. We count on the unit tests to be complete. If some bit of functionality is not covered in unit tests, then it is possible to break that functionality in a PR, and to not learn of that error by running unit tests.

When a test exists only in the test framework, we must rely on a MapR employee to run the pre-commit tests for us. If the test fails, we are not set up to run the test to track down the problem.

The test framework works best when it acts as an end-to-end test at scale: across many data types, many query types and so on.

If this particular test is only for one specific bit of functionality, and uses a small (80MB) file, perhaps it should remain in unit tests?

The alternative is to package the test framework in a way that others can run it. Impala did a nice job of providing scripts to set up a "mini cluster" to allow running the E2E tests on a developer box (typically an Ubuntu system in AWS.)

Is there a way to retain the test in the unit tests, perhaps with a smaller file? One would think that to test Int dictionary encoding, one could do so with a file of, say, a dozen or a hundred rows, which should result in a small Parquet file.

denysord88:
@paul-rogers , this test was added to check a bug [DRILL-4759](https://issues.apache.org/jira/browse/DRILL-4759) in the Drill PR [540](https://github.com/apache/drill/pull/540). I tried to run this test on the Drill - several commits before this PR 540 and it wasn't failed. It's the reason why I'm removing this test from Drill unit tests.

arina-ielchiieva:
This definitely changes the situation. I think removing extra 1.83 MB would be nice.
@paul-rogers what do you think?

paul-rogers:
Seems fine, let's go ahead.

PR #1896, opened 2019-11-10
DRILL-7441: Fix issues with fillEmpties, offset vectors

paul-rogers:
Fixes subtle issues with offset vectors and "fill empties"
logic.

Drill has an informal standard that if a batch has no rows, then
offset vectors within that batch should have zero size. Contrast
this with batches of size 1 that should have offset vectors of
size 2. Changed to enforce this rule throughout.

Nullable, repeated and variable-width vectors have "fill empties"
logic that is used in two places: when setting the value count and
when preparing to write a new value. The current logic is not
quite right for either case. Added tests and fixed the code to
properly handle each case.

Revised the batch validator to enforce the offset-vector length of 0 for
0-sized batches rule. The result was much simpler code.

Added tools to easily print a batch, restoring some code that
was recently lost when the RowSet classes were moved.

Code cleanup in all files touched.

Added logic to "dirty" allocated buffers when testing to ensure
logic is not sensitive to the "pristine" state of new buffers.

Added logic to the column writers to enforce the zero-size-batch rule
for offset vectors. Added unit tests for this case.

Fixed the column writers to set the "lastSet" mutator value for
nullable types since other code relies on this value.

Removed the "setCount" field in nullable vectors: turns out
it is not actually used.


paul-rogers:
Rebased on master and resolved merge conflicts.

exec/vector/src/main/java/org/apache/drill/exec/vector/complex/AbstractRepeatedMapVector.java
```diff
@@ -430,29 +430,35 @@ public boolean isNull(int index) {
...
     @Override
     public void setValueCount(int topLevelValueCount) {
-      emptyPopulator.populate(topLevelValueCount);
-      offsets.getMutator().setValueCount(topLevelValueCount == 0 ? 0 : topLevelValueCount + 1);
-      int childValueCount = offsets.getAccessor().get(topLevelValueCount);
+      if (topLevelValueCount == 0) {
+        return;
+      }
+      int childValueCount;
+      if (topLevelValueCount == 0) {
```
vvysotskyi: Looks like this code wouldn't be triggered because of the return above.

exec/vector/src/main/codegen/templates/NullableValueVectors.java
```diff
@@ -844,29 +841,35 @@ public void generateTestData(int valueCount) {
...
     // For nullable vectors, exchanging buffers (done elsewhere)
     // requires also exchanging mutator state (done here.)
 
     @Override
     public void exchange(ValueVector.Mutator other) {
-      final Mutator target = (Mutator) other;
-      int temp = setCount;
-      setCount = target.setCount;
-      target.setCount = temp;
+      <#if type.major == "VarLen">
```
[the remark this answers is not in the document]
vvysotskyi: Thanks for the explanation!

exec/java-exec/src/test/java/org/apache/drill/test/BaseTestQuery.java
```diff
@@ -377,6 +379,14 @@ public static void resetSessionOption(String option) {
     }
   }
 
+  public static void resetAllSessionOptions() {
+    try {
+      test("ALTER SESSION RESET ALL");
+    } catch(final Exception e) {
```
vvysotskyi:
```suggestion
    } catch (Exception e) {
```

exec/java-exec/src/test/java/org/apache/drill/test/BaseTestQuery.java
```diff
@@ -377,6 +379,14 @@ public static void resetSessionOption(String option) {
...
+      fail("Failed to reset all session option");
```
vvysotskyi:
```suggestion
      fail("Failed to reset all session options");
```

paul-rogers:
@vvysotskyi, thanks much for the review. Addressed your comments; please take another look.

PR #1897, opened 2019-11-11
DRILL-7442: Create multi-batch row set reader

paul-rogers:
Adds a ResultSetReader that works across multiple batches
in a result set. Reuses the same row set and readers if
schema is unchanged, creates a new set if the schema changes.

Adds a unit test for the result set reader.

Adds a "rebind" capability to the row set readers to focus
on new buffers under an existing set of vectors. Used when
a new batch arrives, if the schema is unchanged.

Extends row set classses to be aware of the BatchAccessor class
which encapsulates a container and optional selection vector,
and tracks schema changes.

Moves row set tests into the same package as the row sets.
(Row set classes were moved a while back, but the tests were
not moved.)

Renames some BatchAccessor methods.

Jira - [DRILL-7442](https://issues.apache.org/jira/browse/DRILL-7442).


exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/AbstractTupleReader.java
```diff
@@ -87,6 +87,13 @@ public void bindNullState(NullStateReader nullStateReader) {
     this.nullStateReader = nullStateReader;
   }
 
+  @Override
+  public void rebind() {
+    for (int i = 0; i < readers.length; i++) {
+      readers[i].events().rebind();
+    }
```
vvysotskyi: It can be rewritten with a for-each loop:
```
    for (AbstractObjectReader reader : readers) {
      reader.events().rebind();
    }
```
There is no difference from the performance perspective, but it would be easier to support this code in the future, for example when replacing array with a list, etc.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/UnionReaderImpl.java
```diff
@@ -182,6 +182,15 @@ public void reposition() {
     }
   }
 
+  @Override
+  public void rebind() {
+    for (int i = 0; i < variants.length; i++) {
```
vvysotskyi: This loop also may be rewritten with a for-each loop, or even with a stream.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/ResultSetReader.java
```diff
@@ -0,0 +1,50 @@
...
+ *   <li>Call {@link #release()} to free the memory for the
+ *       incoming batch. Or, to call {@link #detach()} to keep
+ *       the batch memory.</li>
+ *   </ol>
+ * <li>Call {@link #close()} after all batches are read.</li>
+ * </ol>
+ */
+public interface ResultSetReader {
+
+  void start(BatchAccessor batch);
```
vvysotskyi: Could you please add JavaDocs to these methods. Though their purposes are described in the class JavaDoc, it would be handy also to have their own ones.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSetPrinter.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.drill.exec.record.RecordBatch;
+import org.apache.drill.exec.record.VectorContainer;
+import org.apache.drill.exec.record.metadata.ColumnMetadata;
+import org.apache.drill.exec.record.metadata.TupleMetadata;
+
+/**
+ * Print a row set in CSV-like format. Primarily for debugging.
+ */
+
+public class RowSetPrinter {
```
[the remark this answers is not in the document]
paul-rogers: I'd forgotten we made that change. Code is changed to use that class. Several bugs were fixed in the formatter (numbers were printing as control codes.)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/rowSet/RowSet.java
```diff
@@ -98,6 +98,8 @@
 
   SelectionVectorMode indirectionType();
 
+  void print();
```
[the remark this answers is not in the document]
paul-rogers: Actually, we agreed we would not call methods for printing to output in production code. `RowSet` is used often in testing where it is often handy to print the contents of row sets while debugging. Marked this method as `@VisibleForTesting`. While we could do:

```
System.out.println(rowSet.toString());
```

I am lazy, just easier to write:

```
rowSet.print();
```

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/AbstractTupleReader.java
paul-rogers: Generally I do use the iterator form. However, the readers and writers are inner-loop code: old-fashioned index looping is faster and avoids creating heap objects. The code here was specifically designed for index looping.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/UnionReaderImpl.java
paul-rogers: Also performance-sensitive code.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/reader/AbstractTupleReader.java
vvysotskyi: Actually, at least Java 8 rewrites for-each loop for array into the for-i loop, here is an example of the source code and decompiled code:
```
    String[] arr = {"a", "b", "c"};
    for (String x : arr) {
      System.out.println(x);
    }
```
code after decompilation:
```
    String[] var1 = new String[]{"a", "b", "c"};
    String[] var2 = var1;
    int var3 = var1.length;

    for(int var4 = 0; var4 < var3; ++var4) {
      String var5 = var2[var4];
      System.out.println(var5);
    }
```
But anyway, this is a minor comment, so the code may be left as it is.

paul-rogers:
@vvysotskyi, thanks much for the review and for committing the code. Thanks also for showing the disassembly of the generated code for the for loops. You analysis is convincing and I'll change the loops in the next PR.

PR #1899, opened 2019-11-15
DRILL-7445: Create batch copier based on result set framework

paul-rogers:
The result set framework now provides both a reader and writer.
This PR provides a copier that copies batches using this
framework. Such a copier can:

- Copy selected records
- Copy all records, such as for an SV2 or SV4

The copier uses the result set loader to create uniformly-sized
output batches from input batches of any size. It does this
by merging or splitting input batches as needed.

Since the result set reader handles both SV2 and SV4s, the
copier can filter or reorder rows based on the SV associated
with the input batch.

This version assumes single stream of input batches, and handles
any schema changes in that input by creating output batches
that track the input schema. This would be used in, say, the
selection vector remover. A different design is needed for merging
such as in the merging receiver.

Adds a "copy" method to the column writers. Copy is implemented
by doing a direct memory copy from source to destination vectors.

A unit test verifies functionality for various use cases
and data types.

### DIFF
--- a/common/src/main/java/org/apache/drill/common/types/Types.java
+++ b/common/src/main/java/org/apache/drill/common/types/Types.java
@@ -806,16 +806,20 @@ public class Types {
     return typeBuilder;
   }
 
+  public static boolean isSameType(MajorType type1, MajorType type2) {
+    return type1.getMinorType() == type2.getMinorType() &&
+           type1.getMode() == type2.getMode() &&
+           type1.getScale() == type2.getScale() &&
+           type1.getPrecision() == type2.getPrecision();
+  }
+
   public static boolean isEquivalent(MajorType type1, MajorType type2) {
 
     // Requires full type equality, including fields such as precision and scale.
     // But, unset fields are equivalent to 0. Can't use the protobuf-provided
     // isEquals() which treats set and unset fields as different.
 
-    if (type1.getMinorType() != type2.getMinorType() ||
-        type1.getMode() != type2.getMode() ||
-        type1.getScale() != type2.getScale() ||
-        type1.getPrecision() != type2.getPrecision()) {
+    if (! isSameType(type1, type2)) {
       return false;
     }
 
--- a/common/src/main/java/org/apache/drill/common/types/Types.java
+++ b/common/src/main/java/org/apache/drill/common/types/Types.java
@@ -806,6 +806,16 @@ public class Types {
     return typeBuilder;
   }
 
+  /**
+   * Check if two "core" types are the same, ignoring subtypes and
+   * children. Primarily for non-complex types.
+   *
+   * @param type1 first type
+   * @param type2 second type
+   * @return true if the two types are are the same minor type, mode,
+   * precision and scale
+   */
+
   public static boolean isSameType(MajorType type1, MajorType type2) {
     return type1.getMinorType() == type2.getMinorType() &&
            type1.getMode() == type2.getMode() &&
@@ -813,13 +823,15 @@ public class Types {
            type1.getPrecision() == type2.getPrecision();
   }
 
+  /**
+   * Requires full type equality, including fields such as precision and scale.
+   * But, unset fields are equivalent to 0. Can't use the protobuf-provided
+   * isEquals() which treats set and unset fields as different.
+   */
+
   public static boolean isEquivalent(MajorType type1, MajorType type2) {
 
-    // Requires full type equality, including fields such as precision and scale.
-    // But, unset fields are equivalent to 0. Can't use the protobuf-provided
-    // isEquals() which treats set and unset fields as different.
-
-    if (! isSameType(type1, type2)) {
+    if (!isSameType(type1, type2)) {
       return false;
     }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ResultSetCopierImpl.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ResultSetCopierImpl.java
@@ -1,0 +1,321 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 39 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ResultSetCopierImpl.java
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  @Override
+  public void copyRecord(int posn) {
+    verifyWritable();
+    rowReader.setPosition(posn);
+    project();
+  }
+
+  private final void project() {
```
ihuzenko:
```suggestion
  private void copyInputRow() {
```

exec/java-exec/src/main/java/org/apache/drill/exec/record/selection/SelectionVector2Builder.java
```diff
@@ -0,0 +1,47 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.record.selection;
+
+import org.apache.drill.exec.memory.BufferAllocator;
+import org.apache.drill.exec.record.VectorContainer;
+
+public class SelectionVector2Builder {
```
ihuzenko: Seems like the class is only intended for test purposes. If so, please move to test sources.

exec/vector/src/main/codegen/templates/ColumnAccessors.java
```diff
@@ -592,6 +592,32 @@ public final void setDefaultValue(final Object value) {
...
+    public final void copy(ColumnReader from) {
+      ${drillType}ColumnReader source = (${drillType}ColumnReader) from;
+      final DrillBuf sourceBuf = source.buffer();
+      <#-- First cut, copy materialized value.
+    <#if varWidth>
+      byte[] bytes = source.getBytes();
+      setBytes(bytes, bytes.length);
+    <#else>
+      set${label}(source.get${label}());
+    </#if> -->
```
ihuzenko: please remove commented lines

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/convert/AbstractWriteConverter.java
```diff
@@ -143,4 +144,9 @@ public void setTime(LocalTime value) {
   public void setTimestamp(Instant value) {
     baseWriter.setTimestamp(value);
   }
+
+  @Override
+  public void copy(ColumnReader from) {
+    throw new UnsupportedOperationException("Cannot copy values through a type converter");
```
ihuzenko: If type converters shouldn't support copy then maybe makes sense to mark the method as final here.

exec/vector/src/main/java/org/apache/drill/exec/vector/accessor/writer/UnionWriterImpl.java
```diff
@@ -332,6 +334,14 @@ public int writeIndex() {
     return index.vectorIndex();
   }
 
+  @Override
+  public void copy(ColumnReader from) {
+    if (! from.isNull()) {
```
ihuzenko:
```suggestion
    if (!from.isNull()) {
```

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ResultSetCopierImpl.java
```diff
@@ -0,0 +1,321 @@
...
+
+  public ResultSetCopierImpl(BufferAllocator allocator, BatchAccessor inputBatch) {
+    this(allocator, inputBatch, new OptionBuilder());
+  }
+
+  public ResultSetCopierImpl(BufferAllocator allocator, BatchAccessor inputBatch,
+      OptionBuilder outputOptions) {
+    this.allocator = allocator;
+    resultSetReader = new ResultSetReaderImpl(inputBatch);
+    writerOptions = outputOptions;
```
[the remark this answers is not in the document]
paul-rogers: Seems overly clever to save one line.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ResultSetCopierImpl.java
```diff
@@ -0,0 +1,321 @@
...
+    writerOptions.setVectorCache(new ResultVectorCacheImpl(allocator));
+    state = State.START;
```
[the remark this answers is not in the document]
paul-rogers: See above about use of `this.` prefix.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ResultSetCopierImpl.java
paul-rogers: Renamed to `copyColumns()` (which is what "project" means.) The `final` has meaning for a method: it means that there will never be overrides and so the JVM can optimize calls, which is helpful given that this is called once per rows.

exec/java-exec/src/main/java/org/apache/drill/exec/record/selection/SelectionVector2Builder.java
paul-rogers: That's where it is used in this PR. However, when doing the batch fixes, I found that we repeated the same logic (often incorrectly) in multiple places. The intent is to use this class in those cases as well. By putting it in this PR we can review it without also having to understand the operators that will eventually be changed.

paul-rogers:
@ihuzenko, thanks for your detailed review. Addressed most of your comments, added explanations for the others.

To provide a bit of context: this is a first step toward additional cleanup in the operators ("record batches") based on issues I've been finding as I work though the batch count, row count, vector value count issues. Did this change as its own PR to make it easier to review. A subsequent PR will use these changes to modify selected operators.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/resultSet/impl/ResultSetCopierImpl.java
ihuzenko: Thanks for explaining the reasons, but private methods already can't be overridden. Though this was actually minor suggestion.

common/src/main/java/org/apache/drill/common/types/Types.java
```diff
@@ -806,16 +806,20 @@ public static boolean isSortable(MinorType type) {
     return typeBuilder;
   }
 
+  public static boolean isSameType(MajorType type1, MajorType type2) {
```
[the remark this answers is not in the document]
ihuzenko: Yes I got the idea, and this suggestion is part of my personal beliefs. I like to hide everything until it's required to be used elsewhere. Since it's minor comment, I'm ok with leaving it as is.

PR #1901, opened 2019-11-18
DRILL-7388: Kafka improvements

arina-ielchiieva:
Jira - [DRILL-7388](https://issues.apache.org/jira/browse/DRILL-7388).

1. Upgraded Kafka libraries to 2.3.1 (DRILL-6739).
2. Added new options to support the same features as native JSON reader:
  a. store.kafka.reader.skip_invalid_records, default: false (DRILL-6723);
  b. store.kafka.reader.allow_nan_inf, default: true;
  c. store.kafka.reader.allow_escape_any_char, default: false.
3. Fixed issue when Kafka topic contains only one message (DRILL-7388).
4. Replaced Gson parser with Jackson to parse JSON in the same manner as Drill native Json reader.
5. Performance improvements: Kafka consumers will be closed async, fixed issue with resource leak (DRILL-7290), moved to debug unnecessary info logging.
6. Updated bootstrap-storage-plugins.json to reflect actual Kafka connection properties.
7. Added unit tests.
8. Refactoring and code clean up.


contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/KafkaPartitionScanSpecBuilder.java
```diff
@@ -31,30 +31,30 @@
...
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 
-public class KafkaPartitionScanSpecBuilder extends
-    AbstractExprVisitor<List<KafkaPartitionScanSpec>,Void,RuntimeException> {
-  static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(KafkaPartitionScanSpecBuilder.class);
+public class KafkaPartitionScanSpecBuilder
+  extends AbstractExprVisitor<List<KafkaPartitionScanSpec>,Void,RuntimeException>
```
vvysotskyi:
```suggestion
  extends AbstractExprVisitor<List<KafkaPartitionScanSpec>, Void, RuntimeException>
```

contrib/storage-kafka/src/main/java/org/apache/drill/exec/store/kafka/ReadOptions.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public ReadOptions(OptionManager optionManager) {
+    this.messageReader = optionManager.getString(ExecConstants.KAFKA_RECORD_READER);
+    this.pollTimeOut = optionManager.getLong(ExecConstants.KAFKA_POLL_TIMEOUT);
+    this.allTextMode = optionManager.getBoolean(ExecConstants.KAFKA_ALL_TEXT_MODE);
+    this.readNumbersAsDouble = optionManager.getBoolean(ExecConstants.KAFKA_READER_READ_NUMBERS_AS_DOUBLE);
+    this.enableUnionType = optionManager.getBoolean(ExecConstants.ENABLE_UNION_TYPE_KEY);
+    this.skipInvalidRecords = optionManager.getBoolean(ExecConstants.KAFKA_READER_SKIP_INVALID_RECORDS);
+    this.allowNanInf = optionManager.getBoolean(ExecConstants.KAFKA_READER_NAN_INF_NUMBERS);
+    this.allowEscapeAnyChar = optionManager.getBoolean(ExecConstants.KAFKA_READER_ESCAPE_ANY_CHAR);
```
paul-rogers: Nicely done; this is a clean, simple way to hide option complexity from the rest of the code.

arina-ielchiieva:
@vvysotskyi / @paul-rogers thanks for the code review. Addressed code review comments.

@paul-rogers regarding moving Kafka to EVF, does it depend on Json format conversion to EVF or `org.apache.drill.exec.vector.complex.fn.JsonReader` won't be changed only code which is using it? Also looks like we have a couple of examples of format plugins on EVF, do we have EVF based plugins? Only Mock plugin?

paul-rogers:
@arina-ielchiieva, we do now have multiple plugins using EVF: text (which I did) and several which Charles has done including Pcap, HDF5 and regex.

You are correct, since the Kafka reader uses the JSON reader, it makes sense to get that merged as part of the JSON format reader. Once that is stable, we can apply it here (and in Charles' HTTP storage plugin.)

The JSON format plugin will get merged once I finish resolving the batch/container record count issues. Those were causing unexpected test failures when I enabled the new JSON reader.

So, nothing yet to do on Kafka for EVF; but your changes will make the eventual upgrade easier.

PR #1902, opened 2019-11-18
DRILL-7448: Fix warnings when running Drill memory tests

KazydubB:
Jira - [DRILL-7448](https://issues.apache.org/jira/browse/DRILL-7448)

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -102,14 +102,6 @@
         <version>2.9</version>
     </dependency>
 
-    <!-- Conditional processing in logback configuration files requires the Janino library
-         (see logback-test.xml, DRILL-7397, http://logback.qos.ch/setup.html#janino) -->
-    <dependency>
-      <groupId>org.codehaus.janino</groupId>
-      <artifactId>janino</artifactId>
-      <scope>test</scope>
-    </dependency>
-
   </dependencies>
 
   <build>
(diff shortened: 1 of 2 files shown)


vvysotskyi:
@KazydubB, thanks for fixing it. Could you please add also Janino dependency into other places where `drill-common:tests` is used to avoid possible issues for unit tests in other modules? Also, please add a comment above added dependencies with the info why it should be present.

KazydubB:
@paul-rogers, @vvysotskyi I've updated the PR with @vvysotskyi 's suggestion. Thank you!

arina-ielchiieva:
@KazydubB thanks for making the changes. Discussed this issue with @vvysotskyi, it looks like adding janino dependency to all modules (and then control that it is added in new ones), might be slightly an overhead. I understand that we need to do this to support conditional processing which is mainly used to enable Lilith. On the other hand, we can remove condition, and leave Lilith code commented. Those who need Lilith would have to uncomment this code, instead of setting the variable. I suspect Lilith usage is not that popular in project and there won't be many complaint about it.

KazydubB:
@arina-ielchiieva, yes, this is definetely better. Thank you!
@arina-ielchiieva, @vvysotskyi, @paul-rogers I've yet another change (according to Arina's last comment), please take a look again.

arina-ielchiieva:
+1, LGTM.

PR #1910, opened 2019-11-28
DRILL-7393: Revisit Drill tests to ensure that patching is executed b…

agozhiy:
…efore any test run

- Added BaseTest with patchers and extended all tests from it.
- Added a test to java-exec module to ensure that all tests there are inherited from BaseTest.
- Revised exception handling in the patchers, now it's individual for each patching method.

JIRA: [DRILL-7393](https://issues.apache.org/jira/browse/DRILL-7393)

### DIFF
--- a/common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
+++ b/common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
@@ -24,154 +24,157 @@ import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
-import javassist.CannotCompileException;
 import javassist.ClassPool;
 import javassist.CtClass;
 import javassist.CtConstructor;
 import javassist.CtMethod;
 import javassist.CtNewMethod;
-import javassist.NotFoundException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class GuavaPatcher {
   private static final Logger logger = LoggerFactory.getLogger(GuavaPatcher.class);
 
-  private static boolean patched;
+  private static boolean patchingAttempted;
 
   public static synchronized void patch() {
-    if (!patched) {
-      try {
-        patchStopwatch();
-        patchCloseables();
-        patchPreconditions();
-        patched = true;
-      } catch (Throwable e) {
-        logger.warn("Unable to patch Guava classes.", e);
-      }
+    if (!patchingAttempted) {
+      patchingAttempted = true;
+      patchStopwatch();
+      patchCloseables();
+      patchPreconditions();
     }
   }
 
   /**
    * Makes Guava stopwatch look like the old version for compatibility with hbase-server (for test purposes).
    */
-  private static void patchStopwatch() throws Exception {
+  private static void patchStopwatch() {
+    try {
+      ClassPool cp = ClassPool.getDefault();
+      CtClass cc = cp.get("com.google.common.base.Stopwatch");
 
-    ClassPool cp = ClassPool.getDefault();
-    CtClass cc = cp.get("com.google.common.base.Stopwatch");
-
-    // Expose the constructor for Stopwatch for old libraries who use the pattern new Stopwatch().start().
-    for (CtConstructor c : cc.getConstructors()) {
-      if (!Modifier.isStatic(c.getModifiers())) {
-        c.setModifiers(Modifier.PUBLIC);
+      // Expose the constructor for Stopwatch for old libraries who use the pattern new Stopwatch().start().
+      for (CtConstructor c : cc.getConstructors()) {
+        if (!Modifier.isStatic(c.getModifiers())) {
+          c.setModifiers(Modifier.PUBLIC);
+        }
       }
+
+      // Add back the Stopwatch.elapsedMillis() method for old consumers.
+      CtMethod newMethod = CtNewMethod.make(
+          "public long elapsedMillis() { return elapsed(java.util.concurrent.TimeUnit.MILLISECONDS); }", cc);
+      cc.addMethod(newMethod);
+
+      // Load the modified class instead of the original.
+      cc.toClass();
+
+      logger.info("Google's Stopwatch patched for old HBase Guava version.");
+    } catch (Exception e) {
+      logger.warn("Unable to patch Guava classes.", e);
     }
-
-    // Add back the Stopwatch.elapsedMillis() method for old consumers.
-    CtMethod newMethod = CtNewMethod.make(
-        "public long elapsedMillis() { return elapsed(java.util.concurrent.TimeUnit.MILLISECONDS); }", cc);
-    cc.addMethod(newMethod);
-
-    // Load the modified class instead of the original.
-    cc.toClass();
-
-    logger.info("Google's Stopwatch patched for old HBase Guava version.");
   }
 
-  private static void patchCloseables() throws Exception {
+  private static void patchCloseables() {
+    try {
+      ClassPool cp = ClassPool.getDefault();
+      CtClass cc = cp.get("com.google.common.io.Closeables");
 
-    ClassPool cp = ClassPool.getDefault();
-    CtClass cc = cp.get("com.google.common.io.Closeables");
+      // Add back the Closeables.closeQuietly() method for old consumers.
+      CtMethod newMethod = CtNewMethod.make(
+          "public static void closeQuietly(java.io.Closeable closeable) { try{closeable.close();}catch(Exception e){} }",
+          cc);
+      cc.addMethod(newMethod);
 
+      // Load the modified class instead of the original.
+      cc.toClass();
 
-    // Add back the Closeables.closeQuietly() method for old consumers.
-    CtMethod newMethod = CtNewMethod.make(
-        "public static void closeQuietly(java.io.Closeable closeable) { try{closeable.close();}catch(Exception e){} }",
-        cc);
-    cc.addMethod(newMethod);
-
-    // Load the modified class instead of the original.
-    cc.toClass();
-
-    logger.info("Google's Closeables patched for old HBase Guava version.");
+      logger.info("Google's Closeables patched for old HBase Guava version.");
+    } catch (Exception e) {
+      logger.warn("Unable to patch Guava classes.", e);
+    }
   }
 
   /**
    * Patches Guava Preconditions with missing methods, added for the Apache Iceberg.
    */
-  private static void patchPreconditions() throws NotFoundException, CannotCompileException {
-    ClassPool cp = ClassPool.getDefault();
-    CtClass cc = cp.get("com.google.common.base.Preconditions");
+  private static void patchPreconditions() {
+    try {
+      ClassPool cp = ClassPool.getDefault();
+      CtClass cc = cp.get("com.google.common.base.Preconditions");
 
-    // Javassist does not support varargs, generate methods with varying number of arguments
-    int startIndex = 1;
-    int endIndex = 5;
+      // Javassist does not support varargs, generate methods with varying number of arguments
+      int startIndex = 1;
+      int endIndex = 5;
 
-    List<String> methodsWithVarargsTemplates = Arrays.asList(
-      "public static void checkArgument(boolean expression, String errorMessageTemplate, %s) {\n"
-        + "    if (!expression) {\n"
-        + "      throw new IllegalArgumentException(format(errorMessageTemplate, new Object[] { %s }));\n"
-        + "    }\n"
-        + "  }",
+      List<String> methodsWithVarargsTemplates = Arrays.asList(
+          "public static void checkArgument(boolean expression, String errorMessageTemplate, %s) {\n"
+              + "    if (!expression) {\n"
+              + "      throw new IllegalArgumentException(format(errorMessageTemplate, new Object[] { %s }));\n"
+              + "    }\n"
+              + "  }",
 
-      "public static Object checkNotNull(Object reference, String errorMessageTemplate, %s) {\n"
-        + "    if (reference == null) {\n"
-        + "      throw new NullPointerException(format(errorMessageTemplate, new Object[] { %s }));\n"
-        + "    } else {\n"
-        + "      return reference;\n"
-        + "    }\n"
-        + "  }",
+          "public static Object checkNotNull(Object reference, String errorMessageTemplate, %s) {\n"
+              + "    if (reference == null) {\n"
+              + "      throw new NullPointerException(format(errorMessageTemplate, new Object[] { %s }));\n"
+              + "    } else {\n"
+              + "      return reference;\n"
+              + "    }\n"
+              + "  }",
 
-      "public static void checkState(boolean expression, String errorMessageTemplate, %s) {\n"
-        + "    if (!expression) {\n"
-        + "      throw new IllegalStateException(format(errorMessageTemplate, new Object[] { %s }));\n"
-        + "    }\n"
-        + "  }"
-    );
+          "public static void checkState(boolean expression, String errorMessageTemplate, %s) {\n"
+              + "    if (!expression) {\n"
+              + "      throw new IllegalStateException(format(errorMessageTemplate, new Object[] { %s }));\n"
+              + "    }\n"
+              + "  }"
+      );
 
-    List<String> methodsWithPrimitives = Arrays.asList(
-      "public static void checkArgument(boolean expression, String errorMessageTemplate, int arg1) {\n"
-        + "    if (!expression) {\n"
-        + "      throw new IllegalArgumentException(format(errorMessageTemplate, new Object[] { new Integer(arg1) }));\n"
-        + "    }\n"
-        + "  }",
-      "public static Object checkNotNull(Object reference, String errorMessageTemplate, int arg1) {\n"
-        + "    if (reference == null) {\n"
-        + "      throw new NullPointerException(format(errorMessageTemplate, new Object[] { new Integer(arg1) }));\n"
-        + "    } else {\n"
-        + "      return reference;\n"
-        + "    }\n"
-        + "  }"
-    );
+      List<String> methodsWithPrimitives = Arrays.asList(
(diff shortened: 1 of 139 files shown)


exec/java-exec/src/test/java/org/apache/drill/BaseTestInheritance.java
```diff
@@ -0,0 +1,63 @@
...
+    Reflections reflections = new Reflections("org.apache.drill", new SubTypesScanner(false));
+    Set<String> baseTestInheritors = reflections.getSubTypesOf(BaseTest.class).stream()
+        .map(Class::getName)
+        .collect(Collectors.toSet());
+    // Get all tests
+    Set<String> testClasses = reflections.getSubTypesOf(Object.class).stream()
+        .filter(c -> !c.isInterface())
+        .filter(c -> c.getSimpleName().toLowerCase().contains("test"))
+        .filter(c -> {
+          for (Method m : c.getDeclaredMethods()) {
```
vvysotskyi: Please use find any here.

exec/java-exec/src/test/java/org/apache/drill/BaseTestInheritance.java
```diff
@@ -0,0 +1,63 @@
...
+            if (m.getAnnotation(Test.class) != null) {
+              return true;
+            }
+          }
+          return false;
+        })
+        .map(Class::getName)
+        .collect(Collectors.toSet());
+
+    testClasses.removeAll(baseTestInheritors);
```
vvysotskyi: We may avoid this by adding one more filter into the stream above.

exec/java-exec/src/test/java/org/apache/drill/BaseTestInheritance.java
```diff
@@ -0,0 +1,63 @@
...
+    String illegitimateTests = String.join("\n", testClasses);
+    assertTrue(String.format("The following test classes are not inherited from BaseTest:\n%s", illegitimateTests),
```
vvysotskyi: We may use assert equals and specify empty collection, so it wouldn't be required to append the list into the error string.

agozhiy:
@vvysotskyi, thank you for review, I made the changes you requested.

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
```diff
@@ -24,154 +24,157 @@
...
 import javassist.CtNewMethod;
-import javassist.NotFoundException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class GuavaPatcher {
   private static final Logger logger = LoggerFactory.getLogger(GuavaPatcher.class);
 
-  private static boolean patched;
+  private static boolean patchingAttempted;
```
paul-rogers: Maven runs tests concurrently in the same JVM. Does this flag need synchronization?

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
```diff
@@ -24,154 +24,157 @@
...
-      try {
-        patchStopwatch();
-        patchCloseables();
-        patchPreconditions();
-        patched = true;
-      } catch (Throwable e) {
-        logger.warn("Unable to patch Guava classes.", e);
-      }
+    if (!patchingAttempted) {
+      patchingAttempted = true;
```
paul-rogers: This version no longer has a try/catch block. Is the idea that, if patching fails, the test should fail? If this particular test fails, we've marked patching as attempted and we won't try again. This means that future tests that needed the patching will mysteriously fail.

Should we, instead, simply do a `System.exit(1)` if patching fails to shut down the tests? Or, print a warning to stderr that subsequent tests may fail?

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
```diff
@@ -24,154 +24,157 @@
...
 
-    // Load the modified class instead of the original.
-    cc.toClass();
+      // Load the modified class instead of the original.
+      cc.toClass();
 
-    logger.info("Google's Stopwatch patched for old HBase Guava version.");
+      logger.info("Google's Stopwatch patched for old HBase Guava version.");
+    } catch (Exception e) {
+      logger.warn("Unable to patch Guava classes.", e);
```
paul-rogers: We are warning on errors, but then proceeding. The test that caused this patch will still run, but presumably will fail. Is this the desired behaviour?

paul-rogers:
The meta-question is why we still need this after all these years? Shouldn't we try to normalize our usage of Guava? Also, we moved Drill's use into a shaded jar so that, presumably, it won't conflict with other usages. So, do we still need this patching hack?

common/src/test/java/org/apache/drill/test/BaseTest.java
```diff
@@ -15,24 +15,25 @@
...
-  }
-
-  @Test
-  public void pruneSimple() throws Exception {
-    test(String.format("select * from dfs.`%s/csv` where dir1 in ('Q1', 'Q2')", MULTILEVEL));
+  static {
+    /*
+     * HBase and MapR-DB clients use older version of protobuf,
+     * and override some methods that became final in recent versions.
+     * This code removes these final modifiers.
```
paul-rogers: Shouldn't these clients be addressed directly? We're doing patching for tests, but what about production code? Possible solutions:

* Load these clients into their own class loader so their dependencies do not conflict.
* Shade them.
* Upgrade to newer versions that use more recent Guava libraries.
* Do the patching within Drill: perhaps have a plugin-specific startup event that can trigger patching where needed.
* Lobby the creators of the libraries to fix the problems.

My concern is that changing all our tests to work around bugs in external libraries is an unstable solution: we'll find ourselves doing more of this each time some new external conflict arises.

common/src/test/java/org/apache/drill/test/BaseTest.java
```diff
@@ -15,24 +15,25 @@
...
+package org.apache.drill.test;
 
-import org.apache.drill.BaseTestQuery;
-import org.apache.drill.common.util.TestTools;
-import org.junit.Test;
+import org.apache.drill.common.util.GuavaPatcher;
+import org.apache.drill.common.util.ProtobufPatcher;
 
-public class TestPrune extends BaseTestQuery {
+public class BaseTest {
```
paul-rogers: The term `BaseTest` is too generic: soon this will contain all manner of common code. Better solutions are:

* MapRTestBase and HBaseTestBase to patch the code as needed for just those clients.
* Patchers as a service added to tests where needed, such as is done with the "dirTestWatcher"
* The patch-on-load approach suggested below.

The approach here is really far too much of a hack to maintain.

Also, this patching means that code running in tests operates in a different environment than the same code running in production. That will become the cause of very mysterious bugs.

Can we rethink our approach?

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
agozhiy: It is used in synchronized method so it is synchronized.

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
agozhiy: The try/catch block was moved to the component methods. It is simpler this way to pinpoint possible errors.

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
agozhiy: I didn't change the behavior. 
What do you suggest? Tests are executed by JUnit, so in case of an exception they will fail just as well, but with a less clear message (because BaseTest would fail to initialize).

common/src/test/java/org/apache/drill/test/BaseTest.java
agozhiy: We're doing patching both in tests and in production (see Drillbit class). 
This is actually quite extensive topic. We already have shaded Guava that is used by Drill. The problem is: some of libraries we're using have transitive dependencies to different Guava versions. In total we have 3 or 4 versions we should make friends with. So shading is not the way. I guess we can do some playing with classloaders, but it'll be tricky and additionally we'll have bloated jars.
As for your last paragraph, this would definitely be great if every library had the same dependencies versions or if Guava was fully backward compatible... Sadly, there is nothing we can do on this account.

common/src/test/java/org/apache/drill/test/BaseTest.java
agozhiy: The original problem was caused by tests that didn't require patching, they just loaded a Guava class needed to be patched for someone else, an there is no way we can patch a class that was already loaded. Considering this we should apply patching as early as possible.

agozhiy:
> The meta-question is why we still need this after all these years? Shouldn't we try to normalize our usage of Guava? Also, we moved Drill's use into a shaded jar so that, presumably, it won't conflict with other usages. So, do we still need this patching hack?

As I answered earlier, we need to deal with several Guava versions at once. The on used by Drill is shaded but unfortunately this is not enough.
Also this fix do not change the approach, it just bring the tests in line with it.

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
vvysotskyi: Also, as was mentioned in another comment, patching is done for production code, so it should be synchronized.

common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
vvysotskyi: For the case when patching is used when starting Drill, it would be better to avoid killing a drillbit if it will fail. In this case, part of the functionality would still be available.
Regarding unit tests use case, there was introduced a test that checks that patching is produced for all test classes.

common/src/test/java/org/apache/drill/test/BaseTest.java
vvysotskyi: Good ideas, but I have some questions:
1. I'm not sure how to tell maven to keep a specific version to use it in the future but do not use it for other libraries.
2. When I was shading Guava for Drill usage, I considered this approach, but we have a lot of libraries with their specific versions, for example, hbase, hadoop-common, hadoop-client, hadoop-hdfs. It would be redundant to shade them all.
3. Unfortunately, even the latest versions of most libraries use obsolete guava versions (I have checked hbase).

common/src/test/java/org/apache/drill/test/BaseTest.java
vvysotskyi: Patching is done for production code also, so with patching, we make testing env closer to the production one. Also, pathing is required for a newer hadoop libraries version which is used in exec module, so comment in `BaseTest` perhaps should be updated a little bit.

arina-ielchiieva:
Looks like changing previous approach if **it is possible** might be rather hard and might take long  time. Since this Jira is just aligns tests with existing approach and is a release broker (is needed to merge Hadoop 3 upgrade), I suggest the following:
@agozhiy - to address general code review comments (renaming etc) not connected with approach.
@vvysotskyi - to merge the changes and start the release.

@paul-rogers / @vvysotskyi is this acceptable?

vvysotskyi:
@arina-ielchiieva, yes, it is acceptable for me.

paul-rogers:
@agozhiy, thank you for the explanation. To summarize, when we run a Drillbit, patching occurs. This means that any test that starts a Drillbit should *not* need to patch. In fact, we probably *do not want* to patch: we need the tests to verify that the Drillbit does any needed patching on its own.

We have tests that do not start a Drillbit; these tend to test specific bits of the system. Such tests should not need patching unless they pull in one of the affected dependencies. In this case, it makes sense for the test to explicitly declare that it needs patching so we can verify that the dependency works as needed.

From this, we get the following suggestion:

* Any test that derives from `BaseTestQuery` or `ClusterTest` will start a Drillbit. So the test *should not* attempt to patch as doing so will hide any Drillbit patching errors.
* Any other test should optionally invoke patching if needed for the specific item under test. (For example, if we have HBase, MapR or other tests that do unit tests without a Drillbit.) In this case, part of the test is to verify that the patching works for that dependency.

A result of this thinking is that we *should not* create a new base test class that always does patching. 

I wonder, how does the result above differ from what we already do?

vvysotskyi:
@paul-rogers, your suggestion sounds reasonable to me, but one more thing should be mentioned - is that we use guava in test classes also, so there are possible cases when class is loaded before Drill is patched, and therefore patching is failed. I can observe such scenarios at least for `Preconditions` class.

Also, before the changes in this PR, patching was produced in `DrillTest` static block (one more common parent for a lot of test classes).

paul-rogers:
@vvysotskyi, thanks for the additional info. Doing the patching in a static block makes sense; that ensures the patching is done early (though it still depends on when the class is loaded.)

Putting the patching in a test base class solves the ordering issue: the static block will be invoked before any other code in the test. However, the problem with a base class is we'd need a common base class for all tests, as was done in this PR. This means that we then mask the Drillbit patching, which, as explained earlier, we don't want to do.

I wonder, can we define an interface to solve the problem? Something like this:

```
public interface RequiresGuavaPatch {
  GuavaPatcher PATCHER = new GuavaPatcher();
}

public class MyTest implements RequiresGuavaPatch {
 ...
}
``` 

Assuming that creating the patcher does the patches. Else, we can create a specialized "shim" class that does patching in its constructor.

This approach hides the details of patching, yet declares that the test does require patching. Only non-Drillbit tests need worry about this interface. And, only tests that have affected dependencies need implement the interface.

An alternative is to be more specific, `HBaseTest`, `MapRTest` and so on that do whatever patching is needed just for that dependency. As these dependencies upgrade their own dependencies, we can change the code behind, say, `MapRTest` to remove patching no longer needed.

This way, we track patching at the dependency level, rather than as a single global (and thus hard-to-understand) task.

Let me clarify that point a bit. We need to patch due to specific transitive dependencies. As those are removed, we should remove patches. If we know that such-and-so patching is needed for each dependency, we can track when that patching is no longer needed. If patching is just done globally, then we have no good way to know why we do each patch, and when it is no longer needed.

Would something like this solve the problem?

agozhiy:
@paul-rogers, the approach you suggested is reasonable and actually it is similar to the one used in Drill now. But it doesn't solve the original issue. Let me describe it:

**Prerequisites:**
- In java-exec module we have both tests that require patching (or use Drillbit) and standalone tests that don't need patching but might use some Guava classes (e.g. Closeables)

**Steps:**
1. Run tests by command:
`mvn clean test -pl exec/java-exec -DforkCount=4`
Changing forkCount is needed to increase probability of race conditions.
2. Wait for failures.

**Expected result:**
Tests are passed.

**Actual result:**
Some tests will fail because patching didn't happen.

So, let's see what is happening here:
As we cannot guarantee the tests execution order, it is quite possible that a test that doesn't init patching will run first. It is also possible that this test is using some Guava class inside (and it'll be loaded during the test execution). Once the class is loaded it cannot be patched (Javassist constructs a modified class, attempts to load it and fail with error "attempted  duplicate class definition for name"). Then, all patching-dependent tests will fail. I hope my explanation is clear.

P.S.: Thank you for the feedback!

paul-rogers:
@agozhiy, thank you, the explanation is crystal clear. In fact, I think I've been seeing a similar issue with the ZK and Dynamic UDF tests: they randomly fail about 75% of the time due to issues associated with loading/finding classes that appear to be generated on the fly.

The test solution might solve the above race conditions if all tests invoke the same patcher as their first task. We would have to assume that the Drillbit does also. But, even then, we *still* would not solve the problem. SqlLine runs, then loads any number of JDBC drivers (in the general case) followed by the Drill JDBC driver which can, if requested, start a Drillbit. We have little control over when patching is done in this case (unless we add patching to SqlLine startup.)

Generalizing, since any app can load the Drill JDBC driver and run Drill, *every* app would need the magic patcher run at startup. This seems like a solution that won't scale.

We're talking about patching at runtime. I wonder, can we patch at build time? Can we create a Drill-specific version of Guava with the patches applied, stored in a jar file?

This still does not help with third-party apps: they might put their own Guava on the class path before the patched Drill version.

Did some Googling for solutions since we can't be the only folks with the problem. You've probably done the same. Here is a [note from the Guava team](https://github.com/google/guava/wiki/UseGuavaInYourBuild) which recommends using Guava version 21.

> In the simple case, just upgrade your project to the newest version of Guava. ... However, this can fail in some cases:
>
> *One of your dependencies uses a version of Guava from the days when we were still removing non-@Beta APIs.
>
> In these cases, you may be able to find an older version of Guava that satisfies all your dependencies. However, the only general solution is for all projects to update to at least Guava 21...

There seems to be a [Java library compliance checker tool](https://github.com/lvc/japi-compliance-checker) that might help identify actual gaps.

It also seems that that the only real solution that folks mention is [shading](https://bryantsai.com/how-to-resolve-dependency-conflict-out-of-your-control-e75ace79e54f). Can we shade the MapR plugin along with its dependent obsolete Guava version?

Otherwise, the only reliable solution is to get the set of dependencies to use compatible Guava versions. Can we upgrade our Hive/Hadoop dependencies to newer versions that depend on recent Guava versions? 

Look at it from the point of view of MapR and Hadoop: they cannot expect their clients to be on an ancient version of Guava; to support their users (including Drill), they have to stay current. Perhaps a bit of lobbying, searching will solve the problem without patching?

vvysotskyi:
@paul-rogers, all these approaches require a large amount of time. What do you think about @arina-ielchiieva's suggestion from this comment https://github.com/apache/drill/pull/1910#issuecomment-560383657 to unblock the release?

Discussing the approach on how to fix the general problem and implementing it may be done after the release.

paul-rogers:
@vvysotskyi, you got me on that one. If this is a release blocker, then, yes, we have to live with the best fix we can do right now.

vvysotskyi:
@paul-rogers, thank, I have merged this PR.

@agozhiy could you please create a Jira where this discussion may be continued?

agozhiy:
@paul-rogers, @vvysotskyi, I created a JIRA:
https://jira.apache.org/jira/browse/DRILL-7465

PR #1911, opened 2019-11-28
DRILL-7208: Reuse root git.properties file

vvysotskyi:
- Generate git.properties for root module only and copy it to child modules instead of generating the same files again

Jira - [DRILL-7208](https://issues.apache.org/jira/browse/DRILL-7208)


agozhiy:
@vvysotskyi, Could there be a race condition possible between maven-resources-plugin and git-commit-id-plugin?

vvysotskyi:
@agozhiy, good question! Plugins declared for the same phase, will be executed in the order of their appearance in the pom file, but currently, it may cause problems since `git-commit-id-plugin` is applied after `maven-resources-plugin`, so outdated file may be placed into the target/classes folder. Thanks for pointing that, I have corrected plugins order.

PR #1912, opened 2019-11-30
DRILL-7324: Final set of "batch count" fixes

paul-rogers:
Final set of fixes for batch count/record count issues. Enables
vector checking for all operators (except statistics, which do not
produce valid Drill batches.)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/statistics/StatisticsMergeBatch.java
```diff
@@ -82,17 +83,27 @@
  *       "sales_city" : BIGINT - nonnullstatcount(sales_city)
  *       "cnt"        : BIGINT - nonnullstatcount(cnt)
  *   .... another map for next stats function ....
+ * </pre>
+ * <p>
+ * Note that the above schema is not a valid Drill record batch: the varous
```
arina-ielchiieva:
```suggestion
 * Note that the above schema is not a valid Drill record batch: the various
```

arina-ielchiieva: @vvysotskyi could you please take a look at this java doc comment and changes in this class?

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JSONRecordReader.java
```diff
@@ -226,7 +226,8 @@ public int next() {
           }
           ++parseErrorCount;
           if (printSkippedMalformedJSONRecordLineNumber) {
-            logger.debug("Error parsing JSON in " + hadoopPath.getName() + " : line nos :" + (recordCount + parseErrorCount));
+            logger.debug("Error parsing JSON in {}: line nos: {}",
```
arina-ielchiieva: What `nos` means?

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JSONRecordReader.java
```diff
@@ -256,6 +257,7 @@ private void updateRunningCount() {
   public void close() throws Exception {
     if(stream != null) {
```
arina-ielchiieva:
```suggestion
    if (stream != null) {
```

exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestJsonReader.java
```diff
@@ -81,6 +81,8 @@ public void testEmptyList() throws Exception {
 
   @Test
   public void schemaChange() throws Exception {
+    // TODO: What is this testing? Presumably, we are simply
```
arina-ielchiieva: I assume yes, it just tests that test pass, do we need this comment here though?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/statistics/StatisticsMergeBatch.java
vvysotskyi: If I understand these changes from DRILL-1328 correctly, as a result, we should obtain a map with a single value and its children also should have single values. Looks like this is a bug with specifying wrong row count (perhaps in `buildOutgoingRecordBatch()` method).

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JSONRecordReader.java
paul-rogers: I suspect author means "numbers", but of course there is only one. Changed to just "line:".

exec/java-exec/src/test/java/org/apache/drill/exec/vector/complex/writer/TestJsonReader.java
paul-rogers: Revised the comment.

arina-ielchiieva:
+1, LGTM

PR #1913, opened 2019-11-30
DRILL-6953: EVF-based version of the JSON reader

paul-rogers:
Reimplements the JSON reader on top of the EVF. Does not yet
handle a provided schema. New JSON parser does not yet reflect
any changes made to the "V1" JSON parser in the last year.
Does not yet handle the Union and List-of-union types.
Enabling those encountered many issues elsewhere in Drill.

Provides more robust (but still limited) handling of JSON
type ambigutities. Handles runs of nulls before the first
non-null value (within the first batch.) Handles runs of
empty arrays before the first non-empty array (again, within
the first batch.) Handles the case where a null value turns out
to be an object or array. Handles reasonable conversions between
types.

Handling ambiguities makes the new parser more complex than
the "V1" version. The new one uses explict states for each
kind of JSON object, where as the old one used implicit states
expressed via if-statements, which can be a bit hard to follow
as the states get more complex.

The new "V2" JSON scan is controlled by a new option:
store.json.enable_v2_reader, which is false by default in this
PR.

Adds a "projection type" to the column writer so that the
JSON parser can receive a "hint" as to the expected type.
The hint is from the form of the projected column: `a[0]`,
`a.b` or just `a`.

Reimplements a number of JSON tests to test both the original
"V1" and the new "V2" versions of the JSON reader. Adds many
new tests for the new features of the "V2" parser.

Jira - [DRILL-6953](https://issues.apache.org/jira/browse/DRILL-6953).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JsonBatchReader.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JsonBatchReader.java
@@ -1,0 +1,149 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.store.easy.json;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.List;
+
+import org.apache.drill.common.exceptions.UserException;
+import org.apache.drill.common.types.TypeProtos.MajorType;
+import org.apache.drill.exec.ExecConstants;
+import org.apache.drill.exec.ops.OperatorContext;
+import org.apache.drill.exec.physical.impl.scan.file.FileScanFramework.FileSchemaNegotiator;
+import org.apache.drill.exec.physical.impl.scan.framework.ManagedReader;
+import org.apache.drill.exec.physical.resultSet.ResultSetLoader;
+import org.apache.drill.exec.physical.resultSet.ResultVectorCache;
+import org.apache.drill.exec.physical.resultSet.RowSetLoader;
+import org.apache.drill.exec.server.options.OptionSet;
+import org.apache.drill.exec.store.dfs.DrillFileSystem;
+import org.apache.drill.exec.store.easy.json.parser.JsonLoaderImpl;
+import org.apache.drill.exec.store.easy.json.parser.JsonLoaderImpl.JsonOptions;
+import org.apache.drill.exec.store.easy.json.parser.JsonLoaderImpl.TypeNegotiator;
+import org.apache.hadoop.mapred.FileSplit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JsonBatchReader implements ManagedReader<FileSchemaNegotiator> {
+
+  private static final Logger logger = LoggerFactory.getLogger(JsonBatchReader.class);
+
+  private DrillFileSystem fileSystem;
+  private FileSplit split;
+  private final JsonOptions options;
+  private JsonLoader jsonLoader;
+  private InputStream stream;
+
+  private RowSetLoader tableLoader;
+
+  public JsonBatchReader(JsonOptions options) {
+   this.options = options == null ? new JsonOptions() : options;
+  }
+
+  @Override
+  public boolean open(FileSchemaNegotiator negotiator) {
+    this.fileSystem = negotiator.fileSystem();
+    this.split = negotiator.split();
+    OperatorContext opContext = negotiator.context();
+    OptionSet optionMgr = opContext.getFragmentContext().getOptions();
+    Object embeddedContent = null;
+    options.allTextMode = embeddedContent == null && optionMgr.getBoolean(ExecConstants.JSON_ALL_TEXT_MODE);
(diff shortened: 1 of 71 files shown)


paul-rogers:
Some background: this PR includes work completed about two years ago as part of the "row set" (EVF) project. We had to first get the EVF itself reviewed and merged, then we added provided schema support. The first attempt to merge the JSON reader uncovered many issues with batch, record and vector counts. Those have been fixed over the last couple of months. This time, the unit tests pass with the new JSON reader.

This PR leaves the old "V1" reader enabled by default. More testing is required before we enable the "V2" reader by default.

Because this work pre-dated the "provided schema" work, it does not yet support the provided schema. Let's get this version merged, then we can add the additional work needed to support a provided schema.

Also, any work done in the "V1" JSON reader in the last two years is not yet reflected in the "V2" version. We make any such changes after this PR.

JSON is a surprisingly complex and tricky format. Suggestions for further tests or improvements are welcome.

arina-ielchiieva:
@vvysotskyi bested the review, we need to run tests on test cluster with enabled new JSON reader to see if there are any regressions.

exec/java-exec/src/main/java/org/apache/drill/exec/store/easy/json/JsonBatchReader.java
```diff
@@ -0,0 +1,149 @@
...
+        for (int i = 0; i < path.size() - 1; i++) {
+          cache = cache.childCache(path.get(i));
+        }
+        return cache.getType(path.get(path.size()-1));
+      }
+    };
+
+    // Create the JSON loader (high-level parser).
+
+    jsonLoader = new JsonLoaderImpl(stream, rootWriter, options);
```
cgivre: @paul-rogers 
Would it be possible for you to add some logic here so that the JSONBatchReader can accept an `InputStream` as well as file input?  This could be useful for other storage plugins that output JSON.

vvysotskyi:
@paul-rogers, looks like `TestComplexTypeReader` tests fail with enabled v2 reader. Failing tests check obtaining nested array value. There are some other unit tests failures, but most of them are connected with the features which weren't reflected in v2 reader. I think we should fix most of them since in most cases they require just passing additional configs into Jackson.
Also, there are issues with duplicated values in JSON, for example, the following query returns the incorrect result:
```
select rownum, kvgen(varcharcol) from `json_kvgenflatten/kvgen2.json`
```
returns
```
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| rownum<BIGINT(OPTIONAL)>                                                                            | EXPR$1<MAP(REPEATED)>{$offsets$<UINT4(REQUIRED)>,key<VARCHAR(OPTIONAL)>,value<VARCHAR(OPTIONAL)>}   |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| 1                                                                                                   | [{"key":"varchar_1","value":"abc"},{"key":"varchar_2","value":"def"},{"key":"key","value":"value"}] |
| 2                                                                                                   | [{"key":"varchar_2","value":""}]                                                                    |
| 3                                                                                                   | [{"key":"varchar_2","value":"hellohow"}]                                                            |
| 2                                                                                                   | []                                                                                                  |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

```
Data file content:
```
{
  "rownum": 1,
  "varcharcol": {
    "varchar_1": "abc",
    "varchar_2": "def",
    "key": "value"
  }
}
{
  "rownum": 2,
  "varcharcol": {
    "varchar_1": null,
    "varchar_2": ""
  }
}
{
  "rownum": 3,
  "varcharcol": {
    "varchar_1": null,
    "varchar_2": "",
    "varchar_2": null,
    "varchar_2": "hello",
    "varchar_2": "how"
  }
}
{
  "rownum": 2,
  "varcharcol": {
  }
}
```
Previous reader just uses the last value.

paul-rogers:
@vvysotskyi, thanks for running the tests with V2 enabled. My original thought was to get the basics in with V2 disabled by default in order to reduce the work required. But, since you went ahead and did the full testing, I agree we should include those fixes in the PR as well.

paul-rogers:
Turns out that the v1 JSON reader will ignore all-null columns. In the test `TestStarQueries.testSelStarOrderBy`, we execute the following query:

```
select * from cp.`employee.json` order by last_name
```

Against the Foodmart file `customer.json` which has records like this:

```
{"employee_id":53,...","end_date":null,"salary":...}
```

The test is written to exclude `end_date` from a `SELECT *` query.

In V2, however, `end_date` is included in the query. One can argue that 'SELECT *' means "return all columns", not "return all columns except those that happen to be null in the first batch."

See DRILL-7522.

Modified the test to expect the old result for the V1 reader, the correct result for the V2. reader.

paul-rogers:
@vvysotskyi, @arina-ielchiieva - it turns out that a number of fixes are required to resolve the issues that came up in the tests run above. The full set of changes, with JSON and all the fixes, will be large. We should break them into smaller PRs.

There are two ways to proceed. One is to commit this PR with the new reader disabled by default. Then apply fixes one by one for easier to review. Eventually, we enable the new reader when everything works.

The other is that I make the fixes on this branch, then pull each out into a separate PR that we review before this one. The result is much extra work for me to do all the code copying/merging involved.

How do you recommend we proceed?

cgivre:
I would vote for option one.  IMHO we should work for smaller PRs so that we can get code committed faster and hopefully less work for the reviewers.

vvysotskyi:
I also don't mind making additional fixes separately if the full set of changes will be too large.
Could you please create Jira tickets for these additional fixes to be sure that we wouldn't skip something?
In this case, is this PR ready for the review?

paul-rogers:
@vvysotskyi, thanks for the note. If we're OK with the multi-step plan, I'll go ahead and run a full test myself (with the reader turned off) to make sure we're still good after rebasing and the small fixes includes since. I'll post here once that is done.

Good suggestion on the JIRA tickets. Since each PR needs a ticket, we'd need to fine them for that purpose anyway. You've probably seen the discussion on the dev list about some nasty ambiguities related to the way DICT interacts with some of our problem-child types. I filed a big, long ticket for that one. I'll file additional tickets for the other issues as I reproduce them.

paul-rogers:
This PR has grown too large. It will be replaced by a series of smaller, more focused PRs. Will leave this open for now to allow use of this PR as a reference for the new, smaller PRs.

paul-rogers:
New PRs are posted for some of the reworked way to submit this code. So, closing this one.

PR #1914, opened 2019-12-02
DRILL-7458: Base framework for storage plugins

paul-rogers:
Drill has an increasing number of storage plugins.
It turns out, many have the same boilerplate code
copy and pasted into each new plugin. This patch
provides a generic framework to allow easier creation
of storage plugins by abstracting out the
boilerplate.

Many storage plugins provide filter push-down,
again by copying a large amount of boilerplate
code from earlier plugins. This patch provides a
generic filter push-down framework that abstracts out
the common Drill/Calcite work, allowing the plugin
to implement just the functionality unique to that
plugin.

A "dummy" plugin shows how to use both frameworks and
acts as a "test mule" for the filter push-down framework.

Jira - [DRILL-7458](https://issues.apache.org/jira/browse/DRILL-7458).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
@@ -251,26 +251,6 @@ public enum PlannerPhase {
     this.description = description;
   }
 
-  /**
-   * Filter push-down is best done during logical planning so that the result can
-   * influence parallelization in the physical phase. The specific phase differs
-   * depending on which planning mode is enabled. This check hides those details
-   * from storage plugins that simply want to know "should I add my filter
-   * push-down rules in the given phase?"
-   *
-   * @return true if filter push-down rules should be applied in this phase
-   */
-  public boolean isFilterPushDownPhase() {
-    switch (this) {
-    case LOGICAL_PRUNE_AND_JOIN: // HEP is disabled
-    case PARTITION_PRUNING:      // HEP partition push-down enabled
-    case LOGICAL_PRUNE:          // HEP partition push-down disabled
-      return true;
-    default:
-      return false;
-    }
-  }
-
   public abstract RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins);
 
   private static RuleSet getStorageRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins,
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
@@ -105,6 +105,7 @@ import java.util.Set;
 public enum PlannerPhase {
 
   LOGICAL_PRUNE_AND_JOIN("Logical Planning (with join and partition pruning)") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(
           getDrillBasicRules(context),
@@ -116,6 +117,7 @@ public enum PlannerPhase {
   },
 
   WINDOW_REWRITE("Window Function rewrites") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return RuleSets.ofList(
           RuleInstance.CALC_INSTANCE,
@@ -125,6 +127,7 @@ public enum PlannerPhase {
   },
 
   SUBQUERY_REWRITE("Sub-queries rewrites") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return RuleSets.ofList(
           RuleInstance.SUB_QUERY_FILTER_REMOVE_RULE,
@@ -135,6 +138,7 @@ public enum PlannerPhase {
   },
 
   LOGICAL_PRUNE("Logical Planning (with partition pruning)") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(
           getDrillBasicRules(context),
@@ -145,6 +149,7 @@ public enum PlannerPhase {
   },
 
   JOIN_PLANNING("LOPT Join Planning") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       List<RelOptRule> rules = Lists.newArrayList();
       if (context.getPlannerSettings().isJoinOptimizationEnabled()) {
@@ -160,6 +165,7 @@ public enum PlannerPhase {
   },
 
   ROWKEYJOIN_CONVERSION("Convert Join to RowKeyJoin") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       List<RelOptRule> rules = Lists.newArrayList();
       if (context.getPlannerSettings().isRowKeyJoinConversionEnabled()) {
@@ -173,6 +179,7 @@ public enum PlannerPhase {
   },
 
   SUM_CONVERSION("Convert SUM to $SUM0") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(
           RuleSets.ofList(
@@ -184,24 +191,28 @@ public enum PlannerPhase {
   },
 
   PARTITION_PRUNING("Partition Prune Planning") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(getPruneScanRules(context), getStorageRules(context, plugins, this));
     }
   },
 
   PHYSICAL_PARTITION_PRUNING("Physical Partition Prune Planning") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(getPhysicalPruneScanRules(context), getStorageRules(context, plugins, this));
     }
   },
 
   DIRECTORY_PRUNING("Directory Prune Planning") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(getDirPruneScanRules(context), getStorageRules(context, plugins, this));
     }
   },
 
   LOGICAL("Logical Planning (no pruning or join).") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(
           PlannerPhase.getDrillBasicRules(context),
@@ -211,6 +222,7 @@ public enum PlannerPhase {
   },
 
   PHYSICAL("Physical Planning") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.mergedRuleSets(
           PlannerPhase.getPhysicalRules(context),
@@ -220,12 +232,14 @@ public enum PlannerPhase {
   },
 
   PRE_LOGICAL_PLANNING("Planning with Hep planner only for rules, which are failed for Volcano planner") {
+    @Override
     public RuleSet getRules (OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return PlannerPhase.getSetOpTransposeRules();
     }
   },
 
   TRANSITIVE_CLOSURE("Transitive closure") {
+    @Override
     public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins) {
       return getJoinTransitiveClosureRules();
     }
@@ -237,21 +251,40 @@ public enum PlannerPhase {
     this.description = description;
   }
 
+  /**
+   * Filter push-down is best done during logical planning so that the result can
+   * influence parallelization in the physical phase. The specific phase differs
+   * depending on which planning mode is enabled. This check hides those details
+   * from storage plugins that simply want to know "should I add my filter
+   * push-down rules in the given phase?"
+   *
+   * @return true if filter push-down rules should be applied in this phase
+   */
+  public boolean isFilterPushDownPhase() {
+    switch (this) {
+    case LOGICAL_PRUNE_AND_JOIN: // HEP is disabled
+    case PARTITION_PRUNING:      // HEP partition push-down enabled
+    case LOGICAL_PRUNE:          // HEP partition push-down disabled
+      return true;
+    default:
+      return false;
+    }
+  }
+
   public abstract RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins);
 
   private static RuleSet getStorageRules(OptimizerRulesContext context, Collection<StoragePlugin> plugins,
       PlannerPhase phase) {
     final Builder<RelOptRule> rules = ImmutableSet.builder();
-    for(StoragePlugin plugin : plugins){
-      if(plugin instanceof AbstractStoragePlugin){
+    for (StoragePlugin plugin : plugins) {
+      if (plugin instanceof AbstractStoragePlugin) {
         rules.addAll(((AbstractStoragePlugin) plugin).getOptimizerRules(context, phase));
-      }else{
+      } else {
         rules.addAll(plugin.getOptimizerRules(context));
       }
     }
     return RuleSets.ofList(rules.build());
   }
-
 
   static final RelOptRule DRILL_JOIN_TO_MULTIJOIN_RULE =
       new JoinToMultiJoinRule(DrillJoinRel.class, DrillRelFactories.LOGICAL_BUILDER);
@@ -412,9 +445,7 @@ public enum PlannerPhase {
 
     return RuleSets.ofList(pruneRules);
   }
-  /**
-   *
-   */
+
   static RuleSet getIndexRules(OptimizerRulesContext optimizerRulesContext) {
     final PlannerSettings ps = optimizerRulesContext.getPlannerSettings();
     if (!ps.isIndexPlanningEnabled()) {
@@ -480,7 +511,6 @@ public enum PlannerPhase {
         .build();
 
     return RuleSets.ofList(pruneRules);
-
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseGroupScan.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseGroupScan.java
@@ -19,10 +19,8 @@ package org.apache.drill.exec.store.base;
 
 import java.util.List;
 
-import org.apache.drill.common.JSONOptions;
 import org.apache.drill.common.expression.SchemaPath;
 import org.apache.drill.common.logical.StoragePluginConfig;
-import org.apache.drill.exec.metastore.MetadataProviderManager;
 import org.apache.drill.exec.physical.base.AbstractGroupScan;
 import org.apache.drill.exec.physical.base.GroupScan;
 import org.apache.drill.exec.physical.base.PhysicalOperator;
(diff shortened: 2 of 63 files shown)


arina-ielchiieva:
@paul-rogers is there any documentation on how it works?

paul-rogers:
@arina-ielchiieva, excellent point; a reusable framework is more useful if some explanation exists for use. There are two sources, both preliminary. First, there are comments in the code: `package-info.java` files and Javadoc. Second, there is an early draft of a guide on [this wiki](https://github.com/paul-rogers/drill/wiki/Create-a-Storage-Plugin). At present, the pages describe starting from scratch, but that got so complex that I ended up creating this framework instead. The pages need to be updated based on this framework. Any feedback would be greatly appreciated!

paul-rogers:
@arina-ielchiieva, updated some of the comments to make things a bit clearer. Also, did an extensive tutorial [here](https://github.com/paul-rogers/drill/wiki/Create-a-Storage-Plugin).

As it turns out, the "base" stuff is pretty basic; it is just a thin layer on what is there. (Still struggling with how to abstract out some of the remaining complexity.) The key new bit is the filter push-down strategy. The best way to understand that is to look at the unit test cases.

Thanks much for taking another look.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/PlannerPhase.java
```diff
@@ -238,22 +251,41 @@ public RuleSet getRules(OptimizerRulesContext context, Collection<StoragePlugin>
...
+  /**
+   * Filter push-down is best done during logical planning so that the result can
+   * influence parallelization in the physical phase. The specific phase differs
+   * depending on which planning mode is enabled. This check hides those details
+   * from storage plugins that simply want to know "should I add my filter
+   * push-down rules in the given phase?"
+   *
+   * @return true if filter push-down rules should be applied in this phase
+   */
+  public boolean isFilterPushDownPhase() {
```
vvysotskyi: I disagree with limiting the possibility of pushdown to several phases. In some cases, there may be additional rel nodes between filter and scan, so pushdown cannot happen, but in other planning stages, these rel nodes may be switched with filter, or filter may be extracted from join condition, etc. so pushdown may happen only at a specific phase, but not at the phase we want.

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseSubScan.java
```diff
@@ -0,0 +1,101 @@
...
+import com.fasterxml.jackson.annotation.JsonTypeName;
+
+/**
+ * Specification for a specific execution-time scan. Holds the information
+ * passed form the Planner to the Drillbit so that the Drillbit can execute
+ * the scan. Holds the config of the storage plugin associated with the
+ * scan. The config allows creating a storage plugin instance, as needed,
+ * to implement the scan.
+ *
+ * @see {@link BaseGroupScan} for additional details of the scan
```
vvysotskyi: A class name may be used without @link tag in @see:
```suggestion
 * @see BaseGroupScan for additional details of the scan
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseSubScan.java
```diff
@@ -0,0 +1,101 @@
...
+
+@JsonTypeName("base-sub-scan")
+public class BaseSubScan extends AbstractSubScan {
+
+  protected final BaseStoragePlugin<?> storagePlugin;
+  protected final List<SchemaPath> columns;
+
+  public BaseSubScan(BaseGroupScan groupScan) {
+    super(groupScan.getUserName());
+    storagePlugin = groupScan.storagePlugin();
```
vvysotskyi:
```suggestion
    this.storagePlugin = groupScan.storagePlugin();
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseGroupScan.java
```diff
@@ -0,0 +1,481 @@
...
+   */
+  @Override
+  public PhysicalOperator getNewWithChildren(List<PhysicalOperator> children) {
+    Preconditions.checkArgument(children.isEmpty());
+    return pluginOptions().scanFactory.copyGroupShim(this);
+  }
+
+  /**
+   * Create a string representation of the scan formatted for the
+   * <tt>EXPLAIN PLAN FOR</tt> output. Do not override this method,
```
vvysotskyi: This method may be done a final to prohibit overriding.

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/filter/FilterPushDownStrategy.java
```diff
@@ -0,0 +1,416 @@
...
+
+  private boolean isTargetScan(DrillScanRel scan) {
+    return listener.isTargetScan(scan.getGroupScan());
+  }
+
+  public void onMatch(RelOptRuleCall call, DrillFilterRel filter, DrillProjectRel project, DrillScanRel scan) {
+
+    // Skip if rule has already been applied.
+
+    if (!listener.needsApplication(scan.getGroupScan())) {
```
vvysotskyi: What about the case when this rule was applied for one filter, but planner at some point pushed another filter above the scan, for example, if we have such case:
```
Filter(a=2)
  Join(t1.b=t2.b, type=inner)
    Filter(b=3)
    Scan(t1)
    Scan(t2)
```
Filter b=3 will be pushed into scan, planner will push filter above join:
```
Join(t1.b=t2.b, type=inner)
    Filter(a=2)
    Scan(t1, b=3)
    Scan(t2)
```
In this case, check whether filter was pushed is not enough.

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseStoragePlugin.java
```diff
@@ -0,0 +1,229 @@
...
+     * <code>CREATE TABLE AS</code>. <code>false</code
+     * by default.
+     */
+    public boolean supportsWrite;
+
+    /**
+     * If the plugin supports read, the value of the
+     * {@link CoreOperatorType} for the scan operator.
+     */
+    public int readerId;
```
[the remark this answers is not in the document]
arina-ielchiieva: Yes, I did but I though we new framework had already possibility to use both name or id would simply things in future. Anyway, you can leave as is.

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseStoragePlugin.java
```diff
@@ -0,0 +1,229 @@
...
+ * objects during planning and execution.
+ *
+ * @param <C> the storage plugin configuration class
+ */
+
+public abstract class BaseStoragePlugin<C extends StoragePluginConfig>
+    extends AbstractStoragePlugin {
+
+  private static final Logger logger = LoggerFactory.getLogger(BaseStoragePlugin.class);
+  protected static final ObjectMapper DEFAULT_MAPPER = new ObjectMapper();
```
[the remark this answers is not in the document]
arina-ielchiieva: Well, unless it is static would would have to inject it using constructor or though some context (for example, DrillbitContext).

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/filter/ConstantHolder.java
```diff
@@ -0,0 +1,172 @@
...
+    default:
+      throw conversionError(toType);
+    }
+  }
+
+  public ConstantHolder normalize(MinorType toType) {
+    try {
+      return convertTo(toType);
+    } catch (Throwable e) {
+      return null;
```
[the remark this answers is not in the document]
arina-ielchiieva: What is others need to push down more? What they should do?

exec/java-exec/src/test/java/org/apache/drill/exec/store/base/DummyStoragePlugin.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override
+    public DummyGroupScan groupWithColumns(DummyGroupScan group,
+        List<SchemaPath> columns) {
+      return new DummyGroupScan(group, columns);
+    }
+
+    @Override
+    public ScanFrameworkBuilder scanBuilder(DummyStoragePlugin storagePlugin,
+        OptionManager options, DummySubScan subScan) {
+      ScanFrameworkBuilder builder = new ScanFrameworkBuilder();
```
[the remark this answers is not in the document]
arina-ielchiieva: Maybe add comment in Java doc somewhere? I could not have guessed this without your comment...

exec/java-exec/src/test/java/org/apache/drill/exec/store/base/DummyGroupScan.java
```diff
@@ -97,36 +100,33 @@ public RelOp acceptFilter(RelOp relOp) {
...
       return relOp;
     }
 
-    // All other columns apply only if projected
-
-    if (!FILTER_COLS.contains(relOp.colName)) {
+    // "allTypes" table filters everything. All other tables
+    // only project a fixed set of columns. Simulates a plugin
+    // which can project only some columns.
+    if (!scanSpec.tableName.equals("allTypes") && !FILTER_COLS.contains(relOp.colName)) {
```
arina-ielchiieva: Nit: usually constant goes first to avoid NPE `"allTypes".equals(...`

exec/java-exec/src/test/java/org/apache/drill/exec/store/base/PlanVerifier.java
```diff
@@ -0,0 +1,146 @@
...
+ * <p>
+ * The test works by comparing the actual {@code EXPLAIN} output
+ * to a "golden" file with the expected ("golden") plan.
+ * <p>
+ * To create a test, just write it and let it fail due to a missing file.
+ * The output will go to the console. Inspect it. If it looks good,
+ * copy the plan to the golden file and run again.
+ * <p>
+ * If comparison fails, you can optionally ask the verifier to write the
+ * output to {@code /tmp} so you can compare the golden and actual
```
arina-ielchiieva:
```suggestion
 * output to {@code /tmp/drill/test} so you can compare the golden and actual
```

exec/java-exec/src/test/java/org/apache/drill/exec/store/base/PlanVerifier.java
```diff
@@ -0,0 +1,146 @@
...
+public class PlanVerifier {
+
+  private final String basePath;
+
+  /**
+   * Persist test results. Turn this on if you have failing tests.
+   ( Then, you can diff the actual and golden results in your favorite
+   * tool or IDE.
+   */
+  public boolean saveResults;
```
arina-ielchiieva: Maybe add setter and make it private.

exec/java-exec/src/test/java/org/apache/drill/exec/store/base/TestFilterPushDown.java
```diff
@@ -0,0 +1,339 @@
...
+
+  @BeforeClass
+  public static void setup() throws Exception {
+    ClusterFixtureBuilder builder = new ClusterFixtureBuilder(dirTestWatcher);
+    startCluster(builder);
+
+    StoragePluginRegistry pluginRegistry = cluster.drillbit().getContext().getStorage();
+    DummyStoragePluginConfig config =
+        new DummyStoragePluginConfig(true, true, false);
+    pluginRegistry.createOrUpdate("dummy", config, true);
```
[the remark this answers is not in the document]
arina-ielchiieva: It is used in JDBC plugin tests, please leave only one verions than yours or previous one, I don't think we don't need two.

https://github.com/apache/drill/blob/465fae6b9dcea9e02a4bfb71d9a3b9d4b99e169d/contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithH2IT.java#L68

https://github.com/apache/drill/blob/465fae6b9dcea9e02a4bfb71d9a3b9d4b99e169d/contrib/storage-jdbc/src/test/java/org/apache/drill/exec/store/jdbc/TestJdbcPluginWithMySQLIT.java#L85

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseScanFactory.java
```diff
@@ -0,0 +1,82 @@
...
+ * @param <SUB> the sub scan class
+ */
+
+public abstract class BaseScanFactory<
+      PLUGIN extends BaseStoragePlugin<?>,
+      SPEC,
+      GROUP extends BaseGroupScan,
+      SUB extends BaseSubScan> {
+
+  @SuppressWarnings("unchecked")
```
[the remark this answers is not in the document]
arina-ielchiieva: I see, I guess we don't have netter solution than casting...

exec/java-exec/src/test/java/org/apache/drill/exec/store/base/DummyGroupScan.java
```diff
@@ -0,0 +1,214 @@
...
+  public DummyScanSpec scanSpec() { return scanSpec; }
+
+  @JsonProperty("filters")
+  public FilterSpec andFilters() { return filters; }
+
+  public boolean hasFilters() {
+    return filters != null && ! filters.isEmpty();
+  }
+
+  private static final List<String> FILTER_COLS = ImmutableList.of("a", "b", "id");
```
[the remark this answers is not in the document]
arina-ielchiieva: Use `Arrays.asList`

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/BaseGroupScan.java
```diff
@@ -0,0 +1,481 @@
...
+  public abstract SubScan getSpecificScan(int minorFragmentId);
+
+  @JsonIgnore
+  @Override
+  public String getDigest() { return toString(); }
+
+  /**
+   * Create a copy of the group scan with with candidate projection
+   * columns. Calls
+   * {@link BaseScanFactory#groupWithColumns()}
```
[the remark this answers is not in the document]
vvysotskyi: Thanks for the explanation, in this case, please also remove brackets, since it causes error in JavaDoc: `{@link BaseScanFactory#groupWithColumns()}` -> `{@link BaseScanFactory#groupWithColumns}` here and in JavaDoc for method below.

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/filter/FilterPushDownListener.java
```diff
@@ -0,0 +1,140 @@
...
+   * determines that the plugin can't handle (or cannot completely handle)
+   * a term, return the Calcite node for that term back as part of the
+   * return value and it will be left in the query. Any Calcite nodes
+   * not returned are removed from the query and it is the scan's responsibility
+   * to handle them. Either the group scan or the list of Calcite nodes
+   * must be non-null. Or, return null if the filter condition can't be handled
+   * and the query should remain unchanged.
+   */
+  Pair<GroupScan, List<RexNode>> transform(GroupScan groupScan,
+      List<Pair<RexNode, RelOp>> andTerms, Pair<RexNode, DisjunctionFilterSpec> orTerm);
```
vvysotskyi: I'm concerned about such an approach. We will limit possible set of functions which may be applied for pushdown by functions defined in `RelOp.Op`. For example, if we would have some source with partitions, we wouldn't be able to push filter with conditions like this: `id + 10 = 20`. For example, for parquet, can do filter pruning for such conditions.

vvysotskyi: I like the idea of returning both `GroupScan` and `List<RexNode>`, so the rule will be able either to create partial filter above scan or just use the scan or do nothing if null was returned. But I think it would be better in the listener's implementation to either use only range and equals conditions to be able to parallelize scan if it is possible or prune scan segments if we would have some metadata.

exec/java-exec/src/test/java/org/apache/drill/exec/store/base/TestFilterPushDown.java
```diff
@@ -0,0 +1,264 @@
...
+  }
+
+  //-------------------------------------------------
+  // Unsupported filter push-down cases
+
+  // No predicates
+
+  @Test
+  public void testNoPushDown() throws Exception
+  {
```
vvysotskyi: Please move brace to the same line

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/filter/FilterPushDownStrategy.java
vvysotskyi: I'm not sure whether we should limit it to a specific phase. The optimizations I mentioned, theoretically may happen at any phase, so we should be able to do incrementally add predicates that may participate.

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/filter/FilterPushDownListener.java
vvysotskyi: > We should be able to push the filter t2.b = 10 into the scan for t2. Do you know if Drill already handles this case? If not, we should add it.

Yes, I believe Drill able to do such optimizations (at least create the corresponding filter considering join condition), it was added in the scope of [DRILL-6173](https://issues.apache.org/jira/browse/DRILL-6173).

vvysotskyi: Regarding addressing this issue in the scope of another Jira, I don't think that it is a good idea since there are some PRs that are waiting for this one, and they may extend this interface, so updating its signature later and changing the approach will force us to rewrite its implementations.

vvysotskyi: > So, it seemed to me that, in general, different plugins might use different strategies, or combine strategies.
> 
> Is this what you addressed with your comment, or am I missing something?

Yes, I meant to provide plugin the ability to specify its own strategy instead of enforcing the existing one.

Ideally, I thought about implementing abstraction which will be able to cover the existing parquet filter pushdown strategy.

cgivre: @vvysotskyi Depending on the scope of [DRILL-7557](https://issues.apache.org/jira/projects/DRILL/issues/DRILL-7557), I think it's fine to get this initial enhancement committed so that we can start using it in storage plugins. My plan at least is to start with very simple rules to begin with.

Regardless, IMHO, this is a very important PR to get committed as it will greatly facilitate plugin development.

vvysotskyi: @cgivre, I agree that this one is a very important change and will help to simplify developing new plugins.
My objection is when we will merge it as it is, the scope of DRILL-7557 will become larger since except addressing those issues, we would have to modify the implementations which would use a provided interface.

paul-rogers:
Thanks everyone for the review comments. I think there are enough substantial suggestions that we should go ahead and pause this PR. I'll divide this effort into smaller pieces, each of which is more complete to address the issues raised in reviews. This is all good news: shows that there is demand for an improved storage plugin API and that the present work is too modest an improvement. Let me figure out an alternative approach.

exec/java-exec/src/main/java/org/apache/drill/exec/store/base/filter/FilterPushDownListener.java
paul-rogers: @vvysotskyi and @cgivre - all good points. There is nothing special about the strategy offered; it is just a set of rules implemented using Calcite. Other plugins are free to use their own rules (or copy/paste these rules, or extend them.)

The point is valid: we should commit something we agree upon. So, I'm going to rethink this to figure out how to handle the Parquet case. Maybe we need to address the underlying issues first (missing algebra rewrites in Drill) to see can avoid making the API too wide.

My fear is, if we simply expose all of the current complexity, then there is nothing to abstract; we might as well continue to copy/paste big wads of code so each plugin can make the few tweaks that are needed.

Just a matter of finding the time to understand all the use cases well enough to see what might be in common. I'll offer a proposal in a few days.

paul-rogers:
As noted earlier, the general consensus is that we should come up with a better overall solution than the stop-gap offered here. So, I'll separate this PR into (as least) two. One will generalize the projection push-down to avoid the copy/paste solution we use today.

For the plugins themselves, I'm looking into creating a proper "SPI" (service provider interface) based on Java's `ServiceLoader` pattern. Doing that requires digging much deeper into the existing plugin mechanism and refactoring that code to add a third storage plugin source (beyond the current classpath and system mechanisms.)

Because this work will take time, I'd suggest that other PRs don't wait for this work. Let's get the other PRs in. Then, we can convert them to the new mechanism at some future time.

PR #1916, opened 2019-12-04
DRILL-7463: Apache license is not added to the generated classes

vvysotskyi:
Jira - [DRILL-7463](https://issues.apache.org/jira/browse/DRILL-7463).

Added license which should be inserted into the generated java classes. Before that, these files contained the only license required for `.ftl` files which is not inserted since it is treated as ftl comment and therefore was ignored.


arina-ielchiieva:
+1

PR #1917, opened 2019-12-06
DRILL-7468: Metastore unit tests may fail when used sources from the release archive

vvysotskyi:
Jira - [DRILL-7468](https://issues.apache.org/jira/browse/DRILL-7468).


arina-ielchiieva:
+1

PR #1919, opened 2019-12-06
DRILL-7471: DESCRIBE TABLE command fails with ClassCastException when Metastore is enabled

vvysotskyi:
Please refer to [DRILL-7471](https://issues.apache.org/jira/browse/DRILL-7471) for details.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeColumnUtils.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/metastore/analyze/AnalyzeColumnUtils.java
@@ -21,6 +21,7 @@ import org.apache.calcite.sql.SqlKind;
 import org.apache.drill.common.types.TypeProtos;
 import org.apache.drill.metastore.statistics.BaseStatisticsKind;
 import org.apache.drill.metastore.statistics.ColumnStatisticsKind;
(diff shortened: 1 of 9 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/Records.java
```diff
@@ -750,10 +751,8 @@ public Partition(String catalog, String schemaName, String value, PartitionMetad
       this.LAST_MODIFIED_TIME = convertToTimestamp(partitionMetadata.getLastModifiedTime());
     }
 
-    public static List<Partition> fromSegment(String catalog, String schemaName, SegmentMetadata segmentMetadata) {
-      return segmentMetadata.getPartitionValues().stream()
-        .map(value -> new Partition(catalog, schemaName, value, segmentMetadata))
-        .collect(Collectors.toList());
+    public static Partition fromSegment(String catalog, String schemaName, SegmentMetadata segmentMetadata) {
```
arina-ielchiieva: Please revisit this change, I am not sure it's correct.

vvysotskyi: You are right, changed the approach to populate segment partition values correctly. Thanks a lot for pointing this.

arina-ielchiieva:
+1

PR #1921, opened 2019-12-10
DRILL-7478: Add Pull Request Template

cgivre:
Adds a pull request template to github.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,13 @@
+# [DRILL-XXXX](https://issues.apache.org/jira/browse/DRILL-XXXX): PR Title
+
+(Please replace `PR Title` with actual PR Title)
+
+## Description
+
+(Please describe the change. If more than one ticket is fixed, include a reference to those tickets.)
+
+## Documentation
+(Please describe user-visible changes similar to what should appear in the Drill documentation.)
+
+## Testing
+(Please describe how this PR has been tested.)


vvysotskyi:
Could you please move this file under `.github` directory?

vvysotskyi:
Could you please add examples into every section, for example as it was done there: https://github.com/apache/parquet-mr/blob/master/.github/PULL_REQUEST_TEMPLATE.md

cgivre:
Done

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,47 @@
...
+## Types of changes
+
+What types of changes does your code introduce? Put an `x` in all the boxes that apply:
+
+- [ ] Docs change / refactoring / dependency upgrade
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+
+## Checklist
```
vvysotskyi: Could you please add a point to check whether all newly added dependencies are really required.

cgivre: @vvysotskyi Are we good to go here?

paul-rogers:
What is the intent here? Looks like this PR is trying to solve two problems:

1. Provide a standard template for PR messages,
2. Provide a checklist of things we want the contributer to do.

I wonder, should we think about the *commit* message; the one that will travel with the code in Git?

For the commit message, a simple template would be helpful:

```
DRILL-XXXX: Summary

Description of change.
```

I have seen folks include only the bug number. (Of course, I'm not sure anyone ever reads the commit messages...)

For the PR message, do we insist people keep our headings? Only our headings? Should they replace the commit message with a differently worded message? That is, do we want people to describe the change twice, in two different ways?

Each time I've seen shops provide these templates, it ends up that either the templates are off the mark, or they grow to be cumbersome. They are no substitute for people just doing the right thing.

A good reviewer will look for tests (I do that first.) A `pom.xml` file change should be carefully reviewed. What are we solving by adding a template?

agozhiy:
I agree with @paul-rogers, the PR message shouldn't duplicate the commit message. Still, it's good to have some checklist to ensure that nothing is missed accidentally.

cgivre:
@paul-rogers 
The goals here were:
1.  To encourage authors of PRs, especially those PRs introducing new functionality, to make sure that there is enough documentation so that others can understand what the PR does.

2. To save everyone's time by encouraging authors to make sure that their PRs have adequate tests, licenses, etc.  so that the review process will be smoother. 

I'm fine with shortening it, but I really like the idea of a checklist so that authors (myself included) make sure that their PRs are really ready to go.  Ideally, I'd like also to include an issue template on JIRA so that when people have issues we don't have to go back and ask the obvious questions...

paul-rogers:
Makes sense. Maybe provide the commit template that gives the gist of the change. (The JIRA number is not sufficient, requires a check or JIRA just to find out, say, that this was upgrading library X to a new version.)

I like the pattern you have used: to create a README file to explain things. That becomes a permanent source of documentation. For other cases, a `.md` file or `package-info.java` can serve the same purpose.

This leaves the PR note. It should include the same info as the commit (though perhaps in another form.) We can offer two parts:

1. Outline of the topics that might be covered, if relevant. (Just omit the bits that don't apply rather than filling in "Does not apply.") This is where the (reformatted) commit message would go.
2. Checklist of common tasks. The author is free to remove those that don't apply (such a "check license terms for any new libraries" if no libraries are added.) Add those that do apply to this case.

paul-rogers:
To move this along, I'd like to propose we that have two files. The first would be a "Contributing.md" file in the Drill root directory. As a Markdown file, it can be easily viewed by clicking on it. Perhaps add a link to the file from the "README.md" file that shows by default in Github. In this file, add all the contribution guidelines, pointers to style, how to run unit tests, checklist for contributors, and so on. We can make this file as verbose as needed.

Then, have a separate PR template that is concise. Perhaps:

```
<Title should be DRILL-XXXX: Fix Summary>
<describe the change. If more than one ticket is fixed, include a reference to those tickets.)

## Documentation

<Describe user-visible changes in a form close to what should appear in documentation.>

## Tests

<Describe tests including new unit or other tests added which cover the change.>
```

cgivre:
I concur.  Here's the shortened PR template. We can add a `CONTRIBUTING.md` in another PR. I do think that would be a good addition.

arina-ielchiieva:
+1

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,12 @@
+# DRILL-XXXX: Summary
```
vvysotskyi: Could you please replace it with reference:
```suggestion
# [DRILL-XXXX](https://issues.apache.org/jira/browse/DRILL-XXXX): Summary
```

cgivre: Done!

vvysotskyi: Thanks!

PR #1922, opened 2019-12-11
DRILL-7476: Set lastSet on TransferPair copies

paul-rogers:
Variable-width nullable vectors maintain a "lastSet" field in the mutator. This field is used in "fill empties" logic when setting the vector's value count. This is true even if the vector is read-only, or has been transferred from another (read-only) vector. LastSet must be set to the row count or the code will helpfully overwrite existing offsets with 0.

Added batch validation code to catch this problem should it happen again.

Includes minor code cleanup in files visited during debugging.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/validate/BatchValidator.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/validate/BatchValidator.java
@@ -25,7 +25,10 @@ import org.apache.drill.exec.record.VectorContainer;
 import org.apache.drill.exec.record.VectorWrapper;
 import org.apache.drill.exec.vector.BitVector;
 import org.apache.drill.exec.vector.FixedWidthVector;
+import org.apache.drill.exec.vector.NullableVar16CharVector;
+import org.apache.drill.exec.vector.NullableVarBinaryVector;
 import org.apache.drill.exec.vector.NullableVarCharVector;
+import org.apache.drill.exec.vector.NullableVarDecimalVector;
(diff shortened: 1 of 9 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/validate/BatchValidator.java
```diff
@@ -320,6 +321,19 @@ private void validateNullableVector(String name, NullableVector vector) {
           "Outer value count = %d, but inner value count = %d",
           outerCount, valueCount));
     }
+    if (vector instanceof NullableVarCharVector) {
```
vvysotskyi: Could you please add these checks for other nullable `VariableWidthVector` implementations: `NullableVarBinaryVector`, `NullableVar16CharVector` and `NullableVarDecimalVector`.

paul-rogers:
@vvysotskyi, I had a nice unit test that I used when debugging. However, it turns out that we can't use it in Maven: the information in the system table ('connections') changes too much. (One row for one connection when run in the IDE, but 9 (?) rows when run in Maven.)

I can perhaps add a different test: the problem should appear anytime a `LIMIT` is done where the limit amount is the same as the row count. But, to create such a test, I'll need to do it on master before this fix (to verify that the test catches the error.)

Even better, this PR does include a new batch check that *did* catch the error. If the error occurs, all offsets in a VARCHAR offset vector will be 0. A check verifies that the largest offset is the same size as the data vector. If this bug occurs, that check will trigger and we'll see an error.

PR #1923, opened 2019-12-11
DRILL-7479: Partial fixes for metadata parameterized type issues

paul-rogers:
See DRILL-7479 and DRILL-7480 for an explanation. Adds generic
type parameters where needed to avoid the need to supporess
warnings. However, type parameters are probably not needed
at all and should be removed in the future for reasons explained
in DRILL-7480.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ComparisonPredicate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ComparisonPredicate.java
@@ -88,11 +88,11 @@ public class ComparisonPredicate<C extends Comparable<C>> extends LogicalExpress
   @Override
   @SuppressWarnings("unchecked")
   public RowsMatch matches(StatisticsProvider<C> evaluator) {
-    ColumnStatistics leftStat = left.accept(evaluator, null);
+    ColumnStatistics<BigInteger> leftStat = (ColumnStatistics<BigInteger>) left.accept(evaluator, null);
     if (IsPredicate.isNullOrEmpty(leftStat)) {
       return RowsMatch.SOME;
     }
-    ColumnStatistics rightStat = right.accept(evaluator, null);
+    ColumnStatistics<BigInteger> rightStat = (ColumnStatistics<BigInteger>) right.accept(evaluator, null);
     if (IsPredicate.isNullOrEmpty(rightStat)) {
       return RowsMatch.SOME;
     }
@@ -117,7 +117,7 @@ public class ComparisonPredicate<C extends Comparable<C>> extends LogicalExpress
         leftStat = adjustDecimalStatistics(leftStat, rightScale - leftScale);
       }
     }
-    return predicate.apply(leftStat, rightStat);
+    return predicate.apply((ColumnStatistics<C>) leftStat, (ColumnStatistics<C>) rightStat);
   }
 
   /**
@@ -127,7 +127,7 @@ public class ComparisonPredicate<C extends Comparable<C>> extends LogicalExpress
    * @param scale adjustment scale
    * @return adjusted statistics
    */
-  private ColumnStatistics adjustDecimalStatistics(ColumnStatistics<BigInteger> statistics, int scale) {
+  private ColumnStatistics<BigInteger> adjustDecimalStatistics(ColumnStatistics<BigInteger> statistics, int scale) {
     BigInteger min = new BigDecimal(ColumnStatisticsKind.MIN_VALUE.getValueStatistic(statistics))
         .setScale(scale, RoundingMode.HALF_UP).unscaledValue();
     BigInteger max = new BigDecimal(ColumnStatisticsKind.MAX_VALUE.getValueStatistic(statistics))
@@ -139,7 +139,7 @@ public class ComparisonPredicate<C extends Comparable<C>> extends LogicalExpress
   /**
    * If one rowgroup contains some null values, change the RowsMatch.ALL into RowsMatch.SOME (null values should be discarded by filter)
    */
-  private static RowsMatch checkNull(ColumnStatistics leftStat, ColumnStatistics rightStat) {
+  private static RowsMatch checkNull(ColumnStatistics<?> leftStat, ColumnStatistics<?> rightStat) {
     return !IsPredicate.hasNoNulls(leftStat) || !IsPredicate.hasNoNulls(rightStat) ? RowsMatch.SOME : RowsMatch.ALL;
   }
 
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ComparisonPredicate.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/expr/ComparisonPredicate.java
@@ -88,11 +88,11 @@ public class ComparisonPredicate<C extends Comparable<C>> extends LogicalExpress
   @Override
   @SuppressWarnings("unchecked")
   public RowsMatch matches(StatisticsProvider<C> evaluator) {
-    ColumnStatistics<BigInteger> leftStat = (ColumnStatistics<BigInteger>) left.accept(evaluator, null);
+    ColumnStatistics<C> leftStat = (ColumnStatistics<C>) left.accept(evaluator, null);
     if (IsPredicate.isNullOrEmpty(leftStat)) {
       return RowsMatch.SOME;
     }
-    ColumnStatistics<BigInteger> rightStat = (ColumnStatistics<BigInteger>) right.accept(evaluator, null);
+    ColumnStatistics<C> rightStat = (ColumnStatistics<C>) right.accept(evaluator, null);
     if (IsPredicate.isNullOrEmpty(rightStat)) {
       return RowsMatch.SOME;
     }
@@ -112,12 +112,14 @@ public class ComparisonPredicate<C extends Comparable<C>> extends LogicalExpress
       int leftScale = left.getMajorType().getScale();
       int rightScale = right.getMajorType().getScale();
       if (leftScale > rightScale) {
-        rightStat = adjustDecimalStatistics(rightStat, leftScale - rightScale);
+        rightStat = (ColumnStatistics<C>) adjustDecimalStatistics(
+            (ColumnStatistics<BigInteger>) rightStat, leftScale - rightScale);
       } else if (leftScale < rightScale) {
-        leftStat = adjustDecimalStatistics(leftStat, rightScale - leftScale);
+        leftStat = (ColumnStatistics<C>) adjustDecimalStatistics(
+            (ColumnStatistics<BigInteger>) leftStat, rightScale - leftScale);
       }
     }
-    return predicate.apply((ColumnStatistics<C>) leftStat, (ColumnStatistics<C>) rightStat);
+    return predicate.apply(leftStat, rightStat);
   }
 
   /**
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
@@ -207,7 +207,7 @@ public abstract class BaseParquetMetadataProvider implements ParquetMetadataProv
   @Override
   public TableMetadata getTableMetadata() {
     if (tableMetadata == null) {
-      List<StatisticsHolder> tableStatistics = new ArrayList<>(DrillStatsTable.getEstimatedTableStats(statsTable));
+      List<StatisticsHolder<?>> tableStatistics = new ArrayList<>(DrillStatsTable.getEstimatedTableStats(statsTable));
       Map<SchemaPath, TypeProtos.MajorType> fields = ParquetTableMetadataUtils.resolveFields(parquetTableMetadata);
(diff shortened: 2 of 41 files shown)


metastore/metastore-api/src/main/java/org/apache/drill/metastore/util/TableMetadataUtils.java
```diff
@@ -46,21 +46,24 @@
...
    */
-  public static Comparator getComparator(TypeProtos.MinorType type) {
+  @SuppressWarnings("unchecked")
+  public static <T> Comparator<T> getComparator(TypeProtos.MinorType type) {
     switch (type) {
       case INTERVALDAY:
       case INTERVAL:
       case INTERVALYEAR:
-        return Comparator.nullsFirst(UnsignedBytes.lexicographicalComparator());
+        return (Comparator<T>) (Comparator<?>)
```
vvysotskyi:
```suggestion
        return (Comparator<?>)
```

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/FilterEvaluatorUtils.java
```diff
@@ -57,7 +57,6 @@
   private FilterEvaluatorUtils() {
   }
 
-  @SuppressWarnings("RedundantTypeArguments")
```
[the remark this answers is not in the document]
paul-rogers: Odd. In Eclipse, it is marked as unneeded. Reverted.

metastore/metastore-api/src/main/java/org/apache/drill/metastore/util/TableMetadataUtils.java
paul-rogers: Added a comment about why this horrible hack is necessary. Basically, the method is poorly designed, but I didn't want to make major changes just before a release.

paul-rogers:
Revised the PR with some ideas from DRILL-7479. The `ColumnStatistics` class was made generic: it has no type.

`ColumnStatistics` is a metadata description of a column: it, itself, is not a column and does not hold data for the column. It is important that the description describe the column type (which it does via the `type` field). But, the Java type of the metadata description need not reflect the type of the eventual values. (Think of `ColumnMetadata` or `Class` as examples.)

The result was that I could revert a number of changes and the resulting code is a bit cleaner.

This change should not affect any existing, serialized data for this class.

metastore/metastore-api/src/main/java/org/apache/drill/metastore/util/TableMetadataUtils.java
vvysotskyi: Thanks for adding a comment with an explanation.

paul-rogers:
Tried to clean up a few more warnings. Have found even more cases where we want to assign types statically, but work with them generically at runtime. Had to add some horrible hacks to cast away the (unknown) type.

Also, found another issue in `FilterExpression.Visitor`. The predicates are parameterized, but untyped in the visitor class. Typing them causes many tests to fail. So, we've got a case where we *must* abuse the Java type system to get the code to work. That is a strong "code smell" that some adjusting wants to be done.

The one issue I can see is that some of the proposed changes will change the serialized metadata format. If anyone starts to use the features from this release, some form of upgrade will be needed when that user upgrades. This incompatibility could be a reason to fix the issues now rather than later.

paul-rogers:
All that said, hat's off to the folks who did this code. There is an incredible amount of complex functionality in the metadata & Parquet filter push down modules! We just need a bit of tweaking to the very fiddly problem of handling types both statically and dynamically.

vvysotskyi:
@paul-rogers, regarding `FilterExpression.Visitor`, it may seem that generics usage is redundant here, but such design was chosen to make this class extendable as much as possible but with strict API. By design, implementations of this class should be used for specific metastore implementation, so it may be required to provide a storage-specific type for the visitor, as it was done in the existing implementation. Though, there is a single implementation typed with `org.apache.iceberg.expressionsExpression` (non-Drill class), but in the future, it is possible, that other implementations will be added. 
I worked with the code of `iceberg-metastore` module a lot, and every time I ensured that this is the most thoughtful and stainless part of the metastore work.

Regarding changing the serialized metadata format, could you please share some details on how it happens, since I thought that we use only actual types of data that is written to the metastore.

paul-rogers:
Another commit to fix the remaining warnings in metastore.

Just to summarize based on what I've seen, there are two use cases:

* Generic: Example: take two columns, get their min values generically, combine them generically. Compile-time typing is a hindrance here; we need run-time typing.
* Specific: compute or obtain a specific value of a specific type. The NDV, say, to be used in estimating selectivity.

It is easy to go from specific to generic:

```
  int foo = ...;
  metadata.setValue(foo);
...
  public void setValue(Object value) {
...
```

If is harder to go the other way. If we know the value is an `int`:

```
  int foo = metadata.getInt();
  // Or, less ideally:
  assert metadata.getType() == MinorType.INT;
  int foo = (Int) metadata.getValue();
```

Here `getInt()` just does in a method what the second example does inline.

To be type aware for all types:

```
  switch(metadata.getType()) {
    case INT:
      int intFoo = (Int) metadata.getValue();
      ...
    case VARCHAR:
      String strFoo = (String) metadata.getValue();
    ...
```

The above is true if we introduce, say a `ValueHolder<T>` that holds a value of type T. Instead of casting the value, we'd cast the value holder. There is no escaping the issue.

Point is: we need to understand the use cases so we can more precisely refine the metadata data model.

Some example:

* The pending filter push-down PR: uses a value holder with a (type, value) pair.
* The Impala solution: has nodes that represent literals, with subclasses for various kinds of literal: `IntLiteral`, `StringLiteral`, etc.
* The JDBC approach: a separate schema that tells us that when `foo` appears, it is an `int`, so we should use `getInt(colIndex)`.
* The `ColumnMetadata` (and HOCON config) approach; store values as strings, parse to the desired type on demand.

Perhaps one of these (or, more likely, some combination) might help us here.

paul-rogers:
@vvysotskyi, thanks for the explanation on the `FilterExpression.Visitor` class. I think your idea is right. The question is, how do we achieve extensibility? This is a very complex subject to discuss via written form, but I'll try. I'm sure I'll miss something obvious.

Any metastore can add its own metadata fields (`min`, `max`, `myOwnMetric`), etc. When that happens, the rest of Drill should *not* change. So, adding a class to a class hierarchy that Drill must understand may not be the way to go. Instead, we have to identify the parts that are fixed by Drill and the parts that the specific metastore can extend.

Let's take our list of metadata "fields." Drill defines `min` and `max`: I as a metastore author need to honor Drill's definition. (Or, I need to tell that I can't report `min` and `max`, but I can report, say, `NDV` because I'm Hive.) On the other hand, `myOwnMetric` is something I define for use, perhaps, in my own WizzBang store plugin.

We are talking about a predicate visitor. For the code to compile in Drill, the visitor has to know about all possible predicates that can be visited; that's the only way to create the `visit(Foo)` methods in the base class.

I like how you've categorized the predicates. (MUCH easier than using internal function names the way many plugins currently do.) (I might quibble and say that there are only four predicate types: Unary (IS NULL, IS TRUE, etc.), Binary (=, !=, etc.), Boolean (AND, OR) and List (IN).) Yes, Boolean predicates are binary, but they are "special" because, by definition, each of their two arguments is, itself, a predicate (or a Boolean literal.)

So, since Drill defines the predicates (they come from the SQL query), the types should be known to Drill.

For predicates derived from a SQL query, the argument types are also defined: Drill must have been able to parse them out of the SQL text and apply some type (at least Number, String, Boolean). And, predicates can be of mixed types: `10 = '10'` say. So, the predicate *result* is binary, but the arguments are of any type (even if they don't make sense as in my example.)

What, then, can be extended? The underlying type of a column. (Maybe `INT96` in Parquet, or `IPAddress` in some security system.) My custom plugin has to convert between my values (`INT96`) and Drill's values. So, if we see `X = 123456`, our custom code might know that `X` is a column in our DB that is of type `INT96`, so we might then try to convert the Drill-type (`INT`, say) of `123456` into the `INT96` type for our push-down filter.

In stats, I might define `min` to be of my own specific `INT96` type. (That is, the `min` value is of the type of the actual data, not of the type that the reader will convert it to in Drill.)

Drill's description of the min value must be generic, however. (Drill does not know about my type.) One way to do that is to add "meta-meta-data" that says that the value of this metadata field is my private `INT96` type. Only I know how to read those values; Drill treats them as `Object`s.

If something wanted to do a query on my metadata, my "meta-meta-data" could provide a `toString` method to convert the `INT96` `min` value to something that Drill can display.

Oh, one other topic to cover. Different systems provide different stats. Generally, a system provides NDV or a histogram (or nothing). This info exists, primarily, to compute selectivity for a filter. (How many values will `X = 10` match?) Where should this logic be? Generically in Drill? For equality:

```
  if (metadata == null) {
    sel = 0.15;
  } else if (metadata.get("histogram") != null) {
    sel = metadata.getHistorgramProbability(arg);
   } else if (metadata.get("ndv") != null) {
    sel = 1 / metadata.get("ndv");
  } ...
```

Or, we could make this more generic: require that the metadata provide a `computeSelectivity(arg)` method that "does the right thing" for the system in question with the metadata available. (A special "dummy" metadata provider could provide Calcite's default 0.15 guess.)

In short, this is a tricky topic. We have to think about what Drill does and what the extension does. Maybe we can move the discussion to another forum (maybe direct e-mail) so we don't spam the dev list. Then we can work through specific use cases so I can perhaps make more specific suggestions.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ComparisonPredicate.java
```diff
@@ -88,11 +88,11 @@ private ComparisonPredicate(LogicalExpression left,
   @Override
   @SuppressWarnings("unchecked")
   public RowsMatch matches(StatisticsProvider<C> evaluator) {
-    ColumnStatistics leftStat = left.accept(evaluator, null);
+    ColumnStatistics<BigInteger> leftStat = (ColumnStatistics<BigInteger>) left.accept(evaluator, null);
```
vvysotskyi: Actually, the value here may be not only with `BigInteger` type. Please use `C` for generic and do the cast after the check for `VARDECIMAL` type.

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
```diff
@@ -485,6 +486,7 @@ private static SegmentMetadata combineSegmentMetadata(Collection<SegmentMetadata
       if (entries.isEmpty() || !collectMetadata) {
         return Collections.emptyMap();
       }
+      @SuppressWarnings("unused")
```
vvysotskyi: `addRowGroups` is used in the code below, so perhaps it is your IDE false-positive warning, IDEA doesn't highlight it as unused.

metastore/metastore-api/src/main/java/org/apache/drill/metastore/components/tables/BasicTablesTransformer.java
```diff
@@ -99,6 +99,9 @@ public static MetadataHolder all(List<TableMetadataUnit> units) {
         case PARTITION:
           partitions.add(PartitionMetadata.builder().metadataUnit(unit).build());
           break;
+        default:
+          // Throw an error?
```
vvysotskyi: Please either throw an error here or remove this comment. But it is impossible to obtain a non-listed type here.

metastore/metastore-api/src/main/java/org/apache/drill/metastore/components/tables/TableMetadataUnit.java
```diff
@@ -511,6 +511,9 @@ public static Schema of(Class<?> unitClass, Class<?> builderClass) {
               rowGroupColumns.add(name);
               partitionColumns.add(name);
               break;
+            default:
+              // Throw an error?
```
vvysotskyi: The same here.

vvysotskyi:
@paul-rogers, moving the discussion into another place is a good idea. Perhaps we can continue it in a Jira to preserve our ideas for other contributors.

Regarding the additional changes, I think we should merge this PR to avoid errors in Eclipse and continue the release process since it was already postponed a lot due to metastore issues...

arina-ielchiieva:
@vvysotskyi please copy discussion points and link to this PR into https://issues.apache.org/jira/browse/DRILL-7480 .
What else is left for @paul-rogers to do so this PR can be merged?

vvysotskyi:
I have added a couple of comments to the commit which was added after I have approved this PR, so they should be resolved to merge this PR.

@paul-rogers, could you please take a look at them?

exec/java-exec/src/main/java/org/apache/drill/exec/store/parquet/BaseParquetMetadataProvider.java
paul-rogers: Eclipse claims this is dead code, but I can't figure out why. So, I'll just remove the annotation and leave well-enough alone.

paul-rogers:
Make requested changes.

The changes here were the simplest, mechanical changes I could make to remove the many rawtypes warnings and the error which Eclipse flagged.

@vvysotskyi, think of this as one possible way to temporarily solve the raw types issues. Please suggest better solutions, push a commit with your improvements, or offer an alternative PR.

vvysotskyi:
@paul-rogers, thanks for making changes, +1

PR #1925, opened 2019-12-11
DRILL-7472: Fix ser / de for sys and information_schema schemas queries

arina-ielchiieva:
Jira - [DRILL-7472](https://issues.apache.org/jira/browse/DRILL-7472).

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
@@ -39,8 +39,6 @@ public class InfoSchemaGroupScan extends AbstractGroupScan {
   private final InfoSchemaTableType table;
   private final InfoSchemaFilter filter;
 
-  private boolean isFilterPushedDown = false;
-
   public InfoSchemaGroupScan(InfoSchemaTableType table) {
     this(table, null);
   }
@@ -57,7 +55,6 @@ public class InfoSchemaGroupScan extends AbstractGroupScan {
     super(that);
     this.table = that.table;
     this.filter = that.filter;
-    this.isFilterPushedDown = that.isFilterPushedDown;
   }
 
   @JsonProperty("table")
@@ -68,6 +65,12 @@ public class InfoSchemaGroupScan extends AbstractGroupScan {
   @JsonProperty("filter")
   public InfoSchemaFilter getSchemaFilter() {
     return filter;
+  }
+
+  @JsonIgnore
+  @Override
+  public List<SchemaPath> getColumns() {
+    return super.getColumns();
   }
 
   @Override
@@ -81,7 +84,8 @@ public class InfoSchemaGroupScan extends AbstractGroupScan {
     return new InfoSchemaSubScan(table, filter);
   }
 
-  public ScanStats getScanStats(){
+  @Override
+  public ScanStats getScanStats() {
     if (filter == null) {
       return ScanStats.TRIVIAL_TABLE;
     } else {
@@ -103,7 +107,7 @@ public class InfoSchemaGroupScan extends AbstractGroupScan {
 
   @Override
   public String getDigest() {
-    return this.table.toString() + ", filter=" + filter;
+    return table.toString() + ", filter=" + filter;
   }
 
   @Override
@@ -111,12 +115,8 @@ public class InfoSchemaGroupScan extends AbstractGroupScan {
     return new InfoSchemaGroupScan(this);
   }
 
-  public void setFilterPushedDown(boolean status) {
-    this.isFilterPushedDown = status;
-  }
-
   @JsonIgnore
   public boolean isFilterPushedDown() {
-    return isFilterPushedDown;
+    return filter != null;
   }
 }
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTableScan.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTableScan.java
@@ -42,7 +42,6 @@ import org.apache.drill.exec.store.StoragePluginRegistry;
 
 @JsonTypeName("sys")
 public class SystemTableScan extends AbstractGroupScan implements SubScan {
-  // private static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(SystemTableScan.class);
 
   private final SystemTable table;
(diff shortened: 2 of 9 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
```diff
@@ -70,6 +67,12 @@ public InfoSchemaFilter getSchemaFilter() {
     return filter;
   }
 
+  @JsonIgnore
+  @Override
+  public List<SchemaPath> getColumns() {
+    return super.getColumns();
+  }
```
paul-rogers: I believe that we do need to serialize the columns as part of the logical plan. (Project push down occurs before creating the logical plan JSON.)

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
```diff
@@ -111,12 +114,8 @@ public GroupScan clone(List<SchemaPath> columns) {
...
   }
 
-  public void setFilterPushedDown(boolean status) {
-    this.isFilterPushedDown = status;
-  }
-
   @JsonIgnore
   public boolean isFilterPushedDown() {
-    return isFilterPushedDown;
+    return filter != null;
```
paul-rogers: This is tricky. This flag is probably acting more like "has filter push-down analysis been done": it prevents Calcite from pushing filters down multiple times (which won't work if the group scan removes the original `WHERE` clause conditions); and it prevents Calcite from needlessly repeating the analysis multiple times when there are no filters.

The proposed change means we will do that no-filter analysis repeatedly.

(As it turns out, I made this same mistake myself in a recent project.)

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTableScan.java
```diff
@@ -85,6 +83,12 @@ public PhysicalOperator getNewWithChildren(List<PhysicalOperator> children) {
   public void applyAssignments(List<DrillbitEndpoint> endpoints) {
   }
 
+  @JsonIgnore
+  @Override
+  public List<SchemaPath> getColumns() {
+    return super.getColumns();
+  }
```
paul-rogers: See note above. Columns are part of the logical plan and must be serialized. To see this, do an `EXPLAIN PLAN FOR`. You'll see that the JSON includes the column projection for other group scans.

exec/java-exec/src/test/java/org/apache/drill/exec/store/sys/TestSystemTable.java
```diff
@@ -61,6 +61,12 @@ public void optionsOrderBy() throws Exception {
     test("select * from sys.options order by name");
   }
 
+  @Test
+  public void testSerDe() throws Exception {
+    PlanTestBase.testPhysicalPlanExecutionBasedOnQuery("select * from sys.threads");
+    PlanTestBase.testPhysicalPlanExecutionBasedOnQuery("select * from sys.connections limit 10");
```
paul-rogers: This test is sort of OK. (It converts the plan to a physical plan, then executes it.) But, we don't check the results.

A better SerDe test is to convert the object to JSON, then convert back and compare the two objects. (I have such a test I added for the column metadata serialization for which I need to issue a PR after the release.)

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
arina-ielchiieva: I have removed this flag since instead of using additional boolean variable (which by the way was not serialized), it can be replaced with check if filter is set or not. Filter is set only if it was pushed down, so thus checking that filter was set `will prevent Calcite from pushing filters down multiple times`.

https://github.com/apache/drill/blob/master/exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaPushFilterIntoRecordGenerator.java#L94-L99

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
arina-ielchiieva: Sys and InfoSchema plugins always produce all columns but added them to the plan to avoid confusion.

exec/java-exec/src/main/java/org/apache/drill/exec/store/sys/SystemTableScan.java
arina-ielchiieva: Added columns to the plan.

exec/java-exec/src/test/java/org/apache/drill/exec/store/sys/TestSystemTable.java
arina-ielchiieva: Added test to verify ser / de plus results.

arina-ielchiieva:
@paul-rogers addressed code review comment, please take a look.

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
paul-rogers: Ah, I see. So, I'll amend my comment: we need to serialize columns only if the plugin supports project push-down. With EVF, that is all of them. But, of course, this one is old and may not support project push-down.

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
paul-rogers: Sounds good.

paul-rogers:
Thanks for fixing this. Some of the issues discussed here suggest that Drill has grown to have many ways to do the same thing (project push-down or not, etc.) Would be good to fix those over time.

Still, this one looks good, so
+1

exec/java-exec/src/main/java/org/apache/drill/exec/store/ischema/InfoSchemaGroupScan.java
arina-ielchiieva: Yes, these plugins do not support project push-down, I'll remove columns ser / de.

PR #1926, opened 2019-12-11
DRILL-7474: Reduce size of Drill's tar.gz file

agozhiy:
- Excluded aws-java-sdk-bundle dependency, only required aws libraries added instead.
- Cleared format-excel module from unused dependencies.

JIRA: [DRILL-7474](https://issues.apache.org/jira/browse/DRILL-7474)

### DIFF
--- a/contrib/format-excel/pom.xml
+++ b/contrib/format-excel/pom.xml
@@ -50,21 +50,6 @@
       <artifactId>poi-ooxml</artifactId>
       <version>${poi.version}</version>
     </dependency>
-    <dependency>
-      <groupId>org.apache.poi</groupId>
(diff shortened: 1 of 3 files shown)


arina-ielchiieva:
@agozhiy is there any statistics how size has changed?

agozhiy:
@arina-ielchiieva,
Before: 403942K
After: 308749K
(~93M difference)
Slightly bigger than in 1.16.0 (283120K), but not so critical (we have some new features after all).

arina-ielchiieva:
Awesome! Thanks for making the changes. 
Additionally tested S3 connection on your branch, everything works fine.
+1

PR #1927, opened 2019-12-12
DRILL-7482: Fix missing artifact and overlapping classes warnings in Drill build

vvysotskyi:
Jira - [DRILL-7482](https://issues.apache.org/jira/browse/DRILL-7482)

- Added `javax.el` into dependency management block with specifying the version to avoid searching SNAPSHOT versions since some dependencies specify their version range.
- Excluded `com.google.code.findbugs:jsr305` from `jdbc-all` jar since all required annotations are present in `com.google.code.findbugs:annotations` jar.
- Excluded `org.slf4j`classes shaded in `org.apache.calcite.avatica` jar from `jdbc-all`.
- Completely excluded `parquet-hadoop-bundle` shaded  in `hive-exec` to use the version provided by Drill in `drill-hive-exec-shaded`.

### DIFF
--- a/contrib/storage-hive/hive-exec-shade/pom.xml
+++ b/contrib/storage-hive/hive-exec-shade/pom.xml
@@ -92,6 +92,8 @@
           <artifactSet>
             <includes>
               <include>org.apache.hive:hive-exec</include>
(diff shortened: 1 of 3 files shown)


arina-ielchiieva:
+1, LGTM

PR #1931, opened 2019-12-13
DRILL-6332: Allow to provide two-component Kerberos principals

j-4:
Currently drill only allows for three-component kerberos principals `primary/instance@REALM`
I had the requirement to deal with two-part principals `primary@REALM` and thus made these little changes to the principal checks. This enables drill to deal with two- and three-part principals.

Behaviour: The principal splitting function will return an empty string for `instance` if only a two-component principal is given.

### DIFF
--- a/common/src/main/java/org/apache/drill/common/KerberosUtil.java
+++ b/common/src/main/java/org/apache/drill/common/KerberosUtil.java
@@ -38,29 +38,38 @@ public final class KerberosUtil {
    * Returns principal of format primary/instance@REALM.
    *
    * @param primary non-null primary component
-   * @param instance non-null instance component
+   * @param instance non-null instance component, can be empty string
    * @param realm non-null realm component
-   * @return principal of format primary/instance@REALM
+   * @return principal of format primary/instance@REALM or primary@REALM
    */
   public static String getPrincipalFromParts(final String primary, final String instance, final String realm) {
-    return checkNotNull(primary) + "/" +
-        checkNotNull(instance) + "@" +
-        checkNotNull(realm);
+    checkNotNull(primary);
+    checkNotNull(realm);
+
+    return primary +
+        ((instance != "") ? "/" + instance : "")
+        + "@" + realm;
   }
 
   /**
-   * Expects principal of the format primary/instance@REALM.
+   * Expects principal of the format primary/instance@REALM or primary@REALM.
    *
    * @param principal principal
    * @return components
    */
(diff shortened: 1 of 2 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/server/BootStrapContext.java
```diff
@@ -135,10 +135,11 @@ private void login(final DrillConfig config) throws DrillbitStartupException {
...
         if (parts.length != 3) {
           throw new DrillbitStartupException(
-              String.format("Invalid %s, Drill service principal must be of format: primary/instance@REALM",
-                ExecConstants.SERVICE_PRINCIPAL));
+            String.format("Invalid %s, Drill service principal must be of format 'primary/instance@REALM' or 'primary@REALM'",
+              ExecConstants.SERVICE_PRINCIPAL));
         }
-        parts[1] = KerberosUtil.canonicalizeInstanceName(parts[1], hostName);
+
+        parts[1] = (parts[1] == "")? "": KerberosUtil.canonicalizeInstanceName(parts[1], hostName);
```
arina-ielchiieva:
```suggestion
        parts[1] = (parts[1] == "") ? "" : KerberosUtil.canonicalizeInstanceName(parts[1], hostName);
```

common/src/main/java/org/apache/drill/common/KerberosUtil.java
```diff
@@ -38,29 +38,39 @@
...
    */
   public static String getPrincipalFromParts(final String primary, final String instance, final String realm) {
-    return checkNotNull(primary) + "/" +
-        checkNotNull(instance) + "@" +
-        checkNotNull(realm);
+    checkNotNull(primary);
+    checkNotNull(realm);
+
+    return checkNotNull(primary) +
+        ((instance != "")? "/" + instance: "")
```
arina-ielchiieva:
```suggestion
        ((instance != "") ? "/" + instance : "")
```

arina-ielchiieva:
@j-4 overrall looks good, please address code review comments and update commit message (should be the same as PR naming which I have already updated).

cgivre:
@j-4 Thanks for the contribution and welcome to Drill!

j-4:
Thanks for the worm welcome and the really fast code review - sorry for not being able to keep up this pace ;-) I edited the commit message and changed the code according to your suggestions.

Thanks for your great work on Drill!

arina-ielchiieva:

PR #1933, opened 2019-12-13
DRILL-7473: Parquet reader failed to get field of repeated map

KazydubB:
Jira - [DRILL-7473](https://issues.apache.org/jira/browse/DRILL-7473)


exec/java-exec/src/test/java/org/apache/drill/exec/store/parquet/TestParquetComplex.java
```diff
@@ -835,4 +835,19 @@ public void testDictValueInFilter3() throws Exception {
         .baselineValues(2, TestBuilder.mapOfObject("a", 1, "b", 2, "c", 3))
         .go();
   }
+
+  @Test // DRILL-7473
+  public void testSelect() throws Exception {
```
arina-ielchiieva: Please give better naming for the test

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/validate/BatchValidator.java
```diff
@@ -293,14 +292,20 @@ private void validateVector(String name, ValueVector vector) {
       // structure to check.
     } else if (vector instanceof BaseRepeatedValueVector) {
       validateRepeatedVector(name, (BaseRepeatedValueVector) vector);
-    } else if (vector instanceof RepeatedMapVector) {
-      validateRepeatedMapVector(name, (RepeatedMapVector) vector);
+    } else if (vector instanceof AbstractRepeatedMapVector) {
```
ihuzenko: Could you please clarify your comment below, for example

>Checking ```AbstractRepeatedMapVector``` includes DictVector, in order to avoid NPE while getting value from ```NullReader```. The ```NullReader``` may be returned when index points to the absent value in dict array, example query ```SELECT dict_array[3].keyInDict ... ``` .

KazydubB: I've updated the comment - hope it is clearer

PR #1935, opened 2019-12-14
DRILL-7483: Add support for 12 and 13 java versions

vvysotskyi:
- Updated ASM (made required changes in code), javassist and maven-surefire-plugin versions to the versions which support JDK 13
- Updated CircleCI configs to run tests for  JDK 8, 11, 12 and 13
- Disabled Hive tests for JDK 9+ since current version doesn't support newer java versions
- Fixed small metastore issue connected with calculating last modified time when performed analyze for table level
- Updated TestIcebergTableSchema test to generate required classes using ASM since `net.openhft:compiler` doesn't support JDK 12+ (due to new restrictions added in https://bugs.openjdk.java.net/browse/JDK-8210522)

Jira - [DRILL-7483](https://issues.apache.org/jira/browse/DRILL-7483)

### DIFF
--- a/contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveClusterTest.java
+++ b/contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveClusterTest.java
@@ -20,8 +20,6 @@ package org.apache.drill.exec.hive;
 import org.apache.drill.test.ClusterTest;
 import org.junit.BeforeClass;
 
-import static org.junit.Assume.assumeTrue;
-
 /**
  * Base class for Hive cluster tests.
  */
@@ -29,6 +27,6 @@ public class HiveClusterTest extends ClusterTest {
 
   @BeforeClass
   public static void checkJavaVersion() {
-    assumeTrue("Skipping tests since Hive supports only JDK 8.", HiveTestUtilities.supportedJavaVersion());
+    HiveTestUtilities.assumeJavaVersion();
   }
 }
--- a/contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveClusterTest.java
+++ b/contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveClusterTest.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.drill.exec.hive;
+
+import org.apache.drill.test.ClusterTest;
+import org.junit.BeforeClass;
+
+import static org.junit.Assume.assumeTrue;
+
+/**
+ * Base class for Hive cluster tests.
+ */
+public class HiveClusterTest extends ClusterTest {
+
+  @BeforeClass
+  public static void checkJavaVersion() {
+    assumeTrue("Skipping tests since Hive supports only JDK 8.", HiveTestUtilities.supportedJavaVersion());
+  }
+}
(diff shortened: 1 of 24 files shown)


contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveTestUtilities.java
```diff
@@ -128,4 +128,13 @@ public static void assertNativeScanUsed(QueryBuilder queryBuilder, String table)
     assertThat(plan, containsString("HiveDrillNativeParquetScan"));
   }
 
+  /**
+   * Current Hive version doesn't support JDK 9+.
+   * Checks whether used supported for Hive Java version.
```
arina-ielchiieva: Checks if current version is supported by Hive.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveTestUtilities.java
```diff
@@ -128,4 +128,13 @@ public static void assertNativeScanUsed(QueryBuilder queryBuilder, String table)
...
+   *
+   * @return {@code true} if used supported for Hive Java version, {@code false} otherwise
```
arina-ielchiieva: Please re-phase `if used supported for Hive Java version`, similar to the comment above.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/complex_types/TestHiveArrays.java
```diff
@@ -59,6 +60,7 @@
 
   @BeforeClass
   public static void setUp() throws Exception {
+    assumeTrue("Skipping tests since Hive supports only JDK 8.", HiveTestUtilities.supportedJavaVersion());
```
arina-ielchiieva: Maybe its better to create parent class for all Hive tests and add check there so we don't have to repeat the check?

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/impersonation/hive/TestSqlStdBasedAuthorization.java
```diff
@@ -78,6 +80,7 @@
 
   @BeforeClass
   public static void setup() throws Exception {
+    assumeTrue("Skipping tests since Hive supports only JDK 8.", HiveTestUtilities.supportedJavaVersion());
```
arina-ielchiieva: Can't we move check to `BaseTestHiveImpersonation` and extend from some common Hive test class?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/validate/BatchValidator.java
```diff
@@ -173,6 +173,10 @@ public BatchValidator(ErrorReporter errorReporter) {
   }
 
   public static boolean validate(RecordBatch batch) {
+    // This is a handy place to trace batches as they flow up
+    // the DAG. Works best for single-threaded runs with few records.
```
arina-ielchiieva:
```suggestion
    // the DAG. Works best for single-threaded runs with a few records.
```

arina-ielchiieva: Don't we have some print method for this case? So we don't need to make changes in code?

metastore/iceberg-metastore/src/test/java/org/apache/drill/metastore/iceberg/schema/TestIcebergTableSchema.java
```diff
@@ -238,63 +238,120 @@ void addFields(JDefinedClass jDefinedClass) {
...
+      return classWriter;
     }
 
-    abstract void addFields(JDefinedClass jDefinedClass);
+    abstract void addFields(ClassWriter classWriter);
+  }
 
-    private class ByteArrayStreamCodeWriter extends CodeWriter {
+  /**
+   * Helper class for constructing field type signature string.
```
arina-ielchiieva: Please add line breaks between messages so it looks better in java doc.

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveClusterTest.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.junit.Assume.assumeTrue;
+
+/**
+ * Base class for Hive cluster tests.
+ */
+public class HiveClusterTest extends ClusterTest {
+
+  @BeforeClass
+  public static void checkJavaVersion() {
+    assumeTrue("Skipping tests since Hive supports only JDK 8.", HiveTestUtilities.supportedJavaVersion());
```
arina-ielchiieva: Since others classes use this method it should be placed in utility class as before
`HiveTestUtilities.supportedJavaVersion()` but called from this class

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/validate/BatchValidator.java
arina-ielchiieva: So what the next steps?

contrib/storage-hive/core/src/test/java/org/apache/drill/exec/hive/HiveClusterTest.java
vvysotskyi: Agree, thanks, done.

arina-ielchiieva:
+1, LGTM.

PR #1936, opened 2019-12-18
DRILL-7490: LIMIT is not pushed to JDBC storage plugin

vvysotskyi:
See [DRILL-7490](https://issues.apache.org/jira/browse/DRILL-7490) for details.


arina-ielchiieva:
+1. LGTM

PR #1938, opened 2019-12-21
DRILL-7494: Unable to connect to Drill using JDBC driver when using custom authenticator

vvysotskyi:
- Removed excluding `com.fasterxml.woodstox:woodstox-core` since it is used in Hadoop classes, for example in the `Configuration` class.
-  Fixed relocation pattern for `com.fasterxml.woodstox:woodstox-core` classes, since there wasn't package `com.fasterxml.woodstox`.
- Replaced catching `Exception` with catching `Throwable` in `ConnectionMultiListener` since, for some cases, error may be thrown, for example, `NoClassDefFoundError` so it should be reported to the client correctly instead of hanging as it was before.
- Added code to patch Guava for JDBC driver since Hadoop uses some methods from the new version absent in 1.19.
- Added code to use the same class loader, which was used for loading patcher classes, for finding classes to patch instead of using system class loader since some clients use separate class loaders for loading and registering drivers.

Jira - [DRILL-7494](https://issues.apache.org/jira/browse/DRILL-7494).

### DIFF
--- a/common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
+++ b/common/src/main/java/org/apache/drill/common/util/GuavaPatcher.java
@@ -29,6 +29,8 @@ import javassist.CtClass;
 import javassist.CtConstructor;
 import javassist.CtMethod;
 import javassist.CtNewMethod;
+import javassist.scopedpool.ScopedClassPoolRepository;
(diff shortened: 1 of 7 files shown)


KazydubB:
LGTM, +1

PR #1939, opened 2019-12-24
DRILL-7497: Fix warnings when starting Drill on Windows using Java 11

vvysotskyi:
Jira - [DRILL-7497](https://issues.apache.org/jira/browse/DRILL-7497).

`%DRILL_SHELL_JAVA_OPTS%` is expanded when if block is parsed, so as a result, every next line of the if block will override the value set in the previous one.
Replaced it with a single assignment.


arina-ielchiieva:
+1, LGTM

PR #1941, opened 2019-12-24
DRILL-7495: Excel Reader Not Parsing Dates Correctly in First Column

cgivre:
This PR addresses a few minor bugs in the Excel Format Plugin in which dates were not being parsed correctly. 

Also included in this PR are: 
- A minor correction to the `README.md` document [](url)
- An update to the latest version of the Apache POI library. 

[Jira link](https://issues.apache.org/jira/browse/DRILL-7495)


contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -382,6 +382,26 @@ public void testExplicitSomeQueryWithCompressedFile() throws Exception {
     new RowSetComparison(expected).verifyAndClearAll(results);
   }
 
+  @Test
+  public void testFileWithDoubleDates() throws Exception {
+    String sql = "SELECT `Close Date`, `Type` FROM dfs.`excel/comps.xlsx` WHERE style='Contemporary'";
+
+    RowSet results = client.queryBuilder().sql(sql).rowSet();
+    results.print();
```
arina-ielchiieva: Please remove. Printing pollutes the output.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
```diff
@@ -382,6 +382,26 @@ public void testExplicitSomeQueryWithCompressedFile() throws Exception {
     new RowSetComparison(expected).verifyAndClearAll(results);
   }
 
+  @Test
+  public void testFileWithDoubleDates() throws Exception {
+    String sql = "SELECT `Close Date`, `Type` FROM dfs.`excel/comps.xlsx` WHERE style='Contemporary'";
```
arina-ielchiieva: Do we need to add new file rather than updating existing ones? Project size matters...

contrib/format-excel/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
arina-ielchiieva: Please remove we have common logback-test.xml for all tests.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: Removed.

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: I removed this and incorporated the data into the one test file.  However, the additional file was only 15kb.  I understand that we want to keep Drill as lightweight as possible, but does 15kb really make that much of a difference?

contrib/format-excel/src/test/resources/logback-test.xml
cgivre: Removed.

cgivre:
@arina-ielchiieva 
I addressed all review comments and squashed commits.  Thanks for the review and Happy Holidays!

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
arina-ielchiieva: If we’ll look at this from the perspective that everybody will start adding files, this will increase project size.
Generally, we should rather than adding new files:
1. reuse existing files;
2. generate files at runtime, for example for excel files this could have been done fairly easy.

arina-ielchiieva:
+1, happy holidays to you, too :)

contrib/format-excel/src/test/java/org/apache/drill/exec/store/excel/TestExcelFormat.java
cgivre: Ok.  Noted for future.  Thanks!

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -519,7 +519,10 @@ public void load(Cell cell) {
       if (cellValue == null) {
         columnWriter.setNull();
       } else {
-        Instant timeStamp = new Instant(cellValue.getNumberValue());
+        logger.debug("Cell value: {}", cellValue.getNumberValue());
+
+        Date dt = DateUtil.getJavaDate(cellValue.getNumberValue());
```
KazydubB: Is it possible to use either `java.util` or `org.joda.time` classes?

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
```diff
@@ -412,7 +413,7 @@ private void addColumnToArray(TupleWriter rowWriter, String name, TypeProtos.Min
 
   /**
    * Returns the index of the first row of actual data.  This function is to be used to find the header row as the POI skips blank rows.
-   * @return:  The headerRow index, corrected for blank rows
+   * @return int The headerRow index, corrected for blank rows
```
KazydubB: nit: there is no need to specify return type in `@return`, it looks awkward a little bit.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: That was actually the source of the bug.  I didn't realize it when I wrote the plugin but as it turns out, Excel generates all kinds of weird representations of dates.  Specifically in the case dates (no time) it generates a `double` and when it has a time, it generates a `long`.  Using the `DateUtil` class that is part of the POI library solves this problem and gets consistent dates out of Excel.

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
cgivre: Fixed

contrib/format-excel/src/main/java/org/apache/drill/exec/store/excel/ExcelBatchReader.java
KazydubB: Thanks for the explanation, I think we are good then.

KazydubB:
@cgivre, while running tests I've encountered the next test failure:

cgivre:
Which test failure?

KazydubB:
@cgivre, excuse me, posted accidentaly while not finished :)

arina-ielchiieva:
@cgivre travis does not run all the tests, CircleCI runs more tests but you have to enable it for your repo yourself, it is just in case if you need to run tests on the env different than your local.

cgivre:
@KazydubB Thanks for catching that.  I found one more minor error which I fixed that affected reading large files.

arina-ielchiieva:
Should be good to go now, please squash the commits.

cgivre:
Thanks.  Will do!

vvysotskyi:
@cgivre, in future please **do not merge** pull requests which have code-changes without checking that regression tests are passed.

I was preparing the branch with commits to merge (it would include this one), but since you have merged the commit, I have to remove this commit from my branch and rebase it one more time.

Let's agree that in future if you see that your PR is not merged for a couple of weeks after it is approved, please ask someone who able to run regression tests to do that, and only after merge it.

cgivre:
My apologies.   Won't do that again. 
-- C

arina-ielchiieva:
@cgivre since preparing batch commit branch takes time as well as running the tests we don't plan to do batch commits more often than once or twice a week so please be patient.

PR #1942, opened 2019-12-25
Update README.md with information on how to configure env for building site and update .gitignore to ignore files left after checkout from developing branches

vvysotskyi:


README.md
```diff
@@ -1,5 +1,17 @@
 The Apache Drill website is built using [Jekyll](http://jekyllrb.com/).
 
+# Configuring env
+1. Install `ruby`
+2. Install `bundler` and `jekyll` `v2.5.2`:
```
arina-ielchiieva: If we need specific version, please make an explicit note about that, warning that others versions won't work.

vvysotskyi: Thanks, good point, added this note after this list.

arina-ielchiieva:
+1

README.md
```diff
@@ -13,60 +27,60 @@ Note that you can skip the first two commands (and only run `jekyll serve`) if y
...
-          ---
-          title: "Configuring Multitenant Resources"
-          parent: "Configuring a Multitenant Cluster"
-          date: 
-          ---
-
+```
+chmod +x pre-commit
+```
+On any page you create, in addition to the title, and `parent:`, you now need to add `date:` to the front matter of any file you create. For example:
```
paul-rogers: A bit redundant: "On any page you create...of any file you create."

README.md
```diff
@@ -89,9 +104,9 @@ The updates should then be live: <http://drill.apache.org>.
 
 The documentation pages are under `_docs`. Most files, in the YAML front matter, have three important parameters:
 
-* title: - This is the title of the page enclosed in quotation marks. Each page must have a *unique* title
-* date: - This field is needed for Jekyll to write a last-modified date. Initially, leave this field blank.
-* parent - This is the title of the page's parent page. It should be empty for top-level sections/guides, and be identical to the title attribute of another page in all other cases.
+* `title:` - This is the title of the page enclosed in quotation marks. Each page must have a *unique* title
+* `date:` - This field is needed for Jekyll to write a last-modified date. Initially, leave this field blank.
+* `parent:` - This is the title of the page's parent page. It should be empty for top-level sections/guides, and be identical to the title attribute of another page in all other cases.
```
paul-rogers: Does it make sense to move the build stuff after this intro? I wondered what "title:" and so on were above. Then, when I get here, I now learn what they are. Would be great to reverse the order.

js/script.js
```diff
@@ -1,109 +1,109 @@
...
-	//	$("#menu.r").css({ top: ST+"px" });
-	//} else {
-	//	$("#menu").css({ top: "0px" });
-	//}
-
-	if (ST > 400) $("#subhead").addClass("show");
-	else $("#subhead").removeClass("show");
-}
+---
+---
```
paul-rogers: Why does Git think these lines changed? They look the same to me.

README.md
vvysotskyi: Thanks, removed redundant any.

README.md
vvysotskyi: Good point! Moved this section above `# Developing and Previewing the Website`

js/script.js
vvysotskyi: This file was with CRLF line endings, but exactly the same file in https://github.com/apache/drill-site is committed with LF line endings (perhaps due to git warnings it wasn't committed there), so I made LF line endings there also.

arina-ielchiieva:
Looks good, I think this can be merged.

PR #1945, opened 2019-12-31
DRILL-7502: Invalid codegen for typeof() with UNION

paul-rogers:
The gist of the problem is that codegen for a `UNION` using a `FieldReader` for a `@Param` was handled incorrectly. We had code to handle complex and repeated types specially. Turned out we just needed to add `UNION` type to that same logic. Since the fix turned out to be simple (after much sleuthing), I suspect this may have been a regression and this fix simply puts things back the way the original author intended.

Also fixes DRILL-6362: typeof() reports NULL for primitive columns with a NULL value.

typeof() is meant to return "NULL" if a UNION has a NULL value, but the column type when known, such as for non-UNION columns.

Also fixes DRILL-7499: sqltypeof() function with an array returns "ARRAY", not type. This was due to treating REPEATED like LIST.

Includes cleanup in files visited during the work.

### Documentation

User-visible behaviour changes after this fix:

* `typeof()` returns the value type of a specific column value. This is the Drill minor type for most types. For types other than `UNION`, if the value is `NULL`, `typeof() `still reports the column type (since the NULL has a type.) But for a `UNION` column, `typeof()` returns the type of that specific column value, or "NULL" if the column is null. (Since, for UNION, NULL has no type.)
* `sqlTypeOf()` works the same except for arrays. Previously (sometime between Drill 1.15 and Drill 1.17), `sqlTypeOf()` returned "ARRAY" for a repeated column. After this fix, it works as in Drill 1.14: `sqlTypeOf()` returns the type name. Thus a repeated INT returns "INT", not "ARRAY". (The `LIST` type still returns "ARRAY", however.)
* `drillTypeOf()` returns the actual minor type of a column, even for `UNION` and `DICT` columns. Thus, if a column is `UNION`, `typeOf()` returns the value type, but `drillTypeOf()` returns "UNION."

### Tests

Added two new tests to `TestTypeFns`, one which uses the JSON reader to verify `UNION` type behaviour (the test which uncovered this bug.) Reran all unit tests.


common/src/main/java/org/apache/drill/common/types/Types.java
```diff
@@ -906,4 +908,16 @@ public static boolean isNullable(final MajorType type) {
...
+
+  /**
+   * The number of minor types. Actually, the largest minor type ordinal.
+   * (There are holes in the ordering.) Update this if a new type
+   * is added. Use this value when allocating arrays to be indexed
+   * by minor type.
+   *
+   * @return the maximum minor type ordinal
+   */
+  public static final int typeCount() {
```
vvysotskyi: This method in the most places replaces `MinorType.values().length` occurrences, and for arrays, where it is used, are indexed using `Enum.ordinal()` method, so it should work correctly, and the existing code doesn't require updating every time when a new type is added.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/ClassGenerator.java
```diff
@@ -97,31 +98,37 @@
...
    * <p>
-   * For the upper estimation of number of split methods suppose that each expression in the method uses single variable.
-   * Suppose that the max number of indexes within the constant pull occupied by fields and local variables is M,
-   * the number of split methods is N, number of abstract methods in the template is A, then splitted methods count is
-   * N = (M - A * N * 3) / 50 => N = M / (50 + A * 3)
+   * For the upper estimation of number of split methods suppose that each
+   * expression in the method uses single variable. Suppose that the max number
+   * of indexes within the constant pull occupied by fields and local variables
+   * is M, the number of split methods is N, number of abstract methods in the
+   * template is A, then splitted methods count is N = (M - A * N * 3) / 50 => N
```
vvysotskyi: Please leave formulas to start from the new line.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
```diff
@@ -400,8 +403,10 @@ public HoldingContainer visitUnknown(LogicalExpression e, ClassGenerator<?> gene
      * seedValue0 .value = seedValue;
      * </p>
      *
-     * @param e parameter expression
-     * @param generator class generator
+     * @param e
```
vvysotskyi: Please revert this change, initial formatting looks better.

paul-rogers:
Am now seeing failures in the `GUnionFunctions` where a parameter is declared as `UnionHolder` but is now being created as `FieldReader`. Seems that the bug that I fixed may have been introduced to make these functions work, but had the side effect of forcing union parameters to be a `UnionReader` even if declared as a `FieldReader`. Looks like my change now does the opposite. So, more work to be done to make the decision based on the kind of field.

common/src/main/java/org/apache/drill/common/types/Types.java
vvysotskyi: I think there is an issue with misusing `getNumber()` field in places where `ordinal()` should be used. It should guarantee that we will use indexes less than `MinorType.values().length`.

exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
```diff
@@ -146,6 +135,19 @@ public boolean equals(Object obj) {
 
   Stack<Map<ExpressionHolder,HoldingContainer>> mapStack = new Stack<>();
 
+  public EvaluationVisitor() { }
+
+  public HoldingContainer addExpr(LogicalExpression e, ClassGenerator<?> generator) {
+
```
ihuzenko:
```suggestion
```

exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
```diff
@@ -117,9 +106,9 @@ public HoldingContainer addExpr(LogicalExpression e, ClassGenerator<?> generator
   }
 
   private class ExpressionHolder {
```
ihuzenko:
```suggestion
  private static class ExpressionHolder {
```

exec/java-exec/src/main/java/org/apache/drill/exec/expr/EvaluationVisitor.java
```diff
@@ -1282,11 +1285,9 @@ public HoldingContainer visitConvertExpression(ConvertExpression e, ClassGenerat
...
 
-
-
   private class ConstantFilter extends EvalVisitor {
 
-    private Set<LogicalExpression> constantBoundaries;
+    private final Set<LogicalExpression> constantBoundaries;
 
     public ConstantFilter(Set<LogicalExpression> constantBoundaries) {
       super();
```
ihuzenko:
```suggestion
```

exec/vector/src/main/codegen/templates/UnionReader.java
```diff
@@ -31,28 +34,33 @@
...
+  private BaseReader[] readers = new BaseReader[TYPE_COUNT];
   public UnionVector data;
-  
+
   public UnionReader(UnionVector data) {
     this.data = data;
   }
 
-  private static MajorType[] TYPES = new MajorType[45];
+  private static MajorType[] TYPES = new MajorType[TYPE_COUNT];
```
ihuzenko: can be final?

exec/vector/src/main/codegen/templates/AbstractFieldReader.java
```diff
@@ -28,11 +28,11 @@
 /*
  * This class is generated using freemarker and the ${.template_name} template.
  */
-@SuppressWarnings("unused")
 public abstract class AbstractFieldReader extends AbstractBaseReader implements FieldReader {
 
-  public AbstractFieldReader() {
-  }
+  public AbstractFieldReader() { }
```
ihuzenko:
```suggestion
```

common/src/main/java/org/apache/drill/common/types/Types.java
paul-rogers: @vvysotskyi, if I could e-mail you a beer, or even a pizza, I would do so. You uncovered a huge mess in the Union vector and reader. It is amazing it ever worked. Adding the `DICT` type pushed it over the edge. As it turns out, the code has long confused ordinals and Protobuf values. The problem is made worse because the Probuf-generated classes replace the `valueOf()` ordinal function with a lookup based on the Protobuf values. We had the two systems all mixed up.

So, I cleaned up all that, following your advice to use only the ordinal values. Added comments to help people avoid falling into this trap in the future.

Plus, I think this accidentally fixed another issue: DRILL-7510.

Not bad for a single review comment!

paul-rogers:
@vvysotskyi, @ihuzenko thank you for your reviews. I believe I've addressed your comments. Please take another look.

This commit contains some significant changes. As it turns out, we have three different ways we work with Union vectors in code gen. Two worked, the third did not. Figuring that out was a bit of an involved exercise. I added quite a few comments and split up some overly-large methods, to help gain understanding. I added a bit of a hack to work around the missing case. It works; all tests pass including the new one that initially failed. (And DRILL-7510 got magically fixed along the way.)

I say the fix is a "hack" because I had to add more state: to access the Union vector when setting up parameters that use a `FieldReader`. It seems there should be a simpler way; but I could not find it.

@KazydubB, these changes touched some of the same areas you apparently touched for your `DICT` work. Can you take a look to 1) make sure I didn't accidentally break anything, and 2) perhaps see the pattern I may be missing about how to set up a `FieldReader` for a Union vector.

paul-rogers:
Another thing to note is that we may want to go though one more round of sorting out the type functions. In normal SQL, there is just one level of type: a columns is an `INT` or a `DOUBLE`.

In Sqllite (IIRC) there is the `Variant` type: so a column can be `Variant` but the values can be `INT` or `DOUBLE`. The "variant" name comes from Visual Basic. In Drill, the variant is called `UNION`.

In Drill, we also have structured types: `ARRAY<INTEGER>` or `DICT<STRING,DOUBLE>`.

We've been trying to handle all these cases with originally one function (`typeof()`), then later three (adding `sqlTypeOf()` and `drillTypeOf()`).

We probably want four variations. Two forms of the type:

* The full type description: `ARRAY<ARRAY<MAP<a: INTEGER, b: VARCHAR>>>` say, or `UNION<DOUBLE, INTEGER>`.
* The short type of the vector itself: `LIST`, `UNION`.

For Variant types (`UNION`, `LIST`) we then want two forms:

* The vector itself. (`DICT`, `MAP`, `UNION`, `LIST`)
* The value: (`INTEGER`, `DOUBLE`).

Of course, the value can itself be structured: a `UNION` which has a `LIST` of `UNION` types. So, for both the vector and the value, we'd want both the simple and verbose forms.

And, now that @arina-ielchiieva added the nice type description system, we have no function which returns the full description, including the `NOT NULL` attribute, as it appears in the schema file: `INTEGER NOT NULL`.

The present fix gets us closer, but it is hard to solve four or five cases with three functions. I suspect we could do better. Suggestions welcomed!

common/src/main/java/org/apache/drill/common/types/Types.java
vvysotskyi: Thank you for finding and fixing all these issues! union data type was broken for a long time, and it is great that you will fix these code-gen issues.

exec/java-exec/src/main/java/org/apache/drill/exec/planner/sql/handlers/DefaultSqlHandler.java
```diff
@@ -324,7 +323,12 @@ protected DrillRel convertToDrel(RelNode relNode) throws SqlUnsupportedException
   /**
    * A shuttle designed to finalize all RelNodes.
    */
-  private static class PrelFinalizer extends RelShuttleImpl {
+  public static class PrelFinalizer extends RelShuttleImpl {
+
+    // The non-default constructor appears to be called from Calcite
+    // code when processing Hive-based queries.
+    public PrelFinalizer(DefaultSqlHandler handler) { }
```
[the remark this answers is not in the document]
paul-rogers: @vvysotskyi, not sure I can explain this. I got a runtime exception that some code could not create an instance. The error message appeared to indicate that the code was called via introspection, with the above signature. Adding the above code made the problem disappear.

I should have noted the test that failed when run in Eclipse.From the comment, it seemed to be from some of the Hive tests I ran. But, today, those tests run fine without this addition. So, I removed it.

vvysotskyi:
@paul-rogers, could you please squash the commits.

paul-rogers:
@vvysotskyi, thanks much for the review. Addressed the one comment I missed last time. Squashed commits.

PR #1946, opened 2019-12-31
DRILL-7505: PCAP Plugin Fails on IPv6 Packets

cgivre:
In its current implementation, the PCAP parser fails with an exception if it encounters a protocol it does not have in its enumerated list. 
This is not an acceptable strategy in that there are many protocols out there and having Drill crash when it encounters an unknown protocol is not helpful or useful.  This minor PR makes the PCAP plugin more fault tolerant.

I tested this on proprietary PCAP files which are too large to include.  Existing unit tests are unaffected.

Jira: [Drill-7505](https://issues.apache.org/jira/browse/DRILL-7505)

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/PacketConstants.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/PacketConstants.java
@@ -41,6 +41,9 @@ public final class PacketConstants {
   public static final int AUTHENTICATION_V6 = 51;
   public static final int ENCAPSULATING_SECURITY_V6 = 50;
   public static final int MOBILITY_EXTENSION_V6 = 135;
+  public static final int HOST_IDENTITY_PROTOCOL = 139;
+  public static final int SHIM6_PROTOCOL = 140;
+
   public static final int NO_NEXT_HEADER = 59;
   public static final int UDP_HEADER_LENGTH = 8;
   public static final int VER_IHL_OFFSET = 14;
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/store/pcapng/decoder/PacketDecoder.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/store/pcapng/decoder/PacketDecoder.java
@@ -54,7 +54,7 @@ public class PacketDecoder extends Packet {
   protected int processIpV6Packet() {
     try {
       return super.processIpV6Packet();
-    } catch (IllegalStateException ise) {
+    } catch (IllegalStateException | ArrayIndexOutOfBoundsException e) {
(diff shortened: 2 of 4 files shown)


exec/java-exec/src/main/java/org/apache/drill/exec/store/pcapng/decoder/PacketDecoder.java
```diff
@@ -56,6 +56,8 @@ protected int processIpV6Packet() {
       return super.processIpV6Packet();
     } catch (IllegalStateException ise) {
       return -1;
+    } catch (ArrayIndexOutOfBoundsException e) {
```
arina-ielchiieva: Use multi-catch

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/PacketConstants.java
```diff
@@ -40,7 +40,14 @@
   public static final int FRAGMENT_V6 = 44;
   public static final int AUTHENTICATION_V6 = 51;
   public static final int ENCAPSULATING_SECURITY_V6 = 50;
+  public static final int ICMPv6 = 58;
   public static final int MOBILITY_EXTENSION_V6 = 135;
+  public static final int HOST_IDENTITY_PROTOCOL = 139;
+  public static final int SHIM6_PROTOCOL = 140;
+
+  public static final int RESERVED_IPV6_1 = 253;
```
arina-ielchiieva: Please remove unused constants.

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcapng/decoder/PacketDecoder.java
cgivre: Fixed

exec/java-exec/src/main/java/org/apache/drill/exec/store/pcap/decoder/PacketConstants.java
cgivre: Fixed

cgivre:
After looking at the PCAP and PCAP-NG plugins, both of these really do need some work. I'm going to submit some PRs in the next few weeks for some code cleanup and general improvements.  This is in support of a use case at my company, but in general, I'd like to consolidate the PCAP and PCAP-NG plugins into one so that improvements don't have to be duplicated across both plugins.  But... that's not for this PR.

arina-ielchiieva:

cgivre:
Thanks @arina-ielchiieva for the review.  Commits squashed.  Could we please merge this today as this is a blocker for a POC I'm working on?
Thanks and Happy New Year!

PR #1948, opened 2020-01-02
DRILL-7506: Simplify code gen error handling

paul-rogers:
Revises error handling in code generation to push exceptions closer to the source of the error to both provide clearer errors and simplify code.

## Background

Drill evolved to have 2 1/2 ways to report errors. This PR is a step towards simplifying the execution engine to use just one form of error reporting.

The original design of the "Volcano iterator" envisioned a `STOP` status that indicated that an error occurred. When an operator encountered and error, it would:

* Log the error.
* Pass an exception describing the error to the fragment context, telling the fragment context that the fragment has failed.
* Pass `STOP` up the iterator call stack to stop execution.

Different operators implemented this in different ways. Some performed `close()` operations at the time the error was found. Others put themselves into a special "stop" state. Some call `kill()` on incoming batches others do not.

Even from the beginning, the fragment executor had to handle runtime exceptions such as NPE, OOM and so on. As a result, some operators adopted a second way to report errors: simply throw a runtime (unchecked) exception.

At the time that the managed sort was created, we put in effort to ensure that the fragment executor properly closes all operators in a fragment when a runtime exception was thrown. The years since have shown that this solution works.

We also resolved many cases in which it was not clear when `close()` should be called. When the exception is thrown (or `STOP`) is returned? In the `kill()` call? Only by the fragment executor? It turned out that the only reliable solution was for the fragment executor to call `close()`, from the leaf-most to the root-most operators.

The "2 1/2th" solution is that code *should* throw a `UserException` that contains information useful to the user to describe the error. Some operators use `UserException`, others throw other unchecked exceptions.

This PR moves toward using `UserException` as the one reliable way to report errors. This PR focuses on code generation. It pushes code gen error handling close to the code gen itself to allow clearer error messages. Doing so avoids the need to bubble code gen exceptions up the call stack, resulting in cleaner operator code.

## Tests

All unit tests were rerun to ensure there is no behavior change from this PR. A couple of tests were modified to account for the use of a standardized `UserException` instead of the various other exceptions thrown previously.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/PhysicalOperatorUtil.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/PhysicalOperatorUtil.java
@@ -17,6 +17,7 @@
  */
 package org.apache.drill.exec.physical.base;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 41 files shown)


paul-rogers:
Rebased on latest master to resolve conflicts.

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/LateralJoinBatch.java
```diff
@@ -531,13 +523,10 @@ private IterOutcome processRightBatch() {
...
           // produced
-          if (handleSchemaChange()) {
-            container.setRecordCount(0);
-            rightJoinIndex = (right.getRecordCount() > 0) ? 0 : -1;
-            return OK_NEW_SCHEMA;
-          } else {
-            return STOP;
-          }
+          handleSchemaChange();
+          container.setRecordCount(0);
```
KazydubB: Does `container#setEmpty()` work here, or `setRecordCount(0)` should be used?

exec/java-exec/src/main/java/org/apache/drill/exec/physical/base/PhysicalOperatorUtil.java
```diff
@@ -46,10 +46,13 @@ private PhysicalOperatorUtil() {}
   }
 
   /**
-   * Helper method to create a list of MinorFragmentEndpoint instances from a given endpoint assignment list.
+   * Helper method to create a list of MinorFragmentEndpoint instances from a
+   * given endpoint assignment list.
    *
-   * @param endpoints Assigned endpoint list. Index of each endpoint in list indicates the MinorFragmentId of the
-   *                  fragment that is assigned to the endpoint.
+   * @param endpoints
```
KazydubB: nit: no need to start from a new line.

Can you add Java docs to `@return`, please? :)

exec/java-exec/src/main/java/org/apache/drill/exec/physical/impl/join/LateralJoinBatch.java
```diff
@@ -491,6 +482,7 @@ private IterOutcome processLeftBatch() {
               "received NOT_YET");
           }
           break;
+        default:
```
KazydubB: nit: here and below: can you indicate `fall through` with a comment in `default` cases, please?

paul-rogers:
@KazydubB, thank you for the review. Addressed comments. Squashed commits. Rebased on latest master and resolved conflicts.

PR #1949, opened 2020-01-02
DRILL-7507: Convert fragment interrupts to exceptions

paul-rogers:
Modifies fragment interrupt handling to throw a specialized exception, rather than relying on the complex and cumbersome STOP iterator status.

## Description

PR #1948 (DRILL-7506) explains how Drill's execution engine has two error reporting systems. The execution engine has a mechanism to tell operators that the fragment has been cancelled. Today that system uses the `STOP`-based error system. This PR converts the cancellation mechanism to use the exception-based system.

Basically, an operator periodically checks if the fragment should terminate. Rather than return a `STOP` status, the fragment now throws a specialized exception. This check logic is encapsulated in a new `checkContinue()` method.

The fragment executor catches the exception, notices it is the special cancellation exception that it, itself, requested, and proceeds to shut down the fragment as if it completed with a `STOP` status (or with a runtime exception.)

## Tests

Reran all unit tests, including `TestDrillbitResiliance` which checks for query cancellations.

### DIFF
--- a/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
+++ b/exec/java-exec/src/main/java/org/apache/drill/exec/ops/FragmentContext.java
@@ -213,13 +213,34 @@ public interface FragmentContext extends UdfUtilities, AutoCloseable {
 
   interface ExecutorState {
     /**
-     * Tells individual operations whether they should continue. In some cases, an external event (typically cancellation)
-     * will mean that the fragment should prematurely exit execution. Long running operations should check this every so
-     * often so that Drill is responsive to cancellation operations.
+     * Tells individual operations whether they should continue. In some cases,
+     * an external event (typically cancellation) will mean that the fragment
+     * should prematurely exit execution. Long running operations should check
+     * this every so often so that Drill is responsive to cancellation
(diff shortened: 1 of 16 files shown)


arina-ielchiieva:
+1, LGTM

KazydubB:
@paul-rogers, there are conflicts after https://github.com/apache/drill/pull/1949 was merged. Can you resolve it, please?

paul-rogers:
@arina-ielchiieva, thanks much for the quick review. Rebased on the latest master which removed the old external sort. The change to the "managed" sort (not the only external sort) was just a bit different this time than the previous one as we'd done a bit of refactoring during the PR for that one.

arina-ielchiieva:
+1, @paul-rogers please squash the commits.
P.S. There is no need to leave rebase commit, I think reviewer does not need to spend time reviewing rebase and then asking for commits to be squashed.

paul-rogers:
@arina-ielchiieva, thanks much for your update. I left the rebased commit unsquashed previously because I had to redo the external sort changes. Gave you a chance to verify that rework. Anyway, commits squashed so we're good to go.

